PR #6967, opened 2025-09-26
feat(pocket-ic): prune the result of the previous computation

venkkatesh-sekar:


rs/pocket_ic_server/src/state_api/state.rs
```diff
@@ -1065,6 +1073,17 @@ impl ApiState {
...
+                            // Prune the graph if needed
+                            let graph_size = graph_guard.len();
+                            if graph_size > MAX_GRAPH_SIZE {
+                                let mut purge_state = old_state_label;
+                                purge_state.reduce_by(MAX_GRAPH_SIZE);
+                                for _ in 0..graph_size - (MAX_GRAPH_SIZE + 1) {
+                                    graph_guard.remove(&purge_state);
+                                    purge_state.reduce_by(1);
+                                }
+                            }
```
mraszyk:
```suggestion
                            // Prune the result of the previous computation.
                            graph_guard.remove(old_state_label);
```
that's a bit more aggressive, but simpler

mraszyk: for the sake of concurrent operations, we can indeed preserve the last N entries for every instance (then every `tokio::test` could run up to `N` concurrent operations), but that should be done in a different way (since the size of the graph covers all results across all instances, which could lead to non-deterministic test results).

venkkatesh-sekar:
Closed in favor of #6988

PR #6968, opened 2025-09-26
perf(idkg): CON-1593 Use multi-threading in `send_dealing_support`

eichhorl:
This PR makes use of the IDKG thread pool to create IDKG support shares in parallel. We iterate over all validated dealings using multiple threads, and attempt to create a support share for each of them, if we haven't done so already.

In addition, the following change was made for performance reasons: Previously we would resolve the parameter reference first, and then check if we are a receiver of the given dealing. However, this can also be determined based on the parameter reference itself, without resolving it first, in which case the support creation fails sooner.

### DIFF
--- a/rs/tests/consensus/tecdsa/tecdsa_performance_test_template.rs
+++ b/rs/tests/consensus/tecdsa/tecdsa_performance_test_template.rs
@@ -111,7 +111,7 @@ const NETWORK_SIMULATION: FixedNetworkSimulation = FixedNetworkSimulation::new()
 const PRE_SIGNATURES_TO_CREATE: u32 = 30;
 const MAX_QUEUE_SIZE: u32 = 10;
 const CANISTER_COUNT: usize = 4;
-const SIGNATURE_REQUESTS_PER_SECOND: f64 = 3.5;
+const SIGNATURE_REQUESTS_PER_SECOND: f64 = 4.5;
 
 const SMALL_MSG_SIZE_BYTES: usize = 32;
 #[allow(dead_code)]


rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -272,86 +272,91 @@ impl IDkgPreSignerImpl {
...
+                                idkg_pool,
+                                &signed_dealing.idkg_dealing().transcript_id,
+                                &signed_dealing.dealer_id(),
+                                &self.node_id,
+                                &dealing_hash,
+                            )
+                        },
+                    )
+                })
+                .filter_map(|(id, signed_dealing)| {
```
pierugo-dfinity, line 280: Style: Would it make sense to embed the above `filter` into one single big `filter_map`? You embedded two `filter_map`s into the same one already.

In that case, I guess we should also do the same in `send_dealings`...

eichhorl: Good idea, done! I'll leave it as is in `send_dealings`, because `crypto_create_dealing` actually returns a `Vec`, so we need a `flat_map` there.

pierugo-dfinity: Nice, looks better imo!

Concerning `send_dealings`, we could still do so by using one big `flat_map` and returning an empty `Vec` instead of `false`/`None`.

eichhorl: I think returning empty `Vec`s might be a bit confusing, and it would introduce some `let ... else {}` constructs so I think I prefer the current version.

Combining two `filter_map`s into one (like we did here) makes sense to me because they have the same function signature. Also getting rid of the `filter` was a good idea because the `map_or_else` construct was very unreadable. Generally I think chaining the specialized iterators conveys the intend a bit better, but we can also revisit later.

PR #6842, opened 2025-09-19
perf(idkg): CON-1593 Use multi-threading in `send_dealings`

eichhorl:
This PR enables multi-threading in IDKG, by creating a rayon thread pool with a fixed number of threads at instantiation time.

For now, we only apply multi-threading during the creation of new IDKG dealings, using a conservative thread pool size of 8.

PR #6970, opened 2025-09-26
perf(idkg): CON-1593 Use multi-threading in `ThresholdSignerImpl`

eichhorl:
This PR makes use of the IDKG thread pool to create and validate ECDSA/Schnorr/VetKD shares in parallel.

During creation, we iterate over all open request contexts in parallel, and attempt to create a share for each of them, if we haven't done so already.

During validation, we iterate over all unvalidated shares in parallel, and attempt to validate each of them. If there are multiple valid shares for a request by the same signer in the same unvalidated batch, they are filtered out after the parallel iteration.


rs/consensus/idkg/src/signer.rs
```diff
@@ -163,28 +170,40 @@ impl ThresholdSignerImpl {
...
-                            err
-                        );
-                        self.metrics.sign_errors_inc("signature_inputs_malformed");
-                    }
+        self.thread_pool.install(|| {
+            state_snapshot
+                .get_state()
+                .signature_request_contexts()
+                .into_par_iter()
+                .flat_map(|(id, context)| {
```
pierugo-dfinity: [Similarly](https://github.com/dfinity/ic/pull/6842#discussion_r2381097183), I think it would be clearer to use `filter_map` than `flat_map`?

rs/consensus/idkg/src/signer.rs
```diff
@@ -163,28 +170,40 @@ impl ThresholdSignerImpl {
...
                 })
-            })
-            .filter(|(request_id, inputs)| {
-                !self.signer_has_issued_share(idkg_pool, &self.node_id, request_id, inputs.scheme())
-            })
-            .flat_map(|(request_id, sig_inputs)| {
-                self.create_signature_share(idkg_pool, transcript_loader, request_id, sig_inputs)
-            })
-            .collect()
+                .filter(|(request_id, inputs)| {
```
pierugo-dfinity, line 190: Style preference: maybe consider combining the above `flat/filter_map` with this `filter` like you did in the previous PRs?

And maybe also with the `flat_map` below, if we do the same in `send_dealings` (see [comment](https://github.com/dfinity/ic/pull/6968#discussion_r2394794933))

rs/consensus/idkg/src/signer.rs
```diff
@@ -209,40 +228,48 @@ impl ThresholdSignerImpl {
...
+                .filter_map(|(id, share)| {
+                    match Action::new(
+                        &sig_inputs_map,
+                        &share.request_id(),
+                        state_snapshot.get_height(),
+                    ) {
+                        Action::Process(sig_inputs) => {
+                            let key = (share.request_id(), share.signer());
+                            self.validate_signature_share(idkg_pool, id.clone(), share, sig_inputs)
+                                .map(|action| (Some((id, key)), action))
```
pierugo-dfinity: Again style comment.

Reading through the code, one understands that we keep track of this `(id, key)` tuple only below when used to filter out duplicate validated shares. I preferred what you did in [`IDkgPreSignerImpl::validate_dealings`](https://github.com/dfinity/ic/blob/master/rs/consensus/idkg/src/pre_signer.rs#L240-L256) where you do the parallel iteration and then independently introduce `validated_dealings` and do a second processing filtering out duplicate validated dealings. I understand that we cannot do the exact same thing, i.e. compute the `key` variable solely from the output of `self.validate_signature_share` but maybe something like the following could work?

```rust
        let results: Vec<_> = self.thread_pool.install(|| {
            // Iterate over all signature shares of all schemes
            shares
                .into_par_iter()
                .map(|(id, share)| {
                    match Action::new(
                        &sig_inputs_map,
                        &share.request_id(),
                        state_snapshot.get_height(),
                    ) {
                        Action::Process(sig_inputs) => self.validate_signature_share(idkg_pool, id.clone(), share, sig_inputs)
                        Action::Drop => Some(IDkgChangeAction::RemoveUnvalidated(id)),
                        Action::Defer => None,
                    }
                })
                .collect()
        });

        let mut ret = Vec::new();
        // Collection of validated shares
        let mut validated_sig_shares = BTreeSet::new();
        for (share, maybe_action) in shares.iter().zip(results) {
            let Some(action) = maybe_action else {
                continue;
            }

            if let IDkgChangeAction::MoveToValidated(msg) = &action {
                let key = (share.request_id(), share.signer());
                if !validated_sig_shares.insert(key) {
                    self.metrics
                        .sign_errors_inc("duplicate_sig_shares_in_batch");
                    ret.push(IDkgChangeAction::HandleInvalid(
                        id,
                        format!("Duplicate share in unvalidated batch: {msg:?}"),
                    ));
                    continue;
                }
            }
            ret.push(action);
        }

        ret
    }
```

Having it written down in full, I'm now actually not 100% sure if it's clearer than your implementation. But at least it's somewhat consistent with `IDkgPreSignerImpl::validate_dealings` and "separates the responsibility" of the two iterations.

Performance-wise, I guess the second iteration would be over a potentially longer collection (across all shares, instead of filtered out ones) but to be fair, this overhead is probably done by rayon in your current implementation to aggregate the results of the multiple `filter_map`s into a single collection.

Sawchord (approved):

rs/consensus/idkg/src/signer.rs
Sawchord: +1

eichhorl: Done 👍  `build_signature_inputs` returns a result, so I also had to call `.ok()`

rs/consensus/idkg/src/signer.rs
eichhorl: Thanks for this comment, I agree it isn't very readable. I don't think zipping the shares with the results would work, because `into_par_iter()` consumes the shares. Also, if some shares are deferred, then the `results` vector is shorter than `shares`. I tried something else instead, PTAL

pierugo-dfinity: > Also, if some shares are deferred, then the results vector is shorter than shares.

Indeed, that's why I had replaced `shares.into_par_iter().flat_map()` with `shares.into_par_iter().map()`, which also justifies the 
```rust
let Some(action) = maybe_action else {
    continue;
}
```
below.

> because `into_par_iter()` consumes the shares

But that argument alone indeed invalidates my suggestion.

I like what you did instead, thank you!

rs/consensus/idkg/src/signer.rs
eichhorl: I'll leave it as is for the same reason as in my other comment [here](https://github.com/dfinity/ic/pull/6968#discussion_r2402722656), I think chaining the specialized iterators is a bit more idiomatic

PR #6971, opened 2025-09-26
feat(nns): add backend functions for neuron index

ShahriarJavidi:
This PR adds the backend functions required for fetching all neurons/neuron IDs.

### DIFF
--- a/rs/nns/governance/src/governance/tests/get_neuron_index.rs
+++ b/rs/nns/governance/src/governance/tests/get_neuron_index.rs
@@ -41,7 +41,7 @@ fn make_governance_for_neuron_index() -> Governance {
         .with_controller(PrincipalId::new_user_test_id(id))
         .build();
 
-        governance.add_neuron(neuron.id().id, neuron).unwrap();
+        governance.add_neuron(id, neuron).unwrap();
     }
 
     governance
@@ -50,6 +50,7 @@ fn make_governance_for_neuron_index() -> Governance {
 // This user doesn't control any neurons.
 static NEURONLESS_USER_ID: PrincipalId = PrincipalId::new_user_test_id(1001);
 
+#[track_caller]
 fn assert_neuron_index(neurons: &[NeuronInfo], start_id: u64, size: usize) {
     assert_eq!(neurons.len(), size);
     for (i, neuron) in neurons.iter().enumerate() {
@@ -67,12 +68,12 @@ fn assert_neuron_index(neurons: &[NeuronInfo], start_id: u64, size: usize) {
 fn test_get_neuron_index_empty_args() {
     let _temp = temporarily_enable_comprehensive_neuron_list();
 
+    let governance = make_governance_for_neuron_index();
+
     let request = GetNeuronIndexRequest {
         exclusive_start_neuron_id: None,
         page_size: None,
     };
-
-    let governance = make_governance_for_neuron_index();
 
     let response = governance
         .get_neuron_index(request, NEURONLESS_USER_ID)
@@ -85,12 +86,12 @@ fn test_get_neuron_index_empty_args() {
 fn test_get_neuron_index_large_page() {
     let _temp = temporarily_enable_comprehensive_neuron_list();
 
+    let governance = make_governance_for_neuron_index();
+
     let request = GetNeuronIndexRequest {
         exclusive_start_neuron_id: None,
         page_size: Some(500),
     };
-
-    let governance = make_governance_for_neuron_index();
 
     let response = governance
         .get_neuron_index(request, NEURONLESS_USER_ID)
@@ -103,12 +104,12 @@ fn test_get_neuron_index_large_page() {
 fn test_get_neuron_index_multiple_pages() {
     let _temp = temporarily_enable_comprehensive_neuron_list();
 
+    let governance = make_governance_for_neuron_index();
+
     let request = GetNeuronIndexRequest {
         exclusive_start_neuron_id: Some(NeuronId { id: 0 }),
         page_size: Some(MAX_NEURON_PAGE_SIZE),
     };
-
-    let governance = make_governance_for_neuron_index();
 
     let response = governance
         .get_neuron_index(request, NEURONLESS_USER_ID)
@@ -134,12 +135,12 @@ fn test_get_neuron_index_multiple_pages() {
 fn test_get_neuron_index_disabled() {
     let _temp = temporarily_disable_comprehensive_neuron_list();
 
+    let governance = make_governance_for_neuron_index();
+
     let request = GetNeuronIndexRequest {
         exclusive_start_neuron_id: Some(NeuronId { id: 0 }),
         page_size: Some(MAX_NEURON_PAGE_SIZE),
     };
-
-    let governance = make_governance_for_neuron_index();
 
     let response = governance.get_neuron_index(request, NEURONLESS_USER_ID);
 


rs/nns/governance/src/neuron_store.rs
```diff
@@ -532,6 +539,30 @@ impl NeuronStore {
...
+        with_stable_neuron_store(|stable_store| {
+            stable_store
+                .range_neurons((Bound::Excluded(exclusive_start_id), Bound::Unbounded))
+                .take(page_size as usize)
+                .map(|neuron| {
+                    neuron.get_neuron_info(
+                        voting_power_economics,
+                        now_seconds,
+                        requester,
+                        false, // TODO double-check
```
ShahriarJavidi: Should it be true or false??

rs/nns/governance/canister/canister.rs
```diff
@@ -378,9 +378,11 @@ fn list_neurons(req: ListNeurons) -> ListNeuronsResponse {
 }
 
 #[query]
-fn get_neuron_index(_req: GetNeuronIndexRequest) -> Result<NeuronIndexData, GovernanceError> {
+fn get_neuron_index(req: GetNeuronIndexRequest) -> Result<NeuronIndexData, GovernanceError> {
     debug_log("get_neuron_index");
-    Ok(NeuronIndexData::default())
+    governance()
+        .get_neuron_index(req, caller())
+        .map_err(GovernanceError::from)
```
daniel-wong-dfinity-org, line 385: I think you can delete this line. Is it needed for reasons that I'm not seeing??

rs/nns/governance/src/governance.rs
```diff
@@ -3515,6 +3517,39 @@ impl Governance {
...
+    /// The list has an exclusive lower bound on the neuron id, and a maximum
+    /// number of neurons to be returned.
+    /// The number of neurons to be returned is capped by `MAX_NEURON_PAGE_SIZE`.
+    pub fn get_neuron_index(
+        &self,
+        req: GetNeuronIndexRequest,
+        requester: PrincipalId,
+    ) -> Result<NeuronIndexData, GovernanceError> {
+        if !is_comprehensive_neuron_list_enabled() {
+            return Ok(NeuronIndexData { neurons: vec![] });
```
daniel-wong-dfinity-org: Wouldn't it make more sense to return Err instead of Ok??

rs/nns/governance/src/governance.rs
```diff
@@ -3515,6 +3517,39 @@ impl Governance {
...
+        }
+
+        let now = self.env.now();
+
+        let exclusive_start_id = req.exclusive_start_neuron_id.unwrap_or(NeuronId { id: 0 });
```
daniel-wong-dfinity-org:
```suggestion
        let exclusive_start_neuron_id = req.exclusive_start_neuron_id.unwrap_or(NeuronId { id: 0 });
```

This doesn't really matter very much, ofc. But I think you'll find that "unnecessary" inconsistencies increase ambiguity. When there is such a direct correspondence, I think it helps people to use a name that is as close as possible. Exact match is best match. Inexact matches just raise doubts.

rs/nns/governance/src/governance.rs
```diff
@@ -3515,6 +3517,39 @@ impl Governance {
...
+
+        let page_size = req
+            .page_size
+            .unwrap_or(MAX_NEURON_PAGE_SIZE)
+            .min(MAX_NEURON_PAGE_SIZE);
```
daniel-wong-dfinity-org, line 3543: Consider using [clamp] instead. I think it more directly indicates the intent: to make sure that `page_size` does not exceed `MAX_NEURON_PAGE_SIZE`.

Again, I do not think this is so important. Most ppl know that `min` is very frequently used to clamp from above. It's just a little bit confusing, because min is associated with the least that a value can be, but what you are doing is saying that `page_size` will be at most MAX. Do as you see fit.

[clamp]: https://doc.rust-lang.org/std/primitive.u64.html#method.clamp

rs/nns/governance/src/governance.rs
```diff
@@ -3515,6 +3517,39 @@ impl Governance {
...
+
+        let neurons = self.neuron_store.list_all_neurons_paginated(
+            exclusive_start_id,
+            page_size,
+            requester,
+            now,
+            self.voting_power_economics(),
```
daniel-wong-dfinity-org, line 3550: Hmmm. I guess this is used to calculate voting power and such. It is a little surprising that this is necessary, but I do not have any better ideas for what to do here.

rs/nns/governance/src/governance.rs
```diff
@@ -3515,6 +3517,39 @@ impl Governance {
...
+        );
+
+        Ok(NeuronIndexData { neurons })
+    }
```
daniel-wong-dfinity-org, line 3554: I really like how you have done this. I think many ppl would try to inline everything. Whereas, all this does is

1. do some validation (i.e. is this even enabled?)
2. unpack the request
3. call the "backend" (i.e. NeuronStore) to do the "real" work, and
4. assemble the result.

That is plenty of work for one function, but I think ppl believe that it's not enough. No! All of those things, even though they are not "interesting", are in fact separate discrete steps in the whole general workflow of getting anything done. Sloughing the "real" work to the backend is very sound.

rs/nns/governance/src/storage/neurons.rs
```diff
@@ -439,7 +439,6 @@ where
         self.main.len().min(usize::MAX as u64) as usize
     }
 
-    #[cfg(any(test, feature = "canbench-rs", feature = "tla"))]
```
daniel-wong-dfinity-org, line 442: Hmm. I don't know why this was unavailable before. Seems like a pretty "essential" primitive.

No objections to this change.

rs/nns/governance/src/neuron_store.rs
```diff
@@ -531,6 +538,30 @@ impl NeuronStore {
...
+        exclusive_start_id: NeuronId,
+        page_size: u32,
+        requester: PrincipalId,
+        now_seconds: u64,
+        voting_power_economics: &VotingPowerEconomics,
+    ) -> Vec<NeuronInfo> {
+        with_stable_neuron_store(|stable_store| {
+            stable_store
+                .range_neurons((Bound::Excluded(exclusive_start_id), Bound::Unbounded))
+                .take(page_size as usize)
```
daniel-wong-dfinity-org, line 552: Elegant! This is exactly what I was envisioning. This is, ofc, where the "real" work is happening. This is the kernel of this whole change, and it is just two lines! Everything else is just connecting the right output port over there to the right input port over here. For better or worse, to a large extent, that is what programming consists of. Nevertheless, well laid out pipes in the walls is a beautiful thing.

rs/nns/governance/src/neuron_store.rs
daniel-wong-dfinity-org: I agree that this parameter is confusing. WTF does multi_query mean??? Well, looking at how it is actually used in the body, it seems to just control whether the result includes known_neuron_data.

Well, I guess I would rather not include that, because it makes the response bigger. OTOH, people might get confused if it is omitted.

Well, since this is a new API, we would not be "breaking" anyone by omitting it, so I guess we could omit if we really wanted to. I guess that is what you were thinking here? It would just be confusing, because the .did file seems to indicate that known_neuron_data would be populated (when it exists).

I guess if you want to omit it (by passing `false`) the documentation should specially call out that this field (even though the field exists) will not be populated in the case of get_neuron_index...

Let's discuss...

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+use crate::{
+    temporarily_disable_comprehensive_neuron_list, temporarily_enable_comprehensive_neuron_list,
+};
+use ic_base_types::PrincipalId;
+use ic_nns_common::pb::v1::NeuronId;
+use ic_nns_governance_api::GetNeuronIndexRequest;
+use ic_nns_governance_api::{Governance as ApiGovernance, NetworkEconomics};
+use std::sync::Arc;
+
+fn make_governance_for_neuron_index() -> Governance {
```
daniel-wong-dfinity-org, line 18: Consider doing this instead:

```suggestion
lazy_static! {
    static ref GOVERNANCE: Governance = {
```

This makes it clearer that all tests use the same starting value. I mean, a 0-argument function is already pretty clear, but a `lazy_static!` pseudo-constant makes it even clearer.

Do as you see fit.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+                id,
+                DissolveStateAndAge::NotDissolving {
+                    dissolve_delay_seconds: 100,
+                    aging_since_timestamp_seconds: 0,
+                },
+            )
+            .with_controller(PrincipalId::new_user_test_id(id))
+            .build()
+        })
+        .collect::<Vec<_>>();
```
daniel-wong-dfinity-org: Merge this into the `for` loop after `let mut governance`. Something like

```
for id in 1..=500_u64 {
```

Separate the `u64` from the digits with `_`.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+        Arc::new(MockEnvironment::new(Default::default(), 0)),
+        Arc::new(StubIcpLedger {}),
+        Arc::new(StubCMC {}),
+        Box::new(MockRandomness::new()),
+    );
+
+    for neuron in neurons {
+        governance.add_neuron(neuron.id().id, neuron).unwrap();
+    }
+    governance
```
daniel-wong-dfinity-org, line 47:
```suggestion

    governance
```

Add blank line. This esp makes sense with my other suggestion to merge earlier neurons code into here. But even without that, I would seriously consider having this blank line.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[test]
+fn test_get_neuron_index_empty_args() {
+    let _temp = temporarily_enable_comprehensive_neuron_list();
+    // This user doesn't control any neurons.
+    let user_id = PrincipalId::new_user_test_id(1001);
```
daniel-wong-dfinity-org:
```suggestion
    let neuronless_user_id = PrincipalId::new_user_test_id(1001);
```

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+    let request = GetNeuronIndexRequest {
+        exclusive_start_neuron_id: None,
+        page_size: None,
+    };
+
+    let response = governance
+        .get_neuron_index(request, user_id)
+        .expect("Expected valid GetNeuronIndexResponse");
+
+    assert_eq!(response.neurons.len(), MAX_NEURON_PAGE_SIZE as usize);
```
daniel-wong-dfinity-org: Can we have a stronger assert here? E.g.

```
assert_eq!(
    response.neurons.map(|neuron| {
        (neuron.id, neuron.controller, neuron.dissolve_delay_seconds)
    )
    .collect::<Vec<_>>(),
    (1..300_u64)
        .map(...)
        .collect(),
)
```

We should at least partially inspect the ACTUAL CONTENT, not just the cardinality! I know many of the fields are not super easy to calculate (esp voting power), but id, controller, dissolve_delay_seconds is not difficult.

Ditto elsewhere.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,171 @@
...
+    let request = GetNeuronIndexRequest {
+        exclusive_start_neuron_id: Some(NeuronId { id: 300 }),
+        page_size: Some(MAX_NEURON_PAGE_SIZE),
+    };
+
+    let response = governance
+        .get_neuron_index(request, user_id)
+        .expect("Expected valid GetNeuronIndexResponse");
+
+    assert_eq!(response.neurons.len(), 200);
```
daniel-wong-dfinity-org: Maybe explain in comment that 200 is expected because governance has 500 neurons, and the request skipped the first 300.

rs/nns/governance/canister/canister.rs
ShahriarJavidi: Yes. One in `GovernanceError` (`rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs`), the other one is from API `rs/nns/governance/api/src/types.rs`.
Same pattern can be seen for e.g., `get_metrics`

rs/nns/governance/src/governance.rs
ShahriarJavidi: I agree, but then we need to decide whether min should be 0 or 1.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
ShahriarJavidi: Fantastic suggestion! Adopted.

rs/nns/governance/src/neuron_store.rs
ShahriarJavidi: `known_neuron_data` is omitted if `multi_query` is true. So, I wanted to include it.

rs/nns/governance/canister/canister.rs
daniel-wong-dfinity-org: OMG 🤦‍♂️

I truly loath these duplicate types. Maybe I can get used to them, but they just seem like a pervasive mental "trip hazard". At the very least, we can give them different names so that we can always tell what we are actually looking at. E.g. api::GovernanceError vs. pb::GovernanceError. I mean, imagine having (identical) twins, but they ALSO have the exact same name! That should probably be illegal (I would not be surprised if naming your children the same is already illegal).

api::Widget vs. pb::WIdget wouldn't be hard at all. In fact, it would greatly simplify our giant `use` statements.

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: Well, you are already deciding that it's 0, right? It's just that you are doing it with 0 code. It's still a decision.

Anyway, do as you see fit.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,147 @@
...
+            id,
+            DissolveStateAndAge::NotDissolving {
+                dissolve_delay_seconds: 100,
+                aging_since_timestamp_seconds: 0,
+            },
+        )
+        .with_controller(PrincipalId::new_user_test_id(id))
+        .build();
+
+        governance.add_neuron(neuron.id().id, neuron).unwrap();
```
daniel-wong-dfinity-org:
```suggestion
        governance.add_neuron(id, neuron).unwrap();
```

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    governance
+}
+
+// This user doesn't control any neurons.
+static NEURONLESS_USER_ID: PrincipalId = PrincipalId::new_user_test_id(1001);
+
+fn assert_neuron_index(neurons: &[NeuronInfo], start_id: u64, size: usize) {
```
daniel-wong-dfinity-org, line 54: Nice spin out. People inline way too much.


```suggestion
#[track_caller]
fn assert_neuron_index(neurons: &[NeuronInfo], start_id: u64, size: usize) {
```

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -0,0 +1,147 @@
...
+    assert_eq!(neurons.len(), size);
+    for (i, neuron) in neurons.iter().enumerate() {
+        let expected_id = start_id + i as u64;
+        let neuron_id = neuron.id.unwrap().id;
+        assert_eq!(
+            neuron_id, expected_id,
+            "Expected neuron ID {} but found {}",
+            expected_id, neuron_id
+        );
```
daniel-wong-dfinity-org, line 63: Why not also controller? Seems easy enough. This is, ofc, not a big deal. At least you are looking at id.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -43,129 +30,118 @@ fn make_governance_for_neuron_index() -> Governance {
...
-        governance.add_neuron(neuron.id().id, neuron).unwrap();
-    }
 
     let request = GetNeuronIndexRequest {
         exclusive_start_neuron_id: None,
         page_size: None,
     };
 
+    let governance = make_governance_for_neuron_index();
+
```
daniel-wong-dfinity-org:
```suggestion
    let governance = make_governance_for_neuron_index();

    let request = GetNeuronIndexRequest {
        exclusive_start_neuron_id: None,
        page_size: None,
    };
```

Ditto elsewhere.

I swapped the order: moved `let governance` up, because that's part of "preparing the world"; whereas, `let request` is part of "calling the code under test", a later step.

Also, `request` is VERY closely related to calling `get_neuron_index`, so I also recommend no blank line between `let request` vs. `get_neuron_index`, but this is not so important.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
```diff
@@ -43,129 +30,118 @@ fn make_governance_for_neuron_index() -> Governance {
...
 
-    let response = governance
-        .get_neuron_index(request, user_id)
-        .expect("Expected valid GetNeuronIndexResponse");
+    let governance = make_governance_for_neuron_index();
+
+    let response = governance.get_neuron_index(request, NEURONLESS_USER_ID);
 
-    assert_eq!(response.neurons.len(), 0);
+    assert!(response.is_err());
```
daniel-wong-dfinity-org, line 147: Better to inspect it more closely, but I guess this is good enough.

daniel-wong-dfinity-org (approved):

rs/nns/governance/src/neuron_store.rs
ShahriarJavidi: Decided not to keep the known neuron data, as there is a similar API function on the Governance.

rs/nns/governance/src/governance/tests/get_neuron_index.rs
ShahriarJavidi: Nice suggestion!

rs/nns/governance/src/governance/tests/get_neuron_index.rs
ShahriarJavidi: Controller is not in the neuron info.

PR #6900, opened 2025-09-23
feat: [NODE-1679] Track mainnet measurements in repo

Bownairo:
Measurements are either pulled from proposals (prod) or from the uploaded `launch-measurements.json` file (dev, not currently uploaded).

This PR also contains some small fixes, organization, and begins to upload the dev measurements.

PR #6972, opened 2025-09-26
feat: revert "revert: "feat: [NODE-1679] Track mainnet measurements in repo  (#6900)" (#6963)"

basvandijk:
This reverts commit 62b938b04d78fabc2bcef6404e2c40c06b88addd which reverted 3f638b22171a3bfda5bc6c4a9b06ce3d10e5c2cb.

This is now run with `CI_ALL_BAZEL_TARGETS` to check if it fails on CI ([which it does](https://dash.dm1-idx1.dfinity.network/invocation/8357b75c-e1ce-47de-929c-b2946a0fa0b5)) so we can debug and fix it.

The original issue was that colocated tests that set `uses_guestos_mainnet_nns_update = True` like `//rs/tests/consensus/backup:backup_manager_upgrade_test_colocate` would [fail](https://dash.zh1-idx1.dfinity.network/invocation/60b73406-6014-449c-b2bc-471a45cb1fe1?target=%2F%2Frs%2Ftests%2Fconsensus%2Fbackup%3Abackup_manager_upgrade_test_colocate&targetStatus=6) because the `ENV_DEPS__GUESTOS_INITIAL_LAUNCH_MEASUREMENTS_FILE` would point to `../mainnet_nns_guest_img/launch-measurements.json` which somehow works on non-colocated tests but on colocated tests the full path to the `launch-measurements.json` file becomes `/home/root/runfiles/../mainnet_nns_guest_img/launch-measurements.json` which doesn't exist.

So the question is why there's a `..` in `../mainnet_nns_guest_img/launch-measurements.json`. I think it's because the `_get_mainnet_guestos_image` repository rule is defined under the `bazel/` directory.


basvandijk:
> Shouldn't we actually save the hash of the setupos in [mainnet-icos-revisions.json](https://github.com/dfinity/ic/blob/9a440e81cfee2214d0b085e893868659c0fc89fd/mainnet-icos-revisions.json) such that we can set its integrity?

It seems that if we start tracking the version of setupos in `mainnet-icos-revisions.json` we could use the simpler [`http_file`](https://bazel.build/rules/lib/repo/http#http_file) macro to download the setupos image rather than the more complicated `repository_ctx.download`.

Bownairo:
> So the question is why there's a `..` in `../mainnet_nns_guest_img/launch-measurements.json`. I think it's because the `_get_mainnet_guestos_image` repository rule is defined under the `bazel/` directory.

I think this comes from https://github.com/bazelbuild/bazel/issues/3675 (and linked issues). `.short_path` just doesn't seem to work well for external repos. Locally, I guess we were lucky the layout works out?

I noticed that the [mainnet canisters](https://github.com/dfinity/ic/blob/a874c92c3673c85875dac22e3419759faa14aa06/rs/tests/common.bzl#L29) are using external repos and working fine. They use [`rootpath`](https://github.com/dfinity/ic/blob/a874c92c3673c85875dac22e3419759faa14aa06/rs/tests/common.bzl#L120) to get the location, and this looks to work for us, too.

Bownairo:
> @Bownairo IIUC implementation functions of repository rules should return None. However `_get_mainnet_guestos_image_impl` [returns a dict](https://github.com/dfinity/ic/blob/9a440e81cfee2214d0b085e893868659c0fc89fd/bazel/mainnet-icos-images.bzl#L160-L163). Why is that and should it be removed?

I based most of this implementation on https://github.com/bazelbuild/bazel/blob/3ebd6adfdc836f456e0bcabac399f2e985c2933f/tools/build_defs/repo/http.bzl#L200. `http_file` creates a separate repo for each file, so I aimed at this lower level so we could instead have one repo per image.

I like the organization of:
```
@mainnet_latest_guest_img
@mainnet_latest_guest_img//:launch_measurements
```
more than:
```
@mainnet_latest_setupos_disk_image//file
@mainnet_latest_setupos_disk_image_launch_measurements
```
, and I was hoping to get the `setupos` images into the same format. Maybe this is not worth the extra complexity.

---

> I wonder whether the image is cached if `integrity` is not defined.

I think [this line](https://github.com/dfinity/ic/blob/a874c92c3673c85875dac22e3419759faa14aa06/bazel/mainnet-icos-images.bzl#L103) should allow the image to be cached, by trusting and setting the integrity from the download. This part was also pulled from `http_file`.

---

> Shouldn't we actually save the hash of the setupos in [`mainnet-icos-revisions.json`](https://github.com/dfinity/ic/blob/9a440e81cfee2214d0b085e893868659c0fc89fd/mainnet-icos-revisions.json) such that we can set its integrity?

If setting the integrity works like I think it does, then the images _should_ only be pulled once. If not, this is something we could add pretty easily, and setting the hash would give us some extra safety, too. One note is that these hashes do not exist in proposals like the rest of the information we pull (the only exception is `dev` hashes and measurements, but I'd like to move these, too).

PR #6974, opened 2025-09-26
fix: deploy.py wait_func path and guestos_ipv6_address optionality

andrewbattat:
If wait_func is triggered (when there's no IP address in the BMC info), the test fails. Remove this code path.

### DIFF
--- a/ic-os/dev-tools/bare_metal_deployment/deploy.py
+++ b/ic-os/dev-tools/bare_metal_deployment/deploy.py
@@ -149,7 +149,7 @@ class Args:
         self.csv_filename = self.csv_filename or csv_filename_env_var
 
         assert (self.inject_image_ipv6_prefix and self.inject_image_ipv6_gateway) or not (
-            self.inject_image_ipv6_prefix and self.inject_image_ipv6_gateway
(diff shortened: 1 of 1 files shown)


ic-os/dev-tools/bare_metal_deployment/deploy.py
```diff
@@ -472,15 +472,10 @@ def deploy_server(
...
-        # Otherwise, just wait.
         timeout_secs = 5
 
-        def wait_func() -> bool:
-            wait(timeout_secs)
-            return False
-
         def check_connectivity_func() -> bool:
+            log.info(f"Checking guestos ({bmc_info.guestos_ipv6_address}) connectivity...")
             assert bmc_info.guestos_ipv6_address is not None, "Logic error"
```
Bownairo: Maybe we can collapse the optional nature of `guestos_ipv6_address`, while we're at it?

Bownairo (approved):

ic-os/dev-tools/bare_metal_deployment/deploy.py
andrewbattat: Agreed! Resolved!

PR #6976, opened 2025-09-26
test(registry): Calling get_node_operator_record the second time for the same operator ID does cannot modify the record

aterga:
This PR adds a regression test for the `get_node_operator_record` function, which is now tested to be idempotent.


rs/registry/canister/src/mutations/do_add_node_operator.rs
```diff
@@ -119,4 +119,46 @@ mod tests {
...
+
+    #[test]
+    #[should_panic(expected = "Key already present")]
+    fn test_should_panic_if_node_operator_record_exists() {
+        let mut registry = invariant_compliant_registry(0);
+
+        // create a new NO record
+        let node_operator_id = PrincipalId::new_user_test_id(100);
+
+        // Make a proposal to upgrade all unassigned nodes to a new version
```
daniel-wong-dfinity-org, line 131: Delete this comment.

Why:

1. No proposal is being made -> the comment does not describe the code it is attached to.

2. The code is pretty self-explanatory (in particular, where it says `AddNodeOperatorPayload`), and therefore, is in no great need of a comment.

max-dfinity (approved):

rs/registry/canister/src/mutations/do_add_node_operator.rs
```diff
@@ -119,4 +119,46 @@ mod tests {
         };
         assert_eq!(node_operator_record, expected);
     }
+
+    #[test]
+    #[should_panic(expected = "Key already present")]
+    fn test_should_panic_if_node_operator_record_exists() {
+        let mut registry = invariant_compliant_registry(0);
+
+        // create a new NO record
```
daniel-wong-dfinity-org, line 128: Delete this comment.

Similar reasons:

1. Line 129 does NOT create a record.

2. Also line 129 doesn't particularly need a comment; it is pretty self-explanatory

rs/registry/canister/src/mutations/do_add_node_operator.rs
```diff
@@ -119,4 +119,46 @@ mod tests {
...
+            rewardable_nodes: btreemap! { "type1.1".to_string() => 1 },
+            ipv6: Some("bar".to_string()),
+            max_rewardable_nodes: btreemap! { "type1.2".to_string() => 1 },
+        };
+        assert_eq!(node_operator_record, expected);
+
+        // Change the payload, but keep the same node operator id.
+        payload.node_allowance = 2;
+
+        // Test that an existing record cannot be edited using this function.
```
daniel-wong-dfinity-org, line 161:
```suggestion
        // Recall that this is a should_panic test. The next line is where the expected
        // panic is supposed to arise from.
```

rs/registry/canister/src/mutations/do_add_node_operator.rs
```diff
@@ -119,4 +119,46 @@ mod tests {
...
+            node_operator_principal_id: Some(node_operator_id),
+            node_provider_principal_id: Some(PrincipalId::new_user_test_id(1000)),
+            node_allowance: 1,
+            dc_id: "DC1".to_string(),
+            rewardable_nodes: btreemap! { "type1.1".to_string() => 1 },
+            ipv6: Some("bar".to_string()),
+            max_rewardable_nodes: Some(btreemap! { "type1.2".to_string() => 1 }),
+        };
+
+        // Test that a record can be created from this payload.
```
daniel-wong-dfinity-org, line 142:
```suggestion
        // Initially, when there is no NodeOperatorRecord, adding one is (supposed to be) allowed.
        // Later, we'll try to add again, but that attempt should get blocked (with a panic),
        // because of the record being added here.
```

rs/registry/canister/src/mutations/do_add_node_operator.rs
```diff
@@ -119,4 +119,46 @@ mod tests {
...
+            ipv6: Some("bar".to_string()),
+            max_rewardable_nodes: btreemap! { "type1.2".to_string() => 1 },
+        };
+        assert_eq!(node_operator_record, expected);
+
+        // Change the payload, but keep the same node operator id.
+        payload.node_allowance = 2;
+
+        // Test that an existing record cannot be edited using this function.
+        registry.do_add_node_operator(payload);
```
daniel-wong-dfinity-org, line 162: Non-blocking issue: If line 143 causes such a panic (or any other line for that matter), this test will _incorrectly_ pass. That's the problem with `should_panic` (and `panic`); it is extremely imprecise. Even though what you want to say is "the panic (is supposed to) arise from line 162", there is no way to express that (because the `should_panic` covers the ENTIRE body of the test, not just line 162). It can be precise if the panic message itself contains something SPECIFIC, an actual smoking gun; whereas, "Key already present" is super generic, not a smoking gun. I have no suggestion what to do here though; hence, non-blocking.

daniel-wong-dfinity-org (approved):

PR #6981, opened 2025-09-26
fix(node-rewards-canister): Discard DayUtc type in favor of NaiveDate

pietrodimarco-dfinity:
This introduces some breaking changes to the Node Rewards canister API, however that's considered acceptable since nobody is calling it yet.

Changes:

- Removed completely `DayUtc` base custom type in favour of `chrono::NaiveDate`
- Adapted DayUtc in API types to work with `chrono::NaiveDate`
- Adapted tests


rs/node_rewards/canister/Cargo.toml
```diff
@@ -37,6 +37,10 @@ mockall = { workspace = true }
 prost = { workspace = true }
 rewards-calculation = { path = "../rewards_calculation" }
 itertools = { workspace = true }
+chrono = { version = "0.4.41", default-features = false, features = [
+    "alloc",
+    "serde",
+] }
```
daniel-wong-dfinity-org: It is rather unusual to not do

```
foo = { workspace = true }
```

Why are you doing it like this? Specific features? Specific version? At the very least, it should be explained why this unusual way is being used.

I am afraid that this will lead to some kind of breakage down the road. Like, you might depend on X, and X only accepts version 9, and you only accept version 10. Then, you will become broken, and it won't be straightforward to fix; not impossible, but whoever ends up experiencing this will very likely not be familiar with your code, and therefore, be at risk of fixing things incorrectly. I think `{ workspace = true }` avoids laying such landmines for future maintainers.

Ditto elsewhere.

rs/node_rewards/canister/api/src/lib.rs
```diff
@@ -11,33 +12,55 @@ use rewards_calculation::types as native_types;
...
+            value.day.expect("Day is missing"),
         )
+        .ok_or(format!("Invalid date: {:?}", value))
+    }
+}
+
+impl DayUtc {
+    pub fn from_unix_timestamp_nanoseconds(value: u64) -> Self {
+        let naive_date = DateTime::from_timestamp_nanos(value as i64).date_naive();
+        DayUtc::from(naive_date)
```
daniel-wong-dfinity-org: Consider

```suggestion
        Self::from(naive_date)
```

rs/node_rewards/canister/api/src/lib.rs
```diff
@@ -11,33 +12,55 @@ use rewards_calculation::types as native_types;
...
+    }
+    pub fn from_unix_timestamp_seconds(value: u64) -> Self {
+        let naive_date = DateTime::from_timestamp(value as i64, 0)
+            .unwrap()
+            .date_naive();
+        DayUtc::from(naive_date)
```
daniel-wong-dfinity-org: Ditto.

rs/node_rewards/canister/src/metrics/tests.rs
```diff
@@ -475,8 +474,8 @@ async fn daily_metrics_correct_overlapping_days() {
     let subnet_2 = subnet_id(2);
 
     let node_1 = node_id(1);
-    let node_2 = node_id(1);
-    let day_start = DayUtc::from_nanos(0);
+    let node_2 = node_id(2);
```
daniel-wong-dfinity-org, line 477: oops 😅

rs/node_rewards/canister/src/metrics/tests.rs
```diff
@@ -475,8 +474,8 @@ async fn daily_metrics_correct_overlapping_days() {
...
+    let day_start = NaiveDate::from_ymd(1991, 9, 5);
```
daniel-wong-dfinity-org: Consider using a more recent value, like idk 2025-01-01.

rs/node_rewards/canister/src/registry_querier/tests.rs
```diff
@@ -290,23 +279,22 @@ fn test_node_re_registered_after_deletion() {
...
-        DayUtc::try_from("2025-07-08").unwrap(),
-        DayUtc::try_from("2025-07-09").unwrap(),
-        DayUtc::try_from("2025-07-10").unwrap(),
+        to_native_date("2025-07-08"),
+        to_native_date("2025-07-09"),
+        to_native_date("2025-07-10"),
     ];
 
-    while current_day <= to {
+    for day in from.iter_days().take_while(|d| *d <= to) {
```
daniel-wong-dfinity-org, line 290: 👍

rs/node_rewards/rewards_calculation/src/types.rs
daniel-wong-dfinity-org, line 1: 😍

daniel-wong-dfinity-org (approved):
Thanks for this. Re-inventing the wheel is generally suboptimal. Therefore, this un-reinventing of the wheel is very probably a Good Time(TM).

PR #6989, opened 2025-09-29
fix(http-outcalls): Make invalid_ip correctness test less flaky.

mihailjianu1:
**Before**:
The test was asserting a response that it got from calling out to `240.0.0.0`.
This had several issues:
- when API Boundary Nodes will be used as socks proxies for the app subnet, they won't have a static IPv4 address anymore 
- the test was asserting for a timeout (which does happen in 97% of the cases), however in the remaining 3%, the request fails mot because of a timeout, but because the "[Time to live](https://en.wikipedia.org/wiki/Time_to_live)" expires before the 30 seconds deadline (which is non deterministic). This made the error type flaky (either `SysFatal` or `SysTransient`).

**After**:
The test is calling out to `2001:db8::1`, which is a [reserved IPv6 address](https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv6). It now consistently fails with `SysTransient`.


rs/tests/networking/canister_http_correctness_test.rs
```diff
@@ -1121,14 +1121,14 @@ fn test_invalid_domain_name(env: TestEnv) {
...
-fn test_invalid_ip_timeout(env: TestEnv) {
+fn test_invalid_ip(env: TestEnv) {
     let handlers = Handlers::new(&env);
 
     let (response, refunded_cycles) = block_on(submit_outcall(
         &handlers,
         RemoteHttpRequest {
             request: UnvalidatedCanisterHttpRequestArgs {
-                url: "https://240.0.0.0".to_string(),
+                url: "https://[2001:db8::1]".to_string(),
```
kpop-dfinity, line 1133: can we leave a comment what this ip address refers to? Something like
```suggestion
                // `2001:db8::1` is a reserved ipv6 address used in documentation and example source code. 
                // See https://www.rfc-editor.org/rfc/rfc3849
                url: "https://[2001:db8::1]".to_string(),
```

rs/tests/networking/canister_http_correctness_test.rs
```diff
@@ -1148,12 +1148,10 @@ fn test_invalid_ip_timeout(env: TestEnv) {
 
     assert_matches!(
         response,
-        Err(
-            RejectResponse {
-            reject_code: RejectCode::SysFatal,
-            reject_message,
+        Err(RejectResponse {
+            reject_code: RejectCode::SysTransient,
```
kpop-dfinity, line 1154: The PR description says that we should fail with `SysFatal`

mihailjianu1: Oups thanks! Fixed the PR description

PR #6993, opened 2025-09-29
feat(IDX): move weekly secrets to env

cgundy:
This moves weekly secrets to an environment so they can only be accessed on the master branch.


nmattia (approved):
Nice!

PR #6996, opened 2025-09-29
feat(ICRC-Index): handle incorrect blocks, don't schedule timer if bad block

maciejdfinity:


rs/ledger_suite/icrc1/index-ng/index-ng.did
```diff
@@ -134,7 +136,15 @@ type ListSubaccountsArgs = record {
 };
 
 type Status = record {
-  num_blocks_synced : BlockIndex
+  num_blocks_synced : BlockIndex;
+  sync_status : SyncStatus
```
maciejdfinity: We could consider making it optional - this would allow a new client to call the endpoint on an older index.

mbjorkqvist: Yes, I think I'd lean towards making `sync_status` optional. That would also allow us to remove `sync_status` from the `Status` record at a later point in time if we wanted to.

What about a scenario where we add a new variant to `SyncStatus` (such as e.g., `Synced`, if the ledger produced a [`124deactivate`](https://github.com/dfinity/ICRC/pull/135) block indicating that no more blocks would ever be produced, in which case we could stop the timer), how would that be handled? If I interpret [the Candid spec](https://github.com/dfinity/candid/blob/7e1bd3b1ecde4163b86a231291ec7e5ad1250245/spec/Candid.md#upgrading-and-subtyping) correctly, a client would interpret an enum variant it doesn't know about as `null` (as long as `sync_status` is optional)?

rs/ledger_suite/icrc1/index-ng/src/main.rs
```diff
@@ -679,21 +691,64 @@ async fn fetch_blocks_via_icrc3() -> Option<u64> {
...
+    let timer_id = ic_cdk_timers::set_timer_interval(after, || {
         ic_cdk::spawn(async {
             let _ = build_index().await;
         })
-    })
+    });
+    TIMER_ID.with(|tid| *tid.borrow_mut() = Some(timer_id));
+}
+
+fn stop_timer_with_error(error: String) {
```
mbjorkqvist: I can find 6 occurrences where `stop_timer_with_error` is called, in three different functions. With the current approach, I feel it's a bit difficult to get an overview of how the timer stopping is done - it's necessary to check all the code paths, and later on if something is changed, it is easier to e.g., miss some cases. I'm wondering if it would make sense to have some more consistent error handling for the functionality run by the timer (retrieving blocks and appending them to the data structures), e.g.,:

- Define a trait for the errors that can be used to determine if the timer should be stopped or not.
- Create errors that implement the aforementioned trait where we currently simply return `Err(())`.
- In `build_index`, if `num_indexed` is an error, using the trait, check if we should stop the timer or not, and only do it here (not anywhere deeper down in the call stack).
- We could e.g., use `anyhow` to add context and only log at the top level (in `build_index()`) - this is somewhat orthogonal to the changes required to stopping the timer. Or we could potentially set `SYNC_ERROR` using an `anyhow` error.

The above would require quite a lot of code changes, and since this PR is already quite larger, I'd be happy to discuss doing this separately (before or after merging this one).

rs/ledger_suite/icrc1/index-ng/src/main.rs
```diff
@@ -679,21 +691,64 @@ async fn fetch_blocks_via_icrc3() -> Option<u64> {
...
+    log!(P0, "{}", error);
+    ic_cdk::eprintln!("{}", error);
+    if let Some(timer_id) = TIMER_ID.with(|tid| *tid.borrow()) {
+        ic_cdk_timers::clear_timer(timer_id);
+        TIMER_ID.with(|tid| *tid.borrow_mut() = None);
+    }
```
mbjorkqvist: With the current implementation, can it ever happen that `TIMER_ID` is `None` here?

gregorydemay (approved):

rs/ledger_suite/icrc1/index-ng/src/main.rs
maciejdfinity: It should not happen, I could change this to panic if timer id is none?

rs/ledger_suite/icrc1/index-ng/src/main.rs
maciejdfinity: Good point! I return the `SyncError` struct which contains the error message and a boolean indicating whether the error is retriable. I log the error and stop the sync only in one place - `build_index`.

rs/ledger_suite/icrc1/index-ng/src/main.rs
maciejdfinity: I changed it to `expect`, stopping the indexing is now done in `build_index`.

rs/ledger_suite/icrc1/index-ng/index-ng.did
maciejdfinity: As discussed, we removed `sync_status` completely. The error information can be obtained by querying the `/logs` http endpoint.

rs/ledger_suite/icrc1/index-ng/src/main.rs
maciejdfinity: Actually, since we don't report `sync_status` anymore, I changed `TIMER_ID` from`Option<TimerId>` to `TimerId`.

rs/ledger_suite/icrc1/test_utils/src/lib.rs
```diff
@@ -761,15 +774,17 @@ pub fn valid_transactions_strategy_with_excluded_transaction_types(
             amount_strategy(),
             valid_created_at_time_strategy(now),
             arb_memo(),
+            prop::option::of(Just(default_fee)),
```
mbjorkqvist: I thought setting a fee half of the time for `mint` transactions (and the same for `burn_strategy` below) would be a problem for the tests that interact with our production ledgers, which do not support fees in `mint` or `burn` transactions, but the `fee` sent in the `TransferArg` to the ledger is actually always `None`, so I'm wondering if this `prop::option::of(Just(default_fee))` is needed - maybe I'm missing something! Another thing I thought might be an issue is the fact the the check for duplicates is performed *with* the fee, so if we happened to generate two identical `mint` transactions, but one with a fee and one without, we wouldn't treat them as duplicates here when generating the transactions, but the ledger should discard the second one as a duplicate, which could cause some test flakiness? Since there's a lot of other varying input (`to`, `memo`, `amount`), I guess this is very unlikely to happen in practice.

rs/rosetta-api/icrc1/src/common/storage/storage_operations.rs
```diff
@@ -295,7 +295,7 @@ pub fn update_account_balances(connection: &mut Connection) -> anyhow::Result<()
                         )?;
                     }
                 }
-                crate::common::storage::types::IcrcOperation::Mint { to, amount } => {
+                crate::common::storage::types::IcrcOperation::Mint { to, amount, .. } => {
```
mbjorkqvist: Would it make sense here (and for the other operations also) to explicitly list all the fields (prefixed with underscores if they are not used, e.g., `fee: _fee` here), so that we would get compiler errors if/when new fields are added to the structs, and then it would be more difficult for new fields that we should interpret to fall through the cracks?

rs/ledger_suite/icrc1/test_utils/src/lib.rs
maciejdfinity: Yeah, I am not sure what I was thinking here, I added fee to the tx used only for deduplication but not to the transfer args that are actually used for making the transaction. I removed it!

rs/rosetta-api/icrc1/src/common/storage/storage_operations.rs
maciejdfinity: Good point! I added all the fields in this `match` statement. It could make sense to look at the whole ledger_suite and rosetta codebase to see if similar change would make sense but this PR would probably grow too much if we did it here. Should we do it as part of a separate ticket?

rs/ledger_suite/icrc1/test_utils/src/lib.rs
maciejdfinity: I still add the optional mint/burn fee in the blocks_strategy, but that one is used to verify block conversions and by rosetta tests.

rs/rosetta-api/icrc1/src/common/storage/storage_operations.rs
mbjorkqvist: A separate ticket sounds good!

maciejdfinity: added: https://dfinity.atlassian.net/browse/FI-1890

PR #6997, opened 2025-09-29
fix(IDX): fix cargo clippy logic

cgundy:
There was a small issue with this job where it was automatically getting skipped on master and rc branches. This changes to always run on master and rc brances and only run on PRs if a specific file change was made.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -704,7 +704,8 @@ jobs:
         if: |
           steps.filter.outputs.cargo == 'true' ||
           env.BRANCH_NAME == 'master' ||
-          startsWith(env.BRANCH_NAME, 'rc--')
+          startsWith(env.BRANCH_NAME, 'rc--') ||
+          startsWith(env.BRANCH_NAME, 'hotfix-')
         shell: bash
         run: |
           set -eExuo pipefail


.github/workflows/ci-main.yml
```diff
@@ -691,9 +691,6 @@ jobs:
       - name: Filter Rust Files [*.{rs,toml,lock}]
         uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
         id: filter
-        if: |
-          github.event_name == 'pull_request' ||
-          github.event_name == 'merge_group'
```
basvandijk, line 696: The [behaviour](https://github.com/dorny/paths-filter?tab=readme-ov-file#supported-workflows) of dorny/paths-filter for non-PRs is a bit more complex and requires the `base` parameter. It might be easier to understand the behaviour to keep the `pull_request` and `merge_group` condition.

.github/workflows/ci-main.yml
```diff
@@ -702,11 +699,12 @@ jobs:
...
         id: cargo-clippy-linux
+        env:
+          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         if: |
           steps.filter.outputs.cargo == 'true' ||
-          github.event_name == 'schedule' ||
-          github.event_name == 'workflow_dispatch' ||
-          github.event_name == 'workflow_call'
+          env.BRANCH_NAME == 'master' ||
+          startsWith(env.BRANCH_NAME, 'rc--')
```
basvandijk: What about including `master-private` and `hotfix-` as well like:
```suggestion
          env.BRANCH_NAME == 'master' ||
          env.BRANCH_NAME == 'master-private' ||
          startsWith(env.BRANCH_NAME, 'rc--') ||
          startsWith(env.BRANCH_NAME, 'hotfix-')
```

.github/workflows/ci-main.yml
cgundy: So the problem is if we skip this condition then the subsequent job also never gets run - see this example: https://github.com/dfinity/ic/actions/runs/18084667430/job/51453715110

.github/workflows/ci-main.yml
cgundy: I don't think master-private is relevant since we don't run any tests on that branch in ic-private.

basvandijk: Ok make sense.

.github/workflows/ci-main.yml
basvandijk: I see. Ok then ignore it.

basvandijk (approved):

PR #6999, opened 2025-09-29
fix: Return current registry version

michael-weigelt:


rs/registry/canister/src/mutations/do_migrate_canisters.rs
```diff
@@ -10,15 +10,15 @@ impl Registry {
...
         self.maybe_apply_mutation_internal(self.migrate_canisters_to_subnet(
-            self.latest_version(),
+            version,
             canister_ids,
             target_subnet_id,
         ));
 
         MigrateCanistersResponse {
-            registry_version: self.latest_version(),
+            registry_version: version,
```
max-dfinity, line 21: I'm not sure what the bug is.  What version of the Registry is this intended to be?  Is it supposed to be the one before the latest version, and if so why?

I don't quite understand, but is it possible to add some documentation to the types, and also some comments here to explain what the response is supposed to have?

michael-weigelt: The migration canister causes a registry change by calling this endpoint. Then it needs to wait until the two subnets involved in migration have acquired this new registry version. The subnet_info endpoint (on these subnets) returns a version that is off-by one w.r.t. the one I returned in the previous PR. The reason seems to be (and correct me if I am wrong) that the registry canister internal counter self.latest_version() actually indicates the _next_ registry version. So with this PR, I return the version that the other subnets will actually report.

michael-weigelt: If you can confirm that this interpretation and new behaviour is correct, I would add a comment here and document the return type (even though that does not necessarily help external callers...)

max-dfinity: The implementation in registry of "latest_version()" is as follows:

```
 pub fn latest_version(&self) -> Version {
        self.version
    }
```

If you want to have the version of the Registry that has the changes, then the previous behavior was correct, and I'm not quite sure why the other test about subnet_info would be failing, except maybe it hadn't finished syncing the version?

michael-weigelt: (I did check the implementation of latest_version and also checked where the value is incremented - and semi-concluded that the meaning of the variable is other than expected). 

Thanks for clearing that up, then I will close this PR. But then I don't understand how my other test would not work. I gave the subnets 11 (Edit: 40) minutes to report the new version, and they did not.

PR #7004, opened 2025-09-29
test: Systest for migration canister

michael-weigelt:


rs/tests/execution/canister_migration_test.rs
```diff
@@ -192,4 +194,20 @@ async fn test_async(env: TestEnv) {
...
+        .await
+        .expect("Failed to call migration_status.");
+    let decoded_status = Decode!(&status, Vec<MigrationStatus>)
+        .expect("Failed to decode response from migration_status.");
+    println!("status: {:?}", decoded_status);
+
+    assert!(matches!(
+        decoded_status[0],
+        MigrationStatus::Succeeded { .. }
+    ));
```
schneiderstefan: Ideally after the migration, the test should try to interact with the migrated canister and verify that it now runs on the new subnet.

rs/tests/execution/canister_migration_test.rs
```diff
@@ -171,25 +195,77 @@ async fn test_async(env: TestEnv) {
...
     assert_eq!(
-        decoded_status,
-        vec![MigrationStatus::InProgress {
-            status: "Accepted".into()
-        }]
+        subnet_id.unwrap(),
+        app_subnet_2.subnet_id().unwrap().get().0
     );
+    // wait until delegation updates
+    tokio::time::sleep(Duration::from_secs(610)).await;
```
michael-weigelt: This can be adjusted once @schneiderstefan lowers the delegation refresh period.

rs/tests/execution/canister_migration_test.rs
```diff
@@ -171,25 +195,77 @@ async fn test_async(env: TestEnv) {
...
+
     let decoded_result = Decode!(&result, Result<(), ValidationError>)
         .expect("Failed to decode reponse from migrate_canister.");
 
     assert_eq!(decoded_result, Ok(()));
 
-    info!(logger, "Calling migration_status");
+    // The migration canister has a step where it waits for 5 minutes, so we give it a minute more than that.
+    println!("Wait over 5 minutes for processing.");
+    tokio::time::sleep(Duration::from_secs(360)).await;
```
michael-weigelt: @basvandijk this wait is unavoidable due to how the migration canister works. Would you prefer some other means of waiting?

rs/tests/execution/canister_migration_test.rs
schneiderstefan: Instead of just sleeping, we should just retry until it succeeds (with a max). Same for the other sleep. Then you can also print something every time we try.

rs/tests/execution/canister_migration_test.rs
basvandijk: No, this is fine as long as you mark the test as a `long_test` to not run it by default on PRs.

basvandijk: BTW if there's a way to check that the migration_canister is done processing it would be better to repeatedly check that inside a `retry_with_msg!` instead of just waiting. Since that'll inform the user what we're waiting on and we don't need to keep this wait time in sync between the canister and its test.

rs/tests/execution/canister_migration_test.rs
```diff
@@ -23,6 +26,8 @@ fn main() -> Result<()> {
     SystemTestGroup::new()
         .with_setup(setup)
         .add_test(systest!(test))
+        .with_timeout_per_test(Duration::from_secs(1200))
+        .with_overall_timeout(Duration::from_secs(1200))
```
basvandijk: Are these long timeouts necessary? 

Note that the bazel test target of this test (`//rs/tests/execution:canister_migration_test`) is using the [default](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/system_tests.bzl?L208) `test_timeout` of `long` meaning 900 seconds. So the bazel timeout would kick in sooner than these.

If the test really takes this long you want to increase the `test_timeout` to `eternal` (1 hour) in the bazel target. And, as already mentioned above, tag this test as a `long_test`.

PR #7008, opened 2025-09-30
feat: relax validation in copy during state sync

ShuoWangNSL:
…nager started height


rs/state_manager/tests/state_manager.rs
```diff
@@ -4251,6 +4251,18 @@ fn can_recover_from_corruption_on_state_sync() {
             std::fs::write(&canister_100_raw_pb, b"Garbage").unwrap();
             make_readonly(&canister_100_raw_pb).unwrap();
 
+            // Force validation during state sync for testing corruption recovery.
+            // Normally validation only occurs when base checkpoint height <= started_height
+            // (i.e., after state manager restart), but we override this for testing purposes.
+            //
+            // This test is to verify that the validation logic can dectect various types of corruption
```
Andrei1998: typo (dectect)

rs/state_manager/src/state_sync/chunkable.rs
```diff
@@ -347,6 +367,7 @@ impl IncompleteState {
     /// Hardlink unchanged files from previous checkpoint according to diff script.
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn hardlink_files(
+        self: &IncompleteState,
```
schneiderstefan:
```suggestion
        &self,
```

Am I missing something here? Shouldn't clippy complain about that?

rs/state_manager/src/state_sync/chunkable.rs
```diff
@@ -1098,6 +1118,7 @@ impl IncompleteState {
 
             let mut thread_pool = self.thread_pool.lock().unwrap();
             Self::hardlink_files(
+                self,
```
schneiderstefan:
```suggestion
            self.hardlink_files(
```

Same pattern is used at some other places.

rs/state_manager/tests/state_manager.rs
ShuoWangNSL: Good catch.

PR #7009, opened 2025-09-30
test(recovery): use separate SSH keys in recovery system tests

pierugo-dfinity:
This PR enables the generation of new SSH keys for any user, not just `admin`. It is used in recovery system tests as follows:
- NNS recoveries: create a separate SSH key for `backup` access. Additionally, we remove `admin` access on all nodes except the DFINITY-owned one to perfectly mirror the production setup.
- App subnet recoveries: create a separate SSH key for `readonly` access. Unfortunately, we cannot blindly remove `admin` access like above because the last check of the test (`assert_node_is_unassigned`) uses admin SSH access. Thus, we still do so but we keep track of the SSH connections that we opened to perform this operation and re-use those for the last check. We still successfully prevent new `admin` SSH connections from being established.


andrewbattat (approved):

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -523,6 +539,57 @@ fn app_subnet_recovery_test(env: TestEnv, cfg: TestConfig) {
...
+            node.get_ip_addr()
+        );
+
+        let session = disable_ssh_access_to_node(&node, SSH_USERNAME, &admin_auth).unwrap();
+
+        admin_ssh_sessions.insert(node.node_id, session);
+    }
+    // Scenarios where we still need admin access to the download node are:
+    //  - Local recovery
+    //  - Cases where we cannot deploy read-only access, and we need to fall back to admin access
```
eichhorl: Why do we need admin access in these cases?

pierugo-dfinity: In local recoveries, we use `--download-method local` which means we need to run `ic-recovery` on the download node, as well as `--upload-method local`, which means we also need admin access.

In cases where we cannot deploy read-only access (here which happens if we want to test corrupting the CUP), see this failed [run](https://dash.dm1-idx1.dfinity.network/invocation/3f1a8c3f-861f-4817-8063-cee0d229e84a?target=%2F%2Frs%2Ftests%2Fconsensus%2Fsubnet_recovery%3Asr_app_no_upgrade_with_chain_keys_test_colocate&targetStatus=6) if we remove admin access on the download node. We cannot download & merge certifications (which is why they are [skipped](https://github.com/dfinity/ic/blob/a54a2fd77ab14a1502f15ab35c470fb35f3062eb/rs/recovery/src/app_subnet_recovery.rs#L355)) so if this were to happen in production, I am [interpreting](https://github.com/dfinity/ic/blob/a54a2fd77ab14a1502f15ab35c470fb35f3062eb/rs/recovery/src/app_subnet_recovery.rs#L373-L377) the code that we would require admin access on the highest finalized node and use their state & pool to proceed with the recovery. Isn't that correct?

An alternative would actually be to follow what we do in NNS recoveries and use backup access instead, but that would require to add a new backup key to every subnet of the IC just for that edge-case...

eichhorl: > In local recoveries, we use --download-method local which means we need to run ic-recovery on the download node, as well as --upload-method local, which means we also need admin access.

So if I understand correctly, in local recoveries we don't actually use the upload node? Maybe it would be clearer to select the admin nodes beforehand, depending on the test configuration, and then remove admin access on all nodes except the selected admin nodes.

> we would require admin access on the highest finalized node and use their state & pool to proceed with the recovery

The only way the orchestrator's CUP can be corrupted on enough nodes to stall the subnet should be during upgrades. In that case we know that all nodes stalled on the same height (the upgrade height), and already have the latest state. So ideally in this case we don't need to download or replay any artifacts.

I guess the test scenario here isn't very realistic in that case, the point is mostly to test the orchestrator's ability to recover from a broken CUP. So having an extra admin node should be fine.

pierugo-dfinity: Thanks, I agree the style isn't very readable. I followed what you said of choosing the admin nodes in advance.

> So if I understand correctly, in local recoveries we don't actually use the upload node?

Indeed. That wasn't clear to me at first either, so I also changed the definition of `upload_node`. It doesn't actually have an impact because we ignore that parameter in local recoveries anyways, but together with the previous change, it allows to simplify the logic of which nodes to choose when disabling SSH admin access. PTAL

eichhorl: Great, LGTM!

PR #7013, opened 2025-09-30
feat(nns): Disallow following private neurons (with some exceptions)

ShahriarJavidi:
This PR enforces that either of these restrictions when requesting to follow a private neuron:

1. the follower neuron has the same controller as the followee,  or
2. the controller of the follower neuron is among the hotkeys of the followee.

### DIFF
--- a/rs/nns/governance/tests/governance.rs
+++ b/rs/nns/governance/tests/governance.rs
@@ -1904,7 +1904,7 @@ fn test_enforce_public_neuron() {
     );
 
     // Step 3: Inspect results.
-    // TODO[NNS1-4230]:
+    // TODO(NNS1-4230):
     // Here, we just want to assert that for a random principal ID
     // full_neuron is not in the result.
     assert_eq!(neuron_info.visibility, Some(Visibility::Public as i32));


rs/nns/integration_tests/src/neuron_following.rs
```diff
@@ -139,6 +139,7 @@ fn nonexistent_neuron_cannot_follow_neuron() {
 }
 
 #[test]
+#[should_panic]
 fn neuron_follow_nonexistent_neuron() {
```
ShahriarJavidi: change names to reflect the behaviour change.

ShahriarJavidi: Done.

rs/nns/governance/src/governance.rs
```diff
@@ -3315,23 +3315,86 @@ impl Governance {
...
+                    if let Ok(follow_neuron) = self.with_neuron(follow, |neuron| neuron.clone()) {
+                        if follow_neuron.visibility() == Visibility::Public
+                            || follow_neuron.controller() == controller
+                            || hotkeys.contains(&follow_neuron.controller())
+                        {
+                            continue;
+                        } else {
+                            invalid_followees += 1;
+                            error_message.push_str(&format!(
+                                "{}: Not authorized to follow private neuron: {follow:?}\n\t\
```
mstrasinskis: How about adding a comment to prevent unexpected message text changes? Something like:
```
// Note: These error messages are matched in the nns-dapp. In case of changes, please sync with the nns dev team.
```

ShahriarJavidi: Sure. Thanks for pointing out!

rs/nns/governance/init/src/lib.rs
```diff
@@ -281,6 +284,8 @@ impl GovernanceCanisterInitPayloadBuilder {
         }
 
         for result in reader.records() {
+            use ic_nns_governance_api::Visibility;
```
daniel-wong-dfinity-org: Put this at the top of the file. (This is a pattern.)

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+        topic_to_followees: &HashMap<i32, Followees>,
+        topic: i32,
+        new_followees: Followees,
+    ) -> Result<HashMap<i32, Followees>, GovernanceError> {
+        let controller = neuron.controller();
+        let mut updated_followees = topic_to_followees.clone();
+        if new_followees.followees.is_empty() {
+            // If the new followees list is empty, remove the entry for the topic.
+            updated_followees.remove(&topic);
+        } else {
```
daniel-wong-dfinity-org:
```suggestion
        if new_followees.followees.is_empty() {
            // If the new followees list is empty, remove the entry for the topic.
            updated_followees.remove(&topic);
            return Ok(updated_followees);
        }

```

This allows the reader to mentally "retire" the empty case without having to read a page full of code.

> Flat is better than nested.
> --Zen of Python

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+            // Otherwise, update the entry with the new followees list.
+            if is_neuron_follow_restrictions_enabled() {
```
daniel-wong-dfinity-org: Similarly,

```suggestion
            if !is_neuron_follow_restrictions_enabled() {
                update_followees.insert(...);
                return Ok(updated_followees);
            }

```

This will simplify getting rid of the flag, because you won't have to unindent a bunch of flag-protected code.

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+                // A new can follow another neuron if:
+                // 1. the followee neuron is a public neuron
+                // 2. or if the followee neuron is a private neuron, they share a controller.
+                // If in the list of followees, there is any follow relationship
+                // the doesn't adhere to the aforementioned rules, return a GovernanceError.
+                let (mut invalid_followees, mut error_message) = (0_u32, String::new());
```
daniel-wong-dfinity-org: Split this into two separate statements.

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+
+                for followee in &new_followees.followees {
+                    if let Ok(followee_neuron) = self.with_neuron(followee, |neuron| neuron.clone())
```
daniel-wong-dfinity-org: Hmm. This is slightly expensive, but I guess it cannot be avoided. Well, at least, we need not fetch auxiliary fields, right? All we need to know is

a) visibility
b) controller

```suggestion
                    let followee = self.with_neuron(
                        followee,
                        |neuron| (neuron.visibility, neuron.controller),
                    );
                    let (followee_visibility, followee_controller) = match followee {
                        Ok(ok) => ok,
                        Err(err) => {
                            error_message.push_str(...);
                            continue;
                        }
                    };

```

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+                    {
+                        if followee_neuron.visibility() == Visibility::Public
+                            || followee_neuron.controller() == controller
+                            || followee_neuron.hot_keys.contains(&controller)
+                        {
+                            continue;
+                        } else {
```
daniel-wong-dfinity-org:
```suggestion
                        }

```

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+                // If in the list of followees, there is any follow relationship
+                // the doesn't adhere to the aforementioned rules, return a GovernanceError.
+                let (mut invalid_followees, mut error_message) = (0_u32, String::new());
+
+                for followee in &new_followees.followees {
+                    if let Ok(followee_neuron) = self.with_neuron(followee, |neuron| neuron.clone())
+                    {
+                        if followee_neuron.visibility() == Visibility::Public
+                            || followee_neuron.controller() == controller
+                            || followee_neuron.hot_keys.contains(&controller)
```
daniel-wong-dfinity-org:
```suggestion
                        let ok = followee_neuron.visibility() == Visibility::Public
                            || followee_neuron.controller() == controller
                            || followee_neuron.hot_keys.contains(&controller);
                        if ok {
```

In general, it is not so good when the open brace of a block statement is not on the same line as initial the key word. Well, I know this is a personal opinion, but I think it's a pretty good one 😇 Don't be afraid to use "dummy" variables! Semicolon is not evil!

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+                        ));
+                    }
+                }
+
+                if invalid_followees > 0 {
+                    // Note: These error messages are matched in the nns-dapp. In case of changes, please sync with the nns dev team.
+                    error_message = format!(
+                        "The {} followee(s) listed below is(are) invalid:\n{}",
+                        invalid_followees, error_message
+                    );
```
daniel-wong-dfinity-org: Very informative error message!

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
         self.with_neuron_mut(id, |neuron| {
-            neuron.followees = new_followees;
+            neuron.followees = topic_to_followees;
             neuron.refresh_voting_power(now_seconds);
         })?;
 
         Ok(())
     }
 
+    fn modify_followees(
```
daniel-wong-dfinity-org: Thanks for spinning this out! People inline too often.

rs/nns/governance/src/governance.rs
```diff
@@ -3360,23 +3360,95 @@ impl Governance {
...
+        &self,
```
daniel-wong-dfinity-org: The only thing we need is the NeuronStore, not the rest of Governance, right? Better to only pass what's needed. "Need to know" is a good thing to live by. Control access to data.

rs/nns/governance/src/governance.rs
```diff
@@ -5940,18 +6012,21 @@ impl Governance {
         })?;
 
         let now_seconds = self.env.now();
+        let neuron_followees = self.with_neuron(id, |neuron| {
```
daniel-wong-dfinity-org:
```suggestion
        let new_neuron_followees = self.with_neuron(id, |neuron| {
```

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: I think also we want to allow existing followees. We can check that before `with_neuron` to avoid the expense of looking up existing followees.

rs/nns/governance/tests/governance.rs
```diff
@@ -1864,8 +1866,10 @@ fn fixture_for_manage_neuron() -> api::Governance {
         .build()
 }
 
+// @todo currently, `fixture_for_manage_neuron` creates neurons with public visibility.
+// We should add tests for private neurons too.
 #[test]
-fn test_enforce_private_neuron() {
+fn test_enforce_public_neuron() {
```
daniel-wong-dfinity-org: I think I probably wrote this test before, but now that I look at this, it seems like it was wrong. What I should have done on (new) line 1905 is pass some garbage principal ID, and assert that full_neuron is not in the result. I think there is a bug where full_neuron does indeed appear in the result, which this test failed to catch originally.

rs/nns/nns.bzl
```diff
@@ -16,8 +16,8 @@ but other things could be added here later.
 CANISTER_NAME_TO_MAX_COMPRESSED_WASM_SIZE_E5_BYTES = {
     "cycles-minting-canister.wasm.gz": 6,
     "genesis-token-canister.wasm.gz": 3,
-    "governance-canister.wasm.gz": 17,
-    "governance-canister_test.wasm.gz": 20,
+    "governance-canister.wasm.gz": 24,
+    "governance-canister_test.wasm.gz": 24,
```
daniel-wong-dfinity-org: Wow! 17 -> 24?? This is a pretty big jump! Can you try to figure out why this is suddenly happening?? I'm really surprised if there is really such a big jump with this change. No new dependencies or anything!

rs/nns/governance/src/governance.rs
ShahriarJavidi: Absolutely right, but by keeping the governance, then the function `modify_follow` would be method on Governance and can be called by `self`. Otherwise, we would need to use an associated function.

rs/nns/governance/src/governance.rs
ShahriarJavidi: Nice point! Adopted.

rs/nns/governance/src/governance.rs
ShahriarJavidi: Thanks.

rs/nns/nns.bzl
ShahriarJavidi: 18 would be enough. I though of increasing the limit more to avoid future inconveniences. :D

rs/nns/governance/src/governance.rs
ShahriarJavidi: Adopted.

rs/nns/nns.bzl
daniel-wong-dfinity-org: No, we want to be "inconvenienced". What we actually want is to maintain awareness of our growth, because message size limits are not very generous.

Ofc, the real solution is to support chunking, but it would be nice to push that as far down the road as possible.

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: > can be called by `self`

I don't think this is necessarily a good thing. Yes, it is convenient, and this is what most people do. But it does not limit access to data very well.

FWIW, in the highly acclaimed book, [_Effective C++_ (by Scott Mayers)][C++], it says

[C++]: https://ptgmedia.pearsoncmg.com/images/9780321334879/samplepages/0321334876.pdf

> Prefer non-member non-friend functions to member functions.

Or, to put it another way, minimize methods, because they have access to privates. To reiterate, people don't really follow this, but I think they are mistaken.

Anyway, since people don't really follow this, I'm not going to block. I just want to make a second plea for you to consider this.

rs/nns/governance/tests/governance.rs
daniel-wong-dfinity-org: Can you use a different principal on line 1905, or at least add a TODO so that we make that change (+ bug fix) later?

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: Is this tested?

rs/nns/governance/src/governance.rs
```diff
@@ -3390,62 +3390,94 @@ impl Governance {
...
+            .unwrap_or_default();
+
+        for followee in &new_followees.followees {
+            if old_followees.contains(followee) {
+                // The follow relationship already exists, so it was already validated.
+                continue;
+            }
+
+            if let Ok((followee_visibility, followee_controller, followee_hot_keys)) = self
+                .with_neuron(followee, |neuron| {
```
daniel-wong-dfinity-org: Sorry, I forgot to show how to avoid loading auxiliary data. Instead of `with_neuron`, use `with_neuron_sections` with [`hot_keys`][aux] set to `true`, like so:

[aux]: https://sourcegraph.com/github.com/dfinity/ic@616e36391931f48f13515fad0df601d583e87047/-/blob/rs/nns/governance/src/storage/neurons.rs?L52

```
NeuronSections {
    hot_keys: true,
    ..NeuronSections::NONE
}
```

Maybe this doesn't really make a big difference in practice. Maybe this is premature optimization. Do as you see fit.

rs/nns/governance/src/governance.rs
```diff
@@ -3390,62 +3390,89 @@ impl Governance {
...
+        // Otherwise, update the entry with the new followees list.
+        // A new can follow another neuron if:
+        // 1. the followee neuron is a public neuron
+        // 2. or if the followee neuron is a private neuron, either
+        //      * they share a controller or
+        //      * the controller of the follower neuron is in the hot keys of the followee neuron.
+        // If in the list of followees, there are any follow relationships
+        // that don't adhere to the aforementioned rules, return a GovernanceError
+        // including all the invalid followees.
+        let mut invalid_followees = 0_032;
```
daniel-wong-dfinity-org:
```suggestion
        let mut invalid_followees = 0;
```

rs/nns/governance/src/governance.rs
```diff
@@ -3390,62 +3390,89 @@ impl Governance {
         if new_followees.followees.is_empty() {
```
daniel-wong-dfinity-org: Sorry. I think this issue was present before, but...

Isn't topic_to_followees from neuron? Isn't it redundant to pass topic_to_followees (in addition to neuron)?

rs/nns/governance/src/governance.rs
```diff
@@ -3390,62 +3390,89 @@ impl Governance {
...
+        // To avoid looking up the already existing followees of the neuron
+        // (which are not changing) we only validate the new followees.
+        let old_followees = topic_to_followees
+            .get(&topic)
+            .map(|f| f.followees.iter().collect::<HashSet<&NeuronId>>())
+            .unwrap_or_default();
+
+        for followee in &new_followees.followees {
+            if old_followees.contains(followee) {
+                // The follow relationship already exists, so it was already validated.
```
daniel-wong-dfinity-org: Maybe also mention grandfathering? To me, that is the main reason here.

rs/nns/governance/src/governance.rs
ShahriarJavidi: Makes sense. Adopted.

rs/nns/governance/src/governance.rs
ShahriarJavidi: As we also want the controller and visibility and `NeuronSections` doesn't have this yet, would change it once NeuronSections has other fields as well.

rs/nns/governance/src/governance.rs
```diff
@@ -8498,3 +8402,103 @@ impl TimeWarp {
         }
     }
 }
+
+fn modify_followees(
+    neuron_store: &NeuronStore,
+    neuron: &Neuron,
+    topic_to_followees: &HashMap<i32, Followees>,
```
daniel-wong-dfinity-org: I think you can just get this from neuron?

rs/nns/governance/src/governance.rs
```diff
@@ -6090,7 +5993,8 @@ impl Governance {
 
         let now_seconds = self.env.now();
         let new_neuron_followees = self.with_neuron(id, |neuron| {
```
daniel-wong-dfinity-org: I should have pointed this out earlier. Sorry.

To avoid looking up the same neuron again on line 6007, how about we put everything within one `with_neuron_mut` ?

rs/nns/governance/tests/governance.rs
daniel-wong-dfinity-org: > Can you use a different principal on line 1905, or at least add a TODO so that we make that change (+ bug fix) later?

I don't think this is done. Perhaps, you didn't push it?

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: Is there already a test for grandfathering private following?

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: I think those come standard; even if you pass NeuronSections::NONE, I think you still get those?

rs/nns/governance/src/governance.rs
ShahriarJavidi: There is no test for grandfathering.

rs/nns/governance/src/governance.rs
ShahriarJavidi: I will keep is as `with_neuron_sections` is currently private and making it public makes the compiler throw a warning that `type NeuronSections is more private than the item neuron_store::NeuronStore::with_neuron_sections`

rs/nns/governance/src/governance.rs
ShahriarJavidi: I think it doesn't work because `modify_followees` takes an immutable reference to the neuron but `with_neuron_mut`, as the name suggests, takes a mutable reference. It would then break the follow rules and compiler would yell.

rs/nns/governance/src/governance.rs
ShahriarJavidi: In `set_following` we are iteratively populating the `topic_to_followees` and returning it. When the iteration is over, we assign it to the followees of the neuron. Hence, the suggested optimisation wouldn't apply there.

rs/nns/governance/src/governance.rs
ShahriarJavidi: Will add the test in a follow-up PR.

rs/nns/governance/tests/governance.rs
```diff
@@ -1902,8 +1904,11 @@ fn test_enforce_private_neuron() {
     );
 
     // Step 3: Inspect results.
-    assert_eq!(neuron_info.visibility, Some(Visibility::Private as i32));
-    assert_eq!(full_neuron.visibility, Some(Visibility::Private as i32));
+    // TODO: Bug, fix later
```
daniel-wong-dfinity-org: We generally do `TODO(${JIRA_TICKET_ID})` to help make TODOs get done.

rs/nns/governance/src/governance.rs
daniel-wong-dfinity-org: I think it would work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=de788aff4ddd50cbd5d051ade606134c

PR #7015, opened 2025-09-30
feat(canister_logging): improve canister logging metrics

maksymar:
This PR improves canister logging metrics by adding new histogram buckets for total log memory usage and introducing a metric to track per-round delta log memory usage.

- The canister log now records the size of each appended delta log (supporting multiple deltas per round) and exposes `canister_log_delta_memory_usage` metric
  - The code retains and clears delta log sizes at the end of each round to prevent unbounded growth
- `canister_log_memory_usage_v3` has more histogram buckets comparing to `*_v2`

These changes improve observability and tracking of canister log memory usage.

EXC-2115

### DIFF
--- a/rs/monitoring/metrics/src/buckets.rs
+++ b/rs/monitoring/metrics/src/buckets.rs
@@ -2,27 +2,24 @@
 
 use std::cmp::Ordering;
 
-/// Create buckets using divisors of 10 multiplied by powers of 10, e.g.,
-/// ```text
-/// […, 0.1, 0.2, 0.5, 1, 2, 5, 10, 20, 50, …]
-/// ```
-///
-/// The buckets go from `10^min_power` to `5 × 10^max_power`, inclusively.
-/// The total number of buckets is `3 * (max_power - min_power + 1)`.
+/// Create log-scale buckets on a decimal 1–2–5 grid: `{1, 2, 5} × 10^n`.
+/// Range — from `10^min_power` up to `5×10^max_power` (inclusive).
+/// Count — `3 * (max_power - min_power + 1)`.
+/// Returns a strictly increasing `Vec<f64>`.
 ///
 /// # Examples
-///
 /// ```
 /// use ic_metrics::buckets::decimal_buckets;
 ///
-/// assert_eq!(vec![0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0], decimal_buckets(-1, 1));
-/// assert_eq!(vec![1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0], decimal_buckets(0, 2));
+/// assert_eq!(
+///     vec![0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0],
+///     decimal_buckets(-1, 1)
+/// );
 /// assert_eq!(3 * 21, decimal_buckets(-10, 10).len());
 /// ```
 ///
 /// # Panics
-///
-/// * Panics if min_power > max_power.
+/// Panics if `min_power > max_power`.
 pub fn decimal_buckets(min_power: i32, max_power: i32) -> Vec<f64> {
     assert!(
         min_power <= max_power,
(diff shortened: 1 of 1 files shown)


rs/execution_environment/src/scheduler/scheduler_metrics.rs
```diff
@@ -187,6 +190,56 @@ impl SchedulerMetrics {
...
+                    20 * MIB,
+                    50 * MIB,
+                    100 * MIB,
+                    200 * MIB,
+                    500 * MIB,
+                    GIB,
+                    2 * GIB,
+                    5 * GIB,
+                    10 * GIB,
+                ])
```
alin-at-dfinity: FYI, this is exactly what `decimal_buckets()` is for. There's also a function (I forget the name) to add an extra bucket (i.e. the zero).

rs/execution_environment/src/scheduler/scheduler_metrics.rs
```diff
@@ -187,6 +190,56 @@ impl SchedulerMetrics {
                     10 * MIB,
                 ])
             ),
+            canister_log_memory_usage_v3: metrics_registry.histogram(
+                "canister_log_memory_usage_bytes_v3",
+                "Canisters log memory usage distribution in bytes.",
```
alin-at-dfinity, line 189: You don't need another metric, you can simply update the existing one. Even though changing the buckets is not idea, it's also non-breaking. You're essentially adding new values to an existing label.

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -508,8 +508,10 @@ impl SystemStateModifications {
             system_state.global_timer = new_global_timer;
         }
 
-        // Append canister log.
-        system_state.canister_log.append(&mut self.canister_log);
+        // Append delta log to the total canister log.
+        system_state
+            .canister_log
+            .append_delta_log(&mut self.canister_log);
```
alin-at-dfinity, line 514: This method already takes a `logger` argument. You may as well give it a `metrics` (or just the one metric) argument and reduce a lot of the complexity of recording and wiping a collection of sizes by observing the size here (or within `append_delta_log()`).

rs/execution_environment/src/scheduler/scheduler_metrics.rs
maksymar: thanks for pointing that out, especially that there's a `*_with_zero` version. but unfortunately `decimal_buckets*` do not do **exactly** what I need, because the buckets are based on powers of 10, whereas memory sizes scale naturally in powers of 1024. if we used `decimal_buckets` the buckets would be shown at the wrong positions, eg 2 KB (==1.024 KB) vs 1024 (1KiB) or 10 KB (==8.192KB) vs 8192 (8KiB).

maybe it'd make sense to have a parallel set of functions that follow the same pattern as `decimal_buckets` but with base 1024: `binary_buckets` and `binary_buckets_with_zero`.

WDYT?

alin-at-dfinity: Sure, feel free to add a `binary_buckets()` function.

maksymar: Done, added `binary_buckets*` and switched to it.

rs/execution_environment/src/scheduler/scheduler_metrics.rs
maksymar: After offline discussion I decided to go purely with binary buckets (not a mix of 1-2-5 decimal scale for binary values like 1KiB, 2KiB, 5KiB etc). So it's going to be a significant change in bucket labels, therefore I need to go with v2-to-v3 approach.

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
maksymar: As discussed offline propagating metrics was an another option to implement this, but I decided to go with storing metrics data next to canister logs for simplicity (no extra plumbing needed) and therefore lower maintenance cost.

rs/execution_environment/src/scheduler/scheduler_metrics.rs
alin-at-dfinity: Unless you are concerned about how the transition looks like e.g. in a heatmap panel on a dashboard, there is no need to. `histogram_quantile()` (among others) will work just fine across this kind of transition.

I would simply change the buckets of the existing metric, but again, your choice.

PR #7017, opened 2025-09-30
chore(node): clean up nested test and improve code reuse

andrewbattat:


rs/tests/nested/guestos_upgrade.rs
```diff
@@ -103,13 +89,13 @@ pub fn upgrade_guestos(env: TestEnv) {
             &guest_ipv6,
             &original_version,
             &logger,
-            Duration::from_secs(5 * 60),
```
andrewbattat, line 106: 60 seconds should be more than enough to get the expected version

rs/tests/nested/guestos_upgrade.rs
```diff
@@ -119,44 +105,27 @@ pub fn upgrade_guestos(env: TestEnv) {
...
-        );
-
-        // Check that GuestOS is on the expected version (upgrade version)
         wait_for_expected_guest_version(
             &client,
             &guest_ipv6,
             &target_version,
             &logger,
-            Duration::from_secs(7 * 60), // Long wait for GuestOS upgrade to apply and reboot
-            Duration::from_secs(5),
```
andrewbattat, line 159: NODE_UPGRADE_TIMEOUT is 5 minutes, not 7. We bumped the timeout from 5 minutes to 7 in an attempt to fix flakiness, but we've since resolved these issues, and can bring the timeout back down to something more appropriate.

PR #7018, opened 2025-09-30
perf(nns-recovery): reduce resource usage of NNS recovery system tests

pierugo-dfinity:
This PR enables the configuration of resources allocated to nested VMs in system tests. It also reduces the number of vCPUs per VM from 32 to 8 in NNS recovery tests in an effort to reduce their resource usage. AFAIU it cannot be lower than 8 since the final number of CPUs allocated to HostOS is this number divided [by 2](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/driver/src/driver/bootstrap.rs?L728), divided [by 4](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/ic_os/os_tools/guest_vm_runner/templates/guestos_vm_template.xml?L18).

Before:
- `nr_large`:  6 (upstreams VM) + 6 (original single NNS node) + 6 (API BN) + 6 (IcGateway VM) + 2 (vector VM) + *32* * 40 (40 nested VMs for the new NNS nodes) = 1306 vCPUs
- Other `nr_*`: 6 (upstreams VM) + 6 (original single NNS node) + 6 (API BN) + 6 (IcGateway VM) + 2 (vector VM) + *32* * 4 (4 nested VMs for the new NNS nodes) = 154 vCPUs
image

After:
- `nr_large`: 6 (upstreams VM) + 6 (original single NNS node) + 6 (API BN) + 6 (IcGateway VM) + 2 (vector VM) + *8* * 40 (40 nested VMs for the new NNS nodes) = 346 vCPUs
- Other `nr_*`: 6 (upstreams VM) + 6 (original single NNS node) + 6 (API BN) + 6 (IcGateway VM) + 2 (vector VM) + *8* * 4 (4 nested VMs for the new NNS nodes) = 58 vCPUs
image

### DIFF
--- a/rs/tests/driver/src/driver/bootstrap.rs
+++ b/rs/tests/driver/src/driver/bootstrap.rs
@@ -736,9 +736,9 @@ fn create_setupos_config_image(
                 urls: vec![nns_url.clone()],
             },
             vm_resources: deployment_json::VmResources {
-                memory: Some((vm_spec.memory_ki_b / 2 / 1024 / 1024) as u32),
+                memory: (vm_spec.memory_ki_b / 2 / 1024 / 1024) as u32,
                 cpu: cpu.to_string(),
-                nr_of_vcpus: Some((vm_spec.v_cpus / 2) as u32),
+                nr_of_vcpus: (vm_spec.v_cpus / 2) as u32,
             },
         },
     )


rs/tests/driver/src/driver/resource.rs
```diff
@@ -390,14 +390,27 @@ fn vm_spec_from_nested_node(
     node: &NestedNode,
     default_vm_resources: Option<VmResources>,
 ) -> VmSpec {
+    let vm_resources = &node.vm_resources;
     VmSpec {
         name: node.name.clone(),
         // Note that the nested GuestOS VM uses half the vCPUs and memory of this host VM.
```
pierugo-dfinity, line 396:
```suggestion
        // Note that the nested HostOS VM uses half the vCPUs and memory of this host VM.
```
Isn't it HostOS that gets half the vCPUs? We write the deployment config containing half the vCPUs [here](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/ic_os/dev_test_tools/setupos-image-config/src/lib.rs?L127-128) and later read from that file [here](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/ic_os/config/src/main.rs?L141) setting HostOS's number of CPUs to that number.

basvandijk: I don't think so. The HostOS setting `vm_nr_of_vcpus` [configures](https://sourcegraph.com/github.com/dfinity/ic@5300df4673ae591b57a2b5b38baa6690d0f28165/-/blob/rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs?L119) the number of vCPUs for the GuestOS (nested) VM.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -40,6 +40,12 @@ use sha2::{Digest, Sha256};
 use slog::{Logger, info};
 use tokio::task::JoinSet;
 
+pub const NNS_RECOVERY_VM_RESOURCES: VmResources = VmResources {
+    vcpus: Some(NrOfVCPUs::new(8)),
```
basvandijk, line 44: From experience I know guestOS VMs need to have a minimum of 6 vCPUs to operate stably. That's why we [default to it](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/driver/src/driver/resource.rs?L28). Do know this was experimentally determined years ago so things could have improved in the mean time that could allow us to go to fewer vCPUs. 

(For reference Farm uses [this calculation](https://github.com/dfinity-lab/farm/blob/master/backend/Farm/VMManager.hs#L442-L444) to determine the number of vCPU sockets, cores and threads).

So with 8 vCPUs given to the hostOS VM the nested guestOS VM only gets 4 vCPUs. It might work but please do test it thoroughly by running the test many times (i.e. with `--runs_per_test=10`) to spot any flakiness this might introduce.

When in doubt I would go for 16.

basvandijk (approved):

rs/tests/driver/src/driver/resource.rs
pierugo-dfinity: Ah right, thank you!

rs/tests/nested/nns_recovery/common.rs
basvandijk: Something else which we could consider is not doing the [halving](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/driver/src/driver/bootstrap.rs?L728) by either: 
* giving the nested VM as many vCPUs as the host VM (like in production) 
* or giving it slightly less but not halve (like `vcpus - 2` or something and setting `vcpus` to 10).

pierugo-dfinity: Alright, will do.

Why does choosing 6 vCPUs actually work? I thought the number needed to be divisible by 4.

basvandijk: > Why does choosing 6 vCPUs actually work? I thought the number needed to be divisible by 4.

Good question. Farm uses [this calculation](https://github.com/dfinity-lab/farm/blob/master/backend/Farm/VMManager.hs#L442-L444) to determine the number of vCPU sockets, cores and threads. So when vCPUs is set to 6 the vCPU [topology](https://github.com/dfinity-lab/farm/blob/master/backend/libvirt-templates/production.xml#L15) will be:
```
<topology sockets='1' cores='6' threads='1'/>
```

eichhorl (approved):

andrewbattat (approved):
Stamping! I'm not sure the impact on flakiness, if any, but I see that's already being discussed.

rs/tests/nested/nns_recovery/common.rs
pierugo-dfinity: > It might work but please do test it thoroughly by running the test many times (i.e. with --runs_per_test=10) to spot any flakiness this might introduce.

As said on Slack already, `nr_broken_dfinity_node`, `nr_local`, `nr_remote` and `nr_large` all ran 10 times without issues.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -40,6 +40,12 @@ use sha2::{Digest, Sha256};
...
+    memory_kibibytes: Some(AmountOfMemoryKiB::new(25165824)), // 24GiB
```
basvandijk, line 45: Should we also lower `HOSTOS_MEMORY_KIB_PER_VM`?

pierugo-dfinity: I'm not sure whether we also experienced flakiness in the other nested tests (not NNS recovery ones)... Do you think we should go for it just in case?

basvandijk: Oh right, the other nested tests also use this. Let's leave it for now then to not change as many things at once. But if the other nested tests also work well with 24GiB we should go with that eventually to not have as many different resource configurations.

PR #7021, opened 2025-09-30
feat(nns): Allow updating known neuron through RegisterKnownNeuron

jasonz-dfinity:
# Why

Previously, a `RegisterKnownNeuron` proposal can still be made and executed if the neuron is already a known neuron, as long as the "name" is different. This is considered a hack because it technically allows updating an existing a known neuron, while the name check was meant for registering a new known neuron.

On the other hand, it's reasonable to allow "upsert" semantic for RegisterKnownNeuron proposal type, and the recent addition of `committed_topics` and `links` make this PR necessary - existing known neurons would be able to update their information through proposals.

# What

* Relax the "known neuron name existence" check by not failing when both the name and neuron id are the same.


ic-admin
basvandijk, line 1: Was this accidentally committed?

jasonz-dfinity: Yes. Thanks for catching it!

rs/nns/governance/src/known_neuron_index.rs
```diff
@@ -110,13 +110,9 @@ impl<M: Memory> KnownNeuronIndex<M> {
     }
 
     /// Checks whether the known neuron name already exists in the index.
```
max-dfinity: Comment should be updated

jasonz-dfinity: Done

PR #7028, opened 2025-10-01
fix: subnet memory taken for canisters with memory allocation

mraszyk:
This PR fixes subnet memory taken for canisters with memory allocation: so far their contribution to the `execution` field in `MemoryTaken` was the sum of their memory allocation, canister history, wasm chunk store, and canister snapshots memory usage. However, this leads to accounting for canister history, wasm chunk store, and canister snapshots memory usage twice - once in the memory allocation and one more time separately. This PR fixes that by taking the maximum of the canister memory usage (execution, canister history, wasm chunk store, and canister snapshots memory usage) and memory allocation.

This PR also adds a regression test:
- fill a subnet with canisters having 100 GiB of memory allocation each;
- make the memory allocation of the last canister best-effort;
- grow their stable memory to 40 GiB and take a canister snapshot (all within the memory allocation of 100 GiB);
- because of the double accounting the subnet claims to have no more available memory left for the best-effort canister although no canister with memory allocation of 100 GiB exceeds its memory allocation and thus at least 100 GiB (the original memory allocation of the best-effort canister) are definitely available.

### DIFF
--- a/rs/execution_environment/tests/canister_snapshots.rs
+++ b/rs/execution_environment/tests/canister_snapshots.rs
@@ -1005,10 +1005,11 @@ fn load_canister_snapshot_works_on_another_canister() {
 fn canister_snapshots_and_memory_allocation() {
     let env = StateMachine::new();
 
+    // We first fill the subnet with canisters having 100 GiB of memory allocation each.
     let mut canisters = vec![];
     loop {
         let settings = CanisterSettingsArgsBuilder::new()
-            .with_memory_allocation(100_000_000_000)
+            .with_memory_allocation(100 << 30)
             .build();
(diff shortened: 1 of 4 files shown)


michael-weigelt (approved):

pakhomov-dfinity (approved):
Where do we compute the maximum of allocation and usage?

mraszyk:
> Where do we compute the maximum of allocation and usage?

```
canister
                        .memory_allocation()
                        .allocated_bytes(canister.memory_usage())
```
computes the actually allocated bytes as the maximum of pre-allocated bytes (canister memory allocation) and actually used bytes `canister.memory_usage()`.

PR #7033, opened 2025-10-01
feat(canister_logging): charge for fetching logs via update call

maksymar:
This PR implements charging for fetching canister logs for update calls (when feature is enabled).

The actual fees are just placeholders for now, should be adjusted after benchmarking the actual costs.

When charging for logs user is required to provide enough cycles to cover a full response 2MB, but it only charges the actual amount of bytes returned.

EXC-2114


rs/cycles_account_manager/src/lib.rs
```diff
@@ -1370,6 +1371,34 @@ impl CyclesAccountManager {
     pub fn default_reserved_balance_limit(&self) -> Cycles {
         self.config.default_reserved_balance_limit
     }
+
+    pub fn fetch_canister_logs_fee(
+        &self,
+        response_size: usize,
```
michael-weigelt: I think this should be `NumBytes`.

maksymar: Done.

rs/execution_environment/src/execution_environment.rs
```diff
@@ -1570,42 +1570,60 @@ impl ExecutionEnvironment {
...
+                                            fetch_canister_logs(
+                                                sender,
+                                                &state,
+                                                args,
+                                                self.config.fetch_canister_logs_filter,
+                                            )
+                                        })
+                                        .map(|resp| {
+                                            let response_bytes = Encode!(&resp).unwrap();
+                                            Arc::make_mut(request).payment -= self
```
michael-weigelt: I don't understand this line. I could not find any further use of this request.payment, so what is the intended effect?

michael-weigelt: Even in `finish_subnet_message_execution` and downstream I found no use site.

maksymar: This is to address Rust borrow checker rules -- in `execute_subnet_message` function most of the ic00 methods are not paid, so they do not need to modify `request.payment` which is automatically calculated when `msg.take_cycles()` is called.

But for a few methods (like http outcalls, ecdsa signatures) it must be paid, and payment must be modified, which can only be done if initial `msg` is matched as mutable reference `match &mut msg`, which makes in `CanisterCall::Request(request)` request a mutable `Arc`, but the underlying data is not mutable, so you have to call `Arc::make_mut` so that it clones underlying data to a new allocation to ensure unique ownership (clone-on-write) and now payment is modified. Same trick is done for http outcalls and threshold signature methods.

maksymar: I extracted this logic into a separate `CanisterCall::deduct_cycles(fee)` method, this should improve readability.

michael-weigelt (approved):

PR #7034, opened 2025-10-01
feat(nns-recovery): [CON-1597] allow to upgrade without first electing the version

pierugo-dfinity:
This PR allows to upgrade to an already elected version in NNS recoveries, which could be useful if we simply need to revert back to a previous version (that is already elected).

We also add a new system test (`nr_no_bless`) that tests this feature. Ticket CON-1599 will add a similar system test to application subnet recoveries.


rs/tests/nested/nns_recovery/nr_no_bless.rs
```diff
@@ -0,0 +1,54 @@
...
+                    impersonate_upstreams: true,
+                    subnet_size: SUBNET_SIZE,
+                    dkg_interval: DKG_INTERVAL,
+                },
+            )
+        })
+        .add_test(systest!(test; TestConfig {
+            local_recovery: false,
+            break_dfinity_owned_node: false,
+            add_and_bless_upgrade_version: false,
```
eichhorl, line 47: Something else we could consider to conserve resources is to combine multiple test cases into one, as long as we are convinced they shouldn't affect each other. For instance, we could bless a new version in the local test, and upgrade to an existing version in the remote test.

pierugo-dfinity: Thanks for the suggestion! Then I'll just remove `nr_remote`, which is equivalent to what you're suggesting. I also thought that `nr_remote` did not test any "special" scenario.

PR #7035, opened 2025-10-01
feat(ckdoge): Dogecoin address

gregorydemay:
Add the Dogecoin address type to the ckDOGE minter that will be used by the minter to derive addresses for the users to deposits DOGE and as recipient addresses when withdrawing.

This is different from the [BitcoinAddress](https://github.com/dfinity/ic/blob/fb4dff62d3b2b9261e9d486c743b99b0eb7ef438/rs/bitcoin/ckbtc/minter/src/address.rs#L20) type used by the ckBTC minter:

1. Only pay to public key hash ([P2PKH](https://learnmeabitcoin.com/technical/script/p2pkh/)) or pay to script hash ([P2SH](https://learnmeabitcoin.com/technical/script/p2sh/)) address types are supported.
2. Dogecoin uses different 1-byte prefix for both address types.

### DIFF
--- a/rs/dogecoin/ckdoge/minter/src/address/tests.rs
+++ b/rs/dogecoin/ckdoge/minter/src/address/tests.rs
@@ -52,8 +52,7 @@ fn should_fail_to_parse_invalid_addresses() {
 }
 
 fn test_vectors<T: DeserializeOwned>(filename: &str) -> T {
-    // Return something like
-    // "rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_valid.json"
+    // Return something like "rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_valid.json"
     let test_vector_files = std::env::var("TEST_VECTORS").expect(
         "environment variable 'TEST_VECTORS' should be a space-separated list of test vectors",
     );


rs/dogecoin/ckdoge/minter/src/address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        match (bytes[0], network) {
+            (DOGE_MAINNET_PREFIX, Network::Mainnet)
+            | (DOGE_TESTNET_PREFIX, Network::Testnet)
+            | (DOGE_REGTEST_PREFIX, Network::Regtest) => Ok(Self::P2pkh(data)),
+            (DOGE_MAINNET_P2SH_PREFIX, Network::Mainnet)
+            | (DOGE_TESTNET_P2SH_PREFIX, Network::Testnet)
+            | (DOGE_REGTEST_P2SH_PREFIX, Network::Regtest) => Ok(Self::P2sh(data)),
+            (DOGE_MAINNET_PREFIX, _) | (DOGE_MAINNET_P2SH_PREFIX, _) => {
+                Err(ParseAddressError::WrongNetwork {
+                    expected: Network::Mainnet,
```
ninegua: nit: Shouldn't the "expected network" come from the `network` parameter instead of `bytes[0]`?

gregorydemay: yes, good point, the network parameter will come from the canister, so you're totally right.

rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json
```diff
@@ -0,0 +1,152 @@
+[
```
mducroux, line 1: understanding: does this contain Bitcoin/Litecoin priv/pubkey addresses as well? Could you add comments to show which ones are?

rs/dogecoin/ckdoge/minter/src/address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+pub enum ParseAddressError {
+    UnsupportedAddressType,
+    WrongNetwork { expected: Network, actual: Network },
+    MalformedAddress(String),
+    UnexpectedHumanReadablePart { expected: String, actual: String },
+    NoData,
+}
+
+impl DogecoinAddress {
+    pub fn parse(address: &str, network: &Network) -> Result<Self, ParseAddressError> {
```
mducroux, line 34: understanding: why implementing this yourself when parsing addresses is already implemented in the [rust-dogecoin library](https://github.com/dfinity/rust-dogecoin/blob/b24563ead9c61522e05b144d527c4b114befc301/bitcoin/src/dogecoin/address/mod.rs#L533)?

rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json
gregorydemay: See the comments [here](https://github.com/dfinity/ic/blob/a7b083c255616508d391a069c870a6b5b71c7e14/rs/dogecoin/ckdoge/minter/src/address/tests.rs#L9). I want to leave these files as is, since they are copied form the Dogecoin repo.

mducroux: Could you maybe add a comment that this file comes from the Dogecoin repo?

rs/dogecoin/ckdoge/minter/src/address/mod.rs
gregorydemay: Similar approach as the ckBTC minter which has the [BitcoinAddress](https://github.com/dfinity/ic/blob/0756b99d225d3b0526e32d9c236492b367716c3b/rs/bitcoin/ckbtc/minter/src/address.rs#L72) type. I think the motivation is to avoid a dependency on `rust-bitcoin`. Similarly here, we avoid a dependency on rust-dogecoin.

rs/dogecoin/ckdoge/minter/src/address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        if address.len() > 125 {
```
mducroux: Shouldn't this be `> 50` like [here](https://github.com/dfinity/rust-dogecoin/blob/b24563ead9c61522e05b144d527c4b114befc301/bitcoin/src/dogecoin/address/mod.rs#L534).

rs/dogecoin/ckdoge/minter/src/address/mod.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(test)]
+mod tests;
+
+use crate::lifecycle::init::Network;
+
+// See https://github.com/dogecoin/dogecoin/blob/7237da74b8c356568644cbe4fba19d994704355b/src/chainparams.cpp#L167
+const DOGE_MAINNET_PREFIX: u8 = 30;
```
mducroux: why not adding `P2PKH` to this const name?

rs/dogecoin/ckdoge/minter/src/address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+                    expected: *network,
+                })
+            }
+            (DOGE_TESTNET_PREFIX, _) | (DOGE_TESTNET_P2SH_PREFIX, _) => {
+                Err(ParseAddressError::WrongNetwork {
+                    actual: Network::Testnet,
+                    expected: *network,
+                })
+            }
+            (DOGE_REGTEST_PREFIX, _) => Err(ParseAddressError::WrongNetwork {
```
mducroux: Isn't `DOGE_REGTEST_P2SH_PREFIX` missing here?

rs/dogecoin/ckdoge/minter/src/address/tests.rs
```diff
@@ -0,0 +1,121 @@
+use crate::address::DogecoinAddress;
+use crate::lifecycle::init::Network;
+use assert_matches::assert_matches;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use std::io::BufReader;
+use std::path::PathBuf;
+
+//https://github.com/dogecoin/dogecoin/blob/2c513d0172e8bc86fe9a337693b26f2fdf68a013/src/test/data/base58_keys_valid.json
```
mducroux: missing whitespace

rs/dogecoin/ckdoge/minter/src/address/tests.rs
```diff
@@ -0,0 +1,121 @@
...
+    for test_case in test_cases {
+        for network in &ALL_NETWORKS {
+            assert_matches!(DogecoinAddress::parse(&test_case.0, network), Err(_));
+        }
+    }
+}
+
+fn test_vectors<T: DeserializeOwned>(filename: &str) -> T {
+    // Return something like
+    // "rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_valid.json"
```
mducroux: weird formatting

rs/dogecoin/ckdoge/minter/test_vectors/base58_keys_invalid.json
gregorydemay: I updated the comments I linked to. I would like to avoid modifying the `.json` files because then I would need to tweak the deserialization to skip comments.

rs/dogecoin/ckdoge/minter/src/address/mod.rs
gregorydemay: Good point, thanks for the pointer.

rs/dogecoin/ckdoge/minter/src/address/mod.rs
gregorydemay: sounds good, I was following the same style as ckBTC, but I also thinks it helps readability adding that.

rs/dogecoin/ckdoge/minter/src/address/mod.rs
gregorydemay: no, it's because `DOGE_REGTEST_P2SH_PREFIX == DOGE_TESTNET_P2SH_PREFIX`

rs/dogecoin/ckdoge/minter/src/address/tests.rs
gregorydemay: Done.

rs/dogecoin/ckdoge/minter/src/address/tests.rs
gregorydemay: Done.

PR #7043, opened 2025-10-02
chore: rename test_macos to test_all_platforms and test on arm64-linux

mraszyk:
This PR renames the bazel tag `test_macos` to `test_all_platforms` and unifies arm64-darwin and arm64-linux jobs so that all tests tagged with `test_all_platforms` run on both arm64-darwin and arm64-linux.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -302,7 +302,7 @@ jobs:
             bazel build \
               --config=stamped \
               --invocation_id="${{ fromJSON(steps.gen-uuids.outputs.output).build_id }}" \
-              --build_tag_filters=test_macos \
+              --build_tag_filters=test_all_platforms \
               //... --nobuild
 
       - name: Build & Test
@@ -313,7 +313,7 @@ jobs:
             bazel test \
               --config=stamped \
               --invocation_id="${{ fromJSON(steps.gen-uuids.outputs.output).test_id }}" \
-              --test_tag_filters=test_macos \
+              --test_tag_filters=test_all_platforms \
               //...
 
       - name: Upload artifacts
@@ -353,16 +353,24 @@ jobs:
 
       - name: Build and Test
         run: |
-          # run pocket-ic tests
+          # Setup zig-cache
+          mkdir -p /tmp/zig-cache
+
+          bazel_targets=(
+            # make sure codebase builds for local development
+            //rs/...
+            //publish/binaries:pocket-ic.gz
+
+            # necessary artifacts & tests for pocket -ic
+            //packages/pocket-ic:all
+          )
+
           bazel \
             --noworkspace_rc \
             --bazelrc=./bazel/conf/.bazelrc.build --bazelrc=/tmp/bazel-cache.bazelrc \
             test \
-              //packages/pocket-ic:all \
-              //rs/pocket_ic_server:test \
-              //rs/pocket_ic_server:gateway \
-              //rs/pocket_ic_server:pocket-ic-server \
-              //publish/binaries:pocket-ic.gz
+            --test_tag_filters="test_all_platforms,test_all_platforms_slow" \
+            "${bazel_targets[@]}"
 
           mkdir -p build
           cp \
@@ -396,22 +404,17 @@ jobs:
           bazel_targets=(
             # make sure codebase builds for local development
             //rs/...
-            //publish/binaries/...
+            //publish/binaries:pocket-ic.gz
 
             # necessary artifacts & tests for pocket -ic
-            //packages/pocket-ic:all \
-            # NOTE: technically covered by //rs/... above, but
-            # added explicitly for clarity
-            //rs/pocket_ic_server:test \
-            //rs/pocket_ic_server:gateway \
-            //rs/pocket_ic_server:pocket-ic-server
+            //packages/pocket-ic:all
           )
 
           bazel \
             --noworkspace_rc \
             --bazelrc=./bazel/conf/.bazelrc.build --bazelrc=/tmp/bazel-cache.bazelrc \
             test \
-            --test_tag_filters="test_macos,test_macos_slow" \
+            --test_tag_filters="test_all_platforms,test_all_platforms_slow" \
             "${bazel_targets[@]}"
 
           mkdir -p build
--- a/packages/pocket-ic/BUILD.bazel
+++ b/packages/pocket-ic/BUILD.bazel
@@ -77,7 +77,7 @@ rust_test(
         # TODO: remove 'requires-network' tag when the root cause for sporadic error below on Apple Silicon is identified and fixed.
         # Failed to crate http gateway: Failed to bind to address 127.0.0.1:0: Operation not permitted (os error 1)
         "requires-network",
-        "test_macos",
+        "test_all_platforms",
     ],
     deps = [
         # Keep sorted.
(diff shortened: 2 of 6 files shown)


rs/execution_environment/tests/hypervisor.rs
```diff
@@ -9691,7 +9691,7 @@ fn canister_balance_is_roughly(
     );
 }
 
-#[cfg(not(all(target_arch = "aarch64", target_vendor = "apple")))]
+#[cfg(all(target_arch = "x86_64", target_os = "linux"))]
```
mraszyk, line 9694: Should we maybe just disable `aarch64`? Currently, macOS Intel is also disabled.

berestovskyy: We want to run the test on the platforms where the new signal handler is available, which is x86 Linux only.

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,99 @@ jobs:
...
       - name: Set up Bazel cache
         run: |
           # Creates a bazelrc configuration fragment which tells bazel where the cache lives.
           nsc bazel cache setup --bazelrc=/tmp/bazel-cache.bazelrc
+
       - uses: actions/checkout@v4
+
+      - name: Set up zig cache
+        run: |
+          mkdir -p /tmp/zig-cache
```
nmattia: interesting that this wasn't required on linux before. I think we could add an `if: matrix.os == "darwin"`  here or even check if it's still necessary

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,99 @@ jobs:
...
-    steps:
-      - name: Set up Bazel cache
+      # Ensures that local development works on arm64-darwin.
+      - name: Build the entire codebase on darwin
+        if: ${{ matrix.os == 'darwin' }}
         run: |
-          # Creates a bazelrc configuration fragment which tells bazel where the cache lives.
-          nsc bazel cache setup --bazelrc=/tmp/bazel-cache.bazelrc
-      - uses: actions/checkout@v4
+          # We do not use `/...` here to exclude `//publish/...` which is not required for local development.
```
nmattia:
```suggestion
          # We do not use `//...` here to exclude `//publish/...` which is not required for local development.
```

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,99 @@ jobs:
...
+          bazel_build_targets=(
```
nmattia, line 382: I think we can inline those in the command for simplicity, or am I missing something?

MODULE.bazel
```diff
@@ -854,6 +854,23 @@ filegroup(
     ],
 )
 
+http_archive(
+    name = "bitcoin_core_linux_aarch64",
```
nmattia, line 858: do you know why this wasn't needed before? do we care about building those on aarch64-linux?

nmattia (commented):
Ohh super nice! Looking really good. Left a couple comments and will have another deeper look in a bit to double check this is not breaking artifact upload in any way but overall LGTM.

MODULE.bazel
mraszyk: because we explicitly listed the following test targets before:
```
//rs/pocket_ic_server:test
//rs/pocket_ic_server:gateway
```
but not `//rs/pocket_ic_server:bitcoin_integration_tests`.

.github/workflows/ci-main.yml
mraszyk: I just preserved the structure from before: they could have been inlined already so I wasn't sure if there's any style guideline for that.

.github/workflows/ci-main.yml
mraszyk: I'm giving it a try now: https://github.com/dfinity/ic/pull/7043/commits/8cef24df0e366a4f1b7f198868747663b5d55479

mraszyk: And it works!

mraszyk:
> this is not breaking artifact upload in any way

I checked that all combinations
```
https://download.dfinity.systems/ic/dcc124dcfb557957da71562265f79329a5220c97/binaries/{arm64,x86_64}-{darwin,linux}/pocket-ic.gz
```
work on the green commit just before I marked this PR as ready for review.

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,95 @@ jobs:
...
+    name: Bazel Test arm64-${{ matrix.os }}
+    runs-on: ${{ matrix.profile }}
+    timeout-minutes: 120
     if: github.repository == 'dfinity/ic' # only run on public repo, not private since Namespace runners are not configured there, so these CI jobs get stuck otherwise.
     steps:
       - name: Set up Bazel cache
         run: |
           # Creates a bazelrc configuration fragment which tells bazel where the cache lives.
           nsc bazel cache setup --bazelrc=/tmp/bazel-cache.bazelrc
+
```
nmattia:
```suggestion
```

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,95 @@ jobs:
...
       - uses: actions/checkout@v4
+
```
nmattia:
```suggestion
```

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,95 @@ jobs:
...
             test \
-            --test_tag_filters="test_macos,test_macos_slow" \
-            "${bazel_targets[@]}"
-
-          mkdir -p build
-          cp \
-            ./bazel-bin/publish/binaries/pocket-ic.gz \
-            ./build/pocket-ic-server-arm64-darwin.gz
+            --test_tag_filters="test_all_platforms,test_all_platforms_slow" \
+            --build_tests_only \
```
nmattia, line 406: what does this flag do? I'm struggling to make sense of the doc:

```
--[no]build_tests_only

If specified, Bazel will build only what is necessary to run the *_test and test_suite rules that were not filtered due to their [size](https://bazel.build/docs/user-manual#test-size-filters), [timeout](https://bazel.build/docs/user-manual#test-timeout-filters), [tag](https://bazel.build/docs/user-manual#test-tag-filters), or [language](https://bazel.build/docs/user-manual#test-lang-filters). If specified, Bazel will ignore other targets specified on the command line. By default, this option is disabled and Bazel will build everything requested, including *_test and test_suite rules that are filtered out from testing. This is useful because running bazel test --build_tests_only foo/... may not detect all build breakages in the foo tree.
```

.github/workflows/ci-main.yml
```diff
@@ -332,106 +332,95 @@ jobs:
...
-  bazel-test-arm64-darwin:
-    name: Bazel Test macOS Apple Silicon
-    timeout-minutes: 120
-    runs-on: namespace-profile-darwin # profile created in namespace console
-    if: github.repository == 'dfinity/ic' # only run on public repo, not private since Namespace runners are not configured there, so these CI jobs get stuck otherwise.
-    steps:
-      - name: Set up Bazel cache
+      # Ensures that local development works on arm64-darwin.
+      - name: Build the entire codebase on darwin
+        if: ${{ matrix.os == 'darwin' }}
```
nmattia, line 379: I think we could do it on linux as well; we're building most of these targets anyway with `bazel test` below

nmattia (approved):
Looking great! Couple thoughts

.github/workflows/ci-main.yml
mraszyk: All targets are not gonna build on arm64-linux.

.github/workflows/ci-main.yml
mraszyk: I specify a wildcard path (`//rs/...` and `//packages/...`) and filter using test tags. However, all paths matching the wildcard are gonna be built even if test tags are provided (test tags are only used to not run all tests, they do not influence what gets built). Since not all targets build on linux-arm64, I'm passing `--build_tests_only` so only build test targets passing filters.

PR #7051, opened 2025-10-02
feat: NNS Proposals to pause/unpause migrations

schneiderstefan:
This commit introduces NNS proposals to call the `enable_api` and `disable_api` endpoints on the canister migration canister. These proposals will be called as part of subnet splitting to ensure that subnet splitting and canister migrations do not overlap.

In addition, the commit tests the proposals by executing them as part of the canister migration tests and modifies the migrations canister to allow privileged calls from the governance canister.

### DIFF
--- a/Cargo.Bazel.json.lock
+++ b/Cargo.Bazel.json.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "18900e59ecae1f09a87ae378267b24738dc0c59e84dc0b93d17abde71d758c61",
+  "checksum": "c086739eeff79c4623fa65357317e3bdd5a3c8f6e12b9a502bec39f965ccadd3",
   "crates": {
     "abnf 0.12.0": {
       "name": "abnf",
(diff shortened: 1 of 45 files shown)


rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs
```diff
@@ -4707,6 +4707,10 @@ pub enum NnsFunction {
     DeployHostosToSomeNodes = 51,
     /// The proposal requests a subnet rental.
     SubnetRentalRequest = 52,
+    /// Instruct the migration canister to not accept any more migration requests.
+    PauseMigrations = 53,
```
max-dfinity: Can we make this "PauseCanisterMigrations"? Governance may interact in the future with some other canister that handles some other kind of migration.

Similarly, UnpauseCanisterMigrations?

schneiderstefan: Good point. Done.

PR #7052, opened 2025-10-02
chore: Remove old BitVec logic from incremental manifest computation

pakhomov-dfinity:


rs/state_manager/src/manifest.rs
```diff
@@ -702,38 +723,11 @@ fn hash_plan(
...
-                .ok()?;
-            Some((dirty_chunk_bitmap, base_index))
-        };
-
-        if let Some((dirty_chunk_bitmap, base_index)) = compute_dirty_chunk_bitmap() {
-            debug_assert_eq!(num_chunks, dirty_chunk_bitmap.len());
-
+        if let Some(base_index) = get_base_index(relative_path, base_manifest)
+            && files_with_same_inodes.contains(relative_path)
+        {
```
ShuoWangNSL: I think separating these two conditions will make the code clearer and more efficient.
`files_with_same_inodes.contains(relative_path)` is a relatively cheaper operation and can come first to filter paths not in the base_manifest.

After filtering, `get_base_index()` is expected to return `Some` value and we can add a warning if not.

rs/state_manager/src/manifest.rs
ShuoWangNSL: Not directly related to this PR. We are able to calculate the size of final result as in `default_hash_plan()`.  It will be good to preallocate via `with_capacity()`.

ShuoWangNSL, line 758: Sorry for another comment not directly related to this PR. I think `extend()` or `resize() `might be a better fit here.

We could handle these minor improvements in another PR—I just wanted to note it before I forget.

ShuoWangNSL (approved):

rs/state_manager/src/manifest.rs
pakhomov-dfinity: I cannot do `filter()` cause we handle all files, some in the else branch; I tried to add a `debug_assert` for finding a base index, but I feel it makes it worse cause it increases the nesting level

rs/state_manager/src/manifest.rs
pakhomov-dfinity: `extend()` takes a slice, which I cannot construct cause the length is defined at runtime; resize takes length arithmetic which I find dangerous. I could do `extend(vec![...])`, but it's less efficient, which is fine but then why do we bother with capacity.

ShuoWangNSL: That's true. If we do `with_capacity` in the beginning, this comment is not relevant anymore.

PR #7054, opened 2025-10-02
fix(nns-recovery): fix broken DFINITY-owned node edge-cases

pierugo-dfinity:


rs/tests/nested/nns_recovery/common.rs
```diff
@@ -258,6 +258,12 @@ pub fn test(env: TestEnv, cfg: TestConfig) {
                 node.get_ip_addr()
             )
         });
+
+        // By restarting ic-replica, we are restarting the orchestrator. On startup, the
+        // orchestrator purges authorized SSH keys from the node. After figuring out its assigned
+        // subnet, it will re-add the keys that are present in the registry. We thus wait to make
+        // sure that the keys are re-added before proceeding.
+        wait_until_authentication_is_granted(&node.get_ip_addr(), "backup", &backup_mean);
```
basvandijk: I see that the `wait_until_authentication_is_granted` and `wait_until_authentication_fails` functions are calling: `SshSession::default().login(ip, username, mean)` in a hot loop.

It would be better to use [`retry_with_msg!`](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/driver/src/driver/test_env_api.rs?L2052) such that we wait a bit between iterations and log what's happening.

pierugo-dfinity: Indeed, thanks for the suggestion. As this function is called in a lot of different places and needs an additional argument (a `logger`), I propose to do this change in a separate [PR](https://github.com/dfinity/ic/pull/7111).

basvandijk: Sounds good!

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -710,6 +727,7 @@ fn local_recovery(
         {maybe_replay_until_height}\
         {maybe_download_pool_node}\
         --admin-access-location local \
+        {maybe_nb_checkpoints}\
         {maybe_keep_downloaded_state}\
         --wait-for-cup-node {node_ip} \
         {maybe_backup_key_file}\
```
pierugo-dfinity, line 722: `local_recovery()` in `//.../nns_recovery/common.rs` and in `//.../subnet_recovery/common.rs` are very similar yet slightly different/inconsistent. CON-1598 now tracks this.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -299,6 +299,17 @@ pub fn test(env: TestEnv, cfg: TestConfig) {
                 node.get_ip_addr()
             )
         });
+
+        // By restarting ic-replica, we are restarting the orchestrator. On startup, the
+        // orchestrator purges authorized SSH keys from the node. After figuring out its assigned
+        // subnet, it will re-add the keys that are present in the registry. We thus wait to make
+        // sure that the keys are re-added before proceeding.
```
eichhorl: This actually sounds like something we should fix in the orchestrator instead. I don't think we should purge the SSH keys, otherwise if nodes restart and fail to re-add the keys during a disaster, we would lose backup access.

pierugo-dfinity: Very relevant comment, thanks. I implemented the required changes but I think they deserve their own [PR](https://github.com/dfinity/ic/pull/7181).

rs/recovery/src/app_subnet_recovery.rs
```diff
@@ -263,6 +269,13 @@ impl RecoveryIterator<StepType, StepTypeIter> for AppSubnetRecovery {
                     );
                 }
 
+                if self.params.nb_checkpoints.is_none() {
+                    self.params.nb_checkpoints = read_optional(
+                        &self.logger,
+                        "Enter number of checkpoints to download (default 1): ",
```
eichhorl, line 275: How does the operator know which number to enter?

eichhorl: Generally I think if the problem is that we shouldn't use the latest checkpoint, but the one before, then the resulting action should be to only download the previous checkpoint.

But I can see how that is probably significantly more effort to implement, so maybe specifying the number of checkpoints is fine for now.

eichhorl (approved):

PR #7056, opened 2025-10-02
test(ICRC-Ledger): add archiving to the test icrc3 ledger

maciejdfinity:
The archiving has to be performed manually in the test by calling the `archive_blocks` endpoint. The endpoint accepts the `Principal` of the archive and the number of blocks to archive.


rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/tests/tests.rs
```diff
@@ -44,6 +45,22 @@ fn icrc3_test_ledger_wasm() -> Vec<u8> {
...
+        .execute_ingress(canister_id, "icrc1_balance_of", Encode!(&account).unwrap())
+        .expect("Failed to send icrc1_balance_of")
+        .bytes();
+    Decode!(&res, Nat)
+        .expect("Failed to decode icrc1_balance_of response")
+        .0
+        .to_u64()
+        .expect("Balance must be a u64!")
+}
+
```
mbjorkqvist: nit: Instead of duplicating the code here, you may be able to instead call [`balance_of`](https://github.com/dfinity/ic/blob/891c0d9d63b158792f68999a69ad597e6c9130ff/rs/ledger_suite/test_utils/state_machine_helpers/src/lib.rs#L139-L150) from the `ic-ledger-suite-state-machine-helpers` crate (unless it e.g., pulls in a lot of extra dependencies).

rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/tests/tests.rs
```diff
@@ -577,3 +613,179 @@ fn test_supported_standards() {
...
+                start: Nat::from(0u64),
+                length: Nat::from(u64::MAX),
+            }])
+            .unwrap()
+        )
+        .expect("failed to get blocks")
+        .bytes(),
+        GetBlocksResult
+    )
+    .expect("failed to decode icrc3_get_blocks response");
```
mbjorkqvist: You may be able to call [icrc3_get_blocks](https://github.com/dfinity/ic/blob/891c0d9d63b158792f68999a69ad597e6c9130ff/rs/ledger_suite/test_utils/state_machine_helpers/src/lib.rs#L518-L539) from the `ic-ledger-suite-state-machine-helpers` crate here.

rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/src/main.rs
```diff
@@ -1,50 +1,143 @@
...
+                id: Nat::from(block_id),
+                block: block.1,
+            })
+            .await
+            .expect("failed to add block to archive")
+            .candid::<AddBlockResult>()
+            .expect("Could not decode AddBlockResult")
+            .expect("adding block failed");
+        assert_eq!(result, Nat::from(block_id));
+    }
```
mbjorkqvist, line 135: IIUC, if something fails here, the ledger+archive would potentially end up in an inconsistent state, since the blocks have already been `pop`ped in the ledger, but not necessarily stored in the archive (since when we make a call to the archive, any changes made to the ledger before that would be committed, and not rolled back in case we later panic)? Given that this is a test ledger this is not too critical, but would it perhaps make sense to change the return type from `u64` to a `Result`, and then the caller (probably in a unit/integration test) could panic, and that may be easier to debug than having the panics here inside the test ledger canister?

rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/tests/tests.rs
maciejdfinity: Thanks, done!

rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/tests/tests.rs
maciejdfinity: Good point, done!

rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger/src/main.rs
maciejdfinity: I agree! In this case it probably won't matter much, but e.g. in rosetta system tests we might not see the ledger panic at all.

PR #7058, opened 2025-10-02
feat(sns-wasm): Add an option to skip updating latest version in SnsWasm::add_wasm

jasonz-dfinity:
# Why

The ICVC index canister is uninstalled (ran out of cycles for too logn), which causes:

1. The normal SNS upgrade procedure does not work, since `InstallCode { mode : Upgrade}` does not work for uninstalled canisters
2. The auto-upgrade feature requires getting the status of all SNS framework canister, in order to determine the next canister to upgrade. An uninstalled canister does not have a status.

Therefore, if we want to reinstall the ICVC index canister, we want to install a special version of the SNS root canister to do a one-off `InstallCode { mode: Install}`. Such a special root canister wasm needs to be added to SNS-W canister without having all SNSes upgrade to this canister

Note that even though we can hard-code the canister id so that the one-off work only applies to ICVC, we still need to make the special root canister wasm base on the current root version running on ICVC root, for compatibility guarantees. For the same compatibility reason, we don't want such a special root reach all the root canisters for other SNSes since its base would be too old.

# What

1. Add a `skip_update_latest_version: opt bool`, consider `null` as `opt false` 
2. When `skip_update_latest_version = opt true` we avoid updating the upgrade paths and simply add the wasm.


rs/nns/sns-wasm/src/sns_wasm.rs
```diff
@@ -3263,6 +3305,165 @@ mod test {
...
+        let index_1_hash = add_wasm_and_return_hash(SnsCanisterType::Index, 1, false);
+        let governance_2_hash = add_wasm_and_return_hash(SnsCanisterType::Governance, 2, false);
+
+        println!("governance_1_hash: {:?}", governance_1_hash);
+        println!("root_1_hash: {:?}", root_1_hash);
+        println!("ledger_1_hash: {:?}", ledger_1_hash);
+        println!("swap_1_hash: {:?}", swap_1_hash);
+        println!("archive_1_hash: {:?}", archive_1_hash);
+        println!("index_1_hash: {:?}", index_1_hash);
+        println!("governance_2_hash: {:?}", governance_2_hash);
```
max-dfinity: Do we want these printlns to be checked in?

basvandijk (approved):

michael-weigelt (approved):

rs/nns/sns-wasm/src/sns_wasm.rs
jasonz-dfinity: Nice catch! Removed.

PR #7061, opened 2025-10-03
feat(nns): following only existing neurons allowed

ShahriarJavidi:
This PR enforces following those neurons that already exists in the neuron store.


ShahriarJavidi:
Closing in favour of [7013](https://github.com/dfinity/ic/pull/7013).

PR #7065, opened 2025-10-03
feat(canister_logging): add log_size configuration to canister settings

maksymar:


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2088,6 +2094,7 @@ pub mod testing {
             canister_history: Default::default(),
             wasm_chunk_store: WasmChunkStore::new_for_testing(),
             log_visibility: Default::default(),
+            log_size: NumBytes::from(MAX_ALLOWED_CANISTER_LOG_BUFFER_SIZE as u64), // TODO(EXC-2118): replace with default canister log size.
```
maksymar: According to the comment above I checked `CanisterSnapshot::from_canister()` and verified that logs are not preserved, as well as logging related canister settings, so adding `log_size` is not a breaking change.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -571,6 +576,7 @@ impl SystemState {
                 wasm_chunk_store_metadata,
             ),
             log_visibility,
+            log_size,
```
schneiderstefan: Existing canisters will end up with this variable set to 0, while new canisters end up with `MAX_ALLOWED_CANISTER_LOG_BUFFER`. Is that intended?

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
```diff
@@ -464,6 +465,8 @@ message CanisterStateBits {
   reserved "log_visibility";
   // Log visibility for the canister.
   LogVisibilityV2 log_visibility_v2 = 51;
+  // The size of the canister log in bytes.
+  uint64 log_size = 56;
```
schneiderstefan: Should this be `max_log_size`? If I understand the code correctly, the actual log size is just the sum of the records in field 43.

maksymar: maybe `log_size` is not the best name and it should rather be `log_capacity`. it's not an exact sum of fields, it's the capacity of ring-buffer which can have a hole of unused bytes (between tail and head).

rs/replicated_state/src/canister_state/system_state.rs
maksymar: In this particular PR it's not important, but it should be done any ways, so I added to this PR -- I set each canister setting to the default value now.

rs/execution_environment/src/canister_settings.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) struct CanisterSettings {
     pub(crate) freezing_threshold: Option<NumSeconds>,
     pub(crate) reserved_cycles_limit: Option<Cycles>,
     pub(crate) log_visibility: Option<LogVisibilityV2>,
+    pub(crate) log_capacity: Option<NumBytes>,
```
berestovskyy: nit: the `log_capacity` name is different both from the field names around (cycles_limit, memory_limit) and the internal constant name (LOG_BUFFER_SIZE)

rs/nns/cmc/src/lib.rs
```diff
@@ -203,6 +203,7 @@ impl From<CanisterSettingsArgs> for Ic00CanisterSettingsArgs {
             freezing_threshold: settings.freezing_threshold,
             reserved_cycles_limit: settings.reserved_cycles_limit,
             log_visibility: settings.log_visibility.map(LogVisibilityV2::from),
+            log_capacity: Some(candid::Nat::from(MAX_ALLOWED_CANISTER_LOG_BUFFER_SIZE)), // TODO: properly implement log_capacity.
```
berestovskyy: nit: should the TODO be linked to a ticket?

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
```diff
@@ -464,6 +465,8 @@ message CanisterStateBits {
   reserved "log_visibility";
   // Log visibility for the canister.
   LogVisibilityV2 log_visibility_v2 = 51;
+  // The capacity of the canister log in bytes.
+  uint64 log_capacity = 56;
```
berestovskyy: nit: Grouping fields in Proto makes it hard to determine the next available field number. I’d just add the new field at the end.

rs/execution_environment/src/canister_manager.rs
```diff
@@ -471,6 +471,7 @@ impl CanisterManager {
             settings.reserved_cycles_limit(),
             reservation_cycles,
             settings.log_visibility().cloned(),
+            settings.log_capacity(),
```
berestovskyy: Seems, we just copy the user-provided number into a validated canister setting. Should we have an actual validation here?

rs/execution_environment/src/scheduler.rs
```diff
@@ -1582,6 +1582,12 @@ impl Scheduler for SchedulerImpl {
 
                     // TODO(EXC-1124): Re-enable once the cycle balance check is fixed.
                     // cycles_out_sum += canister.system_state.queues().output_queue_cycles();
+
+                    // TODO(EXC-2141): Remove this temporary code of setting the log capacity to default value.
+                    // This code is added to make sure that all the canisters (existing and new ones)
+                    // have the log capacity set to the default value.
```
berestovskyy: Should we set the default on canister creation or loading canister state bits instead of every round?

berestovskyy (approved):

rs/nns/cmc/src/lib.rs
maksymar: done.

rs/execution_environment/src/canister_settings.rs
maksymar: good point, renamed to `log_memory_limit`, thank you!
`LOG_BUFFER_SIZE` is an obsolete name, will be removed when the constant is gone.

rs/execution_environment/src/scheduler.rs
maksymar: done, now it sets the default value when reading state from proto checkpoint data.

rs/execution_environment/src/canister_manager.rs
maksymar: Added validation of the value to be within min/max allowed range.

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
maksymar: yes, I also don't really like breaking the sorted order, but I find grouping is better for readability, therefore I added on top of the message a comment with the next message ID, which should help a bit. it'd be nice to have some kind of linter that forced updating this comment correctly.
I'll keep this code as is.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2088,6 +2094,7 @@ pub mod testing {
             canister_history: Default::default(),
             wasm_chunk_store: WasmChunkStore::new_for_testing(),
             log_visibility: Default::default(),
+            log_memory_limit: NumBytes::from(DEFAULT_LOG_MEMORY_LIMIT as u64),
```
maksymar: According to the comment above I checked `CanisterSnapshot::from_canister()` and verified that logs are not preserved, as well as logging related canister settings, so adding `log_memory_limit` is not a breaking change.

rs/nns/cmc/src/lib.rs
```diff
@@ -203,6 +203,7 @@ impl From<CanisterSettingsArgs> for Ic00CanisterSettingsArgs {
             freezing_threshold: settings.freezing_threshold,
             reserved_cycles_limit: settings.reserved_cycles_limit,
             log_visibility: settings.log_visibility.map(LogVisibilityV2::from),
+            log_memory_limit: None, // TODO(EXC-2118): properly implement log_memory_limit.
```
maksymar: FYI: @aterga

PR #7066, opened 2025-10-03
feat(ckdoge): derive Dogecoin address for deposits

gregorydemay:
Use public key derivation to derive a Dogecoin address from a ledger account that contains a user principal and an optional subaccount, similarly to what is done in the ckBTC minter.

One difficulty is that the `Bitcoin` address type is used as a field in `RetrieveBtcRequest`, which is both used in the minter event log (event `AcceptedRetrieveBtcRequest`) and in the minter state (`pending_retrieve_btc_requests`). Generalizing this to be able to deal with different address types without impacting ckBTC is not trivial: 

- a separate event log would be needed;
- and, the state would need to be made generic, requiring either dynamic dispatch or feature flags (the state is held in a `thread_local!`, implying that the type must be known at compile time)

To be able to reuse the same infrastructure, the 20 bytes of the Dogecoin address is mapped to the corresponding 20-bytes of a Bitcoin address before calling certain methods on the ckBTC minter.


rs/dogecoin/ckdoge/minter/src/lib.rs
```diff
@@ -73,4 +78,35 @@ impl CanisterRuntime for DogeCanisterRuntime {
...
+        let doge_address =
+            DogecoinAddress::parse(address, &doge_network).map_err(|e| e.to_string())?;
+
+        // This convertion is a hack to use the same type of address as in RetrieveBtcRequest,
+        // since this type is used both in the event logs (event `AcceptedRetrieveBtcRequest`)
+        // and in the minter state (field `pending_retrieve_btc_requests`)
+        Ok(match doge_address {
+            DogecoinAddress::P2pkh(bytes) => BitcoinAddress::P2pkh(bytes),
+            DogecoinAddress::P2sh(bytes) => BitcoinAddress::P2sh(bytes),
+        })
```
gregorydemay: @ninegua @mducroux : happy to hear your thoughts on this and whether you see other alternatives (see also reasoning in PR description)

ninegua: It seems that this is quickest way to make progress here, unless we want to abstract out state from all existing ckbtc minter functions and move thread local definitions out of the lib, which would be a much bigger effort.

gregorydemay: > It seems that this is quickest way to make progress here

That's also my current thinking. My idea would be to have something up and running as fast as possible with some decent amount of testing. This would enable us to be able to refactor things later if we feel the need.

rs/bitcoin/ckbtc/minter/src/lib.rs
```diff
@@ -1529,6 +1540,39 @@ impl CanisterRuntime for IcCanisterRuntime {
...
+        BitcoinAddress::parse(address, network).map_err(|e| e.to_string())
+    }
+
+    async fn check_address(
+        &self,
+        btc_checker_principal: Option<Principal>,
+        address: String,
+    ) -> Result<BtcAddressCheckStatus, CallError> {
+        let btc_checker_principal = btc_checker_principal
+            .expect("BUG: upgrade procedure must ensure that the Bitcoin checker principal is set");
```
mducroux, line 1558: Isn't this string outdated?

rs/dogecoin/ckdoge/minter/src/lib.rs
```diff
@@ -73,4 +78,35 @@ impl CanisterRuntime for DogeCanisterRuntime {
...
+    ) -> Result<BitcoinAddress, std::string::String> {
+        let doge_network = match network {
+            ic_ckbtc_minter::Network::Mainnet => Network::Mainnet,
+            ic_ckbtc_minter::Network::Testnet => Network::Testnet,
+            ic_ckbtc_minter::Network::Regtest => Network::Regtest,
+        };
+        let doge_address =
+            DogecoinAddress::parse(address, &doge_network).map_err(|e| e.to_string())?;
+
+        // This convertion is a hack to use the same type of address as in RetrieveBtcRequest,
```
mducroux: typo

rs/dogecoin/ckdoge/minter/src/lib.rs
```diff
@@ -73,4 +78,35 @@ impl CanisterRuntime for DogeCanisterRuntime {
             ic_cdk::trap("ecdsa_key_name is not set");
         }
     }
+
+    fn parse_address(
+        &self,
+        address: &str,
+        network: ic_ckbtc_minter::Network,
+    ) -> Result<BitcoinAddress, std::string::String> {
```
mducroux: nit: `std::string` can be removed

rs/dogecoin/ckdoge/minter/src/updates/get_doge_address.rs
```diff
@@ -0,0 +1,59 @@
...
+    ]
+}
+
+async fn derive_public_key(
+    ecdsa_key_name: String,
+    account: &Account,
+) -> Result<EcdsaPublicKeyResult, ic_cdk::call::Error> {
+    use ic_cdk::management_canister as mgmt;
+
+    mgmt::ecdsa_public_key(&mgmt::EcdsaPublicKeyArgs {
```
mducroux, line 50: understanding: is it a new endpoint and is it the reason why deriving public key is done different with Bitcoin?

rs/dogecoin/ckdoge/minter/src/updates/get_doge_address.rs
```diff
@@ -0,0 +1,59 @@
...
+        .try_into()
+        .expect("BUG: invalid ECDSA compressed public key");
+    Ok(DogecoinAddress::from_compressed_public_key(&public_key).display(&network))
+}
+
+/// Returns the derivation path that should be used to sign a message from a
+/// specified account.
+fn derivation_path(account: &Account) -> Vec<Vec<u8>> {
+    const SCHEMA_V1: u8 = 1;
+    const PREFIX: [u8; 4] = [b'd', b'o', b'g', b'e'];
```
mducroux: nit: not crystal clear that this corresponds to `doge`, maybe add a comment.

rs/dogecoin/ckdoge/minter/src/lib.rs
```diff
@@ -1,12 +1,17 @@
...
 use candid::Principal;
+use ic_ckbtc_minter::address::BitcoinAddress;
 use ic_ckbtc_minter::state::CkBtcMinterState;
+use ic_ckbtc_minter::updates::retrieve_btc::BtcAddressCheckStatus;
 use ic_ckbtc_minter::{
-    CanisterRuntime, CheckTransactionResponse, GetUtxosRequest, GetUtxosResponse, Network, Utxo,
+    CanisterRuntime, CheckTransactionResponse, GetUtxosRequest, GetUtxosResponse, Utxo,
     management::CallError, tx, updates::update_balance::UpdateBalanceError,
 };
 use icrc_ledger_types::icrc1::{account::Account, transfer::Memo};
```
mducroux, line 17: 2 lines below: does the `DOGECOIN_CANISTER_RUNTIME` has anything to do with the Dogecoin canister? If not, consider renaming

rs/dogecoin/ckdoge/minter/src/lib.rs
```diff
@@ -73,4 +78,35 @@ impl CanisterRuntime for DogeCanisterRuntime {
             ic_cdk::trap("ecdsa_key_name is not set");
         }
     }
+
+    fn parse_address(
+        &self,
+        address: &str,
+        network: ic_ckbtc_minter::Network,
+    ) -> Result<BitcoinAddress, std::string::String> {
+        let doge_network = match network {
```
mducroux: nit: why not using the From trait you implemented?

rs/dogecoin/ckdoge/minter/src/updates/get_doge_address.rs
ninegua: It can be replaced by `*b"doge"`

ninegua (approved):
LGTM! Thanks

rs/bitcoin/ckbtc/minter/src/lib.rs
gregorydemay: no I think that's still correct. The BTC checker principal was added afterwards, so that it is optional for backwards-compatibility reasons but it is expected to be set (in the case of ckBTC)

rs/dogecoin/ckdoge/minter/src/updates/get_doge_address.rs
gregorydemay: yes, this endpoint did not exist at the time where ckBTC went live. Actually in the next PR, I'm going to change it to the same way ckBTC does key derivation because the fact that `ecdsa_public_key` is asynchronous is annoying and does not need to be.

rs/dogecoin/ckdoge/minter/src/updates/get_doge_address.rs
gregorydemay: Thanks for the suggestion, much nicer!

rs/dogecoin/ckdoge/minter/src/lib.rs
gregorydemay: > DOGECOIN_CANISTER_RUNTIME has anything to do with the Dogecoin canister?

yes, for example for the get_utxos endpoint it will be used to query corresponding endpoint on the dogecoin canister (instead of the corresponding one on the Bitcoin canister)

rs/dogecoin/ckdoge/minter/src/lib.rs
gregorydemay: good point, I forgot about that one (wrote the `From` conversion afterwards).

PR #7069, opened 2025-10-03
perf(idkg): CON-1593 Use multi-threading in `validate_dealing_support`

eichhorl:
This PR makes use of the IDKG thread pool to validate dealing support shares in parallel. We iterate over all unvalidated support shares in parallel, and attempt to validate each of them.

Note that we keep a cache of signers for which we have already validated a support share for a given dealing. In order to be usable in the multi-threaded setting, the cache is behind a `RwLock`.

Before processing an unvalidated share, we acquire the read lock on this cache, and check if:
1. the same signer has already issued a support share for this dealing, in which case the new share is invalidated
2. we already validated enough (2f+1) support shares for the corresponding dealing, in which case the new share is removed immediately.

If neither condition is true, we will release the read lock and continue with cryptographic validation of the share. 

Note that this introduces the following race condition: Two different threads could each be assigned a different share by the same signer for the same dealing. After reading the cache, both threads determine that the share should be validated (despite the shares being duplicates of each other).

Therefore, if the share turns out to be valid after cryptographic verification, we acquire the _write_ lock on the cache, to insert the new signer and check again if a share by the same signer was validated in the meantime. If so, the newly verified share is invalidated instead.

### Alternatives Considered
An alternative approach (https://github.com/dfinity/ic/pull/7029) was to iterate over the unvalidated shares twice (once sequentially, and once in parallel): During the first iteration, we remove all duplicate shares and select at most 2f+1 shares for each dealing to be cryptographically verified as part of the second iteration. This would avoid the aforementioned race condition and guarantee that no more than 2f+1 shares are validated (which is a performance optimization).

However, this approach turns out to be more complex, and potentially insecure:

Assume we already verified 2f shares for a certain dealing and are now only missing the last one. Then, during the sequential iteration, we will only select a single share to be validated as part of the parallel iteration. If this single share turns out to be invalid, we will select a new share during the next call to `validate_dealing_support`. However, since shares for the same transcript and dealing are sorted by the signer's node ID, we will again pick a share by the same signer (assuming they sent another one/multiple in the meantime). If this signer turns out to be malicious, they could prevent us from ever accumulating enough support shares.

Instead, using locks to guarantee exclusive write access to the cache is simpler, and closer to the previous sequential implementation. The locks are only acquired for a very short amount of time, which does not affect performance according to performance tests.

In prinicpal, this approach could lead to frequently validating more than 2f+1 shares. However, unvalidated shares are ordered by the transcript ID. This means that under high load (where this optimization matters), support shares for the same transcript ID should be assigned to only one or two threads (modulo some work stealing), which should reduce the number of race conditions and therefore the number of unneeded validations.

### DIFF
--- a/rs/consensus/idkg/src/pre_signer.rs
+++ b/rs/consensus/idkg/src/pre_signer.rs
@@ -393,7 +393,7 @@ impl IDkgPreSignerImpl {
         // Collect all unvalidated dealing support shares into a vector to be processed in parallel.
         // The vector is then evenly split between threads of the thread pool. Shares in the vector
         // are sorted by the IDkgTranscriptId. Therefore, this increases the chance of shares for the
-        // same dealing ending up in the same thread, which reduced the number of superfluous shares
+        // same dealing ending up in the same thread, which reduces the number of superfluous shares
         // being validated due to race conditions.
         let unvalidated_supports: Vec<_> = idkg_pool.unvalidated().dealing_support().collect();
         self.thread_pool.install(|| {
@@ -515,7 +515,7 @@ impl IDkgPreSignerImpl {
                                 // Although we already checked the cache for duplicate shares above,
                                 // it could happen that a different thread validated a share for the
                                 // same (signer_id, transcript_id, dealer_id, dealing_hash) in the meantime,
-                                // after released the read lock. Therefore, we acquire the write lock here
+                                // after we released the read lock. Therefore, we acquire the write lock here
                                 // to check again with exclusive access.
                                 let mut valid_dealing_supports = self.validated_dealing_supports.write().unwrap();
                                 let signers = valid_dealing_supports.entry(key).or_default();


rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
-            // Drop shares for xnet reshare transcripts
-            if source_subnet_xnet_transcripts.contains(&support.transcript_id) {
-                self.metrics.pre_sign_errors_inc("xnet_reshare_support");
-                ret.push(IDkgChangeAction::HandleInvalid(
-                    id,
-                    format!("Support for xnet reshare transcript: {support}"),
-                ));
-                continue;
-            }
+        let unvalidated_supports: Vec<_> = idkg_pool.unvalidated().dealing_support().collect();
```
Sawchord, line 398: I assume it is not possible to call `into_par_iter` directly on `dealing_support`?

Maybe we could mention that here

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
                                 support,
                                 idkg_pool.stats(),
                             );
-                            if let Some(IDkgChangeAction::MoveToValidated(_)) = action {
-                                if let Some(signers) = maybe_signers {
-                                    signers.insert(signer);
-                                } else {
-                                    valid_dealing_supports.insert(key, BTreeSet::from([signer]));
+                            if let Some(IDkgChangeAction::MoveToValidated(msg)) = &action {
+                                let mut valid_dealing_supports = self.validated_dealing_supports.write().unwrap();
```
Sawchord, line 520: Maybe put a shorter version of what you wrote in the PR description here as a comment

Sawchord (approved):
PR looks good.

It is also much cleaner, using `return` instead of the highly nested `if...else`.

Thanks!

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
-                            ret.push(IDkgChangeAction::HandleInvalid(
+                // Disable the height check on target subnet side for the initial transcripts.
+                // Since the transcript_id.source_height is from the source subnet, the height
+                // cannot be relied upon. This also lets us process the shares for the initial
+                // bootstrap with higher urgency, without deferring it.
+                let msg_height = if target_subnet_xnet_transcripts.contains(&support.transcript_id) {
+                    None
+                } else {
+                    Some(support.transcript_id.source_height())
+                };
```
pierugo-dfinity, line 418:
```suggestion
                let msg_height = (!target_subnet_xnet_transcripts.contains(&support.transcript_id)).then_some(support.transcript_id.source_height());
```

Just a suggestion, I'm not convinced myself that it's more readable

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
+                        // Dedup dealing support by checking whether a previous (transcript_id, dealer_id,
+                        // dealing_hash) was already signed by the signer
+                        let key = IDkgValidatedDealingSupportIdentifier::from(&support);
+                        {
+                            let valid_dealing_supports = self.validated_dealing_supports.read().unwrap();
+                            let maybe_signers = valid_dealing_supports.get(&key);
+                            if maybe_signers.is_some_and(|signers| signers.contains(&signer)) {
+                                return Some(IDkgChangeAction::HandleInvalid(
+                                    id,
+                                    format!("Duplicate dealing support found in cache: {support}"),
```
pierugo-dfinity: Why change from `"Duplicate dealing support in unvalidated batch: {support}"`? You said yourself that we *check again if a share by the same signer was validated*. Thus this message should be the same as the one below when acquiring the write lock and realizing we've hit the race condition IIUC.

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
+                            }
+
+                            let Some(transcript_params) = self.resolve_ref(
+                                transcript_params_ref,
+                                block_reader,
+                                "validate_dealing_support",
+                            ) else {
+                                return Some(IDkgChangeAction::HandleInvalid(
+                                    id.clone(),
+                                    format!("Failed to translate transcript_params_ref: {}", support),
```
pierugo-dfinity:
```suggestion
                                    format!("Failed to translate transcript_params_ref: {support}"),
```

nit, following the style of other messages.

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
-                                    signers.insert(signer);
-                                } else {
-                                    valid_dealing_supports.insert(key, BTreeSet::from([signer]));
+                            if let Some(IDkgChangeAction::MoveToValidated(msg)) = &action {
+                                let mut valid_dealing_supports = self.validated_dealing_supports.write().unwrap();
+                                let signers = valid_dealing_supports.entry(key).or_default();
+                                if !signers.insert(signer) {
+                                    return Some(IDkgChangeAction::HandleInvalid(
+                                        id,
+                                        format!("Duplicate dealing support in unvalidated batch: {:?}", msg),
```
pierugo-dfinity: Why use `{msg:?}` instead of `{support}` like before/above?

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,143 +390,147 @@ impl IDkgPreSignerImpl {
...
-                                .pre_sign_errors_inc("missing_hash_invalid_dealer");
-                            ret.push(IDkgChangeAction::RemoveUnvalidated(id));
-                            warn!(
-                                self.log,
-                                "validate_dealing_support(): Missing hash, invalid dealer: {:?}",
-                                support
-                            );
+                            action
+
                         } else {
```
pierugo-dfinity:
```suggestion
                            return action;
                        }
```

i.e. consider removing the `else` to further reduce the indentation of what comes below.

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -2432,10 +2435,9 @@ mod tests {
 
                 let block_reader = block_reader.clone().with_fail_to_resolve();
                 let change_set = pre_signer.validate_dealing_support(&idkg_pool, &block_reader);
-                assert_eq!(change_set.len(), 4);
+                assert_eq!(change_set.len(), 3);
                 assert!(is_handle_invalid(&change_set, &msg_id_2));
                 assert!(is_handle_invalid(&change_set, &msg_id_2_dupl));
-                assert!(is_handle_invalid(&change_set, &msg_id_3));
                 assert!(is_removed_from_unvalidated(&change_set, &msg_id_4));
```
pierugo-dfinity, line 2453: Wait, why?

eichhorl: Before this PR we tried to resolve the transcript parameter refs before checking if we have the dealing supported by the unvalidated share. I moved the resolve step down just before calling crypto, because that's the only place where we actually need the resolved parameters. 

So now the share for `id_3` in the test is deferred because we don't attempt to resolve the references (which would fail otherwise). I added a comment in the test.

rs/consensus/idkg/src/pre_signer.rs
eichhorl: Nice, thanks!

rs/consensus/idkg/src/pre_signer.rs
eichhorl: `support` moved into the `crypto_verify_dealing_support` call and cannot be used anymore afterwards

rs/consensus/idkg/src/pre_signer.rs
eichhorl: Done 👍

rs/consensus/idkg/src/pre_signer.rs
eichhorl: I think there is value in having different error messages, so that during debugging we can determine in which of the two places the error happened. But I made both messages a bit more consistent now 👍

rs/consensus/idkg/src/pre_signer.rs
eichhorl: I think I prefer the `if` a bit 😄

rs/consensus/idkg/src/pre_signer.rs
eichhorl: Done 👍

rs/consensus/idkg/src/pre_signer.rs
eichhorl: Good question. `dealing_support()` returns an iterator, which indeed doesn't implement `IntoParallelIterator`. The alternative would be to call `idkg_pool.unvalidated().dealing_support().par_bridge().filter_map(...)`. In that case, rayon creates one producer thread that will advance the iterator and push elements onto a queue, from which they are consumed by worker threads, as new elements arrive.

Since the `dealing_support()` iterator is entirely in memory (and fast to consume), the overhead of doing this extra queueing probably exceeds simply collecting all elements into a vector and then evenly distributing the work from the start.

The latter also has the advantage that elements that are adjacent in the vector are more likely to be assigned to the same thread, which should help with reducing the race conditions I mentioned.

I added a small comment explaining this more.

rs/consensus/idkg/src/pre_signer.rs
pierugo-dfinity: Oh right, makes sense if nothing depends on that message functionality-wise. And the messages look even better now, thanks!

rs/consensus/idkg/src/pre_signer.rs
pierugo-dfinity: Right, thanks!

rs/consensus/idkg/src/pre_signer.rs
pierugo-dfinity: Got it, thanks for the explanation!

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,176 +390,188 @@ impl IDkgPreSignerImpl {
...
-                ret.push(IDkgChangeAction::HandleInvalid(
-                    id,
-                    format!("Support for xnet reshare transcript: {support}"),
-                ));
-                continue;
-            }
+        // Collect all unvalidated dealing support shares into a vector to be processed in parallel.
+        // The vector is then evenly split between threads of the thread pool. Shares in the vector
+        // are sorted by the IDkgTranscriptId. Therefore, this increases the chance of shares for the
+        // same dealing ending up in the same thread, which reduced the number of superfluous shares
```
pierugo-dfinity:
```suggestion
        // same dealing ending up in the same thread, which reduces the number of superfluous shares
```

rs/consensus/idkg/src/pre_signer.rs
```diff
@@ -390,176 +390,188 @@ impl IDkgPreSignerImpl {
...
-                            if let Some(IDkgChangeAction::MoveToValidated(_)) = action {
-                                if let Some(signers) = maybe_signers {
-                                    signers.insert(signer);
-                                } else {
-                                    valid_dealing_supports.insert(key, BTreeSet::from([signer]));
+                            if let Some(IDkgChangeAction::MoveToValidated(msg)) = &action {
+                                // Although we already checked the cache for duplicate shares above,
+                                // it could happen that a different thread validated a share for the
+                                // same (signer_id, transcript_id, dealer_id, dealing_hash) in the meantime,
+                                // after released the read lock. Therefore, we acquire the write lock here
```
pierugo-dfinity:
```suggestion
                                // after we released the read lock. Therefore, we acquire the write lock here
```

PR #7071, opened 2025-10-03
feat(node-rewards-canister): Compute rewards over single day for all node providers

pietrodimarco-dfinity:
Modified the get_rewards_calculation API.
No external changes, as the API is not yet in use.

Changes include:
- Compute rewards for all node providers over a single day, instead of calculating the full reward period for a single provider.
- Add detailed comments explaining the attributes in the rewards calculation results.
- Move API type conversions to the appropriate module for better organization.


rs/node_rewards/canister/api/src/api_native_conversion.rs
```diff
@@ -0,0 +1,249 @@
...
+    type Error = String;
+
+    fn try_from(value: NodeMetricsDaily) -> Result<Self, Self::Error> {
+        Ok(Self {
+            subnet_assigned: SubnetId::from(value.subnet_assigned.unwrap()),
+            subnet_assigned_fr: Decimal::try_from(value.subnet_assigned_fr_percent.unwrap())?,
+            num_blocks_proposed: value.num_blocks_proposed.unwrap(),
+            num_blocks_failed: value.num_blocks_failed.unwrap(),
+            original_fr: Decimal::try_from(value.original_fr_percent.unwrap())?,
+            relative_fr: Decimal::try_from(value.relative_fr_percent.unwrap())?,
```
daniel-wong-dfinity-org: Instead of `unwrap`, how about `.ok_or_else(|| "F is missing (from T).")?` ?

Ditto elsewhere.

rs/node_rewards/canister/api/src/api_native_conversion.rs
```diff
@@ -0,0 +1,249 @@
...
+use rewards_calculation::performance_based_algorithm::results as native_types;
+use rust_decimal::Decimal;
+
+// ================================================================================================
+// Conversion implementations for API types <-> Internal types
+// ================================================================================================
+
+// NodeMetricsDaily conversions
+impl From<native_types::NodeMetricsDaily> for NodeMetricsDaily {
+    fn from(metrics: native_types::NodeMetricsDaily) -> Self {
```
daniel-wong-dfinity-org:
```suggestion
    fn from(src: native_types::NodeMetricsDaily) -> Self {
```

Ditto elsewhere.

This is not a super strong pattern in the `ic` repo (yet!), but there is some precedent for it. I think this is better than `value`, which is hyper generic. Also, since we have two types with the same name, using a snake_case version of the type name is ambiguous and potentially confusing, because it does not indicate WHICH ONE of the two types it is; whereas, `src` does indicate which one (the source type, not the destination type).

rs/node_rewards/canister/api/src/api_native_conversion.rs
daniel-wong-dfinity-org, line 1: I guess I already said this, but I like how this relegates all this stupid code to a dark corner that we can avoid having to look at.

rs/node_rewards/canister/api/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+pub mod api_native_conversion;
```
daniel-wong-dfinity-org: I don't think this needs to be pub. It just contains `From` (and `TryFrom`) implementations, right? If you make this private, please, do not mix public and private mods in the same group.

rs/node_rewards/canister/api/src/provider_rewards_calculation.rs
```diff
@@ -67,9 +68,8 @@ pub struct NodeProviderRewards {
     pub base_rewards_type3: Vec<BaseRewardsType3>,
     pub nodes_results: Vec<NodeResults>,
 }
-
-#[derive(CandidType, candid::Deserialize, Clone, Debug)]
-pub struct NodeProviderRewardsDaily {
-    pub day_utc: Option<DateUtc>,
-    pub node_provider_rewards: Option<NodeProviderRewards>,
+#[derive(candid::CandidType, candid::Deserialize, Clone, PartialEq, Debug)]
```
daniel-wong-dfinity-org, line 72: Why not also `Eq` ? I guess adding `Eq` would be viral. In that case, no need to do that in this PR, but consider doing it later.

rs/node_rewards/canister/api/src/provider_rewards_calculation.rs
```diff
@@ -1,20 +1,21 @@
...
 #[derive(CandidType, Clone, Deserialize)]
 pub struct GetNodeProviderRewardsCalculationRequest {
-    pub from_day: DateUtc,
-    pub to_day: DateUtc,
-    pub provider_id: Principal,
+    pub day: DateUtc,
 }
 
-pub type GetNodeProviderRewardsCalculationResponse = Result<Vec<NodeProviderRewardsDaily>, String>;
+pub type GetNodeProviderRewardsCalculationResponse = Result<DailyResults, String>;
```
daniel-wong-dfinity-org, line 13: Can we (in a later PR) get rid of this?

In general, I think type aliases are a bad idea. If in order to use `GetNodeProviderRewardsCalculationResponse` correct, you need to know its definition, then, it just adds indirection without raising the level of abstraction.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
daniel-wong-dfinity-org, line 1: Thank you!

Seems like these changes could have been in a separate PR? Not ask you to break this out. Just a suggestion for future changes.

The .did file seems like a better place to put these comments. Here, you are only speaking to users of this Rust library. Whereas, the .did file is for anyone who calls the canister, regardless of what language or library they use.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 
-pub type XDRPermyriad = Decimal;
+pub type XDRPermyriad = Decimal; // Rewards unit in XDR scaled by 1/10,000 (permyriad)
 pub type Percent = Decimal;
```
daniel-wong-dfinity-org, line 11: Not asking for change ini this PR, since this issue already existed, but...

Here is another way of using type aliases that I don't like. This way of using `type` does not create new types per se; just new NAMES for existing types. E.g. using these aliases, you can do `xdr_permyriad + percent`, which is just nonsense. Similarly (and more realistically), you can do `xdr_permyriad + xdr_whole`. In other words, it doesn't prevent you from making the kind of mistakes that types are supposed to stop you from making. This not only adds (unnecessary) indirection, it can give a false sense of security; probably, many people think that `xdr_permyriad + percent` is disallowed, because they think that the two operands are of different types, but they aren't.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
 pub type Region = String;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct NodeMetricsDaily {
+    /// The subnet to which this node was assigned on this day
     pub subnet_assigned: SubnetId,
+    /// Subnet Assigned Failure Rate
```
daniel-wong-dfinity-org, line 21:
```suggestion

    /// Subnet Assigned Failure Rate
```

Blank lines between fields, please. BlankspacescontainusefulinformationandarethereforecommonlyusedForexampleMorsecodewithoutspacesisjustonesuperlongtonewhichofcourseconveyszeropracticalinformationAsongwithnorestsisacompletelydifferentsongSpeakingwithoutbreathingishardalsoonthelistenernotjustthespeaker

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
 use std::collections::BTreeMap;
 
-pub type XDRPermyriad = Decimal;
+pub type XDRPermyriad = Decimal; // Rewards unit in XDR scaled by 1/10,000 (permyriad)
 pub type Percent = Decimal;
 pub type Region = String;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct NodeMetricsDaily {
+    /// The subnet to which this node was assigned on this day
```
daniel-wong-dfinity-org: What if a node belongs to more than one subnet in a day? E.g. a node can exit one subnet and join another on the same day, right? Or is this somehow prevented? I guess we just use whatever subnet it was in at midnight UTC?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub subnet_assigned: SubnetId,
+    /// Subnet Assigned Failure Rate
+    ///
+    /// The failure rate of the entire subnet for this day.
```
daniel-wong-dfinity-org: Not sure how having both sentences help. They seem to be saying more or less the same thing.

Maybe, you can explain that this is the number of blocks that failed / the number of blocks proposed BY ALL NODES in the subnet, or something like that.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub subnet_assigned_fr: Percent,
+    /// Number of blocks proposed by this node on this day
     pub num_blocks_proposed: u64,
+    /// Number of blocks that failed to be included on this day
     pub num_blocks_failed: u64,
+    /// Original Failure Rate
+    ///
+    /// Calculated as `num_blocks_failed / (num_blocks_proposed + num_blocks_failed)`.
```
daniel-wong-dfinity-org, line 36: I really don't understand why `+ num_blocks_failed`. Seems to me, failure rate is just `failure_count / attempt_count`, and "proposed" sounds like merely an attempt, not necessarily a SUCCESSFUL attempt. If `num_blocks_proposed` were just SUCCESSFUL attempts, then I would understand, but the name does not indicate that only successful attempts are counted there.

Also, this field is technically redundant, right? This is just provided so that people can double-check our intermediate results, right? Maybe mention this fact?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    /// Represents the raw failure rate of the node before any subnet-level adjustments.
     pub original_fr: Percent,
```
daniel-wong-dfinity-org, line 38: This issue predates this PR; therefore, do not make changes in this PR on account of this issue, but consider changes for future PRs.

I don't see how "original" indicates what this is. "original" sounds like there are later revisions, as in, "well, when we first calculated this, we THOUGHT the true value was X, but later (better) calculations tell us that the true value is closer to X_prime". But there are no better+later calculations of the same thing.

Also, this name does not indicate WHAT operation failed, which is a pretty salient fact. If you ask me how I'm doing, and I just say, "I failed" without saying what I failed at doing, that's not much of an answer. You can't do anything with that. If I said I failed at flying a plane, that is completely different from failing to make a hair appointment. Failing to fly a plane has a high potential to kill many humans. Failing to make a hair appointment is exceedingly unlikely to have such results. Maybe something like `block_making_failure_rate` would make sense?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    /// Relative Failure Rate
+    ///
+    /// Failure rate adjusted for subnet performance.
+    /// Calculated as `max(0, original_fr - subnet_assigned_fr)`.
     pub relative_fr: Percent,
```
daniel-wong-dfinity-org, line 45: Can you maybe explain how this is used? Maybe a URL of a page that explains?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct NodeResults {
+    /// Unique identifier of the node
     pub node_id: NodeId,
+    /// NodeRewardType
     pub node_reward_type: NodeRewardType,
+    /// Geographical region of the node
```
daniel-wong-dfinity-org, line 68: Maybe, give an example?

How does this affect rewards?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+        /// Used to estimate the node's performance when unassigned
+        extrapolated_fr: Percent,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct NodeResults {
+    /// Unique identifier of the node
     pub node_id: NodeId,
+    /// NodeRewardType
```
daniel-wong-dfinity-org: This seems pretty vacuous. I guess, the way to really find out is to look at the definition of NodeRewardType.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    Unassigned {
+        /// Extrapolated Failure Rate (EFR)
+        /// Used to estimate the node's performance when unassigned
+        extrapolated_fr: Percent,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct NodeResults {
+    /// Unique identifier of the node
```
daniel-wong-dfinity-org: Yes, this is exactly what the name `node_id` is literally telling us; therefore, this comment seems rather vacuous. Maybe, you can mention that it is a fingerprint of the node's public key (and therefore, tied to the node's private key), as opposed to a random u64, or something.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub node_id: NodeId,
+    /// NodeRewardType
     pub node_reward_type: NodeRewardType,
+    /// Geographical region of the node
     pub region: String,
+    /// Data center identifier
```
daniel-wong-dfinity-org, line 72: Example?

I guess this does not affect rewards. It is just used to ensure decentralization, right?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub original_fr: Percent,
+    /// Relative Failure Rate
+    ///
+    /// Failure rate adjusted for subnet performance.
+    /// Calculated as `max(0, original_fr - subnet_assigned_fr)`.
     pub relative_fr: Percent,
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub enum NodeStatus {
```
daniel-wong-dfinity-org, line 50: An issue beyond the scope of this PR (therefore, don't do anything about this here), but...

I really do not expect something named "Status" to contain performance stats. What maybe makes more sense is

```
// Our primary measure of an ICP node's "productivity" is the amount of successful
// block making it was able to perform. To determine rewards due to a node's provider
// (on account of the node), we compare the performance of the node with its peers
// (i.e. other members of the subnet that the node belongs to).
enum NodeProductivity {
  SubnetMember(SubnetMemberNodeProductivity),
  NonSubnetMember(NonSubnetMemberNodeProductivity),
}
```

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub region: String,
+    /// Data center identifier
     pub dc_id: String,
+    /// Node status, assigned or unassigned, with associated metrics
     pub node_status: NodeStatus,
+    /// Performance multiplier (1 - rewards_reduction)
+    ///
+    /// Represents how rewards are adjusted based on node performance
     pub performance_multiplier: Percent,
+    /// Rewards reduction applied due to failure rates
```
daniel-wong-dfinity-org, line 84: Can you point to how this is derived from failure rate (of the node, and the subnet as a whole)?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
     pub rewards_reduction: Percent,
+    /// Base rewards before applying performance multipliers
     pub base_rewards: XDRPermyriad,
+    /// Rewards adjusted by the performance multiplier
     pub adjusted_rewards: XDRPermyriad,
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct BaseRewards {
```
daniel-wong-dfinity-org: It is surprising that there is a field named `base_rewards`, but its type is not this... When a key fits exactly into a key-shaped hole, nobody believes that's an accident. But in this case, you are not supposed to plug a `BaseRewards` into the `base_rewards` field!

rs/node_rewards/canister/api/src/api_native_conversion.rs
```diff
@@ -0,0 +1,249 @@
...
+        Ok(Self {
+            node_id: NodeId::from(value.node_id.unwrap()),
+            node_reward_type: NodeRewardType::from(value.node_reward_type.unwrap()),
+            region: value.region.unwrap(),
+            dc_id: value.dc_id.unwrap(),
+            node_status: native_types::NodeStatus::try_from(value.node_status.unwrap())?,
+            performance_multiplier: Decimal::try_from(
+                value.performance_multiplier_percent.unwrap(),
+            )?,
+            rewards_reduction: Decimal::try_from(value.rewards_reduction_percent.unwrap())?,
```
daniel-wong-dfinity-org: I think I said this before, and I guess this issue predates this PR, but...

It seems highly suspicious to me that in one type, field names have units (good!), but in the other type, they don't.

rs/node_rewards/canister/api/src/api_native_conversion.rs
```diff
@@ -0,0 +1,249 @@
...
+            adjusted_rewards: Decimal::try_from(value.adjusted_rewards_xdr_permyriad.unwrap())?,
+        })
+    }
+}
+
+// BaseRewards conversions
+impl From<native_types::BaseRewards> for BaseRewards {
+    fn from(rewards: native_types::BaseRewards) -> Self {
+        Self {
+            monthly_xdr_permyriad: Some(DecimalProto::from(rewards.monthly)),
```
daniel-wong-dfinity-org: This predates this PR (therefore, do not make changes in this PR on account of this issue), but...

If you use a Decimal type, why permyriad? Seems like the whole point of Decimal is that you can use the "natural" scale, i.e. 1 XDR, not 1e-4 XDR! The whole point of using cents, e8s, etc is so that you can stay in integer land. Whereas, Decimal frees you from such shackles.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    /// Region for which base rewards are calculated
     pub region: Region,
+    /// Monthly base rewards in XDR permyriad
     pub monthly: XDRPermyriad,
+    /// Daily base rewards in XDR permyriad
     pub daily: XDRPermyriad,
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct BaseRewardsType3 {
```
daniel-wong-dfinity-org: Pre-existing issue (hence, no change in this PR).

Wouldn't `Type3NodeBaseRewards` make more sense? Also, would be good to point to an explanation of what a type 3 node is.

daniel-wong-dfinity-org: I guess this means that BaseRewards does not cover type 3 nodes? Maybe, name should say so (a pre-existing issue). But at least, comments can say so.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,71 +5,119 @@ use rust_decimal::Decimal;
...
+    /// Node is unassigned; only extrapolated failure rate is available
+    Unassigned {
+        /// Extrapolated Failure Rate (EFR)
+        /// Used to estimate the node's performance when unassigned
+        extrapolated_fr: Percent,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct NodeResults {
```
daniel-wong-dfinity-org: A pre-existing issue.

"result" is super generic. Instead, how about `DailyNodeReward`, or something like that?

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: Agree :)

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: I mainly used decimals to not loose precision when I have to do those aggregation in the reward calculator library. I agree however here we could have just used XDR it was mostly to stay consistent with current API.
But happy to discuss more :)

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Fully agree

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Agree :) Changing it and adding comment as suggested

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: Agree :) Changing it everywhere

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Adding comment

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: This is a good point, I will refactor it in the next one :)

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: Good point :) Will do it

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Yeah I totally see your point. This comes from [consensus](https://github.com/dfinity/ic/blob/e284bc85cab1d41ecd754ca0846bbee9cafc3840/rs/replicated_state/src/metadata_state.rs#L1389-L1396), the idea is that you (as node) are tasked to be the blockmaker and you if you propose a block to the blockchain you have successfully made it.  That's why the total attempt are calculated here as `num_blocks_proposed + num_blocks_failed`

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Agree thx :) Added comment explaining how this is used and why

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Yeah I see your point. My intention here was to leave comments just in a single source of truth.
I do agree that a better place however is the .did file. Will move them ;)

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: I see your reasoning! Thank you :) I think in the following pr I will just use Decimal type for all of them and be more specific using better field's names as I do in the candid...like adding postfix `_percent`

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: It is not prevented to a node to join multiple subnets in a day. In all cases what is used is the subnet assigned to the node in the "latest registry version" before the day ends. Will explain that in comment ;) Thank you!

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Will do thx :)

rs/node_rewards/canister/api/src/provider_rewards_calculation.rs
pietrodimarco-dfinity: Sure we can :) Updating TODO

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: I see what you mean, lets give more context to the `failure rate`. Added to the TODOs

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Sure I will add in a later PR an `.md` with documentation about the algorithm and link it here

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Agree, Added to todo list

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Added to todo list

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: I mean I think that the description of this is already included in its definition. I agree these are vacuous thus will remove the comments

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: same as above ;)

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Adding a todo to point to documentation file once it is in the repo ;)

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: As above will add link to the documentation

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Add a TODO to point explanation to documentation once available

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Yes I agree...This struct is meant to give "explanation" on why base_rewards has a certain value. I will change the name of `BaseRewards` to `BaseRewardsSpec`. Let me know if you can think of better ways to add clarity here :)

rs/node_rewards/canister/api/src/api_native_conversion.rs
```diff
@@ -58,172 +74,226 @@ impl From<native_types::NodeStatus> for NodeStatus {
...
-            base_rewards: Decimal::try_from(value.base_rewards_xdr_permyriad.unwrap())?,
-            adjusted_rewards: Decimal::try_from(value.adjusted_rewards_xdr_permyriad.unwrap())?,
         })
     }
 }
 
 // BaseRewards conversions
-impl From<native_types::BaseRewards> for BaseRewards {
-    fn from(rewards: native_types::BaseRewards) -> Self {
+impl From<native_types::BaseRewardsSpec> for BaseRewardsSpec {
```
daniel-wong-dfinity-org, line 156: Not sure why you call this "base rewards SPEC" instead of "base rewards". That is, I'm not really sure what "spec" is intended to convey? What is wrong with "base rewards"? What issue(s) are you trying to address by adding the word "spec"?

rs/node_rewards/canister/api/src/provider_rewards_calculation.rs
```diff
@@ -5,10 +5,11 @@ use ic_nervous_system_proto::pb::v1::Decimal;
 use std::collections::BTreeMap;
 
 #[derive(CandidType, Clone, Deserialize)]
-pub struct GetNodeProviderRewardsCalculationRequest {
+pub struct GetNodeProvidersRewardsCalculationRequest {
```
daniel-wong-dfinity-org, line 8: Not sure why you changed "Provider" to "Providers"?

rs/node_rewards/canister/src/canister/mod.rs
```diff
@@ -289,16 +289,24 @@ impl NodeRewardsCanister {
...
-        let result = canister.with_borrow(|canister| canister.calculate_rewards(request_inner))?;
-        Ok(DailyResults::from(result.daily_results))
+        let mut result =
+            canister.with_borrow(|canister| canister.calculate_rewards(request_inner))?;
+
+        let day = NaiveDate::try_from(request.day)?;
+        let daily_results = result
+            .daily_results
+            .remove(&day)
+            .expect("Could not find daily results for the requested day");
```
daniel-wong-dfinity-org: Why not return Err instead of panic? I mean, I understand that this is due to an internal bug, not because the caller did something wrong. Still, this can be Err too, right?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -5,38 +5,52 @@ use rust_decimal::Decimal;
...
     ///
-    /// The failure rate of the entire subnet for this day.
+    /// This is the SUBNET_FAILURE_RATE_PERCENTILE of the original_fr all nodes in the subnet.
+    /// It is used to adjust individual node failure rates to account for systematic issues
+    /// affecting the whole subnet.
     pub subnet_assigned_fr: Percent,
-    /// Number of blocks proposed by this node on this day
+
+    /// Number of blocks successfully proposed by this node on this day
     pub num_blocks_proposed: u64,
```
daniel-wong-dfinity-org, line 29: Since this counts only SUCCESSFULly proposed blocks, maybe this can be called num_blocks_successfully_proposed?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
```diff
@@ -46,78 +60,105 @@ pub enum NodeStatus {
...
     pub monthly: XDRPermyriad,
+
     /// Daily base rewards in XDR permyriad
     pub daily: XDRPermyriad,
 }
 
+/// Base rewards for a Type 3 node.
+///
+/// Type3 nodes are defined [rs/protobuf/src/gen/registry/registry.node.v1.rs]
+/// For nodes which are type3 special logic is applied to compute base rewards.
```
daniel-wong-dfinity-org, line 114:
```suggestion
/// For nodes which are type3, special logic is applied to compute base rewards.
```

Also, can you maybe give some high level reason for why type 3 nodes are treated differently? What is different about them that warrants special treatment? Maybe, there isn't enough room here for that, and the pointer is enough.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
daniel-wong-dfinity-org: Absolutely. I'm not suggesting that you COPY the comments; but rather, MOVE them.

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
daniel-wong-dfinity-org: This tells us how much a node would get if it were "fully productive" (based on its type and region), right?

Maybe, NodeTypeAndRegionBaseRewards ?

rs/node_rewards/canister/api/src/api_native_conversion.rs
daniel-wong-dfinity-org: I would think that the result is not affected by the scale. Even if it is, I would think that such rounding errors are acceptable?

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: I wish to stay consistent with what we have in IC code for this

rs/node_rewards/canister/api/src/provider_rewards_calculation.rs
pietrodimarco-dfinity: Because now is daily for all node providers :)

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: Lets maybe discuss about this...basically this is for adding additional specification to why we reward as BaseRewards

rs/node_rewards/canister/api/src/api_native_conversion.rs
pietrodimarco-dfinity: Results yeah are not affect by the scale we could go with u64

rs/node_rewards/canister/src/canister/mod.rs
pietrodimarco-dfinity: Makes sense :)

rs/node_rewards/rewards_calculation/src/performance_based_algorithm/results.rs
pietrodimarco-dfinity: Eah I don't know...I don't particularly like also that

PR #7073, opened 2025-10-03
fix(node): libvirt contention on VM runner

andrewbattat:
We've seen VM contention logs like the following:
```
Guest VM stopped, restarting

Machine qemu-1-guestos terminated.

Creating GuestOS virtual machine

Assembling config media for GuestOS: /tmp/config_mediavQpdwr

Unable to remove disk metadata on vm guestos from /tmp/config_mediag7h6JB (disk target sda)

operation failed: domain 'guestos' is already being removed

libvirt: Domain Config error : operation failed: domain 'guestos' is already being removed

ERROR: Failed to start GuestOS virtual machine.

Error: Failed to define GuestOS virtual machine

    1: error: operation failed: domain 'guestos' is already being removed [code=operationfailed (9), domain=domain (20)]
```

This fixes the try_destroy_existing_vm retry logic, as the libvirt `Domain Config error : operation failed: domain 'guestos' is already being removed` error was not triggering the restrictive conditional. Now, the loop retries for any domain creation error.

### DIFF
--- a/rs/ic_os/os_tools/guest_vm_runner/src/main.rs
+++ b/rs/ic_os/os_tools/guest_vm_runner/src/main.rs
@@ -26,7 +26,6 @@ use tokio::time::sleep;
 use tokio_util::sync::CancellationToken;
 use virt::connect::Connect;
 use virt::domain::Domain;
-use virt::error::{ErrorDomain, ErrorNumber};
 use virt::sys::{VIR_DOMAIN_DESTROY_GRACEFUL, VIR_DOMAIN_NONE, VIR_DOMAIN_RUNNING};
 
 mod boot_args;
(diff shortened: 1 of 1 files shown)


frankdavid (approved):

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -142,21 +141,19 @@ impl VirtualMachine {
...
                 Ok(domain) => break domain,
-                Err(e)
-                    if retries > 0
-                        && e.code() == ErrorNumber::OperationInvalid
-                        && e.domain() == ErrorDomain::Domain =>
-                {
+                Err(e) if retries > 0 => {
+                    println!("Domain creation failed, retrying: {}", e);
                     Self::try_destroy_existing_vm(libvirt_connect, vm_domain_name);
+                    sleep(Duration::from_secs(3)); // Wait to give libvirt a chance to clean up the domain.
```
frankdavid: I don't think sleeping should be necessary. When `destroy_flags` returns, the VM should already be dead. Did you find that this sleep helps with the issue or do you just want to test whether it helps?

andrewbattat: Just wanted to test. I'll remove 👍

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -170,10 +166,13 @@ impl VirtualMachine {
...
         println!("Attempting to destroy existing '{vm_domain_name}' domain");
-        if let Err(e) = Domain::lookup_by_name(libvirt_connect, vm_domain_name)
-            .and_then(|existing| existing.destroy_flags(VIR_DOMAIN_DESTROY_GRACEFUL))
-        {
-            eprintln!("Failed to destroy existing domain: {e}");
+        if let Ok(existing_domain) = Domain::lookup_by_name(libvirt_connect, vm_domain_name) {
+            match existing_domain.destroy_flags(VIR_DOMAIN_DESTROY_GRACEFUL) {
+                Ok(_) => println!("Successfully destroyed existing domain"),
+                Err(e) => eprintln!("Failed to destroy existing domain: {e}"),
+            }
```
Bownairo: nit: 
```
if let Err(e) = existing_domain.destroy_flags(VIR_DOMAIN_DESTROY_GRACEFUL) {
    eprintln!("Failed to destroy existing domain: {e}");
} else {
    println!("Successfully destroyed existing domain");
}
```

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
Bownairo: > with small waits in between.

Let's make sure this is dropped from the description, too.

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
andrewbattat: Done: 279de1ffc13914bf59c27508e3041755995be609

PR #7080, opened 2025-10-04
refactor(crypto): remove obsolete lock in proto secret key store

fspreiss:


altkdf (approved):
Thanks for the cleanup, @fspreiss!

Note: Removing whitespace changes from the diff simplified reviewing this PR significantly.

randombit (approved):

PR #7082, opened 2025-10-06
feat(cmc): support env vars in canister creation settings

viviveevee:
Add support for the canister setting `environment_variables` to `create_canister` and `notify_create_canister` of the CMC. 
This will enable the cycles ledger to also support environment variables in canister settings when creating new canisters. This in turn will enable `icp-cli` and maybe `dfx` to support these settings on canister creation.


rs/nns/cmc/CHANGELOG.md
```diff
@@ -10,7 +10,6 @@ here were moved from the adjacent `unreleased_changelog.md` file.
 
 INSERT NEW RELEASES HERE
 
-
```
daniel-wong-dfinity-org, line 13: This seems extraneous.

PR #7092, opened 2025-10-06
feat(http-outcalls-IC-1948): gossip the full response if the request is not fully replicated

mihailjianu1:
TLDR:
- the full response is now gossiped together with the rest of the share in the case of non replicated http outcalls
- there are still no advertisements performed, as only one node knows the response initially 
- the response is validated during artifact validation (i.e. we check that its hash is equal to the hash in the metadata)
- non replicated outcalls ~4x faster 🚀🚀🚀🚀

### DIFF
--- a/rs/rust_canisters/proxy_canister/src/main.rs
+++ b/rs/rust_canisters/proxy_canister/src/main.rs
@@ -8,6 +8,7 @@
 #![allow(deprecated)]
 use candid::Principal;
 use futures::future::join_all;
+use futures::stream::{FuturesUnordered, StreamExt};
 use ic_cdk::api::call::RejectionCode;
 use ic_cdk::api::time;
 use ic_cdk::{caller, spawn};
@@ -89,31 +90,23 @@ pub async fn start_continuous_requests(
     ))
 }
 
-// TODO: instead of sequentially awaiting on each batch, try to send the next requests anyway, with backoff.
-// This should improve the overall qps, as the canister message queue is the bottleneck, and it's not being saturated.
 async fn run_continuous_request_loop(request: RemoteHttpRequest) {
-    const BATCH_SIZE: usize = 500;
-    let futures_iter = (0..BATCH_SIZE).map(|_| send_request(request.clone()));
-    let results = join_all(futures_iter).await;
+    const PARALLEL_REQUESTS: usize = 500;
 
-    let mut successes = 0;
-    let mut errors = 0;
-    for result in results {
-        match result {
-            Ok(_resp) => {
-                successes += 1;
-            }
-            Err((rejection_code, msg)) => {
-                errors += 1;
-                println!("Request failed: {rejection_code:?} - {msg}");
-            }
-        }
+    let mut futures = FuturesUnordered::new();
+
+    for _ in 0..PARALLEL_REQUESTS {
+        futures.push(send_request(request.clone()));
     }
-    println!("Finished batch of {BATCH_SIZE} requests => successes: {successes}, errors: {errors}");
 
-    spawn(async move {
-        run_continuous_request_loop(request).await;
-    });
+    while let Some(result) = futures.next().await {
+        match result {
+            Ok(_resp) => {}
+            Err((_rejection_code, _msg)) => {}
+        }
+
+        futures.push(send_request(request.clone()));
+    }
 }
 
 #[update]
--- a/rs/tests/networking/canister_http_soak_test.rs
+++ b/rs/tests/networking/canister_http_soak_test.rs
@@ -36,16 +36,9 @@ use ic_types::Cycles;
 use proxy_canister::RemoteHttpRequest;
 use proxy_canister::RemoteHttpResponse;
 use proxy_canister::UnvalidatedCanisterHttpRequestArgs;
-use serde::{Deserialize, Serialize};
 use slog::{Logger, info};
 
-#[derive(Serialize, Deserialize, Debug)]
-struct BenchmarkResult {
-    subnet_size: usize,
-    concurrent_requests: u64,
-    qps: f64,
-    average_latency_s: f64,
-}
+const INSTALLED_CANISTERS: usize = 6;
 
 fn main() -> Result<()> {
     SystemTestGroup::new()
@@ -70,16 +63,29 @@ pub fn test(env: TestEnv) {
             .expect("there is no node in this application subnet");
 
         let runtime = get_runtime_from_node(&node);
+        let webserver_ipv6 = get_universal_vm_address(&env);
+        let mut proxy_canisters = Vec::new();
         // Each requests costs ~6-7 billion cycles, and we make many thousands of requests.
         // The default 100T cycles may not be enough.
-        let proxy_canister =
-            create_proxy_canister_with_cycles(&env, &runtime, &node, Cycles::new(u128::MAX));
 
-        let webserver_ipv6 = get_universal_vm_address(&env);
+        // Install 6 proxy canisters and make them all send requests in paralel.
+        for i in 0..INSTALLED_CANISTERS {
+            let canister_name = format!("canister_name_{}", i);
+            let proxy_canister = create_proxy_canister_with_name_and_cycles(
+                &env,
+                &runtime,
+                &node,
+                &canister_name,
+                Cycles::new(u128::MAX),
+            );
+            proxy_canisters.push(proxy_canister);
+        }
 
         block_on(async {
             let url = format!("https://[{webserver_ipv6}]:20443");
-            leave_proxy_canister_running(&proxy_canister, url.clone(), logger.clone()).await;
+            for canister in &proxy_canisters {
+                leave_proxy_canister_running(canister, url.clone(), logger.clone()).await;
+            }
         });
     }
 }
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -214,10 +214,10 @@ jobs:
               targets_args+=("--skip_long_tests")
             fi
             if [[ '${{ needs.config.outputs.skip_didc_checks }}' == 'true' ]]; then
-              targets_args+=("--skip_didc_checks")
+              targets_args+=("--exclude_tags=didc")
             fi
             if [[ '${{ needs.config.outputs.skip_buf_checks }}' == 'true' ]]; then
-              targets_args+=("--skip_buf_checks")
+              targets_args+=("--exclude_tags=buf")
             fi
             if [[ '${{ needs.config.outputs.diff_only }}' == 'true' ]]; then
               targets_args+=(
@@ -666,9 +666,14 @@ jobs:
           run: bazel build --config=stamped --config=local --invocation_id="${{ fromJSON(steps.gen-uuids.outputs.output).build_id }}" //...
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
 
-  cargo-clippy-linux:
-    name: Cargo Clippy Linux
-    runs-on: &dind-small-setup
+  cargo-linux:
+    strategy:
+      matrix:
+        include:
+          - name: Lint
+          - name: Build
+    name: Cargo  ${{ matrix.name }} Linux
+    runs-on:
       labels: dind-small
     container: *container-setup
     timeout-minutes: 30
@@ -683,12 +688,12 @@ jobs:
               - "**/*.rs"
               - "**/*.toml"
               - "**/*.lock"
-      - name: Run Cargo Clippy Linux
-        id: cargo-clippy-linux
+      - name: Run Cargo ${{ matrix.name }} Linux
         env:
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         if: |
           steps.filter.outputs.cargo == 'true' ||
+          contains(github.event.pull_request.labels.*.name, 'CI_RUN_CARGO_JOBS') ||
           env.BRANCH_NAME == 'master' ||
           startsWith(env.BRANCH_NAME, 'rc--') ||
           startsWith(env.BRANCH_NAME, 'hotfix-')
@@ -696,38 +701,12 @@ jobs:
         run: |
           set -eExuo pipefail
           export CARGO_TERM_COLOR=always # ensure output has colors
-          "$CI_PROJECT_DIR"/ci/scripts/rust-lint.sh
 
-  cargo-build-release-linux:
-    name: Cargo Build Release Linux
-    runs-on: *dind-small-setup
-    container: *container-setup
-    timeout-minutes: 30
-    steps:
-      - *checkout
-      - name: Filter Rust Files [*.{rs,toml,lock}]
-        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
-        id: filter
-        with:
-          filters: |
-            cargo:
-              - "**/*.rs"
-              - "**/*.toml"
-              - "**/*.lock"
-      - name: Run Cargo Build Release Linux
-        id: cargo-build-release-linux
-        env:
-          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
-        if: |
-          steps.filter.outputs.cargo == 'true' ||
-          env.BRANCH_NAME == 'master' ||
-          startsWith(env.BRANCH_NAME, 'rc--') ||
-          startsWith(env.BRANCH_NAME, 'hotfix-')
-        shell: bash
-        run: |
-          set -eExuo pipefail
-          export CARGO_TERM_COLOR=always # ensure output has colors
-          cargo build --release --locked
+          if [[ "${{ matrix.name }}" == "Lint" ]]; then
+            "$CI_PROJECT_DIR"/ci/scripts/rust-lint.sh
+          else
+            cargo build --release --locked
+          fi
 
   check-pull-request-bazel-targets:
     name: Check PULL_REQUEST_BAZEL_TARGETS
--- a/.github/workflows/ci-pr-only.yml
+++ b/.github/workflows/ci-pr-only.yml
@@ -87,8 +87,7 @@ jobs:
 
   lock-generate:
     name: Lock Generate
-    runs-on: *dind-small-setup
-    container: *container-setup
+    runs-on: ubuntu-latest
     timeout-minutes: 30
     if: ${{ github.event_name != 'merge_group' }}
     env:
@@ -168,6 +167,8 @@ jobs:
       pull-requests: write
     env:
       SHELL_WRAPPER: "/usr/bin/time"
+      CARGO_AUDIT_VERSION: "0.21.0"
+      CARGO_WASMPACK_VERSION: "0.12.1"
       CI_MERGE_REQUEST_IID: ${{ github.event.pull_request.number }}
       CI_PROJECT_PATH: ${{ github.repository }}
       CI_PIPELINE_ID: ${{ github.run_id }}
@@ -196,13 +197,15 @@ jobs:
         if: steps.filter.outputs.depcheck == 'true'
         with:
           python-version: "3.12"
-      - name: Setup python deps
-        id: setup-python-deps
+      - name: Setup environment deps
+        id: setup-environment-deps
         if: steps.filter.outputs.depcheck == 'true'
         shell: bash
         run: |
           # Ignore externally-managed-environment pip error, install packages system-wide.
           PIP_BREAK_SYSTEM_PACKAGES=1 pip3 install --ignore-installed -r requirements.txt
+          cargo install cargo-audit --version "${CARGO_AUDIT_VERSION}"
+          cargo install wasm-pack --version "${CARGO_WASMPACK_VERSION}"
       - name: Dependency Scan for Pull Request
         id: dependencies-check
         if: steps.filter.outputs.depcheck == 'true'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10126,6 +10126,7 @@ dependencies = [
  "ic-icrc1-test-utils",
  "ic-icrc1-tokens-u256",
  "ic-icrc1-tokens-u64",
+ "ic-icrc3-test-ledger",
  "ic-ledger-canister-core",
  "ic-ledger-core",
  "ic-ledger-suite-state-machine-helpers",
--- a/ci/scripts/targets.py
+++ b/ci/scripts/targets.py
@@ -170,16 +170,14 @@ def diff_only_query(command: str, base: str, head: str, skip_long_tests: bool) -
 def targets(
     command: str,
     skip_long_tests: bool,
-    skip_didc_checks: bool,
-    skip_buf_checks: bool,
+    exclude_tags: list[str],
     base: str | None,
     head: str | None,
 ):
     """Print the bazel targets to build or test to stdout."""
     # If no base is specified, form a query to return all targets
     # but exclude those tagged with 'long_test' (in case --skip_long_tests was specified)
-    # and exclude those tagged with 'didc' (in case --skip_didc_checks was specified)
-    # and exclude those tagged with 'buf' (in case --skip_buf_checks was specified).
+    # and those with any of the excluded tags.
     # Otherwise return a query for all targets that have modified inputs in the specified
     # git commit range taking several factors into account:
     query = (
@@ -189,12 +187,7 @@ def targets(
     )
 
     # Finally, exclude targets that have any of the excluded tags:
-    excluded_tags = EXCLUDED_TAGS
-    if skip_didc_checks:
-        excluded_tags.append("didc")
-    if skip_buf_checks:
-        excluded_tags.append("buf")
-    excluded_tags_regex = "|".join(excluded_tags)
+    excluded_tags_regex = "|".join(EXCLUDED_TAGS + exclude_tags)
     query = f'({query}) except attr(tags, "{excluded_tags_regex}", //...)'
 
     args = ["bazel", "query", "--keep_going", query]
@@ -275,8 +268,9 @@ def main():
         help="Bazel command to generate targets for. If 'check' then check PULL_REQUEST_BAZEL_TARGETS for correctness",
     )
     parser.add_argument("--skip_long_tests", action="store_true", help="Exclude tests tagged as 'long_test'")
-    parser.add_argument("--skip_didc_checks", action="store_true", help="Exclude tests tagged as 'didc'")
-    parser.add_argument("--skip_buf_checks", action="store_true", help="Exclude tests tagged as 'buf'")
+    parser.add_argument(
+        "--exclude_tags", action="append", default=[], help="Exclude targets tagged with the specified tags"
+    )
     parser.add_argument(
         "--base",
         help="Only include targets with modified inputs in `git diff --name-only --merge-base $BASE $HEAD`. When --head is not provided defaults to HEAD.",
@@ -287,7 +281,7 @@ def main():
     if args.command == "check":
         check()
 
-    targets(args.command, args.skip_long_tests, args.skip_didc_checks, args.skip_buf_checks, args.base, args.head)
+    targets(args.command, args.skip_long_tests, args.exclude_tags, args.base, args.head)
 
 
 if __name__ == "__main__":
--- a/packages/pocket-ic/src/nonblocking.rs
+++ b/packages/pocket-ic/src/nonblocking.rs
@@ -1498,13 +1498,14 @@ impl PocketIc {
     }
 
     pub(crate) async fn do_drop(&mut self) {
-        self.stop_http_gateway().await;
         if self.owns_instance {
             self.reqwest_client
                 .delete(self.instance_url())
                 .send()
                 .await
                 .expect("Failed to send delete request");
+        } else {
+            self.stop_http_gateway().await;
         }
     }
 
--- a/packages/pocket-ic/tests/icp_features.rs
+++ b/packages/pocket-ic/tests/icp_features.rs
@@ -557,7 +557,9 @@ fn test_cycles_ledger() {
         )
         .unwrap()
         .0;
-        assert_eq!(balance, expected_index_balance);
+        // TODO: Once https://github.com/dfinity/ic/pull/6508 is on mainnet we can remove
+        // `expected_index_balance` param and only compare against `expected_ledger_balance`.
+        assert!(balance == expected_index_balance || balance == expected_ledger_balance);
     };
     let check_cycles = |expected: u128| {
         let actual = pic.cycle_balance(canister_id);
--- a/packages/pocket-ic/tests/tests.rs
+++ b/packages/pocket-ic/tests/tests.rs
@@ -17,6 +17,7 @@ use pocket_ic::{
         InitialTime, InstanceConfig, InstanceHttpGatewayConfig, MockCanisterHttpResponse,
         RawEffectivePrincipal, RawMessageId, SubnetConfigSet, SubnetKind,
     },
+    nonblocking::PocketIc as PocketIcAsync,
     query_candid, start_server, update_candid,
 };
 use reqwest::header::CONTENT_LENGTH;
@@ -2813,6 +2814,8 @@ async fn list_http_gateways(server_url: &Url) -> Vec<HttpGatewayDetails> {
 
 #[tokio::test]
 async fn with_http_gateway_config_and_cleanup_works() {
+    // We start a fresh server so that we can easily list instances and HTTP gateways
+    // created by this test (without filtering those created by other tests).
     let server_params = StartServerParams {
         server_binary: None,
         reuse: false,
@@ -2820,6 +2823,17 @@ async fn with_http_gateway_config_and_cleanup_works() {
     };
     let (_child, server_url) = start_server(server_params).await;
 
+    // Assert that
+    // - an instance exists on the server iff `instance_exists` is set to `true`;
+    // - the number of HTTP gateways on the server matches `gateway_count`.
+    let assert_server_state = |server_url: Url, instance_exists: bool, gateway_count: usize| async move {
+        let instances = list_instances(&server_url).await;
+        assert_eq!(instances.len(), 1);
+        assert!(instances[0].contains("Deleted") != instance_exists);
+        assert_eq!(list_http_gateways(&server_url).await.len(), gateway_count);
+    };
+
+    // We create a PocketIC instance and its HTTP gateway.
     let http_gateway_config = InstanceHttpGatewayConfig {
         ip_addr: None,
         port: None,
@@ -2833,18 +2847,34 @@ async fn with_http_gateway_config_and_cleanup_works() {
         .with_auto_progress()
         .build_async()
         .await;
+    assert_server_state(server_url.clone(), true, 1).await;
 
-    let instances = list_instances(&server_url).await;
-    assert_eq!(instances.len(), 1);
-    assert!(!instances[0].contains("Deleted"));
-    assert_eq!(list_http_gateways(&server_url).await.len(), 1);
+    // We create an additional handle for the existing PocketIC instance and start an additional HTTP gateway.
+    let mut pic_handle =
+        PocketIcAsync::new_from_existing_instance(server_url.clone(), pic.instance_id, None);
+    pic_handle.make_live(None).await;
+    assert_server_state(server_url.clone(), true, 2).await;
 
+    // We create yet another handle for the existing PocketIC instance and start an additional HTTP gateway.
+    let mut yet_another_pic_handle =
+        PocketIcAsync::new_from_existing_instance(server_url.clone(), pic.instance_id, None);
+    yet_another_pic_handle.make_live(None).await;
+    assert_server_state(server_url.clone(), true, 3).await;
+
+    // Dropping one of the extra handles for the existing PocketIC instance only stops its new HTTP gateway.
+    pic_handle.drop().await;
+    assert_server_state(server_url.clone(), true, 2).await;
+
+    // The instance is still in auto progress mode.
+    assert!(pic.auto_progress_enabled().await);
+
+    // Dropping the original handle deletes the PocketIC instance and stops all its HTTP gateways.
     pic.drop().await;
+    assert_server_state(server_url.clone(), false, 0).await;
 
-    let instances = list_instances(&server_url).await;
-    assert_eq!(instances.len(), 1);
-    assert_eq!(instances[0], "Deleted");
-    assert!(list_http_gateways(&server_url).await.is_empty());
+    // Dropping the other extra handle for the existing PocketIC instance succeeds, but is a no-op.
+    yet_another_pic_handle.drop().await;
+    assert_server_state(server_url.clone(), false, 0).await;
 }
 
 async fn assert_create_instance_failure(
--- a/rs/http_utils/tests/large_file_download_from.rs
+++ b/rs/http_utils/tests/large_file_download_from.rs
@@ -32,7 +32,7 @@ async fn test_large_file_download() {
     let hash = get_hash(downloader, version.as_ref()).await;
 
     let url = format!(
-        "http://download.proxy-global.dfinity.network:8080/ic/{version}/guest-os/update-img/update-img.tar.zst"
+        "https://download.dfinity.systems/ic/{version}/guest-os/update-img/update-img.tar.zst"
     );
     let downloader = FileDownloader::new_with_timeout(None, std::time::Duration::from_secs(2));
 
--- a/rs/ledger_suite/icrc1/index-ng/BUILD.bazel
+++ b/rs/ledger_suite/icrc1/index-ng/BUILD.bazel
@@ -102,12 +102,14 @@ rust_test(
         data = [
             conf["index_wasm"],
             conf["ledger_wasm"],
+            "//rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger:icrc3_test_ledger_canister.wasm.gz",
         ],
         env = {
             "RUST_TEST_THREADS": "4",
             "CARGO_MANIFEST_DIR": "rs/ledger_suite/icrc1/index-ng",
             "IC_ICRC1_INDEX_NG_WASM_PATH": "$(rootpath " + conf["index_wasm"] + ")",
             "IC_ICRC1_LEDGER_WASM_PATH": "$(rootpath " + conf["ledger_wasm"] + ")",
+            "IC_ICRC3_TEST_LEDGER_WASM_PATH": "$(rootpath //rs/ledger_suite/icrc1/test_utils/icrc3_test_ledger:icrc3_test_ledger_canister.wasm.gz)",
         },
         extra_srcs = ["tests/common/mod.rs"],
         tags = ["cpu:4"],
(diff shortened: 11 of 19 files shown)


berestovskyy:
Thanks, @mihailjianu1! The PR is ~1K lines of code. Any chance we could break it down into smaller pieces?

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -113,22 +124,43 @@ impl MutablePool<CanisterHttpResponseShare> for CanisterHttpPoolImpl {
...
                         is_latency_sensitive: true,
                     }));
                     self.validated.insert(share, ());
                     self.content
                         .insert(ic_types::crypto::crypto_hash(&content), content);
                 }
                 CanisterHttpChangeAction::MoveToValidated(share) => {
-                    if self.unvalidated.remove(&share).is_some() {
+                    if let Some(artifact) = self.unvalidated.remove(&share) {
+                        if let Some(content) = artifact.response {
```
Sawchord, line 158: NIT: Could you add a comment here? It is clear from the PR what this does, but without the context, it might be not.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -828,6 +828,108 @@ fn divergence_error_message() {
...
+    // This test ensures that when a non-replicated request is delegated to
+    // a different node than the block maker, and the response is gossiped
+    // back to the block maker, the payload builder correctly includes the response.
+
+    test_config_with_http_feature(true, 4, |mut payload_builder, canister_http_pool| {
+        // In the test setup, the block maker is node 0. We'll make node 1 delegated.
+        let delegated_node_id = node_test_id(1);
+        let callback_id = CallbackId::from(42);
+
+        // 1. Setup a non-replicated request delegated to our block maker (`delegated_node_id`).
```
Sawchord: I don't understand this sentence. `delegated_node_id` is not the blockmaker according to prior comment.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -828,6 +828,108 @@ fn divergence_error_message() {
...
+        let parsed_payload = bytes_to_payload(&payload).expect("Failed to parse payload");
+
+        // We should have exactly one response in the payload.
+        assert_eq!(
+            parsed_payload.responses.len(),
+            1,
+            "Expected exactly one response in the payload"
+        );
+
+        // The response must contain one seignature.
```
Sawchord:
```suggestion
        // The response must contain one signature.
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -320,6 +320,22 @@ impl CanisterHttpPoolManagerImpl {
...
                     self.metrics.shares_signed.inc();
+
+                    let active_contexts = &self
+                        .latest_state()
+                        .metadata
+                        .subnet_call_context_manager
+                        .canister_http_request_contexts;
+
+                    if let Some(context) = active_contexts.get(&response.id)
+                        && let Replication::NonReplicated(_) = context.replication
```
kpop-dfinity: nit: maybe this reads better?

```suggestion
                    if active_contexts
                         .get(&response.id)
                         .is_some_and(|context| matches!(context.replication, Replication::NonReplicated(_))
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -383,11 +399,69 @@ impl CanisterHttpPoolManagerImpl {
 
                 match active_contexts.get(&share.content.id) {
```
kpop-dfinity: nit: maybe we could use `let-else` here to reduce nestedness?

```suggestion
                let Some(context) = active_contexts.get(&share.content.id) else {
                    return Some(CanisterHttpChangeAction::RemoveUnvalidated(share.clone())); 
                };
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -383,11 +399,69 @@ impl CanisterHttpPoolManagerImpl {
...
                     Some(context) => {
-                        if matches!(context.replication, Replication::NonReplicated(node_id) if node_id != share.signature.signer) {
-                            return Some(CanisterHttpChangeAction::HandleInvalid(
-                                share.clone(),
-                                "Share signed by node that is not the delegated node for the request".to_string(),
-                            ));
+                        if let Replication::NonReplicated(node_id) = context.replication {
```
kpop-dfinity: nit: I'd personally prefer to use `match` here instead of `if let {} else {}`

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -383,11 +399,69 @@ impl CanisterHttpPoolManagerImpl {
...
+                                return Some(CanisterHttpChangeAction::HandleInvalid(
+                                    share.clone(),
+                                    "Content hash does not match the response".to_string(),
+                                ));
+                            }
+
+                            //TODO: we should also check the response size when validating the payload.
+
+                            // An honest replica enforces that response.content.count_bytes() does not exceed max_response_bytes
+                            // when the content is `Success`. However it doesn't enroce anything in the case of `Failure`.
```
kpop-dfinity:
```suggestion
                            // when the content is `Success`. However it doesn't enforce anything in the case of `Failure`.
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -383,11 +399,69 @@ impl CanisterHttpPoolManagerImpl {
...
+                            // As we still want to set a limit for failure, we enforce 1KB, which si reasonable for 
+                            // an error message.
+
+                            let response_size = response.content.count_bytes() as u64;
+
+                            let max_response_size = match context.max_response_bytes {
+                                Some(response_size) => max(MINIMUM_ALLOWED_RESPONSE_BYTES, response_size.get()),
```
kpop-dfinity: nit:
```suggestion
                                Some(max_response_size) => max(MINIMUM_ALLOWED_RESPONSE_BYTES, max_response_size.get()),
```

rs/tests/networking/canister_http_soak_test.rs
```diff
@@ -70,16 +70,27 @@ pub fn test(env: TestEnv) {
             .expect("there is no node in this application subnet");
 
         let runtime = get_runtime_from_node(&node);
+        let webserver_ipv6 = get_universal_vm_address(&env);
```
Sawchord: QQ: Are these changes related to the feature? I am not seeing the connection

rs/rust_canisters/proxy_canister/src/main.rs
```diff
@@ -8,6 +8,7 @@
 #![allow(deprecated)]
 use candid::Principal;
 use futures::future::join_all;
+use futures::stream::{FuturesUnordered, StreamExt};
```
Sawchord: QQ: I assume, these changes are needed to performance test the http feauture?
If that is the case, maybe factor it out into a separate PR

Sawchord (commented):
Looks good so far. I left a couple comments, mostly for clarification.

I have not looked at `rs/https_outcalls/consensus/src/pool_manager.rs` yet, will do that later today

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -383,11 +399,69 @@ impl CanisterHttpPoolManagerImpl {
...
+                                    "Share signed by node that is not the delegated node for the request".to_string(),
+                                ));
+                            }
+                            let Some(artifact) = canister_http_pool.get_unvalidated_artifact(share) else {
+                                // This should never happen
+                                return Some(CanisterHttpChangeAction::HandleInvalid(
+                                    share.clone(),
+                                    "Share exists without artifact".to_string(),
+                                ));
+                            };
```
kpop-dfinity: Maybe we should replace `CanisterHttpPool::get_unvalidated_shares` method with `CanisterHttpPool::get_unvalidated_artifacts` which would return an iterator of all `(share, artifact)` pairs in the unvalidated pool. That way we wouldn't need the check here and a couple of lines below. (If I understand correctly)

rs/rust_canisters/proxy_canister/src/main.rs
mihailjianu1: ditto

rs/tests/networking/canister_http_soak_test.rs
mihailjianu1: yeah, I removed everything related to performance.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
mihailjianu1: done thanks

rs/https_outcalls/consensus/src/payload_builder/tests.rs
mihailjianu1: Bad copy pasta, removed that.

rs/artifact_pool/src/canister_http_pool.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Yeap agreed

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: nice

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Ok but the nestedness we saved in other comment is now back :(
jk I agree you are right

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -94,16 +105,17 @@ impl CanisterHttpPool for CanisterHttpPoolImpl {
     fn lookup_unvalidated(
         &self,
         msg_id: &CanisterHttpResponseId,
-    ) -> Option<CanisterHttpResponseShare> {
-        self.unvalidated.get(msg_id).map(|()| msg_id.clone())
+    ) -> Option<CanisterHttpResponseArtifact> {
+        self.unvalidated.get(msg_id).cloned()
     }
```
eichhorl: This function is only called by tests and has almost the same signature as `get_unvalidated_artifact`. Can we remove it?

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -25,14 +27,16 @@ const POOL_CANISTER_HTTP_CONTENT: &str = "canister_http_content";
 
 type ValidatedCanisterHttpPoolSection = PoolSection<CanisterHttpResponseShare, ()>;
 
-type UnvalidatedCanisterHttpPoolSection = PoolSection<CanisterHttpResponseShare, ()>;
+type UnvalidatedCanisterHttpPoolSection =
+    PoolSection<CanisterHttpResponseShare, CanisterHttpResponseArtifact>;
```
eichhorl, line 31: There is a type alias `pub type CanisterHttpResponseId = CanisterHttpResponseShare;` It seems we are using both  names in this file. Would it make sense to use one of them consistently?

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: I moved the check before the match statement, so we don't check twice. I feel like `CanisterHttpPool::get_unvalidated_artifacts` returns too much when other times only the keys are relevant

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -160,9 +195,20 @@ impl MutablePool<CanisterHttpResponseShare> for CanisterHttpPoolImpl {
...
-    fn get(&self, id: &CanisterHttpResponseId) -> Option<CanisterHttpResponseShare> {
-        self.validated.get(id).map(|()| id.clone())
+impl ValidatedPoolReader<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
+    fn get(&self, id: &CanisterHttpResponseId) -> Option<CanisterHttpResponseArtifact> {
+        // Important: this is actually never used, as Http artifacts are always sent directly (no adverts).
+        // If we ever decide to use adverts, we should make the distinction between artifacts with or without a
+        // response. We should either:
+        //  - only use adverts when gossiping a full response; this way, this should always return the response
+        //  - store a flag in the share which says whether the full response needs to be gossiped or not.
+        // This is to avoid sending the response in full in the fully replicated case.
```
eichhorl, line 200: Do you plan to revisit this as part of flexible outcalls? Should we add a ticket?

rs/https_outcalls/consensus/src/pool_manager.rs
kpop-dfinity: lol

I actually was thinking about it when writing the comment

kpop-dfinity: My main motivation for using `match` is that when we add a new variant to `Replication` we will have to explicitly handle it here. With `If let {} else` we might forget to do it

rs/artifact_pool/src/canister_http_pool.rs
eichhorl, line 31: Just for my understanding: Would this also have worked?
```suggestion
    PoolSection<CanisterHttpResponseShare, Option<CanisterHttpResponse>>;
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -320,6 +320,22 @@ impl CanisterHttpPoolManagerImpl {
                     };
                     self.requested_id_cache.borrow_mut().remove(&response.id);
                     self.metrics.shares_signed.inc();
+
+                    let active_contexts = &self
+                        .latest_state()
+                        .metadata
+                        .subnet_call_context_manager
+                        .canister_http_request_contexts;
```
eichhorl: Is the latest state expected to (meaningfully) change between iterations of this loop?

rs/types/types/src/batch/canister_http.rs
```diff
@@ -328,4 +384,75 @@ mod tests {
         let new_payload = CanisterHttpResponseDivergence::try_from(pb_payload).unwrap();
         assert_eq!(payload, new_payload);
     }
+
+    /// Tests that a roundtrip of protobuf conversions for `CanisterHttpResponse`
+    /// works correctly.
+    #[test]
+    fn canister_http_response_conversion() {
```
kpop-dfinity, line 391: We could consider adding an "exhaustive" test for `CanisterHttpResponse`, something like https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/types/types/src/exhaustive.rs?L1091-1106

(I'm not sure how much extra work it would be, so not necessary do it in this PR, if at all)

rs/types/types/src/canister_http.rs
```diff
@@ -711,23 +711,31 @@ impl crate::crypto::SignedBytesWithoutDomainSeparator for CanisterHttpResponseMe
...
 
-impl IdentifiableArtifact for CanisterHttpResponseShare {
+/// Contains a share and optionally the full response.
+///
+/// This is the artifact that will actually be gossiped.
+#[derive(Clone, Debug, PartialEq)]
+pub struct CanisterHttpResponseArtifact {
+    pub share: CanisterHttpResponseShare,
+    // The response should not be included in the case of fully replicated outcalls.
+    pub response: Option<CanisterHttpResponse>,
```
kpop-dfinity, line 724: Did you consider doing something like this instead (the current design is also good and maybe it's too late to change it now)?
```suggestion
pub enum CanisterHttpResponseArtifact {
    Replicated(CanisterHttpResponseShare),
    NonReplicated(CanisterHttpResponseShare, CanisterHttpResponse),
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +397,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                        if response_size > max_response_size {
+                            return Some(CanisterHttpChangeAction::HandleInvalid(
+                                share.clone(),
+                                format!("Response size {response_size} exceeds the maximum allowed size of {max_response_size}"),
+                            ));
+                        }
                     }
-                    None => {
-                        return Some(CanisterHttpChangeAction::RemoveUnvalidated(share.clone()));
+                    _ => {
```
kpop-dfinity:
```suggestion
                    Replication::FullyReplicated => {
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +397,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                                share.clone(),
+                                "Content hash does not match the response".to_string(),
+                            ));
+                        }
+
+                        //TODO: we should also check the response size when validating the payload.
+
+                        // An honest replica enforces that response.content.count_bytes() does not exceed max_response_bytes
+                        // when the content is `Success`. However it doesn't enroce anything in the case of `Failure`.
+                        // As we still want to set a limit for failure, we enforce 1KB, which si reasonable for
```
kpop-dfinity:
```suggestion
                        // As we still want to set a limit for failure, we enforce 1KB, which is reasonable for
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -320,6 +320,22 @@ impl CanisterHttpPoolManagerImpl {
...
+
+                    let active_contexts = &self
+                        .latest_state()
+                        .metadata
+                        .subnet_call_context_manager
+                        .canister_http_request_contexts;
+
+                    if active_contexts.get(&response.id).is_some_and(|context| {
+                        matches!(context.replication, Replication::NonReplicated(_))
+                    }) {
```
eichhorl: This condition reads as; if there is a response for which we don't have a context, it should be treated as fully replicated. Should we handle the case where there is no context separately?

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: no it shouldn't change, so I guess we can move that before the loop.

rs/artifact_pool/src/canister_http_pool.rs
mihailjianu1: Yeah that would have also worked. But I found the current version cleaner

rs/artifact_pool/src/canister_http_pool.rs
mihailjianu1: Well I don't intend to add adverts for flexible outcalls either, as even for flexible outcalls, we still want all responses from all peers (at least that's the current design I am thinking of).

rs/artifact_pool/src/canister_http_pool.rs
mihailjianu1: Yes nice catch

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +397,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                        if node_id != share.signature.signer {
                             return Some(CanisterHttpChangeAction::HandleInvalid(
                                 share.clone(),
                                 "Share signed by node that is not the delegated node for the request".to_string(),
                             ));
                         }
+
+                        let Some(response) = &artifact.response else {
+                            // The request is not fully replicated, but the response is missing.
+                            return Some(CanisterHttpChangeAction::RemoveUnvalidated(share.clone()));
```
eichhorl: Should we return `HandleInvalid` in this case?

rs/artifact_pool/src/canister_http_pool.rs
mihailjianu1: yeah that's true, it's a bit of a mess. I tried to create some consistency by having the following rule:
- whenever p2p communicates with us, they use `CanisterHttpResponseId` (as they are the ones enforcing `IdentifiableArtifact` etc)
- us internally, we use `CanisterHttpResponseShare`, as we don't have/don't care about the concept of ID / identifiable.
lmk if that sounds fair enough

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +397,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                            Some(response_size) => max(MINIMUM_ALLOWED_RESPONSE_BYTES, response_size.get()),
+                            None => MAX_CANISTER_HTTP_RESPONSE_BYTES,
+                        };
+
+                        if response_size > max_response_size {
+                            return Some(CanisterHttpChangeAction::HandleInvalid(
+                                share.clone(),
+                                format!("Response size {response_size} exceeds the maximum allowed size of {max_response_size}"),
+                            ));
+                        }
```
kpop-dfinity: Should we treat success and failure cases differently?
i.e.
1. for success - check that `response_size <= context.max_response_bytes`
2. for failure - check that `response_size <= max(context.max_response_bytes, MINIMUM_ALLOWED_RESPONSE_BYTES)`

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -21,17 +21,20 @@ use ic_registry_client_helpers::subnet::SubnetRegistry;
...
 use std::{
     cell::RefCell,
+    cmp::max,
     collections::{BTreeSet, HashSet},
     convert::TryInto,
     sync::{Arc, Mutex},
     time::Duration,
 };
 
+const MINIMUM_ALLOWED_RESPONSE_BYTES: u64 = 1024; // 1KB
```
kpop-dfinity: IMHO, the name is a bit misleading

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +398,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                        };
+
+                        if share.content.content_hash != ic_types::crypto::crypto_hash(response) {
+                            return Some(CanisterHttpChangeAction::HandleInvalid(
+                                share.clone(),
+                                "Content hash does not match the response".to_string(),
+                            ));
+                        }
+
+                        //TODO: we should also check the response size when validating the payload.
```
eichhorl: Is this TODO already resolved by the check below?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +398,68 @@ impl CanisterHttpPoolManagerImpl {
...
+
+                        // An honest replica enforces that response.content.count_bytes() does not exceed max_response_bytes
+                        // when the content is `Success`. However it doesn't enroce anything in the case of `Failure`.
+                        // As we still want to set a limit for failure, we enforce 1KB, which si reasonable for
+                        // an error message.
```
eichhorl, line 512: If we want to enforce a limit on the failure size during response validation, then we should also enforce the same limit during response creation. Otherwise it could happen that responses created by honest replicas are rejected.

rs/artifact_pool/src/canister_http_pool.rs
eichhorl: I see, that makes sense 👍

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: No, it would have to be another check on the block payload validator

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Good question, I don't have a strong opinion. My main line of reasoning was that from the protocol's perspective, it's <<ok>> not to gossip the full response to everyone (who knows, perhaps in the future we might only gossip it to 1/2 of the replicas to reduce BW?). So we shouldn't flag it as malicious behaviour  (which I believe `InvalidArtifact` is associated with).

But as I said, I'm happy to hear ideas here.

rs/types/types/src/canister_http.rs
mihailjianu1: Yes! Something like this was the first thing I tried (hence the name of this branch ...-outcalls-gossip-hacky), but then I tried going with something easier. There were a few things I wasn't a fan of about that approach:
- `NonReplicated` implies a certain API, whereas that field could also be used by flexible outcalls in the future
- accessing fields became quite verbose (even more match / if let's etc)
- I think semantically an `Option<Response>` actually makes sense: the artifact either contains the response or doesn't, the rest is the same.

wdyt?

rs/https_outcalls/consensus/src/pool_manager.rs
Sawchord: +1
That case should be logged as an error and then be ignored

rs/https_outcalls/consensus/src/pool_manager.rs
Sawchord: I agree with Leo. In the current evolution of the protocol this is malicious behaviour, and should be treated as such.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +398,68 @@ impl CanisterHttpPoolManagerImpl {
...
                 }
 
-                match active_contexts.get(&share.content.id) {
-                    Some(context) => {
-                        if matches!(context.replication, Replication::NonReplicated(node_id) if node_id != share.signature.signer) {
+                let Some(context) = active_contexts.get(&share.content.id) else {
+                    return Some(CanisterHttpChangeAction::RemoveUnvalidated(share.clone()));
+                };
+
+                let Some(artifact) = canister_http_pool.get_unvalidated_artifact(share) else {
```
Sawchord: If I understand correctly, the share is contained within the artifact.
It should be possible to implement a `get_unvalidated_artifacts` function, that returns an iterator, and use that to iterate over artifacts in L387.

 This would simplify the code and also potentially be faster

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Yes. It's got the same behaviour now (or before non replicated calls were introduced). IE we don't check to see if there is a context in order to add it to the validated pool. 

From my understanding, the context missing would only happen if the response for the request has already been finalized before the HTTP response from the server came back. Then the response would be added to the validated pool -> then immediately purged, and the share will be gossiped only to be dropped immediately by the peers' bouncer.

So skipping and not performing any pool action for this request sounds the most reasonable, but at the same time it could be a risky change

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -381,17 +398,68 @@ impl CanisterHttpPoolManagerImpl {
...
+                                share.clone(),
+                                "Content hash does not match the response".to_string(),
+                            ));
+                        }
+
+                        //TODO: we should also check the response size when validating the payload.
+
+                        // An honest replica enforces that response.content.count_bytes() does not exceed max_response_bytes
+                        // when the content is `Success`. However it doesn't enroce anything in the case of `Failure`.
+                        // As we still want to set a limit for failure, we enforce 1KB, which is reasonable for
```
Sawchord, line 511: Unfortunately, I don't think 1KB is enough.

In the case where we have divergence, we generate a quite verbose error message, where we include the hashes of the responses.
On full divergence, this is larger than  67B * #Nodes, which can easily exceed 1KB.

I think, we should use the same limits for error message as for regular content. There is no security to be gained by treating the case differently, and it introduces complexity.

Sawchord (commented):
Looks good.

I like how the implementation circumvents touching the brittle payload builder.

I only have small nits, with the exception of the failure size checking.

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Done, both the `create_shares_from_responses` and `validate_shares` now call `validate_response_size`.

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: renamed

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Yeah, I refactored everything into `validate_response_size`, please lmk what you think.

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: You're right, great catch, lemme redo this

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: okay, we drop the response in that case and send a warning.

mihailjianu1: Actually getting back to this, requiring the context to be in the state makes like half of our unit tests fail :( Can we maybe do this in another PR?

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: So artifacts can't contain a "divergence" message, as that's only generated in the block payload, however the risk of an error being greater than 1KB is still real, so as recommended by @Sawchord , we'll truncate the error message to fit in 1KB when generating the error. Then the artifact validation can just check agains 1KB.

This needs to be treated as a special case from the "happy path" successful response as that is clearly bounded by `max_response_bytes` (albeit before candid serialization, hence we need an overhead there too..)

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: yeah Kamil recommended this too. I think you're both right, I removed `get_unvalidated_shares` with `get_unvalidated_artifacts` now .

rs/types/types/src/batch/canister_http.rs
mihailjianu1: Just added one. It wasn't much extra work, just made good use of `ExhaustiveSet` :)

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -56,6 +56,49 @@ pub struct CanisterHttpPoolManagerImpl {
     log: ReplicaLogger,
 }
 
+const MINIMUM_ALLOWED_ERROR_MESSAGE_BYTES: u64 = 1024; // 1KB
+const CANDID_OVERHEAD_RESERVE_BYTES: u64 = 1024; // 1KB
+
+// Checks that the response size is withing the allowed limits.
+// In general, the response size must be lower than max_response_bytes, however there are to caveats:
```
eichhorl:
```suggestion
// Checks that the response size is within the allowed limits.
// In general, the response size must be lower than max_response_bytes, however there are two caveats:
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -56,6 +56,49 @@ pub struct CanisterHttpPoolManagerImpl {
...
+                ))
+            } else {
+                Ok(())
+            }
+        }
+        CanisterHttpResponseContent::Reject(reject) => {
+            let response_size = reject.message.len() as u64;
+            if response_size > MINIMUM_ALLOWED_ERROR_MESSAGE_BYTES {
+                Err(format!(
+                    "Reject message size {response_size} exceeds the maximum allowed size of {MINIMUM_ALLOWED_ERROR_MESSAGE_BYTES}"
```
eichhorl: Should it be "maximum"?
```suggestion
                    "Reject message size {response_size} exceeds the maximum allowed size of {MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES}"
```

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: I'm okay with creating a ticket instead 👍

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: yes oups

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -56,6 +56,49 @@ pub struct CanisterHttpPoolManagerImpl {
     log: ReplicaLogger,
 }
 
+const MINIMUM_ALLOWED_ERROR_MESSAGE_BYTES: u64 = 1024; // 1KB
```
```suggestion
const MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES: u64 = 1024; // 1KB
```

AFAIU this constant is used to check the maximum size of an error message, right..?

Same for the definition at line 1735.

rs/https_outcalls/consensus/src/pool_manager.rs
```suggestion
// Checks that the response size is within the allowed limits.
// In general, the response size must be lower than max_response_bytes, however there are two caveats:
```

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -56,6 +56,49 @@ pub struct CanisterHttpPoolManagerImpl {
...
+        CanisterHttpResponseContent::Success(bytes) => {
+            let response_size = bytes.len() as u64;
+            let max_response_size = match max_response_bytes {
+                Some(response_size) => response_size.get(),
+                None => MAX_CANISTER_HTTP_RESPONSE_BYTES,
+            };
+            if response_size > max_response_size + CANDID_OVERHEAD_RESERVE_BYTES {
+                Err(format!(
+                    "Response size {response_size} exceeds the maximum allowed size of {max_response_size}"
+                ))
```
```suggestion
            let response_size_without_candid = response_size - CANDID_OVERHEAD_RESERVE_BYTES;
            if response_size_without_candid > max_response_size {
                Err(format!(
                    "Response size {response_size_without_candid} exceeds the maximum allowed size of {max_response_size}"
                ))
```

Wouldn't it make more sense to display the size of the response without the candid overhead? It just feels weird to me that the error message displays the human-readable format of a comparison which is actually not the performed comparison.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -367,31 +458,69 @@ impl CanisterHttpPoolManagerImpl {
...
+                };
+
+                match context.replication {
+                    Replication::NonReplicated(node_id) => {
+                        if node_id != share.signature.signer {
                             return Some(CanisterHttpChangeAction::HandleInvalid(
                                 share.clone(),
                                 "Share signed by node that is not the delegated node for the request".to_string(),
                             ));
                         }
```
pierugo-dfinity, line 492: Maybe you could add a unit test for this scenario, similar to the invalid cases below ("Artifact should contain response" and "Content hash does not match the response")?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -934,9 +1082,144 @@ pub mod test {
...
+                    pool.get_cache(),
+                    replica_config.clone(),
+                    SubnetType::Application,
+                    Arc::clone(&registry) as Arc<_>,
+                    MetricsRegistry::new(),
+                    log.clone(),
+                );
+
+                // TEST 1: Non-replicated request artifact is missing the response.
+                // It should be removed from the unvalidated pool.
```
```suggestion
                // It should be marked as invalid.
```
Just to follow the style of TEST 2 below

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -934,9 +1082,144 @@ pub mod test {
...
+                        timestamp: UNIX_EPOCH,
+                    });
+
+                    let changes = pool_manager.validate_shares(
+                        pool.get_cache().as_ref(),
+                        &canister_http_pool,
+                        Height::from(0),
+                    );
+
+                    assert_matches!(&changes[0], CanisterHttpChangeAction::HandleInvalid(_, _));
```
```suggestion
                    assert_matches!(
                        &changes[0],
                        CanisterHttpChangeAction::HandleInvalid(_, reason) if reason == "Artifact should contain response"
                    );```

File: rs/https_outcalls/consensus/src/pool_manager.rs [hunk 35]
```diff
@@ -1019,51 +1322,779 @@ pub mod test {
...
+
+                // 5. ASSERT: The artifact should be successfully validated and moved to the validated pool.
+                assert_eq!(changes.len(), 1);
+                assert_matches!(&changes[0], CanisterHttpChangeAction::MoveToValidated(_));
+            })
+        });
+    }
+
+    #[test]
+    fn test_oversized_reject_message_is_pruned_not_invalidated() {
```
pierugo-dfinity, line 1719: An error message generated by a dishonest replica could still be larger than the limit right? And in that case it would be marked as invalid. Maybe you could add a unit test that checks that?

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Ah Leo was too fast

Line 1735 still remains though

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Leo too fast again

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Yes I missed that, added a test, thanks!

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Thanks! just added `test_non_replicated_share_from_wrong_signer_is_invalid`

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: done

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: Yeah good point; I changed the error message to reflect the actual comparison; though I avoided making subtractions, as I don't like negative numbers :(

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -56,6 +56,51 @@ pub struct CanisterHttpPoolManagerImpl {
...
+    match &response.content {
+        CanisterHttpResponseContent::Success(bytes) => {
+            let response_size = bytes.len() as u64;
+            let max_response_size = match max_response_bytes {
+                Some(response_size) => response_size.get(),
+                None => MAX_CANISTER_HTTP_RESPONSE_BYTES,
+            };
+            let max_response_size_including_candid_overhead =
+                max_response_size + CANDID_OVERHEAD_RESERVE_BYTES;
+            if response_size > max_response_size_including_candid_overhead {
```
eichhorl, line 84: Why do we need to consider the candid overhead here? Couldn't we just define `max_response_bytes` as the maximum number of bytes including the candid encoding?

mihailjianu1: I agree we could. It is actually defined like that for transformed responses. 
However if a response is not transformed, currently `max_response_bytes` is only enforced on the sum of HTTP headers' sizes + the body size.

So there are three reasons I'd rather not enforce `max_response_bytes` on the untransformed candid encoding right now:
1. It could break some current http request (those whose size becomes > `max_response_bytes` when encoded)
2. It becomes a bit more difficult to explain to canisters what exactly is enforced by `max_response_bytes`
3. `max_response_bytes` will be deprecated in the near future anyway 

So I'd rather just keep this as is for now, ensuring malicious replicas can't send large artifacts, while keeping the same behaviour as before. wtyd?

eichhorl: Makes sense, thanks for the explanation! 👍

eichhorl (approved):

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: added ticket

rs/https_outcalls/consensus/src/pool_manager.rs
mihailjianu1: went with `get_unvalidated_artifacts` in the end

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Fair point, thanks!

berestovskyy (approved):

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -1062,8 +1402,925 @@ pub mod test {
...
+        });
+    }
+
+    #[test]
+    fn test_reject_message_is_valid_when_context_limit_is_too_low() {
+        ic_test_utilities::artifact_pool_config::with_test_pool_config(|pool_config| {
+            with_test_replica_logger(|log| {
+                use ic_types::canister_http::CanisterHttpReject;
+
+                const MAXIMUM_ALLOWED_RESPONSE_BYTES: u64 = 1024;
```
```suggestion
```

nit

mihailjianu1: doen thanks

rs/types/types/src/canister_http.rs
kpop-dfinity: Sounds good, thanks!

kpop-dfinity (approved):

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -290,10 +333,45 @@ impl CanisterHttpPoolManagerImpl {
...
-                Ok(response) => {
+                Ok(mut response) => {
+                    // Truncate the reject message if it's too long.
+                    //
+                    // The "happy path" response is organically bounded by max_response_bytes, however we need to set a
+                    // limit for the error message as well.
+                    //
+                    // The current limit is 1KB, which should be reasonable for an error message.
+                    if let CanisterHttpResponseContent::Reject(reject) = &mut response.content {
+                        let max_len = MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES as usize;
```
kpop-dfinity: nit: maybe it would make sense for `MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES` to be of `usize` type so we don't have to cast it. If we change the type, I would inline `max_len`

mihailjianu1: done

PR #7093, opened 2025-10-06
feat: [MR-736] Credit the caller any cycles attached to late responses

alin-at-dfinity:
As part of guaranteeing the refund of cycles attached to best-effort calls, we must ensure that cycles attached to any messages dropped silently during induction (namely best-effort responses received too late, once a `SYS_UNKNOWN` response was already enqueued; or after it has been processed, consuming the callback) are always refunded to the caller.

### DIFF
--- a/rs/replicated_state/tests/system_state.rs
+++ b/rs/replicated_state/tests/system_state.rs
@@ -424,7 +424,10 @@ fn induct_messages_to_self_duplicate_best_effort_response() {
     assert_eq!(Some(CanisterMessage::Request(request)), fixture.pop_input());
 
     // Expire the callback.
-    fixture.time_out_callbacks(CoarseTime::from_secs_since_unix_epoch(u32::MAX));
+    assert_eq!(
+        (1, Vec::new()),
+        fixture.time_out_callbacks(CoarseTime::from_secs_since_unix_epoch(u32::MAX))
+    );
 
     // A few rounds later, have the running call context produce a response.
     fixture.push_output_response(response);
@@ -434,6 +437,7 @@ fn induct_messages_to_self_duplicate_best_effort_response() {
     assert!(!fixture.system_state.queues().has_output());
 
     // Pop the timeout reject response and execute it (consuming the callback).
+    // The late response was silently dropped, as a duplicate.
     assert_matches!(fixture.pop_input(), Some(CanisterMessage::Response(resp)) if resp.response_payload == Payload::Reject(RejectContext::new(RejectCode::SysUnknown, "Call deadline has expired.")));
     assert_matches!(
         fixture.system_state.unregister_callback(callback),


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1038,6 +1066,7 @@ impl SystemState {
             (RequestOrResponse::Response(response), CanisterStatus::Stopped)
                 if response.is_best_effort() =>
             {
+                self.credit_refund(response);
```
alin-at-dfinity, line 1069: This; plus other calls to `credit_refund()` below; plus direct increases of `cycles_balance` (where the borrow checker won't let us call `credit_refund()`); are the core of the change.

Everything else is either required changes to method signatures; refactoring to allow validating cycle balances; or tests.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -885,8 +885,8 @@ impl StreamHandlerImpl {
                     // Message silently dropped, all done.
                     Ok(false) => {
                         self.observe_inducted_message_status(msg_type, LABEL_VALUE_DROPPED);
-                        // Cycles were lost.
-                        Accept(msg_cycles)
+                        // Cycles were credited.
```
oggy-dfin, line 888: It seems that we never expect to have `Accept(non-zero-value)` any more (the only case where we do this is a critical error). Can we now remove the payload from the `Accept` variant again?

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1024,6 +1024,34 @@ impl SystemState {
...
+            input_queue_type,
+        );
+
+        #[cfg(debug_assertions)]
+        self.assert_balance_with_messages(
+            balance_before,
+            res.as_ref().err().map(|(_, msg)| msg.cycles()),
+        );
+
+        res
```
alin-at-dfinity, line 1044: This (repeated 4x) is the best I was able to come up with as a way of comparing before and after cycle balances. It will also take into account the "outbound refund pool" in the following PR.

Any better suggestions are welcome.

rs/messaging/src/routing/stream_handler.rs
oggy-dfin: Ideally this would also enable us to remove the `bool` from most of the `push_input` method stack and just have the unit there

rs/replicated_state/src/canister_state/system_state.rs
oggy-dfin: The only alternative that pops to mind is to have a single "no cycles lost" check in `StateMachineImpl::execute_round` that would span all the message routing phases. That sounds nicer conceptually since with your current approach we need to manually remember all the places where we could drop messages, and we don't notice if we forget something, whereas having it on a higher level alleviates this. But it sounds to me like it would be quite complicated to deal with both misrouted messages (e.g., recipient deleted or migrated away) and reject signals; maybe you have some ideas how one could simplify that. Obviously it'd also be less efficient since we'd have to iterate over all canisters but that's hopefully fine for something that's `debug_assertions`.

So, in short, no better suggestion :) But thought I'd mention it in case it inspires a more robust solution.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1855,21 +1978,21 @@ pub(crate) fn push_input(
...
 ///    for a best-effort response (because the callback might have expired and
 ///    been closed; or because the callback is executing -- aborted or paused).
 ///  * `Err(StateError::NonMatchingResponse)` when a matching `callback_id` was
 ///    not found for a guaranteed response.
-///  * `Err(StateError::NonMatchingResponse)` if the response details do not
-///    match those of the callback.
-pub(crate) fn should_enqueue_input(
+///  * `Err(StateError::NonMatchingResponse)` when a matching `callback_id` was
+///    found, but the response details do not match those of the callback.
+pub(crate) fn look_up_callback(
```
oggy-dfin: Nitpick: I expected `look_up_callback` to return the callback. Maybe `has_callback` or similar?

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1391,13 +1451,16 @@ impl SystemState {
...
-                .is_err()
-            {
-                return;
+            match self.queues.induct_message_to_self(self.canister_id) {
+                // Message successfully inducted.
+                Ok(None) => {}
+                // Silently dropped duplicate best-effort response.
+                Ok(Some(response)) => {
+                    // Borrow checker does not allow calling `credit_refund()` here.
+                    self.cycles_balance += response.refund;
```
derlerd-dfinity, line 1460: It seems that `credit_refund` does more than just adding the cycles. Is it intentional that we only do the add cycles part here (and similarly above)?

derlerd-dfinity (approved):
The change looks good to me. 

As a sanity check question: as per my understanding this change is safe because we never attach any cycles to timed out / shed message where we don't know whether a reply might still arrive. Or in other words we never run the risk of "duplicating" cycles. Based on a quick check in the code we only produce refunds for outbound requests (i.e., ones where we know that the canister at hand is the only one who has ever seen them). So this should be in line with that we need to do. 

I thought I'd quickly write this down here to doublecheck whether this matches your understanding as well?

rs/messaging/src/routing/stream_handler.rs
stiegerc: The `bool` on `push_input` is there to distinguish the case of best-effort message dropped and message actually inducted. Something to do with the metrics IIRC. It is not tied directly to cycles.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -986,9 +986,9 @@ impl SystemState {
 
     /// Pushes a `RequestOrResponse` into the induction pool.
     ///
-    /// If the message is a `Request`, reserves a slot in the corresponding
-    /// output queue for the eventual response; and the maximum memory size and
-    /// cycles cost for sending the `Response` back. If it is a `Response`,
+    /// If the message is a `Request`, reserves a slot in the corresponding output
+    /// queue for the eventual response; and guaranteed response memory for the
+    /// maximum `Response` size if it's guaranteed response. If it is a `Response`,
```
stiegerc:
```suggestion
    /// maximum `Response` size if it's a guaranteed response. If it is a `Response`,
```

rs/replicated_state/tests/system_state.rs
```diff
@@ -402,6 +404,50 @@ fn induct_messages_to_self_full_queue() {
...
+
+    // Induct it into the input queue.
+    fixture.induct_messages_to_self();
+
+    // Pop and start executing it (pretend it's waiting multiple rounds for
+    // downstream calls).
+    assert_eq!(Some(CanisterMessage::Request(request)), fixture.pop_input());
+
+    // Expire the callback.
+    fixture.time_out_callbacks(CoarseTime::from_secs_since_unix_epoch(u32::MAX));
```
stiegerc: Maybe assert on the return value here too?

rs/replicated_state/tests/system_state.rs
```diff
@@ -402,6 +404,50 @@ fn induct_messages_to_self_full_queue() {
...
+
+    // A few rounds later, have the running call context produce a response.
+    fixture.push_output_response(response);
+
+    // Try inducting the response and check that it was consumed.
+    fixture.induct_messages_to_self();
+    assert!(!fixture.system_state.queues().has_output());
+
```
stiegerc, line 438: Timing out callbacks pushes a compact response right? So at this point there should be `[time out response, real response]` in the input queue, except the real response got dropped since there already is a response for the call in there so it's just `[time out response]`.
I got quite confused for a moment. Could you add that to the comment?

rs/replicated_state/src/canister_state/system_state.rs
stiegerc: Maybe such a high level check is worth it even as an addition rather than a replacement? Just in case we ever add a new induction path.

alin-at-dfinity: Right. I was thinking of adding such a high level check somewhere in `stream_handler.rs` in a follow-up PR. But just like this, there would be at least 3 different places where the check would need to be made: induction, routing (because we reject some messages that we fail to route) and `induct_messages_to_self()`. So I'd rather look at it as more of a "defense in depth".

rs/messaging/src/routing/stream_handler.rs
alin-at-dfinity: I am getting rid of the cycles amount in the follow-up #7107 change, but it requires rejecting (as opposed to silently dropping) misrouted responses, since that's the one not yet handled instance of lost cycles remaining.

rs/replicated_state/src/canister_state/system_state.rs
alin-at-dfinity: In addition to a couple of `debug_asserts`, `credit_refund()` only calls `add_cycles(_, CyclesUseCase::NonConsumed)`. In turn, `add_cycles()`, calls `observe_consumed_cycles_with_use_case()`. Which simply returns immediately if called with `use_case == CyclesUseCase::NonConsumed`.

So in terms of functionality, this is equivalent with calling `credit_refund()`. Of course, the implementation of the latter might change in the future, which is why this is not ideal. But I don't have a good way of working around the borrow checker here, short of making the call from an `unsafe` block.

rs/replicated_state/src/canister_state/system_state.rs
alin-at-dfinity: Great suggestion. Renamed, thanks.

rs/replicated_state/tests/system_state.rs
alin-at-dfinity: Done.

rs/replicated_state/tests/system_state.rs
alin-at-dfinity: Added, thanks for the suggestion.

rs/replicated_state/src/canister_state/system_state.rs
derlerd-dfinity: Makes sense. Thanks a lot.

PR #7098, opened 2025-10-06
test(crypto): CRP-2684 Add integration tests of VetKD protocol

randombit:
This tests various errornous protocol scenarios to ensure the implementation behaves as expected.

### DIFF
--- a/rs/crypto/tests/vetkd.rs
+++ b/rs/crypto/tests/vetkd.rs
@@ -1,3 +1,6 @@
+use ic_crypto_internal_csp::public_key_store::proto_pubkey_store::ProtoPublicKeyStore;
+use ic_crypto_internal_csp::secret_key_store::proto_store::ProtoSecretKeyStore;
+use ic_crypto_internal_logmon::metrics::CryptoMetrics;
 use ic_crypto_temp_crypto::CryptoComponentRng;
 use ic_crypto_temp_crypto::TempCryptoComponentGeneric;
 use ic_crypto_test_utils::crypto_for;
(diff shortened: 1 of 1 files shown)


randombit:
@fspreiss This tests most of the scenarios identified in CRP-2684 however some seem impossible to test without adding some kind of testing facility to the remote vault. In particular the various scenarios like `should_vetkd_verify_encrypted_key_err_with_invalidargumentmasterpublickey_if_master_public_key_in_store_is_invalid` requires that there is some way to modify/invalidate the information stored on the vault.

We *could* add a `cfg(test)`-only vault API to perform malicious modification to the vault's storage. But I'm not sure if this was intended/expected as part of the ticket?

The tests that I was unsure how to proceed with have `todo!` and will fail.

rs/crypto/tests/vetkd.rs
```diff
@@ -7,98 +10,680 @@ use ic_crypto_test_utils_ni_dkg::{
...
+#[test]
+fn should_vetkd_create_encrypted_key_share_err_with_internal_error_if_master_public_key_in_store_is_invalid()
+ {
+    todo!("how to modify the store??");
+}
+
+#[test]
+fn should_vetkd_create_encrypted_key_share_err_with_transientinternalerror_if_vault_returns_transient_error()
+ {
+    todo!("how to cause vault transient error?");
```
fspreiss: We'd have to use a mock: create a `MockLocalCspVault` and configure it to return a `TransientInternalError` whenever `create_encrypted_vetkd_key_share` is called.

For this mock then to used in a test, we have to be able to construct temp crypto components from a given (mock)vault: currently this is not possible, so we'd have to adapt the `TempCryptoBuilder` with a respective `with_vault` method: I had a quick look and it seems this wouldn't require too much work. We can also look at it together if you want.

rs/crypto/tests/vetkd.rs
```diff
@@ -7,98 +10,680 @@ use ic_crypto_test_utils_ni_dkg::{
...
+        .expect("Share creation unexpectedly failed");
+
+    let corrupted_shares = n_random_nodes_in(
+        server.config.receivers().get(),
+        server.config.max_corrupt_dealers(),
+        &mut rng,
+    );
+
+    for dealer in &corrupted_shares {
+        corrupt_share_signature(shares.get_mut(&dealer).expect("Missing share"), &mut rng);
```
fspreiss: What is corrupted here is the share's `node_signature`. However, `VetKdProtocol::combine_encrypted_key_shares` doesn't care about this field at all (only `VetKdProtocol::verify_encrypted_key_share` [does](https://github.com/dfinity/ic/blob/e963d11d0f5bb9c8fa2049911e89dfa50197038b/rs/crypto/src/vetkd/mod.rs#L304C9-L312)).

So, IIUC, the test currently just exercises the [OK-case](https://github.com/dfinity/ic/blob/e963d11d0f5bb9c8fa2049911e89dfa50197038b/rs/crypto/src/vetkd/mod.rs#L393) of `combine_all`, rather than the [InvalidShares-case](https://github.com/dfinity/ic/blob/e963d11d0f5bb9c8fa2049911e89dfa50197038b/rs/crypto/src/vetkd/mod.rs#L400-L442).

If we want a bullet-proof way to ensure that the test really enters the expected case, we could ensure that [this log](https://github.com/dfinity/ic/blob/e963d11d0f5bb9c8fa2049911e89dfa50197038b/rs/crypto/src/vetkd/mod.rs#L401-L403) is written: [this](https://github.com/dfinity/ic/blob/e963d11d0f5bb9c8fa2049911e89dfa50197038b/rs/crypto/tests/integration_test.rs#L1304-L1351) test shows an example for how we can check for certain logs.

rs/crypto/tests/vetkd.rs
```diff
@@ -7,98 +10,680 @@ use ic_crypto_test_utils_ni_dkg::{
...
+
+#[test]
+fn should_vetkd_verify_encrypted_key_err_with_internal_error_if_pub_coeffs_in_store_are_empty() {
+    todo!("how to modify store?");
+}
+
+#[test]
+fn should_vetkd_verify_encrypted_key_err_with_invalidargumentmasterpublickey_if_master_public_key_in_store_is_invalid()
+ {
+    todo!("how to modify store?");
```
fspreiss: > scenarios like should_vetkd_verify_encrypted_key_err_with_invalidargumentmasterpublickey_if_master_public_key_in_store_is_invalid requires that there is some way to modify/invalidate the information stored on the vault

Note that the "store" here is not the vault (which we could make behave however we want by using a mock vault: see my other comment), but rather the `CryptoComponentImpl`'s [lockable_threshold_sig_data_store: LockableThresholdSigDataStore](https://github.com/dfinity/ic/blob/56ee50fb4b674e6cb46f918eceaa0b5becbc5b85/rs/crypto/src/lib.rs#L58), which is what is populated whenever `NiDkgAlgorithm::load_transcript` is called.

I doubt that we can construct invalid transcripts that we can load that then lead to the errors we are seeking, but my guess is that there are checks in place that prevent loading of invalid transcripts (to be checked).

The clean way to approach this would be to change the (temp) crypto component so that can can provide a custom ThresholdSigDataStore (e.g., via a newly introduced trait, and a respective mock implementation) via, e.g., `TempCryptoBuilder::with_threshold_sig_data_store` or similar.

The question is, whether it is worth the effort.

rs/crypto/tests/vetkd.rs
```diff
@@ -7,98 +10,680 @@ use ic_crypto_test_utils_ni_dkg::{
...
+
+#[test]
+fn should_vetkd_create_encrypted_key_share_err_if_pub_coeffs_in_store_are_empty() {
+    let mut rng = reproducible_rng();
+    let mut server = VetKDTestServer::new(&mut rng);
+    let client = VetKDTestClient::new(&mut rng, &server);
+    let vetkd_args = client.create_args(&server.dkg_id);
+
+    server.modify_stored_keys(&mut rng, |_sks, _pks| {
+        todo!("how to modify store?");
```
fspreiss: It is my understanding that this and all other uncommented `todo!`'s are an instance of [this](https://github.com/dfinity/ic/pull/7098/files#r2414249727).

fspreiss (commented):
Thanks, @randombit! I mainly looked at the `todo!`s and at `should_vetkd_combine_shares_succeed_if_reconstruction_threshold_many_shares_are_valid`, which I'd consider the most important test of the PR/ticket, i.e., I didn't look too closely at the rest of the code.

rs/crypto/tests/vetkd.rs
randombit: Deferred these tests in CRP-2945

rs/crypto/tests/vetkd.rs
randombit: I fixed the test to corrupt the shares in a way that causes `combine_all` to actually fail, and extended the test to capture the logs and verify that the expected log is emitted.

rs/crypto/tests/vetkd.rs
```diff
@@ -7,181 +7,625 @@ use ic_crypto_test_utils_ni_dkg::{
...
+    NiDkgId {
+        dealer_subnet: dkg_id.dealer_subnet,
+        dkg_tag: dkg_id.dkg_tag.clone(),
+        target_subnet: dkg_id.target_subnet,
+        start_block_height: (dkg_id.start_block_height.get() + 1000000).into(),
+    }
+}
+
+#[test]
+fn should_vetkd_consistently_derive_the_same_vetkey_given_sufficient_shares() {
```
fspreiss: With the added `_vetkd`, I find this test name doesn't read well any more. In fact this is true for all test names. How about this:

1. move all tests named `should_vetkd_create_encrypted_key_share_*` in a module `create_encrypted_key_share` and remove `_vetkd_create_encrypted_key_share` from the test names.
2. Similar for tests named `should_vetkd_verify_key_share_*`: move to module `verify_encrypted_key_share` and adapt test names
3. Similar for tests named `should_vetkd_combine_shares_*`
4. Similar for tests named `should_vetkd_verify_encrypted_key_*`
5. Finally, there should only be `should_vetkd_consistently_derive_the_same_vetkey_given_sufficient_shares` remaining that is not in a module: now we can remove the obsolete `_vetkd` from the name.

WDYT?

randombit: I guess my take would be it reads fine to me and I liked being able to run all of (and only) the VetKD tests with `cargo test should_vetkd_` *BUT* I also don't care that strongly about it so I'll adjust as you suggest.

randombit: Done

rs/crypto/tests/vetkd.rs
randombit: Moved to CRP-2944

rs/crypto/tests/vetkd.rs
fspreiss: Ok, now I understand why the tests were named in that way. Running just the vetKD tests should also work with `cargo test -p ic-crypto --test vetkd`, but `cargo test should_vetkd_` is a bit easier to type indeed. Anyway, thanks for adapting the test names!

fspreiss (approved):

PR #7099, opened 2025-10-06
refactor(cmc/root): Fix a few places where ic_cdk_macros are not used

jasonz-dfinity:
# Why

There are a few places (root/cmc) not using the ic_cdk_macros without a good reason.

# What

* Use `ic_cdk_macros::query(hidden = true)` for `total_cycles_minted`
* Switch `total_cycles_minted` to return candid rather than protobuf. Note that the current value on the mainnet is at 70% of u64::MAX, so the u64 doesn't make sense anyway. This is a breaking change, but the method has never been exposed (not in .did)
* Use `#[init]` for root

### DIFF
--- a/rs/nns/cmc/src/main.rs
+++ b/rs/nns/cmc/src/main.rs
@@ -1,14 +1,13 @@
 #![allow(deprecated)]
-use candid::{CandidType, Encode};
+use candid::{CandidType, Encode, Nat};
 use core::cmp::Ordering;
 use cycles_minting_canister::*;
-use dfn_protobuf::ProtoBuf;
 use environment::Environment;
 use exchange_rate_canister::{
     RealExchangeRateCanisterClient, UpdateExchangeRateError, UpdateExchangeRateState,
 };
 use ic_cdk::{
-    api::call::{CallResult, ManualReply, reply_raw},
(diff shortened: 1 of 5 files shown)


rs/nns/cmc/src/main.rs
```diff
@@ -2321,11 +2319,9 @@ fn ensure_balance(
...
 }
 
-#[unsafe(export_name = "canister_query total_cycles_minted")]
-fn total_cycles_minted() {
-    let value: u64 = with_state(|state| state.total_cycles_minted.get().try_into().unwrap());
-    let response = ProtoBuf::new(value).into_bytes().unwrap();
-    reply_raw(&response);
+#[query(hidden = true)]
+fn total_cycles_minted() -> Nat {
+    with_state(|state| state.total_cycles_minted.get().into())
```
max-dfinity, line 2324: Can we see if this is called at all?  Does dashboard rely on this at all?

jasonz-dfinity: Boundary node metrics show that it hasn't been called in the last 30 days (except the ones from myself)

max-dfinity: Thanks!

PR #7105, opened 2025-10-07
test(btc-adapter): fix flaky test in generate_to_address call

ninegua:
In btc adapter's integration test, occasionally the call to `generate_to_address` would return the following error:
```
JsonRpc(Transport(SocketError(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" })))
```
In this case we can retry a couple times to improve the success rate.


basvandijk (approved):
Thanks!

I can imagine the same issue could happen on any method of the RpcClient. A generalized approach would be a good follow up.

gregorydemay (approved):
Thanks for looking into this!

PR #7106, opened 2025-10-07
feat(nns): introduce `get_neuron_index` interface

ShahriarJavidi:


rs/nns/governance/canister/governance.did
```diff
@@ -791,6 +791,15 @@ type NeuronInFlightCommand = record {
   timestamp : nat64;
 };
 
+type GetNeuronIndexRequest = record {
+  exclusive_start_neuron_id: opt NeuronId;
+  page_size: opt nat32;
+};
+
+type GetNeuronIndexResponse = record {
+  neurons: vec Neuron;
```
daniel-wong-dfinity-org: I think `Neuron` is for when you are allowed to see the entire neuron. When a neuron is private (and the caller is not allowed to see the whole neuron), using this type might be ambiguous and/or confusing. For example, one of the fields in `Neuron` is

```
recent_ballots : vec BallotInfo;
```

This is a "private" field, i.e. only privileged principals are allowed to see this (mainly, the controller and hotkeys, but maybe others too...). To prevent unprivileged principals from seeing this field, it would have to be returned empty. However, that is misleading: if you just return an empty `recent_ballots`, a reasonable interpretation is that the neuron has not voted recently. However, that's not the real reason that the caller is getting an empty `recent_ballots`.

Basically, I think what we need is a new type that can support both

1. the caller can see the whole neuron
2. the caller only gets a limited view of the neuron

Alternatively, we can have a type that only gives a limited view, and just use that. It used to be that `NeuronInfo` served that purpose, but then, we decided to take more information away (ballots, and Community Fund membership). The disadvantage of this is that if the caller is allowed to see everything about the neuron, then this API won't actually give them everything. But maybe this disadvantage doesn't matter in practice, since the vast majority of neurons are not public (I think it's on the order of 1%).

Anyway, we probably need to discuss this using a more interactive medium. Consider this feedback an introduction to the issue that I'm seeing here.

rs/nns/governance/canister/canister.rs
```diff
@@ -377,6 +377,12 @@ fn list_neurons(req: ListNeurons) -> ListNeuronsResponse {
     governance().list_neurons(&req, caller())
 }
 
+#[query]
+fn get_neuron_index(_req: GetNeuronIndexRequest) -> GetNeuronIndexResponse {
+    debug_log("get_neuron_index");
+    GetNeuronIndexResponse::default()
```
daniel-wong-dfinity-org: A more forceful alternative would be to panic. I don't feel strongly about this. I like it slightly better because it more clearly says "NOT IMPLEMENTED YET", but in practice, nobody is going to start relying on this before it is ready, so it almost certainly doesn't matter...

rs/nns/governance/canister/governance.did
```diff
@@ -1351,6 +1360,7 @@ service : (Governance) -> {
     ) query;
   get_network_economics_parameters : () -> (NetworkEconomics) query;
   get_neuron_ids : () -> (vec nat64) query;
+  get_neuron_index: (GetNeuronIndexRequest) -> (GetNeuronIndexResponse) query;
```
daniel-wong-dfinity-org: We have many methods that return Result (with the Err always being of type GovernanceError). Not all methods do this, but I wish they did. Nevertheless, maybe we can move in that direction here.

To avoid adding yet another `Result_57` type (these were generated automatically, not by humans), I think we can do (something like)

```
get_widget : (GetWidgetRequest) -> (variant {
    Ok : Widget;
    Err : GovernanceError;
});
```
(the "anonymous" Result approach)

or

```
get_widget : (GetWidgetRequest) -> (GetWidgetResult);

...

type GetWidgetResult = variant {
    Ok : Widget;
    Err : GovernanceError;
};
```
(the "named" Result approach).

rs/nns/governance/canister/governance.did
ShahriarJavidi: We decided to return `NeuronInfo` through this interface.

rs/nns/governance/canister/governance.did
ShahriarJavidi: Chose the named result approach. Furthermore, the result is going to be wrapped in a `Response` with an optional field called `result`. This allows more compatibility.

rs/nns/governance/api/src/types.rs
```diff
@@ -4390,3 +4390,37 @@ pub struct MaturityDisbursement {
...
+        neurons: Vec<NeuronInfo>,
+    }
+
+    #[derive(
+        candid::CandidType, candid::Deserialize, serde::Serialize, Debug, Clone, PartialEq,
+    )]
+    pub enum GetNeuronIndexResult {
+        Ok(NeuronIndex),
+        Err(GovernanceError),
+    }
```
daniel-wong-dfinity-org: Instead of defining a new *Result type, I think we can just use Rust's built-in Result type?

(Keeping the Result type definition in the .did file is fine though.)

rs/nns/governance/api/src/types.rs
```diff
@@ -4390,3 +4390,37 @@ pub struct MaturityDisbursement {
...
+
+#[derive(
+    candid::CandidType, candid::Deserialize, serde::Serialize, Debug, Default, Clone, PartialEq,
+)]
+pub struct GetNeuronIndexRequest {
+    exclusive_start_neuron_id: Option<NeuronId>,
+    page_size: Option<u32>,
+}
+
+pub mod neuron_index {
```
daniel-wong-dfinity-org: I don't think we have (sub)modules for organizing other definitions? We do have submodules, but those are from the days where we use Prost to generate these definitions, and in Prost, when your .proto file has

```
message Outer {
    message Inner { ... }
}
```

Then, Prost would generate

```
struct Outer { ... }

mod outer {
    struct Inner { ... }
}
```

But you are not doing anything like that here (since there is no .proto that this is base off of).

Also, if you want to put all neuron index-related definitions in a submodule, why leave out the request type?

rs/nns/governance/canister/governance.did
daniel-wong-dfinity-org: I'd suggest that we not have this extra layer of wrapping, because

1. That is pretty different from what we've been doing.
2. I don't see how that is practically helping us (down the road)?

rs/nns/governance/api/src/types.rs
ShahriarJavidi: Done.

PR #7107, opened 2025-10-07
feat: [MR-736] Reject misrouted responses

alin-at-dfinity:
Reject responses determined to be incorrectly routed during induction.
 * For guaranteed responses, it is safer to ensure their eventual delivery after rerouting. (And it's a critical error case anyway, we don't expect it to occurr.)
 * Best effort responses may legitimately be routed to the wrong subnet after a manual canister migration. And guaranteed delivery of refunds requires their rejection and rerouting.

Also simplify a bit the logic around observing cycles lost due to dropped messages, now that there are fewer instances of that.


rs/messaging/src/routing/stream_handler.rs
```diff
@@ -963,12 +929,14 @@ impl StreamHandlerImpl {
...
+            // timed out.
             Some(_) if msg.is_best_effort() && matches!(msg, RequestOrResponse::Response(_)) => {
-                self.observe_inducted_message_status(msg_type, LABEL_VALUE_DROPPED);
-                // Cycles were lost.
-                Accept(msg_cycles)
+                self.observe_inducted_message_status(
+                    msg_type,
+                    LABEL_VALUE_RECEIVER_LIKELY_MIGRATED,
+                );
+                Reject(RejectReason::CanisterMigrating, msg)
```
alin-at-dfinity, line 921: This, and the change immediately below are the only change in behavior.

Everything else is refactoring.

derlerd-dfinity: Based on the comment above, I'm surprised that we ended up putting `Accept` here. Was this a bug or was there some reason that we used to accept here before this PR?

derlerd-dfinity (approved):
The change looks good to me. Thank you! Only left one question.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -999,22 +967,13 @@ impl StreamHandlerImpl {
                     LABEL_VALUE_RECEIVER_SUBNET_MISMATCH,
                 );
                 self.metrics.critical_error_receiver_subnet_mismatch.inc();
-                // Cycles are lost.
-                Accept(msg.cycles())
+                Reject(RejectReason::CanisterMigrating, msg)
```
stiegerc, line 952: Are we certain this change for guaranteed responses won't have bad side effects, e.g. sending responses again and again due to a bug or something?
I mean why are we changing the behaviour for guaranteed responses when the cycle refunds are supposed to be about best-effort responses?
Attaching cycles is not a safe store anyways if for example a call takes out cycles from a message, then sends those downstream, then traps. Whatever was taken out is lost right? I mean otherwise you could send cycles back to yourself and then also get them refunded due to the trap.

stiegerc: Then again maybe having such infinite loops would be preferable if they get delivered after a fix.
Anyways just making sure we understand the implications here.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -879,14 +845,13 @@ impl StreamHandlerImpl {
                     Ok(true) => {
                         self.observe_inducted_message_status(msg_type, LABEL_VALUE_SUCCESS);
                         self.observe_inducted_payload_size(payload_size);
-                        Accept(Cycles::zero())
+                        Accept
                     }
 
-                    // Message silently dropped, all done.
+                    // Message silently dropped, any refund was already credited.
```
stiegerc, line 833: I kinda wonder whether it might make sense to refund manually here rather than implicitly in `push_input` 🤔.

rs/messaging/src/routing/stream_handler.rs
alin-at-dfinity: `Accept` in this context meant "drop silently". Which is the most efficient thing to do if we observe a best-effort response that we know to be late. Rerouting would simply end up delivering it to the right subnet, which will then drop it silently.

But with the introduction of guaranteed refunds we must now reject and reroute all responses / refunds until they are eventually delivered to the caller's subnet.

rs/messaging/src/routing/stream_handler.rs
alin-at-dfinity: This is partly about consistency (with the best-effort case) and partly about delivery guarantees. As you noted in your second comment, while this runs the risk of an infinite loop in the case of a bug, we should guarantee delivery regardless of such bugs. Such a bug would trigger a critical error (presumably repeatedly, if we're in an infinite loop) so we can ideally fix the bug and get the response correctly routed.

I suppose that a combination of infinite rerouting bug and large responses could lead to one or more streams stalling until we intervene, but I guess that's one of the risks of guaranteed delivery.

rs/messaging/src/routing/stream_handler.rs
alin-at-dfinity: I suppose it's possible, but I imagine there are a number of other places where we call `push_input()` and ensuring that each of them properly refunds cycles under the right circumstances is tricky. (Even adding `asserts` / `debug_asserts` in all locations where we know no refund will have to be handled; and documenting them; is quite the overhead.)

rs/messaging/src/routing/stream_handler.rs
derlerd-dfinity: Indeed, without refunds this was the right thing to do. Thanks!

rs/messaging/src/routing/stream_handler.rs
stiegerc: Also a dishonest subnet would just get its response rejected continuously which doesn't pose a problem either. At worst we can recover so it should be ok.

rs/messaging/src/routing/stream_handler.rs
stiegerc: You're probably right, I was just thinking about annoying borrow checker issues and the like. But then you'd have to look for the canister twice in any case so never mind.

PR #7112, opened 2025-10-07
refactor(cketh/ckerc20): use `evm_rpc_client` in ckETH minter

lpahlavi:


rs/ethereum/evm-rpc-client/src/lib.rs
```diff
@@ -1,341 +0,0 @@
...
-        config: Option<Config>,
-        args: In,
-    ) -> MultiRpcResult<Out>
-    where
-        In: CandidType + Send + Clone + Debug + 'static,
-        Config: CandidType + Send + Clone + Debug + 'static,
-        Out: CandidType + DeserializeOwned + Debug + 'static,
-    {
-        let mut retries = 0;
-        let mut attached_cycles = self.min_attached_cycles;
```
gregorydemay, line 142: How is this handled in the new client (retrying with double the amount of cycles in case there were not enough cycles)? I think this is something we need.

lpahlavi: Good question! Actually I was hoping we could streamline this to make sure we always need a single call. 

Since, with the new `ic-cdk`, we should be refunded extra cycles attached to perform HTTP outcalls, I thought it could make more sense to simply always attach enough cycles that calls never fail due to insufficient cycles. 

The one risk with this strategy is that if something changes with the EVM RPC canister, there would be a chance we suddenly don't attach sufficient cycles anymore (e.g. if the number of default providers increases). My reasoning here though, was that we anyways already have a maximum number of retries, and hence a maximum number of cycles attached. Why then not simply always attach this many cycles? WDYT? 

If you don't agree, I guess the most elegant thing to do would be to add some functionality to the client itself (i.e. in `evm_rpc_client`) to handle this, as I could see other users wanting to do the same. One option could also be to:

1. Try with a hardcoded number of cycles
2. If that fails, call one of the new `CyclesCost` endpoints
3. Try again with that many cycles + x%

gregorydemay: > Good question! Actually I was hoping we could streamline this to make sure we always need a single call.

> Since, with the new ic-cdk, we should be refunded extra cycles attached to perform HTTP outcalls, I thought it could make more sense to simply always attach enough cycles that calls never fail due to insufficient cycles.

I think that's a good observation, because the EVM RPC canister takes care of doubling the `max_response_bytes` and retrying if necessary. One trouble though is that the max number of retries depends a lot on the strategy chosen by the canister to retry (currently it doubles `max_response_bytes` but it could do something else) and that's I think too brittle. 

Calling `CyclesCost` won't help because at that point you don't know  what the last tried out value for `max_response_bytes`  for your request was. So I think we need a retry strategy in the client. WDYT?

lpahlavi: I've updated the `evm_rpc_client` dependency version to v0.2.0 and added a retry strategy to the client so the behaviour should be the exact same as before now.

nmattia (approved):
Bazel changes LGTM!

gregorydemay (approved):
Thanks @lpahlavi for pushing this endeavor!

PR #7118, opened 2025-10-07
test: add management canister candid backward-compatibility test

mraszyk:
This PR adds a management canister candid backward-compatibility test to detect breaking changes to the public interface of the management canister on CI.

### DIFF
--- a/rs/types/management_canister_types/BUILD.bazel
+++ b/rs/types/management_canister_types/BUILD.bazel
@@ -101,6 +101,7 @@ rust_test(
 did_git_test(
     name = "management_canister_did_git_test",
     did = ":tests/ic.did",
+    tags = ["didc"],  # to be able to override this test using the GitHub PR label `CI_OVERRIDE_DIDC_CHECK`
 )
 
 rust_test(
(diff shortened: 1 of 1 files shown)


rs/types/management_canister_types/BUILD.bazel
```diff
@@ -97,6 +98,12 @@ rust_test(
     crate = ":management_canister_types",
 )
 
+did_git_test(
+    name = "management_canister_did_git_test",
+    did = ":tests/ic.did",
+    tags = ["didc"],  # to be able to override this test using the GitHub PR label `CI_OVERRIDE_DIDC_CHECK`
```
mraszyk: @nmattia Does my comment apply, i.e., is adding a `didc` tag the way to enable overriding this test using the GitHub PR label `CI_OVERRIDE_DIDC_CHECK`?

nmattia: I believe it's [set by default](https://github.com/dfinity/ic/blob/master/bazel/candid.bzl#L72-L86) when you call `did_git_test`:

```python
def did_git_test(name, did, **kwargs):
    """Defines a test checking whether a Candid interface evolves in a backward-compatible way.


    Args:
      name: the test name.
      did: the Candid file, must be a repository file.
      **kwargs: additional keyword arguments to pass to the test rule.
            enable_also_reverse (bool, optional): whether the test should also run candid checks in reverse order
    """
    tags = kwargs.pop("tags", [])
    for tag in ["local", "no-sandbox", "smoke", "didc"]:
        if not tag in tags:
            tags.append(tag)


    _did_git_test(name = name, did = did, tags = tags, **kwargs)
```

mraszyk: Thanks for looking into this! I've removed the explicit tag declaration from the test now.

PR #7121, opened 2025-10-07
test: check if bazel-test-all (sans system-tests) can run on Namespace x86_64 runners

basvandijk:
DO NOT MERGE!
===

This is an experiment to see if the `bazel-test-all` job sans all ic-os and system-test targets can run on Namespace x86_64 runners.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -336,7 +336,7 @@ jobs:
     name: Bazel Test All (Namespace)
     needs: [config]
     if: github.repository == 'dfinity/ic' # only run on public repo, not private since Namespace runners are not configured there, so these CI jobs get stuck otherwise.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-main.yml
```diff
@@ -332,6 +332,114 @@ jobs:
           # Clean up the output base for the next run
           sudo rm -rf /var/tmp/bazel-output
 
+  bazel-test-all-namespace:
```
cgundy, line 335: I'd have to look at it more closely, but I think it could make sense to create a reusable workflow just for bazel-test-all, then we could create logic for setting it up differently if it's kicked off by an internal vs external person (e.g. do or don't run system tests).

cgundy: Or have this depend on another job that checks if it's internal or external. Not that important for now. 😅

PR #7122, opened 2025-10-07
chore: Resize backup LV to 500GB

Bownairo:
The current size is 180GB which was too small leading to an incident. We now increase it to 500GB as calculated in NODE-1261


ic-os/components/upgrade/shared-resources/setup-shared-resources/setup-shared-backup.sh
```diff
@@ -7,5 +7,5 @@ blkid /dev/mapper/store-shared--backup >/dev/null || (
 )
 
 # TODO(NODE-1722): remove once every GuestOS has been upgraded after the LV resize
-e2fsck -pf /dev/mapper/store-shared--backup
+e2fsck -pf /dev/mapper/store-shared--backup || true
```
frankdavid, line 13: Perhaps mention that a return code of non-zero is not always bad (it's 1 or 2 when errors were corrected) so it's not a problem to throw away the return code.

frankdavid (approved):

PR #7123, opened 2025-10-07
feat: adding rate limits for node swapping mechanism

NikolaMilosa:
This PR adds two in memory rate limits for the node swapping mechanism:
* _Subnet limit_: on *1* subnet, at most *1* swap can happen within *2* hours. The goal is to allow controlled number of swaps made with this functionality. 
* _Provider limit_: on *1* subnet, at most *1* swap can happen within *12* hours made by the same *provider*. The goal is to disallow a single node provider from abusing the functionality and not allowing other node providers to swap nodes in a same subnet.

### DIFF
--- a/rs/registry/canister/src/mutations/do_swap_node_in_subnet_directly.rs
+++ b/rs/registry/canister/src/mutations/do_swap_node_in_subnet_directly.rs
@@ -6,6 +6,7 @@ use std::{
 
 use candid::CandidType;
 use ic_nervous_system_rate_limits::{InMemoryRateLimiter, RateLimiterConfig, Reservation};
+use ic_nervous_system_time_helpers::now_system_time;
 use ic_types::{NodeId, PrincipalId, SubnetId};
 use prost::Message;
 use serde::{Deserialize, Serialize};
@@ -46,7 +47,7 @@ impl SwappingRateLimiter {
         &mut self,
         subnet_id: SubnetId,
     ) -> Result<Reservation<SubnetId>, SwapError> {
-        self.reserve_subnet(subnet_id, SystemTime::now())
+        self.reserve_subnet(subnet_id, now_system_time())
     }
 
     fn reserve_subnet(
@@ -65,7 +66,7 @@ impl SwappingRateLimiter {
     }
 
     fn commit_subnet_now(&mut self, reservation: Reservation<SubnetId>) {
-        self.commit_subnet(reservation, SystemTime::now())
+        self.commit_subnet(reservation, now_system_time())
     }
 
     fn commit_subnet(&mut self, reservation: Reservation<SubnetId>, now: SystemTime) {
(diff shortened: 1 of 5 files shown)


rs/registry/canister/src/mutations/do_swap_node_in_subnet_directly.rs
```diff
@@ -14,6 +20,108 @@ use crate::{
...
+        // This would fail only if two in parallel calls were being processed
+        // which is not possible since swapping is an update call.
+        self.provider_subnet_limiter
+            .commit(now, reservation)
+            .unwrap()
+    }
+}
+
+thread_local! {
+    static SWAPPING_LIMITER: RefCell<SwappingRateLimiter> = RefCell::new(SwappingRateLimiter::new());
```
max-dfinity: I would make a separate RateLimiter for each one with its own type, and pass "now" in from do_swap_node_in_subnet_directly to swap_nodes_inner, instead of getting the time from inside the object.  That makes it easy to test edge cases and that the limits are applied in unit tests.

I can't see that there's an advantage to having one object for both.  

One of the APIs is just RateLimiter<SubnetId> and the other could be RateLimiter<(SubnetId, PrincipalId)> I think?  Then you wouldn't even need any of the indirection?

NikolaMilosa: I do agree that the reviewed implementation wasn't good enough as it didn't really solve anything. I think that the caller (the canister code) doesn't have to know that there are two consecutive `try_reserve` and `commit` calls that need to happen with different reservations. Furthermore if I moved the actual logic in the canister code testing becomes a bit more messy as the code evolves and I need to seed the whole registry properly just to test if rate limiting (the two limiters this code needs) work properly. 

Thus I will refactor the code but I still think that having a wrapper makes sense. Let me refactor it a bit and lets discuss if the refactoring makes a bit more sense or not and then we can decide if we should drop the wrapper or keep it.

rs/registry/canister/src/mutations/do_swap_node_in_subnet_directly.rs
```diff
@@ -42,9 +130,12 @@ impl Registry {
         let subnet_id = self.find_subnet_for_old_node(old_node_id)?;
         Self::swapping_allowed_on_subnet(subnet_id)?;
 
-        //TODO(DRE-553): Rate-limiting mechanism
+        let reservation =
+            SWAP_LIMITER.with_borrow_mut(|limiter| limiter.try_reserve(caller, subnet_id, now))?;
```
max-dfinity, line 134: Btw, you'll still want one high level test to make sure rate limits are actually respected in this endpoint, which you can do by just reserving all the available capacity, then attempting to make a call

NikolaMilosa: Sure I'll add one more! I think that the integration test added covers that, but adding one more unit test won't hurt.

PR #7124, opened 2025-10-08
refactor(IDX): combine CI cargo jobs

cgundy:
These two CI jobs are basically identical, just executing different commands, so we combine them into one, so that we can make updates to them more easily.

It also introduces the label `CI_RUN_CARGO_JOBS` in case we want to run the job without making any cargo file changes on our PR.

In the next PR I will move these jobs to gh-hosted runners.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -672,7 +672,7 @@ jobs:
         include:
           - name: Lint
           - name: Build
-    name: Cargo Clippy Linux ${{ matrix.name }}
+    name: Cargo  ${{ matrix.name }} Linux
     runs-on:
       labels: dind-small
     container: *container-setup
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -670,8 +670,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: lint
-          - name: build
+          - name: Lint
+          - name: Build
     name: Cargo Clippy Linux ${{ matrix.name }}
     runs-on:
       labels: dind-small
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-main.yml
```diff
@@ -667,8 +667,13 @@ jobs:
...
 
   cargo-clippy-linux:
-    name: Cargo Clippy Linux
-    runs-on: &dind-small-setup
+    strategy:
+      matrix:
+        include:
+          - name: Lint
+          - name: Build
+    name: Cargo Clippy Linux ${{ matrix.name }}
```
basvandijk:
```suggestion
    name: Cargo ${{ matrix.name }} Linux
```

.github/workflows/ci-main.yml
```diff
@@ -683,51 +688,26 @@ jobs:
               - "**/*.rs"
               - "**/*.toml"
               - "**/*.lock"
-      - name: Run Cargo Clippy Linux
-        id: cargo-clippy-linux
+      - name: Run Cargo Clippy Linux ${{ matrix.name }}
```
basvandijk:
```suggestion
      - name: Run Cargo ${{ matrix.name }} Linux 
```

.github/workflows/ci-main.yml
```diff
@@ -683,51 +688,26 @@ jobs:
...
+        id: cargo-clippy-linux-run
```
basvandijk: I would also remove the "clippy" from the id

basvandijk: Do we actually need an id?

cgundy: I guess we don't - I can remove it completely.

.github/workflows/ci-main.yml
```diff
@@ -667,8 +667,13 @@ jobs:
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
 
   cargo-clippy-linux:
```
basvandijk:
```suggestion
  cargo-linux:
```

basvandijk (approved):

PR #7129, opened 2025-10-08
refactor: Use SubnetConfig in ExecutionTestBuilder to group configuration options in tests

dsarlis:
This PR refactors the `ExecutionTestBuilder` to hold a `SubnetConfig` so that certain options can be overwritten in tests instead of passing explicitly the options.

One benefit of this is that we don't need to keep adding fields in `ExecutionTestBuilder` as more configuration options are added. We only need to expose a setter to overwrite the config value (similar to existing ones).

Another benefit is that it helps in a further refactoring of using `setup_execution` to create the execution services needed for tests instead of manually doing that to reduce code duplication. `setup_execution` requires a `subnet_config` to be passed so that it creates the services with the right options, so having it already in the `ExecutionTestBuilder` streamlines things (otherwise one would need to recreate the config before passing it). This change will follow in a subsequent PR.

### DIFF
--- a/rs/test_utilities/execution_environment/src/lib.rs
+++ b/rs/test_utilities/execution_environment/src/lib.rs
@@ -1808,6 +1808,7 @@ impl ExecutionTest {
 /// A builder for `ExecutionTest`.
 pub struct ExecutionTestBuilder {
     execution_config: Config,
+    subnet_config: SubnetConfig,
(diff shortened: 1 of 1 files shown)


michael-weigelt (approved):

dsarlis:
I found some more things that could be updated, so I'll convert back to draft.

PR #7130, opened 2025-10-08
fix: Only allow migration canister to call `rename_canister`

michael-weigelt:


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -6653,21 +6653,27 @@ fn test_environment_variables() {
 
 /// Creates and deploys a pair of universal canisters with the second canister being controlled by the first one
 /// in addition to both canisters being controlled by the anonymous principal.
+/// If the first state machine is has the NNS canister range, then the first canister has the id of
```
berestovskyy: nit: "is has" typo

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -6653,21 +6653,27 @@ fn test_environment_variables() {
...
-            UNIVERSAL_CANISTER_WASM.to_vec(),
-            vec![],
-            None,
-            INITIAL_CYCLES_BALANCE,
-        )
-        .unwrap();
+
+    // Skip 16 canister IDs so that if env1 is the NNS, canister_id1 will have the migration canister ID.
+    let mut canister_id1 = CanisterId::from_u64(0);
+    for _ in 0..18 {
```
berestovskyy, line 6667: Looks like a good hack 👍 We also might `create_canister_with_cycles` with a specified ID, but up to you...

michael-weigelt: I tried, but this is much simpler.

PR #7142, opened 2025-10-08
test(pocket-ic): test index and ledger with the same balance

maciejdfinity:
The broken test was compensating for the discrepancy in balance calculation between the ledger and index.
https://github.com/dfinity/ic/pull/6508 fixed this discrepancy so both index and ledger have the same balances. Since this PR is not on mainnet yet, we temporarily have to compare against both correct balance (for testing with master) and incorrect (for testing with mainnet).


berestovskyy (approved):

maciejdfinity:
> This won't work for the mainnet version of the canister (yet). I guess we'd need to relax the test for now to checking both cases (pass if either the original or new equality holds).

Thanks, good point, I changed the test!

PR #7283, opened 2025-10-16
feat(sns): Reinstall index canister if missing

jasonz-dfinity:
# Why

In a rare event where the ledger index canister has no WASM module, it's possible to reinstall it without losing any data, so we should.

# What

* Run a daily timer to check if SNS ledger index has a wasm module. Install it if there is none
* A PocketIC test to simulate the uninstalling of the index canister and the recovery


rs/nervous_system/integration_tests/tests/recover_from_deleted_index_canister.rs
```diff
@@ -0,0 +1,98 @@
...
+        .build();
+    let swap_parameters = create_service_nervous_system
+        .swap_parameters
+        .clone()
+        .unwrap();
+
+    // Use the tip-of-the-branch root wasm.
+    let wasm = ensure_sns_wasm_gzipped(build_root_sns_wasm());
+    let proposal_info = add_wasm_via_nns_proposal(&pocket_ic, wasm).await.unwrap();
+    assert_eq!(proposal_info.failure_reason, None);
```
jasonz-dfinity: I wonder if it should be using the tip-of-branch for everything

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+        return false;
+    };
+    let result = ic_nervous_system_clients::canister_status::canister_status::<CdkRuntime>(
+        CanisterIdRecord::from(index_canister_id),
+    )
+    .await;
+    let Ok(canister_status) = result else {
+        return false;
+    };
+    canister_status.module_hash.is_none()
```
jasonz-dfinity, line 79: Was trying to think if this can be true for anything other than the index getting uninstalled. All I can think of is when the initial installation getting stuck somehow or fails, but in that case it doesn't seem harmful to try to install it.

rs/sns/root/Cargo.toml
```diff
@@ -18,19 +18,23 @@ build-info = { workspace = true }
 candid = { workspace = true }
 comparable = { version = "0.5.1", features = ["derive"] }
 futures = { workspace = true }
+hex = { workspace = true }
 ic-base-types = { path = "../../types/base_types" }
 ic-canister-log = { path = "../../rust_canisters/canister_log" }
 ic-cdk = { workspace = true }
 ic-cdk-timers = { workspace = true }
 ic-http-types = { path = "../../../packages/ic-http-types" }
+ic-icrc1-index-ng = { path = "../../ledger_suite/icrc1/index-ng" }
```
max-dfinity, line 27: NCR.  It would be neat if they had an API crate here :)

max-dfinity (approved):

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+struct GetWasmResponse {
+    wasm: Option<SnsWasm>,
+}
+
+#[derive(candid::CandidType, candid::Deserialize)]
+struct SnsWasm {
+    wasm: Vec<u8>,
+    canister_type: i32,
+    proposal_id: Option<u64>,
+}
```
daniel-wong-dfinity-org, line 31: Why not get these definitions from the definitive source?

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl SnsRootCanister {
+    /// Reinstalls the index canister if it's needed (i.e., if it has no wasm module).
+    pub async fn install_index_canister_if_missing(self_ref: &'static LocalKey<RefCell<Self>>) {
```
daniel-wong-dfinity-org, line 35: Do as you see fit, since this code will be deleted in the not too distant future (maybe even next week).

Nevertheless, the following seems better to me:

```suggestion
/// Reinstalls the index canister if it's needed (i.e., if it has no wasm module).
pub async fn install_index_canister_if_missing(self_ref: &'static LocalKey<RefCell<Self>>) {
```

I know we do this (i.e. multiple impl) elsewhere, but I think it's a bad idea, is uncommon, and (here) not necessary.

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+async fn try_install_index_canister(
+    governance_canister_id: Option<PrincipalId>,
+    ledger_canister_id: Option<PrincipalId>,
+    index_canister_id: Option<PrincipalId>,
+) -> Result<(), String> {
+    let governance_canister_id = get_canister_id(governance_canister_id, "Governance")?;
+    let ledger_canister_id = get_canister_id(ledger_canister_id, "Ledger")?;
+    let index_canister_id = get_canister_id(index_canister_id, "Index")?;
```
daniel-wong-dfinity-org: Might make more sense for caller to unwrap. You would think that the main reason for Err being returned is a failure during actual installation, during actual "flight", not this pre-flight check. This is like asking, "ok, do we have a pilot?". Whereas, it "Do we have enough fuel?" is a more appropriate question at this point.

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+async fn should_install_index_canister(index_canister_id: Option<PrincipalId>) -> bool {
+    let Ok(index_canister_id) = get_canister_id(index_canister_id, "Index") else {
+        return false;
+    };
+    let result = ic_nervous_system_clients::canister_status::canister_status::<CdkRuntime>(
+        CanisterIdRecord::from(index_canister_id),
+    )
+    .await;
+    let Ok(canister_status) = result else {
+        return false;
```
daniel-wong-dfinity-org, line 77: Log?

rs/sns/root/src/install_index_canister_if_missing.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+fn get_canister_id(principal_id: Option<PrincipalId>, label: &str) -> Result<CanisterId, String> {
+    let principal_id = principal_id.ok_or(format!("No canister id for {label} provided"))?;
+    CanisterId::try_from_principal_id(principal_id)
+        .map_err(|e| format!("Error getting canister id for {label}: {e}"))
+}
+
+async fn should_install_index_canister(index_canister_id: Option<PrincipalId>) -> bool {
+    let Ok(index_canister_id) = get_canister_id(index_canister_id, "Index") else {
```
daniel-wong-dfinity-org, line 69: I think we want to really restrict this to the one known messed up canister. To do that, this would be replaced with something like

```
if index_canister_id != Some(ICVC_INDEX_CANISTER_ID) {
```

rs/nervous_system/integration_tests/tests/recover_from_deleted_index_canister.rs
```diff
@@ -0,0 +1,98 @@
...
+    create_service_nervous_system_builder::CreateServiceNervousSystemBuilder,
+    pocket_ic_helpers::{self, add_wasm_via_nns_proposal, nns, sns},
+};
+use ic_nns_test_utils::sns_wasm::{build_root_sns_wasm, ensure_sns_wasm_gzipped};
+use ic_sns_swap::pb::v1::Lifecycle;
+use ic_sns_wasm::pb::v1::SnsCanisterType;
+use std::time::Duration;
+
+#[tokio::test]
+async fn test_recover_from_deleted_index_canister() {
```
daniel-wong-dfinity-org, line 14: This seems fine, but wouldn't it be nice if we used golden state to "really see" that this works for ICVC?

daniel-wong-dfinity-org (approved):

PR #7642, opened 2025-11-12
refactor(ic-admin): Migrate ic-admin to ic-agent

jasonz-dfinity:
The `//rs/canister_client` is deprecated in favor of the public version ic-agent. This PR migrates ic-admin to the new library.


rs/registry/admin/bin/helpers.rs
```diff
@@ -88,27 +90,34 @@ pub(crate) fn parse_proposal_url(url: &Option<Url>) -> String {
...
+/// the `NeuronId` and `Identity` passed as argument.
+pub(crate) fn get_proposer_and_identity(
     proposer: Option<NeuronId>,
-    sender: Sender,
+    identity: Arc<dyn Identity>,
     use_test_neuron: bool,
-) -> (NeuronId, Sender) {
+) -> (NeuronId, Arc<dyn Identity>) {
     if use_test_neuron {
+        let pem = TEST_NEURON_1_OWNER_KEYPAIR.to_pem();
```
daniel-wong-dfinity-org, line 103: `pem` is pretty generic, and does not raise the level of abstraction. How about this:

```suggestion
        let identity = Arc::new(
            BasicIdentity::from_pem(TEST_NEURON_1_OWNER_KEYPAIR.to_pem())
                .expect(...)
        );
```

rs/registry/admin/bin/update_subnet.rs
```diff
@@ -1,13 +1,14 @@
...
     summary_from_string_or_file,
 };
 use crate::types::{ProposalMetadata, ProposalPayload, SubnetRecord};
 use crate::{ProposalTitle, SubnetDescriptor};
 use async_trait::async_trait;
 use clap::Parser;
 use ic_admin_derive::derive_common_proposal_fields;
-use ic_canister_client::{Agent, Sender};
+use ic_agent::identity::AnonymousIdentity;
+use ic_agent::{Agent, Identity};
```
daniel-wong-dfinity-org, line 11: I do not understand how it is possible that a file only has `use` changes (unless all the "added" things are just replacements for removed things). It seems to me that if you start introducing (for example) `AnonymousIdentity`, but make no changes other than `use`, then it is not possible that `AnonymousIdentity` is used anywhere, unless there are some files (like tests) that do something like `use super::*;`. But in that case, the test (or more generally, the submodule) should do `use ...AnonymousIdentity;` itself, rather than rely on containing module to indirectly supply it.

rs/registry/nns_data_provider/BUILD.bazel
daniel-wong-dfinity-org, line 1: This seems orthogonal to what the title describes? If so, and esp considering how large this PR is, I would spin this out. Normally, I don't feel that strongly about spinning out PRs, but this one is bigger than most.

(I have not reviewed changes under the nns_data_provider* dirs.)

rs/registry/admin/bin/main.rs
```diff
@@ -154,6 +156,7 @@ use registry_canister::mutations::{
     reroute_canister_ranges::RerouteCanisterRangesPayload,
 };
 use serde::{Deserialize, Serialize};
+use std::sync::Arc;
```
daniel-wong-dfinity-org, line 159: Please, put this back whence it came.

rs/registry/admin/bin/main.rs
```diff
@@ -3952,11 +3954,11 @@ async fn main() {
...
     }
 
-    let sender = if opts.secret_key_pem.is_some() || opts.use_hsm {
-        // Make sure to let the user know that we only actually use the sender
+    let identity: Arc<dyn Identity> = if opts.secret_key_pem.is_some() || opts.use_hsm {
+        // Make sure to let the user know that we only actually use the identity
         // in methods that go through the NNS handlers and not for other methods.
         //
-        // TODO(NNS1-486): Remove ic-admin command whitelist for sender
+        // TODO(NNS1-486): Remove ic-admin command whitelist for identity
```
daniel-wong-dfinity-org, line 3961: Do you understand what this TODO is about? I read the Jira ticket, but failed to actually comprehend it. I guess the question is, what bad thing would happen if we got rid of this check?

rs/registry/admin/bin/main.rs
```diff
@@ -4018,33 +4020,69 @@ async fn main() {
...
+                .expect("Failed to create HardwareIdentity from HSM"),
             )
         } else {
-            Sender::Anonymous
+            Arc::new(AnonymousIdentity)
         }
     } else {
-        Sender::Anonymous
+        Arc::new(AnonymousIdentity)
     };
```
daniel-wong-dfinity-org, line 4076: Ofc, this issue already existed, so do not fix in this PR, but I just have to point out how this is totally backwards. Look at how far away these are from the conditions that trigger them. It's kind of insane. It should be more like this:

```rust
let identity = new_identity_from_opts_or_panic(&opts);

...

fn new_identity_from_opts_or_panic(opts: &Opts) -> impl Identity {
    // TODO(NNS1-486): ...
    if identity_supplied(opts) && !support_identity(&opt.subcmd) {
        panic!("The {} subcommand does not support specifying an identity.");
    }

    if let Some(secret_key_pem) = opts.secret_key_pem.as_ref() {
        return Arc::new(new_pem_identity_or_panic(secret_key_pem));
    }

    if &opts.use_hsm {
        return Arc::new(new_hsm_identity_or_panic(
            &opts.hsm_slot,
            ...
        ));
    }

    Arc::new(AnonymousIdentity)
}
```

When you have a coherent chunk of work, spin out! (E.g. `new_identity_from_opts_or_panic`.) When you have a specialized version of coherent work, spin it out! (E.g. `new_hsm_identity_or_panic`.)

rs/registry/admin/bin/main.rs
```diff
@@ -4018,33 +4020,69 @@ async fn main() {
...
+            // Try to parse as Secp256k1 with RFC 5915 format (EC PRIVATE KEY)
+            else if let Ok(identity) = Secp256k1Identity::from_pem(contents.as_bytes()) {
+                Arc::new(identity)
+            } else {
+                panic!(
+                    "Failed to parse PEM file. Tried Ed25519 (BasicIdentity), Secp256k1 RFC5915, and Secp256k1 PKCS#8 formats."
+                );
+            }
         } else if opts.use_hsm {
-            make_hsm_sender(
```
daniel-wong-dfinity-org, line 4024: This (i.e. spinning out) was a good thing. Please, bring it back.

rs/registry/admin/bin/main.rs
```diff
@@ -4018,33 +4020,69 @@ async fn main() {
...
-                &opts.hsm_slot.expect(
-                    "HSM slot must also be provided for --use-hsm; use --hsm-slot or see --help.",
-                ),
-                &opts.hsm_key_id.expect(
-                    "HSM key ID must also be provided for --use-hsm; use --key-id or see --help.",
-                ),
-                &opts.hsm_pin.expect(
-                    "HSM pin must also be provided for --use-hsm; use --pin or see --help.",
```
daniel-wong-dfinity-org, line 4032: Unwrapping could be delegated to make_hsm_identity.

rs/registry/admin/bin/main.rs
```diff
@@ -4298,52 +4350,53 @@ async fn main() {
...
             cmd.apply_defaults_for_unset_fields();
-            let (proposer, sender) = cmd.proposer_and_sender(sender);
+            let (proposer, identity) = cmd.proposer_and_identity(identity.clone());
+            let ic_agent = make_agent_from_identity(
+                reachable_nns_urls.clone(),
+                identity,
+                opts.verify_nns_responses,
+                opts.nns_public_key_pem_file.clone(),
+            )
+            .await;
```
daniel-wong-dfinity-org, line 4378: An issue that already existed (and therefore, should not be addressed in this PR): This code is HYPER DAIPER SUPER DUPER repetitive...

rs/registry/admin/bin/main.rs
```diff
@@ -5984,14 +6112,13 @@ async fn get_subnet_pk(registry: &RegistryCanister, subnet_id: SubnetId) -> Publ
...
-        )),
-        None,
-    );
+async fn get_api_boundary_node_ids(
+    nns_url: Vec<Url>,
+    verify_nns_responses: bool,
+    nns_public_key_pem_file: Option<PathBuf>,
+) -> Vec<String> {
+    let registry_client =
+        make_registry_client(nns_url, verify_nns_responses, nns_public_key_pem_file).await;
```
daniel-wong-dfinity-org, line 6121: Do as you see fit.

Maybe would make more sense for caller to supply this.

rs/registry/admin/bin/main.rs
```diff
@@ -6012,14 +6139,13 @@ fn get_api_boundary_node_ids(nns_url: Vec<Url>) -> Vec<String> {
...
-        )),
-        None,
-    );
+async fn print_routing_table(
+    nns_urls: Vec<Url>,
+    verify_nns_responses: bool,
+    nns_public_key_pem_file: Option<PathBuf>,
+) -> Vec<(SubnetId, CanisterIdRange)> {
+    let registry_client =
+        make_registry_client(nns_urls, verify_nns_responses, nns_public_key_pem_file).await;
```
daniel-wong-dfinity-org, line 6148: Ditto.

rs/registry/admin/bin/main.rs
```diff
@@ -5984,14 +6112,13 @@ async fn get_subnet_pk(registry: &RegistryCanister, subnet_id: SubnetId) -> Publ
...
-fn get_api_boundary_node_ids(nns_url: Vec<Url>) -> Vec<String> {
-    let registry_client = RegistryClientImpl::new(
-        Arc::new(NnsDataProvider::new(
-            tokio::runtime::Handle::current(),
-            nns_url,
-        )),
-        None,
-    );
+async fn get_api_boundary_node_ids(
+    nns_url: Vec<Url>,
```
daniel-wong-dfinity-org, line 6116:
```suggestion
    nns_urls: Vec<Url>,
```

rs/registry/admin/bin/main.rs
```diff
@@ -6179,7 +6305,10 @@ fn get_root_subnet_pub_key(
 
 /// Fetch registry records from the given `nns_urls`, and update the local
```
daniel-wong-dfinity-org, line 6306: Comment `nns_urls` is now obsolete.

rs/registry/admin/bin/main.rs
```diff
@@ -6250,17 +6378,44 @@ async fn update_registry_local_store(nns_urls: Vec<Url>, cmd: UpdateRegistryLoca
...
+                    .expect("Failed to create BasicIdentity from test keypair"),
+            )
+        }
+        Some(4) => {
+            let pem = TEST_USER4_KEYPAIR.to_pem();
+            Arc::new(
+                BasicIdentity::from_pem(pem.as_bytes())
+                    .expect("Failed to create BasicIdentity from test keypair"),
+            )
+        }
```
daniel-wong-dfinity-org, line 6416: This is hyper repetitive. One way to fix:

```suggestion
    let Some(test_identity) = test_identity else {
        return current_identity;
    };

    let keypair = match test_identity {
        1 => TEST_USER1_KEYPAIR,
        etc...
        
        _ => panic!(...),
    };
    
    Arc::new(... keypair ...)
```

rs/registry/admin/bin/main.rs
```diff
@@ -6339,19 +6483,20 @@ async fn swap_node_in_subnet_directly(
...
-        .await
-        .unwrap()
-    {
-        Some(_) => println!("Nodes swapped successfully."),
-        None => panic!("No response was received from swap_node_in_subnet_directly"),
+    // Convert CanisterId to candid::Principal for ic-agent
+    let principal = candid::Principal::try_from(REGISTRY_CANISTER_ID.get().as_slice())
+        .expect("Failed to convert CanisterId to Principal");
+
+    // Use ic-agent's update API
```
daniel-wong-dfinity-org, line 6490: This does not seem to add value.

rs/registry/admin/bin/main.rs
```diff
@@ -6397,6 +6542,51 @@ fn is_mainnet(url: &Url) -> bool {
...
+    })?;
+
+    // Step 3: Go up to the OpenSC root directory (from bin to root)
+    let opensc_root = real_path
+        .parent() // Remove "pkcs11-tool"
+        .and_then(|p| p.parent()) // Remove "bin"
+        .ok_or_else(|| "Failed to find OpenSC root directory".to_string())?;
+
+    // Step 4: Build the library path
+    let lib_path = opensc_root.join("lib").join("opensc-pkcs11.so");
```
daniel-wong-dfinity-org, line 6580: This seems fragile, but I don't know of a better way. Even Step 5 below 👇 seems to acknowledge this. Dynamic linking seems like something that your OS is supposed to take care of for you, but then again, I have often seen it struggle to do this basic task...

Anyway, since I have no better ideas, I guess there is nothing to do here, unless you know the right way to do this.

rs/registry/admin/bin/main.rs
```diff
@@ -6449,21 +6659,24 @@ impl NnsCanisterClient {
         let mut ids_to_try = vec![self.handler_id];
         ids_to_try.extend(ic_nns_constants::ALL_NNS_CANISTER_IDS.iter().cloned());
```
daniel-wong-dfinity-org, line 6660: This just seems like spam. If you want to call a method, you should have to specify which canister it is on, and not just try them all...

Ofc, this issue predates this PR -> do not try to resolve this in this PR.

rs/registry/admin/bin/main.rs
```diff
@@ -6449,21 +6659,24 @@ impl NnsCanisterClient {
...
 
+        let method = msg.to_string();
+
         for canister_id in ids_to_try {
+            // Convert CanisterId to candid::Principal
+            let principal = candid::Principal::try_from(canister_id.get().as_slice())
+                .map_err(|e| format!("Failed to convert CanisterId to Principal: {:?}", e))?;
```
daniel-wong-dfinity-org, line 6667:
```suggestion
        for callee in ids_to_try {
            let callee = Principal::from(PrincipalId::from(callee))
```

There are a few things going on in my suggestion, but the one I want to call out is converting from CanisterId to Principal. Doing it the way I'm suggesting avoids the need to handle Err, which is nice.

Such conversions should be done like this elsewhere in this PR (and everywhere).

rs/registry/admin/bin/main.rs
```diff
@@ -6449,21 +6659,24 @@ impl NnsCanisterClient {
         let mut ids_to_try = vec![self.handler_id];
         ids_to_try.extend(ic_nns_constants::ALL_NNS_CANISTER_IDS.iter().cloned());
 
+        let method = msg.to_string();
```
daniel-wong-dfinity-org, line 6662: Argument should be named `method` too. `msg` is just wrong (and not a real word). Probably, it's type could (and should) be `&str`.

rs/registry/admin/bin/main.rs
```diff
@@ -6449,21 +6659,24 @@ impl NnsCanisterClient {
...
-                )
+                .update(&principal, &method)
+                .with_arg(arguments.clone())
+                .call_and_wait()
                 .await;
 
             match result {
-                Ok(result) => return Ok(result),
-                Err(error_string) => {
+                Ok(response) => return Ok(Some(response)),
```
daniel-wong-dfinity-org, line 6677: IIUC, it is not possible for this method to return `Ok(None)`. Ideally, in some future PR (maybe never), we would change the return type to just `Result<Vec<u8>, ...>`...

pierugo-dfinity (approved):
`cup_explorer` changes LGTM

PR #7784, opened 2025-11-24
refactor(nns): Remove static mut as much as possible

jasonz-dfinity:
# Why

The previous paradigm for access the global state within NNS Governance can lead to subtle problems if a reference to the global state is retained across an await boundary (therefore can be accessed in a different message).

# What

- Change all `governance().f(..)` to `with_governance(|g| g.f(..))`
- Change most `governance_mut().f(..)` to `with_governance_mut(|g| g.f(..))`
- Change the rest of `governance_mut.f(..)` to `legacy_governance_mut.f(..)` (typically when `f` is async)


rs/nns/governance/canister/canister.rs
```diff
@@ -78,7 +80,7 @@ const SPAWN_NEURONS_INTERVAL: Duration = Duration::from_secs(60);
 fn schedule_spawn_neurons() {
     ic_cdk_timers::set_timer_interval(SPAWN_NEURONS_INTERVAL, || {
         ic_cdk::futures::spawn_017_compat(async {
-            governance_mut().maybe_spawn_neurons().await;
+            legacy_governance_mut().maybe_spawn_neurons().await;
```
daniel-wong-dfinity-org: I guess we can gradually get rid of these.

rs/nns/governance/canister/canister.rs
```diff
@@ -289,58 +295,70 @@ async fn manage_neuron(_manage_neuron: ManageNeuronRequest) -> ManageNeuronRespo
...
     debug_log("update_neuron");
-    governance_mut()
-        .update_neuron(neuron)
-        .err()
-        .map(GovernanceError::from)
+    with_governance_mut(|governance| {
+        governance
+            .update_neuron(neuron)
+            .err()
+            .map(GovernanceError::from)
```
daniel-wong-dfinity-org: Let's do as much as possible outside of with.

Ditto elsewhere.

daniel-wong-dfinity-org (approved):
😎

rs/nns/governance/canister/canister.rs
jasonz-dfinity: Agreed — left these as `legacy_governance_mut()` for the async sites, since `with_governance_mut` cannot hold the RefCell borrow across an await. Will migrate them gradually once those call paths are restructured.

rs/nns/governance/canister/canister.rs
jasonz-dfinity: Done — swept canister.rs to lift protobuf/API conversions, error mapping, and response transformations out of the with_governance / with_governance_mut closures (see 70b48d3a). Also applied the same to canister_state::execute_nns_function during the merge-master resolution.

PR #8217, opened 2026-01-05
feat(ICRC-Ledger): DEFI-1687: implement ICRC-107 fee collector

maciejdfinity:
With this change the ICRC ledger now implements the ICRC-107 fee collector standard as described in https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-107/ICRC-107.md. The legacy fee collector is no longer supported by the ledger. This results in the following notable changes:
- If a fee collector is present, approval fees are also collected.
- If an existing ledger has a legacy fee collector configured, the legacy fee collector will be converted to the ICRC-107 fee collector during the upgrade to this or later version.

The fee collector can be configured using the existing `fee_collector_account` init argument and `change_fee_collector` upgrade argument.

Since the ICRC-107 fee collector change is not backwards compatible, the ledger upgraded to this version cannot be downgraded to an earlier version.


rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -848,7 +848,7 @@ mod withdraw_erc20 {
                     .assert_has_unique_events_in_order(&[EventPayload::ReimbursedErc20Withdrawal {
                         withdrawal_id: cketh_block_index.clone(),
                         burn_in_block: ckerc20_block_index.clone(),
-                        reimbursed_in_block: Nat::from(3_u8),
+                        reimbursed_in_block: Nat::from(4_u8),
```
maciejdfinity, line 851: If the ICRC ledger is installed with a 107 fee collector, the first block is the fee collector setting block and therefore all block indexes that are tested need to be increased by 1.

rs/tests/cross_chain/ic_xc_ledger_suite_orchestrator_test.rs
```diff
@@ -162,7 +162,7 @@ fn ic_xc_ledger_suite_orchestrator_test(env: TestEnv) {
         assert_eq!(
             index_status,
             ic_icrc1_index_ng::Status {
-                num_blocks_synced: Nat::from(0_u8)
+                num_blocks_synced: Nat::from(1_u8)
```
maciejdfinity, line 165: If the ICRC ledger is installed with a 107 fee collector, the first block is the fee collector setting block and therefore even a ledger with no transactions will contain one block.

rs/ledger_suite/icrc1/ledger/src/lib.rs
```diff
@@ -735,6 +735,31 @@ impl Ledger {
...
+            mthd: Some(BTYPE_107_LEDGER_SET.to_string()),
+        };
+        let tx = Transaction {
+            operation: op,
+            created_at_time: None,
+            memo: None,
+        };
+        let now = TimeStamp::from_nanos_since_unix_epoch(ic_cdk::api::time());
+        let block = Block::from_transaction(self.blockchain.last_hash, tx, now, Tokens::ZERO);
+        if let Err(e) = self.blockchain.add_block(block) {
```
maciejdfinity: Instead of creating the block and adding it to the blockchain, we could also call `apply_transaction`. This function however also purges old transactions and prunes expired approvals, which we might not want to do during the upgrade.

rs/ledger_suite/icrc1/ledger/canbench_results/canbench_u256.yml
```diff
@@ -2,70 +2,70 @@ benches:
   bench_icrc1_transfers:
     total:
       calls: 1
-      instructions: 54416296904
-      heap_increase: 264
+      instructions: 54937060262
```
maciejdfinity: u64 diffs:
Only significant changes:
| status | name                                    | calls |    ins |  ins Δ% | HI |  HI Δ% | SMI |  SMI Δ% |
|--------|-----------------------------------------|-------|--------|---------|----|--------|-----|---------|
|  +/-   | bench_icrc1_transfers::icrc2_approve    |     1 | 19.65B |  +6.57% | 24 | -4.00% | 128 |   0.00% |
|   -    | bench_icrc1_transfers::icrc3_get_blocks |     1 |  7.84M |  -7.01% |  0 |  0.00% |   0 |   0.00% |

u256 diffs:
Only significant changes:
| status | name                                       | calls |    ins |  ins Δ% | HI |   HI Δ% | SMI |  SMI Δ% |
|--------|--------------------------------------------|-------|--------|---------|----|---------|-----|---------|
|  +/-   | bench_icrc1_transfers::icrc2_approve       |     1 | 20.55B |  +6.23% | 25 | -13.79% | 128 |   0.00% |
|   +    | bench_icrc1_transfers::icrc2_transfer_from |     1 | 21.03B |  -1.86% |  4 | +33.33% |   0 |   0.00% |
|  +/-   | bench_icrc1_transfers::icrc1_transfer      |     1 | 12.64B |  -2.20% | 34 |  +6.25% |   0 |   0.00% |
|   -    | bench_icrc1_transfers::icrc3_get_blocks    |     1 |  8.22M |  -7.89% |  0 |   0.00% |   0 |   0.00% |

maciejdfinity: The u64 diffs make some sense, `icrc2_approve` now also collects fees, so it uses a bit more instructions. The created blocks don't contain the fee collector or fee collector block index, which results in less heap usage for `icrc2_approve` and less instructions for decoding the blocks in case of `icrc3_get_blocks`.

maciejdfinity: For u256, the instruction diffs are similar for `icrc2_approve` and `icrc3_get_blocks`. We also see improvements for `icrc1_transfer` and `icrc2_transfer_from` (in case of u64 they were significant but below threshold). The latter might be caused by the fact that the fee collector is stored directly as `Option<Account>` and previously it was stored in `Option<FeeCollector<Account>>`. The heap diffs are less consistent, though.

MODULE.bazel
```diff
@@ -1102,11 +1102,13 @@ canisters(
         "ck_btc_ledger": "ic-icrc1-ledger.wasm.gz",
         "ck_btc_ledger_v1": "ic-icrc1-ledger.wasm.gz",
         "ck_btc_ledger_v3": "ic-icrc1-ledger.wasm.gz",
+        "ck_btc_ledger_v5": "ic-icrc1-ledger.wasm.gz",
```
mbjorkqvist, line 1148: Would it make sense to rename this to something more descriptive, and/or perhaps add a comment somewhere (I think comments don't work in `mainnet-canister-revisions.json`, but here would be ok?)? Especially since the `v5` doesn't correspond to the `LEDGER_VERSION`, so it can be confusing.

mainnet-canister-revisions.json
```diff
@@ -27,6 +27,10 @@
     "rev": "2190613d3b5bcd9b74c382b22d151580b8ac271a",
     "sha256": "25071c2c55ad4571293e00d8e277f442aec7aed88109743ac52df3125209ff45"
   },
+  "ck_btc_ledger_v5": {
+    "rev": "512cf412f33d430b79f42330518166d14fc6884e",
+    "sha256": "901bc548f901145bd15a1156487eed703705794ad6a23787eaa04b1c7bbdcf48"
+  },
```
mbjorkqvist, line 33: This points to [ledger-suite-icrc-2025-04-14](https://github.com/dfinity/ic/releases/tag/ledger-suite-icrc-2025-04-14). Is there a particular reason for this specific version, or could we use [ledger-suite-icrc-2025-10-27](https://github.com/dfinity/ic/releases/tag/ledger-suite-icrc-2025-10-27), which is the minimum version we point to from the currently most recent version [ledger-suite-icrc-2026-02-02](https://github.com/dfinity/ic/releases/tag/ledger-suite-icrc-2026-02-02)?

rs/ledger_suite/icrc1/index-ng/tests/tests.rs
```diff
@@ -1323,20 +1410,56 @@ fn test_fee_collector() {
...
+
+#[cfg(not(feature = "icrc3_disabled"))]
+#[test]
+fn test_fee_collector_ranges_107() {
+    test_fee_collector_ranges(false);
 }
 
+// This test uses the test ledger to test edge cases such as
+// specifying the legacy fee collector after the 107 fee collector block
+// was genereated, which could not be tested with the prod ledger.
```
mbjorkqvist:
```suggestion
// was generated, which could not be tested with the prod ledger.
```

rs/ledger_suite/icrc1/index-ng/tests/tests.rs
mbjorkqvist, line 1: Would it make sense to move all the fee collector-related tests to a separate file? `rs/ledger_suite/icrc1/index-ng/tests/tests.rs` is already over 2000 lines.

Also, would it make sense to add tests for the following (unless they already exist and I missed them):

- Trying to set the fee collector account to the minting account, the anonymous account, and an invalid account (e.g., an account with an invalid principal). For the init, upgrade, and endpoint fee collector setting paths. IIUC, there's a check for the minting account in the init and upgrade paths, but not for the endpoint?

rs/ledger_suite/tests/sm-tests/src/icrc_106.rs
```diff
@@ -149,16 +149,19 @@ pub fn test_upgrade_downgrade_with_mainnet_ledger<T, U>(
...
+    // // Downgrade the ledger to the mainnet version that does not support ICRC-106
+    // env.upgrade_canister(canister_id, mainnet_ledger_wasm, encoded_empty_upgrade_args)
+    //     .expect("should successfully downgrade ledger canister");
+    // assert_index_not_set(&env, canister_id, false);
+
+    // // Upgrade to a ledger version that supports ICRC-106, but do not set the index principal
+    // let encoded_empty_upgrade_args = Encode!(&encode_empty_upgrade_args()).unwrap();
+    // env.upgrade_canister(canister_id, ledger_wasm, encoded_empty_upgrade_args)
+    //     .expect("should successfully upgrade ledger canister");
+    // assert_index_not_set(&env, canister_id, true);
```
mbjorkqvist, line 164: Would it help to have a ticket in the ICRC-107 epic to remember to uncomment this, and the other snippet in `rs/ledger_suite/icrc1/tests/upgrade_downgrade.rs`, once this has been released and deployed to mainnet? Or will the tests fail anyway once that happens, and uncommenting cannot be forgotten?

mbjorkqvist (commented):
Thanks @maciejdfinity! I have a few comments and questions, but overall this looks very good!

mainnet-canister-revisions.json
maciejdfinity: No reason, I just used the same version as current mainnet ledger :) Changed!

MODULE.bazel
maciejdfinity: Good point! Maybe we could change it to correspond to the ledger versions, i.e.:
v1 -> no_mem_mgr
v3 -> v1
v5 -> v3
WDYT?

rs/ledger_suite/tests/sm-tests/src/icrc_106.rs
maciejdfinity: It could be forgotten, I created https://dfinity.atlassian.net/browse/DEFI-2657

rs/ledger_suite/icrc1/index-ng/tests/tests.rs
maciejdfinity: Good catch, I did not check for minting or anonymous account in the endpoint and init/upgrade did not check for the anonymous account. I added `test_fee_collector_107_minting_account` and `test_fee_collector_107_anonymous` which cover the 6 cases: (minter, anonymous)x(endpoint, init, upgrade).
As for the invalid Account - since this is a candid endpoint, I don't think we have to check that - we always receive a valid Account, invalid Accounts should be rejected by cdk?

I also moved the index tests to a separate file.

PR #8282, opened 2026-01-09
feat(ICP_Rosetta): DEFI-2588: Remove DeprecatedTestnet variant from Environment

mbjorkqvist:
Remove `DeprecatedTestnet` variant from `Environment` configuration enumeration for ICP Rosetta, and make `Test` the default value. When started without providing `--mainnet` or `--environment production`, Rosetta will connect to the `TESTICP` canister on mainnet. A timeout in tests waiting for the Rosetta port file was also introduced, so that in case Rosetta crashes, the tests do not wait forever in an infinite loop.


rs/rosetta-api/icp/src/main.rs
```diff
@@ -80,21 +78,20 @@ struct NetworkConfig {
...
 
         let url_str = match &config.ic_url {
             Some(url_str) => url_str.as_str(),
-            None => match environment {
-                Environment::DeprecatedTestnet => DEPRECATED_TESTNET_URL,
-                Environment::Production | Environment::Test => MAINNET_URL,
-            },
+            None => MAINNET_URL,
         };
+        let is_mainnet_url = url_str == MAINNET_URL;
```
maciejdfinity: Isn't it better to compare the `Url` structs and not the `str`? What if someone add an extra `/` at the end?

mbjorkqvist: Good idea - done!

PR #8464, opened 2026-01-22
feat: [DSM-94] Asynchronous hashing

michael-weigelt:
This PR introduces a hashing thread in the state manager. Instead of hashing the state during `commit_and_certify`, the state hash is calculated asynchonously, while the batch processor thread can continue and work on state cloning and processing the execution layer logic. The next time `commit_and_certify` is called, the previous state hash is awaited. 

Hashing is skipped during catch-up, if consensus delivers state hashes via certifications. 

Comparing this optimization with master using `rejoin_test_long_rounds` (only one run each, but compared in the steady-state):
```
                       base        async
finalization_rate:     0.77         0.90
dsm_overhead:          0.48 s       0.38 s
commit_and_certify:     100 ms        15 ms
hash_tree:           90-100 ms     80-90 ms
```

The improvements in `finalization_rate` and `dsm_overhead` are straightforward. 
The improvement in `commit_and_certify` is due to moving hashing to a background thread. This time is not necessarily all saved, instead it is spent while other work is ongoing. The overlap seems to be large though, which is good, and is evidenced by the `dsm_overhead` improvement. 
Curiously, the `dsm_overhead` improvement is even larger than what we save in `commit_and_certify`. And, also curiously, calculating the hash tree is also faster now, even though that happens now concurrently with the execution threads, due to which one would expect some deterioration rather than improvements... But I am not complaining. 

Thanks @schneiderstefan and @mraszyk for helping with this.


rs/state_manager/tests/state_manager.rs
```diff
@@ -9509,21 +9457,3 @@ fn commit_and_certify_reuses_certification() {
         );
     });
 }
-
-#[test]
-#[should_panic(
-    expected = "Committed state @1 with hash CryptoHash(0x4e2d174de5daaeb4622d8f5e426ee09274f7ec4fb01d62fb9a3d36ae50961353) which is different from previously computed or delivered hash CryptoHash(0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a)"
-)]
-fn commit_and_certify_panic_on_delivered_fake_certification() {
```
michael-weigelt, line 9517: c&c now uses the previous state hash in `certifications` if present, it does not recalculate it.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9142,112 +9190,6 @@ fn deliver_state_certification_for_future_heights() {
...
-        // the state at height 1 is not hashed in `take_tip` since certification was delivered
-        assert_eq!(tip_hash_count(metrics), 1);
-        let (height, _state) = sm.take_tip();
-        assert_eq!(tip_hash_count(metrics), 1);
-        assert_eq!(height, opt_height);
-    });
-}
-
-#[test]
-fn take_tip_hashes_with_optimization() {
```
michael-weigelt, line 9218: `take_tip` does not hash any more.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9142,112 +9190,6 @@ fn deliver_state_certification_for_future_heights() {
...
-        // the state at height 1 is not hashed in `take_tip` since certification metadata are computed
-        assert_eq!(tip_hash_count(metrics), 1);
-        let (height, _state) = sm.take_tip();
-        assert_eq!(tip_hash_count(metrics), 1);
-        assert_eq!(height, no_opt_height);
-    });
-}
-
-#[test]
-fn take_tip_does_not_hash_with_optimization() {
```
michael-weigelt, line 9176: `take_tip` does not hash any more.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9142,112 +9190,6 @@ fn deliver_state_certification_for_future_heights() {
     });
 }
 
-#[test]
-fn take_tip_does_not_hash_without_optimization() {
```
michael-weigelt, line 9146: `take_tip` does not hash any more.

rs/state_manager/src/lib.rs
```diff
@@ -188,7 +188,6 @@ pub struct StateManagerMetrics {
     latest_hash_tree_max_index: IntGauge,
     fast_forward_height: IntGauge,
     no_state_clone_count: IntCounter,
-    tip_hash_count: IntCounter,
```
mraszyk, line 191: This metric tracks the number of states for which consensus did not deliver their hash while catching-up so we should keep that metric.

rs/state_manager/src/lib.rs
```diff
@@ -2263,6 +2241,7 @@ impl StateManagerImpl {
         // as decisions to retain a checkpoint or an in-memory state are made independently.
         let inmemory_heights_to_keep = std::iter::once(latest_certified_height)
             .chain(extra_inmemory_heights_to_keep.iter().copied())
+            .chain(std::iter::once(tip_height))
```
mraszyk, line 2244: Why is this needed now?

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,63 @@ impl StateManager for StateManagerImpl {
...
+                // must have an entry at height-1, so we can unwrap.
+                let states = self.states.read();
+                if let Some(cert_md) = states.certifications_metadata.get(&prev_height) {
+                    cert_md.certified_state_hash.clone()
+                } else {
+                    error!(
+                        self.log,
+                        "Previous state hash was not available after awaiting the hash thread. This is a bug."
+                    );
+                    // Continue nevertheless, this (impossible situation) should self-heal.
```
mraszyk: How would this self-heal? We are also skipping many steps afterwards.

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,63 @@ impl StateManager for StateManagerImpl {
...
+                    return;
+                }
+            }
+        };
+        // Write the previous state hash to the state.
+        state
+            .metadata
+            .prev_state_hash
+            .replace(CryptoHashOfPartialState::from(prev_state_hash));
```
mraszyk: Out of curiosity: why do you use `replace` instead of `=` (assignment)?

rs/state_manager/src/lib.rs
```diff
@@ -3362,16 +3353,21 @@ impl StateManager for StateManagerImpl {
 
         // If the node is catching up (`height.get() < fast_forward_height`)
         // and this is not a checkpoint height (`matches!(scope, CertificationScope::Metadata)`),
+        // and the state hash @ height is present in states.certifications (via consensus),
```
mraszyk, line 3352: Even if the hash is not available, it could make sense to at least skip cloning. However, this means that if consensus does not deliver the hash, we'd have to compute it synchronously. The trade-off is as follows:
- clone + min(execute_round, hash) -- the current approach from this PR
- execute_round + hash -- the alternative

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3527,147 @@ impl StateManager for StateManagerImpl {
...
+                                    certification_metadata.certified_state_hash
+                                );
+                            }
+
+                            // It's possible that we already computed this state before. We
+                            // validate that hashes agree to spot bugs causing non-determinism as
+                            // early as possible. 
+                            let hash = &certification_metadata.certified_state_hash;
+                            let mut states = states.write();
+                            if let Some(prev_metadata) = states.certifications_metadata.get(&height) {
```
mraszyk, line 3614: We should also compare the hash in `state.certifications` if available.

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3527,147 @@ impl StateManager for StateManagerImpl {
...
+                            states,
+                            latest_state_height,
+                            height,
+                            latest_height_update_time,
+                            scope,
+                            state_layout,
+                        } => {
+                            // If the hash@height is already present via the consensus thread, we can skip calculating it.
+                            let states_read = states.read();
+                            if states_read.certifications.contains_key(&height) {
```
mraszyk: We should not skip hashing (which also produces the state tree) unless the conditions for the catch-up optimization were satisfied. Otherwise, we might have no certified state for a long time.

But if there is a certification in `certifications`, we should copy it to `certification_metadata` right away (so that consensus does not have to deliver it twice).

rs/state_manager/src/lib.rs
```diff
@@ -2824,15 +2761,13 @@ impl StateManager for StateManagerImpl {
...
 
-        let mut new_tip = initialize_tip(
+        let new_tip = initialize_tip(
             &self.log,
             &self.tip_channel,
             &target_snapshot,
             checkpoint_layout,
         );
 
-        new_tip.metadata.prev_state_hash = Some(target_hash);
```
mraszyk, line 2834: Is `prev_state_hash` still set correctly in case of state sync?

rs/state_manager/src/lib.rs
```diff
@@ -3411,17 +3407,18 @@ impl StateManager for StateManagerImpl {
...
-        let mut certification_metadata =
-            Self::compute_certification_metadata(&state, height, &self.metrics, &self.log)
-                .unwrap_or_else(|err| fatal!(self.log, "Failed to compute hash tree: {:?}", err));
-
-        if scope == CertificationScope::Full {
-            info!(
-                self.log,
-                "Certification hash for height {}: {:?}",
+        // Kick off hashing of the new state if necessary, i.e., if not available from consensus via states.certifications.
+        if !hash_at_height_available {
```
mraszyk: At this point, we should always hash (and produce the state tree for) the state (since the conditions for the catch-up optimization are not satisfied).

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,63 @@ impl StateManager for StateManagerImpl {
...
+            self.hash_channel
+                .send(HashRequest::Wait { sender })
+                .expect("Failed to send `Wait` to hash channel");
+            recv.recv().expect("Failed to wait for hash channel");
+            // At prev_height 0, we don't have a hash yet, so we have to compute it.
+            if prev_height.get() == 0 {
+                let states = self.states.read();
+                let initial_state = &states
+                    .snapshots
+                    .front()
```
mraszyk, line 3285: We could also add an assertion that `initial_state` is indeed the state at height 0.

mraszyk (commented):
Only reviewed the production code for now: please let me know if you'd like to go over my feedback in a meeting.

rs/state_manager/src/lib.rs
michael-weigelt: The height that used to be passed from `take_tip` to `commit_and_certify` is now given implicitly by `tip_height`. We don't want to remove the state that c&c might be still using. (This method can be called after state sync at an unpredictable time)

rs/state_manager/src/lib.rs
michael-weigelt: No particular reason. I'll change it.

rs/state_manager/src/lib.rs
michael-weigelt: Yeah we are aware of this other option, but for simplicity I am first going with this approach.

rs/state_manager/src/lib.rs
michael-weigelt: This is done in the early return on L3568

rs/state_manager/src/lib.rs
michael-weigelt: 1) I can introduce an additional condition to actually hash every so often. 

2)
>But if there is a certification in certifications, we should copy it to certification_metadata right away (so that consensus does not have to deliver it twice).

Why? There is an implicit assumption in many places that the `snapshots` and `certification_metadata` key set are always equal (modulo possibly the initial height). Writing the certification over without having the state will violate that assumption.

rs/state_manager/src/lib.rs
mraszyk: But on the OC subnet (for example), cloning used to be the more expensive part.

schneiderstefan: This is [no](https://grafana.mainnet.dfinity.network/d/em7Wo7dVz/state-manager-api-calls?var-period=$__auto&from=2026-03-31T08:52:21.838Z&to=2026-03-31T09:15:50.756Z&timezone=utc&var-datasource=PE62C54679EC3C073&var-ic=mercury&var-ic_subnet=2fq7c-slacv-26cgz-vzbx2-2jrcs-5edph-i5s2j-tck77-c3rlz-iobzx-mqe&var-instance=$__all&var-op=$__all&var-op_checkpoint=$__all&viewPanel=panel-28-clone-1) [longer](https://grafana.mainnet.dfinity.network/d/em7Wo7dVz/state-manager-api-calls?var-period=$__auto&from=2026-03-31T08:52:21.838Z&to=2026-03-31T09:15:50.756Z&timezone=utc&var-datasource=PE62C54679EC3C073&var-ic=mercury&var-ic_subnet=2fq7c-slacv-26cgz-vzbx2-2jrcs-5edph-i5s2j-tck77-c3rlz-iobzx-mqe&var-instance=$__all&var-op=$__all&var-op_checkpoint=$__all&viewPanel=panel-28-clone-3) the case. Cloning is now quite cheap. The caveat however is that this doesn't measure the overhead from potentially having to clone more in the execution round.

rs/state_manager/src/lib.rs
mraszyk: I mean we should compare the hash from the certification against the hash we computed locally.

rs/state_manager/src/lib.rs
mraszyk: I couldn't follow - is there a concrete sequence of actions that leads to an issue?

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,63 @@ impl StateManager for StateManagerImpl {
...
+                    &self.metrics,
+                    &self.log,
+                )
+                .unwrap_or_else(|err| fatal!(self.log, "Failed to compute hash tree: {:?}", err));
+                certification.certified_state_hash.clone()
+            } else {
+                // After awaiting the hashing thread, snapshot and certification_metadata
+                // must have an entry at height-1, so we can unwrap.
+                let states = self.states.read();
+                if let Some(cert_md) = states.certifications_metadata.get(&prev_height) {
```
mraszyk, line 3307: We should also check if we can read the hash from `state.certifications` (since the hash thread does not always push to `certifications_metadata` (currently it does not push to `certifications_metadata` if there's a certification in `certifications`).

rs/state_manager/src/lib.rs
mraszyk: > I can introduce an additional condition to actually hash every so often.

Ideally, `HashRequest::HashState` would tell the hash thread if it should always hash (the case of no catch-up optimization) or only hash if there's no certification in `certifications` (the case of catch-up optimization with consensus being slow to deliver certifications).

> But if there is a certification in certifications, we should copy it to certification_metadata right away (so that consensus does not have to deliver it twice).

What I mean is that we should store the certification from `certifications` (if available) into `certification_metadata` computed [here](https://github.com/dfinity/ic/pull/8464/files#diff-a84c433354ffd59be6c8650af0b10231479e0c93e0b2bc92bac99746599e5b6bR3572).

rs/state_manager/src/lib.rs
```diff
@@ -3362,16 +3353,21 @@ impl StateManager for StateManagerImpl {
...
+        let states = self.states.read();
+        let hash_at_height_available = states.certifications.contains_key(&height);
+        drop(states);
         let fast_forward_height = self.fast_forward_height.load(Ordering::Relaxed);
         if matches!(scope, CertificationScope::Metadata)
             && height.get() < fast_forward_height
             && !height
                 .get()
                 .is_multiple_of(MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING)
+            && hash_at_height_available
```
mraszyk: We should also send a hash request in this early-return case since we no longer compute the hash synchronously in `take_tip`.

mraszyk (commented):
Please also run `ict t rejoin_test_long_rounds` before proceeding with this PR.

rs/state_manager/src/lib.rs
michael-weigelt: 4f8d009a1b

rs/state_manager/src/lib.rs
michael-weigelt: 4f8d009a1b

rs/state_manager/src/lib.rs
michael-weigelt: 4f8d009a1b

rs/state_manager/src/lib.rs
michael-weigelt: I changed that, so I think this is obsolete?

rs/state_manager/src/lib.rs
michael-weigelt: We do nothing until at some point, state sync will bring us up to speed. 
We could also crash and state sync, but if state sync is already running, this is a bit better.

rs/state_manager/src/lib.rs
michael-weigelt: In c&c, 
`state.metadata.prev_state_hash = Some(CryptoHashOfPartialState::from(prev_state_hash));`
should also cover the state sync case.

rs/state_manager/src/lib.rs
michael-weigelt: Why? We want to skip hashing the state if consensus already gave us the hash.

rs/state_manager/src/lib.rs
mraszyk: Your change only applies to height for which we do compute the hash locally. Heights on which the catch-up optimization triggers are not covered by the commit https://github.com/dfinity/ic/commit/4f8d009a1b9de4c53fc0a4485cd52d9a80d53924

rs/state_manager/src/lib.rs
mraszyk: Sorry, my mental model is that `certifications` might not contain the certification (which can't be the case anymore with the stronger condition `&& hash_at_height_available`.

rs/state_manager/src/lib.rs
mraszyk: I see. Essentially you move the point in time when the field is set from `take_tip` to the subsequent (after `execute_round`) `commit_and_certify`. Since the field `prev_state_hash` is not really used by `execute_round`, that seems fine, but it could be a bit fragile (to have a field that is only properly set after `execute_round`), e.g., there are some comments about the field being set by `take_tip` in `rs/replicated_state/src/metadata_state.rs` - maybe the field could be removed from `ReplicatedState` entirely (in a separate PR)?

rs/state_manager/src/lib.rs
mraszyk: > if state sync is already running, this is a bit better.

Won't we crash in the subsequent `take_tip` anyway (because no state was set as the tip after this early return from `commit_and_certify`)?

rs/state_manager/src/lib.rs
michael-weigelt: 1) remove_states_below could be executed just after the hashing thread wrote the snapshot and hash @ tip_height. 
2) 
- take_tip @ x
- state_sync @ y (y>>x)
- remove_states_below around y, removes x
- c&c @ x will find neither certification nor certification_metadata and run into `Previous state hash was not available after awaiting the hash thread`

Maybe there exist other scenarios too.

rs/state_manager/src/lib.rs
mraszyk: EDIT: now that state cloning and hashing is coupled together, this particular metric does not make sense. What we could track instead is the number of times when we could skip cloning (and would do so now before this PR), but don't do so after all because we don't have a certification. Such a metric would show how often we skip the optimization although we could still do something (namely not clone).

rs/state_manager/src/lib.rs
mraszyk: I see, thanks a lot for describing the scenario. So this (the fact we need to preserve `tip_height`) is a consequence of moving the assignment of `prev_state_hash` from `take_tip` to the subsequent `commit_and_certify` (which now potentially refers to `tip_height`). Right?

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3527,165 @@ impl StateManager for StateManagerImpl {
...
+                                    if let Err(err) = state_layout.create_diverged_state_marker(height) {
+                                        error!(
+                                            log,
+                                            "Failed to mark state @{} diverged: {}", height, err
+                                        );
+                                    }
+                                    panic!(
+                                        "Committed state @{height} with hash {hash:?} which is different from delivered hash {delivered_hash:?}"
+                                    );
+                                }
```
mraszyk: This used to be deduplicated using a closure:
```
let assert_prev_hash_matches = |prev_hash| {
            let hash = &certification_metadata.certified_state_hash; 
            if prev_hash != hash {
                if let Err(err) = self.state_layout.create_diverged_state_marker(height) {
                    error!(
                        self.log,
                        "Failed to mark state @{} diverged: {}", height, err
                    );
                }   
                panic!(
                    "Committed state @{height} with hash {hash:?} which is different from previously computed or delivered hash {prev_hash:?}"
                );
            }
        };
```

rs/state_manager/src/lib.rs
```diff
@@ -3411,18 +3404,22 @@ impl StateManager for StateManagerImpl {
             CertificationScope::Metadata => Arc::new(state),
         };
 
-        let mut certification_metadata =
-            Self::compute_certification_metadata(&state, height, &self.metrics, &self.log)
-                .unwrap_or_else(|err| fatal!(self.log, "Failed to compute hash tree: {:?}", err));
+        // Kick off hashing of the new state. At this point, we are either
+        // - not catching up, so there is no delivered_certification, or there is one but
+        // - we are at a height which is a multiple of MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING,
```
mraszyk: another case: we could also be catching up (`height.get() < fast_forward_height`), but there was still no delivered certification; maybe we could omit the case distinction and simply say that we hash even if we potentially have a certification?

rs/state_manager/tests/state_manager.rs
mraszyk: this test should now pass again on heights for which the optimization does not trigger (and consequently the hash thread does hash)

rs/state_manager/src/lib.rs
michael-weigelt: > but don't do so after all because we don't have a certification

Agree, I will add this metric.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9493,12 +9434,13 @@ fn commit_and_certify_reuses_certification() {
         // optimization does not trigger
         let state = sm.take_tip().1;
         sm.commit_and_certify_at_height(state, no_opt_height, CertificationScope::Metadata, None);
+        sm.flush_hash_channel();
         assert_eq!(no_state_clone_count(metrics), 0);
 
-        // `commit_and_certify` reused certification from `states.certifications`
+        // `commit_and_certify` did not clone state or calculate a hash -> no entry in cert_metadata
         assert!(
-            sm.certifications_metadata_certification(no_opt_height)
```
mraszyk, line 9500: `sm.certifications_metadata_certification(no_opt_height).is_some()` should also hold again now

rs/state_manager/src/lib.rs
michael-weigelt: Yes, you can put it that way. And we want that assignment in order to kick off the hash thread shortly after.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9493,12 +9434,13 @@ fn commit_and_certify_reuses_certification() {
         // optimization does not trigger
         let state = sm.take_tip().1;
         sm.commit_and_certify_at_height(state, no_opt_height, CertificationScope::Metadata, None);
+        sm.flush_hash_channel();
         assert_eq!(no_state_clone_count(metrics), 0);
 
-        // `commit_and_certify` reused certification from `states.certifications`
+        // `commit_and_certify` did not clone state or calculate a hash -> no entry in cert_metadata
```
mraszyk: > `commit_and_certify` did not clone state or calculate a hash -> no entry in cert_metadata

that was actually a bug if that happens although the optimization did not trigger

rs/state_manager/tests/state_manager.rs
```diff
@@ -4163,8 +4185,8 @@ fn can_commit_after_prev_state_is_gone() {
 
             assert_eq!(Height(3), dst_state_manager.latest_state_height());
             assert_eq!(
-                dst_state_manager.get_state_at(Height(1)),
-                Err(StateManagerError::StateRemoved(Height(1)))
```
mraszyk, line 4167: I suppose `Height(1)` is the `tip_height` which is now present? Then maybe we could keep an assert saying that the state does exist (to cover the new exception to `remove_states_below`)

rs/state_manager/src/lib.rs
schneiderstefan: I think you're right. So maybe crashing right there is cleaner after all.

michael-weigelt: Aha yes right. Do you prefer crashing here immediately?

michael-weigelt: Unless take_tip can use the one from state sync? It does seem unlikely though

schneiderstefan: [No](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/state_manager/src/lib.rs?L2741), we crash before we even check for finished state syncs.

rs/state_manager/src/lib.rs
michael-weigelt: I'll try to find and adjust these comments. 

We do use the prev_state_hash when we hash the current state. Where should we keep it if not in the replicated state? But perhaps we move that discussion elsewhere.

rs/state_manager/src/lib.rs
michael-weigelt: We check `states.certifications` above, [here](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Adfinity%2Fic.git&branch=mwe%2Fasync_hashing&file=rs%2Fstate_manager%2Fsrc%2Flib.rs&editor=VSCode&version=2.4.0&start_row=3271&start_col=14&end_row=3271&end_col=14), and return it as `prev_state_hash` immediately if present. 
Not sure why I did not mention that before. Does that address the issue or am I misunderstanding it?

rs/state_manager/src/lib.rs
michael-weigelt: Yeah I worded this badly, implying that catch up == the 4 conditions above. Which is not the case.

rs/state_manager/tests/state_manager.rs
michael-weigelt: 732979355f

rs/state_manager/src/lib.rs
michael-weigelt: https://dfinity.atlassian.net/browse/DSM-122

rs/state_manager/src/lib.rs
mraszyk: > We check states.certifications above, here, and return it as prev_state_hash immediately if present.

That's what I missed by the time I was reading here. Sorry for the confusion.

rs/state_machine_tests/src/lib.rs
```diff
@@ -3328,6 +3329,7 @@ impl StateMachine {
         state.put_canister_state(source_state.canister_state(&canister_id).unwrap().clone());
         self.state_manager
             .commit_and_certify(state, CertificationScope::Full, None);
+        self.state_manager.flush_hash_channel();
```
mraszyk, line 3343: Could you please call `flush_hash_channel` after every `commit_and_certify` in StateMachine tests to ensure that the latest committed state is always equal to the tip?

rs/state_manager/tests/state_manager.rs
```diff
@@ -1097,7 +1100,8 @@ fn returns_state_no_committed_for_future_states() {
 }
 
 #[test]
-#[should_panic(expected = "which is different from previously computed or delivered hash")]
+#[should_panic(expected = "failed to wait for hashing thread: RecvError")]
+// The hash thread panics with "failed to wait for hashing thread: RecvError", but we can't relay this.
```
mraszyk:
```suggestion
// The hash thread panics with "which is different from previously computed or delivered hash", but we can't relay this.
```
I suppose

rs/state_manager/tests/state_manager.rs
```diff
@@ -8987,17 +9028,18 @@ fn commit_and_certify_optimization_conditions() {
 
         // heights of 10 and 20 are divisible by 10 => optimization does not trigger at those heights
         let mut expected_no_state_clone_count = 1;
-        for height in 3..21 {
+        for height in 3_u64..21_u64 {
+            if !height.is_multiple_of(10) {
+                expected_no_state_clone_count += 1;
+                sm.deliver_state_certification(fake_certification_for_height(Height::new(height)));
```
mraszyk, line 9034: let's always deliver certification so that the optimization does not trigger because of the height being a multiple of 10 (and not because of the certification not being delivered which should be covered by a separate test)

rs/state_manager/tests/state_manager.rs
```diff
@@ -9338,16 +9277,17 @@ fn get_state_hash_at() {
         // optimization triggers every multiple of 10
         let mut expected_no_state_clone_count = 0;
         for height in 1..checkpoint_height.get() {
+            if !height.is_multiple_of(10) {
+                sm.deliver_state_certification(fake_certification_for_height(Height::new(height)));
```
mraszyk, line 9281: let's always deliver certification so that the optimization does not trigger because of the height being a multiple of 10 (and not because of the certification not being delivered which should be covered by a separate test)

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,59 @@ impl StateManager for StateManagerImpl {
             states.tip_height.increment()
         };
 
-        self.populate_extra_metadata(&mut state, height);
+        self.populate_extra_metadata(&mut state);
+
+        // Get the previous state hash either from consensus via certifications (if we are catching up)
+        // or wait for the hashing thread to finish computing it.
+        let prev_height = height.decrement();
```
mraszyk: This would underflow for the initial state height.

rs/state_manager/src/lib.rs
```diff
@@ -3362,29 +3349,37 @@ impl StateManager for StateManagerImpl {
...
-        let fast_forward_height = self.fast_forward_height.load(Ordering::Relaxed);
-        if matches!(scope, CertificationScope::Metadata)
-            && height.get() < fast_forward_height
-            && !height
-                .get()
-                .is_multiple_of(MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING)
-        {
+        let maybe_delivered_certification = {
+            // Scope to drop this lock.
             let mut states = self.states.write();
```
mraszyk, line 3359: We could start with a read lock and only acquire a write lock later, but I guess the impact is minimal so it could also just stay as-is (just pointing it out as food for thought).

rs/state_manager/src/lib.rs
```diff
@@ -3362,29 +3349,37 @@ impl StateManager for StateManagerImpl {
...
+            let maybe_delivered_certification = states.certifications.get(&height).cloned();
+            let fast_forward_height = self.fast_forward_height.load(Ordering::Relaxed);
+            if matches!(scope, CertificationScope::Metadata)
+                && height.get() < fast_forward_height
+                && !height
+                    .get()
+                    .is_multiple_of(MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING)
+                && maybe_delivered_certification.is_some()
+            {
+                // let mut states = self.states.write();
```
mraszyk: left-over comment

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3526,160 @@ impl StateManager for StateManagerImpl {
...
+enum HashRequest {
+    HashState {
+        state: Arc<ReplicatedState>,
+        states: Arc<parking_lot::RwLock<SharedState>>,
+        latest_state_height: Arc<AtomicU64>,
+        height: Height,
+        latest_height_update_time: Arc<Mutex<Instant>>,
+        /// A certification from consensus. If `Some`, we compare it with the state hash
+        /// we calculate and panic on divergence. It should be `Some` whenever we are catching up and could
+        /// skip hashing, but do so anyway because we are at a height which is a multiple of
```
mraszyk:
```suggestion
        /// skip hashing, but we do hash anyway because we are at a height which is a multiple of
```

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3526,160 @@ impl StateManager for StateManagerImpl {
...
+        /// `MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING`.
```
mraszyk, line 3538: could be renamed to `MAX_CONSECUTIVE_ROUNDS_WITHOUT_STATE_CLONING_AND_HASHING` (independently of my suggestion to drop the field `reference_certification` and consequently this doc comment, too)

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3526,160 @@ impl StateManager for StateManagerImpl {
...
+            })
+            .unwrap(),
+    );
+    (handle, hash_req_sender)
+}
+
+impl StateManagerImpl {
+    /// After this method terminates, both `SharedState.snapshots` and `SharedState.certification_metadata`
+    /// at the height from the previous `commit_and_certify` are populated. It also updates `latest_state_height`
+    /// to the maximum of the value before and the height passed.
```
mraszyk:
```suggestion
    /// to the maximum of the value before and the committed height.
```

rs/state_manager/src/lib.rs
```diff
@@ -3589,6 +3526,160 @@ impl StateManager for StateManagerImpl {
...
+            .spawn(move || {
+                while let Ok(req) = receiver.recv() {
+                    match req {
+                        HashRequest::HashState {
+                            state,
+                            states,
+                            latest_state_height,
+                            height,
+                            latest_height_update_time,
+                            reference_certification,
```
mraszyk, line 3566: I think we can just read the certification from `states.certifications` here instead of passing it via `HashRequest::HashState` - by only reading the certification now that we need it, we do not skip the check in cases when the certification is only populated by consensus in between sending and processing `HashRequest`.

rs/state_manager/src/lib.rs
mraszyk, line 2244: Is this covered by a test, i.e., is there a test failing if this line was (accidentally) removed?

rs/state_manager/tests/state_manager.rs
```diff
@@ -9493,12 +9436,13 @@ fn commit_and_certify_reuses_certification() {
...
         sm.commit_and_certify_at_height(state, no_opt_height, CertificationScope::Metadata, None);
+        sm.flush_hash_channel();
         assert_eq!(no_state_clone_count(metrics), 0);
 
         // `commit_and_certify` reused certification from `states.certifications`
         assert!(
-            sm.certifications_metadata_certification(no_opt_height)
-                .is_some()
+            sm.certifications_metadata_heights()
+                .contains(&no_opt_height)
```
mraszyk: The new assertion is trivial - we should make sure that the certification is populated _inside_ certification metadata (not that certification metadata at the given height are pushed which is trivial).

rs/state_manager/src/lib.rs
```diff
@@ -3328,7 +3263,59 @@ impl StateManager for StateManagerImpl {
...
+        drop(states);
+        let prev_state_hash = if let Some(hash) = maybe_hash {
+            hash
+        } else {
+            // Wait for the hashing thread.
+            let (sender, recv) = bounded(1);
+            self.hash_channel
+                .send(HashRequest::Wait { sender })
+                .expect("Failed to send `Wait` to hash channel");
+            recv.recv().expect("Failed to wait for hash channel");
```
mraszyk: Shouldn't we only do this if the previous height is not 0?

mraszyk (commented):
I've done another round of review.

rs/state_manager/src/lib.rs
schneiderstefan: We could possibly add a state manager test that:
1. Create a checkpoint at height 1
2. take_tip/commit to create an in-memory state at 2. (not sure if this step is needed)
3. Call take_tip
4. Call fetch_state at height 10, with the same hash as state 1.
5. Call remove_states_below
6. commit_and certify

rs/state_machine_tests/src/lib.rs
michael-weigelt: I'd like to do this via the testing trait in a follow-up: https://dfinity.atlassian.net/browse/DSM-126

rs/state_manager/tests/state_manager.rs
michael-weigelt: If we deliver a fake certification, that fake hash will conflict with the hash that we calculate. So I am not sure how to achieve your scenario.

rs/state_manager/src/lib.rs
michael-weigelt: A few lines above, height is incremented, so this should always work.

rs/state_manager/src/lib.rs
michael-weigelt: Could we? If we hold a read lock, there is no atomic upgrade to a write lock, so we would introduce a race condition if we release and then acquire.

rs/state_manager/tests/state_manager.rs
mraszyk: The way this is done in other tests is to deliver a fake certification with a fixed hash derived from the panic message (i.e., I start with a dummy hash, run the test, it panics because of hash mismatch, and then I replace the dummy hash with the one from the panic message). For instance,
```
#[test]
fn remove_inmemory_states_below_prunes_certification() {
    state_manager_test(|metrics, sm| {
        // consensus delivers certification for the next height
        let cert_height = Height::new(1);
        // the state hash was derived from the panic message
        let state_hash = CryptoHash(
            hex::decode("4e2d174de5daaeb4622d8f5e426ee09274f7ec4fb01d62fb9a3d36ae50961353")
                .unwrap(),
        );
```

rs/state_machine_tests/src/lib.rs
mraszyk: That sounds like a good long-term solution, but I'd suggest to also manually ensure this property on this PR.

rs/state_manager/src/lib.rs
mraszyk: I see, how about:
```
let (prev_height, height) = {
            let states = self.states.read();
            let prev_height = states.tip_height;
            let height = prev_height.increment();
            (prev_height, height)
        };
```

rs/state_manager/src/lib.rs
mraszyk: Good point! In principle, there could even be a race condition due to
```
let height = {
            let states = self.states.read();
            states.tip_height.increment()
        };
```
which also holds the read lock only temporarily before.

rs/state_manager/src/lib.rs
michael-weigelt: I am not certain. I think at the time we read this in `commit_and_certify`, the certification is protected from being pruned because it is at `tip_height`. But if we check in the thread directly, the read in this thread might be running after the other thread has passed `tip_height.increment()` in the next c&c, and then a third thread could be running `remove_states_below`, and hence we could not read the desired certification. 
Tbh I am not sure if this scenario is possible. @schneiderstefan wdyt? The recommendation to pass the value in was random or was it this?

mraszyk: I believe that the hash thread must fundamentally complete before we can advance tip to the next state (otherwise, we would not be able to set prev_state_hash in the ReplicatedState).

mraszyk: On the other hand, the impact of the proposed refactoring is minimal so the comment is certainly not blocking this PR.

rs/state_manager/tests/state_manager.rs
michael-weigelt: I am not sure what you mean. In there @no_opt_height, we have a `Certification {height, witness, signed: Hash}`. What should we check about that?

rs/state_manager/src/lib.rs
michael-weigelt: If the height is 0, there is no work being done in the hashing thread, so the `Wait` task will succeed immediately. So it cannot fail at height 0, and it will not take significant time.

rs/state_manager/tests/state_manager.rs
mraszyk: - before, we would look up the struct CertificationsMetadata at the given no_opt_height and assert that the field certification in that struct is set to Some
- now we just assert that the struct CertificationsMetadata exists for the given no_opt_height, but we do not check the field certification inside.

rs/state_manager/src/lib.rs
mraszyk: Sure, but why bother at all instead of moving the code into the else branch?

rs/state_manager/src/lib.rs
michael-weigelt: I linked this comment in https://dfinity.atlassian.net/browse/DSM-126

rs/state_manager/src/lib.rs
schneiderstefan: I think there are pros and cons to both, but overall I agree with Martin.

On the one hand, the current version has the drawback that Martin mentions, namely that the certification could arrive while we calculate the hash, and in that (likely rare) case we would ask for the certificate twice.

On the other hand, delaying the lookup could run into a situation where it was there in c&c, but not anymore when we get to it. This can happen if we change the tip height in take_tip [here](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/state_manager/src/lib.rs?L2824) and then call `remove_states_below`. This case is even rarer though.

On the other hand, I don't think it can happen with the increment of the tip_height in the next c&c, because we flush the hashing thread before we get to the increment. Though maybe we should also flush the thread in the case where we [see the certificaton](https://github.com/dfinity/ic/pull/8464/files#diff-a84c433354ffd59be6c8650af0b10231479e0c93e0b2bc92bac99746599e5b6bR3278) in c&c. Otherwise we might see the certification in c&c, but we are also still computing the hash, and by the end of the hash computation we deleted the certificaton.

michael-weigelt: > I believe that the hash thread must fundamentally complete before we can advance tip to the next state 

Ah I forgot that `tip_height.increment()` at the beginning of c&c is an update to the local variable, and we only write it to `states.tip/states.tip_height` after we have awaited.

rs/state_manager/src/lib.rs
michael-weigelt: Yeah that keeps related code closer together. Thanks.

rs/state_manager/src/lib.rs
michael-weigelt: bb81b7d259

rs/state_manager/tests/state_manager.rs
michael-weigelt: That scales for one delivered fake hash, but here (and below) we are in a loop...

rs/state_machine_tests/src/lib.rs
michael-weigelt: I did it for those tests that required it because they needed to observe the effect. But changing the roughly 260 invocations of c&c that don't seem to need it just to remove those changes again in the next iteration... I'd prefer not to.

rs/state_manager/src/lib.rs
michael-weigelt: Postponed.

rs/state_machine_tests/src/lib.rs
mraszyk: > 260 invocations

I only a couple in rs/state_machine_tests/src/lib.rs, are StateMachine tests committing state manually outside rs/state_machine_tests/src/lib.rs (if so, I wouldn't bother fixing that pattern now).

rs/state_manager/tests/state_manager.rs
mraszyk: I see, but then we could drop the loop and just test that the optimization does not trigger at a single height if no certification is delivered.

mraszyk: But are you sure about the issue with the hash, I quickly asked Claude and it seems the test passes if the certification is always delivered.

michael-weigelt: I ran it as proposed before replying initially. Not sure what, if anything, Claude did.

rs/state_machine_tests/src/lib.rs
michael-weigelt: 0dfb685ff8

rs/state_manager/tests/state_manager.rs
mraszyk:
```
diff --git a/rs/state_manager/tests/state_manager.rs b/rs/state_manager/tests/state_manager.rs
index b8f3a121e5..4716bc2671 100644
--- a/rs/state_manager/tests/state_manager.rs
+++ b/rs/state_manager/tests/state_manager.rs
@@ -9029,9 +9029,9 @@ fn commit_and_certify_optimization_conditions() {
         // heights of 10 and 20 are divisible by 10 => optimization does not trigger at those heights
         let mut expected_no_state_clone_count = 1;
         for height in 3_u64..21_u64 {
+            sm.deliver_state_certification(fake_certification_for_height(Height::new(height)));
             if !height.is_multiple_of(10) {
                 expected_no_state_clone_count += 1;
-                sm.deliver_state_certification(fake_certification_for_height(Height::new(height)));
             }
             let state = sm.take_tip().1;
             sm.commit_and_certify_at_height(
```
Claude claims:
```
● Bash(bazel test //rs/state_manager:state_manager_test --test_filter=commit_and_certify_optimization_conditions  
      --test_output=errors 2>&1 | tail -20)                                                                       
  ⎿  [2,316 / 2,322] Compiling Rust rlib canonical_state v0.9.0 (16 files); 4s remote-cache, processwrapper-s     
     andbox ... (2 actions running)                                                                               
     [2,317 / 2,322] Compiling Rust rlib canonical_state v0.9.0 (16 files); 5s remote-cache, processwrapper-s     
     … +16 lines (ctrl+o to expand)                                                                               
  ⎿  (timeout 5m)                                                                                                 
                                                                                                                  
● Test passes. The change moves deliver_state_certification outside the if so it's called unconditionally for     
  every height, and the counter increment stays gated on the non-multiple-of-10 check. The optimization condition 
  in the implementation independently guards against triggering at multiples of 10.       
```

michael-weigelt:
```
bazel test //rs/state_manager:state_manager_integration --test_output=streamed --cache_test_results=no --test_arg=--nocapture --test_arg=commit_and_certify_optimization_conditions
INFO: Invocation ID: 0b1fbb7a-87bb-49cd-b65b-30b5d3740978
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
INFO: Analyzed target //rs/state_manager:state_manager_integration (986 packages loaded, 36071 targets configured).

running 1 test

thread 'HashThread' (3165076) panicked at rs/state_manager/src/lib.rs:3596:37:
Committed state @10 with hash CryptoHash(0xd90d05f3d971482d2ccdc594de723d786afaed32737b19e42f85a1ebe8e21a21) which is different from delivered hash CryptoHash(0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a)
```

michael-weigelt: 439da396f4

mraszyk: Perfect, thanks a lot! (I could confirm that it also failed for me, Claude ran wrong target... - state_manager_test instead of state_manager_integration).

schneiderstefan (approved):

PR #8469, opened 2026-01-22
feat: CON-1391 Return result of remote NiDKG as soon as sufficient dealings are available

eichhorl:
## Background
Remote NiDKG requests (i.e. for subnet recovery/creation) currently take anywhere between 1 and 2 DKG intervals of the source subnet to complete. This is because once a request is made, the corresponding NiDKG config will be included only in the following summary block, and the final transcript is only created in the subsequent summary block.

We can reduce this latency by creating and delivering the transcripts already as part of a data block, as soon as enough dealings are available on the blockchain.

In https://github.com/dfinity/ic/pull/8644, we introduced the new data payload field required to hold and deliver NiDKG transcripts to execution. Initially, payloads containing these early remote transcripts were not created, and did not pass validation.

## Proposed Changes
This PR changes the data payload builder to create and include transcripts as soon as enough dealings are available on the parent chain. For now, we will include at most 2 transcripts per data payload. The payload validator validates early transcripts by re-creating them with the same inputs, and checking for equality.

Note that `SetupInitialDKG` requests require 2 transcripts, whereas `ReshareChainKey` requests for VetKeys require 1 transcript. Furthermore, execution expects us to deliver **all** transcripts for a given request in the same payload. Therefore, we make sure that for a given target ID, we will either include **all** transcript results in the data payload, or **none** of them.

Generally, the introduced code paths should be seen as an optimization. NiDkg configs and transcripts continue to be created as part of summary blocks (unless the transcript was already created as part of an earlier data block).

One special case is the following: Assume _early_ transcript creation for a `SetupInitialDKG` request fails, but succeeds for _one_ of the transcripts when building the summary payload. Then the summary payload containing the transcript will also contain a config for the second transcript that failed (it will be retried in the next interval).

In order to handle this case as part of a following data block, we would have to identify and clone the completed sibling transcript into the data payload creating the second transcript. This would add some complexity which we avoid for now by handling this case exclusively as part of summary blocks. Specifically: we will only handle target IDs as part of data blocks for which all configs can be found in the previous summary. Any other cases (retries, timeouts) continue to be handled and initiated by the summary payload.


rs/consensus/dkg/src/lib.rs
```diff
@@ -1596,7 +1615,524 @@ mod tests {
...
+                vec![10, 11, 12, 13],
+                None,
+                Some(target_id),
+            );
+
+            deps.pool
+                .advance_round_normal_operation_n(EARLY_DKG_INTERVAL + 1);
+            let remote_dkg_ids = extract_remote_dkg_ids_from_highest_block(&deps.pool, target_id);
+            assert_eq!(remote_dkg_ids.len(), 1);
+            assert_eq!(extract_dkg_configs_from_highest_block(&deps.pool).len(), 4);
```
pierugo-dfinity: Why 4? Local (Low), Local (High), Remote (Reshare request), ...?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -114,12 +124,136 @@ fn create_data_payload(
...
+
+    let state_ref = state.get_ref();
+    let callback_id_map = build_target_id_callback_map(state_ref);
+
+    // Try to create transcripts for all configs of each target_id. Note that we either include
+    // all transcript results for a target_id or none of them.
+    let mut selected_transcripts = vec![];
+    for (target_id, configs) in remote_configs {
+        // Lookup the callback id and the expected number of configs for this target_id
+        let Some((expected_config_num, callback_id)) = callback_id_map.get(&target_id) else {
```
pierugo-dfinity: Is it an error to not find the target_id in the context manager? Should we log an error/increase a metric in that case?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -114,12 +124,136 @@ fn create_data_payload(
...
+            continue;
+        };
+
+        // Check that we have the expected number of configs for this target_id
+        if configs.len() != *expected_config_num {
+            // This may happen if we only managed to create one transcript (out of two) as part
+            // of the last summary block. We will handle this in the next summary block instead.
```
pierugo-dfinity: Maybe
```suggestion
            // This may happen if we did not manage to create all required transcripts as part of
            // the last summary block. We will handle this in the next summary block instead.
```
to be more generic (instead of being specific to `SetupInitialDKG`)

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -114,12 +124,136 @@ fn create_data_payload(
...
+            // Generate the transcript. We need to retry transient errors, as a payload containing
+            // transient errors may not be verifiable by peers.
+            let transcript_result = match create_transcript(crypto, config, &all_dealings, &logger)
+            {
+                Ok(transcript) => Ok(transcript),
+                // Note that we handled the reproducible error case of not having enough dealings
+                // already beforehand.
+                Err(err) if err.is_reproducible() => {
+                    // Including the error in the payload will cause the context to receive
+                    // a reject response.
```
pierugo-dfinity, line 235: I feel it hacky to consider that case separately above. An alternative could be:

```rust
    'process_target_id: for (target_id, configs) in remote_configs {
        ...
        // For each config, ...
        for config in configs.iter() {
            // Generate the transcript. ...
            let transcript_result = match ...
            {
                Ok(transcript) => ...,
                // If a config has not enough dealings, we skip this target_id altogether
                Err(DkgCreateTranscriptError::InsufficientDealings(_)) => continue 'process_target_id,
                // Note that we handled the error case of not having enough dealings above.
                Err(err) if err.is_reproducible() => {
                    ...
                }
                Err(err) => {
                    ...
                }
            };
            ...
        }
        ...
    }
```
at the cost of eventually creating a transcript for nothing in case the first one has enough dealings but not the second one.

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -114,12 +124,136 @@ fn create_data_payload(
...
+                    // Return on transient crypto errors
+                    return Err(DkgPayloadCreationError::DkgCreateTranscriptError(err));
+                }
+            };
+            selected_transcripts.push((config.dkg_id().clone(), *callback_id, transcript_result));
+        }
+
+        if selected_transcripts.len() >= MAX_EARLY_REMOTE_TRANSCRIPTS {
+            break;
+        }
```
pierugo-dfinity: If we successfully create a transcript for a `ReshareChainKey` and then two for a `SetupInitialDKG`, wouldn't we have 3 remote transcripts in the same payload? Then peers would actually validate it because they would deterministically create the same payload, but it would break the purpose of that constant.

An alternative could be:
```rust
        let mut transcripts_for_target_id = vec![];
        // For each config, ...
        for config in configs.iter() {
            // Generate the transcript. ...
            let transcript_result = match ...
            {
                ...
            };
            transcripts_for_target_id.push(...);
        }

        // Skip the current target_id if adding it would exceed the maximum number of transcripts per block
        if selected_transcripts.len() + transcripts_for_target_id.len() > MAX_EARLY_REMOTE_TRANSCRIPTS {
            continue;
        }

        selected_transcripts.extend(transcripts_for_target_id);

        // Optimization: if we reached the maximum number of transcripts per block, we can already return now.
        if selected_transcripts.len() >= MAX_EARLY_REMOTE_TRANSCRIPTS {
            break;
        }
    }
```

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -811,40 +970,67 @@ fn get_node_list(
...
+/// is not used to differentiate two DKGs for the same subnet.
 fn eq_sans_height(dkg_id1: &NiDkgId, dkg_id2: &NiDkgId) -> bool {
     dkg_id1.dealer_subnet == dkg_id2.dealer_subnet
         && dkg_id1.dkg_tag == dkg_id2.dkg_tag
         && dkg_id1.target_subnet == dkg_id2.target_subnet
 }
 
-fn add_callback_ids_to_transcript_results(
-    new_transcripts: BTreeMap<NiDkgId, Result<NiDkgTranscript, String>>,
+fn build_target_id_callback_map(
```
pierugo-dfinity, line 1112: Could you maybe add some documentation on the return type?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -811,40 +970,67 @@ fn get_node_list(
...
+        .map(|(&callback_id, context)| (context.target_id, (2, callback_id)))
         .chain(
             call_contexts
                 .reshare_chain_key_contexts
                 .iter()
-                .map(|(&callback_id, context)| (context.target_id, callback_id)),
+                .map(|(&callback_id, context)| (context.target_id, (1, callback_id))),
         )
-        .collect::<BTreeMap<NiDkgTargetId, CallbackId>>();
+        .collect::<BTreeMap<NiDkgTargetId, (usize, CallbackId)>>()
```
pierugo-dfinity:
```suggestion
        .collect()
```
inferred by the return type

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -859,7 +1045,20 @@ fn add_callback_ids_to_transcript_results(
...
+        .metadata
+        .subnet_call_context_manager
+        .setup_initial_dkg_contexts
+        .len()
+        + state
+            .metadata
+            .subnet_call_context_manager
+            .reshare_chain_key_contexts
+            .len()
+}
```
pierugo-dfinity: This function and `build_target_id_callback_map` would both need to be changed in case there would be a new NiDKG protocol to be run. I have been thinking that maybe this function could simply return `build_target_id_callback_map.len()`? Though it would be less clear, but maybe there could still be an alternative where `setup_initial_dkg_contexts` and `reshare_chain_key_contexts` are chained together at only one place?

My suggested simplification would be correct only if there are no two different callback IDs with the same `NiDkgTargetId`.

rs/consensus/dkg/src/test_utils.rs
```diff
@@ -82,6 +87,86 @@ pub(super) fn complement_state_manager_with_reshare_chain_key_request(
...
+            .dkg
+            .transcripts_for_remote_subnets
+    } else {
+        &block
+            .payload
+            .as_ref()
+            .as_data()
+            .dkg
+            .transcripts_for_remote_subnets
+    }
```
pierugo-dfinity:
```suggestion
    match block.payload.as_ref() {
        BlockPayload::Summary(summary) => summary
            .dkg
            .transcripts_for_remote_subnets
        BlockPayload::Data(data) => data
            .dkg
            .transcripts_for_remote_subnets
    }
```
Does this work? Similarly in `extract_dealings_from_highest_block` and `extract_dkg_configs_from_highest_block`

pierugo-dfinity (commented):
Very nice optimization!

I thought of some test cases that you could maybe add:
- If the state manager has one context for a `SetupInitialDKG` and one context for `ReshareChainKey`, the three corresponding transcripts should not appear in the same block (as constrained by `MAX_EARLY_REMOTE_TRANSCRIPTS`). I have the impression that this could fail with the current implementation (see my corresponding comment)
- If the state manager has one context for a `SetupInitialDKG` and one context for `ReshareChainKey`, but one of `SetupInitialDKG`’s configs has not enough dealings to create a transcript, then there should still be a transcript for `ReshareChainKey` (concretely, it tests that the `continue` at new line payload_builder.rs:217 is not a `break`)
- As you documented in `test_no_early_transcripts_for_single_setup_initial_dkg_config`, you could extend that test to “Instead, the next summary block should contain both transcripts.”, I don’t think you check that.

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Fair, done 👍

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Thanks!

rs/consensus/dkg/src/test_utils.rs
eichhorl: Nice!

rs/consensus/dkg/src/payload_builder.rs
eichhorl: > at the cost of eventually creating a transcript for nothing in case the first one has enough dealings but not the second one

Yes, this is exactly why I think the case should be handled separately. Currently, the DKG for which a dealing is included in a block is somewhat random. So the case where there are enough dealings for one transcript, but not enough for the second may actually be very common. Then we would create the first transcript in every single block until we collected enough dealings for the second transcript. Since creating a transcript is fairly expensive, I think doing it unnecessarily should be avoided.

rs/consensus/dkg/src/lib.rs
eichhorl: The VetKD transcript also needs to be reshared locally

pierugo-dfinity: Thanks 👍

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Okay I see, then let's keep it that way.

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Good point, checking the length of  `build_target_id_callback_map` makes sense to me 👍

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Good catch! I moved the check to the beginning of the loop

rs/consensus/dkg/src/payload_builder.rs
eichhorl: I added a warning. There are no metrics in NiDKG currently, so adding them is a bit more work. I would prefer to do it in a follow up.

rs/consensus/dkg/src/payload_builder.rs
eichhorl: I added a test (and also included your second test case above)

rs/consensus/dkg/src/lib.rs
```diff
@@ -1596,7 +1618,808 @@ mod tests {
...
+    /// dealings for one of its configs:
+    /// - The first setup target is skipped (insufficient dealings).
+    /// - The remaining two contexts compete for MAX_EARLY_REMOTE_TRANSCRIPTS
+    ///   (= 2), and only the first context's transcripts are included.
+    #[test]
+    fn test_early_remote_transcripts_respects_max() {
+        for (setup_target_bytes, reshare_target_bytes, desc) in [
+            ([1u8; 32], [2u8; 32], "SetupInitialDKG first"),
+            ([2u8; 32], [1u8; 32], "ReshareChainKey first"),
+        ] {
```
pierugo-dfinity:
```suggestion
        for (skipped_target_bytes, setup_target_bytes, reshare_target_bytes, desc) in [
            ([0u8; 32], [1u8; 32], [2u8; 32], "SetupInitialDKG first"),
            ([0u8; 32], [2u8; 32], [1u8; 32], "ReshareChainKey first"),
        ] {
```
nit: I find it clearer to have all target bytes (and target ID below) defined similarly

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -811,40 +976,71 @@ fn get_node_list(
...
         && dkg_id1.dkg_tag == dkg_id2.dkg_tag
         && dkg_id1.target_subnet == dkg_id2.target_subnet
 }
 
-fn add_callback_ids_to_transcript_results(
-    new_transcripts: BTreeMap<NiDkgId, Result<NiDkgTranscript, String>>,
+// Build a map from target id to callback id according to contexts in the replicated state.
+// Additionally, for each target ID, return the expected number of DKG instances necessary
+// to answer the request. Specifically, setup initial DKG requests require two DKGs, whereas
+// resharing a chain key requires on DKG instance.
```
pierugo-dfinity:
```suggestion
/// Build a map from target id to callback id according to contexts in the replicated state.
/// Additionally, for each target ID, return the expected number of DKG instances necessary
/// to answer the request. Specifically, setup initial DKG requests require two DKGs, whereas
/// resharing a chain key requires on DKG instance.
```

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -344,6 +493,7 @@ fn compute_remote_dkg_data(
         state.get_ref(),
         validation_context,
         reshared_transcripts,
+        completed_target_ids,
         logger,
     )?;
```
pierugo-dfinity, line 592: Do you think there could be an alternative to passing `completed_target_ids` multiple times up the stack? I.e. letting `process_subnet_call_context` return all configs (even those already completed) and do the filtering in this function (`compute_remote_dkg_data`) instead? That way, there would be only one place where we do
```rust
// If the DKG has already been completed, skip this context
if completed_target_ids.contains(&context.target_id) {
    continue;
}
```
It doesn't look super easy though, so this is not a strong opinion

pierugo-dfinity (approved):
Looks really cool 🚀

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Perfect 👍

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Even better. Thank you! Looks great

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Nice

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Not really, since these function returns three different collections we would have to filter all three of them

rs/consensus/dkg/src/lib.rs
eichhorl: done 👍

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Hmm yeah that's what I feared

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -811,40 +976,71 @@ fn get_node_list(
...
         && dkg_id1.dkg_tag == dkg_id2.dkg_tag
         && dkg_id1.target_subnet == dkg_id2.target_subnet
 }
 
-fn add_callback_ids_to_transcript_results(
-    new_transcripts: BTreeMap<NiDkgId, Result<NiDkgTranscript, String>>,
+/// Build a map from target id to callback id according to contexts in the replicated state.
+/// Additionally, for each target ID, return the expected number of DKG instances necessary
+/// to answer the request. Specifically, setup initial DKG requests require two DKGs, whereas
+/// resharing a chain key requires on DKG instance.
```
pierugo-dfinity:
```suggestion
/// resharing a chain key requires one DKG instance.
```

rs/consensus/dkg/src/test_utils.rs
```diff
@@ -55,33 +97,81 @@ pub(super) fn complement_state_manager_with_reshare_chain_key_request(
...
+        BlockPayload::Summary(_) => vec![],
+        BlockPayload::Data(data) => data.dkg.messages.clone(),
+    }
+}
+
+/// Extract the remote dkg transcripts from the current highest validated block
+pub(super) fn extract_remote_dkg_ids_from_highest_block(
+    pool: &TestConsensusPool,
+    target_id: NiDkgTargetId,
+) -> Vec<NiDkgId> {
```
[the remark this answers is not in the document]
eichhorl: Done

rs/test_utilities/artifact_pool/src/consensus_pool.rs
```diff
@@ -160,7 +160,7 @@ fn dkg_payload_builder_fn(
             &*state_manager,
             validation_context,
             no_op_logger(),
-            10, // at most dealings per block
+            20, // at most dealings per block
```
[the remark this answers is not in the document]
eichhorl: Switched to a constant

rs/consensus/dkg/src/payload_validator.rs
```diff
@@ -195,8 +202,32 @@ fn validate_dealings_payload(
...
+            );
+            return Err(InvalidDkgPayloadReason::InvalidEarlyNiDkgTranscripts.into());
+        }
+
+        info!(
+            log,
+            "Validated {} early remote DKG transcripts in data block payload at height {}",
+            dealings.transcripts_for_remote_subnets.len(),
+            parent.height.increment(),
+        );
```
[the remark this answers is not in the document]
eichhorl: This will happen very infrequently (subnet/engine creation/recovery)

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -114,12 +124,142 @@ fn create_data_payload(
...
+    for config in last_dkg_summary.configs.values() {
+        let dkg_id = config.dkg_id();
+        if completed.contains(dkg_id) {
+            // Skip DKGs that have already been completed
+            continue;
+        }
+        if let NiDkgTargetSubnet::Remote(target_id) = dkg_id.target_subnet {
+            remote_configs.entry(target_id).or_default().push(config);
+        }
+    }
```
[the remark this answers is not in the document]
eichhorl: The current ordering is deterministic across replicas

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -811,40 +976,71 @@ fn get_node_list(
...
         .chain(
             call_contexts
                 .reshare_chain_key_contexts
                 .iter()
-                .map(|(&callback_id, context)| (context.target_id, callback_id)),
+                .map(|(&callback_id, context)| (context.target_id, (1, callback_id))),
         )
-        .collect::<BTreeMap<NiDkgTargetId, CallbackId>>();
+        .collect()
+}
```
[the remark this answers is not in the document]
eichhorl: There should not be two contexts for the same target ID

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Done, thanks!

PR #8623, opened 2026-01-30
feat(consenus): [CON-1565] splitting cup share making/validating/aggregating

kpop-dfinity:
# Background

During subnet splitting, when replicas detect that a subnet should be split at a summary height `H`, they will immediately start producing CUP shares at height `H + dkg_interval_length`, according to the post-split subnet memberships, i.e.
half of the node will produce the shares as members of the original subnet, and the other half as the member of the new subnet.

Implement the logic of subnet splitting CUP share creation, validation, and aggregation.

# CUP shares creation

When a replica detects that a "subnet splitting" block (denoted by a new field, `subnet-splitting-status` in the `DkgSummary` struct) has been finalized and the state at the block's height has been committed, it will create a CUP in the following way:

1. Create a new summary block, on the fly, at the next CUP height, with dkg transcript taken from the registry;
2. Create a dummy random beacon at the next CUP height,
3. Check if node belongs to the high threshold committee, based on the transcript from the newly created summary block, if not, return None,
4. Create a CUP, with the block and random beacon created in steps 1. and 2., respectively, and sign it using the key material from the block.

# CUP shares validation

Validation works in the following way:

1. Check if the hash of the block in the CUP share matches the *recreated* summary block's hash
2. Check if the random beacon in the CUP share matches the *recreated* random beacon
3. Check if the signature can be verified using the key material from the *recreated* summary block
4. Check if the state hash matches our local state's hash

# Note

As the field `subnet_splitting_status` is not yet population, the new code branches are never executed.


rs/types/types/src/consensus/dkg.rs
```diff
@@ -172,9 +189,36 @@ pub struct DkgSummary {
...
+            configs,
+            current_transcripts,
+            next_transcripts,
+            transcripts_for_remote_subnets,
+            interval_length,
+            next_interval_length,
+            height,
+            initial_dkg_attempts,
+        } = self;
+
```
eichhorl, line 207: Don't we need to do something like this?

```suggestion

if let Some(status) = subnet_splitting_status {
    status.hash(state);
}
```

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -13,35 +13,59 @@
...
+        crypto_hash,
+        threshold_sig::ni_dkg::{NiDkgId, NiDkgTag, NiDkgTranscript},
     },
     replica_config::ReplicaConfig,
+    signature::ThresholdSignature,
 };
 use std::sync::Arc;
 
-/// CatchUpPackage maker is responsible for creating beacon shares
+/// [`CatchUpPackage`] maker is responsible for creating beacon shares
```
eichhorl, line 47:
```suggestion
/// [`CatchUpPackage`] maker is responsible for creating catch up package shares
```

rs/types/types/src/consensus/dkg.rs
```diff
@@ -576,17 +680,20 @@ pub enum DkgPayloadCreationError {
...
 #[derive(PartialEq, Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum InvalidDkgPayloadReason {
     CryptoError(CryptoError),
     DkgVerifyDealingError(DkgVerifyDealingError),
     MismatchedDkgSummary(DkgSummary, DkgSummary),
     MissingDkgConfigForDealing,
     DkgStartHeightDoesNotMatchParentBlock,
     DkgSummaryAtNonStartHeight(Height),
+    SubnetSplittingNotEnabled,
```
eichhorl, line 696: What does "not enabled" mean? I don't think these variants are used anywhere (yet?).

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -453,6 +457,22 @@ pub fn get_dkg_summary_from_cup_contents(
...
+        /*subnet_splitting_status=*/ None,
+    )
+}
+
+fn get_dkg_summary_from_cup_contents_with_subnet_splitting(
+    cup_contents: CatchUpPackageContents,
+    subnet_id: SubnetId,
+    registry: &dyn RegistryClient,
+    registry_version: RegistryVersion,
+    subnet_splitting_status: Option<SubnetSplittingStatus>,
```
eichhorl, line 475: I assume the `Option` part will disappear eventually? Do I understand correctly that calling this function with `SubnetSplittingStatus::Scheduled { .. }` doesn't make sense? Maybe we could change the argument to `Option<SubnetId>` where `None` corresponds to `SubnetSplittingStatus::NotScheduled` and Some(id) corresponds to `SubnetSplittingStatus::Done { id }`? 

Or we could use the new `CatchUpPackageType` enum from below

rs/consensus/mocks/src/lib.rs
```diff
@@ -105,6 +111,167 @@ pub struct Dependencies {
     pub canister_http_pool: Arc<RwLock<CanisterHttpPoolImpl>>,
 }
 
+pub struct DependenciesBuilder {
+    pool_config: ArtifactPoolConfig,
+    records: Vec<(u64, SubnetId, SubnetRecord)>,
```
eichhorl, line 116:
```suggestion
    subnet_records: Vec<(u64, SubnetId, SubnetRecord)>,
```

rs/consensus/mocks/src/lib.rs
```diff
@@ -105,6 +111,167 @@ pub struct Dependencies {
...
+            last_version = Some(version);
+        }
+
+        if let Some(last_version) = last_version {
+            add_subnet_list_record(
+                &registry_data_provider,
+                last_version,
+                Vec::from_iter(subnet_ids),
+            );
+        }
```
eichhorl, line 198: Why do we need to do this twice if `last_version != version` (see line 174)?

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -143,132 +169,399 @@ impl CatchUpPackageMaker {
...
+}
+
+fn get_new_subnet_id(
+    registry: &dyn RegistryClient,
+    summary_block: &Block,
+    node_id: NodeId,
+    source_subnet_id: SubnetId,
+    destination_subnet_id: SubnetId,
+) -> Result<SubnetId, String> {
+    let registry_version = summary_block.context.registry_version;
```
eichhorl, line 533: Could we add a comment why this is the correct registry version? Usually determining subnet membership is a bit more complicated.

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -985,6 +1006,40 @@ fn create_remote_dkg_config(
...
+    new_subnet_id: SubnetId,
+    registry: &dyn RegistryClient,
+    last_summary_block: &Block,
+) -> Result<DkgSummary, String> {
+    let last_summary = &last_summary_block.payload.as_ref().as_summary().dkg;
+    debug_assert!(matches!(
+        last_summary.subnet_splitting_status,
+        Some(SubnetSplittingStatus::Scheduled { .. })
+    ));
+    let registry_version = last_summary_block.context.registry_version;
```
eichhorl, line 1020: Could we add a comment explaining why this is the correct registry version? I think usually we take the version at which the cup contents were inserted into registry.

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -143,132 +169,399 @@ impl CatchUpPackageMaker {
...
+                // Skip if the state referenced by finalization tip has not caught up to
+                // this height. This is to increase the chance that states are available to
+                // validate payloads at the chain tip.
+                if pool.get_finalized_tip().context.certified_height < summary_height {
+                    return None;
+                }
+            }
+            CatchUpPackageType::PostSplit { .. } => {
+                // During subnet splitting we don't need to wait for the state at the height to be
+                // certified
```
eichhorl, line 197: I assume each half of the subnet will try to certify its (new) part of the state, which will fail because there aren't enough shares for either? I guess in that case we can't wait for the height to be certified even if we wanted to?
```suggestion
                // During subnet splitting we don't need to wait for the state at the height to be
                // certified, because ...
```

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -143,132 +169,399 @@ impl CatchUpPackageMaker {
...
+        // Skip if this node has already made a share
+        if pool
+            .get_catch_up_package_shares(self.get_cup_height(&start_block, cup_type))
+            .any(|share| share.signature.signer == self.replica_config.node_id)
+        {
+            return None;
+        }
+
+        let cup_block = self
+            .get_cup_block(start_block.clone(), cup_type)
```
eichhorl, line 268: Is this clone necessary?
```suggestion
            .get_cup_block(start_block, cup_type)
```

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -143,132 +169,399 @@ impl CatchUpPackageMaker {
...
+        }
+    }
+
+    fn get_high_dkg_id(
+        &self,
+        pool: &PoolReader<'_>,
+        cup_block: &Block,
+        cup_type: CatchUpPackageType,
+    ) -> Result<NiDkgId, String> {
+        // TODO: can we always take the transcript from the block?
```
eichhorl, line 374: Intuitively I would say yes, because the summary block is the first block of the current interval

rs/consensus/mocks/src/lib.rs
```diff
@@ -105,6 +111,167 @@ pub struct Dependencies {
...
+    }
+
+    pub fn build(self) -> Dependencies {
+        let time_source = FastForwardTimeSource::new();
+        let initial_registry_version = RegistryVersion::from(self.records[0].clone().0);
+        let registry_data_provider = Arc::new(ProtoRegistryDataProvider::new());
+        assert!(
+            !self.records.is_empty(),
+            "Cannot setup a registry without records."
+        );
```
pierugo-dfinity, line 168: I think `self.records[0]` two lines above would have panicked before reaching this assert.

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +188,80 @@ impl ShareAggregator {
...
+        let shares = pool
+            .get_catch_up_package_shares(block.height())
+            .collect::<Vec<_>>();
+
+        if shares.len() < threshold {
+            return Ok(None);
+        }
+
+        let cup_content =
+            CatchUpContent::from_share_content(shares[0].content.clone(), block.clone());
```
eichhorl, line 251: Is this clone necessary?
```suggestion
            CatchUpContent::from_share_content(shares[0].content.clone(), block);
```

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +188,80 @@ impl ShareAggregator {
         }
         Vec::new()
     }
+
+    fn aggregate_catch_up_package_shares_for_summary_block(
+        &self,
+        pool: &PoolReader<'_>,
+        summary_block: &Block,
```
eichhorl, line 195: Does this work?
```suggestion
        summary_block: Block,
```

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +188,80 @@ impl ShareAggregator {
...
+
+                (threshold, dkg_id, post_split_summary_block)
+            }
+        };
+
+        let shares = pool
+            .get_catch_up_package_shares(block.height())
+            .collect::<Vec<_>>();
+
+        if shares.len() < threshold {
```
eichhorl, line 246: I guess despite this check we might still not have enough shares below, because `get_catch_up_package_shares` above might return shares created by the other half of the subnet?

Or would the signature verification fail when validating those shares?

rs/consensus/src/consensus/validator.rs
```diff
@@ -1658,19 +1664,70 @@ impl Validator {
...
+
+                (post_split_block, post_split_random_beacon)
+            }
+            // We don't produce CUPs for the height at which a subnet splitting is happening.
+            CatchUpPackageType::PostSplit { .. } if dkg_summary.height == share_height => {
+                return Err(
+                    InvalidArtifactReason::InvalidHeightInSplittingCatchUpPackageShare.into(),
+                );
+            }
+            CatchUpPackageType::PostSplit { .. } | CatchUpPackageType::Normal => {
```
eichhorl, line 1708: When do we ever expect to see the `CatchUpPackageType::PostSplit { .. }` case here?

rs/consensus/src/consensus/catchup_package_maker.rs
eichhorl: One thing that comes to my mind: Just before the split we should have 3f+2 nodes on the subnet, where f+1 honest nodes compute state A and the second set of f+1 honest nodes compute the state B. Could all of the f malicious nodes decide to compute A, and thereby actually aggregate a full certification for A at the splitting height?  If so, would the nodes that computed B think there is a state divergence when receiving that certification?

pierugo-dfinity:
Closing in favour of https://github.com/dfinity/ic/pull/10980

PR #8653, opened 2026-02-03
test(ICRC_Ledger): DEFI-2646: Add canbench for get_blocks in ICRC ledger

mbjorkqvist:
Add a `canbench` benchmark for the `get_blocks` endpoint in the ICRC ledger.


rs/ledger_suite/icrc1/ledger/src/benches/benches_u256.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(not(feature = "get-blocks-disabled"))]
```
gregorydemay, line 1: nit: probably too late but this seems a bit weird because cargo features should be [additive](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)

> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features.

gregorydemay (approved):
Thanks @mbjorkqvist !

rs/ledger_suite/icrc1/ledger/src/benches/benches_u256.rs
mbjorkqvist: Good point - I see that, in addition to this `"get-blocks-disabled"` feature in the ledger that is not additive, we in the index have *two* features that are not additive: `"get_blocks_disabled"` (confusingly named with underscores instead of hyphens as in the ledger), and `"icrc3_disabled"`. I could think of the following follow-ups (I don't think it makes sense to address either as part of this ticket or PR):

- We could change the `get-blocks-disabled` features in the ledger and index to be additive (and name them `get_blocks_enabled`).
- We could do the same for the `icrc3_disabled` feature in the index.
- Regarding the index, there are some more details to consider: This `icrc3_disabled` feature is used for testing, to verify that the index canister's fallback to the legacy `get_blocks` endpoint works in case the ledger doesn't support ICRC-3. However, we no longer have a corresponding `icrc3_disabled` flag in the ledger, so to test the index against a ledger that doesn't support ICRC-3, we download a ledger wasm from the CDN. This ledger is build from commit https://github.com/dfinity/ic/commit/300dc603a92b5f70dae79229793c902f346af3cc and is referenced from [here](https://github.com/dfinity/ic/blob/b28870fe1008dcdbb6e19445296b06017c645a5d/MODULE.bazel#L572-L583). To avoid having to rely on such an old ledger wasm, we could consider using the `icrc3_test_ledger`, and add a corresponding `icrc3_disabled` feature to it.

I created [DEFI-2894](https://dfinity.atlassian.net/browse/DEFI-2894) to track potential follow-up work.

PR #8773, opened 2026-02-10
chore: run build-ic on stock runner

nmattia:
This moves the `build-ic` out of our custom (self-hosted) runners and onto namespace amd64 linux runners.

We try to make sure `build-ic.sh` is portable, and in order to surface issues early we run it on a non-custom runner, with no access to our internal infrastructure.

### DIFF
--- a/ci/container/build-ic.sh
+++ b/ci/container/build-ic.sh
@@ -109,8 +109,12 @@ export VERSION="$(git rev-parse HEAD)"
 
 BAZEL_TARGETS=()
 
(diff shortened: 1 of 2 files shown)


ci/container/build-ic.sh
```diff
@@ -42,17 +42,17 @@ EOF
 tp() { tput -T xterm "$@"; }
 echo_red() {
     tp setaf 1
-    echo "$1"
+    echo "$@"
```
nmattia: TODO: revert

PR #8848, opened 2026-02-13
test(crypto): CRP-2975 Poll as required in ingress verification system test

randombit:
This test has a bit of flakiness going on, the most common scenario seems to be a v3/v4 API call receiving a 202 instead of the expected 200.  However this can indeed occur if the replica is overloaded; normally v3/v4 API calls are sync but the call is still allowed to timeout with a 202. If that happens, use read state requests to poll until completion.


rs/tests/crypto/ingress_verification_test.rs
fspreiss: IIUC, for completeness and robustness, we should also call `await_ingress_via_read_state` here.

Currently the logic to call `await_ingress_via_read_state` would be unnecessary because `perform_update_with_expiry` is only used once where an error is expected afterwards, but this may change in the future.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1497,49 @@ async fn send_request<C: serde::ser::Serialize>(
...
+            "read_state",
+            content,
+            sender.public_key_der(),
+            None,
+            signature,
+        )
+        .await;
+
+        if response.status() == 200 {
+            return response;
```
fspreiss: A 200 here just means that the read_state request itself was successful, but it doesn't tell us something about the status of the request with id `request_id`, so we have to look into the [status](https://docs.internetcomputer.org/references/ic-interface-spec#state-tree-request-status):
* if the request _status_ is `received` or `processing` we have to keep polling
* if the request _status_ is `replied` or `reject` (or `done`, but that is unlikely) we can stop polling, although it is not fully clear to me yet if/how we can construct the `ReplicaResponse` then because we don't have a (HTTP) `StatusCode` that we need to construct it. Maybe we have to distinguish between sync and async responses by means of some enum!?.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1497,49 @@ async fn send_request<C: serde::ser::Serialize>(
...
+            test,
+            "read_state",
+            content,
+            sender.public_key_der(),
+            None,
+            signature,
+        )
+        .await;
+
+        if response.status() == 200 {
```
fspreiss: If the status is 4xx (except maybe 429: Too Many Requests) or 5xx there is no point in continuing to poll for up to 30 seconds, so I guess we can just fail early then.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1499,105 @@ async fn send_request<C: serde::ser::Serialize>(
...
+                status, response.body
+            );
+        }
+
+        if status == 200 {
+            // A 200 means the read_state HTTP request succeeded, but we must
+            // inspect the certificate to determine the actual request status.
+            match request_status_from_read_state_response(&response, &request_id) {
+                Some(s) if s == "replied" || s == "rejected" || s == "done" => {
+                    return response;
```
fspreiss: Is returning the response here the right thing to do? I was somewhat expecting that the method returns the response for the original update call (e.g., the content of the certificate's `/request_status/<request_id>/reply` if status is `replied`), rather than the response for the read-state request.

But that doesn't fit with how the `ReplicaResponse` works because it assumes there are status-codes involved, so I'm wondering if `ReplicaResponse` should be an enum with variants for sync and async responses.

randombit: No I think it's not

One trouble is that this test is only slightly flaky - so far on my devenv I've not been able to make a retry occur so the logic in this PR is basically completely untested. I think I'm going to have to make the update call slower somehow so that it times out in a semi-consistent way. (How exactly to do this, so far unclear...)

randombit: Ah, now that I am able to replicate this somewhat reliably - this does seem correct to me. The response here contains exactly the response we would have received had the synchronous call not timed out, and returning that response here all of the tests pass.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -901,7 +912,14 @@ pub fn requests_to_mgmt_canister_with_delegations(env: TestEnv) {
                         )
                         .await;
 
-                        assert_eq!(response.status(), 200);
```
kpop-dfinity, line 910: I guess we don't want this assertion anymore?

rs/tests/crypto/ingress_verification_test.rs
kpop-dfinity: nit: Shall we remove/update this comment? We do have a retry loop now

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1542,7 +1660,14 @@ async fn perform_update_call_with_delegations(
...
     )
-    .await
+    .await;
+
+    // v3/v4 sync calls may return 202 under load; fall back to read_state polling
+    if api_ver >= 3 && response.status() == 202 {
+        return await_ingress_via_read_state(test, sender, request_id).await;
+    }
+
+    response
```
kpop-dfinity: nit: since the pattern 

```
   let response = send_request(...);
   if api_ver >= 3 && response.status() == 202 {
       await_ingress_via_read_state(test, sender, request_id)
   } else {
       response
   }
```

appears multiple times in this file, maybe we could export the logic to a function and use it throughout the file?

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1499,105 @@ async fn send_request<C: serde::ser::Serialize>(
...
+                "read_state request failed with HTTP {}: {:?}",
+                status, response.body
+            );
+        }
+
+        if status == 200 {
+            // A 200 means the read_state HTTP request succeeded, but we must
+            // inspect the certificate to determine the actual request status.
+            match request_status_from_read_state_response(&response, &request_id) {
+                Some(s) if s == "replied" || s == "rejected" || s == "done" => {
```
kpop-dfinity: nit: I think this should work
```suggestion
                Some("replied" | "rejected" | "done") => {
```

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1499,105 @@ async fn send_request<C: serde::ser::Serialize>(
...
+                    return response;
+                }
+                Some(s) => {
+                    assert!(
+                        s == "received" || s == "processing",
+                        "Unexpected request status: {s}"
+                    );
+                }
```
kpop-dfinity: nit: maybe?
```suggestion
                Some("received" | "processing") => {
                    // Ingress message has not been processed yet; keep polling
                }
                Some(s) => {
                    panic!("Unexpected request status: {s}");
                }
```

kpop-dfinity (approved):
Left some nits

rs/tests/crypto/ingress_verification_test.rs
fspreiss: As discussed yesterday in person, what is relevant here is the [Request status](https://docs.internetcomputer.org/references/ic-interface-spec#state-tree-request-status) section in the IC spec, so we have to distinguish between sync and async responses.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1497,15 +1551,36 @@ async fn send_request<C: serde::ser::Serialize>(
...
 }
 
+/// Parsed ingress request status from a read_state certificate.
+enum IngressStatus {
+    /// The request was executed and a reply is available.
+    Replied {
+        /// Raw CBOR-encoded certificate bytes (the `certificate` field from
+        /// the read_state response). This is the same blob that the v3/v4
+        /// sync call endpoint would have embedded in its response.
+        certificate_cbor: Vec<u8>,
```
fspreiss: nit: maybe call this `reply` as per the [request status](https://docs.internetcomputer.org/references/ic-interface-spec#state-tree-request-status) section in the spec.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1551,209 @@ async fn send_request<C: serde::ser::Serialize>(
...
+                }
+                other => panic!("Missing or invalid reject_code: {other:?}"),
+            };
+
+            let msg_path: [&[u8]; 3] = [b"request_status", request_id.as_ref(), b"reject_message"];
+            let reject_message = match certificate.tree.lookup(&msg_path) {
+                LookupStatus::Found(MixedHashTree::Leaf(bytes)) => {
+                    String::from_utf8(bytes.clone()).expect("Invalid UTF-8 in reject_message")
+                }
+                // reject_message is optional per the spec
```
fspreiss: IIUC, the `reject_message` is mandatory if the status is `rejected`. What is optional is the `error_code`.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1481,6 +1551,209 @@ async fn send_request<C: serde::ser::Serialize>(
...
+    let status_path: [&[u8]; 3] = [b"request_status", request_id.as_ref(), b"status"];
+
+    let status_str = match certificate.tree.lookup(&status_path) {
+        LookupStatus::Found(MixedHashTree::Leaf(status_bytes)) => {
+            String::from_utf8(status_bytes.clone()).expect("Invalid UTF-8 in request status")
+        }
+        LookupStatus::Found(other) => {
+            panic!("Request status is not a leaf node: {other:?}");
+        }
+        LookupStatus::Absent | LookupStatus::Unknown => return None,
```
fspreiss: IIUC, this cannot/shall not be absent or unknown, so I'd find panicking here acceptable.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1583,10 +1797,15 @@ async fn perform_update_call_with_delegations(
         update: HttpCanisterUpdate {
             canister_id: Blob(test.canister_id.get().as_slice().to_vec()),
             method_name: "update".to_string(),
-            arg: Blob(wasm().reply_data(b"update_reply").build()),
+            arg: Blob(
+                wasm()
+                    .instruction_counter_is_at_least(UPDATE_PAYLOAD_INSTRUCTIONS)
```
fspreiss, line 2131: Is this done so that the call actually takes some time?

In any case, I see `instruction_counter_is_at_least` uses at a few (3) places where `wasm()` is used, but not at all places? Why were exactly these three chosen?

rs/tests/crypto/ingress_verification_test.rs
fspreiss, line 1911: nit: We clone `request_id` both here and at the call sites: I think it would make sense to turn the parameter into a ref (`request_id: &MessageId,`) as then we don't need to clone both at the call site and here:
```
paths: vec![vec!["request_status".into(), request_id.into()].into()],
```

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -855,13 +860,17 @@ pub fn requests_to_mgmt_canister_with_delegations(env: TestEnv) {
...
+                                .await
                         } else {
                             response
                         };
 
                     if include_mgmt_canister_id {
                         response.expect_update_ok(api_ver);
+                        if api_ver >= 3 {
+                            response.verify_update_certificate(&request_id);
+                        }
```
fspreiss: This check is done "manually" here after `expect_update_ok`, but it very much feels like this check should directly be part of `expect_update_ok`.

IIUC, the check is done manually here because the response wasn't produced with `perform_update*`, where the `perform_update*` methods have this check baked in at the very end.

So, wouldn't it make more sense to embed the `verify_update_certificate` check into `expect_update_ok` and remove it from the `perform_update*` methods?

I guess the difficulty with this is that `expect_update_ok` doesn't have the needed `request_id`: one option to solve that would be to add it as a parameter (which would require that the `perform_update*` methods also return the calculated request ID); another option (maybe simpler?) could be that the `request_id` is directly baked into the `ReplicaResponse` struct.

WDYT?

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -912,12 +921,15 @@ pub fn requests_to_mgmt_canister_with_delegations(env: TestEnv) {
...
                             await_ingress_via_read_state(&test_info, sender, request_id.clone())
-                                .await;
+                                .await
                         } else {
                             assert_eq!(response.status(), 200);
-                        }
+                            response
+                        };
+
+                        response.verify_update_certificate(&request_id);
```
fspreiss: If `expect_update_ok` would embed the `verify_update_certificate`, then here we could just call `expect_update_ok`.

The same holds for all other calls to `verify_update_certificate`: ideally we never call this directly, but only indirectly via `expect_update_ok` (calling it manually/directly seems quite brittle).

rs/tests/crypto/ingress_verification_test.rs
fspreiss: The above issue was fixed in https://github.com/dfinity/ic/pull/8848/changes/8638be1f101a3a599449284f3f2d1ad272e1ffc0: Thanks, @randombit. I'm resolving this comment.

rs/tests/crypto/ingress_verification_test.rs
fspreiss: The above issue was fixed in https://github.com/dfinity/ic/pull/8848/changes/8638be1f101a3a599449284f3f2d1ad272e1ffc0: Thanks, @randombit. I'm resolving this comment.

rs/tests/crypto/ingress_verification_test.rs
randombit: We do not. The issue is this can fail (with a 202 aka try again later) and then we have to issue a read state request to find the actual result of the message.

rs/tests/crypto/ingress_verification_test.rs
randombit: renamed

rs/tests/crypto/ingress_verification_test.rs
randombit: That does seem correct, changed to panic here

rs/tests/crypto/ingress_verification_test.rs
randombit: Makes sense, changed to panic here

rs/tests/crypto/ingress_verification_test.rs
randombit: This is extremely arbitrary and ad hoc. Basically if a delay is added to all calls, then either the cycles count is too small (in which case the test never sees the 202 retry behavior that we're trying to test) or it is too large (in which case the entire test ends up timing out).

rs/tests/crypto/ingress_verification_test.rs
randombit: Not ideal! Fixed

fspreiss:
Reminder as discussed offline: also remove the Bazel manual flag so that the test is run by CI.

rs/tests/crypto/ingress_verification_test.rs
randombit: done

rs/tests/crypto/ingress_verification_test.rs
randombit: :+1:

rs/tests/crypto/ingress_verification_test.rs
randombit: done

rs/tests/crypto/ingress_verification_test.rs
randombit: Significantly rewritten now

randombit:
@fspreiss All review comments should be addressed, haven't seen any flakes so I think this is ready to go

fspreiss (approved):

PR #8884, opened 2026-02-16
chore: Get rid of `CanisterLog::delta_log_sizes`

alin-at-dfinity:
It introduces a lot of overhead and complexity into the scheduler, particularly with regard to `Arc<CanisterState>` and active-canisters-only scheduling.

This change replaces the `delta_log_sizes` ring buffers (accumulated by `CanisterLog` and friends; then observed and dropped by `SchedulerImpl::execute_round()` as part of round finalization) with passing the relevant metric (now hosted by `HypervisorMetrics`) into `SystemStateModifications::apply_changes()` to directly instrument each `append_delta_log()` call. In the meantime, there's now also a `canister_log_retention_seconds` metric, which was moved into `ReplicatedStateMetrics`.


mraszyk (approved):
LGTM

Any reason why the delta sizes were stored in a ring-buffer, @maksymar ?

schneiderstefan (approved):

PR #8885, opened 2026-02-16
perf: [DSM-103] Only schedule active canisters

alin-at-dfinity:
Only schedule canisters that have something to execute; or that have negative accumulated priority and are working through their "debt".


alin-at-dfinity:
All the relevant bits have been merged. Closing.

PR #9316, opened 2026-03-11
test(node): Skip GuestOS SSH check when TEE is enabled in bare metal tests

andrewbattat:
This PR fixes a [failure](https://github.com/dfinity/ic/actions/runs/22934706555/job/66683864350) in the nested bare-metal TEE tests introduced by https://github.com/dfinity/ic/pull/9291

`setup-ssh-user-keys` intentionally does not copy `accounts_ssh_authorized_keys` when TEE is active, so GuestOS SSH is unavailable by design.


rs/tests/nested/src/lib.rs
```diff
@@ -126,22 +126,30 @@ pub fn registration(env: TestEnv) {
...
         info!(
             logger,
-            "Node {}: systemctl list-units --failed:\n{}", node.node_id, failed_units
-        );
-        assert!(
-            failed_units.trim().is_empty(),
-            "Node {} has failed systemd units:\n{}",
-            node.node_id,
-            failed_units
+            "Skipping GuestOS SSH systemd check because TRUSTED_EXECUTION_ENVIRONMENT is enabled."
```
andrewbattat, line 133: Alternative fix if we want to support SSH access on dev SEV nodes:
- split `setup-ssh-user-keys-guestos.sh` into dev and prod scripts, and only disable SSH key setup when TEE_active + prod, while allowing it for TEE_active + dev.

This matches how config_tool currently works—we build a prod and dev version of the tool, _and the blessed prod measurement should only ever use the prod tool_.

PR #9342, opened 2026-03-13
chore(firewall): remove obsolete default firewall config path

pierugo-dfinity:
The orchestrator's firewall uses two orthogonal ways to detect that it is running in a testnet: [whether or not the config file is the default value](https://sourcegraph.com/r/github.com/dfinity/ic@18219373852c8818f22ea5854c765f7491d94efd/-/blob/rs/orchestrator/src/firewall.rs?L69-79) and [whether or not there are rules in the registry](https://sourcegraph.com/r/github.com/dfinity/ic@18219373852c8818f22ea5854c765f7491d94efd/-/blob/rs/orchestrator/src/firewall.rs?L375-383). This is confusing and error-prone.

In practice, we now always use the second way since the config file is [set](https://sourcegraph.com/r/github.com/dfinity/ic@18219373852c8818f22ea5854c765f7491d94efd/-/blob/rs/ic_os/config/tool/templates/ic.json5.template?L186) in the `ic.json5` template. Indeed, [no testnet logged](https://kibana.testnet.dfinity.network/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:'2026-01-09T09:44:46.557Z',to:now))&_a=(columns:!(),filters:!(),hideChart:!f,index:testnet-vector-push,interval:auto,query:(language:kuery,query:'MESSAGE:%20%22Firewall%20configuration%20not%20found.%20Orchestrator%20does%20not%20update%20firewall%20rules.%22'),sort:!(!(timestamp,desc)))) the first way in the past months. This PR thus removes the ability to disable the firewall.


r-birkner (approved):
👌

eichhorl (approved):

PR #9363, opened 2026-03-13
perf(crypto): CRP-2907 Improve performance of Ed25519 batch verification

randombit:
Previously the implementation took a random seed from the vault but this imposed an overhead of 50 to 100 microseconds for each batch verification. Instead add a new RNG that can be used to generate batch verification nonces without having to perform a vault RPC. Compared to performance on master, the cost of batch verification is reduced by about 20%.


rs/crypto/src/lib.rs
```diff
@@ -36,9 +36,18 @@ use ic_protobuf::registry::crypto::v1::{PublicKey as PublicKeyProto, X509PublicK
...
+use rand::{CryptoRng, RngCore, rngs::OsRng};
 use std::fmt;
 use std::sync::Arc;
 
+/// A trait combining [`RngCore`] and [`CryptoRng`], allowing use as a trait object.
+///
+/// Rust does not allow multiple non-auto traits in a trait object (error E0225)
+/// so this supertrait is needed. `rand` 0.9 adds exactly this same trait, so this
+/// definition can be removed when `rand` is updated.
+pub trait CryptoRngCore: RngCore + CryptoRng {}
```
fspreiss: While we haven't upgraded to rand 0.9+, I'd prefer this to be `Rng + CryptoRng`. IIUC, the purpose of `RngCore` is to _implement_ (see, e.g., [this](https://docs.rs/rand_core/0.10.0/rand_core/)) an RNG (like ReproducibleRng).

fspreiss: Actually, we have almost exact same thing already: [CryptoComponentRng](https://github.com/dfinity/ic/blob/40ad82dbd8da033dadc089a60ff12d2012f271e5/rs/crypto/temp_crypto/src/lib.rs#L24-L26).

It lives in `ic-crypto-temp-crypto` though, so we'd have to move it here if we want to use it. Moving it here could actually make sense (even if we upgrade to rand 0.9/0.10) if we want `Send + Sync` (and `'static` if needed) to be part of the trait definition. The advantage of having `Send + Sync` (and `'static`) be part of the trait definition is that we don't have to repeat it over and over again.

rs/crypto/src/lib.rs
```diff
@@ -55,6 +64,7 @@ pub struct CryptoComponentImpl<C: CryptoServiceProvider> {
     lockable_threshold_sig_data_store: LockableThresholdSigDataStore,
     vault: Arc<dyn CspVault>,
     csp: C,
+    csprng: CspRwLock<Box<dyn CryptoRngCore + Send + Sync>>,
```
fspreiss: Ideally we would avoid dynamic dispatch here by using generics, but that would cause a bit of a ripple effect, so maybe it is best to start with what we have now and potentially improve things later.

rs/crypto/temp_crypto/src/lib.rs
```diff
@@ -445,6 +445,7 @@ pub mod internal {
                 node_id,
                 metrics,
                 Some(time_source),
+                Box::new(OsRng),
```
fspreiss: If we use `OsRng` here, then the `TempCryptoComponent` will no longer be fully reproducible. We have to make use of the builder's `self.rng` here: probably create a ChaCha RNG seeded from self.rng.

rs/crypto/tests/basic_sig.rs
```diff
@@ -195,6 +195,7 @@ pub fn crypto_component_with_vault(
         NODE_1,
         Arc::new(CryptoMetrics::none()),
         None,
+        Box::new(ReproducibleRng::new()),
```
fspreiss: Ideally, `crypto_component_with_vault` would take an RNG as parameter and use that here, so that the tests that use `crypto_component_with_vault` can instantiate an RNG once at the beginning of the test. We can also do that cleanup in a separate PR.

rs/crypto/tests/canister_threshold_idkg.rs
```diff
@@ -323,6 +323,7 @@ mod create_dealing {
                 env.nodes.random_dealer(&params, rng).id(),
                 Arc::new(CryptoMetrics::none()),
                 None,
+                Box::new(ReproducibleRng::new()),
```
fspreiss: We can use the existing RNG: `Box::new(rng)`

rs/crypto/tests/canister_threshold_idkg.rs
```diff
@@ -2086,6 +2087,7 @@ mod load_transcript_with_openings {
                 node_id_not_in_receivers,
                 metrics,
                 None,
+                Box::new(ReproducibleRng::new()),
```
fspreiss: Let's use the existing RNG from line 2057 here.

rs/crypto/tests/canister_threshold_idkg.rs
```diff
@@ -2898,6 +2900,7 @@ mod verify_dealing_private {
                 node_id,
                 crypto_metrics,
                 time_source,
+                Box::new(ReproducibleRng::new()),
```
fspreiss: Let's use the existing RNG parameter (line 2850) here instead of creating a new one.

rs/crypto/src/common/test_utils/crypto_component.rs
```diff
@@ -21,6 +22,7 @@ pub fn crypto_component_with_csp(
         node_test_id(NODE_ID),
         Arc::new(CryptoMetrics::none()),
         None,
+        Box::new(ReproducibleRng::new()),
```
fspreiss: Same as https://github.com/dfinity/ic/pull/9363/changes#r2947676262.

rs/crypto/src/common/test_utils/crypto_component.rs
```diff
@@ -37,5 +39,6 @@ pub fn crypto_component_with_csp_and_vault(
...
```
fspreiss: Same as https://github.com/dfinity/ic/pull/9363/changes#r2947676262.

rs/crypto/src/keygen/tests.rs
```diff
@@ -1348,6 +1349,7 @@ mod rotate_idkg_dealing_encryption_keys {
             node_id(),
             Arc::new(CryptoMetrics::none()),
             Some(Arc::clone(&time_source) as Arc<_>),
+            Box::new(ReproducibleRng::new()),
```
fspreiss: Maybe create this RNG at the beginning of the test, so we can easily re-use it whenever we need an RNG again somewhere in this test in the future.

rs/crypto/src/keygen/tests.rs
```diff
@@ -1968,6 +1970,7 @@ impl SetupBuilder {
             node_id(),
             crypto_metrics,
             Some(Arc::clone(&time_source) as Arc<_>),
+            Box::new(ReproducibleRng::new()),
```
fspreiss: Ideally, the SetupBuilder would have a field for the RNG, which is by default instantiated with a reproducible RNG but can be customized with a `with_rng` method. How about doing this in a follow-up cleanup PR?

rs/crypto/benches/basic_sig.rs
```diff
@@ -169,10 +171,92 @@ fn crypto_basicsig_verifybypubkey<M: Measurement, R: Rng + CryptoRng>(
...
+
+    // Generate batch_size distinct signers, each with their own key and signature.
+    let mut node_ids = Vec::with_capacity(batch_size);
+    let mut signatures = Vec::with_capacity(batch_size);
+    let bytes_to_sign = {
+        let mut buf = vec![];
+        buf.extend_from_slice(&msg.domain);
+        buf.extend_from_slice(&msg.signed_bytes_without_domain);
+        buf
+    };
```
fspreiss: Or just `msg.as_signed_bytes()` with `use ic_types::crypto::Signable;`

Same in `signature_from_random_keypair`.

rs/crypto/benches/basic_sig.rs
```diff
@@ -169,10 +171,92 @@ fn crypto_basicsig_verifybypubkey<M: Measurement, R: Rng + CryptoRng>(
...
+        .with_rng(ChaCha20Rng::from_seed(rng.r#gen()))
+        .with_remote_vault()
+        .build();
+    let node_pubkeys = crypto.current_node_public_keys().unwrap();
+    add_node_signing_pubkey_to_registry(
+        NODE_1,
+        &node_pubkeys.node_signing_public_key.unwrap().key_value,
+        &registry,
+        &registry_data,
+    );
```
fspreiss: Given that `crypto` only ever combines and verifies batches, the following should be sufficient:
```
let crypto = TempCryptoComponent::builder()
        .with_registry(Arc::clone(&registry) as Arc<_>)
        .with_node_id(NODE_1)
        .with_rng(ChaCha20Rng::from_seed(rng.r#gen()))
        .build();
```

rs/crypto/src/lib.rs
randombit: We can't use `Rand` at least as this is currently being used, because `Rand` is not [dyn compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility) due to containing functions which are generic over other types:

```
error[E0038]: the trait `CryptoComponentRng` is not dyn compatible
   --> rs/crypto/src/sign/canister_threshold_sig/idkg/transcript.rs:42:28
    |
 42 |     csprng: &CspRwLock<Box<dyn CryptoComponentRng + Send + Sync>>,
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `CryptoComponentRng` is not dyn compatible
    |
note: for a trait to be dyn compatible it needs to allow building a vtable
      for more information, visit <https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility>
   --> /home/jack/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rand-0.8.5/src/rng.rs:93:8
    |
 93 |     fn gen<T>(&mut self) -> T
    |        ^^^ ...because method `gen` has generic type parameters
...
```

I'm not really sure how to proceed here.

I guess I'll suggest an alternate approach that would sidestep this issue entirely: entirely abandon the idea of making the crypto component RNG type generic. Instead it's always a `ChaCha20Rng` which is seeded either using the OS RNG or some data pulled from a reproducible RNG, depending on how things are set up. WDYT

fspreiss: I see! Interesting, I didn't know about the dyn compatibility issue! Then your previous approach of using 
`RngCore + CryptoRng` is fine! (`Rng` and `RngCore` are practially interchangable anyway as `Rng: RngCore` and `Rng` is automatically implemented for every type implementing `RngCore`, its just that I find `RngCore` should only be used over `Rng` when its really necessary, which would be the case here).

Using an alternative approach could also be fine. The downside of using `ChaCha20Rng` directly is the lack of forward secrecy, I guess. For our Ed25519 batch verification use case here this might not matter. However, in the future we may have use cases where this then _does_ matter, and then we either (a) have to switch to a different approach, or (b) forget to switch, which could lead to a security issue. Could the automatically re-seeded `ThreadRng` be a compromise/solution? Ah...no, because it is not a `SeedableRng`.

Two other possible approaches that I can think of:
1. Use generics (not dynamic dispatch), but reduce the blast radius (i.e., reduce the generic propagation) by adding a _default type parameter_ (`struct CryptoComponentImpl<C: CryptoServiceProvider, R: CryptoComponentRng = OsRng>`). Not sure how much this reduces the impact in practise though.

2. Avoid generics altogether but still allow for different RNGs to be used, by using _enum dispatching_:
   ```
    struct CryptoComponentImpl {
        csprng: CspRwLock<CryptoComponentRng>,
        ...
    }
    enum CryptoComponentRng {
        Os(OsRng),
        Reproducible(ReproducibleRng),
    }
    impl RngCore for CryptoComponentRng {
        fn next_u32(&mut self) -> u32 {
            match self {
                Self::Os(r) => r.next_u32(),
                Self::Reproducible(r) => r.next_u32(),
            }
        }
        fn next_u64(&mut self) -> u64 {
            match self {
                Self::Os(r) => r.next_u64(),
                Self::Reproducible(r) => r.next_u64(),
            }
        }
        fn fill_bytes(&mut self, dest: &mut [u8]) {
            match self {
                Self::Os(r) => r.fill_bytes(dest),
                Self::Reproducible(r) => r.fill_bytes(dest),
            }
        }
        fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand::Error> {
            match self {
                Self::Os(r) => r.try_fill_bytes(dest),
                Self::Reproducible(r) => r.try_fill_bytes(dest),
            }
        }
    }
    impl CryptoRng for CryptoComponentRng {}
   ```
   With this approach we could also (e.g., later) easily switch to a `ThreadRng` if we need things to be faster.

Not really sure at this point what the best approach for us is now. Merely thinking out loud and pointing out options.

randombit:
Todo, check impact of Use generics (not dynamic dispatch), but reduce the blast radius (i.e., reduce the generic propagation) by adding a default type parameter (struct CryptoComponentImpl<C: CryptoServiceProvider, R: CryptoComponentRng = OsRng>). Not sure how much this reduces the impact in practise though.

Edit: done

rs/crypto/temp_crypto/src/lib.rs
```diff
@@ -445,6 +445,7 @@ pub mod internal {
                 node_id,
                 metrics,
                 Some(time_source),
+                OsRng,
```
randombit: This needs to be fixed

randombit: done

randombit:
@fspreiss This is I think ready to go, handing over to you now :saluting_face:

rs/crypto/src/lib.rs
```diff
@@ -36,9 +36,15 @@ use ic_protobuf::registry::crypto::v1::{PublicKey as PublicKeyProto, X509PublicK
...
 use ic_types::{NodeId, RegistryVersion};
 use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+use rand::rngs::OsRng;
+use rand::{CryptoRng, Rng};
 use std::fmt;
 use std::sync::Arc;
 
+/// A supertrait collecting traits required for an RNG used in the [`CryptoComponentImpl`].
+pub trait CryptoComponentRng: Rng + CryptoRng + 'static + Send + Sync {}
+impl<T: Rng + CryptoRng + 'static + Send + Sync> CryptoComponentRng for T {}
```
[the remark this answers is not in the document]
eichhorl: Dropping `Sync` from `CryptoComponentRng` would break `CspRwLock<R>: Sync` (parking_lot's `RwLock<T>` is `Sync `only when `T: Send + Sync`), so the bound is required.

rs/crypto/src/lib.rs
```diff
@@ -213,6 +222,7 @@ impl CryptoComponentImpl<Csp> {
             lockable_threshold_sig_data_store: LockableThresholdSigDataStore::new(),
             csp,
             vault,
+            csprng: CspRwLock::new_for_rng(OsRng, Arc::clone(&metrics)),
```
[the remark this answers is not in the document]
eichhorl: They are all `CspRwLock`s, so it makes sense that they are called `csprng`

rs/crypto/benches/basic_sig.rs
eichhorl: https://github.com/dfinity/ic/pull/9363/changes/0ac427141fe1005c6d134a72018ccafed1ab6581

rs/crypto/benches/basic_sig.rs
eichhorl: https://github.com/dfinity/ic/pull/9363/changes/29f1bb3b86e2c65f9d62b6f8c378bdffbb21969b

rs/crypto/src/lib.rs
eichhorl: https://github.com/dfinity/ic/pull/9363/changes/93f1e23571a30f453492d9b461706472cc331004#diff-7cf78f29f866560611462d0b0aa533a5c1c1b2d86f0e7c114047a75329da4a7dR60

rs/crypto/temp_crypto/src/lib.rs
eichhorl: https://github.com/dfinity/ic/pull/9363/changes/93f1e23571a30f453492d9b461706472cc331004#diff-ba1dfedaa0e9872de4261ff2cc15218b23432814a0bc01a03d86d3714da4a8deR440

rs/crypto/src/common/test_utils/crypto_component.rs
eichhorl: We now pass in a seed which is generated by the initial `ReproducibleRng`

rs/crypto/src/keygen/tests.rs
eichhorl: We now use a fixed seed instead

rs/crypto/tests/canister_threshold_idkg.rs
eichhorl: Done

rs/crypto/tests/canister_threshold_idkg.rs
eichhorl: Done

rs/crypto/tests/canister_threshold_idkg.rs
eichhorl: Done

rs/crypto/tests/basic_sig.rs
eichhorl: We're using a fixed seed for now, and will consider the cleanup in a separate PR

rs/crypto/src/keygen/tests.rs
eichhorl: We're using a fixed seed for now, and will consider the cleanup in different PR

eichhorl: Actually added it in this PR: https://github.com/dfinity/ic/pull/9363/commits/d5585dcdf8c55fb59b246b70853c4b5d680e8917

rs/crypto/src/keygen/tests.rs
eichhorl: Actually added it in this PR: https://github.com/dfinity/ic/pull/9363/commits/d5585dcdf8c55fb59b246b70853c4b5d680e8917

rs/crypto/src/common/test_utils/crypto_component.rs
eichhorl: Actually added it in this PR: https://github.com/dfinity/ic/pull/9363/commits/d5585dcdf8c55fb59b246b70853c4b5d680e8917

rs/crypto/tests/basic_sig.rs
eichhorl: Actually added it in this PR: https://github.com/dfinity/ic/pull/9363/commits/d5585dcdf8c55fb59b246b70853c4b5d680e8917

eichhorl (approved):

PR #9508, opened 2026-03-20
chore(ic): add flexible message ordering to StateMachine tests

venkkatesh-sekar:
Adds flexible message ordering to `StateMachine` for deterministic control over canister execution order in tests. Users specify an explicit sequence of tasks (`Ingress`, `Request`, `Response`, `Heartbeat`, `Timer`) and
  the framework executes them one-per-round.

  ## Production code changes

  - **`rs/config/src/subnet_config.rs`**: Added `subnet_messages_per_round_instruction_limit` field to `SchedulerConfig`. Moved `SUBNET_MESSAGES_LIMIT_FRACTION` constant here as `pub`.
  - **`rs/interfaces/src/execution_environment.rs`**: Added `config: &SchedulerConfig` to `Scheduler::execute_round` and `checkpoint_round_with_no_execution`. Added `ic-config` dependency to `ic-interfaces`.
  - **`rs/execution_environment/src/scheduler.rs`**: `SchedulerImpl` no longer stores `SchedulerConfig` — all helper methods receive config as a parameter from `execute_round`. Only `scheduler_cores` is stored for thread
  pool initialization. Uses the config field for subnet instruction budget instead of hardcoded formula.
  - **`rs/execution_environment/src/lib.rs`**: Removed `scheduler_config` field from `ExecutionServices`. `SchedulerImpl::new` takes `scheduler_cores: usize` instead of `SchedulerConfig`.
  - **`rs/execution_environment/src/scheduler/round_schedule.rs`**: Relaxed 3 debug assertions to tolerate `scheduler_cores=1`.
  - **`rs/execution_environment/src/scheduler/test_utilities.rs`**: Uses `scheduler_config.subnet_messages_per_round_instruction_limit` instead of recomputing from fraction.
  - **`rs/messaging/src/state_machine.rs`**: `StateMachineImpl` stores `SchedulerConfig` and passes it to `Scheduler::execute_round` and `checkpoint_round_with_no_execution`. Added `induct_loopback_stream` to `StateMachine`
  trait.
  - **`rs/messaging/src/message_routing.rs`**: `BatchProcessorImpl::new`, `MessageRoutingImpl::new`, and `SyncMessageRouting::new` accept and forward `SchedulerConfig`.
  - **`rs/messaging/src/routing/demux.rs`**: Added `induct_loopback_stream` to `Demux` trait, implemented via `process_stream_slices`.
  - **`rs/replica/src/setup_ic_stack.rs`**, **`rs/replay/src/player.rs`**, **`rs/determinism_test/src/setup.rs`**: Pass `scheduler_config` to `MessageRoutingImpl::new`.
  - **`rs/replicated_state/src/canister_state/queues.rs`** and **`queues/input_schedule.rs`**: Added test-only helpers `set_next_input_source` and `rotate_local_sender_to_front` to `InputScheduleTesting` trait.

  ## Test code changes

  - **`rs/state_machine_tests/src/lib.rs`**: Added `FlexibleMessageOrdering` and `FlexibleOrderingScheduler` test infrastructure. `FlexibleOrderingScheduler` holds an `Arc<RwLock<SchedulerConfig>>` shared with
  `FlexibleMessageOrdering` to override config per-round via `execute_round`.
  - **`rs/state_machine_tests/tests/flexible_ordering.rs`**: 22 tests exercising flexible message ordering (ingress, inter-canister calls, heartbeats, timers, subnet messages, DTS, self-calls, impossible orderings).

  ## Cautions

  - `scheduler_cores=1` triggers non-fatal debug assertions in the scheduler (logged, not panicked). Acceptable for testing since no canister uses non-zero compute allocation.
  - `FlexibleOrderingScheduler` zeroes all canister priorities before/after each boosted round — test-only state manipulation.
  - `set_next_scheduled_method` (relaxed mode) mutates canister execution state to control what `initialize_inner_round` adds — test infrastructure only.

  ## Not implemented

  - **Multi-subnet / XNet**: single subnet only.
  - **Direct `SchedulerRoundLimits` control**: approximated via config-level budget limits instead.


rs/state_machine_tests/src/lib.rs
```diff
@@ -234,6 +234,93 @@ impl Verifier for FakeVerifier {
...
+            round_summary,
+            current_round_type,
+            registry_settings,
+        );
+        // After a boosted round, reset the boosted canister's priority so it
+        // doesn't dominate subsequent rounds.
+        if let Some(canister_id) = target {
+            let p = result.metadata.subnet_schedule.get_mut(canister_id);
+            p.accumulated_priority = AccumulatedPriority::new(0);
+            p.priority_credit = AccumulatedPriority::new(0);
```
venkkatesh-sekar: this is wrong

rs/state_machine_tests/src/lib.rs
```diff
@@ -1916,6 +2026,28 @@ impl StateMachine {
...
             None => (SubnetConfig::new(subnet_type), HypervisorConfig::default()),
         };
+        // When flexible ordering is enabled, configure the scheduler so that
+        // exactly one canister message executes per round:
+        //  - max_instructions_per_slice = max_instructions_per_message (no DTS)
+        //  - max_instructions_per_round tuned so the scheduler's round_budget
+        //    (= max_round - max_slice + 1) allows one message to start but
+        //    exhausts the budget before a second can begin.
+        if flexible_ordering {
+            let max_msg = subnet_config.scheduler_config.max_instructions_per_message;
```
venkkatesh-sekar: can't disable DTS

rs/state_machine_tests/src/lib.rs
```diff
@@ -1916,6 +2026,28 @@ impl StateMachine {
...
+            // inner-round iteration. After the message finishes (+ 8M per-canister
+            // overhead), the budget goes very negative and the inner loop breaks.
+            let round_budget = ic_types::NumInstructions::from(1_000);
+            subnet_config.scheduler_config.max_instructions_per_round =
+                round_budget + max_msg - ic_types::NumInstructions::from(1);
+            // Use 2 scheduler cores so only 1 core is available for new
+            // canister executions (the other is reserved for long executions).
+            // Combined with priority boosting, this ensures only the target
+            // canister runs in each round.
+            subnet_config.scheduler_config.scheduler_cores = 2;
```
venkkatesh-sekar: can we set to 1

rs/state_machine_tests/src/lib.rs
```diff
@@ -2565,6 +2707,201 @@ impl StateMachine {
...
+                        _ => unreachable!(),
+                    };
+                    // Keep ticking until the message from `source` is no
+                    // longer in `target`'s sender schedule, meaning it was
+                    // consumed. System tasks (heartbeats, timers) that the
+                    // scheduler runs before popping from the input queue are
+                    // executed implicitly along the way.
+                    //
+                    // Phase 1: wait for the message to appear (it may need
+                    // prior rounds for induction / subnet message processing).
```
venkkatesh-sekar: subnet message needs to be tested as well

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+    let a_payload = wasm()
+        .inter_update(canister_b, CallArgs::default().other_side(b_reply))
+        .build();
+
+    let result = sm.execute_ingress(canister_a, "update", a_payload).unwrap();
+    match result {
+        WasmResult::Reply(data) => assert_eq!(data, b"normal reply"),
+        _ => panic!("Expected reply"),
+    }
+}
```
venkkatesh-sekar: not needed?

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+            source: canister_c,
+            target: canister_d,
+        },
+        OrderedMessage::Response {
+            source: canister_b,
+            target: canister_a,
+        },
+        OrderedMessage::Response {
+            source: canister_d,
+            target: canister_c,
```
venkkatesh-sekar, line 266: it should fail if someone tests with the response order changed (not possible)

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+    assert_eq!(get_reply(&sm, &ingress_a2), b"after upgrade");
+}
+
+/// Canister calls create_canister on IC_00 via inter-canister call.
+#[test]
+fn test_canister_calls_management_canister() {
+    let sm = setup();
+    let canister_a = install_uc(&sm);
+
+    let on_reply = wasm().message_payload().reply_data_append().reply().build();
```
venkkatesh-sekar, line 360: add a reply data here and check that's replied in the assert.

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+            source: IC_00,
+            target: canister_a,
+        },
+    ]));
+
+    let reply = get_reply(&sm, &ingress_a);
+    assert!(
+        !reply.is_empty(),
+        "Expected non-empty reply with canister ID"
+    );
```
venkkatesh-sekar, line 393: check canister exists?

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+        },
+    ]));
+
+    let reply = get_reply(&sm, &ingress_a);
+    let counter = u32::from_le_bytes(reply[..4].try_into().unwrap());
+    assert!(
+        counter >= 1,
+        "Heartbeat should have run, counter={}",
+        counter
+    );
```
venkkatesh-sekar: why is this not strict?

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+        },
+        OrderedMessage::Response {
+            source: canister_b,
+            target: canister_a,
+        },
+    ]));
+
+    let reply = get_reply(&sm, &ingress_a);
+    let counter = u32::from_le_bytes(reply[..4].try_into().unwrap());
+    assert!(counter >= 1, "Timer should have fired, counter={}", counter);
```
venkkatesh-sekar: why is this not strict?

rs/state_machine_tests/tests/flexible_ordering.rs
```diff
@@ -0,0 +1,935 @@
...
+            source: canister,
+            target: canister,
+        },
+    ]));
+
+    assert_eq!(get_reply(&sm, &ingress_id), b"self-reply");
+}
+
+#[test]
+fn test_heartbeat_then_request() {
```
venkkatesh-sekar, line 915: move heartbeat to first

rs/state_machine_tests/src/lib.rs
```diff
@@ -237,6 +239,109 @@ impl Verifier for FakeVerifier {
...
+            }
+            state
+                .metadata
+                .subnet_schedule
+                .get_mut(canister_id)
+                .accumulated_priority = AccumulatedPriority::new(i64::MAX / 4);
+        }
+
+        // Safe to restore: no new messages land in subnet_queues during
+        // execute_round — they stay in output until build_streams.
```
mraszyk: This assumption might be fragile.

rs/state_machine_tests/src/lib.rs
```diff
@@ -1932,6 +2052,20 @@ impl StateMachine {
...
+        // DTS per-slice limit is independent of round budget.
+        if flexible_ordering {
+            let sc = &mut subnet_config.scheduler_config;
+            sc.scheduler_cores = 1;
+            // Scheduler computes canister budget as:
+            //   max_instructions_per_round - max(slice, install_code_slice) + 1
+            // Cap install_code_slice so it doesn't inflate the subtraction.
+            sc.max_instructions_per_install_code_slice = sc.max_instructions_per_slice;
+            sc.max_instructions_per_round = sc.max_instructions_per_slice;
+            sc.subnet_messages_per_round_instruction_limit = ic_types::NumInstructions::from(1);
```
mraszyk: This still allows multiple subnet messages are do not consume instructions, e.g., multiple start_canister messages. (The test `test_one_subnet_message_per_round` passes because the subnet messages are ingress messages and the flexible scheduler inducts them one by one.)

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+    fn set_suppress_subnet_messages(&self, suppress: bool) {
+        *self.suppress_subnet_messages.write().unwrap() = suppress;
+    }
+
+    /// Reset flexible ordering state so subsequent tick()/execute_ingress()
+    /// calls pass through the scheduler without priority boost or suppression.
+    pub fn clear_flexible_ordering(&self) {
+        self.set_ordering_target(None);
+        self.set_suppress_subnet_messages(false);
+    }
```
mraszyk: This doesn't reset the changes performed in `StateMachine::setup_from_dir`: maybe we can omit this and make flexible ordering a permanent choice (at least for now)?

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+            .write()
+            .unwrap()
+            .remove(message_id)
+            .unwrap_or_else(|| panic!("No buffered ingress with id {}", message_id))
+    }
+
+    pub fn execute_with_ordering(&self, ordering: MessageOrdering) {
+        assert!(
+            self.flexible_ordering,
+            "call with_flexible_ordering() first"
```
mraszyk: "build" the StateMachine with flexible_ordering

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+                OrderedMessage::Request { source, target } if target == IC_00 => {
+                    assert!(
+                        self.next_sender_in_subnet_queue(source, target),
+                        "No message from {} to IC_00 in subnet_queues or loopback",
+                        source,
+                    );
+                    self.set_suppress_subnet_messages(false);
+                    self.tick();
+                    self.set_suppress_subnet_messages(true);
+                    self.complete_dts(source, MAX_TICKS);
```
mraszyk: Shouldn't these two lines be flipped (or the former dropped since we don't always clean up anyway)?

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+                    for tick in 0..MAX_TICKS {
+                        match self.ingress_status(ingress_id) {
+                            IngressStatus::Known {
+                                state: IngressState::Completed(_) | IngressState::Failed(_),
+                                ..
+                            } => break,
+                            IngressStatus::Known {
+                                state: IngressState::Processing,
+                                ..
+                            } if !self.has_in_flight_work(target) => break,
```
mraszyk: Why check for "in-flight" work here? This would also make `has_in_flight_work` obsolete (I was about to ask about its intented semantics which was not clear to me).

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+                    };
+                    self.set_next_scheduled_method(canister, method);
+                    self.set_ordering_target(Some(canister));
+                    self.tick();
+                    self.complete_dts(canister, MAX_TICKS);
+                }
+
+                OrderedMessage::Request { source, target } if target == IC_00 => {
+                    assert!(
+                        self.next_sender_in_subnet_queue(source, target),
```
mraszyk: This might fail spuriously: we might need to reshuffle the round-robin schedule instead of asserting that the production schedule yields the expected ordering.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+    /// subnet_queues (local schedule for requests to IC_00, remote schedule
+    /// for responses from IC_00) or in the loopback stream (not yet
+    /// inducted by Demux).
+    fn next_sender_in_subnet_queue(&self, source: CanisterId, target: CanisterId) -> bool {
+        let state = self.get_latest_state();
+
+        let in_schedule = match (source, target) {
+            (IC_00, _) => {
+                // IC_00 response: lands in remote schedule because IC_00's
+                // principal doesn't match own_subnet_id or any canister.
```
mraszyk: Inter-canister calls to the mgmt canister have the [subnet ID](https://github.com/dfinity/ic/blob/4aa273c7cdceddc0b1c7a562b3831d3b227f5516/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs#L498) as the callee (receiver). I suppose the test `test_canister_calls_management_canister` passes because of `has_loopback_messages`.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+                state.subnet_queues().local_sender_schedule().front() == Some(&source)
+            }
+            (_, _) => {
+                panic!("Management canister is not involved; use next_sender_in_queue");
+            }
+        };
+        in_schedule || self.has_loopback_messages()
+    }
+
+    fn has_loopback_messages(&self) -> bool {
```
mraszyk: We don't check the sender/receiver here which makes the check imprecise.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+                    self.set_ordering_target(Some(target));
+                    self.tick();
+                    self.complete_dts(target, MAX_TICKS);
+                }
+
+                OrderedMessage::Request { source, target }
+                | OrderedMessage::Response { source, target } => {
+                    self.set_suppress_subnet_messages(true);
+                    assert!(
+                        self.next_sender_in_queue(source, target),
```
mraszyk: This might fail spuriously: we might need to reshuffle the round-robin schedule instead of asserting that the production schedule yields the expected ordering.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+    }
+
+    /// Checks that a message involving IC_00 is available: either in
+    /// subnet_queues (local schedule for requests to IC_00, remote schedule
+    /// for responses from IC_00) or in the loopback stream (not yet
+    /// inducted by Demux).
+    fn next_sender_in_subnet_queue(&self, source: CanisterId, target: CanisterId) -> bool {
+        let state = self.get_latest_state();
+
+        let in_schedule = match (source, target) {
```
mraszyk: `InputSchedule::next_input_source` also matters

rs/state_machine_tests/src/lib.rs
```diff
@@ -2582,6 +2732,240 @@ impl StateMachine {
...
+    }
+
+    /// Checks that `source` is the next sender in `target`'s local input schedule.
+    fn next_sender_in_queue(&self, source: CanisterId, target: CanisterId) -> bool {
+        let state = self.get_latest_state();
+        let queues = match state.canister_state(&target) {
+            Some(c) => c.system_state.queues(),
+            None => return false,
+        };
+        queues.local_sender_schedule().front() == Some(&source)
```
mraszyk: `InputSchedule::next_input_source also matters`

venkkatesh-sekar: Added a new method `set_next_input_source` to fix the round robin before the check.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: Added a new method `set_next_input_source` to fix the round robin before the check.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: I think with `set_next_input_source`, the only way the assert triggers is if the user provides a wrong ordering.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: done, `next_sender_in_subnet_queue` and `has_loopback_message_for` should handle this correctly.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: done, `next_sender_in_subnet_queue` handles this explicitly now.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: This is redundant, removed.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: Scheduler now takes  `Arc<RwLock<SchedulerConfig>>` which allows setting the instruction limit dynamically for both subnet and canister.  With that change, this is redundant and was removed.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: I don't think there's an easy way to tackle this and I would accept this as an limitation for now WDYT?

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: The `clear_flexible_ordering` should reset back the scheduler limits. Only scheduler_cores = 1 will remain after clearing which I believe is acceptable. 

I want to have this method, so that a user can call `clear_flexible_ordering` from pocket ic and write tests normally. So that the setup and verification of the tests looks like a normal pocket ic test while the execute_with_ordering is enforced in between.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: Fixed with making Scheduler Config mutable.

rs/state_machine_tests/src/lib.rs
venkkatesh-sekar: Unless the known limitation of zero cost mgmt canister methods, everything else should be handled with the new changes.

rs/state_machine_tests/src/lib.rs
```diff
@@ -236,6 +238,129 @@ impl Verifier for FakeVerifier {
     }
 }
 
+/// Tasks not listed here may execute implicitly during any step:
+/// - DTS continuation (PausedExecution) runs with highest priority
+/// - OnLowWasmMemory hook runs before heartbeat/timer/messages if triggered
+/// - induct_messages_on_same_subnet moves output → input queues each round
+/// - build_streams routes output to loopback each round
```
mraszyk, line 245: these should still be controlled by flexible message ordering algorithm so I'm not sure why they're listed

rs/state_machine_tests/src/lib.rs
```diff
@@ -236,6 +238,129 @@ impl Verifier for FakeVerifier {
...
+        config.max_instructions_per_round = config.max_instructions_per_slice;
+        config.max_instructions_per_install_code_slice = config.max_instructions_per_slice;
+        config.subnet_messages_per_round_instruction_limit = NumInstructions::from(0);
+    }
+
+    /// canister_budget = 0, subnet_budget = 1.
+    fn prioritise_subnet_execution(&self) {
+        let mut config = self.scheduler_config.write().unwrap();
+        config.max_instructions_per_round =
+            config.max_instructions_per_slice - NumInstructions::new(1);
```
mraszyk, line 366: Wouldn't this still leave the canister budget at 1 instruction (due to saturating arithmetic)?

rs/state_machine_tests/src/lib.rs
```diff
@@ -1131,6 +1256,7 @@ pub struct StateMachine {
     cycles_account_manager: Arc<CyclesAccountManager>,
     cost_schedule: CanisterCyclesCostSchedule,
     hypervisor_config: HypervisorConfig,
+    fmo: Option<FlexibleMessageOrdering>,
```
mraszyk: Could we please use a more comprehensible variable name?

rs/state_machine_tests/src/lib.rs
```diff
@@ -2063,11 +2206,21 @@ impl StateMachine {
             )
         });
 
+        let ordering_target: Arc<RwLock<Option<CanisterId>>> = Arc::new(RwLock::new(None));
```
mraszyk: This line could be moved inside the if-branch.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2594,6 +2756,245 @@ impl StateMachine {
...
+        let default = SubnetConfig::new(self.subnet_type).scheduler_config;
+        config.max_instructions_per_round = default.max_instructions_per_slice;
+        config.max_instructions_per_install_code_slice = default.max_instructions_per_slice;
+        config.subnet_messages_per_round_instruction_limit =
+            default.subnet_messages_per_round_instruction_limit;
+        drop(config);
+        fmo.set_ordering_target(None);
+    }
+
+    pub fn buffer_ingress_as(
```
mraszyk: - Maybe `ingress_message_with_flexible_ordering` would be more descriptive?
- How about making the other methods for submitting an ingress message panic if the flexible ordering mode is turned on?

rs/state_machine_tests/src/lib.rs
mraszyk: We still need to do this:
```
self.set_next_input_source(target, InputSource::LocalSubnet);
```
and make sure that `source` is at the front of `local_sender_schedule` for subnet queues.

rs/state_machine_tests/src/lib.rs
mraszyk: We still need to make sure that source is at the front of `local_sender_schedule`.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2594,6 +2756,245 @@ impl StateMachine {
...
+    }
+
+    fn next_sender_in_subnet_queue(&self, source: CanisterId, target: CanisterId) -> bool {
+        let state = self.get_latest_state();
+
+        let subnet_canister_id = CanisterId::unchecked_from_principal(self.get_subnet_id().get());
+        match (source, target) {
+            (IC_00, target) => {
+                // IC_00 response: lands in remote schedule.
+                state.subnet_queues().remote_sender_schedule().front() == Some(&IC_00)
```
mraszyk:
```suggestion
                state.subnet_queues().remote_sender_schedule().front() == Some(&subnet_canister_id)
```

rs/state_machine_tests/src/lib.rs
mraszyk: but wouldn't it make sense to also assert on next_input_source here?

rs/execution_environment/src/scheduler.rs
```diff
@@ -158,7 +148,7 @@ pub(crate) struct SchedulerImpl {
 impl SchedulerImpl {
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
-        config: SchedulerConfig,
+        config: Arc<RwLock<SchedulerConfig>>,
```
mraszyk: How about avoiding `RwLock` by
- introducing a method `execute_round_with_config` taking `config: SchedulerConfig`
- and having the current method `execute_round` call `execute_round_with_config` using `SchedulerImpl::config`.

rs/execution_environment/src/scheduler.rs
```diff
@@ -554,7 +549,7 @@ impl SchedulerImpl {
             );
 
             round_limits.instructions -= as_round_instructions(
-                self.config
+                self.config()
                     .instruction_overhead_per_canister_for_finalization
                     * state.num_canisters() as u64,
             );
```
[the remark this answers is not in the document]
mraszyk: See https://github.com/dfinity/ic/pull/9508#discussion_r3084424854

rs/execution_environment/src/lib.rs
```diff
@@ -101,6 +104,7 @@ pub struct ExecutionServices {
     pub query_execution_service: QueryExecutionService,
     pub transform_execution_service: TransformExecutionService,
     pub scheduler: Box<dyn Scheduler<State = ReplicatedState>>,
+    pub scheduler_config: Arc<std::sync::RwLock<SchedulerConfig>>,
     pub query_stats_payload_builder: QueryStatsPayloadBuilderParams,
```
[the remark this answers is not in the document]
mraszyk: See https://github.com/dfinity/ic/pull/9508#discussion_r3084424854

rs/state_machine_tests/src/lib.rs
```diff
@@ -2594,6 +2756,245 @@ impl StateMachine {
...
+    }
+
+    fn set_next_scheduled_method(&self, canister_id: CanisterId, method: NextScheduledMethod) {
+        let (_, mut state) = self.state_manager.take_tip();
+        let canister = state
+            .canister_state_make_mut(&canister_id)
+            .unwrap_or_else(|| panic!("Canister {} not found", canister_id));
+        if let Some(es) = canister.execution_state.as_mut() {
+            es.next_scheduled_method = method;
+        }
```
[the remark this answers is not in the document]
mraszyk: This should work because the message will still fail with a corresponding error - but do we have test coverage for that?

rs/state_machine_tests/src/lib.rs
```diff
@@ -2594,6 +2756,245 @@ impl StateMachine {
...
+                        target,
+                    );
+                    self.set_next_scheduled_method(target, NextScheduledMethod::Message);
+                    fmo.set_ordering_target(Some(target));
+                    self.tick();
+                    self.complete_dts(target, MAX_TICKS);
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
mraszyk: We should forbid mixing flexible ordering API with the existing API.

rs/state_machine_tests/src/lib.rs
```diff
@@ -236,6 +238,129 @@ impl Verifier for FakeVerifier {
...
+        config.subnet_messages_per_round_instruction_limit = NumInstructions::from(0);
+    }
+
+    /// canister_budget = 0, subnet_budget = 1.
+    fn prioritise_subnet_execution(&self) {
+        let mut config = self.scheduler_config.write().unwrap();
+        config.max_instructions_per_round =
+            config.max_instructions_per_slice - NumInstructions::new(1);
+        config.max_instructions_per_install_code_slice = config.max_instructions_per_slice;
+        config.subnet_messages_per_round_instruction_limit = NumInstructions::from(1);
```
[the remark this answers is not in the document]
mraszyk: I think that's fine to ignore.

rs/state_machine_tests/src/lib.rs
```diff
@@ -2594,6 +2756,245 @@ impl StateMachine {
...
+    }
+
+    /// Reset per-round ordering state so subsequent tick()/execute_ingress()
+    /// pass through normally.
+    pub fn clear_flexible_ordering(&self) {
+        let fmo = self.fmo();
+        let mut config = fmo.scheduler_config.write().unwrap();
+        let default = SubnetConfig::new(self.subnet_type).scheduler_config;
+        config.max_instructions_per_round = default.max_instructions_per_slice;
+        config.max_instructions_per_install_code_slice = default.max_instructions_per_slice;
```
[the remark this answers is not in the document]
mraszyk: I think that's like this by design, but do we have test coverage? I'd still prefer to not mix flexible ordering with the production scheduler within the entire StateMachine lifetime though.

rs/state_machine_tests/src/lib.rs
mraszyk: but wouldn't it make sense to also assert on next_input_source here?

rs/state_machine_tests/src/lib.rs
mraszyk: > I want to have this method, so that a user can call clear_flexible_ordering from pocket ic and write tests normally. So that the setup and verification of the tests looks like a normal pocket ic test while the execute_with_ordering is enforced in between.

This can be achieved by creating multiple PocketIC instances and persisting their state in between. This way, we clearly separate the execution modes (flexible vs production) and reduce the number of edge cases to consider.

rs/state_machine_tests/src/lib.rs
mraszyk: As discussed separately, this will be addressed by ensuring that every subnet messages consumes a positive amount of instructions.

venkkatesh-sekar:
A newer version of the implementation was carried out in https://github.com/dfinity/ic/pull/9904 but was closed due to time constraints.

PR #9580, opened 2026-03-25
test(recovery): Mainnet NNS Recovery system test and testnet

pierugo-dfinity:
This PR adds a system test that recovers the NNS with the latest mainnet state end-to-end. It also adds a testnet that prepares the setup for a manual NNS recovery using the latest mainnet state.

The system test replaces the current `nr_large` to also hold mainnet state. As it is expected to be quite hungry resource-wise and time-wise (~1h30-2h), it is moved from release testing to nightly tests.

[CON-1624] Now that relevant changes to `ic-replay`/`ic-recovery` reached mainnet, we can come back to fetching the mainnet versions of these binaries. To make it easier to swap between branch and mainnet versions, this PR also introduces global consts `USE_HEAD_IC_REPLAY`/`USE_HEAD_IC_RECOVERY`. Their intended use is that if the setup breaks suddenly for some reason, we might need to adapt `ic-replay` and/or `ic-recovery`. One can then set the respective const to `true` and switch it back to `false` two weeks later.

The testnet `mainnet_nns_recovery` uses the same setup as the current `nns_recovery` but additionally puts mainnet state on the subnet. Also now by default, no nested VMs will be created (instead of 1 previously). This supports more natively potential future drills where the testnet is setup open to the Internet and NPs can join the testnet with their real machines. In that case, no nested VMs are created and the NNS does not change membership with them: the subnet consists only of the original NNS node (and a manual proposal can be made later to trigger the membership change once enough nodes have joined).

[CON-1624]: https://dfinity.atlassian.net/browse/CON-1624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/prep/src/internet_computer.rs
```diff
@@ -528,7 +528,10 @@ impl IcConfig {
                     dc_id: "".into(),
                     rewardable_nodes: BTreeMap::new(),
                     ipv6: None,
-                    max_rewardable_nodes: BTreeMap::new(),
+                    max_rewardable_nodes: BTreeMap::from([(
+                        "type3.1".into(),
+                        node_allowance.try_into().unwrap_or(u32::MAX),
+                    )]),
```
pierugo-dfinity, line 534: Nested VMs need to register to the registry and use this Node Operator. Without this change, they would not have the required allowance to register.

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -883,7 +883,7 @@ fn local_recovery(node: &IcNodeSnapshot, subnet_recovery: AppSubnetRecovery, log
     );
 
     info!(logger, "Executing local recovery command: \n{command}");
-    match node.block_on_bash_script_from_session(&session, &command) {
+    match node.block_on_bash_script_from_session(&session, &format!("{command} > /dev/null 2>&1")) {
```
pierugo-dfinity, line 886: This change was made to be consistent with the exact same one in `rs/nested/nns_recovery/common.rs`. From my experience, with mainnet state, the output of this command can be very large (many registry versions, big state) and could starve the channel/take very long to send over the network/allocate it in a buffer. Anyways, the output of that command is not used and can be silenced as such.

rs/tests/consensus/utils/src/node.rs
```diff
@@ -177,12 +177,14 @@ pub fn assert_node_is_unassigned_with_ssh_session(
     .expect("Failed to detect that node has deleted its state.");
 }
 
-fn fetch_metric_from_nodes<T>(nodes: Vec<IcNodeSnapshot>, metric_name: &str) -> Result<Vec<T>>
+async fn fetch_metric_from_nodes<T>(nodes: Vec<IcNodeSnapshot>, metric_name: &str) -> Result<Vec<T>>
```
pierugo-dfinity, line 180: The changes in this file are just about turning respective functions to async.

rs/tests/driver/src/driver/vector_vm.rs
```diff
@@ -312,6 +312,11 @@ docker run -d --name vector \
             self.container_running = true;
 
             emit_kibana_url_event(&log, &infra_group_name, &self.start_time);
+        } else {
+            info!(log, "Issuing command to reload vector configuration.");
+            deployed_vm
+                .block_on_bash_script_from_session(&session, "docker kill --signal=HUP vector")
+                .unwrap();
```
pierugo-dfinity, line 319: Interestingly, with mainnet state, it looks like this is required to see logs in Kibana after topology changes (see Vector [documentation](https://vector.dev/guides/level-up/managing-complex-configs/#updating-configs)). I had sent a DM to @NikolaMilosa a few months ago but I do not think this was addressed.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -89,30 +100,62 @@ fn get_host_vm_names(num_hosts: usize) -> Vec<String> {
     (1..=num_hosts).map(|i| format!("host-{i}")).collect()
 }
 
-pub fn replace_nns_with_unassigned_nodes(env: &TestEnv) {
+async fn replace_nns_with_nested_vms(env: &TestEnv, use_mainnet_state: bool) {
```
pierugo-dfinity, line 103: When using mainnet state, now suddenly there are a lot of unassigned nodes. The ones we are interested in are actually the nested VMs, so the logic of this function changes a bit.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -182,28 +237,54 @@ pub fn setup(env: TestEnv, cfg: SetupConfig) {
...
+    }
 
-    let host_vm_names = get_host_vm_names(cfg.subnet_size);
-    NestedNodes::new_with_resource_overrides(
-        &host_vm_names,
-        cfg.nested_nodes_vm_resource_overrides,
-    )
-    .setup_and_start(&env)
-    .unwrap();
+    if cfg.subnet_size > 0 {
```
pierugo-dfinity, line 246: As described in the PR description, if given a `cfg.subnet_size == 0`, then we will not spawn any nested VMs nor change the NNS membership.

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -182,28 +237,54 @@ pub fn setup(env: TestEnv, cfg: SetupConfig) {
...
+        let host_vm_names = get_host_vm_names(cfg.subnet_size);
+        let vm_resource_overrides = if cfg.use_mainnet_state {
+            cfg.nested_nodes_vm_resource_overrides
+                .layer(&MAINNET_NODE_VM_RESOURCE_OVERRIDES)
+                .layer(&NNS_RECOVERY_VM_RESOURCE_OVERRIDES)
+        } else {
+            cfg.nested_nodes_vm_resource_overrides
+                .layer(&NNS_RECOVERY_VM_RESOURCE_OVERRIDES)
+        };
```
pierugo-dfinity, line 255: I think it makes more sense to add the `NNS_RECOVERY_VM_RESOURCE_OVERRIDES` default once here, instead of at each call site. Now call sites can add a specific requirement, otherwise use `Default::default()`.

rs/tests/nested/src/lib.rs
```diff
@@ -85,45 +87,45 @@ pub fn get_bare_metal_login_info() -> LoginInfo {
...
-            .block_for_min_registry_version(ic_types::RegistryVersion::from(1)),
-    )
-    .unwrap();
+    let initial_topology = env.topology_snapshot();
 
-    // Check that there are initially no unassigned nodes.
-    let num_unassigned_nodes = initial_topology.unassigned_nodes().count();
-    assert_eq!(num_unassigned_nodes, 0);
+    // Keep track of the initial number of unassigned nodes.
+    let initial_num_unassigned_nodes = initial_topology.unassigned_nodes().count();
```
pierugo-dfinity, line 102: Again, with mainnet state, we cannot assume that there are no unassigned nodes initially.

rs/tests/testnets/mainnet_nns/src/lib.rs
```diff
@@ -299,21 +312,110 @@ async fn setup_recovered_nns(
...
-    )
-    .unwrap();
+    if USE_HEAD_IC_REPLAY {
+        std::fs::copy(
+            get_dependency_path_from_env("IC_REPLAY_PATH"),
+            env.get_path(PATH_IC_REPLAY),
+        )
+        .unwrap();
+    } else {
+        let logger = env.logger();
```
pierugo-dfinity, line 322: This code block and the one below were taken from the original testnet setup (https://github.com/dfinity/ic/commit/e38941af541aaa4b15b5d5676872077e6010d1ca).

rs/tests/nested/nns_recovery/nr_large_with_mainnet_state.rs
```diff
@@ -52,20 +52,21 @@ fn main() -> Result<()> {
...
-                        ..NNS_RECOVERY_VM_RESOURCE_OVERRIDES
+                        ..Default::default()
                     },
                 },
             )
         })
         .add_test(systest!(test; TestConfig {
-            local_recovery: false,
+            use_mainnet_state: true,
+            local_recovery: true,
```
pierugo-dfinity, line 62: We do a local recovery to:
- Use the same `ic-recovery` version as the GuestOS running on the nodes (otherwise we would use the HEAD one).
- Even better reflect a production recovery.

basvandijk (approved):

rs/tests/consensus/subnet_recovery/common.rs
eichhorl: Is this the ic-recovery output? In my experience it is very helpful to have this when debugging failures

rs/tests/nested/nns_recovery/BUILD.bazel
```diff
@@ -55,17 +60,32 @@ system_test_nns(
     ],
 )
 
+# To run this test,
+#
+#     bazel \
+#         test \
+#         --test_timeout=7200 \  # 2 hours
+#         //rs/tests/nested/nns_recovery:nr_large_with_mainnet_state
```
eichhorl, line 69: Do we also need to specify a datacenter that is colocated with a backup pod? I think that's what we used to do before

rs/tests/testnets/mainnet_nns/src/proposals.rs
```diff
@@ -205,4 +247,38 @@ impl ProposalWithMainnetState {
             "API Boundary Nodes addition proposal {:?} has been executed", proposal_id,
         );
     }
+
+    /// Code duplicate of rs/tests/driver/src/nns.rs:change_subnet_membership
```
eichhorl, line 251: Can this be avoided?

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: Ah, in NNS Recovery we were discarding that output anyways. I didn't see we logged it here. I [reverted](https://github.com/dfinity/ic/pull/9580/commits/6b1f36055a0779ed1778d2014978955fe6b6eb80) the change in that file. I agree it would be useful to keep it for NNS recoveries too, but I could argue that keeping that output only for app subnets should be enough.

rs/tests/nested/nns_recovery/BUILD.bazel
pierugo-dfinity: Yes that's a good point https://github.com/dfinity/ic/pull/9580/commits/da38dc7c6e55bf13a9ab6be63d3b8760564046db

rs/tests/testnets/mainnet_nns/src/proposals.rs
pierugo-dfinity: Yeah almost all functions in this file are some code duplicate, maybe we could refactor that in a follow-up by introducing functions in the test driver code that take a custom proposer.

.github/workflows/schedule-daily.yml
```diff
@@ -52,6 +52,28 @@ jobs:
...
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
 
+  nns-recovery-mainnet-state:
+    name: Bazel Test NNS Recovery with Mainnet State
+    runs-on: *dind-large-setup
+    container: *container-setup
+    timeout-minutes: 180
+    environment: Nightly Tests
+    steps:
+      - *checkout
```
eichhorl, line 62: Not entirely sure what it does, but should we also add this like the other pipelines of this file?
```suggestion
      - *checkout
      - uses: ./.github/actions/netrc
```

rs/tests/nested/nns_recovery/BUILD.bazel
```diff
@@ -55,17 +60,33 @@ system_test_nns(
     ],
 )
 
+# To run this test,
+#
+#     bazel \
+#         test \
+#         --set-required-host-features=dc=zh1
```
eichhorl:
```suggestion
#         --set-required-host-features=dc=zh1 \
```

rs/tests/testnets/mainnet_nns/src/lib.rs
```diff
@@ -62,6 +66,15 @@ const AUX_NODE_NAME: &str = "aux";
 const ORIGINAL_NNS_ID: &str = "tdb26-jop6k-aogll-7ltgs-eruif-6kk7m-qpktf-gdiqx-mxtrf-vb5e6-eqe";
 const MAINNET_NNS_DAPP_CANISTER_ID: &str = "qoctq-giaaa-aaaaa-aaaea-cai";
 
+// In general, the flags below should be set to `false` to download the mainnet version of
+// `ic-replay` and `ic-recovery` respectively and avoid compatibility bugs. Though, if some changes
+// were made to them specifically to make this testnet work, then we should use the HEAD version
+// and switch the corresponding flag flag to `true`. We must then hope that no incompatibles
```
eichhorl:
```suggestion
// and switch the corresponding flag to `true`. We must then hope that no incompatible
```

eichhorl: Alternatively what we used to do before was to push and publish a branch with the fix originating from the bugged mainnet version, and then manually set the replay/recovery version to the fixed version (so not necessarily the head version).

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -719,3 +868,101 @@ fn local_recovery(node: &IcNodeSnapshot, subnet_recovery: NNSRecoverySameNodes,
...
+/// Code duplicate of rs/tests/consensus/utils/src/subnet.rs:assert_subnet_is_healthy
+/// The difference is that we create an agent that does not verify query signatures.
+/// To verify query response signatures from the root subnet, the agent requires the root subnet_id.
+/// This is needed to retrieve the public keys of nodes within the root subnet.
+/// Typically, the agent derives the root subnet_id from the root key. However, when using mainnet
+/// state, the root key is different from the original one, but the subnet_id is reused.
+/// So we create a new agent that does not verify the query response signatures for the time being.
+/// A long-term solution involves modifying the agent to fetch the root subnet_id from the HTTP
+/// status endpoint.
+fn assert_subnet_is_healthy_with_mainnet_state(
```
eichhorl: Maybe something like
```suggestion
fn assert_subnet_is_healthy_without_signature_verification(
```
would be more accurate?

eichhorl (approved):

.github/workflows/schedule-daily.yml
pierugo-dfinity: Thank you! This wasn't there at the time I had written the job. It is used to have higher rate limits [AFAICT](https://github.com/dfinity/ic/blob/master/.github/actions/netrc/action.yaml). https://github.com/dfinity/ic/pull/9580/commits/77bc30738656f7d6d7b8ff92d417c454f49f5e1b

rs/tests/nested/nns_recovery/BUILD.bazel
pierugo-dfinity: Thank you https://github.com/dfinity/ic/pull/9580/commits/c7743ac96a77a5707b02cd9a1d01b1e93e4b3472

rs/tests/testnets/mainnet_nns/src/lib.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9580/commits/67b233ddee748a7abcb0bc0e22943ca3224d6144 + https://github.com/dfinity/ic/pull/9580/commits/b2f2ce08d00fd95a16d2df04f0ae71ff135de993
> Alternatively

Interesting, but I'm not sure I completely got it. What would be your suggested changes precisely?

rs/tests/nested/nns_recovery/common.rs
pierugo-dfinity: Yep https://github.com/dfinity/ic/pull/9580/commits/eef45cc64e4e4b5a85f74c3bc9c87371b23747f1

rs/tests/testnets/mainnet_nns/src/lib.rs
eichhorl: If I understand correctly, right now we have the choice between using:
1. the mainnet version (which may be bugged)
2. the head version (which may be incompatible)

So a third option could be to create a PR against the mainnet version, to fix the bug while still being compatible. The resulting hotifx version of ic-recovery/replay could then be downloaded by the mainnet recovery test.

pierugo-dfinity: I see. Does https://github.com/dfinity/ic/pull/9580/commits/bf500f69e48860c53a6f21de80a6da47ab804781 look like what you had in mind? (Deduplicated in https://github.com/dfinity/ic/pull/9580/commits/9177d4f53c86baf283889f980e5eeec8df6625c9)

rs/tests/nested/nns_recovery/common.rs
Bownairo: Makes sense to me!

Bownairo (approved):

rs/tests/testnets/mainnet_nns/src/lib.rs
eichhorl: Nice, thanks a lot!

PR #9707, opened 2026-04-01
feat(recovery): [CON-1604] allow to override what state height to download

pierugo-dfinity:
This PR aims at deflaking subnet recovery system tests that either:
- Download a state older than the latest CUP because the `download_state_node` has not computed the manifest yet.
- Download a state higher than the latest CUP because the subnet stalled right after computing the manifest but before creating a CUP.
- Download the correct state (equal to the latest CUP) but was considered invalid because it was received by state sync.

It does so by introducing a new `download_state_height` argument to `ic-recovery`. If provided, it will download the checkpoint at that height and throw and error if it does not exist.
If not provided (recommended in most production cases), it will default to the latest checkpoint.

Subnet recovery system tests are then adapted as such:
- Always download the checkpoint that matches the highest CUP on the subnet (solves first two flakes).
- Download it from a node that created a manifest for this CUP (solves third flake).
- Download the consensus pool from a node that has this highest CUP, and choose the one with highest certification share height.


rs/tests/consensus/subnet_recovery/utils.rs
```diff
@@ -163,15 +163,33 @@ pub fn get_node_certification_share_height(node: &IcNodeSnapshot, logger: &Logge
...
+            })
+        })
+        .max_by_key(
+            |NodeHeights {
+                 node: _,
+                 cert_share,
+                 cup,
+             }| (*cert_share, *cup),
+        )
         .expect("No healthy node found")
```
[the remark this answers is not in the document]
pierugo-dfinity: Applied

rs/tests/nested/nns_recovery/common.rs
```diff
@@ -438,6 +442,7 @@ pub fn test(env: TestEnv, cfg: TestConfig) {
         download_pool_node: Some(download_pool_node.get_ip_addr()),
         admin_access_location: Some(DataLocation::Remote(dfinity_owned_node.get_ip_addr())),
         keep_downloaded_state: Some(false),
+        download_state_height: Some(highest_cup),
```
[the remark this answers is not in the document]
pierugo-dfinity: Indeed

rs/tests/consensus/subnet_recovery/sr_nns_failover_nodes_test.rs
```diff
@@ -199,6 +202,7 @@ pub fn test(env: TestEnv) {
         registry_url: None,
         validate_nns_url: nns_node.get_public_url(),
         download_node: Some(download_node.get_ip_addr()),
+        download_state_height: Some(highest_cup),
         upload_method: Some(DataLocation::Remote(upload_node.get_ip_addr())),
```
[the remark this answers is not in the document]
pierugo-dfinity: Indeed

rs/recovery/src/lib.rs
```diff
@@ -1244,9 +1351,66 @@ fn get_member_node_ids_and_ips(
...
 
+    #[test]
+    fn get_ic_state_includes_test() {
+        let ic_checkpoints_path = PathBuf::from(IC_DATA_PATH).join(IC_CHECKPOINTS_PATH);
+        create_dir(&ic_checkpoints_path).unwrap();
+        assert!(
+            read_dir(&ic_checkpoints_path).unwrap().next().is_none(),
+            "Test checkpoints directory {} is not empty. Clear it first before running the test.",
+            ic_checkpoints_path.display()
+        );
```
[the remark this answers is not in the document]
pierugo-dfinity: Unfortunately, `PathBuf::from(IC_DATA_PATH).join(IC_CHECKPOINTS_PATH)` is hard-coded and must thus not exist in the test environment. I prefer to ask the developer to delete it themselves rather than delete it as part of the test to avoid deleting files that were deemed important to the developer.

rs/recovery/src/util.rs
```diff
@@ -52,6 +55,69 @@ pub fn data_location_from_str(s: &str) -> RecoveryResult<DataLocation> {
     )))
 }
 
+/// Helper struct to abstract over local and remote implementations of functions that need to access
+/// the filesystem of the node. Local implementations use the standard library, while remote
+/// implementations use ssh to execute commands on the remote node.
+pub enum MaybeRemote<'a> {
+    Local,
+    Remote(&'a SshHelper),
+}
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/9707/commits/6595691a303c7b4ab3dd5f6336ac07dfe9955e61

rs/recovery/src/app_subnet_recovery.rs
```diff
@@ -306,8 +310,8 @@ impl RecoveryIterator<StepType, StepTypeIter> for AppSubnetRecovery {
 
             StepType::DownloadConsensusPool => {
                 if self.params.download_pool_node.is_none() {
-                    // We could pick a node with highest finalization height automatically, but we
-                    // might have a preference between nodes of the same finalization height.
+                    // We could pick a node with highest finalization and CUP height automatically,
+                    // but we might have a preference between nodes of same heights.
```
kpop-dfinity, line 314: IIRC we don't print CUP heights in `print_height_info`. Should we start doing so?

rs/recovery/src/lib.rs
```diff
@@ -365,27 +373,48 @@ impl Recovery {
...
+    /// exist, an error is returned.
+    /// If it is not given, the latest checkpoint on the remote node is downloaded.
     ///
     /// If there are no checkpoints, this function returns an empty list and does not consider it
     /// an error, as the subnet could have stalled in its first DKG interval before producing any
     /// checkpoint.
-    pub fn get_ic_state_includes(ssh_helper: Option<&SshHelper>) -> RecoveryResult<Vec<PathBuf>> {
+    pub fn get_ic_state_includes(
+        maybe_remote: MaybeRemote<'_>,
+        download_height: Option<u64>,
```
kpop-dfinity, line 388: I'm probably overengineering but maybe we could use an enum here?

```
enum CheckpointHeight {
     Latest,
     Specific(u64),  // I don't know what a good name for this variant would be
}
```

Otherwise, without reading the documentation, it's not immediately clear what `None` means

rs/recovery/src/lib.rs
```diff
@@ -365,27 +373,48 @@ impl Recovery {
...
+    ) -> RecoveryResult<Vec<PathBuf>> {
         let ic_checkpoints_path = PathBuf::from(IC_DATA_PATH).join(IC_CHECKPOINTS_PATH);
-        let maybe_latest_checkpoint_name = if let Some(ssh_helper) = ssh_helper {
-            Self::get_maybe_latest_checkpoint_name_remotely(ssh_helper, &ic_checkpoints_path)?
+
+        let checkpoint_name = if let Some(height) = download_height {
+            let name = format!("{:016x}", height);
```
kpop-dfinity, line 393:
```suggestion
            let name = format!("{height:016x}");
```

r-birkner (approved):

rs/recovery/src/lib.rs
kpop-dfinity, line 388:
```suggestion
        checkpoint_height_to_download: Option<u64>,
```

rs/recovery/src/lib.rs
```diff
@@ -365,27 +373,48 @@ impl Recovery {
...
+            name
         } else {
-            Self::get_maybe_latest_checkpoint_name_and_height(&ic_checkpoints_path)?
-                .map(|(name, _height)| name)
-        };
+            let Some((name, _height)) =
+                maybe_remote.get_maybe_latest_checkpoint_name_and_height(&ic_checkpoints_path)?
+            else {
+                // No checkpoints, return an empty list of includes. This is not an error, as the
+                // subnet could have stalled in its first DKG interval.
```
kpop-dfinity, line 409: Should we at least log a warning?

rs/recovery/src/util.rs
```diff
@@ -52,6 +55,69 @@ pub fn data_location_from_str(s: &str) -> RecoveryResult<DataLocation> {
     )))
 }
 
+/// Helper enum to abstract over local and remote implementations of functions that need to access
+/// the filesystem of the node. Local implementations use the standard library, while remote
+/// implementations use ssh to execute commands on the remote node.
+pub enum MaybeRemote<'a> {
```
kpop-dfinity, line 61: nit: maybe something like

```suggestion
pub enum ExecutionMode<'a> {
```

rs/recovery/src/util.rs
```diff
@@ -52,6 +55,69 @@ pub fn data_location_from_str(s: &str) -> RecoveryResult<DataLocation> {
...
+                    return Ok(None);
+                };
+
+                let name = output.trim();
+                let height = parse_hex_str(name)?;
+
+                Ok(Some((name.to_string(), Height::from(height))))
+            }
+        }
+    }
```
kpop-dfinity, line 118: What do you think about something like this:

```suggestion
impl<'a> MaybeRemote<'a> {
    pub fn execute(&self, command: ...) -> Result<..., ...> {
          match self {
            Self::Local => <<execute command locally>
            Self::Remote(ssh_helper) => ssh_helper.ssh(command)...
    }
    
    pub fn path_exists(&self, path: &Path) -> RecoveryResult<bool> {
        let command = format!("test -e {} && echo y || echo n", path.display());
        self.execute(command).map(|result| result == "y")
        }
    }
    
        pub fn get_maybe_latest_checkpoint_name_and_height(
        &self,
        checkpoints_path: &Path,
    ) -> RecoveryResult<Option<(String, Height)>> {
         let maybe_output = self.execute(format!(
                    "ls -1 {} | sort | tail -n 1",
                    checkpoints_path.display()
                ))?;
                
                         let Some(output) = maybe_output else {
                    return Ok(None);
                };

                let name = output.trim();
                let height = parse_hex_str(name)?;

                Ok(Some((name.to_string(), Height::from(height))))
    }
```

so that we run exactly the same commands locally / remotely?

rs/recovery/src/util.rs
```diff
@@ -52,6 +55,69 @@ pub fn data_location_from_str(s: &str) -> RecoveryResult<DataLocation> {
...
+                    .ssh(command.clone())?
+                    .ok_or_else(|| {
+                        RecoveryError::cmd_error(
+                            &ssh_helper.get_command(command),
+                            None,
+                            "Could not check if path exists remotely".to_string(),
+                        )
+                    })?
+                    .trim()
+                    == "y")
```
kpop-dfinity, line 82: nit: instead of `Ok(bla?.trim() == "y")` maybe `bla.map(|output| output.trim() == "y")`

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -1130,3 +1086,116 @@ fn enable_chain_key_on_new_subnet(
...
     app_keys
 }
+
+struct RecoveryParameters {
+    download_pool_node: IcNodeSnapshot,
+    download_state_node: IcNodeSnapshot,
+    upload_node: IcNodeSnapshot,
+    admin_nodes: Vec<IcNodeSnapshot>,
+    replay_height: u64,
+    download_state_height: u64,
```
kpop-dfinity, line 1096: Maybe we should already return an `Option` here, so the callers don't have to cast it to `Option` on their own 
```suggestion
    download_state_height: Option<u64>,
```

rs/tests/testnets/nns_recovery.rs
```diff
@@ -144,7 +144,11 @@ fn log_instructions(env: TestEnv) {
...
     loop {
-        let (_, highest_cert_share) = node_with_highest_certification_share_height(
+        let NodeHeights {
+            node: _,
+            cup: _,
+            cert_share: highest_cert_share,
+        } = node_with_highest_cup_and_cert_share_heights(
             &env.topology_snapshot().root_subnet(),
             &logger,
         );
```
kpop-dfinity, line 154: nit: maybe?
```suggestion
    let  highest_cert_share = 
         = node_with_highest_cup_and_cert_share_heights(
&env.topology_snapshot().root_subnet(),
&logger,
).cert_share;
```

rs/tests/consensus/subnet_recovery/utils.rs
```diff
@@ -162,15 +162,33 @@ pub fn get_node_certification_share_height(node: &IcNodeSnapshot, logger: &Logge
         .map(|m| m.certification_share_height.get())
 }
 
+pub struct NodeHeights {
+    pub node: IcNodeSnapshot,
+    pub cup: u64,
+    pub cert_share: u64,
+}
+
 /// Select a node with highest certification share height in the given subnet snapshot
```
kpop-dfinity, line 171: This not necessarily holds any more, it could happen that there is a node with a cert share higher than everyone else, yet who hasn't produced the lastest CUP.

Is that an issue?

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -1130,3 +1086,116 @@ fn enable_chain_key_on_new_subnet(
...
+fn get_recovery_parameters(
+    env: &TestEnv,
+    cfg: &TestConfig,
+    app_subnet: &SubnetSnapshot,
+    app_node: &IcNodeSnapshot,
+) -> RecoveryParameters {
+    let logger = env.logger();
+
+    let mut parameters = if cfg.corrupt_cup.can_determine_subnet_id() {
+        // If we can deploy read-only access to the subnet, then:
```
kpop-dfinity, line 1108: so `cfg.corrupt_cup.can_determine_subnet_id()` implies that we can deploy `read-only` access?
Can we instead add another argument to `get_recovery_parameters`, say `can_deploy_read_only_access: bool`, so we don't have to rely on some implicit assumptions?

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -1130,3 +1086,116 @@ fn enable_chain_key_on_new_subnet(
...
+            download_state_node,
+            upload_node,
+            admin_nodes,
+            replay_height: highest_cert_share,
+            download_state_height: highest_cup,
+        }
+    } else {
+        // If we cannot deploy read-only access to the subnet, this would mean that the CUP is
+        // corrupted on enough nodes to stall the subnet which, in practice, should happen only
+        // during upgrades. In that case, all nodes stalled at the same height (the upgrade height)
```
kpop-dfinity, line 1152: `all nodes stalled at the same height (the upgrade height)` except the lagging behind nodes (like you mention below). If a node was very slow and doesn't have all the artifacts to reach the upgrade height, then it might never reach that height

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -1130,3 +1086,116 @@ fn enable_chain_key_on_new_subnet(
...
+                    .map(|m| (n, m.manifest_height.get()))
+            })
+            .find(|(_node, manifest_height)| *manifest_height >= highest_cup)
+            .map(|(node, _)| node)
+            .expect("No node found with last manifest height >= highest CUP height");
+
+        let upload_node = env
+            .topology_snapshot()
+            .unassigned_nodes()
+            .next()
```
kpop-dfinity, line 1136: Do I understand correctly that if there are unassigned nodes then we are in failover nodes recovery? (that's how I understand of the comment above)

I'd rather add another parameter to the `get_recovery_parameters` function, say`failover_mode: bool`, and do here:

```
let upload node = if failover_mode {
     env
            .topology_snapshot()
            .unassigned_nodes()
            .next()
            .expect("To do a failover nodes recovery we must have unassigned nodes")
} else {
     download_state_node
};
```

Instead of relying on some implicit assumptions about the setup of the tests

rs/tests/consensus/subnet_recovery/common.rs
```diff
@@ -1130,3 +1086,116 @@ fn enable_chain_key_on_new_subnet(
...
+
+        let download_pool_node = app_node.clone();
+
+        let download_state_node = download_pool_node.clone();
+
+        let upload_node = env
+            .topology_snapshot()
+            .unassigned_nodes()
+            .next()
+            .unwrap_or_else(|| download_state_node.clone());
```
kpop-dfinity, line 1171: ditto

kpop-dfinity: actually maybe we can even move this outside the `if {} else {}`, to avoid code duplciation

rs/recovery/src/util.rs
kpop-dfinity: or maybe we could even reuse `DataLocation` here somehow?

kpop-dfinity (commented):
This will make @basvandijk very happy

kpop-dfinity:
One more meta thing: I don't think `refactor` tag is appropriate here. I think `feat` makes more sense

basvandijk:
> This will make @basvandijk very happy

It does!

rs/recovery/src/app_subnet_recovery.rs
pierugo-dfinity: We print the debug representation of `NodeMetrics` which has a `catch_up_package_height` field with this PR :)

rs/recovery/src/lib.rs
pierugo-dfinity: I don't dislike it. I was also using an `Option<&SshHelper>` before switching to `MaybeRemote` which I think helps. WDYT of https://github.com/dfinity/ic/pull/9707/commits/8c9b67446da931bc56341cf0ca04d9d7db12a7cf ?

rs/recovery/src/lib.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9707/commits/7b3cfc42804900f6e34416e774d699e5a14e1e0e

rs/recovery/src/lib.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9707/commits/f2d53c910fbcf42708b7b433d826b9fd211613d4

rs/recovery/src/lib.rs
pierugo-dfinity: Good idea, https://github.com/dfinity/ic/pull/9707/commits/b3a6a23f6a83fa4696d2fab70b806e17af34ebf3

rs/recovery/src/util.rs
pierugo-dfinity: > or maybe we could even reuse DataLocation here somehow?

I looked into that but it doesn't look trivial with its usage directly in CLI args. Renamed to `ExecutionMode` https://github.com/dfinity/ic/pull/9707/commits/03813ecafeb4a259731d25f94ed72f10abe83df6

rs/recovery/src/util.rs
pierugo-dfinity: I agree, good idea. https://github.com/dfinity/ic/pull/9707/commits/15b651649b0d5fe81461a8b8573874e205d575bd

rs/recovery/src/util.rs
pierugo-dfinity: 👍 As part of https://github.com/dfinity/ic/pull/9707/commits/15b651649b0d5fe81461a8b8573874e205d575bd

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: Good idea. Though NNS recoveries (both failover and same nodes) do not use this function and would still need to cast it. We could then modify the return value of `node_with_highest_cup_and_cert_share_heights` but I do not think it's appropriate.
Another alternative would be that, in `ic-recovery`, if the given download height is 0, to skip the download. But I do not like this introduced edge-case and I would prefer to just trust the user.
So I do not think it's really useful here :(

rs/tests/testnets/nns_recovery.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9707/commits/64f31e9b6a6c1ccaae72fdc10a31a5f1e56eb799

rs/tests/consensus/subnet_recovery/utils.rs
pierugo-dfinity: Yeah I have thought about that and I think you're right :( I do not think that should lead to a lot of flakiness. The way I understand it:
- Except if the test driver is very fast*, all remaining healthy nodes after halting/breaking the subnet should have the same CUP height (i.e. P2P still runs in both cases).
- Out of these, it should be fine to select the node with highest certification share height.

\*it could happen that:
- A node aggregates a CUP
- Subnet halts/breaks
- The test driver scrapes all metrics
- The node gossips the CUP

In that case, only 1 (or a few) nodes would have the latest CUP and, indeed, may not have the highest certification share available. The test would flake if its certification share height is lower than the subnet's highest certification height in `ValidateReplayOutput` step. This is theoretically possible indeed. Let me think if we could also avoid that, but note that between halting/breaking the subnet and fetching the metrics, the test driver also asserts that the subnet is broken (1 query + update). The racing window looks quite unrealistic.

(I have updated the comment https://github.com/dfinity/ic/pull/9707/commits/5a8a090f414cdc09afebce8e29347f5d7d88e442)

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: Sure https://github.com/dfinity/ic/pull/9707/commits/52f9c66650a053ee81b318cf228efb37e9c6a0e3

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: Indeed. I think there's an assumption that between the moment the subnet halts/breaks and the moment the test driver fetches metrics, there is enough time for the node to catch-up in case it's falling behind (through P2P).

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9707/commits/52f9c66650a053ee81b318cf228efb37e9c6a0e3

rs/tests/consensus/subnet_recovery/common.rs
pierugo-dfinity: Unfortunately, it depends on `download_state_node` so it must be declared after it. Though `admin_nodes` (whose definition is different based on the `if-else`) depends on `upload_nodes` :(
I could introduce a closure that takes `download_state_node` as an argument, but I'm afraid that would badly affect readability.

rs/recovery/src/app_subnet_recovery.rs
kpop-dfinity: Ok

rs/recovery/src/lib.rs
```diff
@@ -1104,15 +1134,25 @@ pub async fn get_node_metrics(logger: &Logger, ip: &IpAddr) -> Option<NodeMetric
...
         let mut parts = line.split(' ');
         if let (Some(prefix), Some(height)) = (parts.next(), parts.next()) {
             match prefix {
+                r#"artifact_pool_consensus_height_stat{pool_type="validated",stat="max",type="catch_up_package"}"# => {
+                    match height.trim().parse::<u64>() {
+                        Ok(val) => node_heights.catch_up_package_height = Height::from(val),
+                        error => {
+                            warn!(logger, "Couldn't parse height {}: {:?}", height, error)
+                        }
+                    }
```
kpop-dfinity: nit: maybe create a function/closure to reduce code duplication? The same block appears 5 times

rs/tests/consensus/subnet_recovery/common.rs
kpop-dfinity: Ack

rs/tests/consensus/subnet_recovery/common.rs
kpop-dfinity: I see, leave it as it is then

rs/tests/consensus/subnet_recovery/utils.rs
kpop-dfinity: 👍

rs/recovery/src/lib.rs
kpop-dfinity: LGTM

rs/recovery/src/util.rs
kpop-dfinity: Ok

rs/recovery/src/util.rs
kpop-dfinity: nice

kpop-dfinity (approved):
LFG!

rs/recovery/src/lib.rs
pierugo-dfinity: Fair point https://github.com/dfinity/ic/pull/9707/commits/8f6f1f7050a66cef4819e6b3a911dd7eea218851

PR #9717, opened 2026-04-02
refactor: Separate adding/removing cycles from refunding/consuming cycles

dsarlis:
This PR refactors how cycles changes to balance and metrics are happening via the `SystemState`. Specifically, it more clearly separates adding/removing cycles to the balance from refunding/consuming cycles which on top of the balance updates the respective consumed metrics as well.

The idea is to separate in two pairs of APIs that will be responsible for each type of update:
```
fn add_cycles(amount: Cycles);
fn remove_cycles(amount: Cycles);
```
to handle balance only changes while 

```
fn refund_cycles<T>(prepayment: CompoundCycles<T>, refund: CompoundCycles<T>);
fn consume_cycles<T>(amount: CompoundCycles<T>);
```
will handle metrics being updated as well.

This change has a two fold benefit.

On one hand, it allows us to perform direct updates to balance (e.g. when depositing cycles or attaching cycles in calls) without having to construct `CompoundCycles<NonConsumed>` which was a somewhat hacky way to say "no need to update metrics". In fact, after this change, `NonConsumed` can be completely removed from `CyclesUseCases`. This allows for some simplification in call sites where they don't need to deal with `CompoundCycles` or doing so incurs unnecessary extra work.

On the other, it allows us to require a prepayment for refunding cycles which would be very helpful in a follow up where a metric which acts as a counter will be introduced and the amount consumed will need to be computed based on prepayment and refund. This way the call sites that don't need or should not deal with prepayments (i.e. ones that need only `add_cycles`), they don't have to be exposed to any of this.

The meat of the changes are in `SystemState` while there are some updates in production code and tests to match the refined API. In many cases, it ends up in a simpler version of the code as a bunch of `CompoundCycles::new` calls are eliminated or the cost schedule does not need to be piped through various calls.


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1807,28 +1805,62 @@ impl SystemState {
...
-            ConsumingCycles::No,
+            prepayment.nominal(),
+            refund.nominal(),
+            use_case,
+            ConsumingCycles::Refund,
         );
     }
 
+    /// Decreases 'cycles_balance' for 'requested_amount' from the canister's
+    /// main balance.
```
mraszyk: This sentence reads a bit strange

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1807,28 +1805,62 @@ impl SystemState {
...
+    /// `consume_cycles` instead.
+    pub fn remove_cycles(&mut self, requested_amount: Cycles) {
+        self.cycles_balance -= requested_amount;
+    }
+
     /// Decreases 'cycles_balance' for 'requested_amount'.
     /// The resource use cases first drain the `reserved_balance` and only after
     /// that drain the main `cycles_balance`.
-    pub fn remove_cycles<T: CyclesUseCaseKind>(&mut self, requested_amount: CompoundCycles<T>) {
+    pub fn consume_cycles<T: CyclesUseCaseKind>(&mut self, requested_amount: CompoundCycles<T>) {
```
mraszyk, line 1867: Maybe the doc comment here could be updated as well analogously to add_cycles/refund_cycles.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1919,7 +1953,30 @@ impl SystemState {
...
 
-        if use_case == CyclesUseCase::NonConsumed || amount.is_zero() {
+        // CyclesUseCase::NonConsumed should never be sent to this function.
+        debug_assert_ne!(
+            use_case,
+            CyclesUseCase::NonConsumed,
+            "Non-consumed cycles should not be tracked in the canister metrics."
+        );
+
+        // `prepayment`` must be greater or equal to `refund`.
```
mraszyk:
```suggestion
        // `prepayment` must be greater or equal to `refund`.
```

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1903,12 +1936,13 @@ impl SystemState {
...
         let balance = self.cycles_balance + self.reserved_balance;
-        self.remove_cycles(CompoundCycles::<Uninstall>::new(balance, cost_schedule));
+        self.consume_cycles(CompoundCycles::<Uninstall>::new(balance, cost_schedule));
     }
 
     fn observe_consumed_cycles_with_use_case(
         &mut self,
-        amount: NominalCycles,
+        prepayment: NominalCycles,
+        refund: NominalCycles,
```
mraszyk, line 1949: Why do we distinguish between `prepayment` and `refund` here when
- we require `refund` to be zero if `consuming_cycles` is `Prepayment`;
- and do not use `prepayment` if `consuming_cycles` is `Refund`?

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1807,28 +1805,62 @@ impl SystemState {
...
     }
 
-    /// Increments 'cycles_balance' and in case of refund for consumed cycles
-    /// decrements the metric `consumed_cycles`.
-    pub fn add_cycles<T: CyclesUseCaseKind>(&mut self, amount: CompoundCycles<T>) {
-        self.cycles_balance += amount.real();
+    /// Increments the canister's cycles balance by the provided amount.
+    ///
+    /// If you need to have metrics about consumed cycles updated use
+    /// `refund_cycles` instead.
```
mraszyk, line 1815: This property can no longer be enforced via `CyclesUseCase`: isn't that a major drawback of this refactoring?

rs/replicated_state/src/canister_state/system_state.rs
mraszyk: I guess it's because of this:
> On the other, it allows us to require a prepayment for refunding cycles which would be very helpful in a follow up where a metric which acts as a counter will be introduced and the amount consumed will need to be computed based on prepayment and refund.

But then I wonder why we don't update the metrics only when we know the refund amount, i.e., why we ever call `consume_cycles` only to later call `refund_cycles` instead of calling `remove_cycles` and later `refund_cycles` (updating the metrics according to the difference between prepaid and refunded cycles).

dsarlis: >  and do not use prepayment if consuming_cycles is Refund?

prepayment will be used in a follow up when we have a metric for consumed cycles that acts as a counter, i.e. only gets updated once when you know the actual amount.

The current metrics follow the pattern of updating during the prepayment and then fixing them up based on the refund amount.

In order to update metrics only during `refund_cycles` you'd have to perform a transition over to another set of metrics that don't follow that pattern. The code prepares for such a metric addition, whether or not you can cleanly convert over is unclear -- basically you can't really enforce that all callbacks for messages that have been sent with the existing pattern will have returned to be able to do a clean transition.

Basically, any outstanding refund has to be accounted for correctly before you can start updating things only during the refund phase. To do that you need to set some point where you know if the metrics have or have not been updated at the time the request was sent because that determines whether you need to subtract the refund (current state) or add the consumed (prepayment - refund) directly (potential future state). It's unclear to me whether such a point can be set definitively in the presence of unbounded wait calls.

rs/replicated_state/src/canister_state/system_state.rs
dsarlis: I don't understand your point. Updating metrics for consumed cycles is not something that's supposed to happen in every case when the balance of the canister is updated -- that is the whole point of the PR.

mraszyk: What I meant is that people are now expected to call `add_cycles`/`remove_cycles` passing `Cycles` instead of `CompoundCycles<...>` which makes it more likely that some new code paths could just use `add_cycles`/`remove_cycles` in cases when metrics should actually be updated. Using `CompoundCycles<NonConsumed>` makes it more explicit that metrics are not supposed to be updated.

dsarlis: `NonConsumed` feels a bit weird though as you pass that and then you have to make sure that metrics are never updated for that use case as that would be incorrect. It's also making certain call sites more verbose when in reality they don't have to be nor they have to do some extra work. Plus why can't someone mix things up now and pass `NonConsumed` when they are not supposed to (or vice-versa)? At least with the proposal here, if you have a `Cycles` amount handy (which is the case with deposits and transfers), you can simply pass it on to `add/remove_cycles`. And similarly, if you get a `CompoundCycles` from some `CyclesAccountManager` API, you will use the other set.

In any case, one of the main motivations here is to be able to have an additional argument to the function that adds (refunds) cycles to so that you can do the correct accounting for the case when you need to update metrics only at refund time. You can also go with a single:

```
// Names suck but I really don't know what to name things with a single function.
fn add_cycles<T>(prepayment: Option<CompoundCycles<T>>, refund: CompoundCycles<T>);
```
and handle everything inside it. If-else statements, `match`es, asserts/panics when wrong combination of arguments is passed, the whole package. I'm not sure that's much better. In my view, separating the functions disentangles some of this. Yes, it expects some discipline from the caller but that's where the types and doc comments come in.

mraszyk: Sounds good, thank you for your comprehensive explanation of your motivation.

> It's also making certain call sites more verbose when in reality they don't have to be nor they have to do some extra work. Plus why can't someone mix things up now and pass NonConsumed when they are not supposed to (or vice-versa)?

I think these ^ are the most convincing points to me so I'm closing this thread.

rs/replicated_state/src/canister_state/system_state.rs
mraszyk: > whether or not you can cleanly convert over is unclear

Very good question! I think you can convert if we just subtract all prepayments once after a subnet upgrade and before we run the first round after the subnet upgrade - then we're immediately in the "new" world, no?

dsarlis: My problem is that I don't know for sure if we have any lingering responses before [this](https://sourcegraph.com/r/github.com/dfinity/ic@86a3e5289b15a1a9fe555ced89beb7c6cdb7e08b/-/blob/rs/types/types/src/methods.rs?L281-L286) point.

In any case, I would not convert the existing metrics as it carries over some risk and potentially it can make things look weird on dashboards etc if we do such a "reset". I would add the new metrics that behave like counters and make a conversion if the above is either confirmed to not be a problem or if people are ok missing requests like the above.

Note that conversion if you add the new metrics is optional in the case where you want to eliminate the old metrics entirely but that would be something to do after lots of verifications that they behave equivalently to the old (i.e. no cases are missed). You should also consider that no matter whether you remove the old metrics from the code or not, they would anyway be kept at least on some prometheus recording rules to ensure that you can show cycles consumption historically.

mraszyk (approved):

rs/replicated_state/src/canister_state/system_state.rs
mraszyk: We'll follow up on the high-level suggestion to only update metrics at refund time separately.

dsarlis: After offline discussion the conclusion was to proceed with the current PR and come back to the point of updating metrics only during the refund phase later.

PR #9729, opened 2026-04-03
feat: Move a few tests away from the blessed versions list

Bownairo:
Since #9772, the registry no longer uses the blessed version list to enforce invariants.

Begin to move tests away from using them as well, so that we can remove the record entirely.


mraszyk (approved):

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -105,21 +114,18 @@ fn test_submit_and_accept_update_elected_replica_versions_proposal() {
...
+            default_version,
+            version_to_elect_and_unelect1,
+            version_to_elect_and_unelect2,
+            unassigned_nodes_version,
+        ] {
+            assert!(is_elected_version(&nns_canisters.registry, version).await);
+        }
+        for version in [version_to_elect] {
+            assert!(!is_elected_version(&nns_canisters.registry, version).await);
+        }
```
jasonz-dfinity: nit: the loop seems unnecessary

jasonz-dfinity (approved):

PR #9754, opened 2026-04-07
feat(execution): set SEV-enabled subnet reference size to 7

r-birkner:
- Introduces `SEV_REFERENCE_SUBNET_SIZE = 7` constant in `CyclesAccountManagerConfig`
- Adds `is_sev_enabled: bool` parameter to `CyclesAccountManagerConfig::application_subnet`, `SubnetConfig::new`, and related methods
- In the replica setup (`setup_ic_stack.rs`), reads `sev_enabled` from `SubnetFeatures` in the registry and passes it through to `SubnetConfig::new`
- All existing call sites default to `false` (no behaviour change)


mducroux (approved):

maksymar (approved):

rs/replay/src/player.rs
```diff
@@ -285,7 +285,7 @@ impl Player {
         };
 
         let metrics_registry = MetricsRegistry::new();
-        let subnet_config = SubnetConfig::new(subnet_type);
+        let subnet_config = SubnetConfig::new(subnet_type, false);
```
eichhorl: Should we also check the registry here, if the subnet is SEV enabled or not (like we do with the subnet type above)?

r-birkner: Good point. I changed that.

daniel-wong-dfinity-org (approved):
Approved for Governance.

rs/replay/src/player.rs
```diff
@@ -276,16 +276,22 @@ impl Player {
...
+        let subnet_type = match registry.get_subnet_record(subnet_id, registry_version) {
             Ok(Some(record)) => {
                 SubnetType::try_from(record.subnet_type).expect("Failed to decode subnet type")
             }
             err => panic!("Failed to extract subnet type of {subnet_id:?} from registry: {err:?}"),
         };
+        let is_sev_enabled = registry
+            .get_features(subnet_id, registry_version)
+            .ok()
+            .flatten()
```
eichhorl: Perhaps we could print something if this fails and we need to fall back to the default

eichhorl (approved):

rs/replica/src/setup_ic_stack.rs
```diff
@@ -126,12 +126,14 @@ pub fn construct_ic_stack(
...
-        subnet_id,
-        registry.get_latest_version(),
-        registry.as_ref(),
-    );
+    let registry_version = registry.get_latest_version();
+    let subnet_type = get_subnet_type(log, subnet_id, registry_version, registry.as_ref());
+    let is_sev_enabled = registry
+        .get_features(subnet_id, registry_version)
+        .ok()
+        .flatten()
```
eichhorl: What are the effects of falling back to `false` if this fails?

r-birkner: It will just be cheaper to run your canister. SEV protection is independent of that.

eichhorl: I'm more worried if it could lead to a state divergence if this fails for some replicas on the subnet but not for others. We could take inspiration of the `get_subnet_type` function above

eichhorl: More generally, is this flag something that may be changed via proposal? If so, it seems strange that it is applied on replica restarts, which may happen unexpectedly (again potentially leading to a state divergence).

r-birkner: At the moment, it can be changed with a proposal since we have a proposal for that out today. We want to remove that functionality again after today. From then on, this can only be set at subnet creation.

rs/replay/src/player.rs
r-birkner: Done

rs/replica/src/setup_ic_stack.rs
r-birkner: We have disabled the option to change whether a subnet is SEV enabled during the subnet's lifetime. A subnet can only be made SEV enabled when it is created.

eichhorl: Note the other part of my comment which has not been addressed yet as far as I can tell:
 
> I'm more worried if it could lead to a state divergence if this fails for some replicas on the subnet but not for others. We could take inspiration of the `get_subnet_type` function above

The API of the `RegistryClient` may return errors, and falling back to a fixed default might cause a replica to use a different value than its peers. Instead we could probably also return the `sev_enabled` value from `get_subnet_type` above (and rename it).

rs/config/src/subnet_config.rs
```diff
@@ -579,20 +586,24 @@ pub struct SubnetConfig {
...
-            SubnetType::Application => Self::default_application_subnet(),
+            SubnetType::Application => Self::default_application_subnet(is_sev_enabled),
             SubnetType::System => Self::default_system_subnet(),
-            SubnetType::VerifiedApplication => Self::default_verified_application_subnet(),
+            SubnetType::VerifiedApplication => {
+                Self::default_verified_application_subnet(is_sev_enabled)
+            }
             SubnetType::CloudEngine => Self::default_cloud_engine(),
         }
     }
```
alin-at-dfinity: Nit: Should this maybe complain if `is_sev_enabled == true` is passed in for `System` or `CloudEngine` subnets?

Or are we simply saying that for `System` and `CloudEngine` subnets it doesn't matter, because we don't charge cycles anyway? In that case maybe a brief doc comment would be useful?

rs/config/src/subnet_config.rs
```diff
@@ -457,10 +460,14 @@ pub struct CyclesAccountManagerConfig {
 }
 
 impl CyclesAccountManagerConfig {
-    pub fn application_subnet() -> Self {
+    pub fn application_subnet(is_sev_enabled: bool) -> Self {
```
mbjorkqvist: nit: Instead of `is_sev_enabled: bool`, would it make sense to introduce an enum (e.g., `SubnetSecurity` with variants `Sev` and `None`) for this (similar to `SubnetType`), to make the code more robust and improve readability?

mbjorkqvist (approved):
Thanks @r-birkner, one minor comment from my side!

rs/config/src/subnet_config.rs
r-birkner: Good idea! Done :)

rs/config/src/subnet_config.rs
r-birkner: I left a comment to explain why we don't pass it in for `System` and `CloudEngine` subnets

yhabib (approved):

PR #9755, opened 2026-04-07
chore(nervous-system): Create nns-sns-release Claude Code skill

jasonz-dfinity:
## Why

The NNS/SNS canister release process involves many manual steps (picking RC commits,
generating proposals, submitting via HSM, posting to the forum and Slack, scheduling
votes, updating changelogs). This skill automates and orchestrates the full workflow
within Claude Code, replacing manual use of `release-runscript`.

## What

- Add `.claude/skills/nns-sns-release/SKILL.md` covering the full release lifecycle:
  RC commit selection, upgrade target analysis with production-vs-test classification,
  upgrade dress rehearsal testing, proposal text generation with TODO handling,
  HSM-based proposal submission, forum post creation with pre-filled URLs and
  aggregation thread reminders, Slack notifications, calendar scheduling, and
  changelog updates
- Includes prerequisite validation, remembered neuron IDs, maintenance release
  detection, and fallbacks for optional tools (Slack MCP)

## Testing

- Used during the 2026-04-06 NNS release (governance + registry) to validate the
  workflow end-to-end; improvements incorporated from that session

### DIFF
--- a/.claude/skills/nns-sns-release/SKILL.md
+++ b/.claude/skills/nns-sns-release/SKILL.md
@@ -1,0 +1,344 @@
+---
+name: nns-sns-release
+description: Guide the NNS/SNS canister release process — pick RC commit, determine targets, create proposal texts, submit proposals, create forum posts, schedule votes, and update changelogs.
+---
+
+# NNS/SNS Canister Release
+
+This skill orchestrates the NNS and SNS canister release process. It replaces the
+`release-runscript` binary (`rs/nervous_system/tools/release-runscript`) by calling
+the same underlying bash scripts directly.
+
+All commands should be run from the repository root.
+
+## Prerequisites
+
+Before starting, verify prerequisites with a single command:
+```
+git fetch origin master --force && gh auth status && brew list coreutils >/dev/null && bazel --version
+```
+If any tool is missing, guide the user:
+- `gh`: `brew install gh && gh auth login`
+- `coreutils`: `brew install coreutils` (needed for `gdate` in changelog scripts)
+- `bazel`: `brew install bazelisk`
+
+Also check for optional tools and inform the user what will be automated vs manual:
+- **Slack MCP** (`mcp__plugin_slack_slack__*`): if available, will auto-post to `#dev-nns`
+  and update the channel topic. If not, will print the message and topic for manual posting.
+
+## Step 1: Pick Release Candidate Commit
+
+Run:
+```
+./rs/nervous_system/tools/release/cmd.sh latest_commit_with_prebuilt_artifacts
+```
+
+This returns the latest master commit with prebuilt artifacts. If it does not match
+`origin/master` HEAD, the artifacts are not ready yet. Poll every 2 minutes in the
+background (using Bash tool's `run_in_background` option) until they match. The user
+can continue other work while waiting. If artifacts have not appeared after 20 minutes,
+warn the user that CI may be broken and ask whether to continue waiting or use the
+latest available commit.
+
+Ask the user to confirm the commit or provide an override.
+
+Store the chosen commit as `RC_COMMIT` for all subsequent steps.
+
+## Step 2: Determine Upgrade Targets
+
+Run:
+```
+./rs/nervous_system/tools/release/list-new-commits.sh $RC_COMMIT
+```
+
+Parse the output to categorize each canister:
+- **Interesting commits**: canisters with at least one non-chore/refactor/test/docs commit (shown in green by the script). These are strong candidates for release.
+- **Chore-only commits**: canisters with commits but all are chore/refactor/test/docs. Generally skip these unless there is a reason to release.
+- **No new commits**: canisters with zero commits since last release. Do not release.
+
+**Exclude SNS ledger, index, and archive** from this process. The ICRC ledger suite
+(ledger/index/archive) is released separately by the DeFi team on a different cadence
+and commit. Do not show them in the table or suggest them for release.
+
+For each canister with interesting commits, inspect what those commits actually changed
+in the canister's source directory. Use `git diff --stat` or `git show --stat` to check
+whether the commit modified production code (e.g. `src/`, `canister/`, `*.rs` non-test files)
+or only test/doc/build files (e.g. `tests/`, `*_test.rs`, `*.md`, `BUILD.bazel`).
+
+Classify each interesting commit by inspecting the actual diff within the canister's
+code directory. Run:
+```
+git show $COMMIT_HASH -- $CANISTER_CODE_DIR
+```
+
+Then determine whether the changes affect production code or only test/doc/build artifacts.
+Be granular — in this codebase, tests often live alongside production code:
+- Changes inside `#[cfg(test)]` modules or `mod tests` blocks are **test only**
+- Changes to files under `tests/` directories or files named `*_test.rs`, `*_tests.rs` are **test only**
+- Changes to `BUILD.bazel`, `*.md`, `CHANGELOG.md` are **build/doc only**
+- Changes to `Cargo.toml` that only affect `[dev-dependencies]` are **test only**
+- Everything else (src code outside test blocks, `.did` files, `Cargo.toml` deps) is a **production change**
+
+A commit can have mixed changes — annotate it as production if ANY part of the diff
+modifies production code within the canister directory.
+
+A canister is a strong release candidate only if it has at least one commit with
+production changes. If all "interesting" commits only affect tests/docs/build within
+the canister directory, downgrade the recommendation to skip.
+
+The canister code directories can be determined from the `list-new-commits.sh` output
+(shown after the `git log` command in each section).
+
+Present each canister group (NNS, then SNS) to the user. For each canister, show:
+- The canister name, total commit count, and interesting commit count
+- **All commits** listed underneath, with interesting commits marked with `>>>` prefix
+  to distinguish them from chore/refactor/test/docs commits
+- For each interesting commit, annotate whether it is a production change or test/doc only
+- A recommendation: release / skip / no changes
+
+Example format:
+```
+### NNS governance (4 commits, 2 interesting) — RELEASE
+  >>> 077cec5e8f feat(nns): tag 8-year gang bonus base ... [production: src/governance.rs]
+  >>> 1855b0fd98 feat(NNS): add snapshot visibility ... [production: src/proposals.rs, canister/canister.rs]
+      a0034d3203 chore(nervous-system): update changelog
+      b608c374f2 chore: 42u64 -> 42_u64
+
+### NNS root (4 commits, 2 interesting) — SKIP (interesting commits are test/perf only)
+  >>> 1855b0fd98 feat(NNS): add snapshot visibility ... [production: src/lib.rs]
+  >>> d739eb3b60 perf: Speed up root canister integration tests [test only: tests/]
+      5458e72af6 chore(governance): Add separator ...
+      a5e8364794 docs(governance): Updated changelogs ...
+```
+
+After presenting all canisters, pre-select those with production changes and ask
+the user to confirm or adjust the selection. The user can add or remove canisters.
+
+Store the selected canisters as `NNS_CANISTERS` and `SNS_CANISTERS`.
+
+## Step 3: Run NNS Upgrade Tests
+
+**Skip this step if no NNS canisters were selected** (i.e. only SNS canisters).
+
+Run the NNS canister upgrade dress rehearsal test. This ensures selected NNS canisters
+remain upgradeable (no panic during pre-upgrade). The test uses golden NNS state.
+
+Before running, check that the working tree is clean (`git status --porcelain`). If
+there are uncommitted changes, warn the user — `git checkout` will fail or lose work.
+
+Print the fully-substituted one-liner command to the user before running it, with
+`$RC_COMMIT` and `$NNS_CANISTERS_CSV` replaced by their actual values (e.g.
+`governance,registry,root`):
+
+```
+git checkout $RC_COMMIT && bazel test --test_env=SSH_AUTH_SOCK --test_env=NNS_CANISTER_UPGRADE_SEQUENCE=$NNS_CANISTERS_CSV --test_output=streamed --test_arg=--nocapture --test_timeout=3600 //rs/nns/integration_tests:upgrade_canisters_with_golden_nns_state
+```
+
+This test can take 5-15 minutes. Run it in the background (using Bash tool's
+`run_in_background` option) so the user can continue working in the session.
+Proceed to step 4 (creating proposal texts) in parallel while the test runs,
+but do not proceed to step 5 (submission) until the test passes.
+
+SNS upgrade testing is covered by `sns_release_qualification` in CI and does not need
+to be run manually.
+
+## Step 4: Create Proposal Texts
+
+Create a proposals directory:
+```
+mkdir -p ../proposals/release-$(date +%Y-%m-%d)
+```
+
+For each NNS canister, run (note: `2>/dev/null` to avoid stderr warnings leaking into the file):
+```
+./rs/nervous_system/tools/release/prepare-nns-upgrade-proposal-text.sh $CANISTER $RC_COMMIT 2>/dev/null > ../proposals/release-$(date +%Y-%m-%d)/nns-$CANISTER.md
+```
+
+Special case: for `cycles-minting`, ask the user for an upgrade arg (default `()`), then:
+```
+./rs/nervous_system/tools/release/prepare-nns-upgrade-proposal-text.sh cycles-minting $RC_COMMIT "()" 2>/dev/null > ../proposals/release-$(date +%Y-%m-%d)/nns-cycles-minting.md
+```
+
+For each SNS canister, run:
+```
+./rs/nervous_system/tools/release/prepare-publish-sns-wasm-proposal-text.sh $CANISTER $RC_COMMIT ../proposals/release-$(date +%Y-%m-%d)/sns-$CANISTER.md
+```
+(Note: the SNS script writes directly to the output file, no `>` redirect needed.)
+
+After generating all proposals, check each file for "TODO" markers. For every
+proposal, compare the Features & Fixes section against the interesting commits
+identified in step 2 to see if any production changes are missing.
+
+- **"TODO: Review this section"** — the script generated Features & Fixes from the
+  unreleased changelog. If all interesting commits are accounted for, simply delete
+  the TODO marker paragraph. Do NOT rewrite the script's content. If commits appear
+  to be missing, **ask the user** whether to add entries rather than auto-editing.
+
+- **"TODO Hand-craft this section"** — the unreleased changelog was empty:
+  1. If any interesting commits from step 2 look like they deserve a changelog entry,
+     **ask the user** whether they should be added (PR authors may have misjudged).
+  2. If the user confirms none need entries (or all commits are chore/test/docs),
+     replace the entire TODO block with: `This is a maintenance release.`
+
+- Show the user the final proposal text for review before proceeding.
+
+Store the proposal file paths as `NNS_PROPOSAL_PATHS` and `SNS_PROPOSAL_PATHS`.
+
+## Step 5: Submit Proposals
+
+**IMPORTANT: This step requires interactive HSM PIN entry, which cannot be done
+inside the Claude Code session.**
+
+Check memory for the user's saved neuron ID. If found, pre-fill it and ask the user
+to confirm. If not found, ask the user for their neuron ID.
+(Lookup: https://www.notion.so/dfinityorg/3a1856c603704d51a6fcd2a57c98f92f?v=fc597afede904e499744f3528cad6682)
+
+Instruct the user to:
+1. Plug in their HSM key and unplug their YubiKey
+2. Test hardware: `pkcs11-tool --list-slots`
+
+Then, print the exact commands for the user to run in a **separate terminal / tmux pane**.
+Use the actual proposals directory path (e.g. `../proposals/release-2026-04-06`), not a
+variable. Each command should `tee` its output to a file so Claude can read the proposal
+IDs afterward.
+
+For each NNS proposal:
+```
+./rs/nervous_system/tools/release/submit-mainnet-nns-upgrade-proposal.sh ../proposals/release-YYYY-MM-DD/nns-$CANISTER.md $NEURON_ID 2>&1 | tee /tmp/proposal-nns-$CANISTER.out
+```
+
+For each SNS proposal:
+```
+./rs/nervous_system/tools/release/submit-mainnet-publish-sns-wasm-proposal.sh ../proposals/release-YYYY-MM-DD/sns-$CANISTER.md $NEURON_ID 2>&1 | tee /tmp/proposal-sns-$CANISTER.out
+```
+
+Note: Do NOT pipe through `sed` — it buffers output and hides the interactive HSM PIN prompt.
+
+The submit scripts output `proposal NNNNN` in the ic-admin response. After the user
+confirms submissions are done, read each `/tmp/proposal-*.out` file and extract the
+proposal ID by grepping for `proposal [0-9]+`.
+
+Store proposal IDs as `NNS_PROPOSAL_IDS` and `SNS_PROPOSAL_IDS`.
+
+## Step 6: Create Forum Post
+
+For NNS proposals, run:
+```
+./rs/nervous_system/tools/release/cmd.sh generate_forum_post_nns_upgrades $NNS_PROPOSAL_PATHS
+```
+
+For SNS proposals, run:
+```
+./rs/nervous_system/tools/release/cmd.sh generate_forum_post_sns_wasm_publish $SNS_PROPOSAL_PATHS
+```
+
+The generated post has two TODOs. Resolve them automatically:
+1. Replace "TODO proposal links" with links to each proposal:
+   `[Proposal NNNNN](https://dashboard.internetcomputer.org/proposal/NNNNN)`
+2. Delete the "TODO - delete if nothing relevant" line under Additional Notes
+   (or replace with actual notes if the user mentioned breaking changes).
+
+The generated post says "DFINITY plans to vote on these proposals the following Monday."
+If today is not Friday, update this to match the actual vote date (3 calendar days from
+today, per step 7). E.g. if today is Monday, change to "this Thursday".
+
+After resolving TODOs, write the final forum post to a file (e.g. `/tmp/forum-post-nns.md`).
+Copy it to clipboard by checking the `$TMUX` environment variable:
+- If `$TMUX` is set: use `tmux load-buffer /tmp/forum-post-nns.md` (user pastes with their tmux paste binding)
+- Otherwise: use `pbcopy < /tmp/forum-post-nns.md`
+
+Then provide a pre-filled forum link. Construct the URL with query parameters:
+
+For NNS:
+```
+https://forum.dfinity.org/new-topic?title=NNS+Updates+YYYY-MM-DD&category=Governance%2FNNS+proposal+discussions&tags=nns%2CProtocol-canister-management
+```
+
+For SNS:
+```
+https://forum.dfinity.org/new-topic?title=SNS+Updates+YYYY-MM-DD&category=Governance%2FNNS+proposal+discussions&tags=SNS%2CService-nervous-system-mgmt
+```
+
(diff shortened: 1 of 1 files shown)


.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,344 @@
...
+Example format:
+```
+### NNS governance (4 commits, 2 interesting) — RELEASE
+  >>> 077cec5e8f feat(nns): tag 8-year gang bonus base ... [production: src/governance.rs]
+  >>> 1855b0fd98 feat(NNS): add snapshot visibility ... [production: src/proposals.rs, canister/canister.rs]
+      a0034d3203 chore(nervous-system): update changelog
+      b608c374f2 chore: 42u64 -> 42_u64
+
+### NNS root (4 commits, 2 interesting) — SKIP (interesting commits are test/perf only)
+  >>> 1855b0fd98 feat(NNS): add snapshot visibility ... [production: src/lib.rs]
```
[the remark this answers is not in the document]
jasonz-dfinity: Removed the production/test annotations entirely — the diff-based classification was unreliable. Now just presents interesting vs non-interesting commits and lets the user decide.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,344 @@
...
+
+SNS upgrade testing is covered by `sns_release_qualification` in CI and does not need
+to be run manually.
+
+## Step 4: Create Proposal Texts
+
+Create a proposals directory:
+```
+mkdir -p ../proposals/release-$(date +%Y-%m-%d)
+```
```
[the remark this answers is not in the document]
jasonz-dfinity: Done — captured `RELEASE_DATE=$(date +%%Y-%%m-%%d)` once at the start of Step 4 and reused throughout.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Step 9: Post to Slack
+
+Prepare the following for `#dev-nns`:
+
+1. **Message:** `PTAL @ release changelog update $PR_URL`
+
+2. **Channel topic update.** Format:
+   ```
+   Apr 6, 2026:
```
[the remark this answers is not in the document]
jasonz-dfinity: Done — replaced hardcoded date with `$RELEASE_DATE` placeholder and added instructions to substitute.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+`release-runscript` binary (`rs/nervous_system/tools/release-runscript`) by calling
+the same underlying bash scripts directly.
+
+All commands should be run from the repository root.
+
+## Prerequisites
+
+Before starting, verify prerequisites with a single command:
+```
+git fetch origin master --force && gh auth status && brew list coreutils >/dev/null && bazel --version
```
basvandijk, line 18: Wouldn't it also be a good idea to checkout `master` to ensure the commands executed below, like `./rs/nervous_system/tools/release/cmd.sh`, are not run from some branch you happen to had checked out at the time of running this skill?

Note we do [the same](https://github.com/dfinity/ic/blob/master/.claude/skills/fix-flaky-tests/SKILL.md?plain=1#L12) in the `fix-flaky-tests` skill.

basvandijk, line 18: Regarding `gh`, shouldn't we explain what to do if not authenticated? 

Like [what we do](https://github.com/dfinity/ic/blob/master/.claude/skills/fix-flaky-tests/SKILL.md?plain=1#L15-L23) in the `fix-flaky-tests` skill?

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+If any tool is missing, guide the user:
+- `gh`: `brew install gh && gh auth login`
+- `coreutils`: `brew install coreutils` (needed for `gdate` in changelog scripts)
+- `bazel`: `brew install bazelisk`
+
+Also check for optional tools and inform the user what will be automated vs manual:
+- **Slack MCP** (`mcp__plugin_slack_slack__*`): if available, will auto-post to `#dev-nns`
+  and update the channel topic. If not, will print the message and topic for manual posting.
+
+## Step 1: Pick Release Candidate Commit
```
basvandijk, line 29: FYI: we recently [changed](https://github.com/dfinity/ic/commit/0ab05215fa680a64cb819c08e2f1823dee170653) when artifacts are uploaded:
* Before: artifacts were uploaded after `bazel test` completed successfully.
* Now: artifacts are uploaded after `bazel build` but before `bazel test`.

So you might want to confirm that your script below checks for passing tests on the picked commit besides just uploaded artifacts.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+description: Guide the NNS/SNS canister release process — pick RC commit, determine targets, create proposal texts, submit proposals, create forum posts, schedule votes, and update changelogs.
+---
+
+# NNS/SNS Canister Release
+
+This skill orchestrates the NNS and SNS canister release process. It replaces the
+`release-runscript` binary (`rs/nervous_system/tools/release-runscript`) by calling
+the same underlying bash scripts directly.
+
+All commands should be run from the repository root.
```
basvandijk, line 12: You might want to consider explicitly cd-ing to the root for every command like we do in [CLAUDE.md](https://github.com/dfinity/ic/blob/faa90763e23ad566f32b333496b6f62e2c0fb567/.claude/CLAUDE.md?plain=1#L11).
Without it I've noticed the Claude often goes into a directory, even when instructed to run commands from root.

basvandijk (approved):
Nice.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
+---
+name: nns-sns-release
```
daniel-wong-dfinity-org, line 2:
```suggestion
name: upgrade-governance-backend
```

because this does not include nns-dapp, nor the (new) Governance UI.

For me, the key word here is "backend".

"governance" seems like a more elegant way to include both NNS and SNS (and whatever else we might own in the future).

"upgrade" seems more concrete than "release", but I do not have a strong preference here.

Not sure what (if any) conventions there are around skill name; in particular, not sure if they should be imperative "do X", or noun "X procedure". But since they are basically scripts, imperative makes more sense to me.

just my 2 cents

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+description: Guide the NNS/SNS canister release process — pick RC commit, determine targets, create proposal texts, submit proposals, create forum posts, schedule votes, and update changelogs.
```
daniel-wong-dfinity-org, line 3:
```suggestion
description: >
    Use multiple
    lines; avoid
    super long lines.
```

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+---
+
+# NNS/SNS Canister Release
+
+This skill orchestrates the NNS and SNS canister release process. It replaces the
+`release-runscript` binary (`rs/nervous_system/tools/release-runscript`) by calling
+the same underlying bash scripts directly.
```
daniel-wong-dfinity-org, line 10: > It replaces the `release-runscript` binary

This seems like a step backwards. When it is highly FEASIBLE to do something in a totally mechanical way, it should be done totally mechanically. That way, we can KNOW WITH PRECISION what is going to happen, not leave things up to the fickle AI gods.

An example of when it makes sense to let AI decide how to do something instead of a program (Rust or Bash): debugging. There is NO completely mechanical procedure for debugging; it requires intelligence. So, since there is no way to do it in a completely mechanical way, we would fall back to AI.

I guess the concrete suggestion here is to (as much as possible) turn this skills file into actual code enhancements to the runscript. I'm not saying to do this right now; this could be done later.

.claude/skills/nns-sns-release/SKILL.md
daniel-wong-dfinity-org, line 18: Why `&&`? Is Claude not smart enough to realize that when one command fails, it should stop, and not just blindly proceed to the next one?

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+  and update the channel topic. If not, will print the message and topic for manual posting.
+
+## Step 1: Pick Release Candidate Commit
+
+Run:
+```
+./rs/nervous_system/tools/release/cmd.sh latest_commit_with_prebuilt_artifacts
+```
+
+This returns the latest master commit with prebuilt artifacts. If it does not match
```
daniel-wong-dfinity-org, line 36: > prebuilt artifacts

Does this mean that all tests passed? If so, that is the MUCH more interesting fact. Otherwise, the mere EXISTENCE of artifacts does not mean that they should be put into production.

If exists implies passed, let's explicitly mention that.

What I've always done is (as described in [Notion]) go to [this] page to verify that the suggested commit passes all tests.

[Notion]: https://www.notion.so/dfinityorg/Backend-Release-6d176f60478c4236a1af5f14462e73fc?source=copy_link#dd14a7a08bd347a18996500555edbab7
[this]: https://github.com/dfinity/ic/actions/workflows/ci-kickoff.yml?query=branch%3Amaster+is%3Asuccess

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Step 2: Determine Upgrade Targets
+
+Run:
+```
+./rs/nervous_system/tools/release/list-new-commits.sh $RC_COMMIT
+```
+
+Parse the output to categorize each canister:
+- **Interesting commits**: canisters with at least one non-chore/refactor/test/docs commit (shown in green by the script). These are strong candidates for release.
```
daniel-wong-dfinity-org, line 55: > non-

It is not clear whether this just applies to chore, or whether it applies to ALL the other CC prefixes mentioned.

Perhaps, a better way to describe this is to say that feat and fix are interesting prefixes? idk. I guess you just matched the script's definition of interesting.

Also, this is a long line. Ditto elsewhere.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+background (using Bash tool's `run_in_background` option) until they match. The user
+can continue other work while waiting. If artifacts have not appeared after 20 minutes,
+warn the user that CI may be broken and ask whether to continue waiting or use the
+latest available commit.
+
+Ask the user to confirm the commit or provide an override.
+
+Store the chosen commit as `RC_COMMIT` for all subsequent steps.
+
+## Step 2: Determine Upgrade Targets
```
daniel-wong-dfinity-org, line 47: Instead of doing what we used to do (base on conventional commit prefixes), we can (in a future PR) maybe instead base this decision on unreleased_changelog.md. If anyone "forgets" to add something to unreleased_changelog.md (despite the reminder bot), and as a result, we do not upgrade, that's on them. If we start doing things this way, there will be a HARD reason for people to update unreleased_changelog.md. This also ensures that the proposal summary file WILL have stuff in the "Features & Fixes" section, and therefore, we maybe won't have to manually check that that section matches the commits section.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+Before running, check that the working tree is clean (`git status --porcelain`). If
+there are uncommitted changes, warn the user — `git checkout` will fail or lose work.
+
+Print the fully-substituted one-liner command to the user before running it, with
+`$RC_COMMIT` and `$NNS_CANISTERS_CSV` replaced by their actual values (e.g.
+`governance,registry,root`):
+
+```
+git checkout $RC_COMMIT && bazel test --test_env=SSH_AUTH_SOCK --test_env=NNS_CANISTER_UPGRADE_SEQUENCE=$NNS_CANISTERS_CSV --test_output=streamed --test_arg=--nocapture --test_timeout=3600 //rs/nns/integration_tests:upgrade_canisters_with_golden_nns_state
```
daniel-wong-dfinity-org, line 102: Spread this over multiple lines, like so:

```
git checkout $RC_COMMIT && \
bazel test \
    --test_env=SSH_AUTH_SOCK \
    etc...
```

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+This test can take 5-15 minutes. Run it in the background (using Bash tool's
+`run_in_background` option) so the user can continue working in the session.
+Proceed to step 4 (creating proposal texts) in parallel while the test runs,
+but do not proceed to step 5 (submission) until the test passes.
+
+SNS upgrade testing is covered by `sns_release_qualification` in CI and does not need
+to be run manually.
```
daniel-wong-dfinity-org, line 111: This need not be mentioned.

.claude/skills/nns-sns-release/SKILL.md
daniel-wong-dfinity-org, line 102: It feels wrong that this skills file could be modified by `git checkout ...` while it is being run. I mean, most likely, `git checkout` will not cause this file to be changed while it is being run, but still... I mean, any automation that runs from within the repo that gets used to test will have this problem...

Anyway, I do not have a specific suggestion here (so this is non-blocking). I just want to be on the record for this...

daniel-wong-dfinity-org, line 102: Does this `bazel test ...` cmd work from mac?? I've always run it from my devenv VM. I guess I've never tried it locally, but in theory, it should work... It will just be a bit slow, I guess, since local Internet is probably not that fast, but hey, if it works at all, that's a win!

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+`RELEASE_DATE=$(date +%Y-%m-%d)`. Use `$RELEASE_DATE` consistently below.
+
+Create a proposals directory:
+```
+mkdir -p ../proposals/release-$RELEASE_DATE
+```
+
+For each NNS canister, run (note: `2>/dev/null` to avoid stderr warnings leaking into the file):
+```
+./rs/nervous_system/tools/release/prepare-nns-upgrade-proposal-text.sh $CANISTER $RC_COMMIT 2>/dev/null > ../proposals/release-$RELEASE_DATE/nns-$CANISTER.md
```
daniel-wong-dfinity-org, line 125: Why redirect stderr?

This could (also) be spread over multiple lines. Ditto elsewhere.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+
+For each SNS canister, run:
+```
+./rs/nervous_system/tools/release/prepare-publish-sns-wasm-proposal-text.sh $CANISTER $RC_COMMIT ../proposals/release-$RELEASE_DATE/sns-$CANISTER.md
+```
+(Note: the SNS script writes directly to the output file, no `>` redirect needed.)
+
+After generating all proposals, check each file for "TODO" markers. For every
+proposal, compare the Features & Fixes section against the interesting commits
+identified in step 2 to see if any production changes are missing.
```
daniel-wong-dfinity-org, line 141: This is an example of where intelligence is really needed. I think the earlier steps are all probably purely mechanical.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+  the TODO marker paragraph. Do NOT rewrite the script's content. If commits appear
+  to be missing, **ask the user** whether to add entries rather than auto-editing.
+
+- **"TODO Hand-craft this section"** — the unreleased changelog was empty:
+  1. If any interesting commits from step 2 look like they deserve a changelog entry,
+     **ask the user** whether they should be added (PR authors may have misjudged).
+  2. If the user confirms none need entries (or all commits are chore/test/docs),
+     replace the entire TODO block with: `This is a maintenance release.`
+
+- Show the user the final proposal text for review before proceeding.
```
daniel-wong-dfinity-org, line 154: Maybe, only show the Features & Fixes section, since the rest is always fully mechanically generated? Maximize signal-to-noise ratio.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+Store the proposal file paths as `NNS_PROPOSAL_PATHS` and `SNS_PROPOSAL_PATHS`.
+
+## Step 5: Submit Proposals
+
+**IMPORTANT: This step requires interactive HSM PIN entry, which cannot be done
+inside the Claude Code session.**
+
+Check memory for the user's saved neuron ID. If found, pre-fill it and ask the user
+to confirm. If not found, ask the user for their neuron ID.
+(Lookup: https://www.notion.so/dfinityorg/3a1856c603704d51a6fcd2a57c98f92f?v=fc597afede904e499744f3528cad6682)
```
daniel-wong-dfinity-org, line 165: As you may know, our Notion is going to be deleted soon. Supposedly, content has been copied to Confluence, but I'm not able to find the equivalent page there...

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+./rs/nervous_system/tools/release/cmd.sh generate_forum_post_sns_wasm_publish $SNS_PROPOSAL_PATHS
+```
+
+The generated post has two TODOs. Resolve them automatically:
+1. Replace "TODO proposal links" with links to each proposal:
+   `[Proposal NNNNN](https://dashboard.internetcomputer.org/proposal/NNNNN)`
+2. Delete the "TODO - delete if nothing relevant" line under Additional Notes
+   (or replace with actual notes if the user mentioned breaking changes).
+
+The generated post says "DFINITY plans to vote on these proposals the following Monday."
```
daniel-wong-dfinity-org, line 212: Maybe, we can change the scripts to say "three days from now" instead of "the following Monday". That way, we do not need any intelligence to reword (this aspect) of the forum post.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+```
+https://forum.dfinity.org/new-topic?title=NNS+Updates+YYYY-MM-DD&category=Governance%2FNNS+proposal+discussions&tags=nns%2CProtocol-canister-management
+```
+
+For SNS:
+```
+https://forum.dfinity.org/new-topic?title=SNS+Updates+YYYY-MM-DD&category=Governance%2FNNS+proposal+discussions&tags=SNS%2CService-nervous-system-mgmt
+```
+
+The user can paste from tmux buffer with their configured paste binding (typically `prefix + ]`).
```
daniel-wong-dfinity-org, line 233: > (typically `prefix + ]`).

?? On mac, you would paste into the web form using cmd+v; therefore, I am not sure what this is about.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+- [Registry: $ID](https://dashboard.internetcomputer.org/proposal/$ID)
+SNS:
+- [Root: $ID](https://dashboard.internetcomputer.org/proposal/$ID)
+```
+
+Instruct the user to:
+1. Duplicate a past calendar event from:
+   https://calendar.google.com/calendar/u/0/r/eventedit/duplicate/MjJvMTdva2xtdGJuZDhoYjRjN2poZzNwM2ogY182NGYwZDdmZDYzYjNlMDYxZjE1Zjk2MTU1NWYzMmFiN2EyZmY3M2NjMWJmM2Q3ZTRkNGI3NGVjYjk1ZWVhM2M0QGc
+2. Use the "NNS Upgrades" calendar
+3. Schedule at 6 PM CET, 3 calendar days from today (use the user's local date).
```
daniel-wong-dfinity-org, line 256: > 6 pm

I think we do 11 am now.

(Ideally, if nns-dapp already created an event, we piggy back off that, but this is more difficult to automate, and reducing Calendar events is not critical.)

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+4. Title: include canister name and proposal ID
+5. Description: link to the proposal
+6. Send email invitations when prompted
+7. If people do not respond, ping @trusted-neurons in #eng-release
+
+## Step 8: Update Changelog
+
+Switch to the RC commit (detached HEAD):
+```
+git switch -d $RC_COMMIT
```
daniel-wong-dfinity-org, line 267: I think from the upgrade test step, we are probably already at $RC_COMMIT, right? I mean, forcing it is not a bad thing. Just asking. Maybe, mention that this likely doesn't actually do anything, but rather is just a "belt and suspenders" measure.

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+5. Create a draft PR:
+   ```
+   gh pr create --draft --title "chore(nervous-system): Update changelog for release YYYY-MM-DD" --body "Update CHANGELOG.md for today's release."
+   ```
+
+Tell the user: "PR can be merged before the proposals are executed."
+
+## Step 9: Post to Slack
+
+Prepare the following for `#dev-nns` (private channel — search with `channel_types=private_channel`):
```
daniel-wong-dfinity-org, line 291: What about #eng-nns-releases?

.claude/skills/nns-sns-release/SKILL.md
```diff
@@ -0,0 +1,319 @@
...
+   SNS_CANISTERS=(
+       $SNS_CANISTER_LIST
+   )
+   ```
+   Before updating, read the existing channel topic and validate it follows this
+   expected format (date, RC=, NNS_CANISTERS, SNS_CANISTERS). If it does not match,
+   warn the user instead of overwriting.
+
+If Slack MCP tools are available (`mcp__plugin_slack_slack__*`), create a draft message
+using `slack_send_message_draft` so the user can review before sending. The Slack MCP cannot update channel topics, so always print the topic
```
daniel-wong-dfinity-org, line 315: What is slack_send_message_draft?

daniel-wong-dfinity-org (approved):

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — added CI verification step and clarified that artifacts don't guarantee tests passed.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: My main rationale is that the skill works better in a non-standard environment, which our laptops unfortunately are. In an environment like this, steps often fail for various reasons that are difficult to foresee (end-to-end tests can't catch them as those run in a standard environment). On the other hand, it's much easier to continue when both the context and errors live in the same Claude Code session, and workarounds can more easily be found and the session can continue.

On the other hand, the main downside of using LLM to wrap script is that it's hard to make sure that each step works correctly. I think it's not so bad for this task, because the only important side effect is the proposal being sent out, and there is a human gate right before it being sent out.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: The `&&` chaining is intentional — it minimizes the number of tool-approval prompts the user needs to click through.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — added CI verification step and clarified that artifacts don't guarantee tests passed.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — clarified that `feat` and `fix` are the interesting prefixes, and wrapped long lines.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Sounds like a good idea, but it would be a change to our release process, let me consider it.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: In practice, Claude Code loads the skill file into the conversation at invocation time, so a subsequent `git checkout` doesn't affect the running session.

jasonz-dfinity: Good question — updated to let the user choose: run locally, via remote bazel MCP, or just print the command for copy-paste into devenv.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — removed stderr redirects and spread commands over multiple lines.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Agreed — this is exactly where AI judgment adds value over a mechanical script.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Keeping the full proposal review — since the LLM is editing the text (vs a human manually), showing the complete proposal before submission serves as an important safety check.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — removed the Notion link. The skill already checks Claude memory for the neuron ID.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — changed the scripts to compute and embed the actual vote date (3 days from now).

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done — simplified to just `pbcopy`, which works fine inside tmux on mac.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Done

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: Right, we should already be on `$RC_COMMIT` from the upgrade test step. The explicit `git switch -d` here is a safety measure.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: We stopped posting to #eng-nns-releases a while ago — nobody has complained, so #dev-nns is sufficient.

.claude/skills/nns-sns-release/SKILL.md
jasonz-dfinity: It's a Slack MCP tool that creates a message draft visible only to the user, so they can review before actually sending. Clarified in the text.

PR #9772, opened 2026-04-08
feat: Stop using blessed versions list for invariants in the registry

Bownairo:

This moves the invariants in the registry to use the list of ReplicaVersionRecords, instead of the blessed versions list, as a step towards removing the redundant blessed versions list entirely.


rs/registry/canister/src/invariants/crypto.rs
```diff
@@ -75,13 +75,13 @@ fn check_node_crypto_keys_exist_and_are_unique(
...
     println!("{LOG_PREFIX}node_crypto_keys_invariants_check_start");
-    let nodes = get_node_records_from_snapshot(snapshot);
+    let nodes = get_all_node_records(snapshot);
     let (pks, certs) = get_all_nodes_public_keys_and_certs(snapshot)?;
 
     let mut ok_node_count = 0;
     let mut bad_node_count = 0;
     let mut maybe_error: Option<Result<(), InvariantCheckError>> = None;
-    for node_id in nodes.keys() {
+    for node_id in nodes.iter().map(|(k, _)| k) {
```
fspreiss: We used to iterate over entries in a BTreeMap, now this is a Vec. Wouldn't it make sense to let `get_all_node_records` return a BtreeMap, expressing on the type level that keys are unique (which they are, because they are derived from the RegistrySnapshot which itself is a BTreeMap.

Bownairo: This makes sense to me! Many of the other `common` `fn`s return `Vec`, but I'll leave those updates for a catchall followup PR.

fspreiss (approved):

rs/registry/canister/src/invariants/crypto.rs
fspreiss: Thanks for adapting the return type of `get_all_node_records` to a `BTreeMap`.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -23,8 +21,7 @@ use prost::Message;
 ///
 /// For each replica version that is either referred to in a SubnetRecord
 /// of a subnet listed in the subnet list, that is in use by an API boundary node,
-/// that is used by the unassigned nodes, or that is contained
-/// the BlessedReplicaVersions-List, the following is checked:
+/// that is used by the unassigned nodes, the following is checked:
```
jasonz-dfinity: Nit: "that is in use by an API boundary node or that is used by the unassigned nodes"

jasonz-dfinity (approved):

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -55,7 +55,7 @@ fn get_hostos_version_record(snapshot: &RegistrySnapshot, version: String) -> Ho
 /// Returns the list of HostOS versions where each version is referred to
 /// by at least one node.
 fn get_all_hostos_versions_of_nodes(snapshot: &RegistrySnapshot) -> Vec<String> {
-    get_node_records_from_snapshot(snapshot)
+    get_all_node_records(snapshot)
         .values()
         .filter_map(|node_record| node_record.hostos_version_id.clone())
```
pierugo-dfinity: I think you can avoid this clone
```suggestion
        .into_values()
        .filter_map(|node_record| node_record.hostos_version_id)
```
Sorry this is orthogonal to this PR.

pierugo-dfinity (approved):

PR #9782, opened 2026-04-09
feat: CON-1716 Allow `SetupInitialDkg` requests to be answered by subnets other than NNS

eichhorl:
## Background
`SetupInitialDkg` is a management canister call that can only be made by canisters residing on the NNS. Typically, this management canister call is triggered by proposals that create or recover a subnet. Similarly, it will also be used to create cloud engines. 

Upon receiving such a `SetupInitialDkg` request, the receiving subnet will perform two instances of the NiDKG protocol. The result yields the key material of the new or recovered subnet, from which also the subnet ID is derived.

## Problem
Today, `SetupInitialDkg` requests are exclusively handled by the NNS subnet. This has some disadvantages, which are exacerbated by the advent of cloud engines:

1. The code implementing the remote NiDKG protocol exclusively runs on NNS.
2. The protocol is comparatively expensive and puts some load on the NNS. 
3. So far, execution of this management canister call was gated by proposals. This may no longer be the case for cloud engines.
4. When handled on the NNS, cloud engine creations compete for the same resources as critical recovery proposals.
5. Problems surfacing on the NNS subnet may have higher impact compared to other subnets.

## Proposed Solution
Instead of always handling `SetupInitialDkg` request on the NNS, this PR proposes the ability to optionally specify the subnet to which the request should be routed. Specifically, this would allow the key material for cloud engines to be created on a different subnet than the one handling critical recovery proposals.

To do this, we extend the `SetupInitialDKGArgs` (and consequently, the `Create/RecoverSubnetPayload`) with an optional `SubnetId` field, to which the request will be routed.

Note that, although these changes have not been deployed to the mainnet NNS canisters yet, system tests using mainnet NNS canisters already pass. This is because the new argument is still ignored by the mainnet canisters, which means they fall back to the old behavior (`SetupInitialDkg` is routed to the NNS).

## Future Work
- Notify dashboard teams of the new arguments added to create and recover subnet proposals.
- Select a designated `SetupInitialDkg` subnet for handling Cloud Engine creations, for instance the II subnet.


basvandijk (approved):

alin-at-dfinity (approved):

pierugo-dfinity (approved):

rs/tests/nns/create_subnet_test.rs
```diff
@@ -112,18 +122,20 @@ pub fn test(env: TestEnv) {
 
         // Submit and adopt the configured number of create subnet proposals
         let mut proposal_ids = vec![];
-        for _ in 0..APP_SUBNETS {
+        for proposal_idx in 0..APP_SUBNETS {
             let nodes = unassigned_nodes.by_ref().take(APP_PRE_MASTER).collect();
+            let initial_dkg_subnet_id = (proposal_idx % 2 == 0).then_some(initial_dkg_subnet_id);
```
daniel-wong-dfinity-org: Old way (`if`) != bad way. Similarly, new way (dot chaining) != good way.

The problem here is that it requires esoteric knowledge (specifically, knowledge of an infrequently used method of bool, which rarely has methods called on it anyway, thus violating the Principle of Least Surprise). Furthermore, it is not significantly more concise that the primitive way. Verdict: the "power to weight ratio" here does not "make sense"; therefore, do it the primitive way, like so:

```suggestion
            let initial_dkg_subnet_id = if proposal_idx % 2 == 0 {
                Some(initial_dkg_subnet_id)
            } else {
                None
            };
```

These kinds of "optimizations" help the writer, but we should be _**gleefully**_ throwing the writer under the bus for the sake of the reader. Requiring esoteric knowledge is reader-hostile.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -112,18 +122,20 @@ pub fn test(env: TestEnv) {
...
             info!(
                 log,
-                "Submitting proposal to create subnet with nodes: {nodes:?}"
+                "Submitting proposal to create subnet with nodes: {nodes:?}, initial_dkg_subnet_id: {initial_dkg_subnet_id:?}"
```
daniel-wong-dfinity-org:
```suggestion
                "Submitting proposal to create subnet with nodes: {nodes:?},
                 initial_dkg_subnet_id: {initial_dkg_subnet_id:?}"
```

Even though we have large monitors (as opposed to the CRT bad ol' days), we should still limit our width so that people can COMFORTABLY have windows side by side, so that they can actually make good use of large monitors.

Perhaps, someday, rustfmt will be smart enough to take care of these for us. You may say, I'm a dreamer. But I'm not the only one.

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -114,8 +114,11 @@ impl Registry {
         };
 
         let create_cup_contents = |nodes| async {
-            let request =
-                SetupInitialDKGArgs::new(nodes, RegistryVersion::new(pre_call_registry_version));
+            let request = SetupInitialDKGArgs::new(
+                nodes,
+                RegistryVersion::new(pre_call_registry_version),
+                None, // Initial DKG request is handled by the NNS subnet
```
daniel-wong-dfinity-org, line 120: 🙏 This is good.

daniel-wong-dfinity-org (approved):
👍 for Governance.

PR #9800, opened 2026-04-10
feat(subnet-splitting): [CON-1569, CON-1625, CON-1626] add a python script for finding a good subnet split 

kpop-dfinity:
The python script uses [pulp](https://github.com/coin-or/pulp) to find a solution to the following MILP:

```
Constants:
1. load[c] - some notion of a load of a canister c, could be the number of ingress messages ingested, total cycles burned, etc (based on historical load)
2. msgs[c1, c2] - number of messages exchanged between canisters c1 and c2 (based on historical load)
3. MAX_CUTS - controls how fragmented the routing table we allow to be
4. epsilon - how close to the optimal load balancing we want to be

Variables:
1. t[c] ∊ {0, 1}, assignment of a canister to a subnet (1 = the canister stays on the original subnet, 0 = the canister is migrated to a new subnet)
2. z[c1, c2]  ∊ {0, 1}, denotes whether  canisters c1, c2 would be on different subnets
3. Load_0 = ∑ load[c] * (1 - t[c]), total estimated "load" on the subnet 0
4. Load_1 = ∑ load[c] * t[c], total estimated "load" on the subnet 1

Constraints:
1. ∑ z[c, c+1] <= MAX_CUTS, in order not to fragment the routing table too much
2. Load_0 <= (∑ load[c]) / 2  * (1 + epsilon), the load on subnet 0 is at most epsilon away from half of the total load 
4. Load_1 <= (∑ load[c]) / 2  * (1 + epsilon), the load on subnet 1 is at most epsilon away from half of the total load 
6. etc

Objective:
Minimize ∑ z[c1, c2]  * msgs[c1, c2]
```

Given load metrics of a subnet, we will be able to use this tool to decide which canisters to migrate to a new subnet.

# Example

Output of  `split_finder -- --help`

```
$ bazel run //rs/recovery/subnet_splitting/split_finder -- --help
usage: split_finder.py [-h] --load-path LOAD_PATH --communication-data-path COMMUNICATION_DATA_PATH --output-path OUTPUT_PATH --load-type
                       {instructions_executed,ingress_messages_executed,remote_subnet_messages_executed,local_subnet_messages_executed,http_outcalls_executed,heartbeats_and_global_timers_executed}
                       [--epsilon-load EPSILON_LOAD] [--max-cuts MAX_CUTS]

Run subnet splitting MILP.

options:
  -h, --help            show this help message and exit
  --load-path LOAD_PATH
                        Path to load data
--communication-data-path COMM_DATA_PATH
                        Path to canister-to-canister communication data
  --output-path OUTPUT_PATH
                        Path to output data
  --load-type {instructions_executed,ingress_messages_executed,remote_subnet_messages_executed,local_subnet_messages_executed,http_outcalls_executed,heartbeats_and_global_timers_executed}
                        Type of load to optimize for
  --epsilon-load EPSILON_LOAD
                        Allowed primary load deviation fraction
  --max-cuts MAX_CUTS   Maximum number of routing cuts
```

# Tests
Integrated the tool with the `subnet_splitting:integration_test`

# Credit
Credit: @bjoernek


rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -133,10 +171,10 @@ fn load_metrics_e2e_test() {
...
         // to the changes to the replicated state / execution. It's mostly a sanity check that the
         // returned values are not too ridiculous and they might have to be updated once in a while.
-        assert_near!(states_sizes_bytes.source, 4662083, 0.1);
-        assert_near!(states_sizes_bytes.destination, 4662072, 0.1);
-        assert_near!(instructions_executed.source, 7664259, 0.1);
-        assert_near!(instructions_executed.destination, 7680855, 0.1);
+        assert_near!(states_sizes_bytes.source, 4572729, 0.1);
+        assert_near!(states_sizes_bytes.destination, 4572699, 0.1);
+        assert_near!(instructions_executed.source, 7663691, 0.1);
+        assert_near!(instructions_executed.destination, 7664789, 0.1);
```
kpop-dfinity: Note: these numbers do not sum to the same numbers as before, because of the recent changes to the replicated state / execution (see the comment above)

basvandijk (approved):

rs/recovery/subnet_splitting/split_finder/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+    visibility = [
+        "//rs:release-pkg",
+        "//rs/recovery/subnet_splitting:__subpackages__",
+    ],
+    deps = [
+        requirement("matplotlib"),
+        requirement("networkx"),
+        requirement("pandas"),
+        requirement("pulp"),
+        requirement("seaborn"),
```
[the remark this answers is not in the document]
kpop-dfinity: done

rs/recovery/subnet_splitting/tests/integration_tests.rs
pierugo-dfinity: Should you update this comment now?

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +91,52 @@ fn load_metrics_e2e_test() {
...
-        // middle half of the canisters
-        let canister_id_ranges = vec![CanisterIdRange {
-            start: canister_ids[5],
-            end: canister_ids[14],
-        }];
+        info!(
+            logger,
+            "Using `split-finder` to find candidate canister ranges to be migrated"
+        );
+        // TODO(CON-1569): use actual connectivity metrics
```
pierugo-dfinity: How will the test look with the actual metrics? Do we plan to regularly collect metrics and test that the tool gives some acceptable split, or do we plan to collect them once and run the test always on the same inputs?

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +91,52 @@ fn load_metrics_e2e_test() {
...
+        let output = std::process::Command::new(split_finder_path)
+            .args(["--load-path", &load_samples_path.display().to_string()])
+            .args([
+                "--comm-data-path",
+                &communication_samples_path.display().to_string(),
+            ])
+            .args(["--output-path", &split_output_path.display().to_string()])
+            .args(["--load-type", "instructions_executed"])
+            .args(["--epsilon-load", &EPSILON.to_string()])
+            .args(["--max-cuts", "10"])
```
pierugo-dfinity:
```suggestion
            .args(["--max-cuts", &MAX_CUTS.to_string()])
```
Should we use a constant? Should we test that the resulting routing table indeed follows this and is not too fragmented?

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +91,52 @@ fn load_metrics_e2e_test() {
...
+        let split_finder_path =
+            std::env::var("SPLIT_FINDER_PATH").expect("SPLIT_FINDER_PATH not set");
+        let output = std::process::Command::new(split_finder_path)
+            .args(["--load-path", &load_samples_path.display().to_string()])
+            .args([
+                "--comm-data-path",
+                &communication_samples_path.display().to_string(),
+            ])
+            .args(["--output-path", &split_output_path.display().to_string()])
+            .args(["--load-type", "instructions_executed"])
```
pierugo-dfinity, line 122: Do you plan to test other load types?

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +91,52 @@ fn load_metrics_e2e_test() {
         let load_samples_path = dir.path().join("load_samples.csv");
         ic_state_tool::commands::canister_metrics::get(checkpoint_dir.clone(), &load_samples_path)
             .expect("Should compute canister metrics for a valid checkpoint");
+        let communication_samples_path = dir.path().join("comm_samples.csv");
+        let mut file = std::fs::File::create(&communication_samples_path).unwrap();
```
pierugo-dfinity:
```suggestion
        let mut comm_samples_file = std::fs::File::create(&communication_samples_path).unwrap();
```
Maybe? For clarity

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,57 @@
+from pathlib import Path
+from typing import Dict, List, Tuple
+
+import pandas as pd
+
+
+def load_subnet_data(load_path: Path, load_type: str, comm_data_path: Path) -> Dict:
```
pierugo-dfinity: It's hard to get a feeling of what the data looks like in the .csv files (i.e. what columns are there?). Maybe it could be useful to have some example .csv files somewhere. Same for the format of the variables in the returned dictionary. Maybe some docs could help here

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,57 @@
...
+
+    edges: List[Tuple[int, int, float]] = []
+    for _, row in comm_data.iterrows():
+        sender_id = row["sender_canister_id"]
+        receiver_id = row["receiver_canister_id"]
+        count = row["count"]
+        if sender_id in canister_id_to_index and receiver_id in canister_id_to_index:
+            i = canister_id_to_index[sender_id]
+            j = canister_id_to_index[receiver_id]
+            edges.append((i, j, count))
```
pierugo-dfinity: Is there a reason why not to follow Copilot's suggestion of vectorizing the operations here? An alternative (from Claude):
```python
# Filter to only rows where both sender and receiver are in the index map
mask = (
    comm_data["sender_canister_id"].isin(canister_id_to_index) &
    comm_data["receiver_canister_id"].isin(canister_id_to_index)
)
filtered = comm_data[mask]

# Map IDs to indices vectorially
edges: List[Tuple[int, int, float]] = list(zip(
    filtered["sender_canister_id"].map(canister_id_to_index),
    filtered["receiver_canister_id"].map(canister_id_to_index),
    filtered["count"],
))
```

pierugo-dfinity (commented):
Will go through the script in more detail tomorrow

rs/recovery/subnet_splitting/tests/integration_tests.rs
pierugo-dfinity: Do you plan to test with a secondary load metric?

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: Yes, thanks, done

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: Similarly, to how it's done with the load metrics. 
At some point, I'll expand the `state-tool` to also return the connectivity metrics of the tested subnet

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: > Do you plan to test other load types?

no, it's just sanity check. I assume that if it works for one column then it should work for the others as well

> Do you plan to test with a secondary load metric?

maybe in the future :D

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: Yes to both, thanks

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: Sure

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: I find the loop based approach a bit more readable.
And regarding copilot's comment:

```
Building edges via comm_data.iterrows() will be a bottleneck for large communication datasets (common for real subnets).
```

it's true that it might be a bottleneck for large datasets but it's *not* true that it's common for real subnets. Our subnets are very sparse in the sense that very few canisters talk to each other. I think python should be able handle iterating through a couple thousands of rows; and the majority of time will be spend in the solver anyways.

But if you prefer the vectorized option, I'll update the code

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: Added some comments, PTAL

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +92,51 @@ fn load_metrics_e2e_test() {
...
+        );
+        let canister_id_ranges: Vec<CanisterIdRange> = read_to_string(&split_output_path)
+            .expect("The split finder script should have produced a valid output")
+            .lines()
+            .map(|line| CanisterIdRange::from_str(line).expect("Not a valid CanisterIdRange"))
+            .collect();
+        assert!(
+            canister_id_ranges.len() <= MAX_CUTS,
+            "The split find script should have produced at most {MAX_CUTS} because we \
+            passed {MAX_CUTS} as the `--max-cuts` argument"
```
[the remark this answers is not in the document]
kpop-dfinity: incorrect

rs/recovery/subnet_splitting/tests/integration_tests.rs
pierugo-dfinity: Got it thanks

rs/recovery/subnet_splitting/tests/integration_tests.rs
pierugo-dfinity: Ty

rs/recovery/subnet_splitting/split_finder/data_io.py
pierugo-dfinity: > I find the loop based approach a bit more readable.

I agree. My comment was mostly about performance so if you think this is fine in production, then that's fine.

Does the communication data csv contain entries for all pairs of canisters (including those with a count of 0) or only pairs of canisters with non-zero count? If the former, then even if the subnet is sparse, we would still iterate over many rows.

> Our subnets are very sparse

For subnets like OpenChat too? I'm just a bit scared since the size of that dataframe will scale quadratically with the number of canisters.

pierugo-dfinity: > Our subnets are very sparse

I guess if we'd ever want to split a production subnet, this hypothesis might not be true anymore, as otherwise we would probably not need to split the subnet in the first place.

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +92,51 @@ fn load_metrics_e2e_test() {
...
+            str::from_utf8(&output.stdout).unwrap(),
+            str::from_utf8(&output.stderr).unwrap(),
+        );
+        let canister_id_ranges: Vec<CanisterIdRange> = read_to_string(&split_output_path)
+            .expect("The split-finder script should have produced a valid output")
+            .lines()
+            .map(|line| CanisterIdRange::from_str(line).expect("Not a valid CanisterIdRange"))
+            .collect();
+        assert!(
+            canister_id_ranges.len() <= MAX_CUTS,
```
pierugo-dfinity: Should this be something like

```suggestion
            canister_id_ranges.len() <= (MAX_CUTS + 1) / 2,
```

instead, since `canister_id_ranges` contains only half the ranges (the destination ones)?

rs/recovery/subnet_splitting/split_finder/data_io.py
pierugo-dfinity: I very much like the csv samples, thank you! Docs read good too 👍

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,58 @@
...
+    load_type -- which kind of load (e.g. instructions executed, ingress messages ingested) we want
+                 to balance for.
+    communication_data_path -- path to a file in the csv format which contains for each pair of
+                               canisters, the number of messages they exchanged. See
+                               `../test_data/fake_communication_sample.csv` for an example.
+
+    """
+    canister_data = pd.read_csv(load_path)
+    canister_data = canister_data[["canister_id", load_type]]
+    canister_data["original_index"] = range(len(canister_data))
```
pierugo-dfinity: Where do we use column `original_index`?

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,88 @@
...
+    prob = pulp.LpProblem("CanisterSubgroupAssignment", pulp.LpMinimize)
+
+    t = pulp.LpVariable.dicts("t", range(N_c), cat=pulp.LpBinary)
+    transitions = pulp.LpVariable.dicts("transition", range(N_c - 1), cat=pulp.LpBinary)
+
+    prob += pulp.lpSum([transitions[i] for i in range(N_c - 1)]) <= max_cuts, "MaxCuts"
+
+    for i in range(N_c - 1):
+        prob += transitions[i] >= t[i] - t[i + 1], f"TransitionMin_{i}"
+        prob += transitions[i] >= t[i + 1] - t[i], f"TransitionMax_{i}"
```
pierugo-dfinity: nit: Are `TransitionMin_` and `TransitionMax_` names appropriate? They are at least not consistent with `InterGroupComm1_`/`InterGroupComm2_` below

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,88 @@
...
+    addition to incoming messages).
+    Returns assignments and key metrics to allow testing and reuse.
+    """
+    N_c = len(load_c)
+    prob = pulp.LpProblem("CanisterSubgroupAssignment", pulp.LpMinimize)
+
+    t = pulp.LpVariable.dicts("t", range(N_c), cat=pulp.LpBinary)
+    transitions = pulp.LpVariable.dicts("transition", range(N_c - 1), cat=pulp.LpBinary)
+
+    prob += pulp.lpSum([transitions[i] for i in range(N_c - 1)]) <= max_cuts, "MaxCuts"
```
pierugo-dfinity: Do we need this `transitions` variable? Couldn't we replace this constraint with

```suggestion
    prob += pulp.lpSum([z[(i, i + 1)] for i in range(N_c - 1)]) <= max_cuts, "MaxCuts"
```

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,128 @@
...
+    "heartbeats_and_global_timers_executed",
+]
+
+################################################################
+# Main pipeline functions
+
+
+def compute_targets(load_c):
+    total_load_c = sum(load_c)
+    max_canister_load = max(load_c) if len(load_c) > 0 else 0
```
pierugo-dfinity: Shouldn't it be an error if `len(load_c) == 0`?

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,127 @@
...
+
+            if is_to_be_migrated:
+                if first_in_range is not None and last_in_range is not None:
+                    output_file.write(f"{first_in_range}:{last_in_range}\n")
+                first_in_range, last_in_range = None, None
+            else:
+                canister_id = index_to_canister_id[k]
+                last_in_range = canister_id
+                if first_in_range is None:
+                    first_in_range = canister_id
```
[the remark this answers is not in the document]
pierugo-dfinity: Isn't that true?

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: > Does the communication data csv contain entries for all pairs of canisters (including those with a count of 0) or only pairs of canisters with non-zero count? If the former, then even if the subnet is sparse, we would still iterate over many rows.

Only for the ones which have non-zero count

> For subnets like OpenChat too? I'm just a bit scared since the size of that dataframe will scale quadratically with the number of canisters.

One thing I forgot to mention is that since the communication metrics (will be) part of the replicated state (like the load metrics currently are) we won't keep track of *all* canisters a given canister is communicating with (because in the absolutely worst case we would need to store 120'000 * 120'000 entries, or even more in the future if we bump the bound on the number of canisters in a subnet), but we will only keep info about something like 100 neighbors for each canister (meaning that in the worst case we would keep track of 120'000 * 100 edges)
(see https://docs.google.com/document/d/1NaTwMpbuw1PoQnMZOk-uOTKLWoe-s9OAae22uoTtw4M/edit?pli=1&tab=t.0#heading=h.rf17b37irubq)

> I guess if we'd ever want to split a production subnet, this hypothesis might not be true anymore, as otherwise we would probably not need to split the subnet in the first place.

I could argue it's the opposite: if the subnet is dense, in the sense that most of the canisters talk to each other, meaning that we would have quadratic number of edges, then (unless the subnet is close to being bipartite) maybe we should not split the subnet, as the as the goal of the optimization is to keep the canisters which talk to each other on the same subnet as much as possible.

Anyways, a couple of months ago we did run some experiments for several subnets (I don't remember if OpenChat was one of them, though) on mainnet and there was no subnet which had more edges than the number of canisters. In general canisters don't talk to other canisters at all; unless it's a BOB-like subnet where we have 1 "orchestrator" canister which talks to thousands of other canisters who only talk to the "orchestrator" canister.

rs/recovery/subnet_splitting/split_finder/split_finder.py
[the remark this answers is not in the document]
kpop-dfinity: it's true, but it's irrelevant whether we return ranges or their complement here. 
I'll flip the bit in the PR description, though, for consistency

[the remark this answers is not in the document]
kpop-dfinity: (unless you prefer what Copilot proposes, I don't mind)

rs/recovery/subnet_splitting/tests/integration_tests.rs
kpop-dfinity: I don't think we need to be so precise, but sure, done :)

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: I don't think it's used anywhere anymore after all the cleanups I did, thanks!

rs/recovery/subnet_splitting/split_finder/solver_core.py
kpop-dfinity: I don't think they're appropriate, thanks, renamed it to be consistent with the other variables, tahks!

rs/recovery/subnet_splitting/split_finder/solver_core.py
kpop-dfinity: Good catch! I don't think we need it. Removed the extra variables.

rs/recovery/subnet_splitting/split_finder/split_finder.py
kpop-dfinity: Good idea, done

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,79 @@
...
+    t = pulp.LpVariable.dicts("t", range(canister_count), cat=pulp.LpBinary)
+
+    z = {}
+    # for each pair of consecutive canisters which don't talk to each other, we add an edge of
+    # 0 weight, so that we can later add an constraint that the number of pairs of consecutive
+    # canisters is bounded
+    for i in range(canister_count - 1):
+        if (i, i + 1) not in edges:
+            edges[(i, i + 1)] = 0
+
```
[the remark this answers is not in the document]
kpop-dfinity: risk accepted

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if is_to_be_migrated:
+            if first_in_range is not None and last_in_range is not None:
+                ranges.append((first_in_range, last_in_range))
+            first_in_range, last_in_range = None, None
+        else:
+            canister_id = index_to_canister_id[k]
+            last_in_range = canister_id
+            if first_in_range is None:
+                first_in_range = canister_id
```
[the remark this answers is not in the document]
kpop-dfinity: duplicate

rs/recovery/subnet_splitting/split_finder/split_finder.py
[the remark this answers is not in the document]
pierugo-dfinity: I think it matters depending on what the script is supposed to output: do we want to output the source or the destination ranges? The current implementation outputs the source ranges, but I would have assumed we would want to output the destination ranges, since this is what the registry endpoint [expects](https://sourcegraph.com/r/github.com/dfinity/ic@446004ea9b91a7ca0c730ded988d1b06e288a3a8/-/blob/rs/registry/canister/src/mutations/do_split_subnet.rs?L402).

You could also mention that in the help text of `--output-path`, instead of `"Path to output data"`.

P.S.: you could also argue that source vs destination doesn't really make a difference, but maybe it does? At least this is what I can interpret from the initialization of `target_load_0`: If there is a single canister responsible for the majority of the load, we would want to keep it in the source subnet instead of migrating it?
I don't have the full context, so maybe there is indeed no difference at all between source vs destination ranges, LMK.

rs/recovery/subnet_splitting/split_finder/data_io.py
pierugo-dfinity: > we will only keep info about something like 100 neighbors

Ah ok great, then it falls back to being linear instead of quadratic. Thanks for the context.

> I could argue it's the opposite

Right, indeed, I see.

Thanks for the details, LGTM then!

rs/recovery/subnet_splitting/split_finder/split_finder.py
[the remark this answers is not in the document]
kpop-dfinity: What I meant in `it's true, but it's irrelevant whether we return ranges or their complement here.`  is that the problem is symmetric* in the sense that if (A, B) is an optimal solution to the problem (where A is the list of canister ranges to stay on the subnet and B is the list of canister ranges to leave the subnet) then (B, A) is an optimal solution as well. So it shouldn't matter whether we return A or B.

\* but like you rightfully pointed out, the problem is not fully symmetric because we have different load targets for subnets 0 and 1, if there is a heavy canister. I changed the code, so the problem is fully symmetric now and we no longer insist that the heavy canister will land on specific subnet, PTAL

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,74 @@
...
+    for (i, j), weight in edges.items():
+        z[(i, j)] = pulp.LpVariable(f"z_{i}_{j}", cat="Binary")
```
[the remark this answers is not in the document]
kpop-dfinity: it's ok

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,74 @@
...
+    edges: Dict[Tuple[int, int], int],
+    max_cuts: int,
+) -> Dict:
+    """
+    Build and solve the MILP for a given load vector(s) and communication edges.
+    Returns the subnet assignment for each of the canisters.
+    """
+    if len(load_constraints) == 0:
+        raise ValueError("The provided load constraints data is empty")
+
```
pierugo-dfinity:
```suggestion
if any(len(constraint.canister_loads) != len(load_constratins[0].canister_loads) for constraint in load_constraints):
    raise ValueError("All load constraints must have the same number of canisters")
```
Maybe

rs/recovery/subnet_splitting/split_finder/solver_core.py
```diff
@@ -0,0 +1,74 @@
...
+    for load_constraint in load_constraints:
+        load_0 = pulp.lpSum([load_constraint.canister_loads[k] * (1 - t[k]) for k in range(canister_count)])
+        load_1 = pulp.lpSum([load_constraint.canister_loads[k] * t[k] for k in range(canister_count)])
+        max_allowed_load = load_constraint.max_allowed_load_per_subnet
+        epsilon = load_constraint.epsilon
+        name = load_constraint.name
+
+        problem += load_0 <= max_allowed_load * (1 + epsilon), f"LoadUpper_0_{name}"
+        problem += load_1 <= max_allowed_load * (1 + epsilon), f"LoadUpper_1_{name}"
```
pierugo-dfinity, line 67: Removing the lower bound constraint was intentional right? (I can see how it could indeed be redundant)

pierugo-dfinity: Ah right, it's actually necessary, since `max_allowed_load` could be higher than half (in the case of a hungry canister).

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ValueError("The provided load data is empty")
+
+    total_canister_loads = sum(canister_loads)
+    max_canister_load = max(canister_loads)
+    average_load = total_canister_loads / 2
+    # If there is a single canister which has is responsible for majority of the load on the subnet
+    # it will be impossible to split the subnet in such a way that the load is balanced equally
+    # across both of the post-split subnets. In that case we relax the constraints a bit.
+    max_allowed_load = max(max_canister_load, average_load)
+    return total_canister_loads, max_canister_load, average_load, max_allowed_load
```
pierugo-dfinity: Do we need to return anything else than `max_allowed_load`?

rs/recovery/subnet_splitting/split_finder/split_finder.py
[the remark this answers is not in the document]
pierugo-dfinity: Right. With your changes, I agree it's symmetric now. It also reads much nicer!

> You could also mention that in the help text of --output-path, instead of "Path to output data".

I still think the help text could be a bit more expressive :D

rs/recovery/subnet_splitting/split_finder/solver_core.py
kpop-dfinity: Yeah, exactly

rs/recovery/subnet_splitting/split_finder/split_finder.py
kpop-dfinity: No. What kind backwards language doesn't warn about unused variables, smh.

Thanks, fixed

rs/recovery/subnet_splitting/split_finder/solver_core.py
kpop-dfinity: Good idea

rs/recovery/subnet_splitting/src/post_split_estimations.rs
```diff
@@ -191,9 +191,15 @@ pub struct Estimates {
     pub destination: u64,
 }
 
+impl Estimates {
+    pub fn total(&self) -> u64 {
+        self.source + self.destination
```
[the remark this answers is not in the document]
kpop-dfinity: I'll take the risk

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,57 @@
+from pathlib import Path
+from typing import Dict, Tuple
+
+import pandas as pd
+
+
+def load_subnet_data(load_path: Path, load_type: str, communication_data_path: Path) -> Dict:
+    """
+    Load canister metadata and communication edges for a subnet.
+    Returns canister edge list (i, j, weight), load vector, and index mappings.
```
pierugo-dfinity: `edges` is not a list anymore :D

rs/recovery/subnet_splitting/split_finder/split_finder.py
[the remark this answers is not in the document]
kpop-dfinity: DONE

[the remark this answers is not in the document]
kpop-dfinity: I mean, done

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,57 @@
...
+
+    edges: Dict[Tuple[int, int], int] = {}
+    for _, row in communication_data.iterrows():
+        sender_id = row["sender_canister_id"]
+        receiver_id = row["receiver_canister_id"]
+        count = row["count"]
+        if sender_id in canister_id_to_index and receiver_id in canister_id_to_index:
+            i = canister_id_to_index[sender_id]
+            j = canister_id_to_index[receiver_id]
+            edges[(i, j)] = count
```
pierugo-dfinity: Shouldn't the program crash loudly if the `if` is `false`? In which case the entire loop body could be simplified to

```suggestion
        sender_id = canister_id_to_index[row["sender_canister_id"]]
        receiver_id = canister_id_to_index[row["receiver_canister_id"]]
        count = row["count"]

        edges[(sender_id, receiver_id)] = count
```

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,127 @@
...
+    return max_allowed_load
+
+
+def find_split(path: Path, communication_data_path: Path, load_type: str, epsilon_load: float, max_cuts: int):
+    data = load_subnet_data(path, load_type, communication_data_path)
+    edges = data["edges"]
+    load = data["load"]
+    index_to_canister_id = data["index_to_canister_id"]
+
+    max_allowed_load_per_subnet = compute_targets(load)
```
pierugo-dfinity: nit: maybe this function's name is now outdated

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -89,16 +92,51 @@ fn load_metrics_e2e_test() {
...
+        let output = std::process::Command::new(split_finder_path)
+            .args(["--load-path", &load_samples_path.display().to_string()])
+            .args([
+                "--communication-data-path",
+                &communication_samples_path.display().to_string(),
+            ])
+            .args(["--output-path", &split_output_path.display().to_string()])
+            .args(["--load-type", "instructions_executed"])
+            .args(["--epsilon-load", &EPSILON.to_string()])
+            .args(["--max-cuts", &MAX_CUTS.to_string()])
```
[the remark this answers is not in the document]
kpop-dfinity: done

rs/recovery/subnet_splitting/split_finder/split_finder.py
```diff
@@ -0,0 +1,120 @@
...
+    first_in_range = None
+    last_in_range = None
+    for k in range(len(assignments)):
+        is_to_be_migrated = assignments[k] == 1
+
+        if is_to_be_migrated:
+            if first_in_range is not None and last_in_range is not None:
+                ranges.append((first_in_range, last_in_range))
+            first_in_range, last_in_range = None, None
+        else:
```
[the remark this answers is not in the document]
kpop-dfinity: changed

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: yes yes

rs/recovery/subnet_splitting/split_finder/data_io.py
kpop-dfinity: Good point, done

rs/recovery/subnet_splitting/split_finder/split_finder.py
kpop-dfinity: Yep

rs/recovery/subnet_splitting/split_finder/data_io.py
```diff
@@ -0,0 +1,58 @@
...
+
+    communication_data = pd.read_csv(communication_data_path)
+    communication_data = communication_data[["sender_canister_id", "receiver_canister_id", "count"]]
+
+    communicating_canister_ids = set(communication_data["sender_canister_id"]).union(
+        set(communication_data["receiver_canister_id"])
+    )
+    communicating_canisters = canister_data[canister_data["canister_id"].isin(communicating_canister_ids)].reset_index(
+        drop=True
+    )
```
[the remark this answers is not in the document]
kpop-dfinity: @bjoernek do you remember why we had this filter? Is it because the optimization problem became too large for the subnets we were testing, and the solver was taking too much time to come up with a solution?

rs/recovery/subnet_splitting/split_finder/split_finder_test.py
```diff
@@ -0,0 +1,54 @@
+import unittest
+from pathlib import Path
+
+from data_io import load_subnet_data
+from split_finder import find_split
+
+TEST_DATA_DIR = Path(__file__).resolve().parents[1] / "test_data"
+FAKE_LOAD_SAMPLE_CSV_PATH = str(TEST_DATA_DIR / "fake_load_sample.csv")
+FAKE_LOAD_BASELINE_SAMPLE_CSV_PATH = str(TEST_DATA_DIR / "fake_load_baseline_sample.csv")
```
pierugo-dfinity, line 9: Does this file exist?

kpop-dfinity: Oups, now it does

pierugo-dfinity (approved):
Looks great! 🚀

rs/recovery/subnet_splitting/split_finder/data_io.py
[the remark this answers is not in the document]
kpop-dfinity: I've created a JIRA ticket for this (https://dfinity.atlassian.net/browse/CON-1713) 

I'll merge this PR and will potentially make another PR addressing this issue

kpop-dfinity:
> Looks great! 🚀

🤓 I think I'll switch to python 100%

PR #9801, opened 2026-04-10
fix: deflake //rs/bitcoin/ckbtc/minter:ckbtc_minter_tests

basvandijk:
## Root Cause

The test `should_cancel_and_reimburse_large_withdrawal` deposits 2,000 UTXOs which takes so long to process that the overall test suite times out. In all 3 flaky runs from the past week, this test was the one that never completed before the timeout.

## Fix

Reduce the number of UTXOs from 2,000 to 1,100 and the withdrawal amount from 1,800×deposit_value to 1,001×deposit_value. The test still properly validates the `TooManyInputs` error path (1,001 inputs > 1,000 max) but processes ~45% fewer UTXOs, preventing the timeout.

## Verification

Ran `bazel test --runs_per_test=3 --jobs=3` — all 3 runs passed (avg 45.0s, max 45.8s).

---
*This PR was created following the steps in `.claude/skills/fix-flaky-tests/SKILL.md`.*

### DIFF
--- a/rs/bitcoin/ckbtc/minter/tests/tests.rs
+++ b/rs/bitcoin/ckbtc/minter/tests/tests.rs
@@ -2952,18 +2952,21 @@ fn should_cancel_and_reimburse_large_withdrawal() {
         subaccount,
     };
 
-    // Step 1: deposit a lot of small UTXOs
-    const NUM_UXTOS: usize = 1_100;
+    // Step 1: deposit enough small UTXOs to exceed the max inputs limit.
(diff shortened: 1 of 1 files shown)


mducroux:
For reference, we had a similar flaky test for ckDOGE which was solved here: https://github.com/dfinity/ic/pull/9011

mducroux (approved):
LGTM!

gregorydemay (approved):

PR #9803, opened 2026-04-10
refactor: Restrict which CompoundCycles specializations can be refundable

dsarlis:
Restrict which type specializations of `CompoundCycles` can be refundable, i.e. they are prepaid for and then they would need to be refunded via a call to `SystemState::refund_cycles`. This hardens the code and ensures that the decision of which ones are refundable and thus can be provided to `SystemState::refund_cycles` is an explicit choice of the code author.

### DIFF
--- a/rs/replicated_state/src/canister_state/tests.rs
+++ b/rs/replicated_state/src/canister_state/tests.rs
@@ -896,7 +896,7 @@ fn update_balance_and_consumed_cycles_correctly() {
     );
     assert_eq!(
         system_state.canister_metrics().consumed_cycles(),
-        NominalCycles::new(initial_consumed_cycles.get())
+        prepaid_cycles.nominal()
     );
 
     let refund = CompoundCycles::<Instructions>::new(Cycles::new(100), cost_schedule);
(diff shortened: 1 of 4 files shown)


rs/replicated_state/src/canister_state/tests.rs
```diff
@@ -889,26 +888,26 @@ fn update_balance_and_consumed_cycles_correctly() {
...
+        CompoundCycles::<Instructions>::new(initial_consumed_cycles, cost_schedule);
     system_state.consume_cycles(prepaid_cycles);
     assert_eq!(
         system_state.balance(),
         INITIAL_CYCLES - initial_consumed_cycles
     );
     assert_eq!(
         system_state.canister_metrics().consumed_cycles(),
-        NominalCycles::new(initial_consumed_cycles.get())
+        prepaid_cycles.nominal()
```
dsarlis, line 899: This and following similar changes are drive-by simplifications since I had to touch the tests to update them to use valid refundable `CyclesUseCase`.

maksymar (approved):

dsarlis:
@basvandijk or @maksymar Please restart CI as I had to fix a small test issue and things should be good to be merged if everything passes and an interface-owner stamps as well.

schneiderstefan (approved):

PR #9804, opened 2026-04-10
fix: Upgrade libcryptsetup-rs and patch segfault bug

frankdavid:
The crate has a bug causing segfaults which was reported here: https://github.com/stratis-storage/libcryptsetup-rs/issues/473

This PR patches it for the IC repo.


bazel/libcryptsetup-rs.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/src/settings.rs b/src/settings.rs
```
basvandijk, line 3: Note we always include a reference to the upstream issue / PR in our `.patch` files so that we know when we can remove the patch again:
```suggestion
# Fixes CryptKeyslotHandle::get_pbkdf() segfault for type Argon2I/Argon2Id keyslots
# See: https://github.com/stratis-storage/libcryptsetup-rs/issues/473
diff --git a/src/settings.rs b/src/settings.rs
```

basvandijk (approved):

PR #9806, opened 2026-04-10
fix: [DSM-124] Clamp cycles burned to available balance

alin-at-dfinity:
Clamp the amount burned by `ic0_cycles_burn128()` to the available cycle balance minus freezing limit. This fixes a regression that caused validation errors in `apply_changes()` leading to rollbacks.

### DIFF
--- a/rs/embedders/tests/system_api.rs
+++ b/rs/embedders/tests/system_api.rs
@@ -1445,6 +1445,47 @@ fn call_perform_not_enough_cycles_does_not_trap() {
     assert_eq!(call_context_manager.callbacks().len(), 0);
 }
 
+/// If the canister does not have the requested number of cycles to burn, then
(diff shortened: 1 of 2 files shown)


rs/embedders/tests/system_api.rs
```diff
@@ -1445,6 +1445,47 @@ fn call_perform_not_enough_cycles_does_not_trap() {
     assert_eq!(call_context_manager.callbacks().len(), 0);
 }
 
+/// If the canister does not have the requested number of cycles to burn, then
+/// it clamps the amount to the available cycles (minus freeze threshold).
```
schneiderstefan: How about a test that includes a freeze theshold?

alin-at-dfinity: Done, although it's not exactly pretty.

schneiderstefan (approved):

PR #9809, opened 2026-04-10
test(delegations): [CON-1696] re-enable mainnet variant delegation system test for cloud engines

pierugo-dfinity:
#9613 has reached mainnet NNS and unblocks the mainnet variant of the delegation system tests for cloud engines.

### DIFF
--- a/rs/tests/networking/nns_delegation_test.rs
+++ b/rs/tests/networking/nns_delegation_test.rs
@@ -778,8 +778,6 @@ fn upgrade_non_nns_subnets_if_necessary(env: &TestEnv) {
             .iter()
             .copied()
             .filter(|t| *t != SubnetType::System)
-            // TODO(CON-1696): Remove next line when #9613 reaches mainnet NNS
-            .filter(|t| *t != SubnetType::CloudEngine)
             .map(|subnet_type| {
                 let env = env.clone();
                 let nns_node = nns_node.clone();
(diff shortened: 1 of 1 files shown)


rs/tests/networking/nns_delegation_test.rs
```diff
@@ -806,11 +804,7 @@ macro_rules! systest_all_subnet_types {
         $group = $group.add_test(systest!($function_name; SubnetType::System));
         $group = $group.add_test(systest!($function_name; SubnetType::Application));
         $group = $group.add_test(systest!($function_name; SubnetType::VerifiedApplication));
-        // TODO(CON-1696): Remove this condition (and always run the test for cloud engines) when
-        // #9613 reaches mainnet NNS
-        if get_guestos_img_version() == get_guestos_update_img_version() {
-            $group = $group.add_test(systest!($function_name; SubnetType::CloudEngine));
-        }
+        $group = $group.add_test(systest!($function_name; SubnetType::CloudEngine));
```
kpop-dfinity, line 807: I forgot: why can't we do it in a loop?

kpop-dfinity (approved):
LGTM

rs/tests/networking/nns_delegation_test.rs
pierugo-dfinity: It's a limitation of the `systest!` macro which doesn't move the argument inside the closure. With a loop, we get `closure may outlive the current function, but it borrows subnet_type, which is owned by the current function`

pierugo-dfinity: #9853

PR #9810, opened 2026-04-10
refactor(delegations): [CON-1701] use subnet type to know whether to contact an API BN

pierugo-dfinity:
[As suggested](https://github.com/dfinity/ic/pull/9595#discussion_r3020102210), this PR uses the subnet type instead of the node reward type to determine whether we should contact an API BN or an NNS node directly when fetching delegations.


kpop-dfinity (approved):
Nice!

PR #9811, opened 2026-04-10
fix(governance): Do not let snapshot proposal execution deadlock when the target is Governance

daniel-wong-dfinity-org:
Deadlock is avoided by performing the operation IN THE BACKGROUND when Governance is the target. For other targets, the Root behaves as before: synchronously.

# How Deadlock Would Occur Otherwise

1. Governance calls Root, and waits for Root to reply.
2. Root orders Governance to stop. Stop does not complete while the canister (in this case, Governance) is waiting for calls to complete.

This is circular waiting, deadlock.

# How Background Avoids Deadlock

By doing the operation in spawn_migratory, the second wait (where Root is waiting for Governance to stop) does not happen. Thus, the first wait (where Governance waits for Root to reply) is able to terminate.

# Changes in Greater Detail

Renamed `exclusively_stop_and_start_canister` to `perform_offline_canister_maintenance`. This is not strictly necessary; the old name is not wrong now, but the new name seems better, because it shifts emphasis away from secondary operations (stop, lock, start, etc.) to the main thing, i.e. maintenance of some kind, such as upgrade, taking snapshot, loading snapshot, etc.

The main enhancement to this function is that it now behaves a little bit differently in the special case where the target canister is Governance. In this special case, the operation is performed in the background (via spawn_migratory).

Also, in this case, "success" is defined as "we were able to launch this operation in the background", not "the operation itself went well".

This is why we need the new `Optimistic` trait: it is used when we need to report success in this much weaker definition of success.

Added some helpers to `ic_nns_test_utils`, very similar to existing ones:

* `nns_get_proposal_info` - Unlike other similar helpers, this returns `Option<ProposalInfo>` instead of `ProposalInfo`, because we (for the first time?) needed a way to query a proposal that does NOT exist.

* `nns_execute_proposal` - This probably should have existed before. It optimizes a very common case: TEST_NEURON_1 has majority voting power, and can therefore be used to make a proposal that passes right away + the test expects the proposal to be executed successfully + the test does not care about the title, summary, url.

# Testing

Made a copy of an existing test for snapshots (via proposal). In the new test, the target canister is Governance (instead of Ledger). The code under test is verified by verifying that after loading a snapshot (via proposal) old proposals are still there, but new ones are not.

# References

Closes [NNS1-4321].

[NNS1-4321]: https://dfinity.atlassian.net/browse/NNS1-4321

### DIFF
--- a/rs/nervous_system/root/src/load_canister_snapshot.rs
+++ b/rs/nervous_system/root/src/load_canister_snapshot.rs
@@ -1,8 +1,10 @@
-use crate::private::exclusively_stop_and_start_canister;
+use crate::{LOG_PREFIX, private::exclusively_stop_and_start_canister};
 use candid::CandidType;
 use ic_base_types::{CanisterId, PrincipalId, SnapshotId};
+use ic_cdk::{futures::spawn_017_compat, println};
 use ic_management_canister_types_private::LoadCanisterSnapshotArgs;
 use ic_nervous_system_clients::management_canister_client::ManagementCanisterClient;
+use ic_nns_constants::GOVERNANCE_CANISTER_ID;
 use serde::Deserialize;
 
 #[derive(Clone, Eq, PartialEq, Hash, Debug, CandidType, Deserialize)]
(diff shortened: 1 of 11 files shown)


rs/nervous_system/root/src/load_canister_snapshot.rs
```diff
@@ -26,71 +26,89 @@ pub struct LoadCanisterSnapshotError {
     pub description: String,
 }
 
+/// Restores a canister to a previously taken snapshot, an earlier state,
+/// which includes its memory (normal and stable) and code.
+///
+/// When the target canister is Governance, the operation is performed in the
```
jasonz-dfinity, line 32: NCR. 

While I think it's OK to take a snapshot of Governance, I cannot think of a safe way to load a snapshot for NNS Governance. As soon as such proposal being made, one can disburse a dissolved neuron and do it again once the snapshot is loaded.

OTOH, I'm sure such consideration will be made if we have to do such operation, so perhaps it's OK. I thought about documenting it here, but I doubt it would provide much value as I don't expect people to read the comments here before making such proposal.

jasonz-dfinity (approved):

rs/nervous_system/root/src/load_canister_snapshot.rs
daniel-wong-dfinity-org: Yes, when loading a snapshot, you ALWAYS lose data from after the snapshot was taken, and in general data is GOLD.

PR #9816, opened 2026-04-13
revert: "fix: move bazel-build-all-no-cache to schedule-daily"

cgundy:
Reverts dfinity/ic#9788

I didn't catch this earlier, but the workflow is invalid with the error `(Line: 57, Col: 13): Job 'bazel-build-all-no-cache' depends on unknown job 'config'.` - let's revert this first and see if the rate limit is still an issue.


basvandijk (approved):

cgundy:
Closed in favor of https://github.com/dfinity/ic/pull/9820

PR #9817, opened 2026-04-13
chore(IDX): split out config and bazel targets

cgundy:
This splits out the `config` job from `infer bazel targets`. The motivation is that `infer bazel targets` takes 2-3 minutes to run and the `Bazel Test arm64-` jobs only need the config job, so they can start 2 minutes earlier without waiting.

If this seems like too much added complexity for a minor gain I can close this PR again, but thought it was worth the suggestion.


basvandijk (approved):
Makes sense.

PR #9819, opened 2026-04-13
feat: support non-SEV bare metal in baremetal_app_subnet testnet

shilingwang:
### Summary
The interactive bare-metal application-subnet testnet is renamed to `baremetal_app_subnet` and no longer assumes AMD SEV-SNP. SetupOS / nested config uses the trusted-execution (TEE) path only when `TRUSTED_EXECUTION_ENVIRONMENT` is present in the environment, consistent with `nested::setup`.

### Changes
Testnet driver (`baremetal_app_subnet.rs`): Chooses create_bare_metal_node(..., enable_tee) from `TRUSTED_EXECUTION_ENVIRONMENT`; documents SEV vs non-SEV setup and ict examples.
Bazel (`BUILD.bazel`): Target renamed to baremetal_app_subnet; drops hard-coded TRUSTED_EXECUTION_ENVIRONMENT=true; inherits `BARE_METAL_HOST_SECRETS` and TRUSTED_EXECUTION_ENVIRONMENT from the caller environment when using `bazel test / --test_env`.
Nested helpers (`nested/src/lib.rs`): Adds create_bare_metal_node with an explicit enable_trusted_execution_environment flag; create_bare_metal_tee_node remains as a thin wrapper for existing callers (e.g. sev_recovery).
Cargo (`testnets/Cargo.toml`): Binary name aligned with the new testnet.
### How to run
SEV-SNP host: set `TRUSTED_EXECUTION_ENVIRONMENT` (e.g. true) in addition to `BARE_METAL_HOST_SECRETS`.
Non-SEV bare metal: set only `BARE_METAL_HOST_SECRETS`; ensure `TRUSTED_EXECUTION_ENVIRONMENT` is not inherited if your shell usually exports it.
### Breaking change
ict testnet create / Bazel target name changes from `sev_baremetal_app_subnet` to `baremetal_app_subnet`.
SEV runs no longer get TEE enabled by default from the BUILD file; you must pass or export TRUSTED_EXECUTION_ENVIRONMENT when you want the SEV-SNP path.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -246,7 +246,7 @@ go.sum                    @dfinity/idx
 /rs/tests/testnets/mainnet_nns/                         @dfinity/consensus
 /rs/tests/testnets/mainnet_nns.rs                       @dfinity/consensus
 /rs/tests/testnets/nns_recovery.rs                      @dfinity/consensus
-/rs/tests/testnets/baremetal_app_subnet.rs          @dfinity/idx @dfinity/node
+/rs/tests/testnets/baremetal_app_subnet.rs              @dfinity/idx @dfinity/node
 /rs/tests/driver/src/driver/simulate_network.rs         @dfinity/consensus @dfinity/idx
 /rs/tests/boundary_nodes/                               @dfinity/node
 /rs/tests/ckbtc/                                        @dfinity/defi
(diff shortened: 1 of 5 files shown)


.github/CODEOWNERS
```diff
@@ -246,7 +246,7 @@ go.sum                    @dfinity/idx
 /rs/tests/testnets/mainnet_nns/                         @dfinity/consensus
 /rs/tests/testnets/mainnet_nns.rs                       @dfinity/consensus
 /rs/tests/testnets/nns_recovery.rs                      @dfinity/consensus
-/rs/tests/testnets/sev_baremetal_app_subnet.rs          @dfinity/idx @dfinity/node
+/rs/tests/testnets/baremetal_app_subnet.rs          @dfinity/idx @dfinity/node
```
basvandijk: Could you fix the indentation?

basvandijk (approved):

.github/CODEOWNERS
shilingwang: Done.

r-birkner (approved):

PR #9821, opened 2026-04-13
feat: CON-1707 Limit the number of `SetupInitialDkg` contexts in flight

eichhorl:
`SetupInitialDkg` is a management canister call used to generate the subnet's initial key material. It is typically called during subnet creation/recovery. 

In the future it will also be used to create cloud engines, which may not necessarily be gated behind the adoption of a governance proposal. For that reason, we should limit the number of open contexts, similar to other management canister calls handled by consensus.

This PR limits the number of open `SetupInitialDkg` to 20, although the exact number may change in the future.


michael-weigelt (commented):
Rate limits introduce DOS vectors. Does this endpoint have a cost for the caller? If not, then perhaps this PR is making the problem worse, because an attacker can completely stop anyone else from calling it.

PR #9822, opened 2026-04-13
feat(ic-admin): support creation of cloud engines in ic-admin

r-birkner:
* CloudEngine subnets now use the same default config params as Application subnets in `get_default_config_params`. Previously they fell through to the NNS defaults.
* Added `--subnet-admins` flag to `propose-to-create-subnet` so subnet admin principals can be specified at creation time, instead of being hardcoded to an empty list.
* Fixed stale `--subnet-type` help text to list all four valid values.


basvandijk (approved):

rs/registry/admin/bin/create_subnet.rs
```diff
@@ -170,6 +170,10 @@ pub(crate) struct ProposeToCreateSubnetCmd {
     #[clap(long)]
     pub features: Option<SubnetFeatures>,
 
+    /// The list of principals that are subnet admins.
+    #[clap(long, num_args(1..))]
```
jasonz-dfinity, line 174: Why `num_args(1..)`? IIUC this would require any subnet creation proposal to have >=1 subnet admins. If so, is that what you want?

r-birkner: Thanks for pointing this out: I have double-checked that it works as intended. For an arg of type `Vec<X>` the argument is not mandatory. So, it is fine to not specify `--subnet-admins`, but if you specify it, you need to have at least one. I hope that makes sense :)

jasonz-dfinity: You are right!

jasonz-dfinity (approved):

PR #9829, opened 2026-04-13
chore: CON-1441 Remove `signature_agreements` from `IDkgPayload` and begin `Hash` migration

eichhorl:
After https://github.com/dfinity/ic/pull/9481, the `signature_agreements` field of the `IDkgPayload` no longer holds any data (signatures are included in the `ChainKeyPayload` instead). Also see the following metrics: [1](https://victoria.ch1-obs1.dfinity.network/select/0/vmui/#/?g0.range_input=7d&g0.end_input=2026-04-13T12%3A02%3A50&g0.relative_time=last_7_days&g0.tab=0&g0.tenantID=0&g0.expr=idkg_payload_metrics%7Btype%3D%22signature_agreements%22%7D+%3E+0), [2](https://victoria.ch1-obs1.dfinity.network/select/0/vmui/#/?g0.range_input=7d&g0.end_input=2026-04-13T12%3A04%3A45&g0.relative_time=last_7_days&g0.tab=0&g0.tenantID=0&g0.expr=consensus_threshold_signature_agreements+%3E+0).

Consequently, this PR removes the field entirely from the protobuf definition and the corresponding rust struct. On the protobuf level, this change is forward/backward compatible, since the serialization of an empty vector of signature agreements is indistinguishable from the field being absent entirely.

However, this poses a challenge for the compatibility of the rust struct: For the derived hash function of the struct, by default, it _does_ make a difference whether the field is absent or set to an empty value. This means, after deserialization, we _need_ to be able to distinguish whether the payload was serialized with an empty or an absent field, in order to calculate the hash correctly. 

To solve this, we introduce a boolean flag to the rust and protobuf definitions of the `IDkgPayload`: If the flag is `false` (the default), then we should include the empty signature agreements field in the hash. If the flag is `true`, then that means we should ignore the field.

Currently, the flag is set to `false` for all old and new `IDkgPayload`s (leading to the same hash as before). In a future release, we will start setting it to `true` for all new payloads. Following that, we will remove the custom `Hash` implementation in a subsequent release (all payloads are now considered to have an absent `signature_agreements` field, regardless of the flag value). In the final release we will then remove the flag again.

Also see a prior instance of this process here: [1](https://github.com/dfinity/ic/commit/0bc85d7c375845d5024ff4638606bff103970d48), [2](https://github.com/dfinity/ic/commit/1b8c37441353cd6ab0b59211158c6a87ccfe6c85), [3](https://github.com/dfinity/ic/commit/60177900fe5929d31fa84f549067d3fdf25b7b4c), [4](https://github.com/dfinity/ic/commit/6057ce233e0decc75cfabc7db32b1963956df10f)


alin-at-dfinity (approved):
Protobuf changes LGTM.

pierugo-dfinity (approved):

PR #9834, opened 2026-04-13
feat(IDX): check critical workflows

cgundy:
This introduces a workflow to check if any critical workflows have failed to run in the last 24 hours. This can happen sometimes if an error is introduced in those workflows and we don't catch it.


basvandijk (commented):
Have you considered implementing the alert using [Superset](https://superset.idx.dfinity.network/alert/list/)?

Any pros and cons versus the two approaches?

cgundy:
> Superset

Ah true! Let me take a look at that :)

cgundy:
Implemented in superset.

PR #9835, opened 2026-04-13
fix: deflake //rs/tests/networking:network_large_test

basvandijk:
The `//rs/tests/networking:network_large_test` [ran]( https://dash.zh1-idx1.dfinity.network/invocation/3b17bb9e-acf3-4e24-95ab-96117657fc9b?target=//rs/tests/networking:network_large_test#@1242
) into a `TimeoutWaitingForResponse` error when calling `try_store_msg(UPDATE_MSG_5)`  after restarting one node to restore the subnet to exactly the minimum consensus threshold.

## Root Cause

With f=16 faulty nodes (out of 49), the subnet runs at the minimum consensus threshold (2f+1=33 nodes). Block production drops to ~0.3-0.4 blocks/s, and DKG transcript loading at interval boundaries (every 100 heights) causes 12-17 second stalls.

The `try_store_msg` call uses `call_and_wait()` with the default 600s `max_polling_time`. The ingress message has a ~4.5 minute expiry (`MAX_INGRESS_TTL - 30s`). When the call coincides with DKG boundary stalls, the ingress message expires before the slow subnet can include and execute it. The agent then polls fruitlessly for the remaining ~5.5 minutes until the 600s timeout, returning `TimeoutWaitingForResponse`.

The flakiness depends on timing: Step 6 (same f=16) usually succeeds because the subnet is running steadily, while Step 8 (after restarting a node at a DKG boundary) fails when transcript loading stalls consensus right when the message is submitted.

## Fix

- Create a dedicated agent with `max_polling_time` of 60s (instead of 600s) for update calls made under degraded conditions (Steps 6, 8, 9).
- Use `retry_with_msg_async!` with a 15-minute timeout and 5s backoff so each retry submits a fresh ingress message with a new expiry.
- This ensures the message eventually lands within its expiry window, even with slow consensus.

## Verification

Ran `bazel test --runs_per_test=3 --jobs=3` — all 3 runs passed (1 run hit an unrelated Farm infrastructure error, which is expected and excluded per the flaky test skill).

---

### DIFF
--- a/rs/tests/driver/src/util.rs
+++ b/rs/tests/driver/src/util.rs
@@ -18,7 +18,7 @@ use futures::{
 use ic_agent::{
     Agent, AgentError, Identity, Signature,
     agent::{
-        CallResponse, EnvelopeContent, RejectCode, RejectResponse,
+        AgentBuilder, CallResponse, EnvelopeContent, RejectCode, RejectResponse,
         http_transport::reqwest_transport::reqwest,
     },
     agent_error::TransportError,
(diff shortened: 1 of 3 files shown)


cgundy (approved):

eichhorl:
> DKG transcript loading at interval boundaries (every 100 heights) causes 12-17 second stalls

Transcripts are loaded asynchronously by a different thread, and one DKG interval ahead of time. Therefore it should not cause subnet stalls. But even if it did, it wouldn't explain why an ingress message with 4.5 minutes expiry expires after a 15 second subnet stall:

> The ingress message has a ~4.5 minute expiry (MAX_INGRESS_TTL - 30s). When the call coincides with DKG boundary stalls, the ingress message expires before the slow subnet can include and execute it.

The kibana logs suggest that the subnet is making progress as expected, so another possible failure reason could be that the ingress message is lost somewhere between the test driver and the subnet. Retrying the ingress message would indeed help in that case, but it could also cover other failure cases.

Maybe we can first try if https://github.com/dfinity/ic/pull/9856 helps?

basvandijk:
> Maybe we can first try if https://github.com/dfinity/ic/pull/9856 helps?

Definitely worth a shot. 

I'll keep this open in case it doesn't help.

PR #9836, opened 2026-04-13
feat(ic-admin): Added canister snapshot commands to ic-admin.

daniel-wong-dfinity-org:
Namely, `propose-to-(take|load)-canister-snapshot`.

Closes [NNS1-4305] and [NNS1-4306].

[NNS1-4305]: https://dfinity.atlassian.net/browse/NNS1-4305
[NNS1-4306]: https://dfinity.atlassian.net/browse/NNS1-4306


rs/nns/integration_tests/src/canister_snapshot.rs
```diff
@@ -182,32 +247,42 @@ fn test_canister_snapshot() {
...
     }
 
-    // Verify initial balances. This just makes sure that we configured this
-    // test correctly; this is not to confirm that the code under test itself is
-    // correct.
-    assert_balance(&state_machine, icp_token_sender, initial_balance);
-    assert_balance(&state_machine, icp_token_receiver, initial_balance);
-    assert_balance(&state_machine, stable_balance_principal, initial_balance);
+    // The initial balances have not changed, but the code that set the initial
+    // balances is a little bitfar away, so this is just to remind the reader
```
jasonz-dfinity, line 269:
```suggestion
    // balances is a little bit far away, so this is just to remind the reader
```

rs/nns/integration_tests/BUILD.bazel
```diff
@@ -164,6 +166,7 @@ DEV_DATA = [
     "//rs/node_rewards/canister:node-rewards-canister",
     "//rs/node_rewards/canister:node-rewards-canister-test",
     "//rs/pocket_ic_server:pocket-ic-server",
+    "//rs/registry/admin:ic-admin",
```
jasonz-dfinity, line 169: I'm not a fan of testing through ic-admin. In my mind it's not part of "NNS". A more practical implication is that any test in rs/nns/integration_tests now depend on building ic-admin. Can we keep the tests here focusing on the interface of the canisters, while having the ic-admin integration tests in its own directory (focusing on a much smaller scope such as verifying the proposal info)?

I'm aware that both points are not that strong so it's not blocking, but just in case if you are also on the fence about this.

jasonz-dfinity (approved):

rs/nns/integration_tests/BUILD.bazel
daniel-wong-dfinity-org: > In my mind it's not part of "NNS".

Yes, but why does that mean that these tests should not also exercise ic-admin? What's the harm?

Here's how I think of it: in actual practice, `ic-admin` is part of our workflow; you don't HAVE to use it (e.g. you can raw dog it with `dfx canister call ...`), but `ic-admin` is generally how we make proposals. Testing this way gives us confidence that the **_entire_** workflow will actually succeed _**IRL**_.

I guess the alternative is to test `ic-admin` by mocking NNS. To me, such mocking constitutes a "mockery", i.e. passing should not really inspire confidence, because we have gutted the "interesting" part of the test (the actual desired effect itself), and replaced it with a dumb mock.

> A more practical implication is that any test in rs/nns/integration_tests now depend on building ic-admin.

Yes, this makes things slower, but you can't make an omelette without cracking some eggs. One thing we can do is make it so that ONLY the tests that actually use ic-admin depend on it. Let me work on that in this PR...

daniel-wong-dfinity-org: Done.

PR #9838, opened 2026-04-13
feat(flexible-outcalls): CON-1708 add is_reject to CanisterHttpResponseMetadata

fspreiss:
## Summary

Adds an `is_reject: bool` field to `CanisterHttpResponseMetadata` so that signed metadata alone can distinguish success from reject responses. This is a prerequisite for fixing the `ResponsesTooLarge` validation logic (CON-1709), where the validator needs to partition shares into OK vs reject sets using only signed metadata — without requiring the full response body.

Follows the same pattern established by PR #9673, which added `content_size` to the metadata.

## Key changes

- **Protobuf**: `bool is_reject = 7` on `CanisterHttpResponseMetadata`, `bool is_reject = 10` on `CanisterHttpResponseWithConsensus`.
- **Rust struct**: New `is_reject: bool` field on `CanisterHttpResponseMetadata`, included in `count_bytes()` and CBOR signing.
- **Pool manager**: Sets `is_reject` from the response content when constructing metadata; validates consistency when receiving shares from peers.
- **Payload validation**: `check_response_consistency` (fully-replicated / non-replicated) and `validate_flexible_response_with_proof` (flexible) both verify `is_reject` matches the response body, returning a new `IsRejectMismatch` error on mismatch.
- **Tests**: Dedicated `is_reject` mismatch tests for `payload.responses`, `payload.flexible_responses`, `payload.flexible_errors` (TooManyRequestErrors), and pool manager share validation (non-replicated + flexible). Updated hash stability golden values. Updated all metadata construction sites including proptests.

## Backward compatibility

No concerns — the canister HTTP pool is purely in-memory (backed by `BTreeMap`, not LMDB) and is emptied on every replica restart. On subnet upgrades all nodes restart simultaneously, so there is no mixed-version period where old shares (missing `is_reject`) coexist with new ones.

The hash stability test golden values changed because the CBOR-serialized (and signed) metadata now includes the new field. This is safe because `CanisterHttpResponseMetadata` is never persisted to disk — it only lives in the in-memory pool and in block payloads, both of which are rebuilt fresh after an upgrade.

### DIFF
--- a/rs/https_outcalls/consensus/src/payload_builder/tests.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder/tests.rs
@@ -600,7 +600,8 @@ fn registry_unavailable_validation() {
 /// the existing helper functions
 #[test]
 fn feature_disabled_validation() {
-    let validation_result = run_non_flexible_validation_test(false, |_, _| {}, &default_validation_context());
+    let validation_result =
+        run_non_flexible_validation_test(false, |_, _| {}, &default_validation_context());
     match validation_result {
         Err(ValidationError::ValidationFailed(
             PayloadValidationFailure::CanisterHttpPayloadValidationFailed(
--- a/rs/https_outcalls/consensus/src/payload_builder/tests.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder/tests.rs
@@ -472,7 +472,7 @@ fn divergence_responses_count_toward_max_responses() {
 /// Test that oversized payloads don't validate
 #[test]
 fn oversized_validation() {
-    let validation_result = run_validatation_test(
+    let validation_result = run_non_flexible_validation_test(
         true,
         |response, _| {
             // Give response oversized content
@@ -493,7 +493,7 @@ fn oversized_validation() {
 /// Test that payloads with wrong registry version don't validate
 #[test]
 fn registry_version_validation() {
-    let validation_result = run_validatation_test(
+    let validation_result = run_non_flexible_validation_test(
         true,
         |_, metadata| {
             // Set metadata to a newer registry version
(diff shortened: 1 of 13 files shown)


rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -554,6 +554,26 @@ fn content_size_validation() {
     );
 }
 
+/// Test that payloads with wrong is_reject flag don't validate
+#[test]
+fn is_reject_validation() {
+    let validation_result = run_validatation_test(
```
eichhorl: At some point we could fix the typo in the function name and maybe also highlight that this is testing non-flexible outcalls

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -368,6 +368,10 @@ impl CanisterHttpPoolManagerImpl {
                         registry_version,
                         content_hash: ic_types::crypto::crypto_hash(&response),
                         content_size: response.content.count_bytes() as u32,
+                        is_reject: matches!(
+                            response.content,
+                            CanisterHttpResponseContent::Reject(_)
+                        ),
```
eichhorl: Alternatively we could implement a function `CanisterHttpResponseContent::is_reject(&self)`

eichhorl (approved):

rs/https_outcalls/consensus/src/payload_builder/tests.rs
fspreiss: Renamed to [run_non_flexible_validation_test](https://github.com/dfinity/ic/pull/9838/commits/86c8eae8efef74c422ba76a01764db06f35de55f) (without the typo).

rs/https_outcalls/consensus/src/pool_manager.rs
fspreiss: Good idea: [Done](https://github.com/dfinity/ic/pull/9838/commits/ab15ab2260683743a4b150c8efff0b237c4d7e24).

mraszyk (approved):

PR #9840, opened 2026-04-13
feat: ExecutionEnvironment::execute_mgmt_operation_on_canister

mraszyk:
This PR introduces a new (helper) function `execute_mgmt_operation_on_canister` on `ExecutionEnvironment` which takes care of creating a clone of a specified canister state and round limits, passing those clones to the (mgmt canister) operation, and
- upon success: committing the (mutated) canister state and round limits to `ReplicatedState`;
- upon error: consuming the recorded amount of cycles for instructions spent before failing.

The motivation for the change is to prevent accidentally committing canister state changes in case of errors (currently the mgmt canister implementation must be careful to first check for all possible error causes and only then perform state changes which is difficult to always get correct).

Performance-wise, installing code to a canister performs [canister state cloning](https://github.com/dfinity/ic/blob/c3e7624f2c7284a736a571f0a88fd2321c029715/rs/execution_environment/src/execution/install_code.rs#L131) already so this PR does not introduce a new bottleneck.


rs/execution_environment/src/execution_environment.rs
```diff
@@ -514,6 +581,85 @@ impl ExecutionEnvironment {
...
+                        *round_limits = saved_round_limits;
+                        consumed_cycles.apply(
+                            canister,
+                            round_limits,
+                            registry_settings.subnet_size,
+                            cost_schedule,
+                            &self.metrics.failed_subnet_message_charge,
+                        );
+                        self.process_canister_manager_result(Err(err), state, msg)
+                    }
```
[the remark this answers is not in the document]
mraszyk: The test `upload_chunk_charges_if_failing` already ensures that the cycles balance is decreased on the failure path and consequently `ExecutionTest` framework ensures the change to `round_limits.instructions` (because it asserts that the number of cycles charged for instructions matches the change to `round_limits.instructions`). The fact that the canister state is unchanged now holds by design (modulo implementation bugs) and thus meaningful tests will only be added in a follow-up PR when `*canister = saved_canister` does actually roll-back state changes.

michael-weigelt:
> Performance-wise, installing code to a canister performs [canister state cloning](https://github.com/dfinity/ic/blob/c3e7624f2c7284a736a571f0a88fd2321c029715/rs/execution_environment/src/execution/install_code.rs#L131) already so this PR does not introduce a new bottleneck.

Not a new _type_ of bottleneck, but more of the same. It might have an impact that all mgmt calls now clone state. It's probably ok, but we should compare the before-after when this is rolled out.

rs/execution_environment/src/execution_environment.rs
```diff
@@ -514,6 +581,85 @@ impl ExecutionEnvironment {
...
+    /// which fails with a corresponding error, but cycles are still
+    /// charged for the work of hashing the uploaded WASM chunk.
+    /// In particular, this means that a dedicated "out of cycles"
+    /// error and no charge should be recorded if the canister is
+    /// completely out of cycles.
+    fn execute_mgmt_operation_on_canister<F, C>(
+        &self,
+        canister_id: CanisterId,
+        op: F,
+        context: C,
```
michael-weigelt: C is unconstrained, so it's hard to guess what it's for. 
All call sites I found had `context: ()`, so is it necessary?

mraszyk: It is not so I removed it. I thought it'd be useful so I added to the function definition (I wrote myself), but Claude decided to not use it when instantiating the function.

rs/execution_environment/src/execution_environment.rs
```diff
@@ -297,6 +297,73 @@ impl RoundLimits {
     }
 }
 
+pub(crate) struct ConsumedCyclesForInstructions<'a> {
```
maksymar, line 303: NIT: would be nice to have a high-level doc comment explaining 'what' and 'why'.

maksymar (approved):

rs/execution_environment/src/execution_environment.rs
michael-weigelt: 💡

michael-weigelt (approved):

PR #9841, opened 2026-04-13
feat: Verify LUKS parameters when opening a disk

frankdavid:
This PR enhances security of LUKS encryption by verifying that LUKS parameters haven't been tempered with - that is the LUKS version, cipher, volume key size and key slot ciphers and key sizes are the same as declared during formatting.

The checks are only enabled on the SEV code path because non-SEV machines may have been set up with different LUKS-parameters in the past (some have been around for a long time). The PR adds code for logging metrics on all types of nodes so we get adequate information about the LUKS parameters used on our nodes. We can later expand the checks to non-SEV machines after we know whether all nodes would pass.


rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -113,15 +142,140 @@ pub fn check_encryption_key(device_path: &Path, encryption_key: &[u8]) -> Result
...
+        .status_handle()
+        .get_cipher()
+        .context("Failed to get cipher")?;
+
+    let cipher_mode = crypt_device
+        .status_handle()
+        .get_cipher_mode()
+        .context("Failed to get cipher mode")?;
+
+    let volume_key_size = crypt_device.status_handle().get_volume_key_size();
```
blind-oracle: nit: get `crypt_device.status_handle()` once and reuse? Also below in `verify_luks_parameters`

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -113,15 +142,140 @@ pub fn check_encryption_key(device_path: &Path, encryption_key: &[u8]) -> Result
...
+        .count();
+
+    let mut keyslot_handle = crypt_device.keyslot_handle();
+    let keyslot_pbkdf = keyslot_handle
+        .get_pbkdf(active_keyslot)
+        .context("Failed to get PBKDF type for active keyslot")?;
+
+    let mut keyslot_cipher = String::new();
+    let mut keyslot_key_size = 0;
+    if let Ok(keyslot_encryption) = keyslot_handle.get_encryption(Some(active_keyslot)) {
```
blind-oracle: Why do we ignore the error case here?

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -113,15 +142,140 @@ pub fn check_encryption_key(device_path: &Path, encryption_key: &[u8]) -> Result
...
+        .context("Failed to get cipher")?;
+
+    let cipher_mode = crypt_device
+        .status_handle()
+        .get_cipher_mode()
+        .context("Failed to get cipher mode")?;
+
+    let volume_key_size = crypt_device.status_handle().get_volume_key_size();
+
+    let num_keyslots = (0..LUKS2_N_KEY_SLOTS)
```
blind-oracle: Should we also check that we have bad keyslots (like in the `Invalid` state etc)?

rs/ic_os/os_tools/guest_disk/src/tests.rs
```diff
@@ -479,3 +488,129 @@ fn test_cannot_open_with_generated_key_if_sev_is_enabled() {
             .expect_err("opening with generated key should fail when SEV is enabled");
     }
 }
+
+#[test]
+fn test_verification_cipher_tampered() {
+    for enable_trusted_execution_environment in [true, false] {
+        let mut fixture = TestFixture::new(enable_trusted_execution_environment);
```
blind-oracle: nit: factor out the loop body into a separate function with a bool arg to make it nicer?

rs/ic_os/os_tools/guest_disk/src/tests.rs
```diff
@@ -479,3 +488,129 @@ fn test_cannot_open_with_generated_key_if_sev_is_enabled() {
             .expect_err("opening with generated key should fail when SEV is enabled");
     }
 }
+
+#[test]
+fn test_verification_cipher_tampered() {
```
blind-oracle: Should we test that it catches that other parameters are wrong (besides cipher)?

That is a bit tedious, but maybe here is a rather critical place that we want to make sure it works.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -113,15 +142,140 @@ pub fn check_encryption_key(device_path: &Path, encryption_key: &[u8]) -> Result
...
+        keyslot_pbkdf_iterations: keyslot_pbkdf.iterations,
+        keyslot_cipher,
+        keyslot_key_size,
+        num_keyslots,
+        passes_verification: verify_luks_parameters(crypt_device).is_ok(),
+    })
+}
+
+/// Verifies that the LUKS parameters match the expected values set in format_crypt_device
+fn verify_luks_parameters(crypt_device: &mut CryptDevice) -> Result<()> {
```
blind-oracle: Maybe we can just get params once in `get_luks_parameters` and attest them here?

blind-oracle (changes requested):

rs/ic_os/os_tools/guest_disk/src/tests.rs
frankdavid: Added a few more tests

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: We could, would you fail the check if there are invalid keyslots?

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: This was used for metrics, so if we cannot get the encryption of the keyslot, we could still log the other metrics. I changed it to error for consistency with the other params.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: I gave it a thought but we'd have had to add more fields and structs because the verification code checks more params. This layer felt unnecessary so I kept the verification code as-is and removed the parameter extraction. Instead, we now extract the params for the metrics directly from the CryptDevice in metrics.rs

rs/ic_os/os_tools/guest_disk/src/crypt.rs
blind-oracle: Probably it's worth it, not sure. Maybe that would indicate that somebody tried messing with a header...

rs/ic_os/os_tools/guest_disk/src/crypt.rs
blind-oracle: Dunno, to me it anyway looks like a more clear and concern-separated approach:
* Load all the needed crypto parameters into a struct in one place (deal with LUKS API only here, like have a `new` method on this struct that gets `&mut CryptDevice` as an arg)
* Validate this struct separately (or even as a method of this struct)
* Create the metrics labels from the same struct (or also a method)

Even if we have a bit more params in the struct than we need for labels - why not, if they're used for checking.
And it's easier to test probably, since we don't need to reformat the disk with different crypto params, we can just mock the struct. WDYT?

blind-oracle (approved):

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: No strong preference, I refactored it into two phases.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: I'm afraid that we may deem a valid disk invalid just because some keyslots are corrupted. An attacker can just mark the keyslot as inactive if they want to.

PR #9845, opened 2026-04-13
feat: CON-1703 Add verification of sender_info field

randombit:


rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1147,6 +1149,131 @@ pub fn requests_with_sender_info(env: TestEnv) {
...
+            }
+
+            let sender_info_error_text = "Invalid sender info:";
+
+            ///////////////////////////////////////////////////////////////////
+            // sender_info with wrong signer canister ID should be rejected
+            for &api_ver in ALL_UPDATE_API_VERSIONS {
+                let mut si = valid_sender_info();
+                si.signer = Blob(CanisterId::ic_00().get().as_slice().to_vec());
+                let response = send_update(api_ver, si).await;
```
aterga, line 1293: Nit - let's also run the unhappy test cases with queries, not just upgrade calls.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1147,6 +1149,131 @@ pub fn requests_with_sender_info(env: TestEnv) {
...
+            ///////////////////////////////////////////////////////////////////
+            // Valid sender_info should be accepted for update calls
+            for &api_ver in ALL_UPDATE_API_VERSIONS {
+                send_update(api_ver, valid_sender_info())
+                    .await
+                    .expect_update_ok(api_ver);
+            }
+
+            ///////////////////////////////////////////////////////////////////
+            // Valid sender_info should be accepted for query calls
```
aterga: Nit. Would it be worth also testing with replicated query calls for completeness?

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -1147,6 +1149,131 @@ pub fn requests_with_sender_info(env: TestEnv) {
     });
 }
 
+/// Tests that requests with valid canister-signed sender_info are accepted,
+/// and that various forms of invalid sender_info are rejected.
+pub fn requests_with_valid_sender_info(env: TestEnv) {
```
aterga, line 1159: Should we extend this test with an additional check that `sender_info`, if successfully verified, could actually be read by the canister? 

This could be done by returning it in the response from the universal canister, I believe.

rs/types/types/src/messages/http.rs
```diff
@@ -333,6 +333,20 @@ pub struct SignedSenderInfo {
     pub sig: Vec<u8>,
 }
 
+/// The content bytes of a sender_info field, used as the signable message
+/// for canister signature verification of sender info.
+///
+/// The signing canister (e.g. Internet Identity) signs the `info` blob
+/// using a canister signature with the domain separator `"ic-sender-info"`.
+#[derive(Clone, Debug)]
+pub struct SenderInfoContent(pub Vec<u8>);
```
eichhorl, line 342: I wonder if this could take a slice instead? Or we could directly change the type of `SignedSenderInfo::info` to `SenderInfoContent`, which would also avoid the extra clone necessary to construct the `SenderInfoContent`. But we can also take a look at this in a follow up.

aterga:
LGTM with minor suggestions about how to improve the system test.

There's no e2e test in this PR (for covering the ability of a canister to actually read the sender_info after successful verification), but maybe that is okay as long as we're clear where that will be added later.

aterga (approved):

rs/validator/src/ingress_validation.rs
```diff
@@ -426,14 +428,109 @@ fn validate_nonce<C: HttpRequestContent>(
...
+            validator
+                .verify_canister_sig(
+                    &canister_sig,
+                    &sender_info_content,
+                    &public_key,
+                    &additional_root_of_trust,
+                )
+                .is_ok()
+        });
+    if !verified_with_additional {
```
eichhorl: I guess if both verifications fail, then we will swallow the first failure. Is that expected?

eichhorl: Should we have tests for this fallback mechanism?

rs/validator/src/ingress_validation.rs
```diff
@@ -426,14 +428,109 @@ fn validate_nonce<C: HttpRequestContent>(
...
+    // it matches the declared signer.
+    let parsed_pk = ic_crypto_iccsa::types::PublicKey::try_from(&pk_bytes)
+        .map_err(|e| InvalidSenderInfo(format!("invalid canister sig public key: {e:?}")))?;
+    let pubkey_canister_id = parsed_pk.signing_canister_id();
+    if pubkey_canister_id != sender_info.signer {
+        return Err(InvalidSenderInfo(format!(
+            "signer {} does not match canister ID {} in sender_pubkey",
+            sender_info.signer, pubkey_canister_id
+        )));
     }
```
eichhorl, line 521: Do we have a test for rejecting malformed public keys?

eichhorl (approved):

rs/validator/src/ingress_validation.rs
```diff
@@ -201,7 +203,7 @@ where
     R::Error: std::error::Error,
 {
     validate_nonce(request)?;
-    validate_sender_info(request)?;
+    validate_sender_info(request, ingress_signature_verifier, root_of_trust_provider)?;
     validate_user_id_and_signature(
```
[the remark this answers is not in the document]
randombit: Seems marginal but sure

rs/validator/src/ingress_validation.rs
```diff
@@ -426,14 +428,109 @@ fn validate_nonce<C: HttpRequestContent>(
...
+    let pk_bytes = public_key_bytes_from_der(sender_pubkey_bytes).map_err(|e| {
+        InvalidSenderInfo(format!(
+            "sender_pubkey is not a valid canister signature public key: {e}"
+        ))
+    })?;
+
+    // Extract the canister ID from the parsed public key and verify
+    // it matches the declared signer.
+    let parsed_pk = ic_crypto_iccsa::types::PublicKey::try_from(&pk_bytes)
+        .map_err(|e| InvalidSenderInfo(format!("invalid canister sig public key: {e:?}")))?;
```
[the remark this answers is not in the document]
randombit: I love it when my error messages have less information

rs/validator/src/ingress_validation.rs
```diff
@@ -426,14 +428,109 @@ fn validate_nonce<C: HttpRequestContent>(
...
+
+    // Construct the signable content (domain = "ic-sender-info")
+    let sender_info_content = SenderInfoContent(sender_info.info.clone());
+    let canister_sig = CanisterSigOf::from(CanisterSig(sender_info.sig.clone()));
+
+    // Try verification with the additional root of trust (mainnet) first,
+    // then fall back to the local root of trust.
+    let verified_with_additional = root_of_trust_provider
+        .additional_root_of_trust()
+        .is_some_and(|additional_root_of_trust| {
```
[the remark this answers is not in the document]
randombit: Macro saves the day

rs/tests/crypto/ingress_verification_test.rs
randombit: Done

rs/tests/crypto/ingress_verification_test.rs
randombit: Done

rs/validator/src/ingress_validation.rs
randombit: Probably so though to me it seems somewhat orthogonal to this change, and also if it breaks the only issue is with PocketIC rather than anything production. Opened CON-1711 regarding this.

rs/validator/src/ingress_validation.rs
randombit: No, added now

rs/tests/crypto/ingress_verification_test.rs
randombit: added

rs/types/types/src/messages/http.rs
randombit: Probably, bumped to CON-1712 for a fast merge here

PR #9847, opened 2026-04-13
feat(governance): derive Mission 70 maturity modulation from XRC price history

jasonz-dfinity:
## Why

Previously, Governance relied on CMC (Cycles Minting Canister) as a middleman for ICP price data used in maturity modulation. This PR cuts out the middleman by having Governance fetch ICP/XDR rates directly from the Exchange Rate Canister (XRC) and derive Mission 70 maturity modulation locally.

https://dfinity.atlassian.net/browse/NNS1-4319

## What

- `IcpPriceHistory`: bounded sorted vector of up to 365 daily ICP/XDR rates in Governance state (new proto field 33)
- `UpdateIcpXdrRateRelatedData`: timer task that fetches rates directly from XRC, backfills the last 365 days on first upgrade (~30 min at 5-second intervals), then switches to daily fetches aligned to UTC midnight; recomputes Mission 70 maturity modulation after each update
- The existing CMC-polled maturity modulation remains active for spawning/disbursement (switchover is in a later PR)

### PR Chain
- ⬅️ Previous: #9846
- ➡️ Next: #9848, #9879

## Testing

Unit tests for the price history buffer, maturity modulation computation, and UTC midnight timer logic.


rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,475 @@
...
+        }
+
+        let rate = SampledPrice::from(&exchange_rate);
+        if rate.xdr_permyriad_per_icp == 0 {
+            println!(
+                "{}UpdateIcpXdrRateRelatedData: received zero XDR/ICP rate; ignoring.",
+                LOG_PREFIX
+            );
+            return None;
+        }
```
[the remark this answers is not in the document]
jasonz-dfinity: Fixed: added a timestamp equality check -- if XRC returns a different timestamp than requested, the rate is rejected with a log message.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,475 @@
...
+        println!(
+            "{}compute_average_icp_xdr_rate: only {} of {} days available in window (current_day={})",
+            LOG_PREFIX, count, window_days, current_day
+        );
+    }
+    if count == 0 {
+        return None;
+    }
+    let sum: u64 = filtered.into_iter().sum();
+    Some(sum / count)
```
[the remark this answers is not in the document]
jasonz-dfinity: Fixed: sum is now computed in u128.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+    ) as i64
+}
+
+pub(super) struct UpdateIcpXdrRateRelatedData {
+    governance: &'static LocalKey<RefCell<Governance>>,
+    xrc_client: Arc<dyn ExchangeRateCanisterClient>,
+}
+
+impl UpdateIcpXdrRateRelatedData {
+    pub fn new_with_client(
```
daniel-wong-dfinity-org, line 179: Why not just call this `new`?

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+        if let Err(err) = validate_exchange_rate(&exchange_rate) {
+            println!(
+                "{}UpdateIcpXdrRateRelatedData: XRC rate failed validation: {}",
+                LOG_PREFIX, err
+            );
+            return None;
+        }
+
+        // Verify that XRC returned a rate for the day we requested. If not, the rate
+        // won't fill the expected slot and backfill would loop on the same day.
```
daniel-wong-dfinity-org, line 248: Can you ask the FI team in what cases this might happen to confirm that it generally wouldn't?

daniel-wong-dfinity-org (approved):

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+/// Callers must ensure the 365-day window is fully populated before calling this function.
+fn update_maturity_modulation(
+    icp_price_history: &IcpPriceHistory,
+    maturity_modulation: &mut MaturityModulation,
+    current_day: u64,
+) {
+    let previous_permyriad = maturity_modulation.current_value_permyriad.unwrap_or(0) as i64;
+    let previous_day = maturity_modulation.updated_at_days_since_epoch.unwrap_or(0);
+
+    let new_permyriad = compute_maturity_modulation_permyriad(
```
tmu0, line 348: I would skip this if `previous_day == current_day`

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+        previous_day,
+    );
+
+    maturity_modulation.current_value_permyriad = Some(new_permyriad as i32);
+    maturity_modulation.updated_at_days_since_epoch = Some(current_day);
+}
+
+#[async_trait]
+impl RecurringAsyncTask for UpdateIcpXdrRateRelatedData {
+    async fn execute(self) -> (Duration, Self) {
```
tmu0, line 361: Task execution looks brittle. If there is a panic in this method, execution never happens again.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+            update_rates_buffer(history, rate);
+        });
+
+        // If the history is still incomplete, keep backfilling at the short interval.
+        if self.get_day_to_fetch(current_day).is_some() {
+            return (Duration::from_secs(BACKFILL_INTERVAL_SECONDS), self);
+        }
+
+        // History is complete for current_day: compute maturity modulation.
+        self.governance.with_borrow_mut(|gov| {
```
tmu0: dup code, just remove this and return with a short duration

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+                     (day={}, buffer_size={})",
+                    LOG_PREFIX,
+                    maturity_modulation.current_value_permyriad.unwrap_or(0),
+                    current_day,
+                    icp_price_history.icp_xdr_rates.len(),
+                );
+            });
+
+            let now = self.governance.with_borrow(|gov| gov.env.now());
+            return (duration_until_next_midnight_utc(now), self);
```
tmu0, line 414: I don't know if timer scheduling is exact enough that it always hits midnight or later or if in practice this might be already executed short before midnight and in the worst case one day might get skipped. Definitely worth testing and eventually adding a small delta that ensures that we're executing after midnight.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
```diff
@@ -0,0 +1,466 @@
...
+    let speed_limited = target_modulation.clamp(
+        previous_permyriad.saturating_sub(max_change) as i128,
+        previous_permyriad.saturating_add(max_change) as i128,
+    );
+
+    // Global bounds have final say. The result is within [MIN, MAX] which fit in i64, so the
+    // cast is safe.
+    speed_limited.clamp(
+        MIN_MATURITY_MODULATION_PERMYRIAD as i128,
+        MAX_MATURITY_MODULATION_PERMYRIAD as i128,
```
tmu0: These are still the old values (-500, 500). What we want is (-1000, 200)

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Right, but in the case of a panic, there isn't really a good way of recovering. OTOH, If the instruction limit is hit unintentionally, NOT rescheduling is strictly better. Do you have a suggestion?

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Do you mean if it can be run BEFORE the intended time? I doubt that is possible, but let me confirm.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Done.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: I read through the XRC source. The result rate's `timestamp` field is set from the `requested_timestamp` (which is `(request.timestamp / 60) * 60` — i.e. minute-aligned, so for our midnight-aligned requests it equals what we sent). That timestamp propagates through the cache (keyed on `(symbol, timestamp)`) and through `Mul`/`Div` (which preserve `self.timestamp`). The forex shift logic only affects the separate `forex_timestamp` metadata field, not the main `timestamp`. So in practice this guard should never fire — it's defensive against XRC behavior changes (or someone forgetting to pass a timestamp on a future code path).

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Done — added an early return in `update_maturity_modulation` when `updated_at_days_since_epoch == Some(current_day)`.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Done — removed the duplicate block. After every successful fetch we now return `BACKFILL_INTERVAL_SECONDS`. The next iteration either fetches the next missing day, or — if history is complete — updates maturity modulation via the early-return branch and returns next-midnight.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Done — added new local constants `MATURITY_MODULATION_MIN_PERMYRIAD_MISSION_70 = -1_000` and `MATURITY_MODULATION_MAX_PERMYRIAD_MISSION_70 = 200`, and switched the clamp to use them. Left the shared `MIN/MAX_MATURITY_MODULATION_PERMYRIAD` in `ic_nervous_system_governance::maturity_modulation` untouched (-500/+500), since CMC and SNS still depend on those for the pre-Mission-70 logic. The reason no test failed before is that the test was importing the shared `MAX_MATURITY_MODULATION_PERMYRIAD` and asserting against it, so it tracked whatever value the production code clamped to. The test now asserts against the new local constant.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Done — added a guard at the top of `execute()` that detects this case: if `maturity_modulation.updated_at_days_since_epoch == Some(current_day)` (i.e. we already updated for the current day but the timer fired again), we skip all work and return `duration_until_next_midnight_utc(now)`. Better than a static delta — handles the case directly instead of always padding the schedule.

rs/nns/governance/src/timer_tasks/update_icp_xdr_rate_related_data.rs
jasonz-dfinity: Discussed offline and agreed to add alerts instead.

bjoernek (approved):

PR #9849, opened 2026-04-14
feat: query method to list all canisters on a subnet

mraszyk:
This PR implements a query method that subnets admins can use to list all canisters on their subnet. The method returns the canister IDs of existing canisters (omitting deleted canisters) encoded as a list of canister IDs pairs with all canister IDs inside those pairs existing on the subnet. In the future, pagination could be added in a backward-compatible way by extending the input and output types in Candid.

### DIFF
--- a/rs/execution_environment/src/query_handler.rs
+++ b/rs/execution_environment/src/query_handler.rs
@@ -57,8 +57,8 @@ use tower::{Service, util::BoxCloneService};
 pub(crate) use self::query_scheduler::QueryScheduler;
 use crate::execution::common::validate_subnet_admin;
 use ic_management_canister_types_private::{
-    CanisterIdRange, CanisterIdRecord, FetchCanisterLogsRequest, ListCanistersResponse, Payload,
-    QueryMethod,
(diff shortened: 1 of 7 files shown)


rs/execution_environment/src/query_handler.rs
```diff
@@ -173,6 +176,47 @@ impl InternalHttpQueryHandler {
...
+        state: &ReplicatedState,
+        caller: &ic_types::PrincipalId,
+        payload: &[u8],
+    ) -> Result<WasmResult, UserError> {
+        match EmptyBlob::decode(payload) {
+            Err(err) => Err(err),
+            Ok(EmptyBlob) => {
+                match state.get_own_subnet_admins() {
+                    Some(ref admins) => {
+                        validate_subnet_admin(admins, caller).map_err(UserError::from)?
```
michael-weigelt, line 190: Why is this gated? Couldn't this be useful on public subnets as well?

michael-weigelt (approved):

rs/execution_environment/src/query_handler.rs
mraszyk: This is gated for now because the response size could be large on public subnets and thus making the endpoint public is blocked on supporting pagination.

PR #9854, opened 2026-04-14
fix(governance): reset INFLIGHT flag on get_node_providers_rewards failure

pietrodimarco-dfinity:
# Motivation

The `INFLIGHT` guard in `get_node_providers_rewards_cached` is never reset when the inter-canister call to the Node Rewards Canister fails. The `?` operator on the `await` line returns early, skipping `INFLIGHT.set(false)`. This permanently blocks all subsequent calls to the cached endpoint with:

```
GovernanceError { error_type: External, error_message: "get_node_provider_rewards called while there is an inflight call to NRC. Please try again in 10 seconds." }
```

This is currently happening on mainnet because the NRC is failing due to a newly created cloud engine subnet (tracked in #9852).

# Changes

- Split the `await?` into separate `await` + `?` steps so that `INFLIGHT.set(false)` always executes regardless of call success or failure.

### DIFF
--- a/rs/nns/governance/src/governance.rs
+++ b/rs/nns/governance/src/governance.rs
@@ -7606,9 +7606,10 @@ impl Governance {
         }
 
         INFLIGHT.set(true);
-        let rewards = self.get_node_providers_rewards().await?;
+        let rewards = self.get_node_providers_rewards().await;
         INFLIGHT.set(false);
 
+        let rewards = rewards?;
         CACHE.set(Some((now, rewards.clone())));
 
         Ok(rewards)
--- a/rs/nns/governance/unreleased_changelog.md
+++ b/rs/nns/governance/unreleased_changelog.md
@@ -17,4 +17,7 @@ on the process that this file is part of, see
 
 ## Fixed
 
+* When result in `get_node_providers_rewards_cached` is `Err`,
+  release INFLIGHT (i.e. set it to `false`).
+
 ## Security


0xFFrancesco (approved):

0xFFrancesco (approved):

rs/nns/governance/src/governance.rs
```diff
@@ -7641,9 +7641,10 @@ impl Governance {
         }
 
         INFLIGHT.set(true);
-        let rewards = self.get_node_providers_rewards().await?;
+        let rewards = self.get_node_providers_rewards().await;
         INFLIGHT.set(false);
```
daniel-wong-dfinity-org, line 7610: Not asking for change.

This implicitly assumes that there are no early returns. As this shows, that is extremely fragile. The way to avoid this is to implement Drop (for example, [Temporary]).

[Temporary]: https://sourcegraph.com/r/github.com/dfinity/ic@0800086808b86b76493c86371b27276c843dbef1/-/blob/rs/nervous_system/temporary/src/lib.rs?L88-94

# Tangent

In the C++ world, this is called "resource manager", or RAII (resource acquisition is initialization, or some shit like this that nobody can remember). Without this, you have to make sure that you execute exactly one `delete` for every `new`, and humans are horrible at getting this right. That's why resource managers are the ONLY way to RELIABLY do this kind of lifecycle management. See Chapter 3 in [Effective C++].

[Effective C++]: https://ptgmedia.pearsoncmg.com/images/9780321334879/samplepages/0321334876.pdf

daniel-wong-dfinity-org (approved):

PR #9859, opened 2026-04-14
feat: Store prepayment for call transmission

dsarlis:
Store the prepayment for call transmission so that it can be used to properly calculate the consumed amount for an inter-canister call when updating the metrics during the refund phase.

### Problem

In the current state of the code, only the prepayment for response transmission is stored so that the refund can be calculated when the actual response is received. This is enough for now when wanting to credit the refund back to the canister's balance or updating the current metrics which act like gauges. However, if we want to have new metrics that will act as counters (which will happen in a follow-up) we cannot rely only on that amount for the consumed part as we would miss the parts that are not refunded -- specifically the base fee and request transmission parts.

### Solution

The proposed solution here is to additionally store the total call transmission fee so that it can be passed into the function that refunds cycles to the canister -- the refund part to the balance remains the same but based on prepayment and refund one can calculate the amount to update the metrics with only at refund step.

The changes include the following parts:
1. Storing the new field in the callback in a checkpoint and updating any code needed because of that.
2. Some logic to handle the case that this field could be empty for callbacks that are outstanding. Since no updates are happening to metrics yet, we can hope that by the time after this code is deployed and the update to metrics will be, there won't be empty values there.
3. A small refactoring in `CyclesAccountManager` that felt reasonable given that we have the total transmission fee computed and we don't need to recompute stuff.


dsarlis:
> Run on [dd37c1d](https://github.com/dfinity/ic/commit/dd37c1d1e3f3522ba166534a6bbf0caf63a27446) URL: https://github.com/dfinity/ic/actions/runs/24399827516

I see some test failures that I missed, I'll push a fix in a bit.

rs/embedders/tests/sandbox_safe_system_state.rs
```diff
@@ -86,8 +96,8 @@ fn push_output_request_fails_not_enough_cycles_for_request() {
             NumBytes::from(0),
             MessageMemoryUsage::ZERO,
             request.clone(),
-            CompoundCycles::new(Cycles::zero(), CanisterCyclesCostSchedule::Normal),
-            CompoundCycles::new(Cycles::zero(), CanisterCyclesCostSchedule::Normal),
+            prepayment_for_response_execution,
+            prepayment_call_response_transmission,
```
dsarlis: The test was passing 0 values here relying on some internal computation that would be non-zero to make things work out. I think with the changes that I've made in the production code the test itself becomes more self-explanatory.

rs/embedders/tests/sandbox_safe_system_state.rs
```diff
@@ -103,34 +113,19 @@ fn push_output_request_fails_not_enough_cycles_for_response() {
...
-        + request_payload_cost
-        + prepayment_for_response_execution.real()
-        + prepayment_for_response_transmission.real();
+    let prepayment_for_call_transmission = cycles_account_manager.xnet_total_transmission_fee(
+        request.payload_size_bytes(),
+        SMALL_APP_SUBNET_MAX_SIZE,
+        CanisterCyclesCostSchedule::Normal,
+    );
+    let total_cost =
+        prepayment_for_response_execution.real() + prepayment_for_call_transmission.real();
```
dsarlis, line 128: Similar to the test above.

rs/embedders/tests/sandbox_safe_system_state.rs
```diff
@@ -270,10 +255,13 @@ fn correct_charging_source_canister_for_a_request() {
...
-        + request_payload_cost
-        + prepayment_for_response_execution.real()
-        + prepayment_for_response_transmission.real();
+    let prepayment_for_call_transmission = cycles_account_manager.xnet_total_transmission_fee(
+        request.payload_size_bytes(),
+        SMALL_APP_SUBNET_MAX_SIZE,
+        cost_schedule,
+    );
+    let total_cost =
+        prepayment_for_response_execution.real() + prepayment_for_call_transmission.real();
```
dsarlis, line 264: Similar to the other updates, based on the production code, the tests need to be adapted to pass the correct amounts to the expected APIs.

dsarlis:
> > Run on [dd37c1d](https://github.com/dfinity/ic/commit/dd37c1d1e3f3522ba166534a6bbf0caf63a27446) URL: https://github.com/dfinity/ic/actions/runs/24399827516
> 
> I see some test failures that I missed, I'll push a fix in a bit.

Pushed a fix.

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -1256,10 +1269,8 @@ impl SandboxSafeSystemState {
         self.system_state_modifications.requests.push(msg);
         *used_slots += 1;
         let consumed_cycles = ConsumedCyclesDuringExecution {
-            instructions: Some(instructions_consumed_cycles),
-            request_and_response_transmission: Some(
-                request_and_response_trasmission_consumed_cycles,
-            ),
+            instructions: Some(prepayment_for_response_execution),
+            request_and_response_transmission: Some(prepayment_for_call_transmission),
```
mraszyk, line 1273: should we also call the field `call_transmission` or `prepayment_for_call_transmission`

rs/execution_environment/src/execution/response.rs
```diff
@@ -210,8 +213,21 @@ impl ResponseHelper {
...
 
+        // `self.prepayment_for_call_transmission` might be zero for any responses
+        // to requests that were produced before this field was stored in callbacks.
+        // In that scenario, use `self.prepayment_for_response_transmission` for some
+        // partial update to consumed metrics (the ones that behave like counters)
+        // during refund phase. This does not affect the refund to balance or the
+        // metrics that behave like gauges as these are updated based on the refund amount.
+        // This code can be dropped when all callbacks have the `prepayment_for_call_transmission`
+        // field set.
+        let prepayment_for_call_transmission = if self.prepayment_for_call_transmission.is_zero() {
```
mraszyk, line 224: I first thought that the if-condition is trivially true on "free" cost schedule and then I noticed that `is_zero` checks both amounts (real and nominal). How about naming it `real_and_nominal_are_zero`?

rs/execution_environment/src/scheduler/test_utilities.rs
```diff
@@ -1484,6 +1484,12 @@ impl TestWasmExecutorCore {
         let prepayment_for_response_transmission = self
             .cycles_account_manager
             .prepayment_for_response_transmission(self.subnet_size, system_state.cost_schedule());
+        let prepayment_for_call_transmission =
+            self.cycles_account_manager.xnet_total_transmission_fee(
+                NumBytes::from(0),
```
mraszyk: why zero?

rs/embedders/tests/sandbox_safe_system_state.rs
mraszyk: The test worked because of the non-zero request transmission cost computed inside `self.cycles_account_manager.withdraw_request_cycles` which is no longer the case so the correct amount must be passed here.

rs/embedders/tests/sandbox_safe_system_state.rs
mraszyk: Here we're passing the same amount, just computed using a different code path in `CyclesAccountManager`.

rs/embedders/tests/sandbox_safe_system_state.rs
mraszyk: Here we're passing the same amount, just computed using a different code path in CyclesAccountManager.

rs/embedders/tests/sandbox_safe_system_state.rs
```diff
@@ -584,6 +572,11 @@ fn test_inter_canister_call(
             SMALL_APP_SUBNET_MAX_SIZE,
             CanisterCyclesCostSchedule::Normal,
         );
+    let prepayment_for_call_transmission = cycles_account_manager.xnet_total_transmission_fee(
+        NumBytes::from((method_name.len() + arg.len()) as u64),
```
mraszyk: Maybe we could move `let request = RequestBuilder` above this and use `request.payload_size_bytes` instead of a manual computation.

rs/types/types/src/methods.rs
```diff
@@ -282,8 +282,12 @@ pub struct Callback {
     pub prepayment_for_response_execution: CompoundCycles<Instructions>,
     /// Cycles prepaid by the caller for response transimission.
     ///
-    /// `Cycles::zero()` if the `Callback` was created before February 2022.
+    /// `CompoundCycles::zero()` if the `Callback` was created before February 2022.
     pub prepayment_for_response_transmission: CompoundCycles<RequestAndResponseTransmission>,
+    /// Cycles prepaid by the caller for call transmission (includes both request and response transmission).
+    ///
+    /// `CompoundCycles::zero()` if the `Callback` was created before April 2026.
```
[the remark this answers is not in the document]
mraszyk: I'd ignore this.

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
dsarlis: I'd call it `prepayment_for_call_transmission` to keep things consistent and I'll change the other place that Copilot found is inconsistent. Basically, what `response` should be renamed to `call` but the remaining should be the same.

rs/execution_environment/src/scheduler/test_utilities.rs
dsarlis: Scheduler uses mock requests (`ΤestCall`) which do not really contain any payload. I can leave a comment since it's not trivial if you asked.

rs/embedders/tests/sandbox_safe_system_state.rs
dsarlis: This was my initial idea. Not possible because the value is needed for creating the callback, which in turn is needed in `RequestBuilder` (so circular dependency).

rs/execution_environment/src/execution/response.rs
dsarlis: I could but it feels like this should be an implementation detail of `CompoundCycles::is_zero` rather than making a long name to explain it. The doc comment says what it's supposed to check already and the type holds two amounts, so it's not so unexpected imo.

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
dsarlis: Actually, no, the field is supposed to match the corresponding `CyclesUseCase`. I'd leave it like this if you don't mind. Otherwise, we might want to change `instructions` too but then it's a renaming thing that should happen separately (these appear in a few spots so it grows the scope here).

rs/types/types/src/methods.rs
```diff
@@ -397,13 +404,19 @@ impl TryFrom<pb::Callback> for Callback {
                     cost_schedule,
                 ),
             };
+        let prepayment_for_call_transmission = try_from_option_field(
+            value.prepayment_for_call_transmission,
+            "Callback::prepayment_for_call_transmission",
+        )
+        .unwrap_or(CompoundCycles::new(Cycles::zero(), cost_schedule));
```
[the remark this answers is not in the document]
dsarlis: This is a good point and actually the code should be changed to fail if it finds a `None` value after 2 upgrades and assuming no "old" callback remains outstanding.

rs/embedders/tests/sandbox_safe_system_state.rs
mraszyk: Maybe we could still store the number of bytes into a variable and add an assert that the value matches `request.payload_size_bytes()` (to avoid spurious errors if the payload size bytes computation ever changes)?

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
mraszyk: Makes sense, let's leave it as is.

mraszyk (dismissed):

rs/embedders/tests/sandbox_safe_system_state.rs
dsarlis: Sure I can do that.

mraszyk (approved):

schneiderstefan (approved):

PR #9860, opened 2026-04-14
feat(flexible-outcalls): CON-1709 harden ResponsesTooLarge validation with all-seen-shares proof

fspreiss:
Strengthens the `ResponsesTooLarge` error for flexible HTTP outcalls to prevent a malicious block proposer from fabricating the error by selectively omitting small OK shares.

- Renamed `metadata_shares` → `all_seen_shares` in `FlexibleCanisterHttpError::ResponsesTooLarge` and the corresponding protobuf message, so the error now carries *all* seen shares (both OK and reject), not just a subset.
- Updated the builder (`find_flexible_result`) to emit all OK and reject shares into `all_seen_shares`.
- Rewrote the validator logic to use `is_reject` (from CON-1708) to partition shares, compute `num_unseen` and `min_known_ok_needed`, and verify that the smallest set of OK responses genuinely exceeds `MAX_CANISTER_HTTP_PAYLOAD_SIZE`. An early check rejects payloads with too few OK shares before computing sizes.
- Replaced `FlexibleInsufficientMetadataShareCount` with the more descriptive `FlexibleResponsesTooLargeInsufficientEvidence` error variant.
- Added comprehensive validation tests: mixed OK/reject shares, unseen committee members, attack-vector scenarios (committee members omitted), and insufficient OK share counts.

### Attack vector closed

Previously, a malicious proposer could craft a `ResponsesTooLarge` error by including only a few hand-picked large OK shares while omitting committee members who submitted small responses. The validator would check only the provided shares and accept the error as valid.

Now, the validator computes `num_unseen = committee_size - all_seen_shares.len()` and `min_known_ok_needed = min_responses - num_unseen`. Since unseen members *could* have submitted zero-size OK responses, the validator conservatively assumes they did. A proposer who omits members increases `num_unseen`, which lowers `min_known_ok_needed`, making it *harder* to prove responses are too large—the opposite of what an attacker wants.

## Test plan

- [x] `flexible_build_responses_too_large` — builder emits all 4 OK shares
- [x] `flexible_build_responses_too_large_with_rejects_reducing_unseen` — builder emits 3 OK + 2 reject shares
- [x] `flexible_error_responses_too_large_valid` — all committee members with huge OK → accepted
- [x] `flexible_error_responses_too_large_valid_with_unseen_members` — legitimate error with 1 unseen member → accepted
- [x] `flexible_error_responses_too_large_valid_with_mixed_ok_and_reject` — 2 huge OK + 2 reject, all seen → accepted
- [x] `flexible_error_responses_too_large_invalid_when_small` — small OK shares + high unseen → rejected (min_known_ok_needed=0)
- [x] `flexible_error_responses_too_large_invalid_when_committee_members_omitted` — attacker omits members, num_unseen defeats the claim → rejected
- [x] `flexible_error_responses_too_large_too_few_ok_shares` — not enough OK shares for min_known_ok_needed → `InsufficientEvidence`
- [x] Existing tests for duplicate signers, callback mismatch, non-committee signers, registry version mismatch, and invalid signatures continue to pass


michael-weigelt (approved):

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -450,12 +450,14 @@ pub(crate) fn find_flexible_result(
             .sum();
 
         if smallest_content_sum > MAX_CANISTER_HTTP_PAYLOAD_SIZE {
+            let all_seen_shares: Vec<_> = all_ok_shares_sorted_asc
+                .iter()
+                .map(|(share, _)| (*share).clone())
+                .chain(reject_responses.iter().map(|(_, share)| (*share).clone()))
+                .collect();
             let error = FlexibleCanisterHttpError::ResponsesTooLarge {
```
eichhorl, line 458: Technically I guess it could happen that, given a set of shares, you could construct either a `ResponsesTooLarge` or a `TooManyRequestErrors` error, and either would be accepted by the validator. Probably `TooManyRequestErrors` would be more interesting to the user, and it seems that is already the one we prioritize in the honest payload builder.

At the same time it doesn't seem worth the extra complexity to reject `ResponsesTooLarge` errors that should be `TooManyRequestErrors` instead, so it's probably fine to keep the behavior as is.

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -778,30 +778,36 @@ impl CanisterHttpPayloadBuilderImpl {
...
                     }
 
-                    if metadata_shares.len() < *min_responses as usize {
+                    let num_unseen = flex_committee.len().saturating_sub(all_seen_shares.len());
+                    let min_known_ok_needed = min_responses.saturating_sub(num_unseen);
+
+                    let ok_count = all_seen_shares
+                        .iter()
+                        .filter(|s| !s.content.is_reject)
+                        .count();
```
eichhorl: We could collect the `ok_entry_sizes` already here and then set `let ok_count = ok_entry_sizes.len()`, to avoid iterating a second time.

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -778,30 +778,36 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                                &context.request.sender,
                                 share.content.content_size as usize,
                                 share,
                             )
                         })
                         .collect();
-                    entry_sizes.sort_unstable();
-                    let smallest_sum: usize = entry_sizes[..*min_responses as usize].iter().sum();
+                    ok_entry_sizes.sort_unstable();
+                    let smallest_sum: usize = ok_entry_sizes[..min_known_ok_needed].iter().sum();
```
eichhorl: Alternatively
```suggestion
                    let smallest_sum: usize = ok_entry_sizes.iter().take(min_known_ok_needed).sum();
```
To avoid the direct array access

eichhorl (approved):

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Is this condition still correct? I guess there could be a scenario where we have less than `min_responses` ok shares, but are already able to proof that we won't be able to fit them into a payload. For example: (min=4, max=6, committee=6), with 2 rejects and 3 large OK responses with a sum > the max payload size.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
fspreiss: Yes, good point, that could happen, and indeed it seems reasonable to not add the extra complexity of rejecting a `ResponsesTooLarge` that should technically be a `TooManyRequestErrors`.

rs/https_outcalls/consensus/src/payload_builder.rs
fspreiss: [Done](https://github.com/dfinity/ic/pull/9860/commits/17b91e1bf37c307ed98251e0c51edccf8df04132).

In fact, the original code was like that, but then I changed it to fail early when there aren't enough OK shares to avoid computing `count_bytes_from_parts`. But I was a bit on the fence, and given that you also point this out, plus the fact that `count_bytes_from_parts` is just arithmetic (no I/O, etc) and the savings are negligible, the simpler and more readable code is worth more.

rs/https_outcalls/consensus/src/payload_builder.rs
fspreiss: [Done](https://github.com/dfinity/ic/pull/9860/commits/c06421ab2170896b5b3d9a901eefa607e147f819).

The same pattern was used in the payload _builder_ side, so I also fixed it there.

Plus, as I was touching that code anyway, I [renamed](https://github.com/dfinity/ic/pull/9860/commits/3210915c123597c3be147ace8639c46741f3fd3b) `min_minus_unseen` to `min_known_ok_needed`: both variables represent the same concept and computed in the same way: the renaming makes the correspondence between the builder and validator sides immediately obvious.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
fspreiss: Good catch! The condition is indeed overly conservative. One indicator was also that we only `.take(min_known_ok_needed)`. The correct guard is `if all_ok_shares_sorted_asc.len() >= min_known_ok_needed {`.

I [fixed](https://github.com/dfinity/ic/pull/9860/commits/49a552084dcbf3489975cd7c81f2e7fe5001567d) the code and also added a respective test.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -2961,6 +2961,66 @@ fn flexible_build_responses_too_large_with_rejects_reducing_unseen() {
...
+
+    // min_responses=4, committee=6.
+    // 3 large OK shares (sum > MAX) from nodes 0..3.
+    // 2 rejects from nodes 3..5 → only 1 unseen member remains.
+    // min_known_ok_needed = 4 - 1 = 3, and we have exactly 3 OK shares whose sum exceeds MAX.
+    // Even though num_ok (3) < min_responses (4), we can already prove impossibility.
+    let body_size = (MAX_CANISTER_HTTP_PAYLOAD_SIZE / 3) + 100_000;
+    setup_test_with_flexible_context(num_nodes, callback_id, committee, 4, 6, |pb, pool| {
+        {
+            use ic_types::canister_http::CanisterHttpReject;
```
eichhorl: This seems to be done multiple times in this file, maybe we could just import it once at the top of the file

eichhorl (approved):

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Good idea, thanks!

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: LGTM 👍

rs/https_outcalls/consensus/src/payload_builder/tests.rs
fspreiss: 👍  [Done](https://github.com/dfinity/ic/pull/9860/commits/f962ab5d463e1366f243ebf709c9ad12f4e431a2).

PR #9861, opened 2026-04-14
fix(subnet-splitting): relax one of checks in the `do_split_subnet` method

kpop-dfinity:
When checking if some registry records changed across an await point, we were assuming that the records must exist, which is not necessarily the case. For example, the `subnet_splitting_v2_test` system test fails because `canister_migrations_record` doesn't exist in the registry when `do_split_subnet` is executed for the first time

Note: this endpoint is not yet used

### DIFF
--- a/rs/registry/canister/src/mutations/do_split_subnet.rs
+++ b/rs/registry/canister/src/mutations/do_split_subnet.rs
@@ -388,12 +388,9 @@ impl Registry {
         &self,
         record_key: &str,
         version: Version,
-    ) -> Version {
+    ) -> Option<Version> {
         self.get(record_key.as_bytes(), version)
             .map(|record| record.version)
-            .unwrap_or_else(|| {
-                panic!("Record for {record_key} not found in registry");
-            })
     }
 }
 
--- a/rs/registry/canister/unreleased_changelog.md
+++ b/rs/registry/canister/unreleased_changelog.md
@@ -17,4 +17,7 @@ on the process that this file is part of, see
 
 ## Fixed
 
+* `do_split_subnet` - don't assume that all the registry entries exist when checking whether the
+  entries changed across await point
+
 ## Security


rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -388,12 +388,9 @@ impl Registry {
         &self,
         record_key: &str,
         version: Version,
-    ) -> Version {
+    ) -> Option<Version> {
```
daniel-wong-dfinity-org-twin, line 391: I am surprised that this breaks nothing. I guess this means that nobody calls this yet?

daniel-wong-dfinity-org-twin (approved):

daniel-wong-dfinity-org-twin:
re reminder bot's message:

Changelog was added, so for 1, you can say done. For 3 and 4, I think the answer is "no need". Not sure how this code change affects externally visible behavior, so I'm not sure about 2, i.e. whether it is somehow breaking.

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: For now, only `check_if_registry_changed_across_versions` above calls it, which is only interested whether a record changed or not. This change thus relaxes the checks a bit, by allowing a record to be absent, and still be absent, instead of panicking (still rejecting if a record becomes present or becomes absent). This helps the system test mentioned in the PR description (which is still disabled, thus why it is not apparent that it is broken).

pierugo-dfinity:
> re reminder bot's message:
> 
> Changelog was added, so for 1, you can say done. For 3 and 4, I think the answer is "no need". Not sure how this code change affects externally visible behavior, so I'm not sure about 2, i.e. whether it is somehow breaking.

@kpop-dfinity had [already](https://github.com/dfinity/ic/pull/9861#event-24488045198) dismissed it

PR #9864, opened 2026-04-14
chore: add a metric for the number of expired ingress messages in the artifact pool

kpop-dfinity:
It could be useful for detecting overloaded subnets and for debugging system tests (such as the system test which is being deflaked in https://github.com/dfinity/ic/pull/9835 and https://github.com/dfinity/ic/pull/9856).

Note that we already have a similar metric in [the execution environment](https://github.com/dfinity/ic/blob/f644cd861299fc2aa9829e9c0ecd7863c9d549fc/rs/execution_environment/src/scheduler.rs#L778-L780) and a similar log in the [http handler](https://github.com/dfinity/ic/blob/master/rs/http_endpoints/public/src/common.rs#L254-L260); meaning that currently, as far as I know, the ingress pool is the only place in the replica code where we don't log expired ingress messages in some way before/after purging them.

### DIFF
--- a/rs/artifact_pool/src/ingress_pool.rs
+++ b/rs/artifact_pool/src/ingress_pool.rs
@@ -320,22 +320,23 @@ impl MutablePool<SignedIngress> for IngressPoolImpl {
                             .map(ArtifactTransmit::Abort),
                     );
 
-                    let expired_messages_count =
-                        expired_validated.len() + expired_unvalidated.len();
-                    if expired_messages_count > 0 {
-                        warn!(
-                            every_n_seconds => 30,
-                            self.log,
-                            "Purging {} expired ingress messages \
-                            from the validated ingress pool and {} \
-                            from the unvalidated ingress pool",
-                            expired_validated.len(),
-                            expired_unvalidated.len()
-                        );
-                        self.metrics
-                            .ingress_messages_expired
-                            .inc_by(expired_messages_count as u64)
-                    }
+                    let log_if_non_zero = |count, pool_type| {
+                        if count > 0 {
+                            warn!(
+                                every_n_seconds => 30,
+                                self.log,
+                                "Purging {count} expired ingress messages from \
+                                the {pool_type} ingress pool"
+                            );
+                            self.metrics
+                                .ingress_messages_expired
+                                .with_label_values(&[pool_type])
+                                .inc_by(count as u64);
+                        }
+                    };
+
+                    log_if_non_zero(expired_validated.len(), "validated");
+                    log_if_non_zero(expired_unvalidated.len(), "unvalidated");
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


rs/artifact_pool/src/ingress_pool.rs
```diff
@@ -314,12 +312,27 @@ impl MutablePool<SignedIngress> for IngressPoolImpl {
...
+                            self.log,
+                            "Purging {expired_messages_count} expired ingress messages \
+                            from the ingress pool"
+                        );
+                        self.metrics
+                            .ingress_messages_expired
+                            .inc_by(expired_messages_count)
+                    }
+
                     let _unused = self.unvalidated.purge_below(expiry);
```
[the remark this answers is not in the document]
kpop-dfinity: done

rs/ingress_manager/benches/handle_ingress.rs
```diff
@@ -322,21 +322,29 @@ fn handle_ingress(criterion: &mut Criterion) {
...
+                                time_source.as_ref(),
+                                expiry_range,
+                                total_messages as usize,
+                            );
+                            let (mut ingress_pool, message_ids) = setup(
+                                time_source.as_ref(),
+                                pool_config.clone(),
+                                log.clone(),
+                                messages,
+                            );
```
kpop-dfinity, line 339: I don't think that the `IngressPool` should be `Clone`able so I removed the `Clone` derivation. As a result I had to recreate the pool here on every iteration

rs/ingress_manager/benches/handle_ingress.rs
```diff
@@ -322,21 +322,29 @@ fn handle_ingress(criterion: &mut Criterion) {
...
                             time_source.reset();
                             // We skip the first MAX_INGRESS_TTL duration in order to save
                             // overall benchmark time. Also by this time, the ingress
                             // history has become fully populated.
                             let start = time_source.get_relative_time() + MAX_INGRESS_TTL;
                             time_source.set_time(start).unwrap();
-                            history.set_history(message_ids.clone());
+                            history.set_history(message_ids);
+                            let bench_start = Instant::now();
```
kpop-dfinity, line 347: I moved this because I found it weird that we time the setup of the bench as well

rs/artifact_pool/src/ingress_pool.rs
```diff
@@ -314,13 +310,29 @@ impl MutablePool<SignedIngress> for IngressPoolImpl {
...
+                        warn!(
+                            every_n_seconds => 30,
+                            self.log,
+                            "Purging {expired_messages_count} expired ingress messages \
+                            from the ingress pool"
+                        );
+                        self.metrics
+                            .ingress_messages_expired
+                            .inc_by(expired_messages_count as u64)
+                    }
```
eichhorl: Maybe we could log the individual values of both validated and unvalidated sections?

eichhorl (approved):
Cool, thanks!

rs/artifact_pool/src/ingress_pool.rs
kpop-dfinity: Sure, PTAL

eichhorl (approved):

PR #9875, opened 2026-04-14
chore: track subnet query message response size in metrics

mraszyk:
This PR tracks subnet query message response size in metrics. The motivation for this change is ability to monitor response sizes of the `list_canisters` query endpoint added by this [PR](https://github.com/dfinity/ic/pull/9849).
For consistency, the label `ErrorCode::CanisterRejectedMessage` is introduced for the case `Ok(WasmResult::Reject(_))`.

### DIFF
--- a/rs/execution_environment/src/metrics.rs
+++ b/rs/execution_environment/src/metrics.rs
@@ -294,7 +294,7 @@ impl QueryHandlerMetrics {
             subnet_query_message_response_bytes: metrics_registry.histogram_vec(
                 "execution_subnet_query_message_response_bytes",
                 "Response size in bytes of a successful subnet query message execution.",
-                decimal_buckets(0, 7),
+                decimal_buckets_with_zero(0, 7),
                 &["method_name"],
             ),
         }
--- a/rs/execution_environment/src/metrics.rs
+++ b/rs/execution_environment/src/metrics.rs
@@ -1,5 +1,5 @@
 use ic_embedders::wasmtime_embedder::system_api::sandbox_safe_system_state::RequestMetadataStats;
-use ic_error_types::UserError;
+use ic_error_types::{ErrorCode, UserError};
 use ic_management_canister_types_private::QueryMethod;
 use ic_metrics::MetricsRegistry;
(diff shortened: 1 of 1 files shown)


rs/execution_environment/src/metrics.rs
```diff
@@ -288,14 +291,20 @@ impl QueryHandlerMetrics {
                 decimal_buckets(-3, 2),
                 &["method_name", "status"],
             ),
+            subnet_query_message_response_bytes: metrics_registry.histogram_vec(
+                "execution_subnet_query_message_response_bytes",
+                "Response size in bytes of a successful subnet query message execution.",
```
alin-at-dfinity, line 296: This may be OK, but "successful" is a bit vague, as it currently covers both `Reply` and `Reject` responses, but not user errors.

Same for the doc comment above.

alin-at-dfinity (approved):

rs/execution_environment/src/metrics.rs
mraszyk: I did it like for the sake of consistency (given this AI [comment](https://github.com/dfinity/ic/pull/9875#discussion_r3081234915)).

I suppose "successful" refers to `Ok(WasmResult)` - [represented](https://github.com/dfinity/ic/blob/defc932c6d821f420ce788815fc62bc71aba98a5/rs/types/types/src/ingress.rs#L30) as `Completed` in the state tree types - instead of `Err(UserError)`, but `WasmResult` can indeed be `Reply` or `Reject` itself.

But I agree it is confusing: should we then change [this](https://github.com/dfinity/ic/blob/c915f4fbdfc7a5045d7d47e7723b0c6bf19cf2d0/rs/execution_environment/src/metrics.rs#L311) line to match on `Ok(WasmResult::Reply(_))` instead?

mraszyk: Fixed: https://github.com/dfinity/ic/pull/9875/commits/d7aed463ea52220dccd94dffd1c7d08ab5aec879

alin-at-dfinity: LGTM, thank you.

PR #9880, opened 2026-04-14
feat(governance): decouple minimum dissolve delay to vote from propose

jasonz-dfinity:
## Why

Under Mission 70, `neuron_minimum_dissolve_delay_to_vote_seconds` was lowered from 6 months
to 2 weeks. Since the same parameter also gated proposal submission, neurons with just 2 weeks
of dissolve delay can now submit proposals. This decouples the two thresholds so that proposal
submission retains the original 6-month requirement.

## What

- Add a `NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS` constant (6 months) and use it in
  `make_proposal` instead of the governable vote minimum.
- Add validation on `neuron_minimum_dissolve_delay_to_vote_seconds` ensuring it cannot exceed
  the propose constant. Every time a neuron submits a proposal it also votes on it, so the vote
  minimum must stay at or below the propose minimum.

## Testing

- Updated `test_neuron_minimum_dissolve_delay_to_vote_seconds_bounds` for the new validation error

### DIFF
--- a/rs/nns/governance/src/governance.rs
+++ b/rs/nns/governance/src/governance.rs
@@ -285,6 +285,12 @@ pub const MAX_NEURONS_FUND_PARTICIPANTS: u64 = 5_000;
 /// in the same limit.
 const NEURON_RATE_LIMITER_KEY: &str = "ADD_NEURON";
 
+/// The minimum dissolve delay (in seconds) a neuron must have to submit a
+/// non-ManageNeuron proposal. This is intentionally decoupled from the voting
+/// eligibility threshold (`neuron_minimum_dissolve_delay_to_vote_seconds`),
+/// which can be lower.
+pub const NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS: u64 = 6 * ONE_MONTH_SECONDS;
+
 // The maximum dissolve delay allowed for a neuron.
 pub const MAX_DISSOLVE_DELAY_SECONDS_PRE_MISSION_70: u64 = 8 * ONE_YEAR_SECONDS;
 pub const MAX_DISSOLVE_DELAY_SECONDS_POST_MISSION_70: u64 = 2 * ONE_YEAR_SECONDS;
@@ -5226,16 +5232,16 @@ impl Governance {
             ));
         }
 
-        let min_dissolve_delay_seconds_to_vote = if action.manage_neuron().is_some() {
+        let min_dissolve_delay_seconds_to_propose = if action.manage_neuron().is_some() {
             0
         } else {
-            self.neuron_minimum_dissolve_delay_to_vote_seconds()
+            NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS
         };
 
-        // The proposer must be eligible to vote. This also ensures that the
-        // neuron cannot be dissolved until the proposal has been adopted or
-        // rejected.
-        if proposer_dissolve_delay_seconds < min_dissolve_delay_seconds_to_vote {
+        // The proposer must have sufficient dissolve delay to submit proposals.
+        // This threshold is intentionally decoupled from the voting eligibility
+        // threshold, which can be lower.
+        if proposer_dissolve_delay_seconds < min_dissolve_delay_seconds_to_propose {
             return Err(GovernanceError::new_with_message(
                 ErrorType::InsufficientFunds,
                 "Neuron's dissolve delay is too short.",
--- a/rs/nns/governance/src/governance.rs
+++ b/rs/nns/governance/src/governance.rs
@@ -5239,11 +5239,9 @@ impl Governance {
         };
 
         // The proposer must have sufficient dissolve delay to submit proposals.
-        // This threshold is intentionally decoupled from the voting eligibility
-        // threshold, which can be lower.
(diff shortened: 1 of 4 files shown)


rs/nns/governance/src/governance.rs
```diff
@@ -285,6 +285,12 @@ pub const MAX_NEURONS_FUND_PARTICIPANTS: u64 = 5_000;
 /// in the same limit.
 const NEURON_RATE_LIMITER_KEY: &str = "ADD_NEURON";
 
+/// The minimum dissolve delay (in seconds) a neuron must have to submit a
+/// non-ManageNeuron proposal. This is intentionally decoupled from the voting
+/// eligibility threshold (`neuron_minimum_dissolve_delay_to_vote_seconds`),
+/// which can be lower.
+pub const NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS: u64 = 6 * ONE_MONTH_SECONDS;
```
daniel-wong-dfinity-org, line 292: No change.

Why not make this a NervousSystemParameter? I guess this is the simplest way; it just seems inconsistent.

rs/nns/governance/src/governance.rs
```diff
@@ -5226,16 +5232,16 @@ impl Governance {
...
+            NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS
         };
 
-        // The proposer must be eligible to vote. This also ensures that the
-        // neuron cannot be dissolved until the proposal has been adopted or
-        // rejected.
-        if proposer_dissolve_delay_seconds < min_dissolve_delay_seconds_to_vote {
+        // The proposer must have sufficient dissolve delay to submit proposals.
+        // This threshold is intentionally decoupled from the voting eligibility
+        // threshold, which can be lower.
```
daniel-wong-dfinity-org: NCR.

I don't think you need to say this. This comment is not really harmful though either.

rs/nns/governance/src/governance.rs
```diff
@@ -5226,16 +5232,16 @@ impl Governance {
...
+        if proposer_dissolve_delay_seconds < min_dissolve_delay_seconds_to_propose {
             return Err(GovernanceError::new_with_message(
                 ErrorType::InsufficientFunds,
```
daniel-wong-dfinity-org: NCR.

This ErrorType seems wrong (ofc, it was already like this, hence NCR).

rs/nns/governance/src/network_economics.rs
```diff
@@ -380,6 +381,14 @@ impl VotingPowerEconomics {
                 );
                 defects.push(defect);
             }
+            if delay > NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS {
```
daniel-wong-dfinity-org, line 385: NCR

Add blank line.

```suggestion

            if delay > NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS {
```

daniel-wong-dfinity-org, line 385: NCR

Not sure why this is necessary. I mean, yeah, it would be strange if proposing neuron could vote on his own proposal, but not on anybody else's, but strange != actually broken.

daniel-wong-dfinity-org (approved):

rs/nns/governance/src/network_economics.rs
```diff
@@ -380,6 +381,14 @@ impl VotingPowerEconomics {
...
+                let defect = format!(
+                    "neuron_minimum_dissolve_delay_to_vote_seconds ({}) must not exceed \
+                     NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS ({}).",
+                    delay, NEURON_MINIMUM_DISSOLVE_DELAY_TO_PROPOSE_SECONDS,
```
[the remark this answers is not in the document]
jasonz-dfinity: Done.

rs/nns/governance/src/network_economics.rs
jasonz-dfinity: I considered 2 alternatives:
* allow the strange case, but do not allow the proposer to vote: it creates a weird edge case, and it's not entirely clear to me that no code is depending on the proposer votes on their own proposal
* allow the strange case, but still allow the proposer to vote: a neuron with lower dissolve delay can vote, this is confusing, even though it's probably ok

Comparing to those alternatives, I think having this validation is better.

rs/nns/governance/src/governance.rs
jasonz-dfinity: Removed the extra lines, keeping just the first line.

rs/nns/governance/src/governance.rs
jasonz-dfinity: Changed to PreconditionFailed.

rs/nns/governance/src/network_economics.rs
jasonz-dfinity: Done.

rs/nns/governance/src/governance.rs
jasonz-dfinity: I didn't see any advantage, and I think we also have maximum dissolve delay as a constant so it's not unprecedented.

PR #9883, opened 2026-04-15
feat: charge cycles for log memory resize in update_settings

maksymar:
This PR adds cycle charging for log memory resize in `update_settings`.

When `update_settings` changes `log_memory_limit`, the log memory store rebuilds its ring buffer by reading and rewriting all stored records. This work is now charged proportionally to stored data.

Implementation follows a two-step pattern:
1. `validate_canister_settings`: pre-check that the canister can afford the resize cost without dropping below the freezing threshold
2. `update_settings`: deduct cycles and account instructions in the round budget after the resize completes

Only triggers when the user explicitly sets `log_memory_limit` and the resize would actually do work (capacity change), not when the default is applied during canister creation.

Changes:
- Add `LogResizeNotEnoughCycles` error variant mapped to `ErrorCode::CanisterOutOfCycles` with a log-resize-specific message (no public API change).
- Add `would_resize()` to predict whether a resize is a no-op, used both for the charge gate and as the early-return guard inside `resize_impl`.
- Add tests for cycle deduction, rejection on insufficient balance, minimal charge on empty buffers, and no extra charge when the feature is disabled.


rs/config/src/execution_environment.rs
```diff
@@ -12,7 +12,7 @@ const GIB: u64 = 1024 * MIB;
 const TIB: u64 = 1024 * GIB;
 
 // TODO(DSM-105): remove after the feature is enabled by default.
-pub const LOG_MEMORY_STORE_FEATURE_ENABLED: bool = false;
+pub const LOG_MEMORY_STORE_FEATURE_ENABLED: bool = true; // TODO: do not sumit!
```
michael-weigelt: Ping

maksymar: ah, sorry I didn't notify you that I decided that the PR was not ready yet and switched back to draft mode.

maksymar: now I think the PR is ready for review.
I checked the CI was green for both feature enabled and disabled.
PTAL

image

michael-weigelt (approved):

PR #1530, opened 2024-09-17
test(ICRC_ledger): FI-1492: Set subnet configs in golden state tests to cover all canister IDs

mbjorkqvist:
Expand the canister ID ranges for the `StateMachine` app subnet in the golden state tests for the NNS, SNS, and fiduciary subnets to cover all possible canister IDs. This is to avoid scenarios where the tests may fail in case there are some messages to canisters on other subnets that cannot be routed.

PR #9891, opened 2026-04-15
fix(state-machine-tests): DEFI-2779: Register non-local subnets from routing table in registry

mbjorkqvist:
Fix flaky `icrc_ledger_suite_integration_golden_state_upgrade_downgrade_test` by registering all subnets referenced in the routing table in the registry. Extract `add_cup_contents_and_key_record` helper to eliminate code duplication.

The golden state tests use `create_routing_table` (introduced in [PR #1530](https://github.com/dfinity/ic/pull/1530)) to route canister IDs outside the local subnet to a non-existent subnet. This ensures that leftover cross-subnet responses in the golden state backup are routed into a remote stream (where they sit harmlessly) rather than triggering a critical error in the stream builder.

This worked until [PR #9449](https://github.com/dfinity/ic/pull/9449) introduced routing table filtering in `try_to_populate_network_topology`: non-CloudEngine subnets now filter the routing table to only include entries for subnets that have registry records. Since `StateMachineBuilder::build()` only registered the local subnet, the non-existent subnet's routing entries were silently dropped, leaving cross-subnet responses unroutable and causing `mr_stream_builder_response_destination_not_found` critical errors in the golden state test.

The test is flaky (rather than always failing) because, depending on the timing of the snapshot, it may or may not contain messages destined for other subnets.

`StateMachineBuilder::build()` now derives the subnet list from the routing table and registers minimal registry records (DKG transcript, subnet record, key record) for each non-local subnet via `register_non_local_subnet`. This ensures their routing table entries survive the filtering.

### DIFF
--- a/rs/state_machine_tests/src/lib.rs
+++ b/rs/state_machine_tests/src/lib.rs
@@ -499,7 +499,29 @@ fn add_subnet_local_registry_records(
         .with_resource_limits(resource_limits)
         .build();
 
-    // Insert initial DKG transcripts
+    add_cup_contents_and_key_record(
+        subnet_id,
+        ni_dkg_transcript,
+        public_key,
+        &registry_data_provider,
+        registry_version,
+    );
+
+    add_single_subnet_record(
+        &registry_data_provider,
+        registry_version.get(),
+        subnet_id,
+        record,
+    );
+}
+
+fn add_cup_contents_and_key_record(
+    subnet_id: SubnetId,
+    ni_dkg_transcript: NiDkgTranscript,
+    public_key: ThresholdSigPublicKey,
+    registry_data_provider: &Arc<ProtoRegistryDataProvider>,
+    registry_version: RegistryVersion,
+) {
     let mut high_threshold_transcript = ni_dkg_transcript.clone();
     high_threshold_transcript.dkg_id.dkg_tag = NiDkgTag::HighThreshold;
     let mut low_threshold_transcript = ni_dkg_transcript;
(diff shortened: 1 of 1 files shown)


mbjorkqvist:
Details of a successful run with the test (using a "known-bad" golden state snapshot) can be found [here](https://dash.idx.dfinity.network/invocation/c716479f-94a4-43ad-86ad-20f65d625a35).

rs/state_machine_tests/src/lib.rs
```diff
@@ -515,19 +537,48 @@ fn add_subnet_local_registry_records(
...
+    add_cup_contents_and_key_record(
         subnet_id,
+        ni_dkg_transcript,
         public_key,
+        registry_data_provider,
+        INITIAL_REGISTRY_VERSION,
+    );
+
+    let record = SubnetRecordBuilder::from(&[])
+        .with_subnet_type(SubnetType::Application)
```
schneiderstefan, line 581: That's of course an assumption in a golden state test.

I think it's fine, but might end up as a problem at some point. Does your test have access to the real registry and would it be possible to set that correctly? Might not be worth it, but at least a comment here might be valuable.

schneiderstefan (approved):
Thank you and sorry for making your test flaky.

rs/state_machine_tests/src/lib.rs
mbjorkqvist: Good point! I added a comment to clarify that this is an assumption, but that it shouldn't matter (at least for the current use cases?). IIUC, setting this correctly should be possible in the golden state test setup, but it doesn't seem worth the extra effort at this point.

PR #9896, opened 2026-04-15
feat(nns): apply 8 Year Gang 10% bonus to voting power calculation

jasonz-dfinity:
## Why

Neurons that had the maximum dissolve delay (8 years) before Mission 70 reduced it to 2 years
should receive a grandfathered 10% voting power bonus. The bonus base (`eight_year_gang_bonus_base_e8s`)
was already tagged on qualifying neurons (#9593) and exposed in neuron info (#9687), but was not yet
wired into the actual voting power calculation.

## What

- Add `eight_year_gang_bonus()` helper in `voting_power.rs` that returns 10% of the bonus base
- Include the bonus in `potential_and_deciding_voting_power`: effective stake becomes
  `stake + eight_year_gang_bonus(bonus_base)` before dissolve delay and age multipliers are applied
- Cap the bonus base to `stake_e8s` to prevent rejection fees from inflating the bonus
- Add a test verifying the cap behavior

## Testing

- Unit test for the capping behavior added in `types/tests.rs`
- All existing governance tests pass (12 bazel test targets)


rs/nns/governance/src/neuron/types.rs
```diff
@@ -367,7 +367,12 @@ impl Neuron {
...
     ) -> (u64, u64) {
-        let potential_voting_power = Decimal::from(self.stake_e8s())
+        // Cap the bonus base to the current stake because rejection fees can cause the
+        // bonus base to exceed stake_e8s.
+        let eight_year_gang_bonus_base_e8s =
+            self.eight_year_gang_bonus_base_e8s.min(self.stake_e8s());
+        let potential_voting_power = (Decimal::from(self.stake_e8s())
+            + eight_year_gang_bonus(eight_year_gang_bonus_base_e8s))
             * dissolve_delay_bonus_multiplier(self.dissolve_delay_seconds(now_seconds))
             * age_bonus_multiplier(self.age_seconds(now_seconds));
```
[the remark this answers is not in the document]
jasonz-dfinity: Done.

rs/nns/governance/src/neuron/types.rs
```diff
@@ -367,7 +367,12 @@ impl Neuron {
         voting_power_economics: &VotingPowerEconomics,
         now_seconds: u64,
     ) -> (u64, u64) {
-        let potential_voting_power = Decimal::from(self.stake_e8s())
+        // Cap the bonus base to the current stake because rejection fees can cause the
+        // bonus base to exceed stake_e8s.
+        let eight_year_gang_bonus_base_e8s =
+            self.eight_year_gang_bonus_base_e8s.min(self.stake_e8s());
+        let potential_voting_power = (Decimal::from(self.stake_e8s())
+            + eight_year_gang_bonus(eight_year_gang_bonus_base_e8s))
```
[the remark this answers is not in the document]
jasonz-dfinity: Done.

rs/nns/governance/src/neuron/types.rs
```diff
@@ -333,8 +333,12 @@ impl Neuron {
...
     }
 
-    /// Voting power is fundamentally based on the amount staked. From that
-    /// base, it is boosted by two factors:
+    /// Voting power is fundamentally based on the amount staked plus the
+    /// 8 Year Gang bonus (10% of the bonus base, capped to the current stake).
+    /// The bonus base is set for neurons that had the maximum dissolve delay
+    /// before Mission 70 reduced it from 8 years to 2 years.
+    ///
+    /// From that effective stake, voting power is boosted by two factors:
```
daniel-wong-dfinity-org: Do not mention 8y gang bonus in the lede, because it is an esoteric feature (e.g. there is no action you can take now to receive it). Instead, mention it later, after the two bonuses, which people still CAN control.

```suggestion
    /// Voting power is fundamentally based on the amount staked. From that
    /// base, it is boosted by two factors:
```

Then, after describing the two bonuses,

> In addition, there is the "8 year gang bonus", which starts with 10% of the staked amount (as of when Mission 70 began taking effect), applies the two bonuses mentioned above (for dissolve delay and age), and adds to the final (potential) voting power.

rs/nns/governance/src/neuron/voting_power.rs
```diff
@@ -30,6 +30,11 @@ pub(crate) fn age_bonus_multiplier(age_seconds: u64) -> Decimal {
     t / Decimal::from(4) + Decimal::from(1)
 }
 
+pub(crate) fn eight_year_gang_bonus(eight_year_gang_bonus_base_e8s: u64) -> Decimal {
+    // 10% of the base e8s.
+    Decimal::from(eight_year_gang_bonus_base_e8s) / Decimal::from(10)
+}
+
```
daniel-wong-dfinity-org: Why not inline this?

rs/nns/governance/src/neuron/types.rs
```diff
@@ -367,7 +371,12 @@ impl Neuron {
...
     ) -> (u64, u64) {
-        let potential_voting_power = Decimal::from(self.stake_e8s())
+        let stake_e8s = self.stake_e8s();
+        // Cap the bonus base to the current stake because rejection fees can cause the
+        // bonus base to exceed stake_e8s.
+        let eight_year_gang_bonus_base_e8s = self.eight_year_gang_bonus_base_e8s.min(stake_e8s);
+        let potential_voting_power = (Decimal::from(stake_e8s)
+            + eight_year_gang_bonus(eight_year_gang_bonus_base_e8s))
             * dissolve_delay_bonus_multiplier(self.dissolve_delay_seconds(now_seconds))
             * age_bonus_multiplier(self.age_seconds(now_seconds));
```
daniel-wong-dfinity-org, line 377: DAYSF

```suggestion
        let stake_e8s = Decimal::from(self.stake_e8s());
        let boost = dissolve_delay_bonus_multiplier(self.dissolve_delay_seconds(now_seconds))
            * age_bonus_multiplier(self.age_seconds(now_seconds))
        let mut potential_voting_power = stake_e8s * boost;

        // 8 year gang bonus.
        potential_voting_power +=
            self.eight_year_gang_bonus_base_e8s.min(stake_e8s) * Decimal::from(0.1) * boost;
```

This way, 8 year gang bonus calculation is shunted off, clearly separated from the main calculation, which affects everyone. This makes the main calculation simpler. No need to cram everything into one expression.

rs/nns/governance/src/neuron/types.rs
```diff
@@ -367,7 +371,12 @@ impl Neuron {
         voting_power_economics: &VotingPowerEconomics,
         now_seconds: u64,
     ) -> (u64, u64) {
-        let potential_voting_power = Decimal::from(self.stake_e8s())
+        let stake_e8s = self.stake_e8s();
+        // Cap the bonus base to the current stake because rejection fees can cause the
+        // bonus base to exceed stake_e8s.
+        let eight_year_gang_bonus_base_e8s = self.eight_year_gang_bonus_base_e8s.min(stake_e8s);
```
daniel-wong-dfinity-org: Really not sure why `.min(stake_e8s)`. I guess this is for fees? But if you do it this way, fees will not get applied to the 8y bonus base if the neuron is topped up.

daniel-wong-dfinity-org (approved):

rs/nns/governance/src/neuron/types.rs
jasonz-dfinity: Done — rewrote the lede to lead with the two controllable bonuses, and moved the 8 Year Gang bonus into its own paragraph after them (using your suggested phrasing).

rs/nns/governance/src/neuron/voting_power.rs
jasonz-dfinity: Done — inlined as `Decimal::from(eight_year_gang_bonus_base_e8s) / Decimal::from(10) * boost` and removed the helper from voting_power.rs.

rs/nns/governance/src/neuron/types.rs
jasonz-dfinity: Done — factored out `stake_e8s` and `boost`, and moved the 8 Year Gang bonus into its own `+=` term with a dedicated comment. Had to tweak your snippet slightly: `.min(stake_e8s)` needs to operate on the `u64` side, so the structure is the same but the cap is applied before the `Decimal::from` conversion.

rs/nns/governance/src/neuron/types.rs
jasonz-dfinity: Yes, and I agree that in the top-up case the penalty is not maximized. However, IMHO it's OK as the 90.9% of the penalty is already carried out by the fees themselves. Maximizing the penalty doesn't seem to worth adding the complexity of introducing another mechanism of changing this bonus base as proposals are made and adopted.

PR #9899, opened 2026-04-16
chore: Use `StateReader` instead of `StateManager` where possible

eichhorl:
This PR changes some consensus crates to use the more restrictive `StateReader` interface where possible. Doing so helps in clarifying the interaction between consensus and DSM.


pierugo-dfinity (approved):

kpop-dfinity (approved):
Cool.

PR #6526, opened 2025-08-29
perf(idkg): [CON-1557] reduce the number of support share validations

pierugo-dfinity:
Currently, for every IDKG dealing, we validate all `n` support shares. However, we actually only need `2f+1` support shares in order to complete the transcript. Since the validation of support shares represents the largest amount of computation required by the IDKG protocol, this PR reduces the number of support share validations and improves overall performance.

It does so by maintaining a separate in-memory map, that keeps track of how many support shares were already validated for a given dealing. If we already successfully validated at least `2f+1` support shares according to this map, any additional shares are removed from the unvalidated pool.

The difference in performance is quite significant!
Before:
image

After:
image

P.S.: We also tried removing the [`has_node_issued_dealing_support`](https://github.com/dfinity/ic/pull/6526/files#diff-dbc1abf369d69171bc8d70760b200793813e40581ab3f54e298e87d18f12e83aL429-L435) check to see if it would improve performance, but it made no difference.

P.S. 2: A similar optimization (of validating only `2f+1` shares instead of `n`) led to a race condition two years ago because a membership change modified the verification threshold. This cannot happen here because the number of receivers (and thus also the verification threshold) is fixed throughout an execution of the IDKG protocol.

PR #9907, opened 2026-04-16
perf(idkg): [CON-1575] reduce the number of signature share validations

pierugo-dfinity:
This PR reduces the number of signature share validations from `N` to `f+1`, which are enough to aggregate the full signature. The strategy is very similar to the one done in #6526 to reduce the number of dealing supports: we introduce an in-memory map that keeps track for each `RequestId`, the set of share signers already validated. Once that set reaches a cardinality of `f+1`, we stop validating more signature shares for that `RequestId`. We follow the same approach as in #7069 to do all that in parallel: check with shared access to the lock, then perform the cryptographic validation and if valid, check again with exclusive access to the lock.

### DIFF
--- a/rs/consensus/idkg/src/signer.rs
+++ b/rs/consensus/idkg/src/signer.rs
@@ -1433,7 +1433,7 @@ mod tests {
             });
         }
 
-        // In the single threaded case only 2f + 1 shares should be accepted, the rest dropped
+        // In the single threaded case only f + 1 shares should be accepted, the rest dropped
         ic_test_utilities::artifact_pool_config::with_test_pool_config(|pool_config| {
             with_test_replica_logger(|logger| {
                 let (mut idkg_pool, signer) =
@@ -1467,7 +1467,7 @@ mod tests {
             })
         });
 
-        // In the multi threaded case at least 2f + 1 shares should be accepted, the rest dropped
+        // In the multi threaded case at least f + 1 shares should be accepted, the rest dropped
         ic_test_utilities::artifact_pool_config::with_test_pool_config(|pool_config| {
             with_test_replica_logger(|logger| {
                 let (mut idkg_pool, signer) = create_signer_dependencies(pool_config, logger);
--- a/rs/consensus/idkg/src/signer.rs
+++ b/rs/consensus/idkg/src/signer.rs
@@ -41,7 +41,7 @@ use rayon::{
 use std::{
     collections::{BTreeMap, BTreeSet},
     fmt::{self, Debug, Formatter},
-    sync::Arc,
+    sync::{Arc, RwLock},
 };
 
 #[derive(Clone, Debug)]
@@ -115,6 +115,7 @@ pub(crate) struct ThresholdSignerImpl {
     crypto: Arc<dyn ConsensusCrypto>,
     thread_pool: Arc<ThreadPool>,
     state_reader: Arc<dyn StateReader<State = ReplicatedState>>,
+    validated_sig_share_signers: RwLock<BTreeMap<RequestId, BTreeSet<NodeId>>>,
     metrics: ThresholdSignerMetrics,
     log: ReplicaLogger,
 }
@@ -133,6 +134,7 @@ impl ThresholdSignerImpl {
             crypto,
             thread_pool,
             state_reader,
+            validated_sig_share_signers: RwLock::new(BTreeMap::new()),
             metrics: ThresholdSignerMetrics::new(metrics_registry),
             log,
         }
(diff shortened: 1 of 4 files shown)


rs/consensus/idkg/src/signer.rs
```diff
@@ -290,13 +303,27 @@ impl ThresholdSignerImpl {
...
+        // different thread validated a share for the same request_id in the meantime, after we
+        // released the read lock. Therefore, we acquire the write lock here to check again with
+        // exclusive access.
+        let mut valid_sig_share_signers = self.validated_sig_share_signers.write().unwrap();
+        let signers = valid_sig_share_signers.entry(request_id).or_default();
+        if !signers.insert(signer) {
+            self.metrics
+                .sign_errors_inc("duplicate_sig_share_cache_miss");
+            return Some(IDkgChangeAction::RemoveUnvalidated(id));
         }
```
eichhorl: Alternatively we could move this into the `Ok(share)` branch above

rs/consensus/idkg/src/signer.rs
```diff
@@ -1020,11 +1111,27 @@ mod tests {
...
                     assert!(change_set.is_empty());
+                    assert_eq!(
+                        signer.validated_sig_share_signers(),
+                        BTreeMap::from([
+                            (ids[0], BTreeSet::from([NODE_1])),
+                            (ids[1], BTreeSet::from([NODE_1])),
+                            (ids[2], BTreeSet::from([NODE_1])),
+                        ])
+                    );
                 }
```
eichhorl: Should this assertion be outside of the `if` block?

rs/consensus/idkg/src/signer.rs
```diff
@@ -1275,6 +1383,129 @@ mod tests {
...
+            let message = create_signature_share(&key_id, *node_id, id);
+            msg_ids.push(message.message_id());
+            artifacts.push(UnvalidatedArtifact {
+                message,
+                peer_id: *node_id,
+                timestamp: UNIX_EPOCH,
+            });
+        }
+
+        // In the single threaded case only 2f + 1 shares should be accepted, the rest dropped
```
eichhorl: Should it be f+1 (for ecdsa/schnorr)?
```suggestion
        // In the single threaded case only f + 1 shares should be accepted, the rest dropped
```

rs/consensus/idkg/src/signer.rs
```diff
@@ -1275,6 +1383,129 @@ mod tests {
...
+                        .get(&id)
+                        .is_some_and(|signers| {
+                            signers.len() == expected_nb_sig_shares
+                                && signers.is_subset(&node_ids.iter().cloned().collect())
+                        })
+                );
+            })
+        });
+
+        // In the multi threaded case at least 2f + 1 shares should be accepted, the rest dropped
```
eichhorl: Same here

rs/consensus/idkg/src/signer.rs
pierugo-dfinity: Indeed https://github.com/dfinity/ic/pull/9907/commits/5cab8252b4971b88407e86abc0541d57c4331d74

rs/consensus/idkg/src/signer.rs
pierugo-dfinity: Thank you https://github.com/dfinity/ic/pull/9907/commits/692a7990c881d071e61868656c840cbf4c61d970

rs/consensus/idkg/src/signer.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/9907/commits/692a7990c881d071e61868656c840cbf4c61d970

rs/consensus/idkg/src/signer.rs
pierugo-dfinity: Good idea https://github.com/dfinity/ic/pull/9907/commits/bfa2d11b13704b1dd03db592e543ad4925c9fc9b

rs/consensus/idkg/src/signer.rs
```diff
@@ -536,14 +567,31 @@ impl ThresholdSignerImpl {
         }
     }
 
+    fn inputs_already_have_enough_shares(
+        &self,
```
eichhorl: I guess we don't technically need the reference to `self` here. One alternative could be to implement this as a function of `ThresholdSigInputs`, but no strong opinion.

eichhorl (approved):

rs/consensus/idkg/src/signer.rs
pierugo-dfinity: Good point. I think it would be interesting to move the function to `ThresholdSigInputs`, named something like `have_enough_shares`. Though in this context, the comment `As this code path is an optimization` does not apply anymore since the function is much more public and can be used in a different context. As we would always return `false` for VetKD (at least until the API is modified), I am scared that some different caller in the future may not realize that and shoot themselves in the foot.

I thus removed `&self`, but kept it as a private function of `ThresholdSignerImpl` such that it can only be used there for now. If the VetKD API is adapted, I would be happy with moving it to `ThresholdSigInputs`.
So actually `signer_has_issued_share` and `should_purge` also didn't need `self`. I removed them there too.

PR #9908, opened 2026-04-16
feat(flexible-outcalls): CON-1714 extend ResponsesTooLarge error for good error message

fspreiss:
### Summary

- Adds `total_requests` and `min_responses` fields to the `FlexibleCanisterHttpResponsesTooLarge` proto message and the corresponding `FlexibleCanisterHttpError::ResponsesTooLarge` Rust enum variant.
- Validates these fields in the payload validator, rejecting payloads where a block maker claims incorrect values.
- Extends the existing tests accordingly and adds two new ones for testing the validation logic

### Motivation
A follow-up change will deliver `ResponsesTooLarge` errors to execution as Candid-encoded `FlexibleHttpRequestResult::Err` values. The error message needs `total_requests` and `min_responses` to properly explain *why* the responses don't fit. Here is an example of the envisioned error message:
```
Responses too large: need at least 2 (= 3 min_responses - 1 unseen) OK responses to fit within 2097152 bytes, but even the smallest 2 of 3 OK responses have sizes [1200000, 1300000] bytes (3 ok + 1 reject + 1 unseen = 5 total_requests)
```
Without the new fields in the error, the delivery code would have to re-derive them from the HTTP request context at `into_messages` time, which it doesn't have access to.

### DIFF
--- a/rs/protobuf/def/types/v1/canister_http.proto
+++ b/rs/protobuf/def/types/v1/canister_http.proto
@@ -89,6 +89,8 @@ message FlexibleCanisterHttpTimeout {}
 
 message FlexibleCanisterHttpResponsesTooLarge {
   repeated CanisterHttpShare all_seen_shares = 1;
+  uint32 total_requests = 2;
+  uint32 min_responses = 3;
 }
 
 message FlexibleCanisterHttpTooManyRequestErrors {
--- a/rs/https_outcalls/consensus/src/payload_builder.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder.rs
@@ -762,8 +762,32 @@ impl CanisterHttpPayloadBuilderImpl {
                     }
                 }
                 FlexibleCanisterHttpError::ResponsesTooLarge {
-                    all_seen_shares, ..
+                    all_seen_shares,
+                    total_requests,
+                    min_responses: claimed_min_responses,
+                    ..
                 } => {
+                    if *total_requests != flex_committee.len() as u32 {
+                        return invalid_artifact(
+                            InvalidCanisterHttpPayloadReason::FlexibleResponsesTooLargeParamMismatch {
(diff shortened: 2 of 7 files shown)


rs/protobuf/def/types/v1/canister_http.proto
```diff
@@ -89,6 +89,8 @@ message FlexibleCanisterHttpTimeout {}
 
 message FlexibleCanisterHttpResponsesTooLarge {
   repeated CanisterHttpShare all_seen_shares = 1;
+  uint32 total_requests = 2;
+  uint32 min_responses = 3;
```
eichhorl, line 93: I guess you could argue that whoever made the request should know which values they chose, but fine with me if it makes the response easier to understand 👍

eichhorl (approved):

rs/protobuf/def/types/v1/canister_http.proto
fspreiss: Indeed, the caller does know which values they chose. Still, making the error message easy to understand seems valuable, and Grégory always stressed that debuggability and the ability to easily understanding why things fail is very important to them.

An alternative approach would be to adapt the `IntoMessages::into_messages` trait, e.g., let it take `&self` as then we could access the state and subsequently the contexts (which have the necessary info) directly there. But there are a few subtleties with this:
1. Which state? We don't just want the latest state but the state height the was used during payload/building validation, so we'd need at least a second parameter (a certified height, or the `ValidationContext`).
2. into_messages should be infallible if validate_payload returned `Ok`; but reading the state introduces a failure mode.
3. Probably some other effects...

Even if we solve/overcome this, it doesn't seem worth it as of now to go that more complex route. So, if we want a meaningful error message (and I think we do), then adding these two fields seems by far the simplest solution.

PR #9913, opened 2026-04-16
test(execution): add sender_info system test

aterga:
## Summary

- Adds a new `sender_info_test` Bazel system test target in `rs/tests/execution/`
- Tests the execution-side behavior of the `sender_info` feature (certified auxiliary data attached to ingress messages by canister-signature callers)
- Complements the existing validation-focused tests in `rs/tests/crypto/ingress_verification_test.rs`

**Test cases covered:**
- `query_reads_sender_info_data_and_signer` — query call reads both `msg_caller_info_data` and `msg_caller_info_signer` via system API
- `update_reads_sender_info_data_and_signer` — update call stores sender_info to stable memory; verified via follow-up queries
- `no_sender_info_returns_empty` — both APIs return empty bytes when no sender_info is present
- `inter_canister_call_does_not_propagate_sender_info` — sender_info is NOT forwarded to downstream inter-canister calls
- `reply_callback_can_access_sender_info` — reply callbacks on the original call frame can still access sender_info

The test uses self-contained infrastructure (CanisterSigner, raw HTTP helpers) adapted from `ingress_verification_test.rs`.

## Test plan

- [ ] `bazel test //rs/tests/execution:sender_info_test` passes on CI (Linux)
- [ ] All 5 test functions exercise distinct sender_info behaviors
- [ ] No existing tests are modified

### DIFF
--- a/rs/tests/execution/BUILD.bazel
+++ b/rs/tests/execution/BUILD.bazel
@@ -177,6 +177,31 @@ system_test_nns(
 )
 
 system_test(
+    name = "sender_info_test",
+    runtime_deps = UNIVERSAL_CANISTER_RUNTIME_DEPS,
+    deps = [
+        # Keep sorted.
+        "//rs/crypto/internal/crypto_lib/basic_sig/der_utils",
+        "//rs/crypto/sha2",
+        "//rs/crypto/test_utils",
+        "//rs/registry/subnet_type",
+        "//rs/tests/driver:ic-system-test-driver",
+        "//rs/types/types",
+        "//rs/universal_canister/lib",
+        "@crate_index//:anyhow",
+        "@crate_index//:ic-agent",
+        "@crate_index//:ic-certification",
+        "@crate_index//:reqwest",
+        "@crate_index//:serde",
+        "@crate_index//:serde_bytes",
+        "@crate_index//:serde_cbor",
+        "@crate_index//:simple_asn1",
+        "@crate_index//:slog",
+        "@crate_index//:tokio",
+    ],
+)
+
+system_test(
     name = "system_api_security_test",
     cpus = MIN_LOCAL_CPUS + 2 * DEFAULT_VCPUS_PER_VM,  # 2 IC Node VMs * 6 vCPUs.
     deps = [
--- a/rs/tests/execution/BUILD.bazel
+++ b/rs/tests/execution/BUILD.bazel
@@ -189,7 +189,6 @@ system_test(
         "@crate_index//:anyhow",
         "@crate_index//:ic-agent",
         "@crate_index//:ic-certification",
-        "@crate_index//:reqwest",
         "@crate_index//:serde",
         "@crate_index//:serde_bytes",
         "@crate_index//:serde_cbor",
--- a/rs/tests/execution/BUILD.bazel
+++ b/rs/tests/execution/BUILD.bazel
@@ -181,9 +181,7 @@ system_test(
     runtime_deps = UNIVERSAL_CANISTER_RUNTIME_DEPS,
     deps = [
         # Keep sorted.
-        "//rs/crypto/internal/crypto_lib/basic_sig/der_utils",
         "//rs/crypto/sha2",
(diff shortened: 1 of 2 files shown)


michael-weigelt (approved):

mraszyk:
We discussed offline that this system test will only be merged after some fixes.

rs/tests/execution/sender_info_test.rs
```diff
@@ -0,0 +1,317 @@
...
+// ---------------------------------------------------------------------------
+
+/// Given an already-installed UC that will act as both signer and target,
+/// returns an `ic-agent` that carries a canister-signed `sender_info` on every
+/// request via `InfoAwareIdentity`, plus the raw `info` bytes for assertions.
+async fn build_info_aware_agent(
+    node_url: &str,
+    bootstrap_agent: ic_agent::Agent,
+    canister_id: Principal,
+) -> (ic_agent::Agent, Vec<u8>) {
```
mraszyk: no need to return the bytes if they're constant anyway

rs/tests/execution/sender_info_test.rs
```diff
@@ -0,0 +1,317 @@
...
+    block_on(async move {
+        let bootstrap_agent = node.build_default_agent_async().await;
+        let canister = UniversalCanister::new_with_retries(
+            &bootstrap_agent,
+            node.effective_canister_id(),
+            &logger,
+        )
+        .await;
+        let canister_id = canister.canister_id();
+        let signer_bytes = canister_id.as_slice().to_vec();
```
mraszyk: I'd move this down to just before the assertion since otherwise it's unclear why this is defined.

rs/tests/execution/sender_info_test.rs
```diff
@@ -0,0 +1,317 @@
...
+            .query(&canister_id, "query")
+            .with_arg(wasm().msg_caller_info_signer().append_and_reply().build())
+            .call()
+            .await
+            .expect("query for msg_caller_info_signer failed");
+        assert_eq!(reply, signer_bytes);
+    });
+}
+
+pub fn update_reads_sender_info(env: TestEnv) {
```
mraszyk: I guess this could be deduplicated via an enum over {Update,Query}

rs/tests/execution/sender_info_test.rs
aterga: Dropped the `Vec<u8>` return from `build_info_aware_agent` and now use `INFO_BYTES` directly in the assertions.

mraszyk (approved):

rs/tests/execution/sender_info_test.rs
aterga: Already addressed in commit 6caa6043bb: `signer_bytes` local was removed and `canister_id.as_slice()` is now inlined at the assertion site.

rs/tests/execution/sender_info_test.rs
aterga: Deduped in 6f941e23d6: added a `CallMode { Query, Update }` enum and a single `run_sender_info_test` that dispatches on it; the two `pub fn` bodies are now one-line entry points.

rs/tests/execution/BUILD.bazel
```diff
@@ -176,6 +176,28 @@ system_test_nns(
     ],
 )
 
+system_test(
+    name = "sender_info_test",
```
basvandijk, line 180: @aterga could you add the following such that the `_local` variant of this system-tests requests the required amount of CPUs from bazel:
```suggestion
    name = "sender_info_test",
    cpus = MIN_LOCAL_CPUS + 1 * DEFAULT_VCPUS_PER_VM,  # 1 IC Node VM with 6 vCPUs.
```
This is important once we switch CI to Namespace and start running system-tests on their Remote Build Execution infra instead of on our Farm.

aterga: Addressed in follow-up PR https://github.com/dfinity/ic/pull/10784.

PR #9915, opened 2026-04-16
feat(governance): enable Mission 70 voting rewards

jasonz-dfinity:
`rs/nns/governance/src/lib.rs`: replace `Cell::new(cfg!(feature = "test"))` with `Cell::new(true)`.

Effects that turn on:
- Max neuron dissolve delay capped at 2 years, with existing neurons clamped (#9688)
- Voting rewards pool scaled by 0.6329 (#9698, later re-calculated)
- Quadratic dissolve delay bonus, replacing linear (#9719)
- Reduced minimum dissolve delay required to vote; vote/propose thresholds decoupled (#9689, #9880)
- 8 year gang 10% voting power bonus (#9687, #9896)
- Voting power snapshots cleared on activation (#9731)


bjoernek:
I think this PR superseded by 
https://github.com/dfinity/ic/pull/9923/

PR #9917, opened 2026-04-17
fix: SyncMessageRouting::expected_batch_height

mraszyk:
Fix `SyncMessageRouting::expected_batch_height` by considering the `tip_height` (which is now additionally exposed via `trait StateManager`) since `tip_height` can be ahead of `latest_state_height` if the fast-forward mode in `commit_and_certify` triggers or due to async hashing in a follow-up PR.

### DIFF
--- a/rs/interfaces/state_manager/mocks/src/lib.rs
+++ b/rs/interfaces/state_manager/mocks/src/lib.rs
@@ -42,6 +42,8 @@ mock! {
     }
 
     impl StateManager for StateManager {
+        fn tip_height(&self) -> Height;
+
         fn take_tip(&self) -> (Height, ReplicatedState);
 
         fn take_tip_at(&self, h: Height) -> StateManagerResult<ReplicatedState>;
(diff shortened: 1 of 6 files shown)


schneiderstefan:
Could we also fix (prod) [MessageRouting::expected_batch_height](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/messaging/src/message_routing.rs?L1804)?

michael-weigelt (approved):

mraszyk:
> Could we also fix (prod) [MessageRouting::expected_batch_height](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/messaging/src/message_routing.rs?L1804)?

This one does not need any fix because tip height cannot be ahead of both the recent batch (regular execution flow) and latest state height (state sync flow). Or did I/Claude miss something?

schneiderstefan (approved):
I think you're right. If we are in catch-up mode then last_seen_height will still increase.

PR #9919, opened 2026-04-17
feat: enable canister creation with subnet ID in the URL

mraszyk:
This PR adds a new public HTTP endpoint `/api/v4/subnet/<effective_subnet_id>/call` on the replica that supports call to the `create_canister` endpoint of the management canister (`aaaaa-aa`). It is specified in this [PR](https://github.com/dfinity/portal/pull/6224).

### DIFF
--- a/rs/http_endpoints/public/src/call/call_sync.rs
+++ b/rs/http_endpoints/public/src/call/call_sync.rs
@@ -136,7 +136,7 @@ pub(crate) fn route(version: Version) -> &'static str {
     }
 }
 
-pub(crate) fn route_v4_subnet() -> &'static str {
+pub(crate) fn route_subnet_v4() -> &'static str {
     "/api/v4/subnet/{effective_subnet_id}/call"
 }
 
@@ -188,7 +188,7 @@ pub fn new_service(
     BoxCloneService::new(router.into_service())
 }
 
-pub(crate) fn new_router_v4_subnet(
+pub(crate) fn new_subnet_v4_router(
     call_handler: IngressValidator,
     ingress_watcher_handle: IngressWatcherHandle,
     metrics: HttpHandlerMetrics,
@@ -207,7 +207,7 @@ pub(crate) fn new_router_v4_subnet(
     };
 
     Router::new().route_service(
-        route_v4_subnet(),
+        route_subnet_v4(),
         axum::routing::post(call_sync_subnet)
             .with_state(state)
             .layer(ServiceBuilder::new().layer(DefaultBodyLimit::disable())),
--- a/rs/http_endpoints/public/src/call/call_sync.rs
+++ b/rs/http_endpoints/public/src/call/call_sync.rs
@@ -54,6 +54,8 @@ pub enum Version {
     V3,
     // Synchronous endpoint with the NNS delegation using the tree format of the canister ranges.
     V4,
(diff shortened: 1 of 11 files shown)


rs/http_endpoints/public/src/call/call_sync.rs
eichhorl, line 135: Why not just add a new variant here?

mraszyk: My thinking was that we aren't really introducing a new version of the API, but rather a "subnet" variant at the existing V4.

rs/http_endpoints/public/tests/test.rs
```diff
@@ -1730,3 +1786,136 @@ fn test_call_response_when_p2p_not_running(
...
+                    .with_method_name(method_name.to_string()),
+            )
+            .await;
+
+        assert_eq!(
+            StatusCode::BAD_REQUEST,
+            response.status(),
+            "{:?}",
+            response.text().await
+        );
```
eichhorl, line 1924: Should we also check the response message (like in `test_call_v4_subnet_wrong_subnet_id`)

eichhorl (approved):

rs/http_endpoints/public/src/call/call_sync.rs
eichhorl: Sure, but for the purpose of this file I feel like that's exactly what it is (we could call it `V4Subnet` or something). Then we could likely avoid introducing all the `_subnet` function variants.

venkkatesh-sekar (approved):

rs/http_endpoints/public/tests/test.rs
mraszyk: [7ddcae0](https://github.com/dfinity/ic/pull/9919/commits/7ddcae0df492cae7b87acb8fa4fd056430c60b0e)

rs/http_endpoints/public/src/call/call_sync.rs
mraszyk: [2899676](https://github.com/dfinity/ic/pull/9919/commits/289967660bd902e66f68ef34f37fa63dc89f06a4)

rs/http_endpoints/public/src/call/call_sync.rs
```diff
@@ -187,21 +199,56 @@ pub fn new_service(
...
+    State(state): State<SynchronousCallHandlerState>,
+    WithTimeout(Cbor(request)): WithTimeout<Cbor<HttpRequestEnvelope<HttpCallContent>>>,
+) -> SyncCallResponse {
+    let delegation_filter = match state.version {
+        Version::V3 => CanisterRangesFilter::Flat,
+        Version::V4 => CanisterRangesFilter::Tree(effective_canister_id),
+        Version::SubnetV4 => CanisterRangesFilter::None,
+    };
+    call_sync_impl(
+        EffectiveDestination::Canister(effective_canister_id),
```
eichhorl: Couldn't we also determine `EffectiveDestination` based on `state.version`, and then get rid of `call_sync_subnet` below?

mraszyk: https://github.com/dfinity/ic/pull/9919/commits/7d06f2f6158ed9d2678b07a700527e0d0900b2f1

eichhorl: Thanks!

PR #9922, opened 2026-04-17
feat: Add counter metrics for consumed cycles

dsarlis:
Add counter versions of metrics for consumed cycles that are stored in `ReplicatedState`. The existing ones behave like gauges (so their values can go down when prepayments are made and up when refunds are issued) which makes it more challenging for consumers to build automated monitoring tools to perform aggregations over them. By having them monotonically increase, it's easier to calculate rates of change, show aggregates over time etc.

The key idea is to introduce a second map of `<CyclesUseCase, NominalCycles>` in the `ReplicatedState` that will only be updated once per use case: either at the payment stage if we know the precise amount or only at refund stage if a prepayment is made with an expected refund later. The second map is quite similar to the existing in all other aspects (how they are stored in checkpoints or how they are exposed as prometheus metrics) besides how the values are updated.

A new map is introduced to ease the transition as migrating from the old map to new is non-trivial given that a proper cutoff point needs to be introduced to handle outstanding callbacks that might have been created before the metric introduction. This is left for a follow-up if and when people decide to do it. The new map will be used in a follow up that will implement the [new management canister endpoint](https://github.com/dfinity/portal/pull/6217) to retrieve canister level metrics.

Additionally, the new metrics include the use case `HttpsOutcalls` in the canister level metrics as it's useful to determine how much each canister uses this feature. I've opted to not change existing metrics to do the same as it would make things less clean imo than the current approach -- a single specific API is used to perform this update in exactly one place where it's needed.

The changes in the PR are mostly driven by the addition of the new map of metrics, updates in protobuf files to store the new metrics, the changes to support having the `HttpsOutcalls` use case additionally included as well as some changes in tests to support the new metrics.

### DIFF
--- a/rs/execution_environment/src/execution/response/tests.rs
+++ b/rs/execution_environment/src/execution/response/tests.rs
@@ -98,8 +98,18 @@ fn execute_response_refunds_cycles() {
 
         // Canister A calls canister B.
         let cycles_sent = Cycles::new(1_000_000);
+        let b_callback = wasm()
+            .accept_cycles(cycles_sent / 2_u64)
+            .message_payload()
+            .append_and_reply()
+            .build();
         let wasm_payload = wasm()
-            .call_with_cycles(b_id, "update", call_args(), cycles_sent)
+            .call_with_cycles(
+                b_id,
+                "update",
+                call_args().other_side(b_callback.clone()),
+                cycles_sent,
+            )
             .build();
 
         // Enqueue ingress message to canister A and execute it.
@@ -107,16 +117,11 @@ fn execute_response_refunds_cycles() {
         assert_matches!(test.ingress_state(&msg_id), IngressState::Received);
         test.execute_message(a_id);
 
-        // Create response from canister B to canister A.
-        let response = ResponseBuilder::new()
-            .originator(a_id)
-            .respondent(b_id)
-            .originator_reply_callback(CallbackId::from(1))
-            .refund(cycles_sent / 2_u64)
-            .build();
-        let response_payload_size = response.payload_size_bytes();
+        // Execute message on B.
+        test.induct_messages();
+        test.execute_message(b_id);
 
-        // Execute response.
+        // Execute response on A.
         let balance_before = test.canister_state(a_id).system_state.balance();
         let consumed_cycles_before = *test
             .canister_state(a_id)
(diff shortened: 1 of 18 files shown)


rs/execution_environment/src/execution/response/tests.rs
```diff
@@ -98,25 +98,30 @@ fn execute_response_refunds_cycles() {
 
         // Canister A calls canister B.
         let cycles_sent = Cycles::new(1_000_000);
+        let b_callback = wasm()
```
dsarlis, line 101: Drive-by improvement: using actual callbacks and canister requests (instead of injecting a response) to make the test more robust.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1978,26 +1998,53 @@ impl SystemState {
             ConsumingCycles::Refund => {}
         }
 
-        // Skip if the amounts are zero and no metric updates are needed.
-        if (consuming_cycles == ConsumingCycles::Prepayment && prepayment.is_zero())
-            || (consuming_cycles == ConsumingCycles::Refund && refund.is_zero())
-        {
+        // Skip if the consumed cycles are zero and no metric updates are needed.
+        if prepayment - refund == NominalCycles::zero() {
```
dsarlis, line 2002: The previous one was not good enough after the changes here as it'd miss cases where you really used all of the prepayment (i.e. refund would be 0). Instead simplify the condition to be "consumption happened" whether it was during prepayment or refund phase.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1978,26 +1998,53 @@ impl SystemState {
...
+                    CyclesUseCase::ECDSAOutcalls
+                    | CyclesUseCase::SchnorrOutcalls
+                    | CyclesUseCase::VetKd
+                    | CyclesUseCase::HTTPOutcalls
+                    | CyclesUseCase::DeletedCanisters
+                    | CyclesUseCase::DroppedMessages
+                    | CyclesUseCase::NonConsumed => {
+                        // These use cases should not be tracked on the canister level.
+                    }
+                }
```
dsarlis, line 2042: I know it's not ideal that we need to "redo" a match here when we've had debug_asserts earlier in this function. I didn't have any quick ideas for improvements, besides trying to introduce even more type hierarchy and sort of differentiate types more from each other (e.g. can they be canister level or not). It's unclear to me how much effort that would be so opted for a match instead.

dsarlis (commented):
Open points:

1. Naming is something that I deliberately did not spend too much on. I can totally understand if a suffix `_as_counters` is not considered good enough and open to ideas for alternate names.
2. In tests you'll notice that I've added a few places where I check the new counters additionally to old ones. One might wonder "what if we try to add such checks universally in `ExecutionTest`". I tried it and as it turns out it's not easy to make all tests work as some of them cut corners and you're not always able to do the check at "clear points", i.e. points where you know that there can be no outstanding callbacks (`execute_all` would be the best candidate and you get majority of tests to work but you still miss some special ones, highly concentrated in `hypervisor_tests`). I'd leave as an improvement if anyone wants to pick it up.

pierugo-dfinity (approved):
Approving changes in `rs/monitoring/metrics`

michael-weigelt (approved):

PR #9925, opened 2026-04-17
feat(flexible-outcalls): CON-1714 deliver flexible errors to execution

fspreiss:
Implements the `IntoMessages` path for flexible HTTP outcall errors, converting each `FlexibleCanisterHttpError` from the consensus payload into a single Candid-encoded `ConsensusResponse` carrying a `FlexibleHttpRequestResult::Err` for delivery to the execution environment.

For each flexible error, a `FlexibleHttpRequestErr` is constructed with:
- `global_error`: `Timeout`, `TooManyRequestErrors`, or `ResponsesTooLarge` (matching the error variant).
- `node_details`: per-node entries with `code` and `message`.
  - `Timeout`: empty (no per-node info is carried in the payload today).
  - `TooManyRequestErrors`: one entry per reject with the adapter's reject code and message.
  - `ResponsesTooLarge`: one entry per seen share with `code` being `"ok"` or `"reject"` and `message` being the content size in bytes.
- `message`: a diagnostic summary. For `ResponsesTooLarge`, this includes the explicit calculation of why the limit could not be met, e.g. `"Responses too large: need at least 5 (= 8 min_responses - 3 unseen) OK responses to fit within 2097152 bytes, but even the smallest 5 of 7 OK responses have sizes [...] bytes (7 ok + 3 reject + 3 unseen = 13 total_requests)"`.

`Success` entries inside a `TooManyRequestErrors` batch are filtered out as this path is unreachable due to payload building/validation ensuring that no ok responses are included in the error-payload.

If Candid encoding fails, the delivery of the response is skipped and a timeout will eventually gracefully end the outstanding callback.

A new `TooManyRequestErrors(candid::Reserved)` variant is added to `FlexibleHttpGlobalError`, mirroring the existing `Timeout` and `ResponsesTooLarge` variants. Note that https://github.com/dfinity/ic/pull/9927 will rename `TooManyRequestErrors` to `TooManyRejects`.

Consensus metrics are extended to track `canister_http_flexible_errors_delivered` and to include flexible-error Candid failures in the existing `canister_http_flexible_candid_failures` counter.

### Known limitation

The `ResponsesTooLarge` mapping to `node_details` is a pragmatic fit but not ideal: the `FlexibleHttpNodeDetail.error` field is shaped for error information, yet we use it to convey both OK and reject shares (with `code` set to `"ok"` or `"reject"`). This overloads the field's intent. We can improve this in a future PR, e.g. by extending `FlexibleHttpNodeDetail` with a variant that cleanly distinguishes OK vs reject shares for the `ResponsesTooLarge` case.

### DIFF
--- a/rs/https_outcalls/consensus/src/payload_builder.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder.rs
@@ -1120,12 +1120,19 @@ fn flexible_error_into_consensus_response(
                 })
                 .collect();
 
-            let relevant_ok_sizes: Vec<_> = all_seen_shares
+            let mut ok_sizes: Vec<_> = all_seen_shares
                 .iter()
                 .filter(|s| !s.content.is_reject)
-                .take(min_known_ok_needed as usize)
-                .map(|share| share.content.content_size.to_string())
+                .map(|share| share.content.content_size)
                 .collect();
+            // Sort defensively, as validator doesn't enforce ordering on `all_seen_shares`
+            ok_sizes.sort_unstable();
+            let relevant_ok_sizes: Vec<_> = ok_sizes
+                .iter()
+                .take(min_known_ok_needed as usize)
+                .map(|size| size.to_string())
+                .collect();
+
             let message = format!(
                 "Responses too large: need at least {min_known_ok_needed} \
                  (= {min_responses} min_responses - {num_unseen} unseen) \
--- a/rs/https_outcalls/consensus/src/payload_builder.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder.rs
@@ -1134,10 +1134,10 @@ fn flexible_error_into_consensus_response(
                 .collect();
 
             let message = format!(
-                "Responses too large: need at least {min_known_ok_needed} \
-                 (= {min_responses} min_responses - {num_unseen} unseen) \
+                "Responses too large: need at least {min_responses} \
                  OK responses to fit within {MAX_CANISTER_HTTP_PAYLOAD_SIZE} bytes, \
-                 but even the smallest {min_known_ok_needed} of {num_ok} \
+                 but even the smallest {min_known_ok_needed} \
+                 (= {min_responses} min_responses - {num_unseen} unseen) of {num_ok} \
                  OK responses have sizes [{}] bytes \
                  ({num_ok} ok + {num_reject} reject + {num_unseen} unseen = {total_requests} total_requests)",
                 relevant_ok_sizes.join(", "),
--- a/rs/https_outcalls/consensus/src/payload_builder.rs
+++ b/rs/https_outcalls/consensus/src/payload_builder.rs
@@ -31,7 +31,9 @@ use ic_interfaces_registry::RegistryClient;
 use ic_interfaces_state_manager::StateReader;
 use ic_logger::{ReplicaLogger, warn};
 use ic_management_canister_types_private::{
-    CanisterHttpResponsePayload, FlexibleHttpRequestResult,
+    CanisterHttpResponsePayload, FlexibleHttpGlobalError, FlexibleHttpNodeDetail,
+    FlexibleHttpNodeError, FlexibleHttpRequestErr, FlexibleHttpRequestResult,
+    HttpRequestResourceReport,
 };
 use ic_metrics::MetricsRegistry;
 use ic_registry_client_helpers::subnet::SubnetRegistry;
@@ -77,6 +79,8 @@ pub struct CanisterHttpBatchStats {
     pub single_signature_responses: usize,
     pub flexible_ok_responses: usize,
     pub flexible_ok_responses_candid_failures: usize,
+    pub flexible_errors: usize,
+    pub flexible_errors_candid_failures: usize,
     pub payload_bytes: usize,
 }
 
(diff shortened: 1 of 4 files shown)


alin-at-dfinity (approved):
Management canister types changes LGTM.

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1018,6 +1033,124 @@ fn flexible_ok_responses_into_consensus_response(
...
+                .take(min_known_ok_needed as usize)
+                .map(|share| share.content.content_size.to_string())
+                .collect();
+            let message = format!(
+                "Responses too large: need at least {min_known_ok_needed} \
+                 (= {min_responses} min_responses - {num_unseen} unseen) \
+                 OK responses to fit within {MAX_CANISTER_HTTP_PAYLOAD_SIZE} bytes, \
+                 but even the smallest {min_known_ok_needed} of {num_ok} \
+                 OK responses have sizes [{}] bytes \
+                 ({num_ok} ok + {num_reject} reject + {num_unseen} unseen = {total_requests} total_requests)",
```
eichhorl: The phrase "need at least [...] min_responses - X [...] Ok responses to fit". Is a bit strange to me, because ultimately we of course need **all** min_response Ok responses to fit. It's just that we might already know that they won't before seeing all of them. Maybe something like this?
```suggestion
                "Responses too large: need at least {min_responses} \
                 OK responses to fit within {MAX_CANISTER_HTTP_PAYLOAD_SIZE} bytes, \
                 but even the smallest {min_known_ok_needed} (= {min_responses} min_responses - {num_unseen} unseen) \
                 of {num_ok} OK responses have sizes [{}] bytes \
                 ({num_ok} ok + {num_reject} reject + {num_unseen} unseen = {total_requests} total_requests)",
```

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1018,6 +1033,124 @@ fn flexible_ok_responses_into_consensus_response(
...
+                message,
+            )
+        }
+    };
+
+    let err = FlexibleHttpRequestErr {
+        global_error: Some(global_error),
+        node_details,
+        message,
+    };
```
eichhorl: We could also skip the intermediate triple:
```suggestion
            FlexibleHttpRequestErr {
                global_error: Some(FlexibleHttpGlobalError::ResponsesTooLarge(candid::Reserved)),
                node_details,
                message,
            }
        }
    };
```

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1018,6 +1033,124 @@ fn flexible_ok_responses_into_consensus_response(
...
+                    }
+                })
+                .collect();
+
+            let relevant_ok_sizes: Vec<_> = all_seen_shares
+                .iter()
+                .filter(|s| !s.content.is_reject)
+                .take(min_known_ok_needed as usize)
+                .map(|share| share.content.content_size.to_string())
+                .collect();
```
eichhorl: Below we claim that these are the smallest responses, does that mean we need to sort this vector?

rs/https_outcalls/consensus/src/payload_builder.rs
fspreiss: [Done](https://github.com/dfinity/ic/pull/9925/commits/b14643b76e5da8e3c7649a4ba53a2e0410fd22dc).

rs/https_outcalls/consensus/src/payload_builder.rs
fspreiss: Actually, yes (for the error message to be factually correct). This happened to work because the collection in the error is constructed from a _sorted_ collection in the payload builder, but a buggy/malicious payload builder could create an ordering that is not ascending and then the error message would be wrong. The code in the validator is correct (it sorts the collection itself) but it does not _enforce_ the ordering.

We could of course enforce the ordering in the validator, i.e. reject unsorted `all_seen_shares`, but this adds a new validation rule that could reject otherwise harmless blocks, probably overkill for a cosmetic error message.

I [fixed](https://github.com/dfinity/ic/pull/9925/commits/0cf7183fc37a9c04c077a934145418ae85e252e4) this now by sorting the collection before constructing the error message.

Let me know if you still prefer the new rule in the validator.

eichhorl: Thanks! Sorting the collection just before constructing the error message sounds good to me 👍

rs/https_outcalls/consensus/src/payload_builder.rs
fspreiss: Good point. I like your suggestion. [Fixed](https://github.com/dfinity/ic/pull/9925/commits/e323d472c5f1980d999ce3383d4686b74f147a88).

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -2796,6 +2797,185 @@ fn flexible_ok_responses_into_messages_decode_failure_is_skipped() {
...
+    let share_a = metadata_share_with_content_size(callback_id.get(), 0, 1_200_000);
+    let share_b = metadata_share_with_content_size(callback_id.get(), 1, 1_300_000);
+
+    let share_c = metadata_share_with_content_size(callback_id.get(), 2, 1_400_000);
+    let share_reject = reject_metadata_share(callback_id.get(), 3);
+
+    let payload = CanisterHttpPayload {
+        flexible_errors: vec![FlexibleCanisterHttpError::ResponsesTooLarge {
+            callback_id,
+            all_seen_shares: vec![share_a, share_b, share_c, share_reject],
```
eichhorl, line 2898: Maybe we can also adapt this test such that the share sizes aren't already sorted (to test the new sorting when converting into messages)

eichhorl (approved):

rs/https_outcalls/consensus/src/payload_builder/tests.rs
fspreiss: Good idea! [Done](https://github.com/dfinity/ic/pull/9925/commits/fe55ca6943a3fb453d7aae98a2c1510f93fbb66f).

fspreiss: And [here](https://github.com/dfinity/ic/pull/9925/commits/1080a35b158a9d5ce068b035f71f7250ed66067a).

PR #9926, opened 2026-04-17
feat: CON-1445 Helper functions to create `NiDkgConfig`s from request contexts

eichhorl:
Previously, configs for Remote DKG requests were only created and included as part of summary blocks. This implies that a request has to wait until the end of the current interval to even be started.

This PR introduces two new functions that will be used to start working on remote DKG requests as soon as they appear in the replicated state.

### `build_callback_id_config_map`
This function iterates over all remote DKG request contexts and attempts to derive the corresponding DKG configs. This is done by using the start height of the most recent summary block. In case of malformed inputs, errors are returned instead.

One special aspect is the use of the `initial_dkg_attempts` map, which is maintained by the summary payload. Entries of this map are incremented whenever there is still an outstanding request context in the replicated state, when the summary block is created. If the number of attempts exceeds the maximum, the new function will generate a timeout error instead. Furthermore, we will use this map to track remote DKG requests that have been completed as part of the previous interval. Target IDs for such requests will be inserted into this map with a count of 0. This ensures that they will not be attempted again in the subsequent interval.

### `merge_configs`
This function takes configs from the recent summary, and merges them with configs derived from the latest state using `build_callback_id_config_map`. The output represents the combined set of configs for currently ongoing DKGs.

Note that the new functions are still unused.

### DIFF
--- a/rs/consensus/dkg/src/lib.rs
+++ b/rs/consensus/dkg/src/lib.rs
@@ -32,6 +32,8 @@ use std::{
 pub mod dkg_key_manager;
 pub mod payload_builder;
 pub mod payload_validator;
+#[allow(dead_code)]
+pub(crate) mod remote;
 
 pub use crate::utils::get_vetkey_public_keys;
 
--- a/rs/consensus/dkg/src/payload_builder.rs
+++ b/rs/consensus/dkg/src/payload_builder.rs
@@ -798,7 +798,6 @@ fn process_reshare_chain_key_contexts(
         .metadata
         .subnet_call_context_manager
         .reshare_chain_key_contexts;
-    let get_reshare_transcript = |tag: &NiDkgTag| reshared_transcripts.get(tag).cloned();
 
     for (_callback_id, context) in contexts.iter() {
         // if we haven't reached the required registry version yet, skip this context
@@ -811,20 +810,33 @@ fn process_reshare_chain_key_contexts(
             continue;
         };
 
-        match create_remote_dkg_config_for_key_id(
-            key_id,
+        let dkg_id = NiDkgId {
             start_block_height,
-            this_subnet_id,
-            context.target_id,
+            dealer_subnet: this_subnet_id,
+            dkg_tag: NiDkgTag::HighThresholdForKey(key_id),
+            target_subnet: NiDkgTargetSubnet::Remote(context.target_id),
+        };
+        let Some(resharing_transcript) = reshared_transcripts.get(&dkg_id.dkg_tag).cloned() else {
+            let err = format!(
+                "Failed to find resharing transcript for a remote dkg for tag {:?}",
+                &dkg_id.dkg_tag
+            );
+            errors.push((dkg_id, err));
+            continue;
+        };
+
+        match create_remote_dkg_config(
+            dkg_id.clone(),
+            resharing_transcript.committee.get().clone(),
             context.nodes.clone(),
-            get_reshare_transcript,
             &context.registry_version,
+            Some(resharing_transcript),
         ) {
             Ok(config) => {
                 new_configs.push(vec![config]);
                 valid_target_ids.push(context.target_id);
             }
-            Err(err) => errors.push(*err),
+            Err(err) => errors.push((dkg_id, format!("{err:?}"))),
         }
     }
     (new_configs, errors, valid_target_ids)
@@ -1015,42 +1027,7 @@ pub(crate) fn create_low_high_remote_dkg_configs(
     }
 }
 
-pub(crate) fn create_remote_dkg_config_for_key_id(
-    key_id: NiDkgMasterPublicKeyId,
-    start_block_height: Height,
-    dealer_subnet: SubnetId,
-    target_id: NiDkgTargetId,
-    receivers: BTreeSet<NodeId>,
-    get_reshare_transcript: impl Fn(&NiDkgTag) -> Option<NiDkgTranscript>,
-    registry_version: &RegistryVersion,
-) -> Result<NiDkgConfig, Box<(NiDkgId, String)>> {
-    let dkg_id = NiDkgId {
-        start_block_height,
-        dealer_subnet,
-        dkg_tag: NiDkgTag::HighThresholdForKey(key_id),
-        target_subnet: NiDkgTargetSubnet::Remote(target_id),
-    };
-
-    // Find the resharing transcript corresponding to the remote dkg id
-    let Some(resharing_transcript) = get_reshare_transcript(&dkg_id.dkg_tag) else {
-        let err = format!(
-            "Failed to find resharing transcript for a remote dkg for tag {:?}",
-            &dkg_id.dkg_tag
-        );
-        return Err(Box::new((dkg_id, err)));
-    };
-
-    create_remote_dkg_config(
-        dkg_id.clone(),
-        resharing_transcript.committee.get().clone(),
-        receivers,
-        registry_version,
-        Some(resharing_transcript),
-    )
-    .map_err(|err| Box::new((dkg_id, format!("{err:?}"))))
-}
-
-fn create_remote_dkg_config(
+pub(crate) fn create_remote_dkg_config(
     dkg_id: NiDkgId,
     dealers: BTreeSet<NodeId>,
     receivers: BTreeSet<NodeId>,
@@ -1221,7 +1198,7 @@ mod tests {
     }
 
     #[test]
-    fn test_create_remote_dkg_config_for_key_id_dealers_match_reshared_transcript_committee() {
+    fn test_remote_dkg_config_dealers_match_reshared_transcript_committee() {
         let transcript_committee: Vec<_> = (0..4).map(node_test_id).collect();
         let receivers: BTreeSet<_> = (4..8).map(node_test_id).collect();
         let start_block_height = Height::from(500);
@@ -1242,18 +1219,18 @@ mod tests {
             777,
         );
 
-        let mut reshared_transcripts = BTreeMap::new();
-        reshared_transcripts.insert(tag, resharing_transcript.clone());
-
-        let get_reshare_transcript = |tag: &NiDkgTag| reshared_transcripts.get(tag).cloned();
-        let config = super::create_remote_dkg_config_for_key_id(
-            key_id,
+        let dkg_id = NiDkgId {
             start_block_height,
             dealer_subnet,
-            target_id,
+            dkg_tag: tag,
+            target_subnet: NiDkgTargetSubnet::Remote(target_id),
+        };
+        let config = super::create_remote_dkg_config(
+            dkg_id,
+            resharing_transcript.committee.get().clone(),
             receivers,
-            get_reshare_transcript,
             &registry_version,
+            Some(resharing_transcript.clone()),
         )
         .expect("expected remote DKG config for resharing");
 
--- a/rs/consensus/dkg/src/payload_builder.rs
+++ b/rs/consensus/dkg/src/payload_builder.rs
@@ -1008,7 +1008,7 @@ impl<'a> RemoteDkgContext<'a> {
                         start_block_height,
                         dealer_subnet,
                         dkg_tag,
-                        target_subnet: NiDkgTargetSubnet::Remote(self.target_id().clone()),
+                        target_subnet: NiDkgTargetSubnet::Remote(*self.target_id()),
                     },
                     REMOTE_DKG_REPEATED_FAILURE_ERROR.to_string(),
                 )
--- a/rs/consensus/dkg/src/payload_builder.rs
+++ b/rs/consensus/dkg/src/payload_builder.rs
@@ -1073,7 +1073,13 @@ fn create_remote_dkg_config(
 
 #[cfg(test)]
 mod tests {
-    use crate::{test_utils::create_dealing, tests::test_vet_key_config};
+    use crate::{
+        test_utils::{
+            create_dealing, local_dkg_id, make_test_config, remote_dkg_id,
+            remote_dkg_id_with_target,
+        },
+        tests::test_vet_key_config,
+    };
 
     use super::{super::test_utils::complement_state_manager_with_setup_initial_dkg_request, *};
     use ic_consensus_mocks::{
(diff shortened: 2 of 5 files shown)


rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -810,13 +810,14 @@ fn process_reshare_chain_key_contexts(
...
+        let get_reshare_transcript = |tag: &NiDkgTag| reshared_transcripts.get(tag).cloned();
         match create_remote_dkg_config_for_key_id(
             key_id,
             start_block_height,
             this_subnet_id,
             context.target_id,
             &context.nodes,
-            reshared_transcripts,
+            get_reshare_transcript,
             &context.registry_version,
```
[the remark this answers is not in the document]
eichhorl: Done

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -1031,20 +1032,21 @@ fn create_remote_dkg_config_for_key_id(
...
+    let dealers = resharing_transcript.committee.clone();
 
     create_remote_dkg_config(
         dkg_id.clone(),
-        resharing_transcript.committee.get(),
+        dealers.get(),
         receivers,
         registry_version,
-        Some(resharing_transcript.clone()),
+        Some(resharing_transcript),
```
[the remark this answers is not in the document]
eichhorl: Done

rs/consensus/dkg/src/lib.rs
```diff
@@ -32,6 +32,8 @@ use std::{
 pub mod dkg_key_manager;
 pub mod payload_builder;
 pub mod payload_validator;
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
eichhorl: The entire module is indeed dead code

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,646 @@
...
+            callback_id,
+            context.create_configs(
+                dkg_summary.height,
+                this_subnet_id,
+                registry_client,
+                get_reshare_transcript,
+                logger,
+            )?,
+        );
+    }
```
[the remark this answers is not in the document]
eichhorl: This is a good point, but only enforcing the `MAX_REMOTE_DKGS_PER_INTERVAL` here (similar to the old code) would still allow a potential unlimited number of requests to pile up, as request contexts in the replicated state are created.

Therefore, I'm proposing to remove `MAX_REMOTE_DKGS_PER_INTERVAL` in the new implementation, and instead enforce a limit on the maximum number of open call contexts directly in https://github.com/dfinity/ic/pull/9821. This is simpler and also more similar to the implementation of other management canister calls handled by consensus.

Note that another problem with limiting the number of requests is that it introduces a DOS vector. This should be fine for the creation of cloud engines (which will presumably be expensive enough to cover this resource). However, limiting the number of requests could also affect critical recovery proposals competing with cloud engine creations. For that reason, I'm proposing to handle the creation of cloud engines on a different subnet than the one used for critical create/recover/split subnet proposals in https://github.com/dfinity/ic/pull/9782.

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+}
+
+enum RemoteDkgContext<'a> {
+    SetupInitialDKG(&'a SetupInitialDkgContext),
+    ReshareChainKey(NiDkgReshareChainKeyContext<'a>),
+}
+
+impl<'a> RemoteDkgContext<'a> {
+    fn target_id(&self) -> &'a NiDkgTargetId {
```
pierugo-dfinity:
```suggestion
    fn target_id(&self) -> &NiDkgTargetId {
```

Is this lifetime annotation necessary?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -1014,13 +1015,13 @@ fn create_low_high_remote_dkg_configs(
...
-fn create_remote_dkg_config_for_key_id(
+pub(crate) fn create_remote_dkg_config_for_key_id(
     key_id: NiDkgMasterPublicKeyId,
     start_block_height: Height,
     dealer_subnet: SubnetId,
     target_id: NiDkgTargetId,
-    receivers: &BTreeSet<NodeId>,
-    reshared_transcripts: &BTreeMap<NiDkgTag, NiDkgTranscript>,
+    receivers: BTreeSet<NodeId>,
+    get_reshare_transcript: impl Fn(&NiDkgTag) -> Option<NiDkgTranscript>,
```
pierugo-dfinity: Feel free to ignore.

The `get_reshare_transcript` closure feels a little off to me and this function does not do much more than creating a `dkg_id` with the many passed arguments and checked if a reshared transcript is given before forwarding that to `create_remote_dkg_config`. The returned `Box<(NiDkgId, String)>` as error type is also quite cryptic. Couldn't all that work be done by the callers instead and delete this function?

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+}
+
+/// Builds a map from callback ids to config results for all remote DKG contexts
+/// found in the given replicated state.
+pub(crate) fn build_callback_id_config_map(
+    this_subnet_id: SubnetId,
+    registry_client: &dyn RegistryClient,
+    state: &ReplicatedState,
+    registry_version: RegistryVersion,
```
pierugo-dfinity, line 152:
```suggestion
    validation_context: &ValidationContext,
```

nit: Maybe it would be clearer to pass the validation context directly and use `validation_contex.registry_version` below?

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+            continue;
+        }
+
+        if let Some(&attempts) = remote_dkg_attempts.get(context.target_id()) {
+            if attempts == 0 {
+                // An attempt count of 0 means that the context has already been completed
+                // in the last interval. We skip the context to avoid handling it again in
+                // the current interval.
+                continue;
+            }
```
pierugo-dfinity, line 188: Alternatively, we could maybe change the type of `DkgSummary.initial_dkg_attempts` to be a `BTreeMap<NiDkgTargetId, NiDkgAttempts>` where `NiDkgAttempts` is an enum `Completed` or `InProgress(u32)` and then map in protobuf `Completed` to 0?

pierugo-dfinity: Actually, why do we still need to keep this entry in the map? Couldn't we remove it once completed, instead of setting to `Completed`/0?

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+            &no_op_logger(),
+        );
+        assert!(matches!(
+            result,
+            Err(DkgPayloadCreationError::FailedToGetSubnetMemberListFromRegistry(_))
+        ));
+    }
+
+    #[test]
+    fn test_build_callback_id_config_map_reshare_chain_key() {
```
fspreiss, line 480: Maybe also assert that a reshare context at `MAX_REMOTE_DKG_ATTEMPTS` produces a single `HighThresholdForKey(...)` timeout error.

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+            errors[0]
+                .1
+                .contains("Failed to find resharing transcript for a remote dkg")
+        );
+    }
+
+    #[test]
+    fn test_merge_configs_merges_successes_and_ignores_errors() {
+        let summary_only_id = local_dkg_id(NiDkgTag::LowThreshold);
+        let overlap_id = remote_dkg_id_with_target(NiDkgTag::HighThreshold, [11_u8; 32]);
```
fspreiss, line 624: Do I understand correctly that the configs behind the `overlap_id` are always identical, i.e., it doesn't matter which source won? If not, then it might be worth extending the test so that also the content of the config.

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+                );
+                continue;
+            }
+        }
+
+        // Create configs for the context and insert them into the map
+        callback_id_config_map.insert(
+            callback_id,
+            context.create_configs(
+                dkg_summary.height,
```
pierugo-dfinity: I understand that this needs to be the height of the last summary instead of the current height (as this function will be called in data blocks) for validation purposes. Are we 100% sure that we're not breaking anything by not being the current height? (also inside crypto?)

I'm saying this because the current documentation of `NiDkgId.start_block_height` currently says
> This field refers to the height of the block denoting the start of the computation of this DKG instance (resulting into a transcript later).

which would not be true after this change. If this is not needed for correctness, then we could also consider updating these docs.

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+    type Error = &'static str;
+
+    fn try_from(context: &'a ReshareChainKeyContext) -> Result<Self, Self::Error> {
+        let key_id = NiDkgMasterPublicKeyId::try_from(context.key_id.clone())?;
+        Ok(NiDkgReshareChainKeyContext { context, key_id })
+    }
+}
+
+enum RemoteDkgContext<'a> {
+    SetupInitialDKG(&'a SetupInitialDkgContext),
```
fspreiss, line 43: nit: Do we usually use `Dkg` (not all capitals)?

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+            .map_err(|err_box| vec![*err_box]),
+        };
+        Ok(config_result)
+    }
+}
+
+/// Builds a map from callback ids to config results for all remote DKG contexts
+/// found in the given replicated state.
+pub(crate) fn build_callback_id_config_map(
+    this_subnet_id: SubnetId,
```
fspreiss: Is the `this_` relevant here?

rs/consensus/dkg/src/remote.rs
```diff
@@ -0,0 +1,623 @@
...
+                    target_id,
+                    RegistryVersion::from(1),
+                    (60..64).map(node_test_id).collect(),
+                ),
+            );
+
+        let dkg_summary = test_dkg_summary(
+            Height::from(101),
+            BTreeMap::from([(tag.clone(), current_transcript.clone())]),
+            BTreeMap::from([(tag, transcript.clone())]),
```
fspreiss: Given that this is the "next_transcripts" param, I wonder if the test would be more clear (especially the assertions further below) if the `transcript` variable also contained `next_` or so.

fspreiss (approved):

rs/consensus/dkg/src/remote.rs
eichhorl: It is not: https://github.com/dfinity/ic/pull/9926/commits/30e02b709dbeb181572212b78a90fc35f8d9ab2d

rs/consensus/dkg/src/payload_builder.rs
eichhorl: That's fair, I kept the function because now it is called in two places. But inlining both of them seems fine, and we will hopefully get rid of the old location entirely soon. https://github.com/dfinity/ic/pull/9926/commits/564ac16f3b43c502b0ee48084ebc175ab7db1731

rs/consensus/dkg/src/remote.rs
eichhorl: This way, the function doesn't need to know what a validation context is and the semantics surrounding it, which I prefer slightly

rs/consensus/dkg/src/remote.rs
eichhorl: > `NiDkgAttempts`

It's a cool idea, although we would have to implement `Hash` manually for compatibility reasons, and somehow work around `NiDkgAttempts::InProgress(0)`. I can take a look once we remove the old code, where these types wouldn't make a lot of sense right now.

rs/consensus/dkg/src/remote.rs
eichhorl: Good point, but actually, `SetupInitialDkgContext` seems to be the exception, since we also use: `SubnetCallContext::SetupInitialDKG`, `Ic00Method::SetupInitialDKG`, `SetupInitialDKGArgs` and `SetupInitialDKGResponse`.

So instead, I would prefer to rename the context in a follow-up.

rs/consensus/dkg/src/remote.rs
eichhorl: Renamed to `dealer_subnet` for consistency: https://github.com/dfinity/ic/pull/9926/commits/fca3aaffadc9269f9b4b663b4cc65043da6af525

rs/consensus/dkg/src/remote.rs
eichhorl: Good point: https://github.com/dfinity/ic/pull/9926/commits/092da1c109e89a3a92bdf6c2029619d29e6c013d

rs/consensus/dkg/src/remote.rs
eichhorl: Yes, they should be identical. But soon, both sets will actually even be disjoint: The summary will only contain local configs, while the state will only contain remote configs.

rs/consensus/dkg/src/remote.rs
eichhorl: Good catch, added: https://github.com/dfinity/ic/pull/9926/commits/2dfa208c737940b72fa34190effe4d5c50c3254b

rs/consensus/dkg/src/remote.rs
eichhorl: > Actually, why do we still need to keep this entry in the map? Couldn't we remove it once completed, instead of setting to Completed/0?

If we remove it once completed, then in the following summary, we wouldn't be able to determine if the context is new (i.e. it didn't exist before the last summary, which is why it isn't part of the map), or if it has already been completed in the previous interval.

rs/consensus/dkg/src/remote.rs
eichhorl: Crypto doesn't access the `start_block_height` anywhere outside of tests, so it's only used as an identifier. 
Using the height of the last summary when creating the config means we backdate it, and pretend _as if_ the request context had already existed when the summary was created. It's just that we will only start including dealings a bit later. In that sense the `start_block_height` denotes the beginning of the DKG interval responsible for completing the request.

Alternatively we could initially use the height at which the context was inserted into the state. But if a transcript creation needs to be retired in the following interval, then we need to again set the start height to the current summary height, which adds some inconsistency.

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Nice looks great!

rs/consensus/dkg/src/remote.rs
pierugo-dfinity: Fair!

rs/consensus/dkg/src/remote.rs
pierugo-dfinity: > implement Hash manually for compatibility reasons, and somehow work around NiDkgAttempts::InProgress(0).

Hmmm yes I see

> I can take a look once we remove the old code, where these types wouldn't make a lot of sense right now.

Makes sense, sounds good

> determine if the context is new, or if it has already been completed in the previous interval

Right I see, thank you!

rs/consensus/dkg/src/remote.rs
pierugo-dfinity: Right thank you, then all good if crypto doesn't use it.

pierugo-dfinity (approved):

PR #9927, opened 2026-04-17
refactor(flexible-outcalls): CON-1714 rename TooManyRequestErrors to TooManyRejects

fspreiss:
When the `too_many_request_errors` global error was originally introduced, its name was chosen somewhat ad-hoc. Now that the error is fully wired up and user-facing, it's a good time to align its name with the terminology the rest of the IC uses.

The IC specification and the existing HTTPS outcalls API distinguish replies from **rejects**, and every reject carries one of the well-known [reject codes](https://internetcomputer.org/docs/references/ic-interface-spec#reject-codes) (`SYS_FATAL`, `SYS_TRANSIENT`, `CANISTER_REJECT`, …). Canister developers already recognize "reject" from `http_request` responses, inter-canister call rejects, and ingress-call rejects, so `TooManyRejects` is considerably more evocative than `TooManyRequestErrors`:
- it names the **IC-level concept** (a reject coming back from an HTTP adapter) rather than a generic "request error" that doesn't map to any spec term;
- it matches what actually appears in `node_details[].error.code`, which will carry a standard reject code string;

This error was now also introduced as a flexible_http_request_err.global_error in the public IC spec here: https://github.com/dfinity/portal/pull/5890/commits/defe26d020a5f34d00ed9bcd6739e72a490a2b6f


eichhorl (approved):

rs/types/management_canister_types/src/http.rs
```diff
@@ -435,8 +435,8 @@ pub enum FlexibleHttpGlobalError {
     OutOfCycles(candid::Reserved),
     #[serde(rename = "responses_too_large")]
     ResponsesTooLarge(candid::Reserved),
-    #[serde(rename = "too_many_request_errors")]
-    TooManyRequestErrors(candid::Reserved),
+    #[serde(rename = "too_many_rejects")]
+    TooManyRejects(candid::Reserved),
```
alin-at-dfinity, line 439: Totally clueless, but could the Candid change result in any sort of compatibility issue?

alin-at-dfinity (approved):

rs/types/management_canister_types/src/http.rs
mraszyk: The feature has not been rolled out yet so it is fine to do any breaking changes at this point.

fspreiss: Exactly as Martin said: as the feature hasn't been activated in production yet, as of now, we can still do breaking changes. If it was released, this would have been an issue, so thanks for the comment. Closing the comment.

PR #9928, opened 2026-04-17
chore: CON-1710 Reject missing subnet type when checking canister sig delegations

randombit:


mraszyk (approved):

rs/certification/src/lib.rs
```diff
@@ -162,7 +162,7 @@ pub fn verify_certified_data_with_cache_for_canister_sig(
     let (time, delegation_info) =
         verify_certified_data_internal(certificate, canister_id, root_pk, certified_data, true)?;
     if let Some(info) = delegation_info
-        && info.subnet_type.as_deref() == Some("cloud_engine")
+        && (info.subnet_type.is_none() || info.subnet_type.as_deref() == Some("cloud_engine"))
```
fspreiss: or `matches!(info.subnet_type.as_deref(), None | Some("cloud_engine"))`

fspreiss (approved):

fspreiss:
@andreacerulli, this is what you had in mind for this PR, right?

andreacerulli (approved):

rs/certification/src/lib.rs
randombit: Thanks that's better I was not so happy about that clippy

PR #9929, opened 2026-04-17
fix: deflake //rs/tests/consensus/upgrade:upgrade_downgrade_nns_subnet_test

basvandijk:
## Root cause

`//rs/tests/consensus/upgrade:upgrade_downgrade_nns_subnet_test` was flaking with:

```
Orchestrator of node <node_id> did not shut down gracefully
```

Inspection of the node journals for the 3 flaky runs in the past week
(2026-04-10, 2026-04-15, 2026-04-16) shows that every "failing" node
had actually logged `Orchestrator shut down gracefully` a few hundred
milliseconds *before* the test panicked. Example (2026-04-16 run):

```
10843:2026-04-16 14:46:23.843445 Orchestrator shut down gracefully
```
vs the test panic timestamp `14:46:24.229`.

The old `assert_orchestrator_stopped_gracefully` used a single long-lived
SSH session running `journalctl --follow --lines=1 | grep …`. As soon as
the orchestrator writes the shutdown message, the GuestOS reboots into
the new version. The SSH transport is torn down mid-pipeline before
`grep` flushes the matching line back to the driver. `execute_bash_script_from_session`
then returns `Err(...)` (non-zero exit / transport read error), and the
code used `.contains(...).unwrap_or_default()`, silently converting that
`Err` into `false`, causing a spurious assertion failure.

## Fix

Replace the fragile live-follow stream with a polling retry using
`retry_with_msg!`:

* Each attempt opens a fresh SSH session and searches the persisted
  journal (across all boots, non-follow mode) for
  `"Orchestrator shut down gracefully"`.
* On SSH/journal errors the retry loop simply tries again (bounded by a
  5 min overall timeout and `RETRY_BACKOFF`).
* Because we no longer rely on `--follow`, the search still succeeds
  after the node has rebooted into the new version (the message is
  present in a previous boot's journal).

## Verification

Ran 3 parallel executions:

```
  //rs/tests/consensus/upgrade:upgrade_downgrade_nns_subnet_test
```

Result:

```
//rs/tests/consensus/upgrade:upgrade_downgrade_nns_subnet_test   PASSED in 916.9s
  Stats over 3 runs: max = 916.9s, min = 780.7s, avg = 847.5s, dev = 55.6s
```

---

### DIFF
--- a/rs/tests/consensus/upgrade/common.rs
+++ b/rs/tests/consensus/upgrade/common.rs
@@ -427,7 +427,8 @@ fn find_latest_computed_root_hashes_from_logs(
 /// grep match can be flushed back over the channel. On reboot the SSH session is simply
 /// re-established and the message is found in the previous boot's journal.
 ///
-/// This function will never return if an upgrade is not scheduled.
+/// If no upgrade is scheduled, this keeps retrying until the overall timeout elapses and then
+/// panics.
 fn assert_orchestrator_stopped_gracefully(node: &IcNodeSnapshot) {
     const GRACEFUL_SHUTDOWN_MSG: &str = "Orchestrator shut down gracefully";
     const OVERALL_TIMEOUT: Duration = Duration::from_secs(5 * 60);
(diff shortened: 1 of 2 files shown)


basvandijk:
Closed in favour of https://github.com/dfinity/ic/pull/9934.

PR #9932, opened 2026-04-20
fix: relax expectations in system test for nodes catching up

mraszyk:
This PR relaxes the expectation in the system test `//rs/tests/message_routing:rejoin_test_long_rounds` so that the restarted node is only expected to catch up with the remaining nodes and become healthy until the _second_ next CUP (the condition used to be until the very next CUP). The motivation for this change is increased test flakiness (nodes not catching up until the very next CUP) which makes sense given the recent optimizations of the DSM overhead which lower the "advantage" (due to skipping state cloning and hashing) of a node that is catching up. Hence, it is expected that the node takes longer to catch up with the remaining nodes of the subnet.

The system test has been executed 59 times over the weekend and only failed 2 times (~3.4% failure rate).


alin-at-dfinity (commented):
LGTM, thank you for doing this.

alin-at-dfinity (approved):

PR #9933, opened 2026-04-20
chore: add batch_number into SystemMetadata

mraszyk:
This PR adds `batch_number` to `SystemMetadata`. The motivation for this change is to fix the execution environment (in a follow-up PR) to pass that `batch_number` instead of `state_reader.latest_state_height() + Height::from(1)` to `completed_execution_messages_tx` (channel notifying the synchronous update call HTTP handler) since using the height of the latest state stored in the state manager is imprecise (that state need not match the actual state on which the execution environment operates).

Once this PR is rolled out to all subnets, the critical error `mr_non_increasing_batch_number` could be further tightened to require that `batch_number` in `SystemMetadata` increases by exactly one (since batches must not be skipped). We cannot assert that stricter property now since `batch_number` is not recorded in existing `SystemMetadata` and thus will be implicitly initialized to `0` when deserializing the checkpoint.

### DIFF
--- a/rs/replicated_state/src/metadata_state.rs
+++ b/rs/replicated_state/src/metadata_state.rs
@@ -24,7 +24,7 @@ use ic_registry_routing_table::{
 use ic_registry_subnet_features::SubnetFeatures;
 use ic_registry_subnet_type::SubnetType;
 use ic_types::{
-    CountBytes, CryptoHashOfPartialState, NodeId, NumBytes, PrincipalId, SubnetId,
+    CountBytes, CryptoHashOfPartialState, Height, NodeId, NumBytes, PrincipalId, SubnetId,
     batch::BlockmakerMetrics,
     crypto::CryptoHash,
     ingress::{IngressState, IngressStatus},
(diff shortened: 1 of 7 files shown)


rs/replicated_state/src/metadata_state.rs
```diff
@@ -87,6 +87,11 @@ pub struct SystemMetadata {
     /// increasing).
     pub batch_time: Time,
 
+    /// The Consensus-determined height of this batch.
+    /// NOTE: this height is monotonically increasing (and not strictly
+    /// increasing).
```
mraszyk, line 92: This is a bit confusing (just like the same comment on `batch_time` above): the time and height are expected to be strictly increasing, but if they are not, only a critical error is raised (i.e., there's no crash and the time/height could indeed be not strictly increasing).

schneiderstefan:
This is something we intentionally did not do in the past. Only the state manager should cares what height a state has, and in places where we need both, e.g. in the return type of `get_latests_state`, we explicitly bundle them. This simplifies things like disaster recovery, where we take a state and simply move it forward to a new height. 

Instead, I suggest execution passes the [execution_round](https://sourcegraph.com/r/github.com/dfinity/ic@dd6e1e09e1f310c5205b8cb2f32469449c3db256/-/blob/rs/execution_environment/src/scheduler.rs?L1152) it gets passed in to the right places to use, but not store it in the checkpoint.

PR #9938, opened 2026-04-20
fix: notify sync update call HTTP handler with correct state height

mraszyk:
This PR sends `current_round` instead of `state_reader.latest_state_height() + Height::from(1)` to `completed_execution_messages_tx` (channel notifying the synchronous update call HTTP handler) since using the height of the latest state stored in the state manager is imprecise (that state need not match the actual state on which the execution environment operates).


schneiderstefan (approved):
Thank you!

PR #9941, opened 2026-04-20
refactor: merge canister and subnet read state endpoint code

mraszyk:
This PR merges the code for _canister_ (`/api/{v2,v3}/canister/.../read_state`) and _subnet_ (`/api/{v2,v3}/subnet/.../read_state`) read state HTTP endpoints. The motivation for this change is avoiding code duplication which would get even more problematic when allowing `/request_status` paths via the subnet endpoint.

In terms of behavioral changes, the PR introduces ingress signature validation for subnet read state requests (which does make sense even though at the moment no user information is returned by subnet read state requests).


alin-at-dfinity (approved):

rs/http_endpoints/public/src/lib.rs
```diff
@@ -360,30 +359,21 @@ pub fn start_server(
...
-            state_reader.clone(),
-            nns_subnet_id,
-            version,
-        )
-        .with_health_status(health_status.clone())
-        .build_router()
-    };
-    let subnet_read_state_v2_router = subnet_read_state_router(read_state::subnet::Version::V2);
-    let subnet_read_state_v3_router = subnet_read_state_router(read_state::subnet::Version::V3);
+    let subnet_read_state_router = |version| read_state_router(version, read_state::Target::Subnet);
```
eichhorl, line 374: Why a closure for the `subnet_read_state_router`, but not for the `canister_read_state_router`?

rs/http_endpoints/public/src/read_state.rs
```diff
@@ -1,23 +1,310 @@
...
+    } else {
+        RegistryRootOfTrustProvider::new(registry_client, registry_version)
+    };
+    let request_c = request.clone();
+
+    let response = tokio::task::spawn_blocking(move || {
+        let targets = match validator.validate_request(
+            &request_c,
+            time_source.get_relative_time(),
+            &root_of_trust_provider,
```
eichhorl, line 253: Should we add a test for an invalid authenticated request to one of the subnet-only endpoints, which previously would have been accepted, but is now rejected?

eichhorl (approved):
I reviewed the diffs of `rs/http_endpoints/public/src/read_state.rs` <> `rs/http_endpoints/public/src/read_state/canister.rs` and `rs/http_endpoints/public/src/read_state.rs` <> `rs/http_endpoints/public/src/read_state/subnet.rs` separately, and they look good to me.

As usual, Claude removed a bunch of comments in the process which I wouldn't mind keeping. Is there a way to add them back?

rs/http_endpoints/public/src/lib.rs
mraszyk: Claude... I'll make this consistent.

mraszyk:
> I reviewed the diffs of `rs/http_endpoints/public/src/read_state.rs` <> `rs/http_endpoints/public/src/read_state/canister.rs` and `rs/http_endpoints/public/src/read_state.rs` <> `rs/http_endpoints/public/src/read_state/subnet.rs` separately, and they look good to me.

Thanks a lot! I also reviewed those diffs manually and they looked good to me (after minor manual fixes).

> 
> As usual, Claude removed a bunch of comments in the process which I wouldn't mind keeping. Is there a way to add them back?

I noticed that, too, but I wasn't sure if the comments are worthwhile. I'll look into that and restore them.

rs/http_endpoints/public/src/read_state.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/9941/commits/f24f5672c38a5b104ff2f19a25bce685dfd1583b

rs/http_endpoints/public/src/lib.rs
mraszyk: Fixed: https://github.com/dfinity/ic/pull/9941/commits/24bb354b143a41e53f191121707833100821e61b

PR #9942, opened 2026-04-20
fix: deflake //rs/tests/networking:canister_http_time_out_test_head_nns

basvandijk:
Increase the IPv4 polling timeout inside `start_httpbin_on_uvm` from 120s to 300s and emit a clear error on timeout. 120s has been observed to be insufficient on slow/busy Farm hosts, causing the test to flake with a silent exit 1.

### DIFF
--- a/rs/tests/networking/canister_http/canister_http.rs
+++ b/rs/tests/networking/canister_http/canister_http.rs
@@ -189,11 +189,15 @@ pub fn start_httpbin_on_uvm(env: &TestEnv) {
 
         if [[ "$ipv4" == "" ]] && ip link show dev enp2s0 >/dev/null; then
             count=0
+            # Wait up to 5 minutes for DHCP to assign a globally scoped IPv4 address
+            # on enp2s0. 120s has been observed to be insufficient on slow/busy Farm
+            # hosts, causing this test to flake.
             until ipv4=$(ip -j address show dev enp2s0 \
                         | jq -r -e \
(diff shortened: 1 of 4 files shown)


cgundy (approved):

rs/tests/networking/canister_http/canister_http.rs
```diff
@@ -170,36 +170,45 @@ pub fn start_httpbin_on_uvm(env: &TestEnv) {
...
+    // Retrieve the UVM's IPv4 address directly from the guest if it's configured
+    // with an IPv4 interface. This polls the guest until DHCP assigns a globally
+    // scoped address.
+    let has_ipv4_iface = deployed_universal_vm
+        .block_on_bash_script_from_session(
+            &session,
+            "if ip link show dev enp2s0 >/dev/null 2>&1; then echo yes; else echo no; fi",
+        )
+        .map(|s| s.trim() == "yes")
+        .unwrap_or(false);
```
eichhorl: This will silently default to `false` if `block_on_bash_script_from_session` fails or returns anything other than `yes`. Is that intentional?

eichhorl (approved):

rs/tests/networking/canister_http/canister_http.rs
basvandijk: Good point. It's actually better to just use `.unwrap()` such that the test panics when `block_on_bash_script_from_session` fails.

PR #9946, opened 2026-04-20
test(DSM): Speed up compilation bench tests

adambratschikaye:
Speed up the test for our compilation benches by having the benches which don't measure compilation itself take compilation artifacts from existing results.


alin-at-dfinity (approved):
🎉

Thank you, LGTM.

PR #9948, opened 2026-04-20
feat(DSM): add canister_log_resize_duration_seconds metric

maksymar:
## Summary

- Add `canister_log_resize_duration_seconds` histogram to observe how long `log_memory_limit` resizes take on the canister log memory store.
- Samples are emitted only for real resizes: no-op calls (same limit) are filtered out, and first-time allocation during canister creation is not observed so the distribution stays focused on user-triggered resizes.
- Uses the shared `duration_histogram` helper for consistency with existing execution-environment duration metrics.
- Includes a regression test covering both the observed and no-op paths.


rs/execution_environment/src/canister_manager.rs
```diff
@@ -631,10 +638,19 @@ impl CanisterManager {
...
+            // resize would actually do work (`would_resize`); avoids no-op
+            // samples and keeps the histogram scoped to real resizes.
+            // Dropping the timer records the elapsed duration into the
+            // histogram, so drop it right after the resize to avoid timing
+            // any unrelated work that might be added below.
+            let resize_timer = metrics
+                .filter(|_| log_memory_store.would_resize(limit))
+                .map(|m| m.canister_log_resize_duration.start_timer());
+            log_memory_store.resize(limit, self.fd_factory.clone());
+            drop(resize_timer);
```
alin-at-dfinity: You could also just put the resize in a block by its own.

```suggestion
            (
                let _maybe_timer = metrics
                    .filter(|_| log_memory_store.would_resize(limit))
                    .map(|m| m.canister_log_resize_duration.start_timer());
                log_memory_store.resize(limit, self.fd_factory.clone());
            }
```

alin-at-dfinity (approved):

rs/execution_environment/src/canister_manager.rs
maksymar: done, thank you!

PR #9950, opened 2026-04-20
feat(ic-admin): add --canister-cycles-cost-schedule option to propose-to-create-subnet

r-birkner:
## Summary
- Expose `canister_cycles_cost_schedule` as a CLI flag on `ic-admin propose-to-create-subnet` (previously hard-coded to `Normal`), so cloud engine subnets — which require `Free` — can be proposed through ic-admin.
- Defaults to `Normal` to preserve existing behavior; accepts `normal` or `free`.
- Adds `Display` / `EnumString` derives to `CanisterCyclesCostSchedule` so clap can parse the flag via `FromStr`, matching the `SubnetType` pattern.

### DIFF
--- a/rs/registry/admin/bin/create_subnet.rs
+++ b/rs/registry/admin/bin/create_subnet.rs
@@ -166,6 +166,11 @@ pub(crate) struct ProposeToCreateSubnetCmd {
     #[clap(long)]
     pub max_number_of_canisters: Option<u64>,
 
+    /// The canister cycles cost schedule for this subnet.
+    /// Can be "normal" (default) or "free" (used by cloud engine subnets).
+    #[clap(long)]
+    pub canister_cycles_cost_schedule: Option<do_create_subnet::CanisterCyclesCostSchedule>,
+
     /// The features that are enabled and disabled on the subnet.
     #[clap(long)]
     pub features: Option<SubnetFeatures>,
@@ -272,6 +277,8 @@ impl ProposeToCreateSubnetCmd {
                 .get_or_insert(ReplicaVersion::default());
             self.max_number_of_canisters.get_or_insert(0);
             self.features.get_or_insert(SubnetFeatures::default());
+            self.canister_cycles_cost_schedule
+                .get_or_insert(do_create_subnet::CanisterCyclesCostSchedule::Normal);
         }
     }
 
@@ -329,7 +336,8 @@ impl ProposeToCreateSubnetCmd {
             max_number_of_canisters: self.max_number_of_canisters.unwrap_or_default(),
             chain_key_config,
             canister_cycles_cost_schedule: Some(
-                do_create_subnet::CanisterCyclesCostSchedule::Normal,
+                self.canister_cycles_cost_schedule
+                    .expect("canister_cycles_cost_schedule must be specified."),
             ),
             subnet_admins: Some(self.subnet_admins.clone()),
             resource_limits: self.resource_limits,
@@ -411,6 +419,7 @@ mod tests {
             features: None,
             subnet_admins: vec![],
             resource_limits: None,
+            canister_cycles_cost_schedule: None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


rs/registry/admin/bin/create_subnet.rs
```diff
@@ -166,6 +166,11 @@ pub(crate) struct ProposeToCreateSubnetCmd {
     #[clap(long)]
     pub max_number_of_canisters: Option<u64>,
 
+    /// The canister cycles cost schedule for this subnet.
+    /// Can be "normal" (default) or "free" (used by cloud engine subnets).
+    #[clap(long)]
+    pub canister_cycles_cost_schedule: Option<do_create_subnet::CanisterCyclesCostSchedule>,
```
daniel-wong-dfinity-org, line 172: Don't we usually do something like

```
use do_create_subnet::CanisterCyclesCostSchedule;
```

at the top of the file? Ignore this remark if there is another type with the same name in this context (because create/module prefix is the best and typical way to disambiguate such cases).

rs/registry/admin/bin/create_subnet.rs
```diff
@@ -272,6 +277,8 @@ impl ProposeToCreateSubnetCmd {
                 .get_or_insert(ReplicaVersion::default());
             self.max_number_of_canisters.get_or_insert(0);
             self.features.get_or_insert(SubnetFeatures::default());
+            self.canister_cycles_cost_schedule
+                .get_or_insert(do_create_subnet::CanisterCyclesCostSchedule::Normal);
```
daniel-wong-dfinity-org, line 281: No(t asking for any) change.

It is weird to me that get is called, and then the result is not used. Ofc, this is an existing pattern (i.e. predates this PR); hence, "No change".

rs/registry/canister/src/mutations/do_create_subnet.rs
```diff
@@ -503,14 +504,28 @@ impl TryFrom<KeyConfigRequest> for KeyConfigRequestInternal {
...
+)]
 pub enum CanisterCyclesCostSchedule {
     /// Use the cost schedule associate with the subnet's type.
     #[default]
+    #[strum(serialize = "normal")]
     Normal,
 
     /// This is used by rented subnets. This is because rented subnets get paid
     /// for in a different way.
+    #[strum(serialize = "free")]
```
daniel-wong-dfinity-org, line 528: Do as you see fit.

Why `serialize`?

A problem with this is, you are signing (our future selves) up for manual labor. That is an inevitable result of ad hoc, irregular. Unless there is some overriding reason, go with regular; eschew ad hoc.

# Soap Box/Plea/Beating a Dead Horse

Another problem with ad hoc-ing is that people cannot rely on the standard behavior, and thus, when they try to predict behavior, they are pretty likely to get it wrong. Wrong predictions = bugs. Principle of least surprise.

Is the issue in this tiny case life-threatening in isolation? Ofc not. But you can say similar things about every case. This is how code gradually slides into spaghetti; people are not willing to hold the line on consistency. Well, maybe in the AI promised land, this won't matter. I suspect that even then, Ai will be more productive when everything is regular.

Think about how annoying it is to learn to spell in English. The reason is this: English is not regular (or should I say, it is violently IRREGULAR). For example: https://www.youtube.com/shorts/3ipFdRfFvK4 . Don't create "mice". Do not force people to learn that "mice" = many mouse when everything else is just "slap an -s on it".

daniel-wong-dfinity-org (approved):

rs/registry/admin/bin/create_subnet.rs
r-birkner: Yes, agree: imported `CanisterCyclesCostSchedule` at the top and dropped the module prefix wherever it was called.

rs/registry/canister/src/mutations/do_create_subnet.rs
r-birkner: You convinced me. I removed the serialize and people just have to use "Normal" and "Free". Actually, they only have to use "Free" as "Normal" is anyway the default behavior.

PR #9952, opened 2026-04-20
feat(DSM): add canister_log_retention_seconds metric

maksymar:
## Summary

- Add `canister_log_retention_seconds` histogram to `SchedulerMetrics` — wall-clock span between the oldest and newest records currently in a canister's log buffer.
- Activity-gated observation at round finalization; cost scales with activity, not canister count.
- Works on both log stores (branched on `LOG_MEMORY_STORE_FEATURE_ENABLED`), giving directly-comparable before/after-flip data on the same histogram. On the new path, `first_timestamp` is cached in a transient `Option<u64>` on `LogMemoryStore`, populated eagerly from every mutation site — no header change, no checkpoint migration.
- Buckets: `decimal_buckets_with_zero(1, 6)` → 19 buckets, `[0, 10 s .. ~58 d]`.

## Stacked on

Depends on #9948. Base is `maksym/canister-log-metrics`; retarget to `master` once #9948 lands.


rs/config/src/execution_environment.rs
```diff
@@ -12,7 +12,7 @@ const GIB: u64 = 1024 * MIB;
 const TIB: u64 = 1024 * GIB;
 
 // TODO(DSM-105): remove after the feature is enabled by default.
-pub const LOG_MEMORY_STORE_FEATURE_ENABLED: bool = false;
+pub const LOG_MEMORY_STORE_FEATURE_ENABLED: bool = true; // TODO: do not submit! only for CI testing.
```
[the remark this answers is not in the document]
alin-at-dfinity: Is this addressed on your local branch? Otherwise I'd leave the comment open to protect against accidental merging.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -148,13 +159,20 @@ impl LogMemoryStore {
             self.save_ring_buffer(ring_buffer);
         } else {
             self.header_cache = OnceLock::new();
+            self.first_timestamp_cache = OnceLock::new();
```
alin-at-dfinity: Just FYI: in other parts of the replicated state we use `debug_asserts` to check that stats and other transient state are consistent with the underlying data. So e.g. at the end of this and other methods that touch `first_timestamp_cache`, `header_cache` or the ring buffer you would have something like `debug_assert!(self.stats_ok())`.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -71,6 +72,15 @@ pub struct LogMemoryStore {
...
     header_cache: OnceLock<Option<Header>>,
 
+    /// Caches the timestamp of the oldest live record. Populated alongside
+    /// `header_cache` inside `save_ring_buffer` (while the first-record page
+    /// is still hot from the preceding `append_log`), lazy-initialised after
+    /// checkpoint reload, and invalidated in `clear` / `deallocate`.
+    ///
+    /// Avoids a cold page read at observation time for the retention metric.
+    #[validate_eq(Ignore)]
+    first_timestamp_cache: OnceLock<Option<u64>>,
```
alin-at-dfinity: I only have a vague understanding of `LogMemoryStore` and friends, but doesn't the fact that it's a ring buffer mean that the first record (and its timestamp) keep changing over time? If so, the `OnceLock` type sends the wrong message.

And regardless of the above, do we really need the value to be wrapped in a mutex? Virtually all of the replicated state relies on single-threaded access. Definitely single-threaded mutations.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
maksymar: good point, thank you, added `debug_assert!(self.stats_ok())` and corresponding `fn stats_ok()` which is only does expensive work in debug asserts.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
maksymar: good catch, thank you. I did it by example from the header, but it's not needed for `first_timestamp_cache` because it's only updated in write paths, whereas header cache can be updated on read paths, so it needs `OnceLock`.

rs/config/src/execution_environment.rs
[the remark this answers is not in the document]
maksymar: CI was green with the feature enabled, so I disabled in now.
image

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -193,6 +206,30 @@ impl LogMemoryStore {
...
+            && *cached != actual_header
+        {
+            return false;
+        }
+        // `first_timestamp_cache` is kept eagerly in sync, so must always match.
+        let actual_first_ts = ring_buffer
+            .as_ref()
+            .zip(actual_header.as_ref())
+            .and_then(|(rb, h)| rb.first_timestamp(h));
+        self.first_timestamp_cache == actual_first_ts
```
alin-at-dfinity: Shouldn't this be a conditional `return false`, like above?

rs/execution_environment/src/scheduler.rs
```diff
@@ -1511,6 +1524,11 @@ impl Scheduler for SchedulerImpl {
                         .canister_log_delta_memory_usage
                         .observe(size as f64);
                 }
+                if let Some(retention) = retention {
+                    self.metrics
+                        .canister_log_retention
+                        .observe(retention.as_secs_f64());
+                }
```
alin-at-dfinity, line 1531: [Hmm... GitHub ate my homework. I wrote a similar comment before here, but I must have failed to submit it.]

Should this really be a scheduler metric? I.e. only observing retention for mutated logs will heavily bias the observations toward active canisters. E.g. if there's a canister with a heartbeat producing log messages every round, it will likely dominate the metric, making it not very useful.

I would recommend making this (and `canister_log_delta_memory_usage`) `ReplicatedStateMetrics` fields and observing all canisters' logs, not just the mutated ones. If you are concerned about performance, you can do it e.g. once every 10 rounds. And the reason for the existence of `ReplicatedStateMetrics` is to move the `observe()` call to a background thread and off the critical path (something I hope to do, eventually), so as long as it's not longer than a DSM round, it doesn't matter much just how (in)efficient it is.

alin-at-dfinity (approved):

rs/execution_environment/src/scheduler.rs
maksymar: fair point on the bias. I'd like to keep this scoped to the scheduler for now to ship quickly. the observation is biased toward active cansiters, but I'd argue that's the data we are actually interested in.
this can be revisited later if we needed better fairness on dashboards.

I'll resolve this question to continue with the merge. thank you!

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
maksymar: done, thanks.

PR #9953, opened 2026-04-20
feat(guestos): build patched kernel with folio_split() race fix

basvandijk:
Builds a custom `linux-hwe-6.17` kernel in the GuestOS base image that includes the upstream fix for a pagecache folio_split() / folio_try_get() race that has been observed on IC nodes.

## Background

- Upstream commit: [`577a1f495fd78d8fb61b67ac3d3b595b01f6fcb0`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=577a1f495fd78d8fb61b67ac3d3b595b01f6fcb0) ("mm/huge_memory: fix a folio_split() race condition with folio_try_get()") by Zi Yan, merged into mainline on 2026-03-04; first tagged release v7.0-rc4.
- Backport to [`linux-6.18.y`](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.18.y&id=08b2b65c63bb26dbb2a4e2adc2ce96e2929b8b60): already applied (2026-03-25).
- Ubuntu `linux-hwe-6.17` on noble (24.04): **does NOT yet contain the fix** — verified against both `6.17.0-22.22~24.04.1` (currently installed) and the proposed `6.17.0-24.24~24.04.1` via a grep of their upload diff. The fix does not appear in Ubuntu's stable-queue either.
- Reproducer: https://github.com/dfinity/thp-madv-remove-test.

The bug was introduced by commit `00527733d0dc` in kernel 6.14 (folio_split() feature), so it affects the HWE kernel we currently ship.

## Changes

* New `kernel-build` stage in `ic-os/guestos/context/Dockerfile.base`:
  * `apt-get source linux-hwe-6.17`, applies every `*.patch` file under `kernel-patches/` in lexicographic order.
  * Bumps the Debian changelog with a `+dfinity` local suffix so the resulting kernel is identifiable via `uname -r`.
  * Builds only the amd64 `generic` flavor `.deb`s with `skipdbg=true skipretpoline=true` to cut build time.
* Main stage installs the locally built `linux-image-unsigned-*-generic`, `linux-modules-*-generic` and `linux-modules-extra-*-generic` `.deb`s instead of pulling `linux-image-virtual-hwe-24.04` from apt, keeping the kernel and extra modules in sync by construction.
* New directory `ic-os/guestos/context/kernel-patches/` with:
  * `0001-mm-huge_memory-fix-folio_split-race-condition.patch` — the patch Zi Yan confirmed applies cleanly to 6.17.13.
  * `README.md` documenting conventions and how to drop custom patches when they land upstream.

## Removing this customization later

Once Ubuntu ships the fix (either as a new `linux-hwe-6.17` SRU or via a future `linux-hwe-6.18` / `-7.0` HWE track on noble), delete the patch file(s) under `kernel-patches/`. With no patches, the kernel-build stage still runs and produces a repackaged stock kernel; if we prefer to remove the custom build entirely at that point, revert this commit.

## Testing

- CI will exercise the kernel-build stage as part of the base image build.
- Local smoke tests (optional):
  * Build only the kernel-build stage: `docker build --target kernel-build -f ic-os/guestos/context/Dockerfile.base ic-os/guestos/context`.
  * Inspect `/debs` inside the resulting image for three `+dfinity1` `.deb`s.
  * Boot a GuestOS dev image and check `uname -r` ends in `+dfinity1`.
- Functional test against the reporter's reproducer at https://github.com/dfinity/thp-madv-remove-test.

## Notes for reviewers

- The kernel build adds ~20–40 min to the `deploy-guest-os-baseimg` job.
- The kernel-build stage requires network access during the bazel action (for `apt-get source` / `build-dep`). If the sandbox blocks egress, the rule will need appropriate tags.
- GuestOS signs kernels as part of the IC image build (see `ic-os/guestos/docs/DiskLayout.adoc`), so we deliberately consume the `linux-image-unsigned-*` flavor.


Bownairo:
The bug should be gone with #10022 and #10025. We are working to update to the new ubuntu and kernel with #9989.

PR #3044, opened 2024-12-09
fix(registry): Ensure subnet features can be changed for subnets without any features

aterga:
This PR fixes the validation of the SEV (AMD Secure Encrypted Virtualization) feature that prevented change-subnet proposals that attempted to set a (non-SEV-related) feature for the first time (i.e., when the subnet doesn't yet have any features).

PR #9956, opened 2026-04-21
revert(registry): restore strict sev_enabled check in update_subnet

r-birkner:
Partially reverts #9038:

- **Reverts** the relaxation of `validate_update_sev_feature` in `do_update_subnet` — `sev_enabled` can once again only be set at subnet creation; any update that sets `features.sev_enabled` now panics, as it did before.
- **Keeps** the invariant added in the same PR requiring SEV-enabled subnets to consist only of nodes with a chip ID (`check_sev_subnet_invariants` in `rs/registry/canister/src/invariants/subnet.rs` and its tests).


rs/registry/canister/src/mutations/do_update_subnet.rs
```diff
@@ -143,27 +143,24 @@ impl Registry {
...
-        let Some(update_features) = payload.features else {
+        let Some(features) = payload.features else {
             return;
         };
 
-        // panic if SEV is enable and the update tries to disable it
-        let attempting_to_disable =
-            subnet_features.sev_enabled == Some(true) && update_features.sev_enabled == Some(false);
-        if attempting_to_disable {
+        if let Some(sev_enabled) = features.sev_enabled {
```
jasonz-dfinity: It doesn't seem to be right, because (surprisingly to me at least) the features are REPLACED when updating the subnet, and `null` seems to be treated the same as `Some(false)`. In other words, a proposal can still disable SEV by setting the features to `Some(SubnetFeatures { ..., sev_enabled: None})`.

I think the problem was introduced by #3044, and I believe the right solution is to compare against the existing sev_enabled like pre-#3044, while still avoiding the unnecessary panic.

Other solutions came to mind but I don't think they are the right choice:

- do a "merge" of the features from the payload to the existing record, it changes semantics and could be confusing when the features has a mixture of `bool` and `opt bool`
- separate the types of SubnetFeatures between record and payload which introduces API changes and potentially data change

r-birkner: Thanks @jasonz-dfinity for the thorough review. I was just reverting the previous change assuming everything was good before. I fixed it now. Please take another look. Thank you very much!

jasonz-dfinity: Looks good now. Thanks!

jasonz-dfinity (approved):

PR #9966, opened 2026-04-21
feat(governance): Relax 8 year gang requirements.

daniel-wong-dfinity-org:
In particular, >= 8 * 365 days dissolve delay + aging since < March 30 (midnight UTC) (and not already a member of the eight year gang).

This is because there were neurons with very very very nearly 8 year dissolve delay.

# References

Closes [NNS1-4349].

[Discussion with Thomas, Jason, and Björn][dis].

[NNS1-4349]: https://dfinity.atlassian.net/browse/NNS1-4349
[dis]: https://dfinity.slack.com/archives/C01EWN833KN/p1776847314352599

### DIFF
--- a/rs/nns/governance/src/neuron_store.rs
+++ b/rs/nns/governance/src/neuron_store.rs
@@ -1,11 +1,14 @@
 use crate::{
     CURRENT_PRUNE_FOLLOWING_FULL_CYCLE_START_TIMESTAMP_SECONDS, Clock, IcClock,
-    governance::{LOG_PREFIX, TimeWarp},
+    governance::{
+        LOG_PREFIX, MAX_DISSOLVE_DELAY_SECONDS_PRE_MISSION_70,
+        RELAXED_EIGHT_YEAR_GANG_MIN_DISSOLVE_DELAY_SECONDS, TimeWarp,
+    },
     neuron::Neuron,
     neurons_fund::neurons_fund_neuron::pick_most_important_hotkeys,
     pb::v1::{
         GovernanceError, NeuronDissolveStateSnapshot, Topic, VotingPowerEconomics,
-        governance_error::ErrorType,
+        governance_error::ErrorType, neuron_dissolve_state_snapshot,
     },
     storage::{
         neuron_indexes::CorruptedNeuronIndexes, neurons::NeuronSections,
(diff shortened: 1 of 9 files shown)


rs/nns/governance/src/storage/neurons.rs
```diff
@@ -783,6 +787,42 @@ where
         }
     }
 
+    pub fn set_relaxed_eight_year_gang_bonus_base_e8s_or_panic(
+        &mut self,
+        neuron_ids_with_sufficiently_large_dissolve_delay: Vec<u64>,
+    ) {
+        for neuron_id in neuron_ids_with_sufficiently_large_dissolve_delay {
+            let neuron_id = NeuronId { id: neuron_id };
+            self.with_main_part_mut(neuron_id, |abridged_neuron| {
```
tmu0, line 796: Is it guaranteed that the neuron id from the snapshot still exists or could it be split/merged/etc in the meantime?

rs/nns/governance/src/neuron_store.rs
```diff
@@ -783,6 +786,44 @@ impl NeuronStore {
         });
     }
 
+    pub fn set_relaxed_eight_year_gang_bonus_base_e8s_or_panic(
+        &mut self,
+        pre_clamp_states: &HashMap<u64, NeuronDissolveStateSnapshot>,
```
tmu0, line 791: Did clamping, i.e., snapshot creation, happen at the same time when the `eight_year_gang_bonus_base_e8s` was set?

rs/nns/governance/src/storage/neurons.rs
```diff
@@ -783,6 +787,42 @@ where
...
+                        Skipping...",
+                        LOG_PREFIX, neuron_id.id,
+                    );
+                    return;
+                }
+
+                // This is to avoid newly staked ICP from receiving the eight year
+                // gang bonus.
+                if abridged_neuron.aging_since_timestamp_seconds
+                    > RELAXED_EIGHT_YEAR_GANG_MAX_AGING_SINCE_TIMESTAMP_SECONDS
```
tmu0, line 812: I don't understand how this check should avoid that newly staked ICP receive the eight year gang bonus: IIUC if X new ICP are added to a neuron which already has Y ICP, the age will be reduced by the factor Y/(X+Y). However, if the previous age was large and/or X is small, the resulting new aging timestamp could still be before `RELAXED_EIGHT_YEAR_GANG_MAX_AGING_SINCE_TIMESTAMP_SECONDS`.

rs/nns/governance/src/neuron_store.rs
```diff
@@ -783,6 +786,44 @@ impl NeuronStore {
...
+
+                // Skip neurons dissolve delay that was too small.
+                if d < RELAXED_EIGHT_YEAR_GANG_MIN_DISSOLVE_DELAY_SECONDS {
+                    return None;
+                }
+
+                // Skip neurons that are already in the eight year gang.
+                if d == MAX_DISSOLVE_DELAY_SECONDS_PRE_MISSION_70 {
+                    return None;
+                }
```
jasonz-dfinity, line 811: This

- could be problematic, as the neuron could have increased their dissolve delay between April 9th and April 20th, and I don't see a reason punishing them
- seems unnecessary, as we already guard against `eight_year_gang_bonus_base_e8s` later

jasonz-dfinity (approved):

rs/nns/governance/src/storage/neurons.rs
daniel-wong-dfinity-org: We do not delete neurons. So, for example, after A gets merged into B, A still exists; it is just empty.

rs/nns/governance/src/neuron_store.rs
daniel-wong-dfinity-org: No, clamping dissolve delay took place week(s) later.

rs/nns/governance/src/storage/neurons.rs
daniel-wong-dfinity-org: Yes, that is possible. This is not an absolute guarantee, but at least if you had a 1 ICP neuron, and you suddenly added a gazillion ICP in order to take advantage of this second round of induction, you would not be able to do THAT.

Ideally, what we want is to know the "net stake" (i.e. cached - fees + staked maturity) that the neuron HAD, but we do not have that (on chain). This is the next best thing.

rs/nns/governance/src/neuron_store.rs
daniel-wong-dfinity-org: > already guard

Yes, this is belt and suspenders.

> punish

They are already getting the bonus. I am pretty sure that nobody will complain.

One of the main ideas here is "minimally invasive surgery is best surgery".

PR #9977, opened 2026-04-22
feat: enable listing canisters via query call with subnet ID in the URL

mraszyk:
This PR adds a new public HTTP endpoint `/api/v3/subnet/<effective_subnet_id>/query` on the replica that supports call to the `list_canisters` endpoint of the management canister (`aaaaa-aa`). It is specified in this [PR](https://github.com/dfinity/portal/pull/6224).

Note. PocketIC support will be added in a separate PR.

### DIFF
--- a/rs/http_endpoints/public/src/query.rs
+++ b/rs/http_endpoints/public/src/query.rs
@@ -234,7 +234,7 @@ pub(crate) async fn query(
             if canister_id != CanisterId::ic_00() && canister_id != effective_canister_id {
                 let status = StatusCode::BAD_REQUEST;
                 let text = format!(
-                    "Specified CanisterId {canister_id} does not match effective canister id in URL {effective_canister_id}"
+                    "Specified canister ID {canister_id} does not match effective canister ID in URL {effective_canister_id}"
                 );
                 return (status, text).into_response();
             }
@@ -244,7 +244,7 @@ pub(crate) async fn query(
             if effective_subnet_id != subnet_id {
                 let status = StatusCode::BAD_REQUEST;
                 let text = format!(
-                    "Specified SubnetId {effective_subnet_id} does not match the subnet id of this node {subnet_id}"
+                    "Specified subnet ID {effective_subnet_id} does not match the subnet ID of this node {subnet_id}"
                 );
                 return (status, text).into_response();
             }
--- a/rs/http_endpoints/public/src/query.rs
+++ b/rs/http_endpoints/public/src/query.rs
@@ -1,4 +1,4 @@
-//! Module that deals with requests to /api/v2/canister/.../query and /api/v3/subnet/.../query
+//! Module that deals with requests to /api/v{2,3}/canister/.../query and /api/v3/subnet/.../query
 
 use crate::{
     ReplicaHealthStatus,
--- a/rs/http_endpoints/public/src/query.rs
+++ b/rs/http_endpoints/public/src/query.rs
@@ -1,4 +1,4 @@
-//! Module that deals with requests to /api/v2/canister/.../query
+//! Module that deals with requests to /api/v2/canister/.../query and /api/v4/subnet/.../query
 
 use crate::{
     ReplicaHealthStatus,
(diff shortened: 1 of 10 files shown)


alin-at-dfinity (approved):

rs/tests/networking/http_endpoints_public_spec_test.rs
```diff
@@ -784,6 +784,42 @@ fn get_canister_test_ids(snapshot: &TopologySnapshot) -> (CanisterId, [CanisterI
     )
 }
 
+fn query_calls_subnet_v4(env: TestEnv) {
```
eichhorl: Should it be `v3`?
```suggestion
fn query_calls_subnet_v3(env: TestEnv) {
```

rs/http_endpoints/public/src/query.rs
```diff
@@ -217,12 +226,41 @@ pub(crate) async fn query(
...
+
+    // Validate effective destination.
+    match version {
+        Version::V2 | Version::V3 => {
+            let effective_canister_id = CanisterId::unchecked_from_principal(id);
+            if canister_id != CanisterId::ic_00() && canister_id != effective_canister_id {
+                let status = StatusCode::BAD_REQUEST;
+                let text = format!(
+                    "Specified CanisterId {canister_id} does not match effective canister id in URL {effective_canister_id}"
+                );
```
eichhorl, line 238: If `canister_id == CanisterId::ic_00() && canister_id != effective_canister_id` then the canister ID in the URL would not match the one specified by the request, but we would not return this error. Is that intentional?

rs/http_endpoints/public/src/query.rs
```diff
@@ -263,8 +301,12 @@ pub(crate) async fn query(
         Version::V2 => {
             nns_delegation_reader.get_delegation_with_metadata(CanisterRangesFilter::Flat)
         }
-        Version::V3 => nns_delegation_reader
-            .get_delegation_with_metadata(CanisterRangesFilter::Tree(effective_canister_id)),
+        Version::V3 => nns_delegation_reader.get_delegation_with_metadata(
+            CanisterRangesFilter::Tree(CanisterId::unchecked_from_principal(id)),
```
eichhorl, line 305: Is it really safe to use `CanisterId::unchecked_from_principal(id)` here? In the `canister_id == CanisterId::ic_00()` above, we didn't actually check that `id` is indeed a valid canister ID

mraszyk: At least, this is what we used to do before this change:
```
● Here's how axum extraction works for CanisterId:                                                                                                                                            
                                                                                                                                                                                              
  Mechanism: Path extractor + FromStr                                                                                                                                                         
                                                                                                                                                                                              
  There's no custom FromRequest/FromRequestParts implementation. Axum's built-in Path extractor automatically calls FromStr on the path parameter type, and CanisterId implements it.         
                                                                                                                                                                                              
  FromStr implementation                                                                                                                                                                      
                                                                                                                                                                                              
  rs/types/base_types/src/canister_id.rs:260-268:                                                                                                                                             
  impl std::str::FromStr for CanisterId {                                                                                                                                                     
      type Err = CanisterIdError;                                                                                                                                                             
      fn from_str(input: &str) -> Result<Self, Self::Err> {                                                                                                                                   
          let principal_id = PrincipalId::from_str(input)                                                                                                                                     
              .map_err(CanisterIdError::PrincipalIdParseError)?;
          Ok(CanisterId::unchecked_from_principal(principal_id))                                                                                                                              
      }                                                                                                                                                                                       
  }    
```

rs/http_endpoints/public/src/query.rs
mraszyk: Yes, because management canister calls have `canister_id` equal to `ic_00` (mgmt canister), but the effective canister ID is the actual canister ID of the (application) canister that the request targets. One could cover this case by extracting that actual canister ID from the request body, but this would require candid parsing in the HTTP handler and thus was omitted.

eichhorl: I see, thanks

rs/http_endpoints/public/src/query.rs
```diff
@@ -217,12 +226,41 @@ pub(crate) async fn query(
...
+                );
+                return (status, text).into_response();
+            }
+        }
+        Version::SubnetV3 => {
+            let effective_subnet_id = SubnetId::from(id);
+            if effective_subnet_id != subnet_id {
+                let status = StatusCode::BAD_REQUEST;
+                let text = format!(
+                    "Specified SubnetId {effective_subnet_id} does not match the subnet id of this node {subnet_id}"
```
eichhorl: It's a bit strange that we use `SubnetId` in the first half, but `subnet id` in the second

eichhorl (approved):

rs/http_endpoints/public/src/query.rs
mraszyk: Fixed: https://github.com/dfinity/ic/pull/9977/changes/c619600ef7f4402c889e451925b261a54e6d519f

PR #9981, opened 2026-04-22
feat(icrc-ledger-types): add ICRC-123 block schema validators

bogwar:
Phase 1 of ICRC-123/153 implementation. Adds ICRC-123 block schema validators for 4 block types with permissive and strict validation modes.

### DIFF
--- a/packages/icrc-ledger-types/src/icrc123/schema.rs
+++ b/packages/icrc-ledger-types/src/icrc123/schema.rs
@@ -1,0 +1,440 @@
+use std::borrow::Cow;
+use std::sync::Arc;
+
+use crate::icrc::generic_value::Value;
+use crate::icrc::generic_value_predicate::{
+    ItemRequirement, ValuePredicate, ValuePredicateFailures, and, is, is_account, is_blob,
+    is_equal_to, is_map, is_more_or_equal_to, is_principal, is_text, item, len, or,
+};
+
+// Block type identifiers (ICRC-123)
+pub const BTYPE_123_FREEZE_ACCOUNT: &str = "123freezeaccount";
+pub const BTYPE_123_UNFREEZE_ACCOUNT: &str = "123unfreezeaccount";
+pub const BTYPE_123_FREEZE_PRINCIPAL: &str = "123freezeprincipal";
+pub const BTYPE_123_UNFREEZE_PRINCIPAL: &str = "123unfreezeprincipal";
+
+// Method discriminators (ICRC-153)
+pub const MTHD_153_FREEZE_ACCOUNT: &str = "153freeze_account";
+pub const MTHD_153_UNFREEZE_ACCOUNT: &str = "153unfreeze_account";
+pub const MTHD_153_FREEZE_PRINCIPAL: &str = "153freeze_principal";
+pub const MTHD_153_UNFREEZE_PRINCIPAL: &str = "153unfreeze_principal";
+
+/// Build a block-level predicate for an ICRC-123 account freeze/unfreeze block.
+///
+/// * `btype` – block type identifier
+/// * `strict` – when `true`, `caller`, `mthd`, and `ts` (in tx) are required (ICRC-153);
+///   when `false` they are optional (ICRC-123).
+fn account_block_validator(btype: &'static str, strict: bool) -> ValuePredicate {
+    use ItemRequirement::*;
+    let strict_req = if strict { Required } else { Optional };
+
+    let is_timestamp = is_more_or_equal_to(0);
+    let is_parent_hash = and(vec![is_blob(), len(is_equal_to(32))]);
+    let is_created_at_time = is_more_or_equal_to(0);
+    let is_transaction = and(vec![
+        is_map(),
+        item("account", Required, is_account()),
+        item("caller", strict_req.clone(), is_principal()),
+        item("mthd", strict_req.clone(), is_text()),
+        item("ts", strict_req, is_created_at_time),
+        item("reason", Optional, is_text()),
+        item("policy_ref", Optional, is_text()),
+    ]);
+    and(vec![
+        is_map(),
+        item("phash", Optional, is_parent_hash),
+        item("btype", Required, is(Value::text(btype))),
+        item("ts", Required, is_timestamp),
+        item("tx", Required, is_transaction),
+    ])
+}
+
+/// Build a block-level predicate for an ICRC-123 principal freeze/unfreeze block.
+///
+/// * `btype` – block type identifier
+/// * `strict` – when `true`, `caller`, `mthd`, and `ts` (in tx) are required (ICRC-153);
+///   when `false` they are optional (ICRC-123).
+fn principal_block_validator(btype: &'static str, strict: bool) -> ValuePredicate {
+    use ItemRequirement::*;
+    let strict_req = if strict { Required } else { Optional };
+
+    let is_timestamp = is_more_or_equal_to(0);
+    let is_parent_hash = and(vec![is_blob(), len(is_equal_to(32))]);
+    let is_created_at_time = is_more_or_equal_to(0);
+    let is_transaction = and(vec![
+        is_map(),
+        item("principal", Required, is_principal()),
+        item("caller", strict_req.clone(), is_principal()),
+        item("mthd", strict_req.clone(), is_text()),
+        item("ts", strict_req, is_created_at_time),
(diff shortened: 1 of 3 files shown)


packages/icrc-ledger-types/src/icrc123/schema.rs
```diff
@@ -0,0 +1,442 @@
...
+pub fn validate(block: &Value) -> Result<(), ValuePredicateFailures> {
+    or(vec![
+        Arc::new(|v| validate_freeze_account(&v)),
+        Arc::new(|v| validate_unfreeze_account(&v)),
+        Arc::new(|v| validate_freeze_principal(&v)),
+        Arc::new(|v| validate_unfreeze_principal(&v)),
+    ])(Cow::Borrowed(block))
+}
+
+#[cfg(test)]
```
gregorydemay, line 159: nit: there are several school of thoughts on this but I would always separate test code from productive logic (read move the tests to a separate file). For example as a benefit, no need to rebuild the library if only test code changes.

packages/icrc-ledger-types/src/icrc123/schema.rs
```diff
@@ -0,0 +1,442 @@
...
+pub const BTYPE_123_FREEZE_PRINCIPAL: &str = "123freezeprincipal";
+pub const BTYPE_123_UNFREEZE_PRINCIPAL: &str = "123unfreezeprincipal";
+
+// Method discriminators (ICRC-153)
+pub const MTHD_153_FREEZE_ACCOUNT: &str = "153freeze_account";
+pub const MTHD_153_UNFREEZE_ACCOUNT: &str = "153unfreeze_account";
+pub const MTHD_153_FREEZE_PRINCIPAL: &str = "153freeze_principal";
+pub const MTHD_153_UNFREEZE_PRINCIPAL: &str = "153unfreeze_principal";
+
+use crate::icrc::generic_value_predicate::is_account;
```
gregorydemay: nit: it's weird that it sits there after the const declaration, I would put it with the other imports L.5

packages/icrc-ledger-types/src/icrc123/schema.rs
```diff
@@ -0,0 +1,442 @@
...
+
+/// Build a block-level predicate for an ICRC-123 account freeze/unfreeze block.
+///
+/// * `btype` – block type identifier
+/// * `strict` – when `true`, `caller`, `mthd`, and `ts` (in tx) are required (ICRC-153);
+///   when `false` they are optional (ICRC-123).
+fn account_block_validator(btype: &'static str, strict: bool) -> ValuePredicate {
```
gregorydemay, line 27: nit: this seems to be almost the same as `principal_block_validator`. If you want to keep the expressivity (2 different method names) I would refactor so that they delegate to a common internal method.

packages/icrc-ledger-types/src/icrc123/schema.rs
```diff
@@ -0,0 +1,442 @@
...
+/// Permissive ICRC-123 validation: `caller`, `mthd`, and `ts` (in tx) are optional.
+/// Use [`validate_153_freeze_account`] for strict ICRC-153 validation.
+///
+/// Block structure:
+/// ```text
+/// { btype: "123freezeaccount", phash?: Blob(32), ts: Nat,
+///   tx: { account: Account, caller?: Principal, mthd?: Text, ts?: Nat,
+///         reason?: Text, policy_ref?: Text } }
+/// ```
+pub fn validate_freeze_account(block: &Value) -> Result<(), ValuePredicateFailures> {
```
gregorydemay: understanding question: the ICRC number appears in certain methods, like `validate_153_freeze_account` but not all, like this one (`validate_freeze_account`). Is there a reason?

packages/icrc-ledger-types/src/icrc123/schema.rs
```diff
@@ -0,0 +1,442 @@
...
+    or(vec![
+        Arc::new(|v| validate_freeze_account(&v)),
+        Arc::new(|v| validate_unfreeze_account(&v)),
+        Arc::new(|v| validate_freeze_principal(&v)),
+        Arc::new(|v| validate_unfreeze_principal(&v)),
+    ])(Cow::Borrowed(block))
+}
+
+#[cfg(test)]
+mod tests {
```
gregorydemay, line 160: Understanding question: what's the purpose of those validate methods? How critical are they? Example: do you want to have a proptest to ensure they don't panic?

gregorydemay (approved):
Thanks @bogwar I mostly have some understanding questions but otherwise code LGTM

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: The existing pattern in `icrc-ledger-types` is inline tests — ICRC-122's `schema.rs` does the same, and ICRC-107's `schema.rs` has no tests at all. No module in this crate uses separate test files. I'd prefer to keep it consistent with the existing codebase for now, but happy to refactor all of them if we want to establish a new convention.

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: Fixed — merged `is_account` into the existing import group at line 5.

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: The two functions differ only in the required tx field (`account` with `is_account()` vs `principal` with `is_principal()`). I could extract a common helper that takes the field name and predicate as parameters, but it would add indirection for very little deduplication — the outer block structure (phash, btype, ts, tx) is the same but the tx field predicates are different. Happy to refactor if you feel strongly.

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: Good point — renamed all permissive validators to include the standard number: `validate_123_freeze_account`, `validate_123_unfreeze_account`, etc. The convention is now: `validate_123_*` = permissive (ICRC-123 block standard), `validate_153_*` = strict (ICRC-153 endpoint standard). Created DEFI-2791 to apply the same rename to the ICRC-122 module (`validate_mint` → `validate_122_mint`).

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: These validators are used by test infrastructure to verify that blocks produced by the ledger conform to the ICRC-123/153 schema. They return `Result` (don't panic), so they're safe to call on any input. A proptest could add value for ensuring the validators don't accidentally reject valid blocks — something like: generate a valid block via BlockBuilder, assert the validator accepts it. We have this coverage indirectly through the unit tests that construct blocks and validate them, but a proptest with random valid blocks would be more thorough. Happy to add one if you'd like.

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: On reflection, I'd rather keep them separate. The deduplication gain is minimal (one line differs) and a shared helper with field name + predicate parameters would be harder to read. Keeping them explicit makes it clear at a glance what each block type expects.

packages/icrc-ledger-types/src/icrc123/schema.rs
bogwar: I don't think a proptest adds enough value here. The validators are simple predicate compositions and the existing hand-crafted tests cover both valid and invalid cases for each block type. A proptest generating random valid blocks would mostly exercise the same code paths. The risk of a false rejection that existing tests miss is low.

packages/icrc-ledger-types/src/icrc123/schema.rs
gregorydemay: makes sense to stay consistent. I leave it up to you whether you want to refactor, it's really not critical.

packages/icrc-ledger-types/src/icrc123/schema.rs
gregorydemay: > A proptest generating random valid blocks would mostly exercise the same code paths.

I was suggesting proptest more as a kind of fuzzing: ensure that methods don't panic, if it's a critical requirement. From your description (used by test infrastructure) this doesn't seem to be the case.

PR #9982, opened 2026-04-22
feat: enable asynchronous canister creation with subnet ID in the URL

mraszyk:
This PR enables to retrieve the status of an (asynchronous) update call via the (existing) public HTTP endpoint `/api/v3/subnet/<effective_subnet_id>/read_state` on the replica. It is specified in this [PR](https://github.com/dfinity/portal/pull/6224).

In more detail, this PR is needed if the synchronous update call endpoint `/api/v4/subnet/.../call` returns 202 and thus the user needs to process the update call (to create canister as nothing else is allowed on `/api/v4/subnet/.../call`) asynchronously (by polling `/api/v3/subnet/.../read_state` for the `/request_status/...`).

### DIFF
--- a/rs/http_endpoints/public/src/read_state.rs
+++ b/rs/http_endpoints/public/src/read_state.rs
@@ -1334,6 +1334,10 @@ mod test {
         )
         .expect_err("Should fail");
         assert_eq!(err.status, StatusCode::FORBIDDEN);
+        assert_eq!(
+            err.message,
+            "The user tries to access Request ID not signed by the caller."
+        );
     }
 
     #[rstest]
@@ -1367,5 +1371,10 @@ mod test {
         )
         .expect_err("Should fail");
         assert_eq!(err.status, StatusCode::FORBIDDEN);
+        assert_eq!(
+            err.message,
+            "The user tries to access request IDs for canisters \
+                not belonging to sender delegation targets."
+        );
     }
 }
--- a/rs/http_endpoints/public/src/read_state.rs
+++ b/rs/http_endpoints/public/src/read_state.rs
@@ -633,7 +633,11 @@ mod test {
         SUBNET_0, SUBNET_1, canister_test_id, subnet_test_id, user_test_id,
     };
     use ic_types::{
-        SubnetId, batch::RawQueryStats, consensus::certification::Certification, time::UNIX_EPOCH,
+        NumBytes, SubnetId,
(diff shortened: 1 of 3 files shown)


rs/tests/networking/read_state_test.rs
eichhorl: Should we now also enable these tests for the subnet endpoint?

eichhorl (commented):
The PR title doesn't seem to match the changes

rs/tests/networking/read_state_test.rs
mraszyk: Good catch! https://github.com/dfinity/ic/pull/9982/changes/a64891bf4eddef895d4454e9005c47c71fd177bf

mraszyk:
> The PR title doesn't seem to match the changes

What I meant to say is that this PR is needed if the synchronous update call endpoint /api/v4/subnet/.../call returns 202 and thus the user needs to process the update call (to create canister as nothing else is allowed on /api/v4/subnet/.../call) asynchronously (by polling /api/v3/subnet/.../read_state for the /request_status).

rs/http_endpoints/public/src/read_state.rs
```diff
@@ -1270,4 +1283,89 @@ mod test {
...
+            &[Path::new(vec![
+                Label::from("request_status"),
+                message_id.as_bytes().to_vec().into(),
+            ])],
+            &CanisterIdSet::try_from_iter(vec![other_canister]).unwrap(),
+            effective_principal_id,
+            NNS_SUBNET_ID,
+        )
+        .expect_err("Should fail");
+        assert_eq!(err.status, StatusCode::FORBIDDEN);
```
eichhorl, line 1373: We could also check the error message here and above

rs/tests/networking/read_state_test.rs
```diff
@@ -13,7 +13,7 @@ Success::
     . public key and canister ranges for all subnets
     . public keys of nodes on the subnet
     . no public keys of nodes on other subnets
-. Malformed status requests are rejected by /api/{v2,v3}/canister/.../read_state
+. Malformed status requests are rejected by /api/{v2,v3}/canister/.../read_state and /api/v3/subnet/.../read_state
```
eichhorl: Maybe we could update the documentation below as well

eichhorl (approved):

alin-at-dfinity (approved):

PR #9985, opened 2026-04-22
feat: [DSM-103] Improve scheduler fairness

alin-at-dfinity:
This is a collection of scheduler improvements backported from the active canister scheduler dev branch, with the goal of having the scheduler efficiency proptest pass (with reduced success thresholds). And, by applying the proptest and all other `RoundSchedule` tests onto `master`, hugely reducing the size of that change.

 * Charge heap delta rate limited canisters, so that they actually skip execution rounds instead of being merely delayed while accumulating priority.
 * Charge immediately for the first round of every long execution (which was scheduled as a new execution).
 * Apply an exponential decay to AP outside the `[-2000, 500]` range (`-2000` because of 20 max DTS rounds, the `500` is somewhat arbitrary, but seems to work well). Due to the interaction between long and short executions, runaway priorities are inevitable. This provides a soft bound for runaway AP, while still preserving relative priorities to some extent. It also makes priority resets unnecessary (to be removed later).
 * Fully distribute all positive free compute, even if it means exceeding 100 priority per canister (by distributing it equally to all canisters).
 * Allow long executions to use all scheduler cores when there are no new executions.
 * Fully segregate long and new executions across cores, to prevent inversion of priority when lower priority long executions get a slice executed.


rs/execution_environment/src/scheduler/round_schedule/tests.rs
```diff
@@ -0,0 +1,2131 @@
...
+/// Fixture for testing `RoundSchedule` in isolation: a `RoundSchedule` and a
+/// `ReplicatedState`, with helpers to manage canisters and inspect canister metrics
+/// and priority.
+struct RoundScheduleFixture {
+    round_schedule: RoundSchedule,
+    state: ReplicatedState,
+    current_round: ExecutionRound,
+    next_canister_id: u64,
+    // Keeps MetricsRegistry alive for self.metrics.
+    _registry: MetricsRegistry,
```
schneiderstefan: Do we need this?

schneiderstefan (approved):

rs/execution_environment/src/scheduler/round_schedule/tests.rs
alin-at-dfinity: Apparently not needed, dropped it.

Not sure why I thought it would be. Probably had some tests that queried it for the metrics before using the `SchedulerMetrics` fields directly instead. And then just kept it around.

PR #9986, opened 2026-04-22
feat(governance): Batch proposal

daniel-wong-dfinity-org:
All sub-actions must have same topic.

Topic is inherited from sub-action(s).

# Implementation

To make this work, I (well, Claude, really) had to split `perform_action` into two separate steps:

1. `try_perform_action` (new), followed by
2. `set_proposal_execution_status` (existing)

This way, `perform_batch` can call `try_perform_action` for each sub-action, and NOT set the proposal's final execution status.

Furthermore, we also had to make sure that the implementation of every action (called by `try_perform_action`) did not try to set the proposal's execution status.

In particular, significant rework was needed for `ExecuteNnsFunction`. Such rework includes getting rid of `execute_nns_function` in the `Environment` trait. It was just a shitty version of `call_canister_method` anyway; therefore, deleting `execute_nns_function` is really not as bad as it sounds.

# References

Closes [NNS1-4296].

[NNS1-4296]: https://dfinity.atlassian.net/browse/NNS1-4296

### DIFF
--- a/rs/nns/governance/src/canister_state.rs
+++ b/rs/nns/governance/src/canister_state.rs
@@ -1,7 +1,5 @@
-use crate::{
-    governance::{Environment, Governance, HeapGrowthPotential, RandomnessGenerator, RngError},
-    pb::v1::{GovernanceError, governance_error::ErrorType},
-    proposals::execute_nns_function::ValidExecuteNnsFunction,
+use crate::governance::{
+    Environment, Governance, HeapGrowthPotential, RandomnessGenerator, RngError,
 };
 
 use async_trait::async_trait;
@@ -11,7 +9,6 @@ use ic_nervous_system_canisters::ledger::IcpLedgerCanister;
 use ic_nervous_system_runtime::CdkRuntime;
 use ic_nervous_system_runtime::Runtime;
 use ic_nervous_system_time_helpers::now_seconds;
-use ic_nns_common::types::ProposalId;
 use ic_nns_constants::LEDGER_CANISTER_ID;
 use rand::{RngCore, SeedableRng};
 use rand_chacha::ChaCha20Rng;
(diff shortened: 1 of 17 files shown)


rs/nns/governance/src/canister_state.rs
```diff
@@ -183,60 +180,6 @@ impl Environment for CanisterEnv {
         *self.time_warp.write().unwrap() = new_time_warp;
     }
 
-    fn execute_nns_function(
```
jasonz-dfinity, line 186: I very much like this change, but I strongly suggest splitting it into a smaller PR.

rs/nns/governance/src/governance.rs
```diff
@@ -4200,148 +4148,161 @@ impl Governance {
     }
 
     async fn perform_action(&mut self, pid: u64, action: ValidProposalAction) {
```
jasonz-dfinity, line 4150: We have an existing issue that could be made worse with the batch execution - the execution of a proposal is within the same method execution as everything else (most notably, changing the execution status). That means, if the instruction limit is hit for a synchronous proposal (e.g. manage neuron) and if such proposal execution triggered through deadline, then it will keep trying. I considered this case when disabling "making proposal as part of manage neuron proposal" (https://github.com/dfinity/ic/commit/cbebf5d2d74bb97b616a8204df43d2ddfc3560aa).

I haven't gone through all the proposal types, but I suppose for most types it's not an issue. However, I can think of a case - a ManageNeuron proposals with 100 RegisterVote, and each one triggers a lot of following (sadly, for the same reason, the proposal execution triggered by RegisterVote causing an absolute majority, also executes within the RegisterVote message, and vote cascading can be expensive).

Therefore, my suggestion is to change ic_cdk::spawn=>ic_cdk::spawn_migatory (if I understood that correctly, I'm not 100% sure about the behavior of different spawn methods)

rs/nns/governance/src/proposals/mod.rs
```diff
@@ -151,6 +155,68 @@ impl TryFrom<Option<Action>> for ValidProposalAction {
     }
 }
 
+fn try_into_valid_batch(batch: Batch) -> Result<Vec<ValidProposalAction>, GovernanceError> {
```
jasonz-dfinity, line 158: I would assume the "all actions have the same topic" validation happens somewhere below, but I can't find it.

PR #9987, opened 2026-04-22
perf(crypto): CRP-2550 Enable Rayon in NIDKG again

randombit:
Originally added in #9007 reverted in #9968


randombit:
The general flavor of the test flakes seem to be a number of test binaries want to concurrently spawn up 8-20 IC instances running in the same process (probably already at the limits of flaking IMO) then when we add a few more threads to each IC instance everything falls over. 33155e3 attempts to reduce this by sharing the NIDKG pool but likely as not that's simply causing starvation where one of the test ICs claims the entire pool and then some other IC times out waiting for progress to be made. Likely fixing this (and/or being able to make use of any parallelism in crypto) will require breaking apart (or serializing) these big tests.

PR #9989, opened 2026-04-22
Update IC-OS images to ubuntu 26.04

Bownairo:


Bownairo:
Performance testing done with this branch, in https://dfinity.atlassian.net/browse/NODE-1954. Will break up into component changes to merge.

PR #9990, opened 2026-04-23
chore: clean up orphaned StopCanisterCall

mraszyk:
This PR cleans up `StopCanisterCall`s from `SubnetCallContextManager` that have no corresponding `StopCanisterContext` in the target canister's stop contexts. These can arise from a bug fixed in commit 52e7b89 where a `StopCanisterCall` was pushed but never removed when the `stop_canister` call failed, e.g., because the target canister was already stopped or the sender was not a controller.

### DIFF
--- a/rs/execution_environment/src/execution_environment.rs
+++ b/rs/execution_environment/src/execution_environment.rs
@@ -4574,7 +4574,8 @@ impl ExecutionEnvironment {
     /// Removes `StopCanisterCall`s from `SubnetCallContextManager` that have no
     /// corresponding `StopCanisterContext` in the target canister's stop contexts.
     /// These can arise from a bug fixed in commit 52e7b89 where a `StopCanisterCall`
-    /// was pushed but never removed when the target canister was already stopped.
+    /// was pushed but never removed when the `stop_canister` call failed, e.g., because
+    /// the target canister was already stopped or the sender was not a controller.
     fn cleanup_orphaned_stop_canister_calls(
         &self,
         state: &mut ReplicatedState,
--- a/rs/execution_environment/src/execution_environment.rs
+++ b/rs/execution_environment/src/execution_environment.rs
@@ -4580,25 +4580,25 @@ impl ExecutionEnvironment {
         state: &mut ReplicatedState,
         canister_states: &BTreeMap<CanisterId, Arc<CanisterState>>,
     ) {
+        // Build a set of all call IDs that have a matching StopCanisterContext.
+        let call_ids_with_context: BTreeSet<StopCanisterCallId> = canister_states
+            .values()
+            .flat_map(|canister| match canister.system_state.get_status() {
+                CanisterStatus::Stopping { stop_contexts, .. } => stop_contexts.as_slice(),
+                _ => &[],
+            })
+            .filter_map(|ctx| *ctx.call_id())
+            .collect();
+
         let orphaned: Vec<(StopCanisterCallId, CanisterId)> = state
             .metadata
             .subnet_call_context_manager
             .iter_stop_canister_calls()
             .filter_map(|(call_id, stop_canister_call)| {
-                let canister_id = stop_canister_call.effective_canister_id;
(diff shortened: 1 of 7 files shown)


rs/execution_environment/src/execution_environment.rs
```diff
@@ -4571,6 +4571,50 @@ impl ExecutionEnvironment {
...
+            .metadata
+            .subnet_call_context_manager
+            .iter_stop_canister_calls()
+            .filter_map(|(call_id, stop_canister_call)| {
+                if call_ids_with_context.contains(call_id) {
+                    None
+                } else {
+                    Some((*call_id, stop_canister_call.effective_canister_id))
+                }
+            })
```
alin-at-dfinity: Just for my own edification: these calls / contexts are all subnet-local, right? I.e. it's the management canister keeping track of a local canister stopping.

rs/execution_environment/src/execution_environment.rs
```diff
@@ -4586,6 +4630,8 @@ impl ExecutionEnvironment {
         let mut canister_states = state.take_canister_states();
         let time = state.time();
 
+        self.cleanup_orphaned_stop_canister_calls(&mut state, &canister_states);
```
alin-at-dfinity: Nit: Wouldn't it be sufficient to call this once, e.g. during checkpoint loading? But feel free to leave it as it is if that would require too many changes.

alin-at-dfinity (approved):

rs/execution_environment/src/execution_environment.rs
mraszyk: Indeed, they are created if the mgmt canister processes `stop_canister` call for a canister within the local `ReplicatedState` and supposed to be dropped once the call completes. Due to bugs, some instances were not properly cleaned up in the past and this PR removes such "orphan" instances.

rs/execution_environment/src/execution_environment.rs
mraszyk: Agreed, also because currently metrics are updated based on the orphans which does not make sense. I did the change in [this](https://github.com/dfinity/ic/pull/9990/commits/a568c33b7687877522522d0f395f68fc9a63cb74) commit and would appreciate if you could please take one more look if it does the job (and does not breaking anything else).

schneiderstefan: It's ok to do it during checkpoint loading, as long as we are very sure that it is a no-op when there are no orphans left, and that there cannot be any orphans except during the load just after replica upgrade. If we are wrong about these assumptions we can trigger a state divergence.

In my opinion it would be a very bad idea to do it in checkpoint loading and relying on these assumptions, if it weren't for the fact that this is temporary code that will be deleted very soon. But since it is, I guess it's ok.

mraszyk: > If we are wrong about these assumptions we can trigger a state divergence.

How about moving `state.remove_orphaned_stop_canister_calls();` to just before "[round preparation](https://github.com/dfinity/ic/blob/2abc4da5e8921f7e90b68ee6c9d9a4d3f0a7a27a/rs/execution_environment/src/scheduler.rs#L1179)" in `SchedulerImpl::execute_round` then? This way:
- the cleanup is decoupled from checkpoint loading (i.e., deterministic);
- no metrics are updated based on orphans (improvement over the initial approach from this PR).

schneiderstefan: I think that's a better way. It addresses the two (minor) points of possible nondeterminism and confusing metrics, and the performance impact should be minimal, even if we do it more often that really necessary. You could also guard it on `current_round_type` being checkpoint, so that it doesn't have to run every round.

mraszyk: Done: https://github.com/dfinity/ic/pull/9990/changes/3fa582a75c42e870f79f11d7c454ff034086e79d

schneiderstefan (approved):

PR #9994, opened 2026-04-23
refactor: use ExecutionEnvironment::execute_mgmt_operation for stop_canister

mraszyk:
This PR refactors mgmt canister implementation to use `ExecutionEnvironment::execute_mgmt_operation` for `stop_canister` (analogously to other methods). It also add tests ensuring that no orphan `StopCanisterCall` emerges.

### DIFF
--- a/rs/execution_environment/src/execution_environment.rs
+++ b/rs/execution_environment/src/execution_environment.rs
@@ -622,6 +622,7 @@ impl ExecutionEnvironment {
             Some(canister) => {
                 let saved_canister = canister.clone();
                 let saved_round_limits = round_limits.clone();
+                let saved_msg = msg.clone();
                 match op(canister, msg, round_limits, &mut consumed_cycles) {
                     Ok(response) => self.process_canister_manager_result(
                         Ok(response),
@@ -632,6 +633,7 @@ impl ExecutionEnvironment {
                     Err(err) => {
                         *canister = saved_canister;
                         *round_limits = saved_round_limits;
+                        *msg = saved_msg;
                         consumed_cycles.apply(
                             canister,
                             round_limits,
--- a/rs/execution_environment/src/execution_environment.rs
+++ b/rs/execution_environment/src/execution_environment.rs
@@ -580,20 +580,30 @@ impl ExecutionEnvironment {
             .saturating_sub(state.callback_count()) as i64
     }
 
-    /// Executes a (mgmt canister) operation on the canister state
+    /// Executes a management operation on the canister state
     /// for a given canister ID.
-    /// Changes to the canister state, message, and round limits
-    /// are discarded if the operation fails with an error.
     ///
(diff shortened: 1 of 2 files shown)


rs/execution_environment/src/execution_environment.rs
alin-at-dfinity, line 654: I'm having a hard time parsing this. It seems to me like restoring `msg` simply restores the attached cycles to the original amount.

I'm guessing we want to revert any charges made by the operation and instead apply the charge for the failed operation. But if I were to look at this code independently of this PR, I'd have a hard time understanding that.

Maybe a brief comment might be in order?

alin-at-dfinity (approved):

rs/execution_environment/src/execution_environment.rs
mraszyk: I've updated the doc comment for this function:
```
Changes to the canister state, message, and round limits
are discarded if the operation fails with an error.
```
There's also test coverage for cycles attached to `stop_canister` calls: In rs/`execution_environment/src/execution_environment/tests.rs`, there's a helper `stop_canister_refunds_cycles` and four test cases. The helper attaches 1T cycles to the stop_canister call and asserts two invariants:
  1. The combined balance of the proxy + target canister must not increase (cycles_after <= cycles_before).
  2. The loss is bounded to at most 1B cycles (cycles_before <= cycles_after + 1_000_000_000).

alin-at-dfinity: I suppose I have a hard time understanding what exactly `consumed_cycles.apply()` does. And why it's only called on the `Err` branch. The rest is more or less obvious.

mraszyk: This is described here:
```
    /// If the operation fails with an error, the amount of cycles
    /// recorded in the mutable argument of type `ConsumedCyclesForInstructions`
    /// is charged and the charge is expected to succeed.
    /// An example is charging for uploading an existing WASM chunk
    /// which fails with a corresponding error, but cycles are still
    /// charged for the work of hashing the uploaded WASM chunk.
    /// In particular, this means that a dedicated "out of cycles"
    /// error and no charge should be recorded if the canister is
    /// completely out of cycles.
```

mraszyk: The idea is that
- either the operation succeeds and all changes are reflected in CanisterState, CanisterCall, RoundLimits;
- or the operation fails and then we revert to the very original CanisterState, CanisterCall, and RoundLimits, but because we actually did some work, we charge cycles recorded in `consumed_cycles` and update instructions in `RoundLimits`.

alin-at-dfinity: I see now. `consumed_cycles.apply(canister, ...)` applies the charge to the canister. That's what was confusing me. Thanks.

PR #9996, opened 2026-04-23
refactor: validate_and_update_canister_settings

mraszyk:
This PR refactors `CanisterManager::validate_and_update_canister_settings` to eliminate the validate/apply split for canister settings, removing all `.expect()` panics.

This is possible now that
- changing canister settings operates on a clone of canister state and round limits; and
- canister creation only commits the canister state upon success and operates on a clone of round limits (the latter was changed in this PR).

At the same time, memory accounting for canister log buffer resize is fixed and memory matrix tests are added to cover the fixes.

Notes:
- some tests now create canisters with log memory limit set to zero so that canister creation does not fail due to memory checks for canister log buffer size (added by this PR);
- `LogResizeNotEnoughCycles` is dropped because the generic helper for mgmt canister memory usage accounting uses `InsufficientCyclesInMemoryGrow`;
- default canister log memory limit is implemented analogously to default `wasm_memory_limit` (the current implementation uses the default whenever `None` is specified in settings, including when updating settings, which is wrong).


rs/execution_environment/tests/canister_logging.rs
```diff
@@ -2634,8 +2633,10 @@ fn test_fetch_canister_logs_update_call_deducts_cycles() {
 
 #[test]
 fn test_log_memory_store_feature_flag_via_execution_test_builder() {
-    // With the flag disabled (default), canister creation does not allocate a ring buffer.
-    let mut test = ExecutionTestBuilder::new().build();
+    // With the flag disabled, canister creation does not allocate a ring buffer.
+    let mut test = ExecutionTestBuilder::new()
+        .with_log_memory_store_feature_disabled()
```
mraszyk, line 2638: more explicit

rs/execution_environment/src/canister_manager.rs
```diff
@@ -1548,30 +1488,32 @@ impl CanisterManager {
...
         );
 
-        // Canister creation's first-time allocation is a different event class
-        // from user-triggered `log_memory_limit` resize — don't mix their
-        // distributions. Pass `None` to skip observation here.
-        // If validation fails the canister is not inserted into state, so there
-        // is no need to revert round_limits: validate_and_update_canister_settings
-        // only modifies round_limits after all checks pass.
-        self.validate_and_update_canister_settings(
+        // If validation fails, the canister is not inserted into state, but
```
schneiderstefan, line 1491: This system with 2 of the arguments in the middle of a long list of arguments being &mut seems unnecessarily fragile to me. Not a problem with your PR of course, as it was like that before, but more a general observation.

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -7558,7 +7578,14 @@ fn create_canister_when_compute_capacity_is_oversubscribed() {
         .set_balance(Cycles::new(2_000_000_000_000_000));
 
     // Create a canister with default settings.
```
schneiderstefan: Why do so many unrelated tests have to explicitly set the log limit now? Also, this comment calling this default settings is a bit misleading now.

mraszyk: > Why do so many unrelated tests have to explicitly set the log limit now?

Before this PR, memory usage checks were not performed for the default value of log memory limit (4KiB), this PR fixes that so that memory usage checks are always performed for log memory limit.

> Also, this comment calling this default settings is a bit misleading now.

Thanks for catching that! Changed the comment to "Create a canister with no compute allocation." consistently with the subsequent comments "Create a canister with zero compute allocation." and "Create a canister with compute allocation.".

rs/execution_environment/src/canister_manager.rs
mraszyk: Moved the `&mut` args to be the first args of `validate_and_update_canister_settings`. Or did I misunderstand your feedback?

schneiderstefan (approved):

PR #9998, opened 2026-04-23
feat(governance): Store snapshot ID in TakeCanisterSnapshot proposals.

daniel-wong-dfinity-org:
This makes use of a field that was introduced recently: `successful_value`. Prior to this, there was no good way of knowing the ID of the snapshot.

Limitation: this does not work when Governance is being snapshotted. This is because snapshotting is executed asynchronously in this case; the proposal is marked as "Successful" once snapshottting is LAUNCHED, not when it has completed successfully. The reason for async is to avoid deadlock.

# References

Closes [NNS1-4339].

[NNS1-4339]: https://dfinity.atlassian.net/browse/NNS1-4339

### DIFF
--- a/rs/nns/governance/api/src/types.rs
+++ b/rs/nns/governance/api/src/types.rs
@@ -2065,12 +2065,19 @@ pub struct ProposalInfo {
 #[derive(candid::CandidType, candid::Deserialize, serde::Serialize, Clone, PartialEq, Debug)]
 pub enum SuccessfulProposalExecutionValue {
     CreateCanisterAndInstallCode(CreateCanisterAndInstallCodeOk),
+    TakeCanisterSnapshot(TakeCanisterSnapshotOk),
 }
 
 /// The result of a successful CreateCanisterAndInstallCode proposal execution.
 #[derive(candid::CandidType, candid::Deserialize, serde::Serialize, Clone, PartialEq, Debug)]
 pub struct CreateCanisterAndInstallCodeOk {
     pub canister_id: Option<PrincipalId>,
+}
+
+/// The result of a successful TakeCanisterSnapshot proposal execution.
+#[derive(candid::CandidType, candid::Deserialize, serde::Serialize, Clone, PartialEq, Debug)]
+pub struct TakeCanisterSnapshotOk {
+    pub snapshot_id: Vec<u8>,
 }
 
 /// Network economics contains the parameters for several operations related
--- a/rs/nns/governance/api/src/types.rs
+++ b/rs/nns/governance/api/src/types.rs
@@ -2080,6 +2080,18 @@ pub struct TakeCanisterSnapshotOk {
     pub snapshot_id: Vec<u8>,
 }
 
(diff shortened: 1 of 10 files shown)


rs/nns/governance/api/src/types.rs
```diff
@@ -2073,6 +2074,24 @@ pub struct CreateCanisterAndInstallCodeOk {
...
+#[derive(candid::CandidType, candid::Deserialize, serde::Serialize, Clone, PartialEq, Debug)]
+pub struct TakeCanisterSnapshotOk {
+    pub snapshot_id: Vec<u8>,
+}
+
+impl From<CreateCanisterAndInstallCodeOk> for SuccessfulProposalExecutionValue {
+    fn from(ok: CreateCanisterAndInstallCodeOk) -> Self {
+        Self::CreateCanisterAndInstallCode(ok)
+    }
+}
```
jasonz-dfinity: Is this intended?

jasonz-dfinity, line 2087: Is this intended?

rs/nns/governance/src/proposals/take_canister_snapshot.rs
```diff
@@ -47,6 +50,25 @@ impl CallCanister for TakeCanisterSnapshot {
     }
 }
 
+impl CallCanisterReply for TakeCanisterSnapshotOk {
+    fn try_decode(encoded_reply: &[u8]) -> Result<Option<Self>, GovernanceError> {
+        let response = Decode!(encoded_reply, TakeCanisterSnapshotResponse).map_err(|err| {
+            GovernanceError::new_with_message(
+                crate::pb::v1::governance_error::ErrorType::External,
```
jasonz-dfinity:
```suggestion
                ErrorType::External,
```

rs/nns/governance/src/pb/conversions/mod.rs
jasonz-dfinity: Now that there are multiple variants, this could be confusing if it ever happens. I think a simple fix is to `impl From<pb::SuccessfulProposalExecutionValue> for Option<api::SuccessfulProposalExecutionValue>`. The change is low value, so it's not blocking, but I think it's also low effort, if you agree it makes sense.

jasonz-dfinity (approved):

rs/nns/governance/api/src/types.rs
daniel-wong-dfinity-org: Yes, because the other inner success value (type) also has an "upgrade" conversion.

rs/nns/governance/src/proposals/take_canister_snapshot.rs
daniel-wong-dfinity-org: Doh. This is me not carefully reading everything Claude writes...

rs/nns/governance/src/pb/conversions/mod.rs
daniel-wong-dfinity-org: SG.

PR #9850, opened 2026-04-14
fix: move certified height channel for HTTP handler to StateManagerImpl

mraszyk:
Move the `max_certified_height` channel from `CertifierImpl` to `StateManagerImpl`.

The reason for this change is that the consensus certifier was updating the channel after every `deliver_state_certification`, but the certified state is not available for heights that have not been committed and cloned by the state manager.

This issue has also been observed in production (see [slack](https://dfinity.slack.com/archives/C020G13AS4F/p1773656957580139)) and mitigated by a work-around that checks if the channel was not updated too early ([PR](https://github.com/dfinity/ic/pull/9691)). This PR implements a proper fix by ensuring that the channel is not updated too early at the first place.

PR #10001, opened 2026-04-23
fix: Revert async hashing

michael-weigelt:
This PR reverts https://github.com/dfinity/ic/pull/9850 and https://github.com/dfinity/ic/pull/8464 because the latter may have introduced a bug. The former depends on the latter, so reverting them together leads to the smallest conflict.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7926,6 +7926,7 @@ dependencies = [
  "mockall",
  "prometheus",
  "slog",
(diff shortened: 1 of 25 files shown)


pierugo-dfinity (approved):

cgundy (approved):

alin-at-dfinity (approved):
☹️

PR #10002, opened 2026-04-23
fix: deflake //rs/tests/execution:canister_migration_test_head_nns

basvandijk:
## Root cause

The test `//rs/tests/execution:canister_migration_test_head_nns` flaked 3 times in the last week for the same underlying reason: `ic-agent` calls to the replica's public API (`:8080`) intermittently failed with transient `AgentError::TransportError` variants (`ConnectionRefused`, `BrokenPipe`, `TimedOut`) during brief replica restart windows, and the test treated every such transient error as fatal via `.expect(...)`.

Representative failure messages from the logs:

- `Failed to call migration_status.: TransportError(Reqwest(... ConnectError("tcp connect error", ... ConnectionRefused ...)))`
- `Failed to call migration_status.: TransportError(Reqwest(... hyper::Error(Io, Custom { kind: BrokenPipe ... })))`
- `Failed to call migrate_canister.: TransportError(Reqwest(... hyper::Error(Io, Kind(TimedOut)) on /api/v3/.../read_state))`

In the 2026-04-17 run the replica was restarted by the orchestrator (`orchestrator_replica_process_start_attempts_total = 2 > 1`), explaining the `ConnectionRefused`/`BrokenPipe` during the restart window.

## Fix

- Wrap the three `migrate_canister` update calls with `retry_agent_on_transport_errors!` (120 s / 5 s defaults), which only retries on `AgentError::TransportError` and surfaces all other agent errors immediately, preserving the existing `Result<(), Option<ValidationError>>` semantics via the macro's `other => Ok(other)` arm.
- Inside the existing `retry_with_msg_async!` polling loop for `migration_status`, match on the agent result: on `TransportError`, `bail!` to trigger a retry; on any other agent error, `panic!` so that real protocol/logic bugs still fail loudly instead of being silently retried away.

This keeps the test's diagnostic value for real regressions while absorbing the short-lived transport blips the replica exhibits during restarts.

## Verification

```
bazel test --test_output=errors --runs_per_test=3 --jobs=3 //rs/tests/execution:canister_migration_test_head_nns
```

passed all 3 runs (max 564 s, min 550 s).

---

### DIFF
--- a/rs/tests/execution/canister_migration_test.rs
+++ b/rs/tests/execution/canister_migration_test.rs
@@ -6,7 +6,7 @@ use anyhow::{Result, bail};
 use candid::{CandidType, Decode, Deserialize, Encode, Principal, Reserved};
 use canister_test::Canister;
 use futures::future::join_all;
-use ic_agent::Agent;
+use ic_agent::{Agent, AgentError};
 use ic_consensus_system_test_utils::rw_message::install_nns_and_check_progress;
 use ic_nns_constants::{GOVERNANCE_CANISTER_ID, MIGRATION_CANISTER_ID, REGISTRY_CANISTER_ID};
 use ic_nns_test_utils::governance::{pause_canister_migrations, unpause_canister_migrations};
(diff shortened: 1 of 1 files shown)


basvandijk:
The `ConnectionRefused` and `BrokenPipe` transport errors where caused by a panic in the replica at [rs/replicated_state/src/page_map/page_allocator/mmap.rs:74](https://github.com/dfinity/ic/blob/4881f66b198e0dfca23f7447649a46d7699212c5/rs/replicated_state/src/page_map/page_allocator/mmap.rs#L74). I think these panics were caused by the kernel bug which was fixed by reverting to an earlier kernel in https://github.com/dfinity/ic/commit/7bfc38f6c709f89884f797211f93d64dd209f03a.

I'm not sure why the Timeout happened. Maybe the replica was just too slow and needs more resources.

In light of this retrying on transport errors would not have helped in these cases. So closing this for now.

PR #10004, opened 2026-04-23
DO NOT MERGE - based on composite-proposal branch. Once that goes in, rebase from 5b07908e013864631779c9a2abde0d6f40eadaa4 feat(ic-admin): Added --snapshot-before to upgrades.

daniel-wong-dfinity-org:
More precisely, this is part of `propose-to-change-nns-canister` (which, btw is a horrible name).

This new flag has an advanced mode: by doing `--snapshot-before=replace=${SNAPSHOT_ID}`, you can make the new snapshot supplant a pre-existing one.

Makes use of the new Batch proposal type. First, TakeCanisterSnapshot is used. Then, InstallCode (like when the new flag is not used).

Written with Claude.

# References.

Closes [NNS1-4307].

[NNS1-4307]: https://dfinity.atlassian.net/browse/NNS1-4307

### DIFF
--- a/rs/registry/admin/bin/main.rs
+++ b/rs/registry/admin/bin/main.rs
@@ -14,6 +14,7 @@ use helpers::{
     parse_proposal_url, shortened_pid_string, shortened_subnet_string,
 };
 use ic_admin::get_routing_table;
+use ic_base_types::SnapshotId;
 use ic_btc_interface::{Fees, Flag, SetConfigRequest};
 use ic_canister_client::{Agent, Sender};
 use ic_canister_client_sender::SigKeys;
@@ -1396,8 +1397,12 @@ struct ProposeToChangeNnsCanisterCmd {
 
     /// If set, a canister snapshot is taken before installing the new wasm.
     /// The proposal becomes a Batch: [TakeCanisterSnapshot, InstallCode].
-    #[clap(long)]
-    snapshot_before: bool,
+    ///
+    /// Optionally, an existing snapshot can be replaced:
+    ///   --snapshot-before                   (no replace)
+    ///   --snapshot-before=replace=<hex-id>  (replace that snapshot)
+    #[clap(long, num_args = 0..=1, default_missing_value = "")]
+    snapshot_before: Option<String>,
 }
 
 #[async_trait]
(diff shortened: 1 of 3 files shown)


rs/registry/admin/bin/main.rs
```diff
@@ -1392,6 +1394,15 @@ struct ProposeToChangeNnsCanisterCmd {
...
     arg_sha256: Option<String>,
+
+    /// If set, a canister snapshot is taken before installing the new wasm.
+    /// The proposal becomes a Batch: [TakeCanisterSnapshot, InstallCode].
+    ///
+    /// Optionally, an existing snapshot can be replaced:
+    ///   --snapshot-before                   (no replace)
+    ///   --snapshot-before=replace=<hex-id>  (replace that snapshot)
+    #[clap(long, num_args = 0..=1, default_missing_value = "")]
+    snapshot_before: Option<String>,
```
jasonz-dfinity, line 1405: Nit: would it make sense to add an additional flag `--replace_snapshot`, to be consistent with the TakeSnapshot proposal? I think the complexity from having an additional format ("replace=...") isn't worth the benefit of having one fewer argument.

jasonz-dfinity (approved):

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org: Yeah, I thought about that originally. What I don't want is for people to HAVE to supply both. I guess one way is to make `--snapshot-before` optional (but not banned either) when `--replace-snapshot` is used. So, supplying both would be supported, but not required.

Another approach is

1. `--snapshot-before=without-replacement` (instead of just `--snapshot-before`) vs.
2. `--snapshot-before=replace=${ID}` (same as what I sent out originally)

Then, it is more symmetrical. WDYT?

I just kind of hate how `--snapshot-before` and `--replace-snapshot` heavily overlap (e.g. when using `--replace-snapshot`, is `--snapshot-before` REQUIRED??).

OTOH, I understand how `replace=${ID}` (which results in having two `=` in one `argv` element) is a bit ugly; OTOH, that's not unheard of. E.g. in `bazel` there's `--test_env=ENV=VALUE`.

FWIW, `--snapshot-before replace=${ID}` is equivalent to `--snapshot-before=replace=${ID}`, so at least you do not see two `=` in one `argv`.

PR #10005, opened 2026-04-23
fix: deflake //rs/tests/ckbtc:ckdoge_minter_basics_test

basvandijk:
## Root cause

Two recent flaky runs of `//rs/tests/ckbtc:ckdoge_minter_basics_test` failed inside `test_retrieve_doge_status` with:

```
failed to call retrieve_doge_status: AgentError(TransportError(Reqwest(reqwest::Error { kind: Request, url: \"http://[...]:8080/api/v3/canister/.../query\", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Kind(TimedOut))) })))
```

In both cases the test had already been polling the minter every 5 s and received `Submitted { .. }` several times successfully. Then a single `ic-agent` query over IPv6 hit a TCP-level timeout (`hyper::Error(Io, Kind(TimedOut))`) — a transient connectivity/replica slowness blip, not a logic bug. The second attempt of the same bazel invocation passed, confirming the underlying infrastructure was fine.

The retry loop in `test_retrieve_doge_status` already tolerates non-terminal statuses (`Pending`, `Signing`, `Submitted`) but called `.expect(...)` on the agent result, so any transport error panicked immediately and failed the whole attempt.

## Fix

On `ic_agent::AgentError::TransportError`, log and retry (still counting the attempt against the existing 30-retry budget / 5 s sleep) instead of panicking. All other errors still panic.

## Verification

```
bazel test --test_output=errors --runs_per_test=3 --jobs=3 //rs/tests/ckbtc:ckdoge_minter_basics_test
```
3/3 runs passed.

Created following the steps in `.claude/skills/fix-flaky-tests/SKILL.md`.

### DIFF
--- a/rs/tests/ckbtc/ckdoge_minter_basics_test.rs
+++ b/rs/tests/ckbtc/ckdoge_minter_basics_test.rs
@@ -3,7 +3,7 @@ use anyhow::Result;
 use bitcoin::{Amount, Txid, dogecoin, dogecoin::Address};
 use candid::Nat;
 use candid::{Decode, Encode, Principal};
-use ic_ckdoge_agent::CkDogeMinterAgent;
+use ic_ckdoge_agent::{CkDogeMinterAgent, CkDogeMinterAgentError};
 use ic_ckdoge_minter::{
     UpdateBalanceArgs, UtxoStatus,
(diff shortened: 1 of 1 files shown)


gregorydemay (approved):
Thanks @basvandijk !

PR #10013, opened 2026-04-24
feat: [DSM-103] Actual priority queue for long executions

alin-at-dfinity:
Instead of a binary prioritized / opportunistic flag, explicitly (record and) prioritize long executions based on number of slices executed, AP and round when the long execution started. This ensures that we don't starve low priority canisters (which may happen with bounded AP and just the right distribution across execution cores).

Also switch from persisting `SubnetSchedule` spread across individual canister states to persisting it as part of the subnet's `SystemMetadata`.

### DIFF
--- a/rs/execution_environment/src/scheduler/round_schedule.rs
+++ b/rs/execution_environment/src/scheduler/round_schedule.rs
@@ -8,9 +8,7 @@ use ic_config::flag_status::FlagStatus;
 use ic_logger::{ReplicaLogger, error};
 use ic_replicated_state::canister_state::NextExecution;
 use ic_replicated_state::{CanisterPriority, CanisterState, ReplicatedState};
-use ic_types::{
-    AccumulatedPriority, ComputeAllocation, ExecutionRound, LongExecutionMode, NumInstructions,
-};
+use ic_types::{AccumulatedPriority, ComputeAllocation, ExecutionRound, NumInstructions};
 use ic_utils::iter::left_outer_join;
 use more_asserts::debug_assert_gt;
 use num_traits::SaturatingSub;
@@ -30,7 +28,8 @@ const ZERO: AccumulatedPriority = AccumulatedPriority::new(0);
 const ONE_PERCENT: AccumulatedPriority = AccumulatedPriority::new(1 * MULTIPLIER);
 
 /// 100% in accumulated priority.
-const ONE_HUNDRED_PERCENT: AccumulatedPriority = AccumulatedPriority::new(100 * MULTIPLIER);
+pub(super) const ONE_HUNDRED_PERCENT: AccumulatedPriority =
+    AccumulatedPriority::new(100 * MULTIPLIER);
 
 const fn from_ca(ca: ComputeAllocation) -> AccumulatedPriority {
     AccumulatedPriority::new(ca.as_percent() as i64 * MULTIPLIER)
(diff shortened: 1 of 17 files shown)


rs/state_manager/src/checkpoint.rs
```diff
@@ -889,8 +888,14 @@ pub fn load_canister_state(
...
-        priority_credit: canister_state_bits.priority_credit,
-        long_execution_mode: canister_state_bits.long_execution_mode,
+        // We can only be loading an aborted execution, so zero executed slices.
+        executed_slices: 0,
+        // Set a long execution start round where necessary.
+        long_execution_start_round: if canister_state.has_long_execution() {
+            Some(0.into())
+        } else {
+            None
+        },
```
alin-at-dfinity, line 898: We will lose `long_execution_mode` during the next replica upgrade; and `long_execution_start_round` if there is a corresponding replica downgrade. But this is acceptable (i) as all replicas are restarted during an upgrade/downgrade, so there's no risk of divergence; and (ii) canister priorities are transient information (as proved by the fact that we reset them once a day, on average).

rs/execution_environment/src/scheduler/round_schedule.rs
```diff
@@ -47,21 +46,32 @@ pub(super) struct CanisterRoundState {
...
     pub fn new(canister: &CanisterState, canister_priority: &CanisterPriority) -> Self {
+        // Ensure that `long_execution_start_round` matches the canister state.
+        debug_assert_eq!(
+            canister.has_long_execution(),
+            canister_priority.long_execution_start_round.is_some(),
+            "canister: {:?}, task_queue: {:?}, canister_priority: {:?}",
+            canister.canister_id(),
+            canister.system_state.task_queue,
+            canister_priority,
+        );
```
alin-at-dfinity, line 66: Note: We set and reset `long_execution_start_round` in `end_iteration()` below. And we have a `debug_assert` here to help prevent regressions. Plus, there is logic during checkpoint loading to correctly populate `long_execution_start_round` with `Some` / `None`, depending on the actual canister state.

But the only potential issue with an incorrectly set / cleared `long_execution_start_round` would be that the canister gets scheduled into the wrong pool (long vs new executions). But as soon as the canister gets one round of execution, its `long_execution_start_round` would be updated correctly.

rs/execution_environment/src/scheduler/round_schedule.rs
```diff
@@ -72,17 +82,34 @@ impl CanisterRoundState {
...
+            // priority as a newly started long execution (executed slices == 1). This is to
+            // avoid starvation of aborted executions.
+            (Some(self_start_round), Some(other_start_round)) => other
+                .executed_slices
+                .max(1)
+                .cmp(&self.executed_slices.max(1))
+                .then_with(|| other.accumulated_priority.cmp(&self.accumulated_priority))
+                .then_with(|| self_start_round.cmp(&other_start_round))
+                .then_with(|| self.canister_id.cmp(&other.canister_id)),
+        }
```
adambratschikaye:
```suggestion
            // Among long executions, sort by executed slices descending; AP descending; start round
            // ascending; then break ties by canister ID.
            //
            // An aborted execution (executed slices == 0) is considered to have the same
            // priority as a newly started long execution (executed slices == 1). This is to
            // avoid starvation of aborted executions.
            (Some(self_start_round), Some(other_start_round)) => other
                .executed_slices
                .max(1)
                .cmp(&self.executed_slices.max(1))
                .then_with(|| other.accumulated_priority.cmp(&self.accumulated_priority))
                .then_with(|| self_start_round.cmp(&other_start_round))
                .then_with(|| self.canister_id.cmp(&other.canister_id)),
        }
```

adambratschikaye (approved):

rs/execution_environment/src/scheduler/round_schedule.rs
alin-at-dfinity: Good catch, thank you.

PR #10014, opened 2026-04-24
perf(crypto): avoid some unnecessary clones

pierugo-dfinity:
I recommend reviewing by commit:
- https://github.com/dfinity/ic/pull/10014/commits/d25992fde4cf217345de4b18c095fe8141d1151c
  - `payload_verifier.rs`: avoid clones for which we only need borrows anyways
  - `batch_delivery.rs`: avoid cloning the internal transcripts as we do not need ownership when converting to protobuf.
- https://github.com/dfinity/ic/pull/10014/commits/c48f195c9e9b2416be1ceb3a0905fec2bfac08ef
  - Avoid cloning dealings as we can instead pass ownership.
- Half of the more "invasive" changes: https://github.com/dfinity/ic/pull/10014/commits/90e517ac4ebf1be1f0cb01d18e787a2a07359677
  - Change the interface of `NiDkgAlgorithm::create_transcript` to now take ownership of the dealings which allows to avoid a second clone down the stack. This comes for free from a Consensus perspective.
- Other half of the more invasive changes: https://github.com/dfinity/ic/pull/10014/commits/4ad048b7d2730a7577e101bf0338a55c01544095
  - Similarly, change the interface of `IDkgProtocol::create_transcript` to now take ownership of the dealings which allows to avoid cloning them and the commitments down the stack. This also positively affects `verify_transcript`. This also comes for free from Consensus.


randombit (approved):
Looks good to me thanks @pierugo-dfinity 

PS this means you're now in charge of NIDKG, take care of her! :)

rs/consensus/src/consensus/batch_delivery.rs
eichhorl: Can we also avoid these clones?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -381,12 +382,8 @@ pub(super) fn create_summary_payload(
 fn create_transcript(
```
eichhorl: Do we need this function?

rs/crypto/benches/idkg.rs
```diff
@@ -303,7 +303,7 @@ fn bench_create_transcript<M: Measurement, R: RngCore + CryptoRng>(
             },
             |(receiver, dealings)| {
                 let (_, _, params) = bench_context.get().unwrap();
-                create_transcript_or_panic(receiver, params, dealings)
+                create_transcript_or_panic(receiver, params, dealings.clone())
```
eichhorl: should we move the clone into the bench setup?

rs/crypto/benches/ni_dkg.rs
```diff
@@ -153,7 +153,12 @@ fn bench_create_transcript<M: Measurement, R: Rng + CryptoRng>(
                 (config, dealings, creator_node_id, &env.crypto_components)
             },
             |(config, dealings, creator_node_id, crypto_components)| {
-                create_transcript(config, crypto_components, dealings, **creator_node_id)
+                create_transcript(
+                    config,
+                    crypto_components,
+                    dealings.clone(),
```
eichhorl, line 155: should we move the clone into the bench setup?

eichhorl (approved):
Very nice!

kpop-dfinity:
Maybe we should enable [redundant_clone](https://rust-lang.github.io/rust-clippy/master/index.html?search=clone#redundant_clone) and [needless_pass_by_value](https://rust-lang.github.io/rust-clippy/master/index.html?search=needless#needless_pass_by_value) clippy lints (at least in the CON owned crates). Some of these cloned you're removing should be caught by them

rs/consensus/src/consensus/batch_delivery.rs
pierugo-dfinity: We can! + a few more (batch messages + initial dealings for IDKG). Though it required a bit of refactoring, PTAL https://github.com/dfinity/ic/pull/10014/commits/9f32e684ab4255d27661bb2ad6d78fdf8548650e. In particular, this heavily relies on the fact that we are the sole owner of the `Arc` in the `Payload`. From what I can tell, this should be the case, testing in https://github.com/dfinity/ic/pull/10014/commits/e8814284ea7626e11da61e49a41e8b9872b95a54...

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: No indeed, thanks! https://github.com/dfinity/ic/pull/10014/commits/76ab7f9b2e4b8ac8cfbc9cdbf6da9ff22178a338

rs/crypto/benches/idkg.rs
pierugo-dfinity: I'm not sure I see what you mean. The setup already returns owned dealings and the routine is given only a (mutable) reference to those dealings. Is there a way to still avoid the clone in the routine?

pierugo-dfinity: Ah, you're right, I see we can if we use `iter_batched` instead of `iter_batched_ref`. Is this what you're suggesting?

rs/consensus/src/consensus/batch_delivery.rs
eichhorl: Ah, I just meant to change `RemoteDkgResults` to hold references instead

rs/crypto/benches/idkg.rs
eichhorl: Yes, in principal it should be possible to instrument the create_transcript function without including the cloning of dealings

rs/crypto/benches/ni_dkg.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10014/commits/771686015990a61aa355c46c7ced859203b06d0a

rs/crypto/benches/idkg.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10014/commits/2a573480abebe23fc78756aedefae83ff50cee86

rs/consensus/src/consensus/batch_delivery.rs
pierugo-dfinity: Oops.

> testing in https://github.com/dfinity/ic/commit/e8814284ea7626e11da61e49a41e8b9872b95a54...

So it [looks like](https://kibana.testnet.dfinity.network/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:'2026-04-27T00:26:10.291Z',to:now))&_a=(columns:!(MESSAGE,ic_subnet,ic_node,ic),filters:!(),hideChart:!f,index:testnet-vector-push,interval:auto,query:(language:kuery,query:Pierugo),sort:!(!(timestamp,desc)))) we may not always be the only owner of the `Arc`.
But it happened just twice in the entire CI, so I still think there's some benefit to it?

pierugo-dfinity:
@eichhorl PTAnotherL, I also adapted the transcript creation in data blocks after merging master.

rs/consensus/src/consensus/batch_delivery.rs
eichhorl: Is there a way to collect some mainnet data? I considered `try_unwrap` a couple times before, but maybe we should open that can of worms in a separate PR.

pierugo-dfinity: > maybe we should open that can of worms in a separate PR

Right, I reverted the changes and instead followed your suggestion of

> I just meant to change RemoteDkgResults to hold references

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -223,10 +225,16 @@ pub(crate) fn create_early_remote_transcripts(
 
         // For each config, try to build the necessary (dkg_id, callback_id, transcript_result) triple
         for config in configs.iter() {
+            let dealings = all_dealings.remove(config.dkg_id()).unwrap_or_else(|| {
+                unreachable!(
+                    "We checked that all configs have enough dealings above. This is a bug."
+                )
```
eichhorl: Alternatively we could use `debug_unreachable!` and `break;`

eichhorl (approved):

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: What about `let dealings = all_dealings.remove(config.dkg_id()).unwrap_or_default();` to keep the same functionality as before my change? This would hit `Err(err) if err.is_reproducible()` and include a reject response. My concern if we `break`, is that we could potentially include a single transcript for a config that requires two (in case we hit the `break` in the second transcript).

eichhorl: Good point, then maybe let's keep it as is 👍

PR #10015, opened 2026-04-24
chore(crypto): CRP-1799 Don't panic on invalid ZK proof instances in NIDKG

randombit:
Previously in NIDKG a proof instance type could be instantiated even if manifestly invalid; then check_instance would return an error. In some cases this error would propagate to the caller, but in others it would simply panic.

Change this so that creating a ZK proof instance is falliable, and handle the error directly. This also eliminates any possibility of forgetting to call check_instance in the right spot.


randombit:
Claude banged this out based on a 4 year old incomplete branch I had lying around, plus the ticket description. It looks good to me, and these panics were on the top of my list of remaining NIDKG nits. Merge or not as you please.

PR #10027, opened 2026-04-27
feat: CON-1716 Allow specifying initial DKG subnet for subnet splitting and rental

eichhorl:
Creating, recovering or splitting a subnet requires new key material for the nodes comprising the subnet. This key material needs to be generated by a different subnet. Until recently, this key generation was exclusively done by the NNS. Since https://github.com/dfinity/ic/pull/9782 however, it may be done by any subnet (and specifically, _not_ the NNS).

The linked PR already updated proposals for creating and recovering subnets to optionally specify which subnet should be responsible for generating the initial key material.

This PR similarly updates the remaining two proposal types that also cause new key material to be generated, namely subnet splitting and subnet rental (which is a subnet creation).


rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -725,6 +755,26 @@ mod tests {
         },
         Err(PayloadValidationError::SourceSubnetHalted)
     )]
+    #[case::initial_dkg_subnet_is_source_subnet(
```
daniel-wong-dfinity-org, line 758: No(t asking for) change, because this is not your fault. Nevertheless, I have to go on the record as being emphatically against this type of code:

Is this just a way of generating tests? Couldn't that be done with a `for` loop? E.g.

```
for (input, expected_output) in [(a, A), ...] {
  assert_code_under_test(input, expected_output);
}
```

Notice how this ordinary Rust code requires ZERO esoteric knowledge, AND that it is NOT verbose. All it requires is knowledge of Rust itself. This really shows how fancy != good, primitive != bad. To the contrary, primitive (code) = highly sophisticated (engineer), at least in this case.

Requiring esoteric knowledge is peak "hostile to the reader". Whereas, one of the highest commandments is this:

> Optimize for the reader.

rs/registry/canister/unreleased_changelog.md
```diff
@@ -8,7 +8,9 @@ on the process that this file is part of, see
 # Next Upgrade Proposal
 
 ## Added
-
+* Added an optional field `initial_dkg_subnet_id` to `SplitSubnetPayload` and `FulfillSubnetRentalRequest`,
+  which, when present, determines the subnet to which the resulting `SetupInitialDKG` management
+  canister call should be routed.
```
daniel-wong-dfinity-org: This seems like a highly "implementation detail". What abilities do users actually gain from this?

daniel-wong-dfinity-org:
```suggestion
* Added an optional field `initial_dkg_subnet_id` to `SplitSubnetPayload` and `FulfillSubnetRentalRequest`,
  which, when present, determines the subnet to which the resulting `SetupInitialDKG` management
  canister call should be routed.

```

Related things close; unrelated things apart.

daniel-wong-dfinity-org (approved):

rs/registry/canister/unreleased_changelog.md
eichhorl: Thanks, I resolved the merge conflict which removed the old entries 👍

eichhorl: That's fair, I made it a bit more user focused.

pierugo-dfinity (approved):

basvandijk (approved):
I seem to remember there was a very similar PR that did the same recently. Or maybe I'm dreaming things.

yhabib (approved):

PR #10034, opened 2026-04-27
feat: [DSM-103] Actual priority queue for long executions, take 2

alin-at-dfinity:
Instead of a binary prioritized / opportunistic flag, explicitly (record and) prioritize long executions based on number of slices executed, AP and round when the long execution started. This ensures that we don't starve low priority canisters (which may happen with bounded AP and just the right distribution across execution cores).

Also switch from persisting `SubnetSchedule` spread across individual canister states to persisting it as part of the subnet's `SystemMetadata`.


alin-at-dfinity:
This broke `//rs/tests/consensus:subnet_splitting_test_colocate`, so it was rolled back (#10030).

I'm resubmitting it with the fix, plus necessary adjustments to other tests.

adambratschikaye (approved):

PR #10038, opened 2026-04-27
test(guestos_no_failed_systemd_units): log journal of failed units

basvandijk:
When `systemctl list-units --failed` returns non-empty output, parse each failed unit name and fetch its journalctl logs over SSH before the `assert!` fails the test, to aid debugging of flaky failures.

In particular we would like to debug the failing `reload_nftables.service` that failed this test twice on [April 27th, 2026 at 12:36:36 pm](https://dash.dm1-idx1.dfinity.network/invocation/b861247c-5114-4b38-985c-717e4ca91bfb?target=//rs/tests/node:guestos_no_failed_systemd_units#@217) and on [April 23rd, 2026 at 6:28:42 pm](https://dash.zh1-idx1.dfinity.network/invocation/72a57525-3631-4f39-92f8-2ce152e7b207?target=//rs/tests/node:guestos_no_failed_systemd_units#@226).

Note we can't rely on the logs being shipped to ElasticSearch since the test completes either too quickly for the logs to be shipped or something in the nodes fails such that shipping can't be done successfully.

### DIFF
--- a/rs/tests/node/guestos_no_failed_systemd_units.rs
+++ b/rs/tests/node/guestos_no_failed_systemd_units.rs
@@ -31,7 +31,7 @@ fn check_no_failed_systemd_units(env: TestEnv) {
             .expect("Failed to establish SSH session to GuestOS node");
 
         let failed_units = node
-            .block_on_bash_script("systemctl list-units --failed --no-legend --no-pager")
+            .block_on_bash_script("systemctl list-units --failed --no-legend --no-pager --plain")
             .expect("Failed to run systemctl list-units --failed on GuestOS node");
         info!(
             logger,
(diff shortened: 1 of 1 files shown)


rs/tests/node/guestos_no_failed_systemd_units.rs
```diff
@@ -37,6 +37,40 @@ fn check_no_failed_systemd_units(env: TestEnv) {
...
+                let line = line.trim();
+                if line.is_empty() {
+                    continue;
+                }
+                // Lines from `systemctl list-units --failed --no-legend` typically
+                // start with a status glyph (e.g. `●`) followed by the unit name.
+                // Skip the glyph if present and take the unit name as the next token.
+                let unit = line
+                    .split_whitespace()
+                    .find(|tok: &&str| !tok.chars().all(|c| !c.is_alphanumeric()));
```
Bownairo: I _think_ we can use `--plain` above instead of this parsing? It looks like it should drop the glyph and any tree structure.

basvandijk: Good idea. I'll get AI on ASAP.

basvandijk: Done in 7ada207ca9 — switched to `systemctl list-units --failed --no-legend --no-pager --plain` and now just take the first whitespace-separated token of each line as the unit name.

Bownairo (approved):

PR #10040, opened 2026-04-28
test: rejoin_test_long_rounds causes more expensive canister state cloning

mraszyk:
This PR fixes `rejoin_test_long_rounds` to cause more expensive canister state cloning. The motivation for this change is that cloning many (100,000) canisters whose state does not change is cheap now that their canister state is wrapped in an `Arc`. Hence, the updated test cycles through all those many canisters and keeps bumping their freezing threshold so that their canister state under an `Arc` must actually be cloned. The cyclic iteration is a busy computation and thus the original busy canisters with expensive heartbeats could be dropped now.


rs/rust_canisters/statesync_test/test/test.rs
```diff
@@ -160,6 +160,54 @@ fn test_create_many_canisters() {
...
+                        .unwrap()
+                        .unwrap()
+                        .version()
+                        >= initial_version + 2
+                });
+        if all_at_version {
+            break;
+        }
+        env.tick();
+    }
```
alin-at-dfinity: Could you please add a comment for what this code is doing? Because I don't actually understand it.

alin-at-dfinity (approved):
LGTM, thank you.

rs/rust_canisters/statesync_test/test/test.rs
mraszyk: Sure, I made the tests better readable in this commit: https://github.com/dfinity/ic/pull/10040/changes/82f4cac4fa66d8e5aa4c8da7438ed4deb5c14a8f

PR #10046, opened 2026-04-28
refactor: canister ID generation

mraszyk:
This PR refactors canister ID generation into two phases:
- "peeking" the (next) canister ID yields the next canister ID for canister creation (without mutating `ReplicatedState`);
- "committing" that canister ID updates `ReplicatedState` to record that the canister ID is used now (after successful canister creation).

The motivation for this change is a follow-up refactoring that combines canister settings validation and updates (to make the code more robust and avoid potential panics), but that requires ability to derive a canister ID without actually consuming it (currently canister settings are first validated, then a canister ID is consumed, and finally canister state with validated settings is created, but that separation of settings validation and application is fragile and subject to a follow-up refactoring).

### DIFF
--- a/rs/replicated_state/src/metadata_state.rs
+++ b/rs/replicated_state/src/metadata_state.rs
@@ -590,13 +590,9 @@ impl SystemMetadata {
         Ok(())
     }
 
-    /// Generates a new canister ID.
+    /// Computes a new canister ID for canister creation.
     ///
-    /// If a canister ID from a second canister allocation range is generated, the
-    /// first range is dropped. The last canister allocation range is never dropped.
-    ///
-    /// Returns the next canister ID that would be generated, without mutating
-    /// state.  Returns `Err` iff no more canister IDs can be generated.
+    /// Returns `Err` iff no more canister IDs are available.
     pub fn peek_new_canister_id(&self) -> Result<CanisterId, String> {
         // Start off with
         //     (canister_allocation_ranges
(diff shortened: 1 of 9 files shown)


pierugo-dfinity (approved):

rs/replicated_state/src/metadata_state.rs
```diff
@@ -626,26 +623,32 @@ impl SystemMetadata {
...
-                self.canister_allocation_ranges.drop_first();
-            }
+    /// Records `canister_id` as the last generated canister ID and drops any
+    /// exhausted allocation ranges. Must be called with the canister ID returned
+    /// by the immediately preceding `peek_new_canister_id` call, once canister
+    /// creation succeeded and the canister ID should be permanently consumed.
+    ///
+    /// Note. The last canister allocation range is never dropped.
+    pub fn commit_new_canister_id(&mut self, canister_id: CanisterId) {
+        self.last_generated_canister_id = Some(canister_id);
```
alin-at-dfinity, line 640: I would at the very least add a `debug_assert_eq!(canister_id, self.peek_new_canister_id())`. Maybe even an `assert_eq!`, this is not performance critical.

alin-at-dfinity (approved):

rs/replicated_state/src/metadata_state.rs
mraszyk: Makes sense: https://github.com/dfinity/ic/pull/10046/changes/bc0160a74940252d2ba6cb8aea40f485e222ef93

I still only went for a debug_assert to avoid a crash in production since we could still reasonably continue and fix such a bug asynchronously.

jasonz-dfinity (approved):
LGTM for rs/nns/constants and rs/registry/routing_table changes

PR #10047, opened 2026-04-28
feat: Switch GuestOS TEE detection to dfinity.tee kernel cmdline

frankdavid:
- Replace GuestOS TEE/SEV detection based on the SEV_ACTIVE environment variable with detection based on the dfinity.tee kernel command-line flag which is populated in guestos_boot_args.template. The related host changes were submitted in #9575.
- The launch measurements now only contain TEE-specific kernel command lines.
- Rename the shared helper from is_sev_active() to is_tee_enabled() and use the more generic tee instead of sev in the codebase.


rs/orchestrator/src/registration.rs
pierugo-dfinity: Should we also adapt these docs and the log messages for consistency?

pierugo-dfinity (approved):
Approving orchestrator changes

ic-os/defs.bzl
```diff
@@ -292,7 +292,8 @@ def icos_build(
                     # Create GuestLaunchMeasurements JSON for each CPU generation, vCPU count, and boot slot
                     (for vcpu_flag in """ + vcpu_type_flags + """; do
                         for vcpus in """ + vcpu_configs + """; do
-                            for cmdline in "$$BOOT_ARGS_A" "$$BOOT_ARGS_B"; do
+                            # Note: we only add TEE boot args to the launch measurements
```
Bownairo: What does this mean? Something like:
> We only create launch measurements for the TEE boot arg variants

Bownairo (approved):

ic-os/defs.bzl
frankdavid: Fixed.

rs/orchestrator/src/registration.rs
pierugo-dfinity: Maybe also the log messages?

PR #10054, opened 2026-04-29
chore: introduce the protected long-lived featurebranch feat-ai-on-engines

basvandijk:
For Cloud Engines we would like to have a long-lived feature branch for experimenting with AI called: 
`feat-ai-on-engines`.

This branch should be treated almost the same as `master` in particular:

* PRs will be created targeting `feat-ai-on-engines`.

* The branch should be protected such that nobody can push directly to it and you have to go via PRs. The @dfinity/dre team will be allowed to directly push to this branch such that they can push merge commit with `master` without going via a PR which would require many approvals.

* The `Kickoff / CI Main` workflow should be triggered on pushes to this branch and it should upload artifacts.

* The `upload-artifacts` environment, and thus its secrets, are accessible on this branch.

### DIFF
--- a/.github/workflows/ci-kickoff.yml
+++ b/.github/workflows/ci-kickoff.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - feat-ai-on-engines # special protected long-lived feature branch for experimenting with AI on cloud engines.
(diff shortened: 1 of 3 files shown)


NikolaMilosa (approved):
LGTM!

PR #10055, opened 2026-04-29
feat: CON-1445 Start remote DKG as soon as request appears in state

eichhorl:
## Background
The `NiDkgConfig` contains the data necessary for a subnet to run one instance of the NiDKG protocol. For remote DKG (where a subnet generates key material for a different subnet), these configs are derived from request contexts stored in the replicated state.

## Problem
Before this PR, these remote `NiDkgConfig`s were only created and included as part of summary blocks. This means, a remote DKG request had to wait in the replicated state until consensus reaches the next summary block, at which point the config is created and the remote DKG protocol is started. This increases the worst case latency of remote DKG requests (i.e. subnet/engine creations) by up to one DKG interval.

## Proposed Changes
Instead of waiting until the next summary, we can create the `NiDkgConfig`s on demand, as soon as the corresponding request appears in the replicated state. We continue to use the `start_height` of the most recent summary block for these configs, essentially treating them "as if" they had already existed when the last summary was created. The only difference is that the number of remaining blocks to complete the remote DKG may be reduced, since it may be started at any point throughout the interval.

For simplicity, this PR removes the handling of remote DKG as part of summary blocks entirely, which implies that remote `NiDkgConfig`s are no longer stored in summary blocks. Therefore, when creating/validating dealings and transcripts, we now iterate over the union of local configs in the summary and remote configs derived from the replicated state. 

Additionally, this means that config creation errors and timeouts are now also handled by data blocks:
- Config creation errors are included in the data block, which will generate a reject response to the context. 
- A remote DKG request context increases its attempt counter (stored in the summary), whenever a new summary block is created. Once this counter exceeds the maximum number of attempts, the next data block will include a timeout response.
- Completed requests of the current interval (successful or unsuccessful) are detected by traversing previous data blocks, such that the same request isn't handled multiple times.
- Completed requests of the past interval (successful or unsuccessful) are tracked by the summary with an attempt counter of 0. These requests are ignored when creating future `NiDkgConfig`s, such that the same request isn't handled multiple times.
- Note that we consider a request completed if any one of it's transcripts were completed.

## Future Work
- Currently, we may start a remote DKG request near the end of the interval, without enough data blocks left to include enough dealings to finish all required transcripts. This means the request will be retried at the start of the next interval. To improve efficiency, we should not create and include dealings for remote DKGs that were received with an insufficient number of data blocks left. Instead, we should only start working on the request at the start of the next interval.

- Consider increasing the number of dealings per block and the number of prioritized remote DKGs per interval.


rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -406,26 +422,19 @@ pub(super) fn create_summary_payload(
...
     let (mut all_dealings, completed_dkgs) = utils::get_dkg_dealings(pool_reader, parent);
-    let mut transcripts_for_remote_subnets = BTreeMap::new();
     let mut next_transcripts = BTreeMap::new();
     // Try to create transcripts from the last round.
     for (dkg_id, config) in last_summary.configs.iter() {
-        if completed_dkgs.contains(dkg_id) {
-            // Skip DKGs that have already been completed as part of data blocks
+        if dkg_id.target_subnet != NiDkgTargetSubnet::Local {
+            // Skip remote DKGs
             continue;
```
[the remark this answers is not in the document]
eichhorl: Yes, technically including dealings in the last data block is pointless now. More generally, it is pointless to include dealings for transcripts that cannot be completed anymore in the current interval, given a limited amount of remaining data blocks and dealings per block.

These transcripts will instead be retried during the next interval. Fixing this is only an optimization and requires enough work to be deferred to a follow up PR, for which I opened a ticket here: https://dfinity.atlassian.net/browse/CON-1717

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -480,29 +489,15 @@ pub(super) fn create_summary_payload(
...
-            &previous_transcripts,
-            &reshared_transcripts,
-            &completed_target_ids,
-            &last_summary.initial_dkg_attempts,
-            &logger,
-        )?;
+    let remote_dkg_attempts = get_updated_remote_dkg_attempts(
+        last_summary,
+        state.get_ref(),
+        get_completed_target_ids(&completed_dkgs),
```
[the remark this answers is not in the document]
eichhorl: `context.registry_version` should never be ahead of the subnet for more than a couple seconds. If it is, then that indicates a bug and removing such contexts seems like a good idea. Otherwise malformed contexts might never time out, if their registry version is too far in the future.

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -109,26 +115,49 @@ fn create_data_payload(
...
+        .get_state_at(validation_context.certified_height)
+        .map_err(DkgPayloadCreationError::StateManagerError)?;
+    let remote_config_results = build_callback_id_config_map(
+        this_subnet_id,
+        registry_client,
+        state.get_ref(),
+        validation_context.registry_version,
+        last_dkg_summary,
+        &logger,
+    )?;
```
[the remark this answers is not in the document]
eichhorl: I'm proposing to instead limit the number of remote contexts in this PR: https://github.com/dfinity/ic/pull/9821

rs/consensus/dkg/src/lib.rs
```diff
@@ -1537,105 +1724,6 @@ mod tests {
         });
     }
 
-    #[test]
-    fn test_dkg_payload_has_transcripts_for_initial_dkg_requests() {
```
eichhorl, line 1541: Refactoring this test would lead to a duplicate of `test_create_dealing_works_for_remote_dkg` and `test_early_setup_initial_dkg_transcripts`

rs/consensus/dkg/src/lib.rs
```diff
@@ -2397,180 +2427,66 @@ mod tests {
...
+                assert_highest_block_validates(&deps);
+                assert_eq!(extract_dealings_from_highest_block(&deps.pool).len(), 0);
+                let remote_dkgs = extract_remote_dkgs_from_highest_block(&deps.pool);
+                check_transcripts(!setup_first, remote_dkgs);
             });
         }
     }
 
-    #[test]
-    fn test_dkg_payload_has_transcript_for_reshare_chain_key_request() {
```
eichhorl, line 2443: Refactoring this test would lead to a duplicate of `test_early_reshare_chain_key_transcripts`

rs/tests/consensus/subnet_splitting_test.rs
```diff
@@ -55,7 +55,7 @@ use candid::Principal;
 use slog::{Logger, info};
 use std::{thread, time::Duration};
 
-const DKG_INTERVAL: u64 = 9;
+const DKG_INTERVAL: u64 = 29;
```
eichhorl, line 58: This test turned flaky with an interval of 9. We no longer create/deliver transcripts as part of summary blocks, so the effective interval length usable for remote transcripts decreased by one. Luckily, with this change increasing the interval length no longer leads to higher latency of remote DKG.

rs/consensus/dkg/src/lib.rs
```diff
@@ -211,9 +223,13 @@ impl DkgImpl {
         }
 
         // If the dealing refers a config which is not among the ongoing DKGs,
-        // we reject it.
+        // we reject it, unless it is a remote DKG, in which case we defer it
+        // until the request appears in the state, or the dealing is purged.
         let config = match configs.get(message_dkg_id) {
             Some(config) => config,
+            None if message_dkg_id.target_subnet != NiDkgTargetSubnet::Local => {
```
pierugo-dfinity: Is
```suggestion
            None if matches!(message_dkg_id.target_subnet, NiDkgTargetSubnet::Remote(_)) => {
```
more appropriate?

rs/consensus/dkg/src/lib.rs
```diff
@@ -829,14 +874,88 @@ mod tests {
...
+                let change_set = dkg.on_state_change(&*dkg_pool.read().unwrap());
+                let remote_dealings = change_set
+                    .iter()
+                    .filter(|change| match change {
+                        ChangeAction::AddToValidated(message) => {
+                            message.content.dkg_id.target_subnet
+                                == NiDkgTargetSubnet::Remote(target_id)
+                        }
+                        _ => false,
+                    })
```
pierugo-dfinity: nit: alternatively
```suggestion
                    .filter(|change|
                        matches!(
                            change,
                            ChangeAction::AddToValidated(message)
                                if message.content.dkg_id.target_subnet == NiDkgTargetSubnet::Remote(target_id)
                        )
                    )
```

rs/consensus/dkg/src/lib.rs
```diff
@@ -867,12 +986,40 @@ mod tests {
...
+                assert_eq!(data.dkg.transcripts_for_remote_subnets.len(), 2);
+                let mut tags = BTreeSet::new();
+                for dkg in data.dkg.transcripts_for_remote_subnets.iter() {
+                    assert!(dkg.2.is_err());
+                    assert!(tags.insert(dkg.0.dkg_tag.clone()));
+                }
+                assert_eq!(
+                    tags,
+                    BTreeSet::from([NiDkgTag::LowThreshold, NiDkgTag::HighThreshold])
+                );
```
pierugo-dfinity, line 997: Should we also assert that the target subnet is `Remote(target_id)` like you did previously?

rs/consensus/dkg/src/lib.rs
```diff
@@ -1669,11 +1757,32 @@ mod tests {
...
+                dkg_tag: NiDkgTag::LowThreshold,
+                target_subnet: NiDkgTargetSubnet::Remote(target_id),
+            },
+            NiDkgId {
+                start_block_height: Height::from(EARLY_DKG_INTERVAL + 1),
+                dealer_subnet: deps.replica_config.subnet_id,
+                dkg_tag: NiDkgTag::HighThreshold,
+                target_subnet: NiDkgTargetSubnet::Remote(target_id),
+            },
+        ];
```
pierugo-dfinity, line 1777: Just an idea. Maybe it would be clearer to get these `remote_dkg_ids` from the highest block following an `advance_round_normal_operation` (it would also be the opportunity to assert that these configs exist). Then `advance_round_normal_operation_n(EARLY_DKG_INTERVAL)` and the rest.

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: 👍

rs/consensus/dkg/src/lib.rs
```diff
@@ -2397,180 +2427,66 @@ mod tests {
...
+                            "[{desc}] Expected 1 ReshareChainKey transcript, got {}",
+                            remote_dkgs.len()
+                        );
+                        let (dkg_id, _, result) = &remote_dkgs[0];
+                        assert_eq!(
+                            dkg_id.target_subnet,
+                            NiDkgTargetSubnet::Remote(reshare_target_id),
+                            "[{desc}] transcript should be for ReshareChainKey target id"
+                        );
+                        assert!(result.is_ok(), "[{desc}]");
```
pierugo-dfinity, line 2459: nit: should we also assert on the tag like you do in the `expect_setup_initial_dkg` branch above?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -153,59 +172,44 @@ pub(crate) fn create_early_remote_transcripts(
...
+                        error!(
+                            logger,
+                            "Failed to create remote transcript config for dkg id {:?} at height {}: {}",
+                            dkg_id,
+                            parent.height.increment(),
+                            err
+                        );
+                        // Including the error in the payload will cause the context to receive
+                        // a reject response.
+                        selected_transcripts.push((dkg_id, callback_id, Err(err)));
```
pierugo-dfinity: We might include more than `MAX_EARLY_REMOTE_TRANSCRIPTS` transcripts in the same payload if we process a successful `ReshareChainKey` and then a failed `SetupInitialDKG` which contains 2 errors.

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -406,26 +410,19 @@ pub(super) fn create_summary_payload(
     logger: ReplicaLogger,
 ) -> Result<DkgSummary, DkgPayloadCreationError> {
     let (mut all_dealings, completed_dkgs) = utils::get_dkg_dealings(pool_reader, parent);
-    let mut transcripts_for_remote_subnets = BTreeMap::new();
     let mut next_transcripts = BTreeMap::new();
     // Try to create transcripts from the last round.
     for (dkg_id, config) in last_summary.configs.iter() {
-        if completed_dkgs.contains(dkg_id) {
-            // Skip DKGs that have already been completed as part of data blocks
+        if dkg_id.target_subnet != NiDkgTargetSubnet::Local {
```
pierugo-dfinity: Is
```suggestion
        if matches!(dkg_id.target_subnet, NiDkgTargetSubnet::Remote(_)) {
```
more appropriate?

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: nit: you could inline all of this. Maybe also replace `unreachable` with `panic`? It's not that unreachable imo

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -1467,138 +1064,141 @@ mod tests {
...
+                }
 
-                // Two configs are created for this remote target.
+                // The first data block after the last summary should contain repeated-failure
+                // errors for remote DKG transcripts.
+                deps.pool.advance_round_normal_operation();
+                let data_block = deps.pool.get_cache().finalized_block();
+                let dkg_data = data_block.payload.as_ref().as_data().dkg.clone();
+                assert_eq!(dkg_data.messages.len(), 0);
+                assert_eq!(dkg_data.transcripts_for_remote_subnets.len(), 3);
```
pierugo-dfinity: Though `MAX_EARLY_REMOTE_TRANSCRIPTS == 2` 👀

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Ah, this happens in a test, see my comment below

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -1467,138 +1064,141 @@ mod tests {
...
-                    .unwrap();
-
-                // No configs are created for this remote target any more.
-                assert_eq!(configs.len(), 0);
-                // No transcripts or errors are returned for this target.
-                assert_eq!(transcripts_for_remote_subnets.len(), 0);
-                // The corresponding entry is removed from the counter.
-                assert_eq!(initial_dkg_attempts.len(), 0);
+                let callback_map = build_callback_map(dkg_summary);
+                assert!(callback_map.is_empty(), "{callback_map:?}");
```
pierugo-dfinity, line 1217: Maybe to test the remaining lifecycle of the request, you could now delete the requests from the state and assert that the next summary will not contain the requests in the `initial_dkg_attempts` anymore (i.e. went from `Some(&0)` back to `None`).

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -98,6 +102,8 @@ pub fn create_payload(
 }
 
 fn create_data_payload(
+    this_subnet_id: SubnetId,
```
pierugo-dfinity: nit: Do we need `this_`? (c.f. Franz-Stefan's [comment](https://github.com/dfinity/ic/pull/9926#discussion_r3117701093))

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -281,15 +285,15 @@ pub(crate) fn create_early_remote_transcripts(
 /// 2. Among remote targets, prioritize dealings for targets that are closer to their threshold.
 /// 3. Use `target_subnet` as a tie-breaker between dealings for targets with the same remaining capacity.
 fn select_dealings_for_payload(
-    configs: &BTreeMap<NiDkgId, NiDkgConfig>,
+    configs: BTreeMap<&NiDkgId, &NiDkgConfig>,
```
pierugo-dfinity:
```suggestion
    configs: &BTreeMap<&NiDkgId, &NiDkgConfig>,
```
Do you need ownership?

rs/consensus/src/consensus/batch_delivery.rs
```diff
@@ -524,6 +524,15 @@ fn generate_dkg_response_payload(
...
+        (Some(Ok(transcript)), None) | (None, Some(Ok(transcript))) => {
+            Some(Payload::Reject(RejectContext::new(
+                RejectCode::CanisterReject,
+                format!(
+                    "Data payload contains only the {:?} transcript for SetupInitialDKG request",
+                    transcript.dkg_id.dkg_tag
+                ),
+            )))
+        }
         _ => None,
```
pierugo-dfinity:
```suggestion
        (None, None) => None,
```
nit: just for completeness

rs/tests/consensus/subnet_splitting_test.rs
pierugo-dfinity: 👍

rs/types/types/src/consensus/dkg.rs
```diff
@@ -205,7 +203,7 @@ impl DkgSummary {
                 .collect(),
             current_transcripts,
             next_transcripts,
-            transcripts_for_remote_subnets,
+            transcripts_for_remote_subnets: vec![],
```
pierugo-dfinity, line 206: Do you plan to remove this field altogether in the future?

pierugo-dfinity (commented):
Something that crosses my mind when reading
> remote `NiDkgConfig`s are no longer stored in summary blocks

Can something break on upgrading the subnet if the old replica version included a `NiDkgConfig` in the CUP's summary block? From what I can tell, it looks like we do not re-validate the CUP's block when starting up (which would fail on the new replica version) but [directly insert](https://github.com/dfinity/ic/blob/36b8b711477a52f03dc0ddfb4c59460cdcde8fec/rs/artifact_pool/src/lmdb_pool.rs#L760) it into our validated pool instead. So it appears to look fine. Also from NiDKG's PoV, it does not make a difference whether the remote config comes from the state or the summary since `merge_configs` unifies them.
But more generally, couldn't having an invalid block in our validated pool break stuff elsewhere?

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: 👍

rs/consensus/dkg/src/lib.rs
eichhorl: I introduced new functions instead, WDYT: https://github.com/dfinity/ic/pull/10055/commits/90c61ad0616a9b1bca5cce9d131c76e0d26ebd4b

rs/consensus/dkg/src/payload_builder.rs
eichhorl: I introduced new functions instead, WDYT: https://github.com/dfinity/ic/pull/10055/commits/90c61ad0616a9b1bca5cce9d131c76e0d26ebd4b

rs/consensus/dkg/src/lib.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10055/commits/2836411bc3932ae16de7b5e791dfe8bd145d048b

rs/consensus/dkg/src/lib.rs
eichhorl: Good catch, thanks! https://github.com/dfinity/ic/pull/10055/commits/9e29073fff59da3774c630d7dbdf0d1c55730fb9

rs/consensus/dkg/src/lib.rs
eichhorl: Which block do you mean? The remote DKG configs should no longer be stored in blocks.

rs/consensus/dkg/src/lib.rs
eichhorl: Good catch, thanks! https://github.com/dfinity/ic/pull/10055/commits/9e29073fff59da3774c630d7dbdf0d1c55730fb9

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Inlined here: https://github.com/dfinity/ic/pull/10055/commits/3c136fd2afd03a70f559d887498646f574920d87

Changed to panic: https://github.com/dfinity/ic/pull/10055/commits/db4284eb0be196c585074d9f5a8814abb9cc010a

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Good idea, done: https://github.com/dfinity/ic/pull/10055/commits/2c6d626571c1fb81f0c1e6762bbbaa5d840610a0

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10055/commits/b05ac4716f2c33d6a035df8d2b0eeb157bce887c

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Nope, thanks: https://github.com/dfinity/ic/pull/10055/commits/fb00e942a78a791038d5deed8dd89a73d5dca239

rs/consensus/src/consensus/batch_delivery.rs
eichhorl: Good point, done: https://github.com/dfinity/ic/pull/10055/commits/feda9aa0458fa0e173130215567f36b13af80aa3

rs/types/types/src/consensus/dkg.rs
eichhorl: I'm still on the fence about it. Technically the old code had the following optimization for setup initial DKG specifically (which is now removed):
If in the previous interval we only manage to complete one of the two transcripts, that transcript is still stored in the summary. In the following interval we then only need to complete the second transcript, and clone the previous transcript into the new payload.

We could also implement this optimization with the new code (which would then use this field to store the transcript).

However, I'm convinced that this never happened in production so far. The chance of this happening now is a bit higher (although currently dealings for both transcripts have the same priority when selecting them for the payload, so typically they would anyway complete within the same couple blocks).

However, after increasing the number of dealings per block to maybe 5 or higher, it should only take 5 blocks to complete both transcripts anyway, so the chance of only one of them being completed before the summary is reached should still be quite low. Therefore I'm not sure it's a good idea to spend extra complexity on this. So probably I will indeed remove the field at some point.

eichhorl:
> Can something break on upgrading the subnet if the old replica version included a NiDkgConfig in the CUP's summary block? From what I can tell, it looks like we do not re-validate the CUP's block when starting up (which would fail on the new replica version) but [directly insert](https://github.com/dfinity/ic/blob/36b8b711477a52f03dc0ddfb4c59460cdcde8fec/rs/artifact_pool/src/lmdb_pool.rs#L760) it into our validated pool instead. So it appears to look fine. Also from NiDKG's PoV, it does not make a difference whether the remote config comes from the state or the summary since merge_configs unifies them.
But more generally, couldn't having an invalid block in our validated pool break stuff elsewhere?

This is not something we usually pay attention to. Technically, I guess every upgrade summary is "invalid" on the new version, because it contains the wrong replica version. 

The only place where the upgrade CUP is validated again after the upgrade should be when it is fetched by an orchestrator of a different node that is still on the old version. But in that case we only validate the threshold signature. There, what's more important is that the new replica still returns the exact same bytes that the CUP was created with on the old version, which is why we store the original bytes separately in the pool. Otherwise the byte representation could change if an old CUP was deserialized and serialized again using the new version.

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Good point, probably including more errors doesn't matter as much as including more transcripts. But then the behavior is still inconsistent in the case where we include the errors in an earlier iteration (because then they will count towards the limit). Should be fixed here: https://github.com/dfinity/ic/pull/10055/changes/764e2855bae510779be993d96ae110fe61819e31

rs/consensus/dkg/src/payload_builder.rs
eichhorl: Thanks, I updated the test: https://github.com/dfinity/ic/pull/10055/changes/764e2855bae510779be993d96ae110fe61819e31

rs/consensus/dkg/src/lib.rs
```diff
@@ -183,7 +195,7 @@ impl DkgImpl {
     fn validate_dealings_for_dealer(
         &self,
         dkg_pool: &dyn DkgPool,
-        configs: &BTreeMap<NiDkgId, NiDkgConfig>,
+        configs: &BTreeMap<&NiDkgId, &NiDkgConfig>,
         dkg_start_height: Height,
         messages: Vec<&Message>,
```
pierugo-dfinity: I do not think we need ownership here (which saves a `.to_vec()` in the caller).

Sorry, not your changes.

rs/consensus/dkg/src/lib.rs
```diff
@@ -211,9 +223,13 @@ impl DkgImpl {
...
 
         // If the dealing refers a config which is not among the ongoing DKGs,
-        // we reject it.
+        // we reject it, unless it is a remote DKG, in which case we defer it
+        // until the request appears in the state, or the dealing is purged.
         let config = match configs.get(message_dkg_id) {
             Some(config) => config,
+            None if message_dkg_id.target_subnet.is_remote() => {
+                return Mutations::new();
+            }
```
pierugo-dfinity, line 232: Not a blocker since I don't think it would be really significant.

If an honest peer sends a remote dealing for a config in the future, it could be that we would not validate any of the peer's dealings until the request appears in the state. This is because we select only one dealing at the beginning of the function, and since the messages are ordered, the chosen one will always be the same one if it is "small" enough.

rs/consensus/dkg/src/lib.rs
pierugo-dfinity:
```suggestion
            .into_values()
```

Sorry, not your changes.

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: Right, of course, I think for some reason I thought configs would be stored in data blocks

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: Hmmm this actually applies to all `return Mutations::new();` in this function...

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: That's great thanks!

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Nice. A bit unfortunate to do the check twice, but couldn't think of a cleaner alternative

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: That's great thanks!

rs/types/types/src/consensus/dkg.rs
pierugo-dfinity: > Therefore I'm not sure it's a good idea to spend extra complexity on this.

I think I agree indeed.

> So probably I will indeed remove the field at some point.

Sounds good

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -406,28 +418,21 @@ pub(super) fn create_summary_payload(
...
     let (mut all_dealings, completed_dkgs) = utils::get_dkg_dealings(pool_reader, parent);
-    let mut transcripts_for_remote_subnets = BTreeMap::new();
     let mut next_transcripts = BTreeMap::new();
     // Try to create transcripts from the last round.
     for (dkg_id, config) in last_summary.configs.iter() {
-        if completed_dkgs.contains(dkg_id) {
-            // Skip DKGs that have already been completed as part of data blocks
+        if dkg_id.target_subnet.is_remote() {
+            // Skip remote DKGs
             continue;
```
pierugo-dfinity, line 426: Technically, this could be removed in the future, correct? Maybe not right now, as I mentioned in my comment that the old replica version could still include a remote config in the summary, but otherwise, we would not expect any remote configs in a summary, right? Not asking to actually do it, just to understand.

pierugo-dfinity (approved):
Nice! 🚀 🚀

pierugo-dfinity:
> every upgrade summary is "invalid" on the new version, because it contains the wrong replica version.

Good point. Then, this addresses my concern.

> There, what's more important is that the new replica still returns the exact same bytes that the CUP was created with

Good to know, right, thanks!

rs/consensus/dkg/src/lib.rs
eichhorl: To me it looks like we group and select the dealings using the key `(dealer ID, DKG ID)`. So I think what you're saying is true for dealings from the same dealer for the same DKG (of which there should anyway only be one). But we would still validate dealings from that dealer for different DKGs. I clarified the comment here: https://github.com/dfinity/ic/pull/10055/changes/6c57e9b17338041d119837843f202ec80a6c82e5

rs/consensus/dkg/src/lib.rs
eichhorl: Cool, thanks: https://github.com/dfinity/ic/pull/10055/changes/272d542053206c99b1b0cc52d0ee23ef44f6ac04

rs/consensus/dkg/src/lib.rs
eichhorl: Cool, thanks: https://github.com/dfinity/ic/pull/10055/changes/272d542053206c99b1b0cc52d0ee23ef44f6ac04

rs/consensus/dkg/src/payload_builder.rs
eichhorl: I agree, this could be removed in the future

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: Right. I was confused by the name of the function `validate_dealings_for_dealer`. That works then, thanks!

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Thanks

PR #10058, opened 2026-04-29
feat: Move LUKS header to a detached file

frankdavid:
Move the store LUKS header to a detached header file on the private var partition. The private partition is encrypted which provides additional protection for the LUKS header.

For backwards compatibility, we don't remove the attached header. This is necessary, so we can still downgrade / rollback to previous versions.

Flow:
- During the upgrade flow, we exchange the LUKS header and the upgrade VM writes the header to var.
- When upgrading from the current version (no LUKS header sent), the implementation takes the attached header and backs it up as a detached header on the var partition.
- When opening the disk, we always use the detached header on SEV nodes / store device.
- When formatting the disk, we create both an attached and a detached header. This is necessary such that both upgrades and downgrades work. All changes (if any) are made to the detached header.


rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -37,19 +36,29 @@ mod tls;
 const NNS_PUBLIC_KEY_PATH: &str = "/run/config/nns_public_key.pem";
 
 type ServiceClientType = DiskEncryptionKeyExchangeServiceClient<Channel>;
+
+/// (store_device, previous_key_path, store_luks_header_path, sev_firmware) -> success
 pub type CanOpenStore =
```
blind-oracle: nit: move boxed funcs to a Trait

rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -188,10 +207,46 @@ impl DiskEncryptionKeyExchangeClientAgent {
             .key
             .context("GetKeyResponse does not contain a key")?;
 
+        self.persist_disk_encryption_artifacts(
+            disk_encryption_key,
+            // This can be None when upgrading from older GuestOS-s that do not populate this field.
+            // TODO: Error on None once all GuestOS-s support detached headers
+            get_key_response.luks_header,
```
blind-oracle: nit: maybe rename `get_key_response` and relevant stuff like `get_disk_encryption_key()` method to account for that it now also sends a LUKS header?

I dunno something more generic like `get_disk_encryption_data()` or similar.

rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -188,10 +207,46 @@ impl DiskEncryptionKeyExchangeClientAgent {
...
+        )?;
+
+        Ok(())
+    }
+
+    fn persist_disk_encryption_artifacts(
```
blind-oracle: use Tokio IO wrappers here and everywhere else to avoid blocking the main thread pool

rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -188,10 +207,46 @@ impl DiskEncryptionKeyExchangeClientAgent {
...
+                })?;
+            }
+            None => {
+                println!(
+                    "GetKeyResponse does not contain a store LUKS header; recovering it locally from {}",
+                    self.store_device_path.display()
+                );
+                (self.backup_luks_header)(&self.store_device_path, &self.store_luks_header_path)
+                    .context(
+                    "GetKeyResponse does not contain a store LUKS header and local backup failed",
```
blind-oracle: nit: we can drop the `GetKeyResponse does not contain a store LUKS header` probably since it's anyway printed above?

rs/ic_os/guest_upgrade/server/src/service.rs
```diff
@@ -144,18 +153,28 @@ impl DiskEncryptionKeyExchangeServiceImpl {
         let mut sev_firmware = self.sev_firmware_factory.deref()()
             .map_err(|e| Status::internal(format!("Failed to create SEV firmware: {e:?}")))?;
 
+        let luks_header = if self.send_luks_header {
+            let detached_header_path = &self.store_luks_header_path;
+            Some(std::fs::read(detached_header_path).map_err(|e| {
```
blind-oracle: Use Tokio IO wrappers here too

rs/ic_os/guest_upgrade/server/src/service.rs
```diff
@@ -144,18 +153,28 @@ impl DiskEncryptionKeyExchangeServiceImpl {
         let mut sev_firmware = self.sev_firmware_factory.deref()()
             .map_err(|e| Status::internal(format!("Failed to create SEV firmware: {e:?}")))?;
 
+        let luks_header = if self.send_luks_header {
+            let detached_header_path = &self.store_luks_header_path;
```
blind-oracle: nit: why the need for the variable? To pass ref? I guess if we remove lifetimes on structs it won't be needed

rs/ic_os/guest_upgrade/tests/src/lib.rs
```diff
@@ -151,6 +162,10 @@ impl DiskEncryptionKeyExchangeTestFixture {
         };
 
         let previous_key = NamedTempFile::new().unwrap();
+        let server_store = NamedTempFile::new().unwrap();
+        let server_store_luks_header = NamedTempFile::new().unwrap();
+        let client_store_luks_header = NamedTempFile::new().unwrap();
+        std::fs::write(server_store_luks_header.path(), b"store LUKS header bytes").unwrap();
         // Increment port for each test case so tests can run in parallel
         let server_port = FREE_PORT.fetch_add(1, Ordering::Relaxed);
```
blind-oracle, line 171: nit: ideally you might want to listen on port 0 - that will assign a random free port w/o conflicts. But you'll need to change API a bit to spin up a listener first and extract assigned port from it.

rs/ic_os/guest_upgrade/tests/src/lib.rs
```diff
@@ -225,29 +245,62 @@ impl DiskEncryptionKeyExchangeTestFixture {
         vsock_client: MockVSockClient,
     ) -> DiskEncryptionKeyExchangeServerAgent {
         let server_sev_firmware = self.server_sev_firmware.clone();
-        DiskEncryptionKeyExchangeServerAgent::new_for_testing(
+        DiskEncryptionKeyExchangeServerAgent::new(
             tokio::runtime::Handle::current(),
             Box::new(vsock_client),
             Arc::new(move || Ok(Box::new(server_sev_firmware.clone()))),
             SevRootCertificateVerification::TestOnlySkipVerification,
```
blind-oracle, line 253: nit: not relevant to this PR, but should we gate this enum variant with `#[cfg(test]]` then maybe for a stronger guarantee?

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -37,31 +40,63 @@ pub(crate) struct KeyslotParameters {
     pub(crate) key_size: Option<usize>,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum LuksHeaderLocation<'a> {
```
blind-oracle, line 44: nit: do we really need refs & lifetimes here? I guess we can just clone & own a `PathBuf` for simplicity.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -37,31 +40,63 @@ pub(crate) struct KeyslotParameters {
...
+    /// Use the attached LUKS header on the device.
+    Attached,
+    /// Use the detached LUKS header at the specified path.
+    Detached(&'a Path),
+}
+
+/// Initializes a cryptsetup handle for `device_path`.
+/// `header_location` selects whether the LUKS header is read from the device itself or from a
+/// detached header file while `device_path` remains the data device.
+fn init_crypt_device(
```
blind-oracle: nit: I don't like the `init_` prefix (probably comes from `libcryptsetup_rs` terminology) - this makes me think it wipes/formats it. Maybe rename to at least `init_crypt_device_handle` to be specific, or `obtain_crypt_device_handle`.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -132,27 +177,30 @@ pub fn format_crypt_device(device_path: &Path, encryption_key: &[u8]) -> Result<
     Ok(crypt_device)
 }
 
-/// Opens a LUKS2 device at the specified path and loads its context.
-fn open_luks2_device(device_path: &Path, verify_luks_params: bool) -> Result<CryptDevice> {
-    let mut crypt_device =
-        CryptInit::init(device_path).context("Failed to initialize cryptographic device")?;
+/// Opens a LUKS2 device at the specified path and loads its context. Does not activate the device.
+pub fn open_luks2_device(
```
blind-oracle, line 182: Why we don't verify LUKS params anymore here?

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -162,6 +210,41 @@ pub fn check_encryption_key(device_path: &Path, encryption_key: &[u8]) -> Result
...
+        .parent()
+        .context("LUKS header path does not have a parent directory")?;
+    fs::create_dir_all(parent_dir).with_context(|| {
+        format!(
+            "Failed to create parent directory for LUKS header {}",
+            header_path.display()
+        )
+    })?;
+
+    let temp_dir = tempfile::tempdir_in(parent_dir)
```
blind-oracle, line 227: What's the rationale for using the temporary dir? So that we don't overwrite existing header unless we really were able to get the new one?

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -1,23 +1,55 @@
...
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, bail};
 use config_types::GuestVMType;
 use sev_guest::firmware::SevGuestFirmware;
 use sev_guest::key_deriver::{Key, derive_key_from_sev_measurement};
-use std::fs;
 use std::path::Path;
 use tracing::{info, warn};
 
 pub struct SevDiskEncryption<'a> {
```
blind-oracle: nit: not relevant to this PR: like above I'd drop the lifetimes everywhere and just own the `PathBuf`, easier to use and we don't really save anything here performance-wise anyway.

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -138,14 +173,31 @@ impl DiskEncryption for SevDiskEncryption<'_> {
         Ok(())
     }
 
-    fn format(&mut self, device_path: &Path, _partition: Partition) -> Result<()> {
+    fn format(&mut self, device_path: &Path, partition: Partition) -> Result<()> {
+        if partition == Partition::Store && self.store_luks_header_path.exists() {
+            bail!(
```
blind-oracle, line 178: Is this error actionable by any means?

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -156,17 +208,31 @@ impl DiskEncryption for SevDiskEncryption<'_> {
 pub fn can_open_store(
     device_path: &Path,
     previous_key_path: &Path,
+    store_luks_header_path: &Path,
     sev_firmware: &mut dyn SevGuestFirmware,
 ) -> Result<bool> {
-    // The logic should be kept consistent with open above
+    // Keep key selection consistent with open() above. Unlike open(), this check must not create
+    // or refresh the detached header, so it probes the detached header first and then falls back
+    // to the attached header.
```
blind-oracle: Where does the fall back to the attached header happens? In this function I see only `Detached`

blind-oracle (changes requested):

rs/ic_os/os_tools/guest_disk/src/sev.rs
frankdavid: Makes sense, fixed.

rs/ic_os/os_tools/guest_disk/src/sev.rs
frankdavid: I added it as a sanity check, we should not format store when the header exists.

rs/ic_os/os_tools/guest_disk/src/sev.rs
frankdavid: I think I had updated this comment but then AI rewrote it again incorrectly. There is no fallback in the open path. We always read from the detached header. The detached header is created on the upgrade path. In addition, when we format the disk, we create both an attached and a detached header.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: Done.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: This is only used to pass it as a function arg (a wrapper around &Path). We never need an owned variant, so here a reference makes sense because callers can just pass LuksHeaderLocation::Detached(...) where ... can come from PathBuf (`&variable`) or &Path (`variable`) without having to clone.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: I didn't want to store garbage if header_backup fails midway for whatever reason. If it does, the header won't be created. Probably not a big gain, I can simplify it if you prefer that.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: There was exactly one place that passed `verify_luks_params=true` which was `activate_luks2_device`.

`activate_luks2_device` also extracts the params for logging so the extraction logic was duplicated. It made more sense to do the verification only in `activate_luks2_device` so the extraction is done once.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
blind-oracle: I just think that having this redundant lifetime marker `<'_>` everywhere is a bit confusing. Easier to clone, but up to you.

rs/ic_os/guest_upgrade/tests/src/lib.rs
frankdavid: I don't think t's worth the effort to change the prod code just to support this use case. In prod, we want a fixed port anyway and just assigning ports naively in the test has worked well - if we ever find issues with port assignment, we can always change the API.

rs/ic_os/guest_upgrade/tests/src/lib.rs
frankdavid: `#[cfg(test]]` doesn't work across crate-boundaries. So if I'm testing crate `A` which uses `A_dep`, I'll get `#[cfg(test]]` behavior from crate `A` but not from `A_dep`.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: I checked and we don't need lifetimes at all, `LuksHeaderLocation` is always passed/used without any `<'_>` or `<...>`, it's all implicit.

blind-oracle (approved):

PR #10062, opened 2026-04-29
feat: CON-1445 Add a function to update the number of ongoing remote DKG attempts

eichhorl:
This PR introduces a function which will be used to generate the map tracking the number of ongoing remote DKG attempts. This map is included and maintained by the DKG summary payload. In particular:
- We increment the number of attempts of all request contexts that are still open when the summary is being built.
- Attempt counters for request contexts that are no longer part of the state are removed.
- Remote DKG requests that were completed in the past interval are tracked in the new summary with their attempt counter set to 0. This is to avoid handling the same context again in the following interval.

Note that the function is currently still unused.

### DIFF
--- a/rs/consensus/dkg/src/remote.rs
+++ b/rs/consensus/dkg/src/remote.rs
@@ -8,7 +8,9 @@ use ic_interfaces_registry::RegistryClient;
 use ic_logger::ReplicaLogger;
 use ic_replicated_state::{
     ReplicatedState,
-    metadata_state::subnet_call_context_manager::{ReshareChainKeyContext, SetupInitialDkgContext},
+    metadata_state::subnet_call_context_manager::{
+        ReshareChainKeyContext, SetupInitialDkgContext, SubnetCallContextManager,
+    },
 };
 use ic_types::{
     Height, RegistryVersion, SubnetId,
@@ -19,7 +21,7 @@ use ic_types::{
     },
     messages::CallbackId,
 };
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 /// The result of creating DKG configs for a given request context.
 pub(crate) type ConfigResult = Result<Vec<NiDkgConfig>, Vec<(NiDkgId, String)>>;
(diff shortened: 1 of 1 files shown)


rs/consensus/dkg/src/remote.rs
```diff
@@ -220,6 +228,36 @@ pub(crate) fn merge_configs<'a>(
...
 
+/// Update the remote DKG attempts for the given target IDs based on the previous attempts and the
+/// newly completed target IDs.
+pub(crate) fn get_updated_remote_dkg_attempts(
+    summary: &DkgSummary,
+    state: &ReplicatedState,
+    completed_target_ids: BTreeSet<NiDkgTargetId>,
+) -> BTreeMap<NiDkgTargetId, u32> {
+    let call_contexts = &state.metadata.subnet_call_context_manager;
+    let previous_remote_dkg_attempts = &summary.initial_dkg_attempts;
```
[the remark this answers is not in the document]
eichhorl: Updated PR description

rs/consensus/dkg/src/remote.rs
```diff
@@ -220,6 +228,36 @@ pub(crate) fn merge_configs<'a>(
...
+
+    // Iterate over all remote DKG contexts
+    for (_, context) in remote_dkg_contexts(call_contexts) {
+        let target_id = *context.target_id();
+        let attempts = previous_remote_dkg_attempts
+            .get(&target_id)
+            .copied()
+            .unwrap_or(0)
+            .saturating_add(1);
+        remote_dkg_attempts.insert(target_id, attempts);
```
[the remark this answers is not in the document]
eichhorl: This is by design, otherwise a context with a faulty registry version might never time out

rs/consensus/dkg/src/remote.rs
```diff
@@ -220,6 +228,36 @@ pub(crate) fn merge_configs<'a>(
     merged_configs
 }
 
+/// Update the remote DKG attempts for the given target IDs based on the previous attempts and the
+/// newly completed target IDs.
+pub(crate) fn get_updated_remote_dkg_attempts(
+    summary: &DkgSummary,
+    state: &ReplicatedState,
+    completed_target_ids: BTreeSet<NiDkgTargetId>,
+) -> BTreeMap<NiDkgTargetId, u32> {
```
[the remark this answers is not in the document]
eichhorl: Target IDs that were passed in via this map are moved into the return value

rs/consensus/dkg/src/remote.rs
```diff
@@ -220,6 +228,36 @@ pub(crate) fn merge_configs<'a>(
...
+    let mut remote_dkg_attempts = BTreeMap::new();
+
+    // Iterate over all remote DKG contexts
+    for (_, context) in remote_dkg_contexts(call_contexts) {
+        let target_id = *context.target_id();
+        let attempts = previous_remote_dkg_attempts
+            .get(&target_id)
+            .copied()
+            .unwrap_or(0)
+            .saturating_add(1);
```
pierugo-dfinity: Is there a possibility that there is a `callback_id` for which the previous summary's `initial_dkg_attempts` is 0 (in the case where it completed in a data block during the previous interval), while also `completed_target_ids` does not contain the `callback_id` (because there were no dealings for it in the newest interval)? In that scenario, we would increment from 0 to 1 and start generating new dealings & transcript for that `target_id` no?

eichhorl: If the previous summary's `initial_dkg_attempts` is 0, then that means we delivered a response to DSM at some point in the previous interval. This response will remove the request context. The assumption is that a certified state without the request context will appear before we reach the next summary block (meaning we would not increment the attempts from 0 to 1). This assumption should hold, because the notarized-certified gap is capped to 70 blocks.

pierugo-dfinity: What about when the DKG interval is smaller than 70 like in system tests?

eichhorl: Ideally after this feature we don't need to reduce the DKG interval in system tests doing remote DKG anymore. But even if we did handle the context a second time, it wouldn't really be a problem, because the second response will simply be ignored by DSM if the context doesn't exist anymore.

That being said I agree it would be cleaner to simply carry over entries with 0 as long as the context still exists in the state (instead of incrementing them). Implemented here: https://github.com/dfinity/ic/pull/10062/commits/b7ad52ea58f56d63ab609476d8d8984cefb24a81

Hopefully we can later clean this up more with your suggestion of introducing a new type.

pierugo-dfinity: Nice looks great thanks!

> Hopefully we can later clean this up more with your suggestion of introducing a new type.

👍

pierugo-dfinity (approved):

PR #4473, opened 2025-03-24
test: [CON-1144] modify `cup_compatibility_test` to also test the compatibility against the version deployed on the `io67a` subnet

kpop-dfinity:
Currently, we only test against the version deployed on the NNS subnet, which is the oldest deployed replica version.
With this change, we will also test against the newest deployed replica version, i.e., the one running on the `io67a` subnet

PR #10063, opened 2026-04-29
test(upgrade): [CON-1144] more exhaustive upgrade system tests

pierugo-dfinity:
This PR refactors a bit the upgrade system tests to be more exhaustive. Here are the proposed changes:
- Currently, we upgrade and downgrade app and NNS subnets from the oldest version on mainnet (`mainnet_nns`) to the branch back and forth. This means that when a release cut is decided, it has not been tested against the version that will be deployed once the new release is rolled out. Thus, we replace `mainnet_nns` with `mainnet_app` to test the branch against the latest version on mainnet.
  - Though, imagine the NNS (or any subnet running the same version as the NNS) requires a hot-fix. We would make a new release cut and upgrade it directly. Here, we would also need to test the upgradability of `mainnet_nns` with the branch. Effectively, it is more complete to test upgrading from and to the latest *and* the oldest version in mainnet, and this PR does exactly that.
- This PR also upgrades & downgrades unassigned nodes instead of only upgrading, also with `mainnet_nns` and `mainnet_app` for the same reason as above. It is also much more simplified and reads more easily.
- Finally, this PR removes the system tests `upgrade_app_subnet_test` and `downgrade_app_subnet_test` since `upgrade_downgrade_app_subnet_test` should already cover them. Though, they had a difference which was running a chain key signature test after upgrading. I figured that this was not necessary either since `upgrade_downgrade_app_subnet_test` already runs a workload of chain key signatures for the entire duration of the system test but let me know if this is indeed fine.


rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
```diff
@@ -0,0 +1,103 @@
...
+    InternetComputer::new()
+        .add_fast_single_node_subnet(SubnetType::System)
+        .with_unassigned_nodes(1)
+        .setup_and_start(&env)
+        .expect("failed to setup IC under test");
+
+    install_nns_and_check_progress(env.topology_snapshot());
+}
+
+fn upgrade_assigned_nodes(
```
eichhorl:
```suggestion
fn upgrade_unassigned_nodes(
```

rs/tests/consensus/upgrade/common.rs
```diff
@@ -251,12 +250,6 @@ pub fn upgrade(
     ));
     info!(logger, "Could store and read message '{}'", msg_2);
 
-    if let Some((canister, public_keys)) = ecdsa_canister_key {
-        for (key_id, public_key) in public_keys {
-            run_chain_key_signature_test(canister, &logger, key_id, public_key.clone());
-        }
-    }
```
eichhorl, line 258: The motivation between the signature workload and the `run_chain_key_signature_test` is slightly different: The former is to ensure that the subnet can still upgrade even under very high signature load. The latter is to ensure that signing still works and the public key doesn't change during the upgrade.

We can't really assert that signing works while there is also a workload going on, because signing might fail simply because the queue is completely filled due to the workload. maybe it would be enough to only check that the public key didn't change?

eichhorl (approved):
Nice! I think the one way upgrade tests used to exist because they were faster then the roundtrip tests, and could be run on every PR. But I see that is no longer the case anyway, so I agree we can remove them.

We should also (not this PR) extend the `cup_compatibility_test` to check the compatibility between more versions, similar to how the `queues_compatibility_test` does it.

rs/tests/consensus/upgrade/common.rs
pierugo-dfinity: Right, then I put that check back and replaced it with just checking that the public key hasn't changed.

rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
pierugo-dfinity: Thanks

rs/tests/consensus/upgrade/common.rs
eichhorl: Looks good, thanks!

pierugo-dfinity:
> extend the cup_compatibility_test to check the compatibility between more versions

@eichhorl I just checked that test and am not sure which other pair of versions you would test? It looks like `queues_compatibility_test` tests the same:
- Serialize with mainnet & deserialize with branch
- Serialize with branch & deserialize with mainnet

Repeat for both the latest and oldest mainnet versions.

eichhorl:
> > extend the cup_compatibility_test to check the compatibility between more versions
> 
> @eichhorl I just checked that test and am not sure which other pair of versions you would test? It looks like `queues_compatibility_test` tests the same:
> 
> * Serialize with mainnet & deserialize with branch
> * Serialize with branch & deserialize with mainnet
> 
> Repeat for both the latest and oldest mainnet versions.

Ah, seems like this was already done in https://github.com/dfinity/ic/pull/4473. Even better!

PR #10065, opened 2026-04-30
chore: account for CREATE_EXECUTION_STATE_BASE_COST

mraszyk:
**Charge a fixed base cost for create_execution_state**
                                                                                                                                                
Canister installation and loading canister snapshot call `Hypervisor::create_execution_state` which has a constant ~10 ms overhead independent of module size. Previously only the per-instruction Wasm compilation cost was charged, leaving this fixed overhead unaccounted.

This PR introduces `CREATE_EXECUTION_STATE_BASE_COST` (20M instructions, calibrated via benchmark) and adds it on top of the existing per-byte compilation cost every time `Hypervisor::create_execution_state` is called. The constant is configurable via `EmbeddersConfig` and defaults are validated in updated tests.


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -2472,7 +2473,7 @@ fn failed_upgrade_hooks_consume_instructions() {
             + if fails_before_compiling_upgrade_wasm {
                 NumInstructions::new(0)
             } else {
-                compilation_cost
+                DEFAULT_CREATE_EXECUTION_STATE_BASE_COST + compilation_cost
             };
```
alin-at-dfinity, line 2480: Just wondering, no clue: does the 10 ms overhead not apply in the failed upgrade case?

rs/execution_environment/src/execution/install_code/tests.rs
```diff
@@ -74,6 +74,7 @@ fn dts_resume_works_in_install_code() {
     let mut test = ExecutionTestBuilder::new()
         .with_install_code_instruction_limit(INSTRUCTION_LIMIT)
         .with_install_code_slice_instruction_limit(1_000)
+        .with_create_execution_state_base_cost(0)
```
alin-at-dfinity, line 77: Just wondering (again, no idea whether it would result in more changes or fewer) whether it would make sense for the base cost to default to zero. Maybe not.

Otherwise, something I did for the similar `SchedulerConfig` was to define a `zero_instruction_overhead_config()` function, so I could write something like `SchedulerConfig { scheduler_cores: 2, ..zero_instruction_overhead_config() }`. Although that won't work with a builder. And this appears to be the only field being reset to zero. So maybe also not applicable.

rs/execution_environment/src/hypervisor.rs
```diff
@@ -146,12 +147,14 @@ impl Hypervisor {
             Ok(size) => std::cmp::max(size, canister_module.len()),
             Err(_) => canister_module.len(),
         };
+        let mut total_cost = self.create_execution_state_base_cost;
```
alin-at-dfinity: Ultimate nitpicking: I don't think that defining `total_cost` up top helps readability. I'd find it much easier to parse something like:
```rust
let total_cost = self.create_execution_state_base_cost + compilation_cost;
round_limits.instructions -= as_round_instructions(total_cost);
(total_cost, Err(err))
```

alin-at-dfinity (approved):
LGTM, just a couple of questions and a nitpick that you can ignore.

rs/execution_environment/src/canister_manager/tests.rs
mraszyk: `fails_before_compiling_upgrade_wasm` indicates whether the upgrade fails during the pre-upgrade hook (before the upgrade WASM is ever compiled), as opposed to failing during or after compilation of the upgrade WASM; hence it makes sense that no overhead applies in such a case

rs/execution_environment/src/execution/install_code/tests.rs
mraszyk: > whether it would make sense for the base cost to default to zero

I think the base cost should match the production cost to reduce the risk of production misconfiguration (and for the sake of consistency with other parameter defaults).

The alternative (dedicated function producing zeroed config) might not be appropriate here either since we should by default test with the production config (as much as possible) unless we know it does not matter and override it explicitly. In your case, it sounds like this assumption was true for an entire class of tests so it was maybe more helpful.

PR #10070, opened 2026-04-30
fix: CON-1651 Disallow PUT/DELETE for flexible outcalls unless `min_responses` = `max_responses` = `total_requests`

eichhorl:
Currently, PUT and DELETE requests are allowed for non-replicated outcalls, but not for traditional HTTP outcalls. Similarly, this PR disallows PUT and DELETE for flexible outcalls, if the number of returned responses is potentially non-deterministic, depending on the request parameters. This is to avoid confusing race conditions that may occur.

For flexible outcalls, a deterministic number of responses means that all delegated requests must complete before a response is delivered to the canister, i.e. min_responses == max_responses == total_requests.

Otherwise, if first a flexible DELETE outcall with min_responses < total_requests is made for a resource R, directly followed by a PUT or POST outcall for R, then it may happen that R is actually deleted after the PUT/POST outcall has finished, because the IC does not necessarily wait for all outcalls to complete before a result is delivered back to the canister: The IC only waits for sufficient calls to complete to reach consensus on the result.

### DIFF
--- a/rs/types/types/src/canister_http.rs
+++ b/rs/types/types/src/canister_http.rs
@@ -1715,18 +1715,19 @@ mod tests {
     }
 
     #[rstest]
-    #[case(HttpMethod::PUT, 2, 3, false)]
-    #[case(HttpMethod::DELETE, 2, 3, false)]
-    #[case(HttpMethod::PUT, 3, 3, true)]
-    #[case(HttpMethod::DELETE, 3, 3, true)]
-    #[case(HttpMethod::PUT, 2, 2, false)]
+    #[case(HttpMethod::PUT, 3, 2, 3, false)]
+    #[case(HttpMethod::DELETE, 3, 2, 3, false)]
+    #[case(HttpMethod::PUT, 3, 3, 3, true)]
+    #[case(HttpMethod::DELETE, 3, 3, 3, true)]
+    #[case(HttpMethod::PUT, 3, 2, 2, false)]
+    #[case(HttpMethod::DELETE, 3, 2, 2, false)]
     fn flexible_methods_require_deterministic_response_counts(
         #[case] method: HttpMethod,
+        #[case] total_requests: u32,
         #[case] min_responses: u32,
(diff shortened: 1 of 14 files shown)


rs/types/types/src/canister_http.rs
```diff
@@ -1732,6 +1714,59 @@ mod tests {
         );
     }
 
+    #[rstest]
+    #[case(HttpMethod::PUT, 2, 3, false)]
+    #[case(HttpMethod::DELETE, 2, 3, false)]
+    #[case(HttpMethod::PUT, 3, 3, true)]
+    #[case(HttpMethod::DELETE, 3, 3, true)]
+    #[case(HttpMethod::PUT, 2, 2, false)]
```
pierugo-dfinity: nit: maybe also `#[case(HttpMethod::DELETE, 2, 2, false)]`

rs/types/types/src/canister_http.rs
```diff
@@ -1732,6 +1714,59 @@ mod tests {
...
+    fn flexible_methods_require_deterministic_response_counts(
+        #[case] method: HttpMethod,
+        #[case] min_responses: u32,
+        #[case] max_responses: u32,
+        #[case] should_succeed: bool,
+    ) {
+        let total_requests = 3;
```
pierugo-dfinity: nit: maybe you could move the total requests to an argument of the function. Even though it's always the same, I think it's clearer to have it defined close to the minimum and maximum responses.

pierugo-dfinity (approved):

rs/types/types/src/canister_http.rs
eichhorl: Done 👍

rs/types/types/src/canister_http.rs
eichhorl: Done 👍

schneiderstefan (approved):
I like the rstest syntax. I'll have to clean up some old tests of mine when I have a chance.

PR #10071, opened 2026-04-30
fix: Treat certifications like other in-memory states to keep around when pruning. Reverts the revert of async hashing.

michael-weigelt:
When pruning states below a certain height, we keep around some special heights anyway (such as the `tip_height`). This must be true for both `certification_metadata` and `certifications`. 

This also reverts the revert of https://github.com/dfinity/ic/pull/8464


basvandijk (approved):

rs/state_manager/tests/state_manager.rs
```diff
@@ -8985,26 +9035,48 @@ fn commit_and_certify_optimization_conditions() {
...
+        assert_eq!(no_state_clone_count(metrics), expected_no_state_clone_count);
 
-        // height of 21 is not less than `fast_forward_height` => optimization does not trigger
-        assert_eq!(sm.fast_forward_height(), 21);
+        // height of 11 is not less than `fast_forward_height` => optimization does not trigger
+        let height = 11;
+        assert_eq!(sm.fast_forward_height(), height);
         let state = sm.take_tip().1;
-        sm.commit_and_certify_at_height(state, Height::new(21), CertificationScope::Metadata, None);
+        sm.commit_and_certify_at_height(
```
mraszyk, line 9099: We should still
```
sm.deliver_state_certification(fake_certification_for_height(Height::new(height)));
```
so that the optimization does not trigger because of fast_forward_height only.

rs/state_manager/src/lib.rs
```diff
@@ -3329,12 +3292,64 @@ impl StateManager for StateManagerImpl {
...
+                certification.certified_state_hash.clone()
+            } else {
+                // Wait for the hashing thread.
+                let (sender, recv) = bounded(1);
+                self.hash_channel
+                    .send(HashRequest::Wait { sender })
+                    .expect("Failed to send `Wait` to hash channel");
+                recv.recv().expect("Failed to wait for hash channel");
+                // After awaiting the hashing thread, snapshot and certification_metadata
+                // must have an entry at height-1, so we can unwrap.
```
mraszyk:
```suggestion
                // must have an entry at prev_height, so we can unwrap.
```

rs/state_manager/src/lib.rs
```diff
@@ -3329,12 +3292,64 @@ impl StateManager for StateManagerImpl {
...
+                let states = self.states.read();
+                if let Some(cert_md) = states.certifications_metadata.get(&prev_height) {
+                    cert_md.certified_state_hash.clone()
+                } else {
+                    fatal!(
+                        self.log,
+                        "Previous state hash was not available after awaiting the hash thread. This is a bug."
+                    );
```
mraszyk, line 3366: Instead of `fatal`, we could still do
```
let certification = StateManagerImpl::compute_certification_metadata(
                    state,
                    prev_height,
                    &self.metrics,
                    &self.log,
                )
                .unwrap_or_else(|err| fatal!(self.log, "Failed to compute hash tree: {:?}", err));
certification.certified_state_hash.clone()
```
and raise a critical error. Let's also add `debug_assert!(false)`.

rs/state_manager/src/lib.rs
```diff
@@ -3595,6 +3561,179 @@ impl StateManager for StateManagerImpl {
...
+                }
+            })
+            .unwrap(),
+    );
+    (handle, hash_req_sender)
+}
+
+impl StateManagerImpl {
+    /// After this method terminates, both `SharedState.snapshots` and `SharedState.certification_metadata`
+    /// at the height from the previous `commit_and_certify` are populated. It also updates `latest_state_height`
```
mraszyk, line 3740: We could also mention that `latest_certified_state_height` might also be updated (I wouldn't elaborate on the exact conditions though).

rs/state_manager/src/lib.rs
```diff
@@ -3595,6 +3561,179 @@ impl StateManager for StateManagerImpl {
...
+                                    let now = Instant::now();
+                                    metrics
+                                        .height_update_time_seconds
+                                        .observe((now - *last_height_update_time).as_secs_f64());
+                                    *last_height_update_time = now;
+                                }
+                                // Only update the certified height and notify the channel if the
+                                // certification is actually being stored. We must not fire the channel
+                                // when the snapshot already existed (e.g., due to state sync), because
+                                // in that case `certification_metadata` is never inserted and the
```
mraszyk:
```suggestion
                                // in that case `certification_metadata` is not updated and the
```

rs/state_manager/src/lib.rs
```diff
@@ -2331,6 +2331,11 @@ impl StateManagerImpl {
             .set(latest_certified_height.get() as i64);
 
         let mut certifications = states.certifications.split_off(&last_height_to_keep);
+        for h in inmemory_heights_to_keep.iter() {
+            if let Some(cert) = states.certifications.remove(h) {
+                certifications.insert(*h, cert);
+            }
+        }
```
mraszyk, line 2357: Could we have a test for this?
- set fast forward height to 42
- deliver certification at height 1 
- commit and certify at height 1 (should trigger the optimization - assert that)
- take_tip
- push state at height 2, updating latest_state_height (just like state sync would do it)
- remove_states_below at height 2
- assert that the certification is not gone
- commit and certify at height 2

Let's please also sanity check that this test fails if this block of code is temporarily removed again.

schneiderstefan:
Does this intentionally also revert the revert? Can't the fix go in independently?

rs/state_manager/tests/state_manager.rs
mraszyk: or rather one with a proper hash

rs/state_manager/tests/state_manager.rs
```diff
@@ -8985,26 +9035,48 @@ fn commit_and_certify_optimization_conditions() {
         sm.commit_and_certify_at_height(state, Height::new(2), CertificationScope::Full, None);
```
mraszyk, line 9042: We should still
```
sm.deliver_state_certification(...)
```
so that the optimization does not trigger because of Scope::Full only.

rs/state_manager/tests/state_manager.rs
```diff
@@ -8969,12 +9015,16 @@ fn fake_certification_for_height_with_hash(height: Height, hash: CryptoHash) ->
 fn commit_and_certify_optimization_conditions() {
```
mraszyk, line 9015: We should add a scenario when a missing certification is the only reason to not trigger optimization (i.e., the other conditions are all satisfied).

mraszyk:
> Does this intentionally also revert the revert? Can't the fix go in independently?

I think it's better to revert the revert so that we don't need to separate the changes...

schneiderstefan:
>I think it's better to revert the revert so that we don't need to separate the changes...

That's ok, but then the title and description should reflect that too.

rs/state_manager/tests/state_manager.rs
michael-weigelt: 567d07efcd

rs/state_manager/tests/state_manager.rs
```diff
@@ -9527,3 +9636,41 @@ fn commit_and_certify_panic_on_delivered_fake_certification() {
         assert_eq!(no_state_clone_count(metrics), 0);
     });
 }
+
+#[test]
+fn certification_not_pruned() {
+    state_manager_test(|metrics, sm| {
+        // update `fast_forward_height` to enable optimization
+        sm.update_fast_forward_height(Height::new(5));
```
schneiderstefan, line 9655: Could you do a variant where we are not in the fast forward mode, please?

michael-weigelt: 39a47fbc41

rs/state_machine_tests/src/lib.rs
```diff
@@ -3665,6 +3667,7 @@ impl StateMachine {
 
         env.state_manager
             .commit_and_certify(state, CertificationScope::Full, None);
+        self.state_manager.flush_hash_channel();
```
eichhorl: Should it be `env`?
```suggestion
        env.state_manager.flush_hash_channel();
```

rs/state_manager/src/lib.rs
```diff
@@ -3329,12 +3311,75 @@ impl StateManager for StateManagerImpl {
...
+                        self.log,
+                        "Previous state hash was not available after awaiting the hash thread. This is a bug."
+                    );
+                    debug_assert!(false);
+                    let certification = StateManagerImpl::compute_certification_metadata(
+                        &state,
+                        prev_height,
+                        &self.metrics,
+                        &self.log,
+                    )
```
eichhorl: Is the goal to compute the certification of the previous height? In that case, should we also use the state of the previous height?

rs/state_machine_tests/src/lib.rs
michael-weigelt: What happened here? Thanks for catching

rs/state_manager/tests/state_manager.rs
```diff
@@ -9644,21 +9655,118 @@ fn certification_not_pruned() {
...
+
+        // This needs the hash @ prev_height = 2 to
+        // still be available, which is only the case if the tip-height
+        // protection kept the certification metadata at height 2 around.
+        sm.commit_and_certify(state, CertificationScope::Metadata, None);
+    })
+}
+
+#[test]
+fn remove_states_below_protects_tip_height_with_optimization() {
```
mraszyk, line 9726: Could you please check that the optimization triggered via metrics? I'm not sure it actually did because update_fast_forward_height happened only after commit_and_certify

rs/state_manager/tests/state_manager.rs
```diff
@@ -9644,21 +9655,118 @@ fn certification_not_pruned() {
...
+        // This clones the verified checkpoint @1 to @10 and advances
+        // `latest_state_height` to 10, but does not touch `tip_height`.
+        sm.fetch_state(Height(10), hash_at_1.clone(), Height::new(999));
+        sm.flush_hash_channel();
+        assert_eq!(sm.latest_state_height(), Height(10));
+
+        // height 2 should be retained
+        sm.remove_states_below(Height(10));
+        assert!(
+            sm.list_state_heights(CERT_ANY).contains(&Height(2)),
```
mraszyk: Here we also need the certification delivered by consensus in advance so maybe let's add an assertion for that?

rs/state_manager/tests/state_manager.rs
```diff
@@ -9527,3 +9647,126 @@ fn commit_and_certify_panic_on_delivered_fake_certification() {
         assert_eq!(no_state_clone_count(metrics), 0);
     });
 }
+
+#[test]
+fn certification_not_pruned() {
```
mraszyk, line 9652: Is this test still needed: isn't it now subsumed by remove_states_below_protects_tip_height_with_optimization?

rs/state_manager/src/lib.rs
mraszyk: Good point! This "trick" used to work when we computed the hash at take_tip (before computing the next state), but not when we only compute the hash in (the next) commit_and_certify. So I guess we need to panic after all...

rs/state_manager/src/lib.rs
```diff
@@ -3329,12 +3311,75 @@ impl StateManager for StateManagerImpl {
...
+                certification.certified_state_hash.clone()
+            } else {
+                // Wait for the hashing thread.
+                let (sender, recv) = bounded(1);
+                self.hash_channel
+                    .send(HashRequest::Wait { sender })
+                    .expect("Failed to send `Wait` to hash channel");
+                recv.recv().expect("Failed to wait for hash channel");
+                // After awaiting the hashing thread, snapshot and certification_metadata
+                // must have an entry at prev_height, so we can unwrap.
```
mraszyk: looks outdated, but not for long (since we'd likely need to get back to unwrapping here.

rs/state_manager/tests/state_manager.rs
michael-weigelt: No, the one below uses fetch_state, simulating state sync just ending between take_tip and c&c

schneiderstefan (approved):

schneiderstefan (approved):

PR #10080, opened 2026-05-04
feat(crypto): support Ed25519/EdDSA in COSE parser for WebAuthn

sea-snake:
## Problem

The IC's COSE parser at `rs/crypto/internal/crypto_lib/basic_sig/cose/src/lib.rs` only accepts ECDSA-P256 (`kty=EC2`, `alg=ES256`) and RSA-PKCS1-SHA256 (`kty=RSA`, `alg=RS256`) keys. WebAuthn authenticators that produce `kty=OKP`, `alg=EdDSA`, `crv=Ed25519` keys (e.g. NitroKey 3A) are rejected with the misleading error:

> Invalid delegation: Invalid public key: Algorithm Unspecified not supported: Algorithm not supported in COSE parser

The "Unspecified" comes from the wrapper at `parse_cose_public_key`, which hardcodes `AlgorithmId::Unspecified` for any `AlgorithmNotSupported` outcome — so the actual EdDSA `alg` value never surfaces.

### Why Ed25519 recovery phrases work but Ed25519 WebAuthn keys don't

The IC already has full Ed25519 support for **non-WebAuthn** keys. Recovery phrases in Internet Identity generate raw Ed25519 keys that are DER-encoded with the standard Ed25519 algorithm identifier (RFC 8410, OID `1.3.101.112`). These take the `ed25519_algorithm_identifier()` path in `user_public_key_from_bytes` and are verified via `verify_basic_sig_by_public_key(AlgorithmId::Ed25519, ...)` with a plain signature — no WebAuthn envelope, no COSE parsing.

WebAuthn authenticators like the NitroKey 3A take a **different code path**: the public key is COSE-encoded and DER-wrapped with the IC's COSE OID (`1.3.6.1.4.1.56387.1.1`). This hits the `cose_algorithm_identifier()` branch, which calls `parse_cose_public_key` — and that parser only knew ECDSA-P256 and RSA-SHA256. The Ed25519 cryptographic primitives were always there; the COSE parser just never learned to unwrap Ed25519/EdDSA keys from the COSE map format that WebAuthn produces.

This was reported via dfinity/internet-identity#3835 with a captured COSE key from a NitroKey 3A. The captured key has only the standard 4 COSE map entries (no `key_ops` or other extension fields), confirming the root cause is the parser's algorithm allowlist, not extra entries.

### `cose` crate
- Add `CosePublicKey::Ed25519(Vec<u8>)` variant.
- Add `parse_eddsa_ed25519` helper that decodes `kty=OKP / alg=EdDSA / crv=Ed25519 / x=<32 bytes>` and returns the RFC 8410 SPKI DER via `ic_ed25519::PublicKey::deserialize_raw().serialize_rfc8410_der()`.
- Wire the new branch into `CosePublicKey::from_cbor`.
- Add `ic-ed25519` to the crate's `Cargo.toml` and `BUILD.bazel`.

### `ic-crypto-standalone-sig-verifier`
- Add `KeyBytesContentType::Ed25519PublicKeyDerWrappedCose`.
- Map `AlgorithmId::Ed25519` to it in `cose_key_bytes_content_type`.

### `ic-validator`
- `ingress_validation.rs`: include `Ed25519PublicKeyDerWrappedCose` in the WebAuthn signature path (both in `validate_signed_request` and `validate_signed_delegation`).
- `webauthn.rs`: add an `AlgorithmId::Ed25519` arm to `basic_sig_from_webauthn_sig`. Per [WebAuthn §6.5.6](https://www.w3.org/TR/webauthn-2/#sctn-signature-attestation-types), EdDSA WebAuthn signatures are 64 raw bytes (no DER wrapping) — they pass through unchanged.

### Tests
- `cose/tests/tests.rs`: parsing of the RFC 8032 TEST 1 keypair as a COSE key, end-to-end signature verification through the parsed DER, parsing of the captured NitroKey 3A WebAuthn key, rejection of `crv=Ed448` and short-x malformed keys.
- `validator/src/webauthn.rs`: a new `ed25519` test module mirroring the existing `ecdsa` and `rsa` modules, with a valid-signature, wrong-message, and malformed-signature test. Updates the existing `should_return_error_if_algorithm_id_is_not_supported` test (which previously asserted Ed25519 was *un*supported) to use `AlgorithmId::EcdsaSecp256k1` instead.

## Verification

```
cargo test -p ic-crypto-internal-basic-sig-cose -p ic-crypto-standalone-sig-verifier -p ic-validator
cargo clippy --all-features --tests -p ic-crypto-internal-basic-sig-cose -p ic-crypto-standalone-sig-verifier -p ic-validator -- -D warnings -D clippy::all
```

All clean locally. Bazel build/test was not run in the development environment — please verify in CI.

## Related

- dfinity/internet-identity#3835 — diagnostic flow that captured the COSE key from a NitroKey 3A and confirmed the parse failure.


sea-snake:
Superseded by #10081 (same changes, branch on dfinity/ic instead of fork).

PR #10081, opened 2026-05-04
feat(crypto): support Ed25519/EdDSA in COSE parser for WebAuthn

sea-snake:
## Problem

The IC's COSE parser at `rs/crypto/internal/crypto_lib/basic_sig/cose/src/lib.rs` only accepts ECDSA-P256 (`kty=EC2`, `alg=ES256`) and RSA-PKCS1-SHA256 (`kty=RSA`, `alg=RS256`) keys. WebAuthn authenticators that produce `kty=OKP`, `alg=EdDSA`, `crv=Ed25519` keys (e.g. NitroKey 3A) are rejected with the misleading error:

> Invalid delegation: Invalid public key: Algorithm Unspecified not supported: Algorithm not supported in COSE parser

The "Unspecified" comes from the wrapper at `parse_cose_public_key`, which hardcodes `AlgorithmId::Unspecified` for any `AlgorithmNotSupported` outcome — so the actual EdDSA `alg` value never surfaces.

### Why Ed25519 recovery phrases work but Ed25519 WebAuthn keys don't

The IC already has full Ed25519 support for **non-WebAuthn** keys. Recovery phrases in Internet Identity generate raw Ed25519 keys that are DER-encoded with the standard Ed25519 algorithm identifier (RFC 8410, OID `1.3.101.112`). These take the `ed25519_algorithm_identifier()` path in `user_public_key_from_bytes` and are verified via `verify_basic_sig_by_public_key(AlgorithmId::Ed25519, ...)` with a plain signature — no WebAuthn envelope, no COSE parsing.

WebAuthn authenticators like the NitroKey 3A take a **different code path**: the public key is COSE-encoded and DER-wrapped with the IC's COSE OID (`1.3.6.1.4.1.56387.1.1`). This hits the `cose_algorithm_identifier()` branch, which calls `parse_cose_public_key` — and that parser only knew ECDSA-P256 and RSA-SHA256. The Ed25519 cryptographic primitives were always there; the COSE parser just never learned to unwrap Ed25519/EdDSA keys from the COSE map format that WebAuthn produces.

This was reported via dfinity/internet-identity#3835 with a captured COSE key from a NitroKey 3A. The captured key has only the standard 4 COSE map entries (no `key_ops` or other extension fields), confirming the root cause is the parser's algorithm allowlist, not extra entries.

### `cose` crate
- Add `CosePublicKey::Ed25519(Vec<u8>)` variant.
- Add `parse_eddsa_ed25519` helper that decodes `kty=OKP / alg=EdDSA / crv=Ed25519 / x=<32 bytes>` and returns the RFC 8410 SPKI DER via `ic_ed25519::PublicKey::deserialize_raw().serialize_rfc8410_der()`.
- Wire the new branch into `CosePublicKey::from_cbor`.
- Add `ic-ed25519` to the crate's `Cargo.toml` and `BUILD.bazel`.

### `ic-crypto-standalone-sig-verifier`
- Add `KeyBytesContentType::Ed25519PublicKeyDerWrappedCose`.
- Map `AlgorithmId::Ed25519` to it in `cose_key_bytes_content_type`.

### `ic-validator`
- `ingress_validation.rs`: include `Ed25519PublicKeyDerWrappedCose` in the WebAuthn signature path (both in `validate_signed_request` and `validate_signed_delegation`).
- `webauthn.rs`: add an `AlgorithmId::Ed25519` arm to `basic_sig_from_webauthn_sig`. Per [WebAuthn §6.5.6](https://www.w3.org/TR/webauthn-2/#sctn-signature-attestation-types), EdDSA WebAuthn signatures are 64 raw bytes (no DER wrapping) — they pass through unchanged.

### Tests
- `cose/tests/tests.rs`: parsing of the RFC 8032 TEST 1 keypair as a COSE key, end-to-end signature verification through the parsed DER, parsing of the captured NitroKey 3A WebAuthn key, rejection of `crv=Ed448` and short-x malformed keys.
- `validator/src/webauthn.rs`: a new `ed25519` test module mirroring the existing `ecdsa` and `rsa` modules, with a valid-signature, wrong-message, and malformed-signature test. Updates the existing `should_return_error_if_algorithm_id_is_not_supported` test (which previously asserted Ed25519 was *un*supported) to use `AlgorithmId::EcdsaSecp256k1` instead.

## Verification

```
cargo test -p ic-crypto-internal-basic-sig-cose -p ic-crypto-standalone-sig-verifier -p ic-validator
cargo clippy --all-features --tests -p ic-crypto-internal-basic-sig-cose -p ic-crypto-standalone-sig-verifier -p ic-validator -- -D warnings -D clippy::all
```

All clean locally. Bazel build/test was not run in the development environment — please verify in CI.

## Related

- dfinity/internet-identity#3835 — diagnostic flow that captured the COSE key from a NitroKey 3A and confirmed the parse failure.


sea-snake:
Addressed Copilot review (99ee480):

- Dropped the redundant `.clone()` — `webauthn_sig.signature()` already returns an owned `Blob`, so `.0` moves the Vec out without a second allocation.
- Added an explicit 64-byte length check with a clear error message (`Invalid Ed25519 signature length: expected 64 bytes, got N`), so malformed signatures are caught early in the validator rather than surfacing a generic error from the crypto layer.
- Tightened the malformed-signature test to assert the new error text.

rs/validator/src/webauthn.rs
eichhorl: Not your change, but does this work?
```suggestion
    webauthn_sig: &WebAuthnSignature,
```

eichhorl (commented):
Thanks @sea-snake, is there a corresponding interface spec PR?

Some ideas for more tests:
- Extend `rs/validator/src/ingress_validation/tests.rs` with a test validating Ed25519
- Incorrect public key should be rejected in `rs/validator/src/webauthn.rs`
- Incorrect signature with correct length should be rejected in `rs/validator/src/webauthn.rs`
- Parsing of the new `Ed25519PublicKeyDerWrappedCose` variant in `rs/crypto/tests/request_id_signatures.rs`
- Extend the `rs/tests/crypto/ingress_verification_test.rs` system test with the new scheme

sea-snake:
Thanks for the review! Addressing both threads:

**Spec PR**: dfinity/portal#6239. Adds EdDSA on curve Ed25519 to the WebAuthn allowed-schemes list, and clarifies that Ed25519 WebAuthn signatures are the raw 64-byte `R || s` concatenation from RFC 8032 §5.1.6, not DER-wrapped (which only applies to ECDSA).

**Tests** (commit edaadbc1):
- `rs/crypto/tests/request_id_signatures.rs`: `should_correctly_parse_cose_encoded_der_wrapped_ed25519_pk` asserts `AlgorithmId::Ed25519` and `KeyBytesContentType::Ed25519PublicKeyDerWrappedCose` for a COSE-DER-wrapped Ed25519 pk.
- `rs/validator/src/ingress_validation/tests.rs`: `validate_signature_webauthn_ed25519` covers an end-to-end validator check on `message_test_id(13)` signed with the RFC 8032 TEST 1 keypair.
- `rs/validator/src/webauthn.rs` (ed25519 mod):
  - `should_return_error_on_incorrect_public_key` uses RFC 8032 TEST 2 pk against a TEST 1 signature.
  - `should_return_error_on_correct_length_but_invalid_ed25519_signature` uses 64 zero bytes, which passes the length check and is then rejected by crypto verification.
- `rs/tests/crypto/ingress_verification_test.rs`: extended with a `WebAuthnEd25519` identity type, including the COSE wrapping and signing helpers that mirror the existing ECDSA/RSA equivalents.

rs/validator/src/webauthn.rs
sea-snake: Yes, that works. The function was declared with `&&WebAuthnSignature` to match the call site, which was passing `&webauthn_sig` even though `webauthn_sig` was already `&WebAuthnSignature`. Dropped both in the latest commit.

eichhorl (approved):

PR #10086, opened 2026-05-05
fix: instructions for Operator::ReturnCallIndirect

mraszyk:
This PR fixes the instruction count for `Operator::ReturnCallIndirect` to be 2x that of `Operator::ReturnCall` consistently with `Operator::CallIndirect` being 2x that of `Operator::Call`.


schneiderstefan (approved):

ggreif:
@mraszyk @schneiderstefan just for the record, I got the word from Alex Uta that the early version of `wasmtime` this was originally benchmarked on actually arrived at the crazy cost of 60. It might have well went down in the meantime. Were those benches automatised for an easy re-run?

schneiderstefan:
>@mraszyk @schneiderstefan just for the record, I got the word from Alex Uta that the early version of wasmtime this was originally benchmarked on actually arrived at the crazy cost of 60. It might have well went down in the meantime. Were those benches automatised for an easy re-run?

@ggreif Thanks for pointing that out. I wasn't aware. I suggest we just revert the PR, as the previous number seems to have had a better claim at being data driven.

PR #10088, opened 2026-05-05
feat: Add canister_metrics management canister endpoint

dsarlis:
Implement the `canister_metrics` endpoint as described in https://github.com/dfinity/portal/pull/6217 to allow controllers or subnet admins to retrieve canister level metrics for a target canister. The metrics currently contain some basic cycles related ones but can be extended in the future to contain more.

The necessary boilerplate is added to wire the new endpoint through the code stack. As defined in the interface spec, the API is available both in replicated mode as well as a query call.

A few tests are also added to ensure that the endpoint correctly returns the values that are present in the replicated state.

### DIFF
--- a/rs/execution_environment/src/canister_manager/tests.rs
+++ b/rs/execution_environment/src/canister_manager/tests.rs
@@ -40,9 +40,9 @@ use ic_limits::SMALL_APP_SUBNET_MAX_SIZE;
 use ic_logger::replica_logger::no_op_logger;
 use ic_management_canister_types_private::{
     CanisterChange, CanisterChangeDetails, CanisterChangeOrigin, CanisterIdRecord,
-    CanisterInstallMode, CanisterInstallModeV2, CanisterSettingsArgsBuilder,
+    CanisterInstallMode, CanisterInstallModeV2, CanisterMetricsResult, CanisterSettingsArgsBuilder,
     CanisterStatusResultV2, CanisterStatusType, CanisterUpgradeOptions, ChunkHash,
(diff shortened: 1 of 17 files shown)


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -8354,3 +8367,128 @@ fn subnet_admin_cannot_install_code() {
...
+    test.canister_state_mut(canister_id)
+        .system_state
+        .consume_cycles(burned_cycles);
+
+    let expected_cycles_consumed = CyclesConsumed::new(
+        memory_cycles.nominal(),
+        compute_allocation_cycles.nominal(),
+        ingress_induction_cycles.nominal(),
+        // Instructions will include both the "fake" value set as well as the cost
+        // of executing `canister_metrics` for the canister.
```
michael-weigelt, line 8447: `canister_metrics` is called as a query, so do we expect the instructions to be counted? Or is this for in replicated mode (not even sure if that makes a difference)?

dsarlis: This test operates in replicated mode (follow the path in `ExecutionTest::canister_metrics`) where instructions will be counted for executing `canister_metrics` so without this adjustment the test would fail. The case of query calls is covered in another test in another file below where access to the endpoint is mainly verified.

I felt that this way we get enough coverage and we don't need to have lots of setup steps in each case, the two paths (replicated and non-replicated) use the same `CanisterManager::get_canister_metrics` on purpose to ensure consistency in the results returned.

michael-weigelt (approved):

schneiderstefan (approved):

PR #10092, opened 2026-05-05
fix: [DSM-103] Ignore executed_rounds when computing total AP

alin-at-dfinity:
In tests checking that the accumulated priority (AP) sum across all canisters is zero, ignore `executed_rounds`: the zero lower-bound invariant strictly applies to AP now.


schneiderstefan (approved):
Nit: Could you replace AP by "accumulated priority" at least at one place in either the title or the description, for clarity?

alin-at-dfinity:
> Nit: Could you replace AP by "accumulated priority" at least at one place in either the title or the description, for clarity?

Good point, done.

PR #10093, opened 2026-05-05
feat(ckbtc-minter): support ICRC-21 consent messages

Dfinity-Bjoern:
## Summary

Implements the [ICRC-21](https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md) canister call consent message standard on the ckBTC minter so that ICRC-21-aware wallets can display a human-readable description of `retrieve_btc_with_approval` calls before the user signs.

Both display variants from the standard are supported:

- **GenericDisplay** — a Markdown message intended for software wallets with a full screen.
- **FieldsDisplay** — a list of `(label, Value)` fields tailored for hardware wallets (e.g. Ledger Nano S+).

Also adds an `icrc10_supported_standards` query that advertises ICRC-10 and ICRC-21.

## Implementation

- New module `rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs` holds the consent-message construction and `icrc10_supported_standards`.
- `main.rs` exposes the two new endpoints (`icrc21_canister_call_consent_message` as `update`, `icrc10_supported_standards` as `query`).
- `ckbtc_minter.did` declares the ICRC-10 / ICRC-21 types and methods.
- The handler enforces the standard's 500-byte argument-size limit and returns `Icrc21Error::UnsupportedCanisterCall` for unknown methods or undecodable args.

## Test plan

- [x] 13 new unit tests in `updates::icrc21::tests` covering: argument-size limit, unsupported method, decode failures, GenericDisplay output for `retrieve_btc_with_approval` (with and without subaccount) and `retrieve_btc`, FieldsDisplay output for short and long addresses, subaccount chunking, label suffixing, chunk-text edge cases (empty / exact boundary / multi-byte UTF-8), and round-trip reassembly of chunked values.
- [x] `cargo clippy -p ic-ckbtc-minter --lib --bins` clean with workspace lint set.
- [x] `cargo test -p ic-ckbtc-minter --bin ic-ckbtc-minter check_candid_interface_compatibility` passes (verifies the `.did` matches the implementation).
- [ ] Try the FieldsDisplay output on an actual Ledger Nano S+ to verify chunked labels and values fit on screen as expected.


rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,602 @@
...
+use icrc_ledger_types::icrc21::responses::{ConsentInfo, ConsentMessage, FieldsDisplay, Value};
+
+/// The token symbol used in consent messages. The minter burns ckBTC to
+/// release the equivalent in BTC.
+pub const TOKEN_SYMBOL: &str = "ckBTC";
+
+/// The number of decimals used to display token amounts.
+/// Both ckBTC and BTC use 8 decimals (1 BTC = 10^8 satoshis).
+pub const DECIMALS: u8 = 8;
+
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Good catch — fixed in 9a2df03. Introduced a `TokenSymbols` struct that maps `Network::Mainnet → (ckBTC, BTC)` and `Network::Testnet | Regtest → (ckTESTBTC, TESTBTC)`, matching the convention in [dashboard.rs:56-68](https://github.com/dfinity/ic/blob/master/rs/bitcoin/ckbtc/minter/src/dashboard.rs#L56-L68) and [update_balance.rs:267-269](https://github.com/dfinity/ic/blob/master/rs/bitcoin/ckbtc/minter/src/updates/update_balance.rs#L267-L269). The symbols are read from `s.btc_network` and threaded through both the GenericDisplay Markdown and FieldsDisplay fields (intent, amount symbol, address label). Added `test_token_symbols_for_network`, `test_retrieve_btc_with_approval_uses_testnet_symbols` and `test_retrieve_btc_with_approval_generic_uses_testnet_symbols` to lock this in.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,602 @@
...
+/// Pushes a `(label, Value::Text)` field, splitting `content` into chunks of at
+/// most [`FIELDS_DISPLAY_TEXT_CHUNK_LEN`] characters when needed. When chunked,
+/// the label gets a "(i/n)" suffix so a hardware-wallet user can tell where
+/// each chunk fits in the whole value.
+///
+/// Bitcoin addresses and the chunked subaccount hex are pure ASCII, so chunking
+/// on byte boundaries is identical to chunking on character boundaries here.
+/// We still split on `char_indices` to keep the helper safe for any future
+/// caller that passes non-ASCII text.
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Fixed in 9a2df03. Updated the doc comment to say the helper iterates Unicode scalar values via `char`, which matches the implementation. The intent — keeping the helper safe for non-ASCII callers — is unchanged.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,713 @@
...
+                ckbtc: "ckTESTBTC",
+                btc: "TESTBTC",
+            },
+        }
+    }
+}
+
+/// Maximum number of bytes that an argument to a ckBTC minter method can have
+/// when passed through the ICRC-21 endpoint.
+pub const MAX_CONSENT_MESSAGE_ARG_SIZE_BYTES: usize = 500;
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Fixed in e2efdac. Replaced the local constant with `icrc_ledger_types::icrc21::lib::MAX_CONSENT_MESSAGE_ARG_SIZE_BYTES`. The shared constant is a `u16`, so I cast to `usize` at the comparison and test sites.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+use icrc_ledger_types::icrc21::errors::{ErrorInfo, Icrc21Error};
+use icrc_ledger_types::icrc21::lib::MAX_CONSENT_MESSAGE_ARG_SIZE_BYTES;
+use icrc_ledger_types::icrc21::requests::{
+    ConsentMessageMetadata, ConsentMessageRequest, DisplayMessageType,
+};
+use icrc_ledger_types::icrc21::responses::{ConsentInfo, ConsentMessage, FieldsDisplay, Value};
+
+/// The number of decimals used to display token amounts.
+/// Both ckBTC and BTC use 8 decimals (1 BTC = 10^8 satoshis).
+pub const DECIMALS: u8 = 8;
```
mbjorkqvist: nit: This is only used within the module:
```suggestion
const DECIMALS: u8 = 8;
```

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+/// variant. Hardware wallets like the Ledger Nano S+ have a small display and
+/// cannot fit a full Bitcoin address (up to 62 characters for bech32m) or a
+/// 64-character hex subaccount on a single screen. Long values are split into
+/// multiple fields with a "(N/M)" pagination suffix in the label so each chunk
+/// fits on a single screen with the label on top.
+///
+/// Field labels are kept short (e.g. "BTC address", "From subaccount") so that
+/// even with a "(N/M)" suffix the label still fits on one line at the typical
+/// hardware-wallet font size.
+pub const FIELDS_DISPLAY_TEXT_CHUNK_LEN: usize = 18;
```
mbjorkqvist: nit: This is also not used outside the module:
```suggestion
const FIELDS_DISPLAY_TEXT_CHUNK_LEN: usize = 18;
```

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+/// Maximum number of characters per `Value::Text` field for the FieldsDisplay
+/// variant. Hardware wallets like the Ledger Nano S+ have a small display and
+/// cannot fit a full Bitcoin address (up to 62 characters for bech32m) or a
+/// 64-character hex subaccount on a single screen. Long values are split into
+/// multiple fields with a "(N/M)" pagination suffix in the label so each chunk
+/// fits on a single screen with the label on top.
+///
+/// Field labels are kept short (e.g. "BTC address", "From subaccount") so that
+/// even with a "(N/M)" suffix the label still fits on one line at the typical
+/// hardware-wallet font size.
```
mbjorkqvist: The [ICRC-21.did](https://github.com/dfinity/wg-identity-authentication/blob/c72de52f21240d973501dc7e9362b93bfcb2072a/topics/ICRC-21/ICRC-21.did#L22-L30) canister interface spec describes the `FieldsDisplay` and `GenericDisplay` devices as having line-wrapping/pagination capability, which suggests the canister should provide one logical field per value and trust the device to render it. Pre-chunking with (N/M) label suffixes works around the spec's design assumption rather than following it, so if possible, I'd recommend not performing this chunking in the minter.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+                 the Bitcoin address below."
+            ));
+            message.push_str(&format!(
+                "\n\n**Amount to convert:** `{} {ckbtc}`",
+                format_amount(args.amount, DECIMALS)
+            ));
+            message.push_str(&format!(
+                "\n\n**Bitcoin destination address:**\n`{}`",
+                args.address
+            ));
```
mbjorkqvist: nit: There's a possible Markdown injection vector here, where the user could provide `args` such as `retrieve_btc { address: "bc1q…\n# You will receive 100 BTC", amount: 50_000 }`, which would lead the minter to construct an invalid/confusing consent message. The subsequent call to `retrieve_btc_with_approval` with the same, signed arguments would fail, since at that point the minter tries to parse the address and would return an error. We could add a check here too, but there's (currently) no risk of e.g., losing funds, so this is not a blocker.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+                 (minus network and minter fees) to the Bitcoin address below."
+            ));
+            message.push_str(&format!(
+                "\n\n**Amount to convert:** `{} {ckbtc}`",
+                format_amount(args.amount, DECIMALS)
+            ));
+            message.push_str(&format!(
+                "\n\n**Bitcoin destination address:**\n`{}`",
+                args.address
+            ));
```
mbjorkqvist: Same Markdown injection issue as in `build_retrieve_btc_with_approval_message`.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,710 @@
...
+
+pub fn icrc10_supported_standards() -> Vec<StandardRecord> {
+    vec![
+        StandardRecord {
+            name: "ICRC-10".to_string(),
+            url: "https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-10/ICRC-10.md".to_string(),
+        },
+        StandardRecord {
+            name: "ICRC-21".to_string(),
+            url: "https://github.com/dfinity/wg-identity-authentication/blob/main/topics/ICRC-21/icrc_21_consent_msg.md".to_string(),
```
mbjorkqvist: This URL is different from the one in [the ICRC-21 spec](https://github.com/dfinity/wg-identity-authentication/blob/c72de52f21240d973501dc7e9362b93bfcb2072a/topics/ICRC-21/ICRC-21.did#L169). However, the ICRC-21 standard is not (yet) available at the URL in the spec: [https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md](https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md). Fixing this as part of this PR doesn't necessarily make sense, but we may want to consider publishing the ICRC-21 standard in the ICRC repo, and then fixing the URL in our other canisters also.

mbjorkqvist (commented):
Thanks @Dfinity-Bjoern, just a few minor comments!

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: Agreed and fixed in 3bf0482 (refactor) + e96631c (rebase). Confirmed by reading the Ledger ICP app: [`msg.rs:render_item`](https://github.com/Zondax/ledger-icp/blob/main/app/rust/src/parser/consent_message/msg.rs) reads the raw `Value::Text { content }` and passes it to [`handle_ui_message(text, message, page)`](https://github.com/Zondax/ledger-icp/blob/main/app/rust/src/utils.rs#L630-L652), which slices the value into `out.len()-1`-sized device pages (`item.chunks(m_len).nth(page)`) and returns the total page count so the device-side UI handles scrolling natively. Pre-chunking on the canister side was actively harmful — one logical address turned into 3 separate fields the user had to step through with confusing `(1/3)` labels, and the device couldn't paginate cleanly. Dropped `push_chunked_text` / `chunk_text` / `FIELDS_DISPLAY_TEXT_CHUNK_LEN` and now emit each long value as a single `Value::Text`. Short labels (`Amount`, `BTC address`, `From subaccount`) are kept since titles in the Ledger app are truncated rather than paginated.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: Good catch — fixed in 6285499. Added a `validate_address` step that runs `BitcoinAddress::parse(&args.address, network)` before the address gets interpolated into any consent message. If parsing fails the endpoint returns `Icrc21Error::UnsupportedCanisterCall` with a generic "Invalid Bitcoin destination address" message — the malicious payload is never echoed back. Using the same parser as `retrieve_btc[_with_approval]` keeps the consent endpoint and the actual call in agreement on what's accepted. Added `test_malformed_address_is_rejected` covering Markdown-injection payloads (newlines, backticks, `#`) plus an address from the wrong network.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: Same fix in 6285499 — `validate_address` runs for both `retrieve_btc` and `retrieve_btc_with_approval` paths.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: I decided to just do this and created [the PR](https://github.com/dfinity/ICRC/pull/165), will attempt to get it merged tomorrow. Easy one to clean up, will change in the code. Thanks for pointing this out!

rs/bitcoin/ckbtc/minter/src/main.rs
```diff
@@ -265,6 +268,18 @@ fn decode_ledger_memo(arg: DecodeLedgerMemoArgs) -> DecodeLedgerMemoResult {
...
+fn icrc21_canister_call_consent_message(
+    consent_msg_request: ConsentMessageRequest,
+) -> Result<ConsentInfo, Icrc21Error> {
+    updates::icrc21::icrc21_canister_call_consent_message(consent_msg_request)
+}
+
+#[query]
+fn icrc10_supported_standards() -> Vec<updates::icrc21::StandardRecord> {
+    updates::icrc21::icrc10_supported_standards()
+}
```
gregorydemay, line 281: nit: let's have a smoke integration test for those new endpoints.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
mbjorkqvist: Thanks! I created [DEFI-2803](https://dfinity.atlassian.net/browse/DEFI-2803) to update the URL in other affected canisters.

mbjorkqvist (approved):

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,620 @@
...
+        "retrieve_btc" => {
+            let args =
+                Decode!(consent_msg_request.arg.as_slice(), RetrieveBtcArgs).map_err(|e| {
+                    Icrc21Error::UnsupportedCanisterCall(ErrorInfo {
+                        description: format!("Failed to decode RetrieveBtcArgs: {e}"),
+                    })
+                })?;
+            validate_address(&args.address, network)?;
+            build_retrieve_btc_message(&args, &display_type, symbols)
+        }
```
gregorydemay: `retrieve_btc` is deprecated (`retrieve_btc_with_approval` should be used instead) so I would avoid building new functionality for it unless strictly required.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,620 @@
...
+) -> ConsentMessage {
+    let TokenSymbols { ckbtc, btc } = symbols;
+    let amount = format_amount(args.amount, DECIMALS);
+    match display_type {
+        DisplayMessageType::GenericDisplay => {
+            let mut message = format!(
+                "# Convert {ckbtc} to {btc}\n\n\
+                 Authorize the {ckbtc} minter to burn {ckbtc} from your account and \
+                 send the equivalent amount in {btc} (minus network and minter fees) to \
+                 the Bitcoin address below.\n\n\
```
gregorydemay, line 144: understanding question: do ledger devices do line wrapping or do we have to manually respect a maximal number of characters per line?

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,620 @@
...
+        format!("{whole}")
+    } else {
+        let frac_str = format!("{frac:0width$}", width = decimals as usize);
+        let trimmed = frac_str.trim_end_matches('0');
+        format!("{whole}.{trimmed}")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
gregorydemay: nit: there is already a test file (`rs/bitcoin/ckbtc/minter/src/updates/tests.rs`) so I would move the tests there (into a new icrc21 moduile). There is a benefit of separating test code from prod code (not in the same file) so that changing test code does not require re-compiling the crate.

gregorydemay (commented):
Thanks for that PR @Dfinity-Bjoern ! The main comment is regarding *not* supporting ICRC-21 for `retrieve_btc`. Otherwise, the rest is minor

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: The Ledger devices use the `FieldsDisplay` variant. They do support line breaks for the values, but not for the caption.

rs/bitcoin/ckbtc/minter/src/main.rs
Dfinity-Bjoern: ... and how would I do that?

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: Moved to the `tests.rs` file, thanks for pointing this out!

rs/bitcoin/ckbtc/minter/src/main.rs
gregorydemay: Look at the tests in `rs/bitcoin/ckbtc/minter/tests/tests.rs`. (In Rust integration tests are usually in `<crate>/tests`)

Dfinity-Bjoern: Added in 3dc1a91. New `test_icrc21_endpoints_smoke` in `rs/bitcoin/ckbtc/minter/tests/tests.rs` installs a minter via the existing `CkBtcSetup::new()` (mainnet) helper and exercises the full Candid path end-to-end:

- `icrc10_supported_standards` advertises ICRC-10 and ICRC-21, with the ICRC-21 entry pointing at the canonical `dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md` URL.
- `icrc21_canister_call_consent_message` for `retrieve_btc_with_approval` returns a GenericDisplay markdown message (header + amount + address) and a FieldsDisplay structured message (intent + (Amount, BTC address) fields).
- Unsupported methods — including `retrieve_btc`, which we intentionally do not render consent for — come back as `Icrc21Error::UnsupportedCanisterCall`.

This complements the existing `updates::tests::icrc21` unit tests, which cover the inner builder directly without going through the canister boundary.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -15,20 +15,20 @@ use icrc_ledger_types::icrc21::responses::{ConsentInfo, ConsentMessage, FieldsDi
 
 /// The number of decimals used to display token amounts.
 /// Both ckBTC and BTC use 8 decimals (1 BTC = 10^8 satoshis).
-const DECIMALS: u8 = 8;
+pub(super) const DECIMALS: u8 = 8;
```
gregorydemay, line 18: Here and below, you actually don't need `pub(super)` so I would remove it (the tests compile fine without it). Claude justification is 💩 
```
Bumps the visibility of the items the tests reach into — DECIMALS,
TokenSymbols, TokenSymbols::for_network, format_amount, and
build_consent_info — to pub(super), which is the minimum needed for
the sibling `crate::updates::tests::icrc21` module to use them while
keeping them out of the crate-public API.
```
This is completely wrong. See, https://comprehensive-rust.mo8it.com/modules/visibility.html#visibility

> Parent and sibling items are always visible.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,222 @@
...
+            validate_address(&args.address, network)?;
+            build_retrieve_btc_with_approval_message(&args, &display_type, symbols)
+        }
+        method => {
+            return Err(Icrc21Error::UnsupportedCanisterCall(ErrorInfo {
+                description: format!(
+                    "The method '{method}' is not supported by the ckBTC minter ICRC-21 endpoint."
+                ),
+            }));
+        }
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Updated PR description.

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
```diff
@@ -0,0 +1,222 @@
+//! Implementation of the [ICRC-21](https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md)
+//! Canister Call Consent Message standard for the ckBTC minter.
+
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Adapted PR description.

rs/bitcoin/ckbtc/minter/tests/tests.rs
```diff
@@ -203,6 +203,161 @@ fn test_install_ckbtc_minter_canister() {
...
+        Encode!(&args).unwrap(),
+        Some(DisplayMessageType::GenericDisplay),
+    ))
+    .expect("consent message should be produced for retrieve_btc_with_approval");
+    let message = match info.consent_message {
+        ConsentMessage::GenericDisplayMessage(m) => m,
+        other => panic!("expected GenericDisplayMessage, got {other:?}"),
+    };
+    assert!(
+        message.starts_with("# Convert ckBTC to BTC"),
```
gregorydemay: nit: don't we want to assert_eq on the full message string in that case? Any change of character would be a change of behavior visible to the user (it's fine to make such a change but we should be aware of it and make it conscientiously)

gregorydemay (approved):
A couple of nits but ottherwise LGTM!

rs/bitcoin/ckbtc/minter/src/updates/icrc21.rs
Dfinity-Bjoern: I just tried this — removing the `pub(super)` annotations does not compile here. With everything left as default-private, `cargo test --no-run -p ic-ckbtc-minter --lib` fails with four `E0603 ... is private` errors:

```
error[E0603]: function `format_amount` is private
    --> rs/bitcoin/ckbtc/minter/src/updates/tests.rs:1170:78
     |
1170 |     use crate::updates::icrc21::{DECIMALS, TokenSymbols, build_consent_info, format_amount};
     |                                                                              ^^^^^^^^^^^^^ private function
```

(Same for `DECIMALS`, `TokenSymbols`, and `build_consent_info`.)

I read "Parent and sibling items are always visible" from comprehensive-rust as referring to module *declarations* being mutually accessible — i.e. `mod tests` can refer to `crate::updates::icrc21` as a path. But the items *inside* that module are still governed by their own visibility, and an item without any `pub` annotation is only visible inside its declaring module and its descendants ([Rust Reference: visibility and privacy](https://doc.rust-lang.org/reference/visibility-and-privacy.html)). `crate::updates::tests::icrc21` is a sibling of `crate::updates::icrc21`, not a descendant, so it can't reach private items in there — hence `pub(super)`.

Happy to be wrong here — if there's a tweak you have in mind that makes the tests compile without the annotations, I'd love to see it. Otherwise I'd rather keep the `pub(super)` so the build stays green.

gregorydemay: Interesting I tried with Bazel (Cargo is "basically" not supported in the monorepo)
```
bazel test //rs/bitcoin/ckbtc/minter:ckbtc_minter_lib_unit_tests
```

gregorydemay: Ok my bad, I did try out but instead of removing `pub(super)` I only removed `super`, so yeah unsurprisingly it did work 🙈 . Sorry about that.

rs/bitcoin/ckbtc/minter/tests/tests.rs
Dfinity-Bjoern: Good call — done in bd563a2. Replaced the `starts_with` + `contains` pair with a full `assert_eq!` against the verbatim Markdown the production builder produces (mainnet symbols, 0.0025 ckBTC, the WITHDRAWAL_ADDRESS, no subaccount section). Any future wording change will now have to be updated here consciously.

PR #10096, opened 2026-05-05
feat: [EXC-1783] No execution overhead on low cycles

alin-at-dfinity:
When a message execution is skipped due to low cycle balance, don't deduct the message execution overhead from the round instruction limit. The instruction overhead is intended to account for entering and exiting the sandbox, which does not apply here.

This will allow the "execution" of arbitrarily many messages per round on frozen (or close to frozen) canisters, which could drive up round duration. However, there are a lot of other limits in place (ingress messages in blocks, canister call fees, queue sizes) that make this a non-issue. Particularly given that the overhead for skipping a message execution is at least an order of magnitude lower than that of executing one.

### DIFF
--- a/rs/execution_environment/src/scheduler.rs
+++ b/rs/execution_environment/src/scheduler.rs
@@ -568,13 +568,14 @@ impl SchedulerImpl {
                 "`subnet_available_callbacks` is a lower bound estimate for the number of available callbacks"
             );
 
-            if instructions_consumed == RoundInstructions::new(0) {
+            // Stop iterating if no work was done (no instructions consumed and no system
+            // generated reject responses produced).
+            if instructions_consumed.get() == 0 && low_cycle_balance_canisters.is_empty() {
                 break state;
-            } else {
-                self.metrics
-                    .inner_loop_consumed_non_zero_instructions_count
-                    .inc();
             }
+            self.metrics
+                .inner_loop_processed_non_zero_inputs_count
+                .inc();
 
             if total_heap_delta >= self.config.max_heap_delta_per_iteration {
                 break state;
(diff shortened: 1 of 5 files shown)


rs/execution_environment/src/scheduler.rs
```diff
@@ -228,11 +228,11 @@ impl SchedulerImpl {
                 as_num_instructions(instructions_before - round_limits.instructions);
 
             let messages = match execute_subnet_message_result_type {
-                ExecuteSubnetMessageResultType::Finished => NumMessages::from(1),
-                ExecuteSubnetMessageResultType::Processing => NumMessages::from(0),
-                ExecuteSubnetMessageResultType::Paused => NumMessages::from(0),
+                ExecuteSubnetMessageResultType::Finished => NumMessages::new(1),
+                ExecuteSubnetMessageResultType::Processing => NumMessages::new(0),
+                ExecuteSubnetMessageResultType::Paused => NumMessages::new(0),
```
alin-at-dfinity, line 233: This is a drive-by cleanup: `AmountOf::new()` is `const`; `AmountOf::from()` cannot be made `const`.

rs/execution_environment/src/scheduler/tests/limits.rs
```diff
@@ -147,34 +147,36 @@ fn test_message_limit_from_message_overhead() {
         + 1;
 
     let mut callee = canister0;
-    let mut call = other_side(callee, 0);
+    let mut call = other_side(callee, 1);
```
alin-at-dfinity, line 145: Another drive-by improvement (although this one fixes a test that was broken by this change): a message execution returning "zero instructions executed" is interpreted by the scheduler logic as "message not executed due to low cycles"; that's not what this test was aiming for (which is why it broke).

schneiderstefan (approved):

PR #10099, opened 2026-05-06
test(cketh): Add retry mechanism to `assert_has_unique_events_in_order`

gregorydemay:
Upcoming message-execution changes (#10090) caused some events to not yet be visible at the moment cketh integration tests called `assert_has_unique_events_in_order`, leading to test failures. Add a retry mechanism to `assert_has_unique_events_in_order` to make the tests more robust.

### DIFF
--- a/rs/ethereum/cketh/test_utils/src/events.rs
+++ b/rs/ethereum/cketh/test_utils/src/events.rs
@@ -1,6 +1,7 @@
 use crate::CkEthSetup;
 use ic_cketh_minter::endpoints::events::{Event, EventPayload};
 use std::collections::BTreeMap;
+use std::fmt;
 
 pub struct MinterEventAssert<T> {
     setup: T,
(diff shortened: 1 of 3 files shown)


mducroux (approved):
LGTM!

rs/ethereum/cketh/test_utils/src/events.rs
```diff
@@ -67,3 +49,40 @@ impl<T> MinterEventAssert<T> {
...
+                    .is_some()
+            {
+                return Err(format!("Event {expected_event:?} occurs multiple times"));
+            }
+        }
+        if !found_event_indexes.contains_key(&index_expected_event) {
+            return Err(format!(
+                "Missing event {:?}. All events: {:?}",
+                expected_event, events
+            ));
```
mbjorkqvist: IIUC, it only makes sense to retry for this error case, and fail early for the other errors (duplicate or out-of-order events). Would it make sense to return a new `CheckError` enum from this function, and then have the caller check, and only retry on `CheckError::MissingEvent`?

mbjorkqvist (approved):
Thanks @gregorydemay, just one minor comment!

rs/ethereum/cketh/test_utils/src/events.rs
gregorydemay: Good call — done in 3bde11432e. Introduced a `CheckError` enum with `MissingEvent` / `DuplicateEvent` / `OutOfOrder` variants; only `MissingEvent` triggers a retry, the other two panic immediately. Had to box it (`Result<(), Box<CheckError>>`) to satisfy `clippy::result_large_err`.

PR #10100, opened 2026-05-06
chore(CI): don't notify teams on slack if the PR targets ai feature branch

NikolaMilosa:
Just to reduce the noise on slack for non actionable branches.


basvandijk (approved):
Let's merge this after https://github.com/dfinity/ic/pull/10101 is merged.

PR #10101, opened 2026-05-06
chore(CI): auto-close forks targeting the ai feature branch

NikolaMilosa:
Since this branch is meant for internal development, this PR is closing all prs that target specific feature branches


basvandijk (commented):
It's unexpected to find the `check-external-branch-policy` job in the `external-contrib-slack-notification.yaml‎` workflow since the former has nothing to do with slack notifications. What about moving it into a dedicated workflow?

.github/workflows/external-contrib-branch-policy.yaml
```diff
@@ -0,0 +1,65 @@
+name: External Contributor Branch Policy
+
+on:
+  pull_request_target:
+    types: [opened, reopened]
```
basvandijk: What happens when you retarget an already open PR? Will this still get triggered?

NikolaMilosa: Based on [this](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) I don't see an event type for that. Which events would you suggest we add?

basvandijk: I would guess `edited` since you have to click the edit title button to change the target branch. But you'll have to test it to be sure.

basvandijk (approved):

PR #10111, opened 2026-05-07
chore: revert that log_memory_limit cannot be set together with memory allocation

mraszyk:
At canister creation time, the log memory limit is initialized to 4KiB by default and thus the implementation must be able to deal with simultaneous changes to log memory limit and memory allocation after all.


schneiderstefan (approved):
Wait, so was canister creation broken? Or would it have been broken if the log feature was enabled?

mraszyk:
> Wait, so was canister creation broken? Or would it have been broken if the log feature was enabled?

Because the default value of log memory limit is only inserted into CanisterSettings after validating that log memory limit and memory allocation are not set simultaneously, no flow was broken. However, this means that the code must really support this case and be tested properly.

PR #10117, opened 2026-05-07
chore: Inline DeFi Bazel dependencies

nmattia:


mducroux (approved):
Merci !

PR #10120, opened 2026-05-07
feat: [DSM-106] Observe `ReplicatedState` metrics in a background thread

alin-at-dfinity:
Based on a recent CPU profile of the `alin/DSM-103-active-canister-only-scheduling` branch (i.e. after scheduler optimizations), `ReplicatedStateMetrics::observe()` accounts for 20%-25% of the DSM round duration. This is all read-only work (scanning the `ReplicatedState` and instrumenting stats) so it can be safely done on a background thread, particularly after the `StateManager` has already created a cheap to clone `Arc<ReplicatedState>`.

The change moves `ReplicatedStateMetrics::observe()` off the `commit_and_certify()` critical path by introducing a generic `WorkerThread` in `ic-utils-thread` and wiring `StateManagerImpl` to defer observations through it. There's a single‑slot bounded channel (one in‑flight + one queued); excess work is dropped and counted via `state_manager_skipped_state_observations`.


rs/state_machine_tests/src/lib.rs
```diff
@@ -3078,6 +3083,11 @@ impl StateMachine {
         }
         assert_eq!(self.state_manager.latest_state_height(), batch_number);
 
+        // Wait until the enqueued `ReplicatedStateMetrics::observe()` call has been
+        // processed by the background metrics thread.
+        if self.flush_replicated_state_metrics {
+            self.state_manager.flush_metrics_channel();
+        }
```
alin-at-dfinity, line 3072: Not sure how much of an effect this will have on the average `StateMachine` test. In my benchmark runs it added about 30% overhead (and a lot of variability), but those runs actually had 100k canisters for the metrics thread to iterate over.

I could also switch the default to not wait for metrics to be collected; and only override it in tests that actually look at `ReplicatedState` metrics. But I did that and was pretty sure it was done right; and then had a test that I missed flake out on me. So rather than look for every `ReplicatedState` metric name; and risk similar flakiness in the future, when I will likely have forgotten about this quirk; I'd rather be safe.

rs/execution_environment/src/scheduler/round_schedule.rs
```diff
@@ -331,68 +299,110 @@ impl RoundSchedule {
...
+                        // just by being idle for a round. Similarly, idle canisters with positive AP
+                        // burn it down as if they had executed full rounds until they (almost) reach 0.
+                        if is_first_iteration && !canister.must_be_in_schedule() {
+                            let canister_priority = subnet_schedule.get(canister_id);
+                            if canister_priority.accumulated_priority >= ZERO
+                                && canister_priority.accumulated_priority <= ONE_HUNDRED_PERCENT
+                            {
+                                subnet_schedule.remove(canister_id);
+                            }
+                        }
```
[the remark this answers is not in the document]
alin-at-dfinity: This is a bad merge artifact. This change was introduced as part of #10094.

rs/utils/thread/src/worker_thread.rs
```diff
@@ -0,0 +1,143 @@
...
+//! sender closes the channel, the worker drains any enqueued workloads, its
+//! `recv()` returns `Err`, and the `JoinOnDrop` handle joins the thread.
+
+use crate::JoinOnDrop;
+use crossbeam_channel::{Sender, TrySendError, bounded};
+use prometheus::IntCounter;
+
+/// A workload to be executed by the worker thread.
+type Workload = Box<dyn FnOnce() + Send>;
+
```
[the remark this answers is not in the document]
alin-at-dfinity: Made `Workload` public.

rs/execution_environment/src/scheduler/test_utilities.rs
```diff
@@ -595,6 +601,18 @@ impl SchedulerTest {
             round_type,
             self.registry_settings(),
         );
+
+        // Explicitly observe the Replicated State metrics after every round. This used
+        // to be done by the scheduler, so there are a number of tests depending on
+        // various state metrics, but instrumentation was moved into the State Manager.
+        // So we "manually"update the metrics after every round.
```
mraszyk:
```suggestion
        // So we "manually" update the metrics after every round.
```

rs/state_machine_tests/src/lib.rs
mraszyk: Agreed, let's flush by default.

rs/state_manager/tests/state_manager.rs
```diff
@@ -9648,6 +9649,29 @@ fn commit_and_certify_panic_on_delivered_fake_certification() {
     });
 }
 
+#[test]
+fn commit_and_certify_observes_replicated_state_metrics() {
+    state_manager_test(|metrics, sm| {
+        let state = sm.take_tip().1;
+        // Sanity check: metrics should be zero before any observations are made.
+        assert_matches!(
+            fetch_histogram_stats(metrics, "scheduler_canister_paused_execution"),
```
mraszyk, line 9629: Why this metric?

rs/utils/thread/src/worker_thread.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    /// Test-only: blocks until the worker thread has completed any previous
+    /// workload.
+    #[doc(hidden)]
+    pub fn flush_channel(&self) {
+        // Blocking send waits for the worker thread to complete any in-progress job.
+        let res = self.sender.send(Job::Flush());
+        println!("flush_channel: {:?}", res);
```
mraszyk:
```suggestion
```

mraszyk (approved):

rs/state_manager/tests/state_manager.rs
alin-at-dfinity: Most of the other metrics are either gauges (and they likely stay at zero with an empty state; or their values are somewhat unpredictable); or per-canister histograms (and we have zero canisters, so they get no observations). The only viable metrics are one of the 4 paused/aborted executions/install codes histograms.

Clarified that this is an arbitrary metric that is observed once per round.

rs/utils/thread/src/worker_thread.rs
alin-at-dfinity: Oops, thanks for the catch. Removed.

PR #10121, opened 2026-05-07
feat: CON-1670 Add some DKG metrics & avoid proposing empty IDKG payloads

eichhorl:
Add metrics for
* NiDKG payload/transcript creation errors
* NiDKG payload creation time
* Number of dealings & remote transcripts included in data payloads
* Sum and size of DKG attempts map in summary payloads

Additionally, we change the way _IDKG_ payload creation errors are handled: Previously, if no IDKG payload could be created, the block maker would instead propose a block _without_ an IDKG payload. This is dangerous since we could silently loose the master keys on the subnet, even if the reason for the payload creation failure was potentially transient.

After this PR, if IDKG payload creation fails, we will not propose a block. Instead we will try again later, or wait until we receive a block from different block maker.


rs/consensus/dkg/src/metrics.rs
```diff
@@ -0,0 +1,135 @@
...
+pub struct DkgPayloadMetrics {
+    payload_errors: IntCounterVec,
+    payload_duration: HistogramVec,
+}
+
+impl DkgPayloadMetrics {
+    pub fn new(metrics_registry: MetricsRegistry) -> Self {
+        Self {
+            payload_errors: metrics_registry.int_counter_vec(
+                "dkg_payload_errors",
```
pierugo-dfinity: Would it make sense to name them `consensus_dkg_...` instead of `dkg_...` for consistency with the other `Dkg` and `IDkg` metrics?

pierugo-dfinity (approved):
> This is dangerous since we could silently loose the master keys on the subnet

AFAICT, others would not validate the block as they would either compute the actual summary or also hit an error no? I still agree that it makes more sense to not propose the block in case of an error though.

eichhorl:
> > This is dangerous since we could silently loose the master keys on the subnet
> 
> AFAICT, others would not validate the block as they would either compute the actual summary or also hit an error no? I still agree that it makes more sense to not propose the block in case of an error though.

Good point, and data blocks would be rejected with `InvalidIDkgPayloadReason::MissingIDkgDataPayload` as well 👍

rs/consensus/dkg/src/metrics.rs
eichhorl: The IDKG payload metrics follow the `idkg_...` style, but it's probably better to be consistent in the same crate: https://github.com/dfinity/ic/pull/10121/commits/f090a575dbdc79f113d55cf820c4a6ba0a782b4f

pierugo-dfinity: Ah indeed, I had only seen the ones [here](https://sourcegraph.com/r/github.com/dfinity/ic@9fd53d30b7275bd777679f3d16510db80bf4abc3/-/blob/rs/consensus/src/consensus/metrics.rs?L239-258) :D

PR #10131, opened 2026-05-08
test: [DSM-126] Add synchronous commit_and_certify for tests

michael-weigelt:
Co-authored by Claude

### DIFF
--- a/rs/state_manager/src/lib.rs
+++ b/rs/state_manager/src/lib.rs
@@ -4358,6 +4358,23 @@ pub mod testing {
             batch_summary: Option<BatchSummary>,
         );
 
+        /// Testing only: Like `commit_and_certify_at_height`, but waits for hashing/certification metadata.
+        fn commit_and_certify_at_height_sync(
+            &self,
+            state: ReplicatedState,
+            height: Height,
+            scope: CertificationScope,
+            batch_summary: Option<BatchSummary>,
+        );
+
+        /// Testing only: Commits, then waits for hashing/certification metadata to be populated.
+        fn commit_and_certify_sync(
+            &self,
+            state: ReplicatedState,
+            scope: CertificationScope,
+            batch_summary: Option<BatchSummary>,
+        );
+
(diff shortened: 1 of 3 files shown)


rs/state_manager/src/lib.rs
```diff
@@ -4358,6 +4358,23 @@ pub mod testing {
             batch_summary: Option<BatchSummary>,
         );
 
+        /// Testing only: Like `commit_and_certify_at_height`, but waits for hashing/certification metadata.
+        fn commit_and_certify_at_height_sync(
```
schneiderstefan, line 4364: Can these also be used in the state_machine_tests crate and/or pocket_ic?

rs/state_manager/src/lib.rs
```diff
@@ -4508,5 +4554,15 @@ pub mod testing {
                 &mut states,
             );
         }
+
+        fn fetch_state_sync(
```
schneiderstefan: That is an unfortunate name, because it's simultaneously not related to state sync and kinda is.

rs/state_manager/src/lib.rs
```diff
@@ -4413,6 +4438,27 @@ pub mod testing {
...
+
+        fn commit_and_certify_sync(
+            &self,
+            state: ReplicatedState,
+            scope: CertificationScope,
+            batch_summary: Option<BatchSummary>,
+        ) {
+            self.commit_and_certify(state, scope, batch_summary);
+            self.flush_hash_channel();
+        }
```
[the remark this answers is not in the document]
michael-weigelt: This will not slow tests

rs/state_manager/src/lib.rs
michael-weigelt: And worse, it's slop.

rs/state_manager/src/lib.rs
michael-weigelt: Yes, done

schneiderstefan (approved):

PR #10132, opened 2026-05-08
chore: [DSM-122] Add metric for skipped optimization

michael-weigelt:


rs/state_manager/src/lib.rs
```diff
@@ -3417,18 +3424,23 @@ impl StateManager for StateManagerImpl {
                 && !height
                     .get()
```
michael-weigelt, line 3425: Line-based diff being awful. I merely put the old block inside a new if statement.

schneiderstefan (approved):

PR #10135, opened 2026-05-08
feat: don't install LLVM in ic-build image

nmattia:
This moves the aflplusplus build into a new build stage and only copies
the relevant build output into the main build image. This avoids
inheriting the LLVM install which makes the final image smaller, leading
to faster download times (3.31GB -> 2.52GB).

Note: The permissions on `mold` had to be updated; it turns out the
linker was installed incorrectly and removing llvm surfaced the issue.


nmattia:
Note: I might do the same for the cargo-audit install. Would do the same for e2fsdroid but I'd actually rather get rid of it entirely if possible.

PR #10136, opened 2026-05-08
chore: move FARM_METADATA to volatile status file

basvandijk:
Move the `FARM_META` from stable to volatile such that they won't invalidate cached system-tests.

### DIFF
--- a/bazel/workspace_status.sh
+++ b/bazel/workspace_status.sh
@@ -22,11 +22,11 @@ else
 fi
 
 # Used as farm metadata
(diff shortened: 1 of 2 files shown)


bazel/workspace_status.sh
nmattia, line 1: I think I understand your worry from the other day. Basically the `farm_metadata.txt` will never get rebuilt and will be cached; every system test will inherit the FARM_METADATA of the first ever system that was run.

basvandijk:
Closing since this is now done directly in https://github.com/dfinity/ic/pull/10122.

PR #10137, opened 2026-05-08
chore: remove testonly for non-test targets

nmattia:
The `testonly` tag was used very liberally throughout the codebase, and was applied to many targets that definitely were not test only (like ic-admin, published pocket-ic executables, etc). This removes some of the uses of `testonly` (it tends to spread like wildfire).

### DIFF
--- a/rs/crypto/iccsa/test_utils/BUILD.bazel
+++ b/rs/crypto/iccsa/test_utils/BUILD.bazel
@@ -7,7 +7,6 @@ package(default_visibility = [
 
 rust_library(
     name = "test_utils",
-    testonly = True,
     srcs = glob(["src/**"]),
     crate_name = "ic_crypto_iccsa_test_utils",
     version = "0.9.0",
--- a/rs/nns/init/src/lib.rs
+++ b/rs/nns/init/src/lib.rs
@@ -2,7 +2,6 @@
 //! initial payloads and install nns canisters in a fresh IC.
 //! Was used at genesis, but now only used to install the nns in testnets.
 
-use canister_test::Wasm;
 use ic_canister_client::Sender;
 use ic_interfaces_registry::{RegistryDataProvider, ZERO_REGISTRY_VERSION};
 use ic_nns_constants::NNS_CANISTER_WASMS;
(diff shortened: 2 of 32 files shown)


nmattia:
(will need to update the cargo deps as well)

basvandijk (approved):
I remember there was some reason to mark them as `testonly` but can't remember the details. But I guess if CI passes it's good!

rs/crypto/iccsa/test_utils/BUILD.bazel
```diff
@@ -7,7 +7,6 @@ package(default_visibility = [
 
 rust_library(
     name = "test_utils",
-    testonly = True,
```
schneiderstefan, line 10: A lot of them, like this one, is in a directory containing test_utils. Shouldn't we keep them for those?

nmattia: Good catch, I'll need to double check this. Might have been too zealous.

Bownairo (approved):

rs/nns/init/src/lib.rs
```diff
@@ -154,12 +153,12 @@ pub fn set_up_env_vars_for_all_canisters<P: AsRef<Path>>(wasm_dir: P) {
             let mut path = wasm_dir.as_ref().to_path_buf();
             path.push(file_part.as_str());
             if path.is_file() {
-                // Sets up the env var following the pattern expected by
-                // WASM::from_location_specified_by_env_var
```
jasonz-dfinity, line 158: Should we still keep the reference to `WASM::from_location_specified_by_env_var`? It seems more important now that the convention becomes implicit (but still exists)

jasonz-dfinity (approved):

rs/crypto/iccsa/test_utils/BUILD.bazel
nmattia: Here's something from the bazel docs: 

> For example, stubs and fakes that are useful for unit tests may also be useful for integration tests involving the same binaries that will be released to production, and therefore should probably not be marked testonly. Conversely, rules that are dangerous to even link in, perhaps because they unconditionally override normal behavior, should definitely be marked testonly.

This `test_utils` target is an indirect dependency of pocket-ic-lib, meaning anything depending on it has to be `testonly` as well. I think it's not dangerous to depend on it and there is no reason (and very little risk) to do so in prod, so we might as well remove `testonly`

schneiderstefan (approved):

PR #10139, opened 2026-05-08
chore: use GitHub case syntax instead of ternary "if"

nmattia:
Replace ternary ifs (`condition && trueValue || falseValue`) GitHub Actions workflow files with the new first-class `case()` expression (actions/runner#409).

The `case()` expression is now the correct tool for the job and avoids some the "conditional if"
[footguns](https://github.com/actions/runner/issues/409#issuecomment-1013325196).


basvandijk (approved):
Nice

PR #10140, opened 2026-05-08
feat(ckdoge-minter): support ICRC-21 consent messages

Dfinity-Bjoern:
## Summary

Implements the [ICRC-21](https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md) Canister Call Consent Message standard on the ckDOGE minter so that ICRC-21-aware wallets can display a human-readable description of `retrieve_doge_with_approval` calls before the user signs.

Both display variants from the standard are supported:

- **GenericDisplay** — a Markdown message intended for software wallets with a full screen.
- **FieldsDisplay** — a list of `(label, Value)` fields tailored for hardware wallets (e.g. Ledger Nano S+).

Also adds an `icrc10_supported_standards` query that advertises ICRC-10 and ICRC-21.

This mirrors the ckBTC minter implementation from #10093, with the PR-review learnings applied up front rather than as follow-ups.

## Implementation

- New module `rs/dogecoin/ckdoge/minter/src/updates/icrc21.rs` holds the consent-message construction and `icrc10_supported_standards`.
- `main.rs` exposes the two new endpoints (`icrc21_canister_call_consent_message` as `update`, `icrc10_supported_standards` as `query`).
- `ckdoge_minter.did` declares the ICRC-10 / ICRC-21 types and methods.
- The handler enforces the standard's argument-size limit (reusing `MAX_CONSENT_MESSAGE_ARG_SIZE_BYTES` from `icrc-ledger-types`) and returns `Icrc21Error::UnsupportedCanisterCall` for unknown methods or undecodable args.
- **Scope**: only `retrieve_doge_with_approval` produces a consent message; the legacy `retrieve_doge` flow intentionally returns `UnsupportedCanisterCall`.
- **Network-aware token symbols**: `Mainnet ⇒ ckDOGE/DOGE`, `Regtest ⇒ ckTESTDOGE/TESTDOGE`, threaded through the GenericDisplay Markdown, the FieldsDisplay intent, the `TokenAmount` symbol, and the address label.
- **No canister-side pagination**: long values (Dogecoin address, subaccount hex) are emitted as a single `Value::Text` per the ICRC-21 spec; wallets handle pagination (cf. the Ledger ICP app).
- **Markdown-injection defence**: the destination address is parsed via `DogecoinAddress::parse` before any interpolation. Bad addresses (including newlines/backticks/`#` payloads or addresses from the wrong network) are rejected with a generic error and never echoed back.
- ICRC-10 / ICRC-21 URLs point at the canonical `dfinity/ICRC` repo.

## Test plan

- [x] 11 new unit tests in `updates::tests::icrc21` covering: argument-size limit, unsupported method (incl. explicit `retrieve_doge`), decode failures, GenericDisplay output (with and without subaccount), FieldsDisplay output (with and without subaccount), regtest token symbols (both display variants), and Markdown-injection / wrong-network address rejection. Tests live in `updates/tests.rs` (separate from prod code).
- [x] New `test_icrc21_endpoints_smoke` integration test in `tests/tests.rs` installs a mainnet minter via `Setup::new(Network::Mainnet)` and exercises the full Candid path end-to-end: `icrc10_supported_standards` advertises both standards with the canonical URL; `icrc21_canister_call_consent_message` returns the expected GenericDisplay markdown and FieldsDisplay structured message; unsupported methods (including `retrieve_doge`) are rejected.
- [x] `cargo clippy --all-features -p ic-ckdoge-minter -p ic-ckdoge-minter-test-utils` clean with the workspace lint set.
- [x] `bazel test //rs/dogecoin/ckdoge/minter:unit_tests //rs/dogecoin/ckdoge/minter:ckdoge_candid_api_tests //rs/dogecoin/ckdoge/minter:ckdoge_minter_replay_events_tests` all pass (the candid API test verifies the `.did` matches the canister).
- [ ] `bazel test //rs/dogecoin/ckdoge/minter:integration_tests` (Linux x86_64 only — runs in CI; contains the smoke test).
- [ ] Try the FieldsDisplay output on an actual Ledger Nano S+ once a ckDOGE app is available.


rs/dogecoin/ckdoge/minter/ckdoge_minter.did
```diff
@@ -507,4 +570,12 @@ service : (minter_arg : MinterArg) -> {
...
     // NOTE: this method exists for debugging purposes and backwards-compatibility is **not** guaranteed.
     get_events : (record { start: nat64; length : nat64 }) -> (vec Event) query;
+
+    // Returns the list of supported ICRC standards.
+    icrc10_supported_standards : () -> (vec Icrc10StandardRecord) query;
+
+    // Returns a human-readable consent message describing the requested
+    // canister call. See the ICRC-21 standard for details:
+    // https://github.com/dfinity/ICRC/blob/main/ICRCs/ICRC-21/ICRC-21.md
+    icrc21_canister_call_consent_message : (Icrc21ConsentMessageRequest) -> (variant { Ok : Icrc21ConsentInfo; Err : Icrc21Error });
```
gregorydemay: nit: I would put those before maybe before `get_minter_info` since they are more important. Also get_events is a debug endpoint so I think it makes sense to keep it at the end of the list.

rs/dogecoin/ckdoge/minter/BUILD.bazel
```diff
@@ -84,6 +84,7 @@ rust_test(
             # Keep sorted.
             ":ckdoge_minter_lib",
             "//packages/ic-http-types",
+            "//packages/icrc-ledger-types:icrc_ledger_types_storable",
```
gregorydemay, line 87: nit: that dependency should also be added to the  Cargo.toml file.

gregorydemay (approved):
Thanks @Dfinity-Bjoern !

rs/dogecoin/ckdoge/minter/ckdoge_minter.did
Dfinity-Bjoern: Done in 0afc38b37f — moved both ICRC endpoints to sit right after `retrieve_doge_status` and pushed `get_events` back to the end of the service so the debug endpoint is last.

rs/dogecoin/ckdoge/minter/BUILD.bazel
Dfinity-Bjoern: The dependency is already declared in [`rs/dogecoin/ckdoge/minter/Cargo.toml:25`](https://github.com/dfinity/ic/blob/claude/inspiring-neumann-b9d915/rs/dogecoin/ckdoge/minter/Cargo.toml#L25) (`icrc-ledger-types = { path = "../../../../packages/icrc-ledger-types" }`). Cargo binaries share the crate's `[dependencies]` block, so it's already available to `src/main.rs`. The Bazel `rust_canister` target lists deps explicitly per-target though, which is why I had to add the entry in BUILD.bazel separately.

PR #10144, opened 2026-05-08
feat: [DSM-106] Move `observe_canisters_memory_usage()` into background thread

alin-at-dfinity:
Move the logic from `observe_canisters_memory_usage()` into `ReplicatedStateMetrics::observe()`, which now runs in a background thread. Unfortunately, we still need to compute the total canister memory usage on the critical path in order to populate `SubnetMetrics::canister_state_bytes`, but we don't need to do that every round.


rs/messaging/src/message_routing.rs
```diff
@@ -1508,13 +1428,15 @@ impl<RegistryClient_: RegistryClient> BatchProcessor for BatchProcessorImpl<Regi
...
-            .metadata
-            .subnet_metrics
-            .canister_state_bytes = total_memory_usage;
+        // Calculating the total memory usage across all canisters is expensive and
+        // we do not need it to be perfectly accurate. Only do it every 10 rounds.
+        if batch_number.get().is_multiple_of(10) {
+            let total_memory_usage = state_after_round.total_canister_memory_usage();
+            state_after_round
+                .metadata
+                .subnet_metrics
```
mraszyk, line 1437: Wouldn't it make sense to record all [publicly](https://docs.internetcomputer.org/references/ic-interface-spec/#state-tree-subnet) visible metrics consistently only at heights that are multiples of 10, i.e., record all of them at multiples of 10 and nothing in between?

mraszyk (commented):
Left one suggestion, otherwise LGTM

rs/messaging/src/message_routing.rs
alin-at-dfinity: All the `SubnetMetrics` fields except `canister_state_bytes` and `num_canisters` are (sometimes buggy) counters, i.e. they must get updated every round or even multiple times per round (e.g. `update_transactions_total`).

So we'd only be reducing the frequency at which we update `num_canisters`, and it is basically free to update.

I think it makes sense to stick with this approach. Or else, bite the bullet and keep updating it every round, as before. But I don't think the benefits (to the extent that they exist at all) are worth the cost.

mraszyk: Why is it not possible to only update them every 10 rounds?

mraszyk (approved):

rs/messaging/src/message_routing.rs
alin-at-dfinity: As discussed offline, `update_transactions_total` is updated at the end of each round with the number of messages executed during the round; and `consumed_cycles` is updated throughout the round, as messages are sent, timed out, canisters are deleted, etc.

So we'd have to jump through hoops just to keep these numbers hidden and only "publish" them every 10 rounds.

PR #10145, opened 2026-05-08
chore: prune streams to deleted subnets

mraszyk:
This PR implements DSM adjustments to support subnet deletion under the following assumptions:
- only bounded-wait calls between the local and deleted (remote) subnet;
- no cycles attached to those calls.

To this end, two changes are needed:

1. When a subnet is deleted from the registry, the local subnet's outgoing stream to it is now explicitly discarded as part of the same round's execution. This is safe because no certified stream slices from the deleted subnet are included in a block after the outgoing stream is dropped (such slices would require the existence of the outgoing stream for the sake of garbage collection). The stream builder's existing route() returning None for deleted subnets already prevents a fresh stream from being recreated. Prior to this change, stale streams to deleted subnets would persist in ReplicatedState indefinitely.

2. No critical errors are raised for discarded bounded-wait responses that cannot be routed to the deleted subnet (which is now expected).

A new test ensures that the (outbound) stream is indeed deleted, no critical errors are raised when dropping a bounded-wait response to the deleted subnet (both canister request and subnet message), and rejects responses are generated for (bounded-wait) requests already in the output queues to the mgmt canister and regular canister on the deleted subnet. Further tests ensure that the (inbound) certified stream is garbage collected and not included in blocks after the subnet is deleted.


rs/messaging/src/state_machine.rs
```diff
@@ -223,6 +223,9 @@ impl StateMachine for StateMachineImpl {
 
         induction_timer.observe_duration();
 
+        // Discard streams to subnets no longer present in the network topology.
+        state_with_messages.discard_streams_for_deleted_subnets();
```
alin-at-dfinity: Nit: This isn't covered by a "MR phase timer". It's also somewhat randomly invoked before execution. I believe it would fit better somewhere in the "message routing phase" (`PHASE_MESSAGE_ROUTING`) below.

It doesn't matter much, since it's going to be virtually instant anyway, but I just moved things around to ensure that all the work we do in MR is covered by a timer (as was already the case in the scheduler). Because most of

rs/replicated_state/src/replicated_state.rs
```diff
@@ -857,6 +857,22 @@ impl ReplicatedState {
         self.metadata.streams.get(destination_subnet_id)
     }
 
+    /// Discards streams to subnets no longer present in the network topology.
+    ///
+    /// Safe to call because by the time subnet deletion takes effect in the registry,
+    /// all certified stream slices from the deleted subnet have already been inducted
+    /// and no new certified slices can be produced.
```
alin-at-dfinity: Hmm... Is this actually true? (And necessary, if we're going to only allow this for streams with best-effort calls and no attached cycles?)

There's a potential race condition between deleting the subnet and other subnets consuming its streams. Even if we were to first delete all canisters on the subnet to be deleted; or even pause it first; there is no way to guarantee that the streams in both directions (which are reflected by either stream: outgoing messages plus signals for incoming messages) have been completely consumed.

I believe this should simply say that because we only allow this for streams with best-effort calls and no attached cycles, it is OK to drop the stream regardless of contents once the remote subnet is gone.

alin-at-dfinity (approved):

rs/messaging/src/state_machine.rs
mraszyk: https://github.com/dfinity/ic/pull/10145/commits/a92547a8e5520efcdcb9fe1928c12a18d5cef5d3

rs/replicated_state/src/replicated_state.rs
mraszyk: https://github.com/dfinity/ic/pull/10145/commits/5870257553ed8d270732ed2f2e1b29497018f85c

mraszyk: and https://github.com/dfinity/ic/pull/10145/commits/d8c7e89b2eba234014f448355e2b34e2e4412ff9 pointing out that the key property that no stream slices from the deleted subnet are pulled after deleting the outgoing stream

mraszyk: and further hardening the tests: https://github.com/dfinity/ic/pull/10145/commits/8cc78f2c62df39283df787884c3cc34c0bbe1603

alin-at-dfinity (approved):
LGTM, thank you.

PR #10147, opened 2026-05-08
ci(defi): check endpoints exported in a canister's WASM against its Candid specification

gregorydemay:
## Summary

Use `ic-wasm` to check the endpoints of a canister’s WASM exports against its Candid interface. This is to ensure that all exported endpoints are intentional and match the Candid specification, helping to detect any accidental, unexpected or potentially malicious exports.

Wraps the existing `ic-wasm` invocation in `bazel/canisters.bzl`'s `finalize_wasm` rule with an optional `check-endpoints` step, then opts the five DeFi-team canisters in `dfinity/ic` into it. The check fails the build if a canister WASM exports a method that is neither in the candid `service` declaration nor explicitly allowlisted in a per-canister `hidden_endpoints.conf` — catching accidental public exports (a dropped `#[query(hidden = true)]`, a leaked FFI symbol, etc.) before they ship.

Closes DEFI-2501.
Closes DEFI-2502.
Closes DEFI-2503.
Closes DEFI-2504.
Closes DEFI-2511.


mbjorkqvist (approved):

bazel/canisters.bzl
```diff
@@ -181,27 +185,56 @@ def motoko_canister(name, entry, deps, **kwargs):
...
-            .format(input_wasm = "$(location {})".format(src_wasm), ic_wasm = "$(location @crate_index//:ic-wasm__ic-wasm)", version_file = "$(location {})".format(version_file), pigz = "$(location @pigz)", keep_name_section = "--keep-name-section" if keep_name_section else ""),
+        cmd_bash = " && ".join(steps)
+            .format(
+            input_wasm = "$(location {})".format(src_wasm),
+            ic_wasm = "$(location @crate_index//:ic-wasm__ic-wasm)",
+            version_file = "$(location {})".format(version_file),
+            pigz = "$(location @pigz)",
+            keep_name_section = "--keep-name-section" if keep_name_section else "",
+            hidden_endpoints = hidden_endpoints if hidden_endpoints != None else "",
+        ),
```
nmattia: In the future try to avoid refactoring blocks out when adding a feature, it makes reviewing harder

bazel/canisters.bzl
```diff
@@ -181,27 +185,56 @@ def motoko_canister(name, entry, deps, **kwargs):
...
+
+    Args:
+      name: target name; the rule outputs a file `name`.
+      src_wasm: the input wasm label.
+      service_file: optional candid `.did` file to embed as `candid:service` metadata.
+      version_file: text file whose contents become the `git_commit_id` metadata.
+      testonly: testonly attribute on the generated genrule.
+      visibility: visibility of the output target.
+      keep_name_section: pass `--keep-name-section` to every `ic-wasm` invocation.
+      hidden_endpoints: optional `hidden_endpoints.conf` label; enables `ic-wasm check-endpoints --hidden <conf>`.
```
nmattia: Appreciate the added docs; for next time please submit a separate PR for unrelated changes

rs/bitcoin/checker/hidden_endpoints.conf
nmattia, line 1: Are those `.conf` files read by anyone else than ic-wasm? Otherwise I'd be tempted to avoid adding new files and instead add the list of endpoints as starlark string args

bazel/canisters.bzl
```diff
@@ -181,27 +185,56 @@ def motoko_canister(name, entry, deps, **kwargs):
         actual = final_name,
     )
 
-def finalize_wasm(*, name, src_wasm, service_file = None, version_file, testonly, visibility = ["//visibility:public"], keep_name_section = False):
+def finalize_wasm(*, name, src_wasm, service_file = None, version_file, testonly, visibility = ["//visibility:public"], keep_name_section = False, hidden_endpoints = None):
```
nmattia: Instead of modifying the wasm build, why not add a test with `sh_test`? That'll better separate the concerns

bazel/rust.MODULE.bazel
```diff
@@ -771,6 +771,7 @@ crate.spec(
 crate.spec(
     default_features = False,
     features = [
+        "check-endpoints",
```
nmattia, line 778: is it correct to assume that all the new deps are pulled in for this feature?

gregorydemay: yes, I only added that feature. The upgrade to a version that supports that feature was done in #10157.

rs/bitcoin/checker/hidden_endpoints.conf
gregorydemay: > Are those .conf files read by anyone else than ic-wasm? 

No..


> Otherwise I'd be tempted to avoid adding new files and instead add the list of endpoints as starlark string args

I would avoid that. ic-wasm is a tool that's agnostic whether Cargo/Bazel is used, hence require as input such a file (sure the Bazel rule could be expanded to generate such a file on the fly). We also set up this check in different repos where only Cargo is used so it makes sense to stay consistent.

bazel/canisters.bzl
gregorydemay: Good idea! Done in 4d0ab5ee69 — moved to a dedicated `sh_test` (`<canister>_check_endpoints_test`) and reverted `finalize_wasm` back to its master shape. Each `rust_canister(... hidden_endpoints = "...")` now emits a separate test target that runs `ic-wasm check-endpoints` on the finalized wasm, modeled on the existing `inspect_stamped_trivial_wasm` pattern. Call sites are unchanged.

bazel/canisters.bzl
gregorydemay: sorry about that should be now much cleaner thanks to your suggestion below!

bazel/canisters.bzl
gregorydemay: Reverted with the changes form your suggestion below.

nmattia (approved):
LGTM, thanks for the round of edits!

PR #10150, opened 2026-05-09
feat(https_adapter): extending default https outcalls timeout

NikolaMilosa:
Since LLMs take some time, and we still use regular https outcalls, this PR extends the timeout from default 30 seconds to 120 seconds making it much more probable to receive a decent response.

### DIFF
--- a/rs/https_outcalls/adapter/src/config.rs
+++ b/rs/https_outcalls/adapter/src/config.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
 const DEFAULT_HTTP_CONNECT_TIMEOUT_SECS: u64 = 2;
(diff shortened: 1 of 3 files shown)


rs/https_outcalls/adapter/src/config.rs
```diff
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
 const DEFAULT_HTTP_CONNECT_TIMEOUT_SECS: u64 = 2;
-const DEFAULT_HTTP_REQUEST_TIMEOUT_SECS: u64 = 30;
+const DEFAULT_HTTP_REQUEST_TIMEOUT_SECS: u64 = 120;
```
mraszyk, line 6: [This](https://github.com/dfinity/ic/blob/e878738de15674de57a348005b0ece87ec27bc47/rs/config/src/execution_environment.rs#L204) should also be bumped to ensure throughput, @NikolaMilosa .

PR #10154, opened 2026-05-11
chore: Remove NonConsumed cycles use case

dsarlis:
The `NonConsumed` use case is no longer needed after previous refactorings of the code that more clearly separated the cases using an approach with generics. The `NonConsumed` was mainly meant to mark cases which should not be tracked in canister metrics (e.g. transfers of cycles from one canister to another which are not consumed), however, this is now obsolete as is obvious from the code removed, there were a couple of `debug_assert`s verifying that this variant is not present and it appeared in some match statements due to compilation requirements.

Removing the variant from the protobuf enum should also be safe as it should not be present in any checkpoint files.


alin-at-dfinity (dismissed):
LGTM, thank you.

alin-at-dfinity:
Looks like the `compatibility_for_cycles_use_case` test is failing because of the newly missing enum variant.

dsarlis:
> Looks like the `compatibility_for_cycles_use_case` test is failing because of the newly missing enum variant.

Makes sense. I updated the test to remove the missing variant's value from the expected result which should be fine given that it was never serialized (and so it never needs to be deserialized) by the implementation.

alin-at-dfinity (approved):

PR #10155, opened 2026-05-11
chore: Remove deprecated protobuf fields

dsarlis:
Several fields regarding storing `Cycles` prepayments are now obsolete after introducing new versions of them that deal with `CompoundCycles`. The relevant code has been deployed to production for a couple weeks already so we can now remove the fallback code that was trying to backfill data from the old fields.

Additionally, we make the parsing code for a new field `prepayment_for_call_transmission` that was introduced stricter as it should also be populated now in production.

### DIFF
--- a/rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
+++ b/rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
@@ -75,8 +75,8 @@ message Callback {
   types.v1.CanisterId respondent = 7;
   // If non-zero, this is a best-effort call.
   uint32 deadline_seconds = 10;
-  state.queues.v1.CompoundCycles prepayment_for_response_execution = 11;
-  state.queues.v1.CompoundCycles prepayment_for_response_transmission = 12;
+  state.queues.v1.CompoundCycles prepayment_for_response_execution_compound = 11;
+  state.queues.v1.CompoundCycles prepayment_for_response_transmission_compound = 12;
   state.queues.v1.CompoundCycles prepayment_for_call_transmission = 13;
 }
 
@@ -225,7 +225,7 @@ message ExecutionTask {
     }
     // The execution cost that has already been charged from the canister.
     // Retried execution does not have to pay for it again.
-    state.queues.v1.CompoundCycles prepaid_execution_cycles = 7;
+    state.queues.v1.CompoundCycles prepaid_execution_compound_cycles = 7;
   }
 
   message AbortedInstallCode {
@@ -239,7 +239,7 @@ message ExecutionTask {
     optional uint64 call_id = 4;
     // The execution cost that has already been charged from the canister.
     // Retried execution does not have to pay for it again.
-    state.queues.v1.CompoundCycles prepaid_execution_cycles = 5;
+    state.queues.v1.CompoundCycles prepaid_execution_compound_cycles = 5;
   }
 
   oneof task {
--- a/rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
+++ b/rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
@@ -64,6 +64,8 @@ message WasmClosure {
 
 message Callback {
   reserved 6;
+  reserved 8;
+  reserved 9;
 
   uint64 call_context_id = 1;
   WasmClosure on_reply = 2;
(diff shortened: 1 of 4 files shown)


rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
```diff
@@ -64,21 +64,19 @@ message WasmClosure {
...
-  state.queues.v1.Cycles prepayment_for_response_execution = 8;
-  state.queues.v1.Cycles prepayment_for_response_transmission = 9;
   // If non-zero, this is a best-effort call.
   uint32 deadline_seconds = 10;
-  // To replace `prepayment_for_response_execution`.
-  state.queues.v1.CompoundCycles prepayment_for_response_execution_compound = 11;
-  // To replace `prepayment_for_response_transmission`.
-  state.queues.v1.CompoundCycles prepayment_for_response_transmission_compound = 12;
+  state.queues.v1.CompoundCycles prepayment_for_response_execution = 11;
+  state.queues.v1.CompoundCycles prepayment_for_response_transmission = 12;
```
dsarlis: I decided to reuse the old field names to avoid some extra clutter but I can also keep the new names if people prefer.

rs/types/types/src/methods.rs
```diff
@@ -374,44 +363,18 @@ impl TryFrom<pb::Callback> for Callback {
...
-                    )?,
-                    cost_schedule,
-                ),
-            };
-        let prepayment_for_call_transmission = match value.prepayment_for_call_transmission {
-            Some(value) => CompoundCycles::try_from(value)?,
-            // Temporary code until the change is deployed and the field starts
-            // being populated in checkpoints. If no "old" callbacks exist,
-            // this can be changed to fail on `None` instead and always
-            // expect to find some value.
```
dsarlis, line 412: Should be fine to remove this now but if someone can double-check it would be appreciated.

mraszyk: I think so, too, since the current code would turn `None` to `CompoundCycles::new(Cycles::zero(), cost_schedule)` and then serialize to `Some(item.prepayment_for_call_transmission.into())` so that when this PR is deployed, it'd already see that `Some(item.prepayment_for_call_transmission.into())`.

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
mraszyk: I'd suggest to keep the new names, in particular, since we only use them in the conversion to Rust types.

rs/replicated_state/src/canister_state/system_state/proto.rs
```diff
@@ -100,12 +98,7 @@ impl From<&ExecutionTask> for pb::ExecutionTask {
                     task: Some(pb::execution_task::Task::AbortedExecution(
                         pb::execution_task::AbortedExecution {
                             input: Some(input),
-                            prepaid_execution_cycles: Some(
-                                (prepaid_execution_cycles.real()).into(),
-                            ),
-                            prepaid_execution_compound_cycles: Some(PbCompoundCycles::from(
```
mraszyk, line 106: Minor remark: I find the form `PbCompoundCycles::from(...)` better readable.

rs/types/types/src/methods.rs
dsarlis: Do note that there's also [this](https://github.com/dfinity/ic/blob/e878738de15674de57a348005b0ece87ec27bc47/rs/execution_environment/src/execution/response.rs#L216-L223) code which requires state inspection to ensure it can be removed, i.e. checking that values are not really 0 -- this is why it was left out from the current PR.

For the code here, as long as the field has `Some(value)`, even if `value == 0`, it should be fine to drop the temporary code.

mraszyk (approved):

dsarlis:
> Run on [f49d15b](https://github.com/dfinity/ic/commit/f49d15bf05f89441480422595e5656a37a5a9fff) URL: https://github.com/dfinity/ic/actions/runs/25655953578

I noticed a failure for some system tests which seem to have failed due to:
```
failed to setup IC under test: Retried too many times: sending a request to Farm
```
and one which failed with
```
-- could not kill VM: TooManyRetries { message: "sending a request to Farm" }
```
so I would assume it's due to flakiness.

dsarlis:
@mraszyk Can you please retry [this](https://github.com/dfinity/ic/actions/runs/25661248563/job/75322154082) job? It seems there was some issue downloading certain dependencies...

PR #10187, opened 2026-05-12
feat(ic-recovery): allow upgrading without blessing + make some steps non-optional

pierugo-dfinity:
This PR makes the following changes to `ic-recovery`:
- Allow to upgrade to a version without first blessing/electing it in case we want to do a simple rollback. This was possible by supplying `--skip BlessVersion` but the tool now asks for the upgrade question again during `UpgradeVersion` if not supplied in `BlessVersion`. If it was actually supplied in the latter, then it re-uses it during `UpgradeVersion` (unchanged).
- Makes `DownloadConsensusPool`, `DownloadState` and `UploadState` non-optional by asking for the input over and over again if not supplied/invalid. Note that it is still possible to hard skip them with `--skip ...`.
  - `DownloadState` and `UploadState` could be optional if we know for sure that there no blocks that were executed after a checkpoint (like at an upgrade height, for example if there is a bug in the orchestrator). Though this is relatively rare compared to bugs in the replica + I would argue that we could probably still want to download it to manually inspect it.
  - `DownloadConsensusPool` could also be optional if we have not downloaded any state. But if we did, even if no blocks are replayed, we still need it for the CUP, which validates the downloaded state.

### DIFF
--- a/rs/recovery/src/app_subnet_recovery.rs
+++ b/rs/recovery/src/app_subnet_recovery.rs
@@ -1,7 +1,7 @@
 use crate::{
     DataLocation, NeuronArgs, Recovery, RecoveryArgs, RecoveryResult, Step,
     cli::{
-        consent_given, print_height_info, read, read_optional, read_optional_data_location,
+        consent_given, print_height_info, read, read_data_location, read_optional,
         read_optional_node_ids, read_optional_subnet_id, read_optional_type, wait_for_confirmation,
     },
     error::{GracefulExpect, RecoveryError},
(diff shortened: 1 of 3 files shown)


rs/recovery/src/app_subnet_recovery.rs
```diff
@@ -324,16 +324,16 @@ impl RecoveryIterator<StepType, StepTypeIter> for AppSubnetRecovery {
...
                     self.params.download_pool_node =
-                        read_optional(&self.logger, "Enter consensus pool download IP:");
+                        Some(read(&self.logger, "Enter consensus pool download IP:"));
                 }
             }
 
             StepType::DownloadState => {
                 if self.params.download_state_method.is_none() {
-                    self.params.download_state_method = read_optional_data_location(
+                    self.params.download_state_method = Some(read_data_location(
```
eichhorl, line 333: > DownloadState and UploadState could be optional if we know for sure that there no blocks that were executed after a checkpoint (like at an upgrade height, for example if there is a bug in the orchestrator). Though this is relatively rare

Historically speaking I would say upgrade failures are actually the most common, since they happen whenever there is a version incompatibility (i.e. between CUPs or in the replicated state). But in thoses cases we usually don't even run `ic-recovery` so it should probably be fine.

eichhorl (approved):

rs/recovery/src/app_subnet_recovery.rs
pierugo-dfinity: I see, sounds good 👍

PR #10190, opened 2026-05-12
chore: inline P2P, HTTP endpoints & Monitoring bazel crate dependencies

nmattia:


mraszyk (approved):

pierugo-dfinity:
Concerning Copilot's suggestion, it looks like the [commit](https://github.com/dfinity/ic/commit/30949dfcb26b3de36c67ed1e4c503f14297c1a19#diff-ac7b42caad0a5a95818c6f4d62c9c71404cf2cbb355a141f92d9231a07facb3c) introducing the comment had the same dependencies as today (modulo protobuf) so it is fine. I guess he has a point that the comment is in the middle of dependencies.

We could also rename `NET` to `Core Protocol`.

nmattia:
@pierugo-dfinity I've accepted the suggestions as is. I would have preferred to leave the original comments untouched and handle this separately but I guess this is fine. I'll let you follow up with another PR if you need the comment moved/refined/etc.

PR #10191, opened 2026-05-12
chore: inline DSM bazel crate dependencies

nmattia:


mraszyk (approved):
Approving without

PR #10198, opened 2026-05-12
chore: account for heap delta increase due to canister log buffer resize

mraszyk:
This PR accounts for heap delta increase due to canister log buffer resize. The heap delta increase is equal to the new log buffer memory usage since resizing might reshuffle the existing log entries in memory. In particular, there's no heap delta increase at canister creation time since there are no entries to reshuffle in memory.

This PR also fixes `Ic00MethodPermissions` to enable instructions charges in round limits due to canister log buffer resize.


rs/execution_environment/src/canister_manager.rs
```diff
@@ -335,7 +335,9 @@ impl CanisterManager {
         subnet_size: usize,
         cost_schedule: CanisterCyclesCostSchedule,
         metrics: Option<&ExecutionEnvironmentMetrics>,
-    ) -> Result<(), CanisterManagerError> {
+    ) -> Result<NumBytes, CanisterManagerError> {
```
mraszyk, line 342: doc comment

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -5429,6 +5429,181 @@ fn upload_chunk_increases_subnet_heap_delta() {
...
+        canister_id: canister_id.get(),
+        settings: CanisterSettingsArgsBuilder::new()
+            .with_freezing_threshold(100)
+            .build(),
+        sender_canister_version: None,
+    }
+    .encode();
+    test.subnet_message(Method::UpdateSettings, args).unwrap();
+
+    assert_eq!(test.state().metadata.heap_delta_estimate, NumBytes::from(0));
```
mraszyk: also assert zero before update settings in all tests

rs/execution_environment/src/canister_manager.rs
```diff
@@ -553,6 +555,9 @@ impl CanisterManager {
                 .system_state
                 .log_memory_store
                 .would_resize(requested_limit.get() as usize);
+            if log_resize_needed {
+                heap_delta_increase = canister.log_memory_store_memory_usage();
+            }
```
alin-at-dfinity: Nit: At first sight, this looked incorrect to me: we were already using some memory for the logs, so it can't be all "increase". Took me a while to realize that it's a "heap delta increase".

Maybe a one-line comment would help.

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -5429,6 +5429,181 @@ fn upload_chunk_increases_subnet_heap_delta() {
...
+}
+
+// log_memory_limit decreased → would_resize = true → heap delta = old log store size.
```
alin-at-dfinity: A couple of questions:
1. Should deleting 2 MB and writing 1 MB count as 2 MB worth of heap delta?
2. Would we actually write to disk 1 MB of zeroes (or whatever) in this particular instance?

alin-at-dfinity (approved):

rs/execution_environment/src/canister_manager.rs
mraszyk: I actually extended the comment further above and noticed that using `log_bytes_used` for the heap delta increase makes more sense for consistency with instruction charging. Ensuring test coverage for that then revealed some issues with instruction charging (that was untested before) which are also fixed in this PR.

rs/execution_environment/src/canister_manager/tests.rs
mraszyk: The log memory store is backed by `PageMap` so zeros are implicit. It actually seems reasonable to take the newly used bytes as the heap delta since we have to dirty all the newly used bytes in the worst case.

alin-at-dfinity (approved):
Still LGTM, thanks.

PR #10199, opened 2026-05-12
fix: update lockfiles

nmattia:
The lockfiles were out of date. This is the result of running `./bin/bazel-pin.sh`.


nmattia:
Superseded by #10201

PR #10200, opened 2026-05-12
chore: use PerformanceOptimizedAllocation for all performance tests

basvandijk:
In https://github.com/dfinity-lab/farm/pull/276 we would like to increase the maximum vCPUs for performance hosts from 64 to the default 256 such that we can run more regular system-tests on them.

However, because of the extra room, this will cause performance system-tests to allocate more than one VM per host which will make their results unreliable. 

To work around this Farm recently gained a new `VmAllocationMode` called `PerformanceOptimizedAllocation` (https://github.com/dfinity-lab/farm/pull/275) which behaves the same as the default `MinIntraDistanceLoadBalanceAllocation` except that it doesn't have the first ordering property that compares the number of VMs of the group that the host has already allocated. This will have the effect of load-balancing VMs over the hosts in a single DC instead of colocating them to single hosts as much as possible.

This commit configures all performance system-tests to use this new `VmAllocationMode::PerformanceOptimizedAllocation`.

This also removes the ability to configure the `VmAllocationMode` per subnet and node since we never used that ability.

### DIFF
--- a/rs/tests/boundary_nodes/api_bn_performance_test.rs
+++ b/rs/tests/boundary_nodes/api_bn_performance_test.rs
@@ -1,10 +1,12 @@
 use anyhow::Result;
 use ic_boundary_nodes_performance_test_common::{query_calls_test, setup, update_calls_test};
+use ic_system_test_driver::driver::farm::VmAllocationMode;
 use ic_system_test_driver::{driver::group::SystemTestGroup, systest};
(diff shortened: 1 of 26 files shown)


frankdavid (approved):

mraszyk (approved):

nmattia (approved):
Rubber stamp

PR #10202, opened 2026-05-12
feat: [DSM-106] Check scheduler invariants in a background thread 

alin-at-dfinity:
In benchmarks ran after replicated state metrics (including memory usage) have been moved to a background thread, `SchedulerImpl::check_invariants()` accounts for about 15% of `SchedulerImpl::execute_round()` time. Moving it into the background metrics thread is a bit more involved, as it requires plumbing through the execution config; and introduces the possibility that the invariant check won't run after every single round or in all tests; but this is a debugging mechanism, so always having it on the critical path is overkill.


rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -3046,7 +3046,7 @@ impl PocketIc {
                             tokio::sync::watch::channel(ic_types::Height::from(0)).0,
                             Some(ReplicatedStateInvariants::new(
                                 &metrics_registry,
-                                &icp_config.hypervisor_config,
+                                &execution_environment::Config::default(),
                             )),
```
mraszyk: This StateManager is only temporary to parse state metadata so we could as well just pass `None`.

mraszyk (approved):

jasonz-dfinity (approved):
LGTM for rs/registry

basvandijk (approved):

rs/pocket_ic_server/src/pocket_ic.rs
alin-at-dfinity: Good point, done.

PR #10206, opened 2026-05-12
feat(PocketIC): canister creation with cycles

mraszyk:
This PR makes the existing PocketIC (Rust) library functions create canisters with a default amount of 100T cycles and adds a new function `PocketIc::create_canister_with_params` to create a canister with parameters specified in a struct with optional fields to use defaults as appropriate.

The motivation for this change is that the (upcoming) canister log memory store feature allocates canister log buffer memory at canister creation time and thus creating a canister with zero cycles is no longer allowed.

### DIFF
--- a/packages/pocket-ic/src/lib.rs
+++ b/packages/pocket-ic/src/lib.rs
@@ -574,6 +574,24 @@ impl TryFrom<Time> for SystemTime {
     }
 }
 
+/// Specifies where to place a newly created canister.
+#[derive(Debug)]
+pub enum CreateCanisterPlacement {
+    /// Place the canister on the given subnet.
+    SubnetId(SubnetId),
+    /// Create the canister with the given specific canister ID.
+    CanisterId(CanisterId),
+}
+
+/// Parameters for [`PocketIc::create_canister_with_params`].
+#[derive(Debug, Default)]
+pub struct CreateCanisterParams {
+    /// Initial cycles balance; defaults to 100T if `None`.
+    pub cycles: Option<u128>,
+    pub settings: Option<CanisterSettings>,
+    pub placement: Option<CreateCanisterPlacement>,
+}
+
 /// Main entry point for interacting with PocketIC.
 pub struct PocketIc {
     pocket_ic: PocketIcAsync,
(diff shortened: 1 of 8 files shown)


mbjorkqvist (approved):

packages/pocket-ic/src/lib.rs
```diff
@@ -574,6 +574,24 @@ impl TryFrom<Time> for SystemTime {
...
+#[derive(Debug)]
+pub enum CreateCanisterPlacement {
+    /// Place the canister on the given subnet.
+    SubnetId(SubnetId),
+    /// Create the canister with the given specific canister ID.
+    CanisterId(CanisterId),
+}
+
+/// Parameters for [`PocketIc::create_canister_with_params`].
+#[derive(Debug, Default)]
```
eichhorl: Since this is a public type, should we derive `Clone`?

packages/pocket-ic/src/lib.rs
```diff
@@ -574,6 +574,24 @@ impl TryFrom<Time> for SystemTime {
...
+pub struct CreateCanisterParams {
+    /// Initial cycles balance; defaults to 100T if `None`.
+    pub cycles: Option<u128>,
+    pub settings: Option<CanisterSettings>,
+    pub placement: Option<CreateCanisterPlacement>,
```
eichhorl, line 594: Should we also document what `None` means for the other two parameters?

packages/pocket-ic/src/nonblocking.rs
```diff
@@ -1109,54 +1090,81 @@ impl PocketIc {
...
-        .map(|(x,)| x)
-        .unwrap();
-        canister_id
+        .map(|(x,)| x);
+        match res {
+            Ok(CanisterIdRecord {
+                canister_id: actual_canister_id,
+            }) => Ok(actual_canister_id),
+            Err(e) => Err(format!("{e:?}")),
+        }
```
eichhorl: Alternatively
```suggestion
        .map(|(x,)| x.canister_id)
        .map_err(|e| format!("{e:?}"))
```

eichhorl (approved):

PR #10215, opened 2026-05-13
feat: Allow xnet on engines

schneiderstefan:
This commit opens engines up to send and receive XNet messages with 2
restrictions on messages that involve engines (engine->subnet,
subnet->engine, engine->engine, but not messages staying on the same
engine):
1. Only bounded wait calls are allowed
2. Messages cannot contain attached cycles

Messages that are not allowed will be rejected by the protocol and the
canister receives an error.

The implementation is in 3 places:
1. The StreamBuilder rejects these messages on the sending side
2. The XNet PayloadBuilder rejects these messages on the receiving side.
No honest subnet would send any of these messages, but this protects
from malicious subnets.
3. The NetworkTopology now contains all other subnets again. Previously,
when engines were not able to do any XNet, the NetworkTopology would
only contain other subnets if it could send messages to them. The
exception to this is the list of ecsda subnets, which does not contain
other subnets that do not allow sending cycles to them. This is because
the threshold ecdsa endpoint expect to be called with cycles attached.


rs/messaging/src/message_routing/tests.rs
```diff
@@ -1914,31 +1905,163 @@ fn try_read_registry_nns_subnet_has_full_topology_with_engines() {
...
             .subnets_for_certification()
             .keys()
             .copied()
             .collect();
-        assert!(all_keys.contains(&app_subnet_id));
-        assert!(all_keys.contains(&engine_subnet_id));
-        assert!(all_keys.contains(&nns_subnet_id));
+        assert!(cert_keys.contains(&app_subnet_id));
+        assert!(cert_keys.contains(&engine_subnet_id));
+        assert!(cert_keys.contains(&nns_subnet_id));
```
alin-at-dfinity, line 1922: This appears to be the same as the previous two tests (only written differently). Do we still need `full_topology` now that all subnets see all subnets in every "view"?

rs/messaging/src/message_routing.rs
```diff
@@ -1130,6 +1111,22 @@ impl<RegistryClient_: RegistryClient> BatchProcessorImpl<RegistryClient_> {
...
+        // always both. The local subnet is always reachable (loopback); any other
+        // subnet is reachable only if neither end is a CloudEngine. Pruning such
+        // subnets here means routing returns a clean `ChainKeyError` (and the cost
+        // API an `UnknownKey`) instead of routing a doomed request to the boundary.
+        let chain_key_subnet_is_reachable = |id: &SubnetId| match subnets.get(id) {
+            None => false,
+            Some(topo) => {
+                *id == own_subnet_id
+                    || (own_subnet_type != SubnetType::CloudEngine
+                        && topo.subnet_type != SubnetType::CloudEngine)
```
alin-at-dfinity, line 1125: Just wondering (and potentially showing my ignorance of chain keys) would it eventually make sense for cloud engines to also use chain keys provided by other cloud engines? (Although that would likely require a way of designating which cloud engines have access to which cloud engines' chain keys, since they're not going to pay using cycles.)

So very much a philosophical question.

rs/messaging/src/message_routing.rs
```diff
@@ -1070,43 +1077,17 @@ impl<RegistryClient_: RegistryClient> BatchProcessorImpl<RegistryClient_> {
...
-                .collect::<BTreeMap<_, _>>()
-                .try_into()
-                .map_err(|err| Persistent(format!("'filtered routing table', err: {:?}", err)))?;
-            (subnets, routing_table)
-        };
+        // All subnets see the full topology, including CloudEngine subnets.
+        let subnets: BTreeMap<_, _> = all_subnets
+            .iter()
+            .map(|(id, topo)| (*id, topo.clone()))
+            .collect();
```
alin-at-dfinity, line 1084: To answer my earlier question, `subnets` now looks to be an identical copy of `all_subnets` (which later becomes `full_subnets`). I.e. `full_subnets` appears no longer necessary.

OTOH, I went down the rabbit hole and asked AI to drop `full_subnets` / `FullSubnets` altogether (here it is, FWIW: #10411). Problem is, doing it all at once would break backward compatibility, which is why, I imagine, you didn't do it in this PR. We could try to be smart about it and encode the regular topology into `full_topology` for the NNS only; or we could just leave it as is for another release and then drop it altogether. By that point it would be dead code (or rather data), so it wouldn't matter if it simply disappeared (particularly since, as I just tested it, not even doing it all in one go breaks any compatibility test). )o:

rs/messaging/src/message_routing.rs
alin-at-dfinity: We no longer filter out subnets:

```suggestion
// Look up the default subnet for `SetupInitialDKG`. The key may be
// unset, in which case `SetupInitialDKG` requests fall back to the
// calling subnet.
```

alin-at-dfinity (commented):
First round of comments, still going through it.

rs/messaging/src/message_routing/tests.rs
```diff
@@ -1914,31 +1905,163 @@ fn try_read_registry_nns_subnet_has_full_topology_with_engines() {
...
+    };
+
+    let mut routing_table = RoutingTable::new();
+    routing_table_insert_subnet(&mut routing_table, app_subnet_id).unwrap();
+    routing_table_insert_subnet(&mut routing_table, engine_subnet_id).unwrap();
+    routing_table_insert_subnet(&mut routing_table, nns_subnet_id).unwrap();
+
+    let chain_key_enabled_subnets = btreemap! {
+        shared_chain_key() => Valid(vec![app_subnet_id, engine_subnet_id]),
+        engine_only_chain_key() => Valid(vec![engine_subnet_id]),
```
alin-at-dfinity, line 1968: Clueless question: do system subnets have access to / use chain keys?

rs/messaging/src/message_routing.rs
schneiderstefan: Ah sorry already responded to the other PR. Yes indeed the reason why I didn't just drop it in this PR is because upgrade/downgrades are a bit tricky.

rs/messaging/src/message_routing.rs
schneiderstefan: It's a valid question. I think if we go down the route of specifying which engines can talk to which other engines, then threshold signing by another engine makes sense. In the short term, I think only loopback signings really fit the model.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -799,6 +799,260 @@ fn build_streams_with_best_effort_messages() {
     }
 }
 
+/// Tests that a guaranteed-response request from a CloudEngine subnet (own subnet) to a
+/// non-engine subnet is rejected with a synthetic reject response.
+#[test]
+fn build_streams_engine_src_rejects_gr_request() {
```
alin-at-dfinity: Nit: AFAIK we've never used the GR abbreviation before (although I might have just missed it). It took me a second to parse it. I'd rather just spell it out.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -799,6 +799,260 @@ fn build_streams_with_best_effort_messages() {
...
+    let local_canister_id = canister_test_id(0);
+    let remote_canister_id = canister_test_id(1);
+    with_test_replica_logger(|log| {
+        let response = Arc::new(Response {
+            originator: remote_canister_id,
+            respondent: local_canister_id,
+            originator_reply_callback: CallbackId::from(1),
+            refund: Cycles::new(100),
+            response_payload: Payload::Data(vec![]),
+            deadline: NO_DEADLINE,
```
alin-at-dfinity: Nit: Maybe we should make this a best-effort response, to make it clear that it's dropped because of the refund, not because it's guaranteed response.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -799,6 +799,260 @@ fn build_streams_with_best_effort_messages() {
...
+            // The refund must NOT have been routed into the REMOTE_SUBNET stream.
+            let routed_refunds = result_state
+                .streams()
+                .get(&REMOTE_SUBNET)
+                .map_or(0, |s| s.refund_count());
+            assert_eq!(
+                0, routed_refunds,
+                "Refund leaked across engine boundary (own_subnet_type={own_subnet_type:?}, \
+                 remote_subnet_type={remote_subnet_type:?})",
+            );
```
alin-at-dfinity, line 1126: Should we also check that the refund is also gone from the refund pool (if that's what it's called)?

rs/messaging/src/routing/stream_handler/tests.rs
```diff
@@ -3235,6 +3292,368 @@ fn process_stream_slices_with_invalid_messages() {
...
+            metrics.assert_eq_critical_errors(CriticalErrorCounts::default());
+        },
+    );
+}
+
+/// Tests that a best-effort request carrying cycles from a CloudEngine subnet is dropped
+/// at the engine boundary: cycles observed as lost, accept signal pushed, and a critical
+/// error raised. Mirrors the sender-side test
+/// `build_streams_engine_src_rejects_cycles_request` on the receiving side, which is
+/// the security-critical filter against a malicious engine.
```
alin-at-dfinity: Just thinking out loud:
1. Should we be raising critical errors (i.e. page) if an engine misbehaves?
2. Should we be recording such cycles as lost to begin with? (Assuming it's trivial to not do so. I don't think this matters enough to add another 100 or even 50 lines of code.)

alin-at-dfinity: Further thinking out loud: It may be possible to turn these tests (and the similar ones in `stream_builder/tests.rs`. I.e. have an `induct_stream_slices_test_impl()` that takes as arguments the types of the two subnets and the message induct / route. Then either have a `for` loop with sets of inputs and a `match` statement for the expected outcomes (critical error / metric bump / success); or just write a whole lot of short tests that call the helper and test the outcome.

Just an idea, feel free to leave it as is. It would make it simpler to write tests for every single combination. But it's also quite a bit of extra work.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -487,9 +491,52 @@ impl StreamBuilderImpl {
...
+                            engine_requests_to_reject.push(req);
+                        }
+
+                        // Response that should not exist at an engine boundary: a
+                        // guaranteed-response response, or one carrying cycles. Neither
+                        // guaranteed-response calls nor cycles are allowed across the
+                        // boundary, so such a response can only come from a buggy local
+                        // canister. Drop it; any cycles are lost. Kept above the
+                        // oversized-response arm, which truncates and routes; matching
+                        // here first ensures such a response is dropped, not routed.
```
alin-at-dfinity: This looks more like a critical error. A buggy canister can neither change the deadline of a call; nor attach a refund if the payment was zero.

We should still retain the behavior (modulo cycles lost) to avoid serious breakage in case a bug does sneak in, but this branch should never be reached in the absence of serious bugs.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -685,11 +750,27 @@ impl StreamBuilderImpl {
...
                     let is_loopback_stream = dst_subnet_id == self.subnet_id;
+                    let is_engine_dst = !is_loopback_stream
+                        && network_topology
+                            .subnets()
+                            .get(&dst_subnet_id)
+                            .is_some_and(|t| t.subnet_type == SubnetType::CloudEngine);
+                    let is_engine_src = !is_loopback_stream && own_is_engine;
+                    if is_engine_dst || is_engine_src {
+                        // Refund destined to cross the engine boundary: drop, cycles lost.
+                        cycles_lost += refund.amount();
```
alin-at-dfinity, line 809: Same here: this sounds more like a bug than lost cycles. An (actual cross-subnet) refund implies that a message with cycles made it across in the first place.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -750,6 +784,27 @@ impl StreamHandlerImpl {
...
+                // Refunds bypass sender validation; apply the engine filters here.
+                if is_at_engine_boundary {
+                    error!(
+                        self.log,
+                        "{}: Dropping refund from {}: {:?}",
+                        CRITICAL_ERROR_ENGINE_MESSAGE,
+                        remote_subnet_id,
+                        refund,
+                    );
+                    self.metrics.critical_error_engine_message.inc();
```
alin-at-dfinity, line 797: As already observed elsewhere, this could reasonably be replaced by a plain counter. We want to learn about it if it happens, but I don't know if we want to be paged.

Although it's probably best to just leave it as is and wait and see if it turns into a source of spam. Very likely not. We can always silence it in Alertmanager if it does.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -762,6 +817,74 @@ impl StreamHandlerImpl {
...
+                        RequestOrResponse::Request(req) if req.deadline == NO_DEADLINE => {
+                            error!(
+                                self.log,
+                                "{}: Rejecting guaranteed-response request at engine boundary (from {}): {:?}",
+                                CRITICAL_ERROR_ENGINE_MESSAGE,
+                                remote_subnet_id,
+                                req,
+                            );
+                            self.metrics.critical_error_engine_message.inc();
+                            self.observe_inducted_message_status(msg_type, LABEL_VALUE_DROPPED);
```
alin-at-dfinity: `dropped` is likely the wrong status here. We have definitely not dropped the request, we only rejected it.

We probably need a new status label to do with engines (`EngineNotAllowed`?). And if we add that, we might as well also use it for responses and refunds too. (Using `dropped` for refunds that we couldn't apply further below, way before your change, was likely also an error. It should have been `CanisterNotFound`, because that's what it actually is. `dropped` should only apply to best-effort responses that arrived too late.)

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -762,6 +817,74 @@ impl StreamHandlerImpl {
...
+                            state.observe_lost_cycles_due_to_dropped_messages(CompoundCycles::new(
+                                req.payment,
+                                own_cost_schedule,
+                            ));
+                            stream.push_accept_signal();
```
alin-at-dfinity: I believe this is a bug:
```suggestion
                            stream.push_reject_signal();
```

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -762,6 +817,74 @@ impl StreamHandlerImpl {
...
+                                CRITICAL_ERROR_ENGINE_MESSAGE,
+                                remote_subnet_id,
+                                req,
+                            );
+                            self.metrics.critical_error_engine_message.inc();
+                            self.observe_inducted_message_status(msg_type, LABEL_VALUE_DROPPED);
+                            state.observe_lost_cycles_due_to_dropped_messages(CompoundCycles::new(
+                                req.payment,
+                                own_cost_schedule,
+                            ));
```
alin-at-dfinity: Nit: Regarding recording cycles coming from engines as lost, we would also need to do this for guaranteed response requests above too. I think we're better off not doing it at all.

Actually, having looked a bit closer, this `match` branch and the one above should only differ in the critical error log message. But that already includes the request, so we may make it generic and unify the two branches.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -762,6 +817,74 @@ impl StreamHandlerImpl {
...
+                        // exist (GR or carrying cycles) is dropped just below.
+                        _ => {}
+                    }
+                    // A response that should not exist at the engine boundary: a
+                    // guaranteed-response response, or one carrying cycles. Neither is
+                    // allowed across the boundary, so this can only come from a buggy or
+                    // malicious peer. Drop it (any cycles are lost) and raise a critical
+                    // error.
+                    if let RequestOrResponse::Response(ref rep) = msg {
+                        let is_gr = rep.deadline == NO_DEADLINE;
```
alin-at-dfinity: Nit: Same observation regarding the GR abbreviation. You can either expand it or not bother with the intermediate boolean and just plug `rep.deadline == NO_DEADLINE` into the `if` condition below.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -762,6 +817,74 @@ impl StreamHandlerImpl {
...
+                                rep,
+                            );
+                            self.metrics.critical_error_engine_message.inc();
+                            self.observe_inducted_message_status(msg_type, LABEL_VALUE_DROPPED);
+                            if has_cycles {
+                                state.observe_lost_cycles_due_to_dropped_messages(
+                                    CompoundCycles::new(rep.refund, own_cost_schedule),
+                                );
+                            }
+                            stream.push_accept_signal();
```
alin-at-dfinity: Taking this even further, this branch only differs from the (two) `Request` branch(es) in that it produces an `Accept` instead of `Reject(EngineNotAllowed)`.

Even if you don't feel like unifying them all, you can still put them under the same `match` block, there appears to be no need for a separate `if`.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -1210,6 +1340,10 @@ fn generate_reject_response_for(reason: RejectReason, request: &Request) -> Requ
             RejectCode::SysFatal,
             "Inducting request failed due to an unknown error".to_string(),
         ),
+        RejectReason::EngineNotAllowed => (
+            RejectCode::SysFatal,
+            "Guaranteed-response calls from CloudEngine subnets are not allowed".to_string(),
```
alin-at-dfinity: Nit:

```suggestion
            "Guaranteed-response calls and cycle transfers to / from CloudEngine subnets are not allowed".to_string(),
```

rs/canonical_state/src/encoding/types.rs
```diff
@@ -67,6 +67,8 @@ pub struct RejectSignals {
     pub out_of_memory_deltas: Vec<u64>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub unknown_deltas: Vec<u64>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub engine_not_allowed_deltas: Vec<u64>,
```
alin-at-dfinity, line 71: Damn, this just occurred to me now: whether we go with an explicit certification version or not, a `canonical_state` change requires a staged rollout: first deploy the new "variant" (to all subnets, in this case, since it affects stream structure) with the full logic required to handle it (in this case, to inflate the signal into a reject response, so not much); and only then deploy replica binaries that may produce it.

If we're to do this properly, the new certification version would be introduced with the field; and it would become the "current version" once we have code that can produce it. An explicit version would also allow us to rely on tests to tell us when it's safe to merge follow-up changes.

You should also make a copy of the original `RejectSignals` type (no `engine_not_allowed_deltas` field) under `rs/canonical_state/encoding/old_types.rs` (likely as `RejectSignalsV25`) and update `encoding/tests/compatibility.rs` and `encoding/tests/test_fixtures.rs` to also cover the new variant.

rs/replicated_state/src/metadata_state/tests.rs
```diff
@@ -2287,7 +2287,7 @@ fn compatibility_for_reject_reason() {
         RejectReason::iter()
             .map(|reason| reason as i32)
             .collect::<Vec<i32>>(),
-        [1, 2, 3, 4, 5, 6, 7]
+        [1, 2, 3, 4, 5, 6, 7, 8]
```
alin-at-dfinity, line 2291: Similar (but less stringent) observation here as for `rs/canonical_state`. As per `rs/replicated_state/best-practices-replicated-state.md`, you should first deploy the `ic_types` and `protobuf` additions, without using them anywhere. Only once that has made it into a replica release can you proceed with the rest of the changes.

alin-at-dfinity (commented):
More nitpicks, for the most part.

rs/xnet/payload_builder/src/lib.rs
```diff
@@ -489,16 +489,7 @@ impl XNetPayloadBuilderImpl {
...
-            .into_iter()
-            .filter(|subnet_id| {
-                self.registry
-                    .get_subnet_type(*subnet_id, validation_context.registry_version)
-                    .is_ok_and(|maybe_t| {
-                        maybe_t.is_some_and(|t| t != SubnetType::CloudEngine.into())
-                    })
-            })
-            .collect();
+        let subnet_ids: Vec<_> = all_subnet_ids.into_iter().collect();
```
alin-at-dfinity: `all_subnet_ids` is already a `Vec<SubnetId>`.
```suggestion
        let subnet_ids = self
            .registry
            .get_subnet_ids(validation_context.registry_version)
            .map_err(Error::RegistryGetSubnetsFailed)?
            .unwrap_or_default();
```

rs/tests/message_routing/xnet/xnet_cloud_engine_isolation_test.rs
alin-at-dfinity: Not part of your change, but these two appear redundant. AFAICT there is no difference between "CloudEngine 1" and "CloudEngine 2".
```suggestion
```

rs/tests/message_routing/xnet/xnet_cloud_engine_isolation_test.rs
```diff
@@ -306,10 +315,10 @@ fn test(env: TestEnv) {
             ("CloudEngine 2", &uc_ce_2a, "CloudEngine 1", &uc_ce_1a),
```
alin-at-dfinity: Ditto.
```suggestion
```

rs/xnet/payload_builder/src/tests.rs
```diff
@@ -555,77 +576,50 @@ async fn cloud_engine_validate_rejects_non_empty_payload() {
...
-async fn validate_rejects_slice_from_cloud_engine_subnet() {
+async fn validate_accepts_slice_from_cloud_engine_subnet() {
     with_test_replica_logger(|log| {
         let cloud_engine_subnet = SUBNET_1;
         let fixture = PayloadBuilderTestFixture::with_xnet_state_and_subnet_types(
-            1,
+            0,
             btreemap![cloud_engine_subnet => SubnetType::CloudEngine],
             None,
         );
```
alin-at-dfinity, line 605: Rant: This fixture is amazingly hard to read. Does this mean we get 4 subnets out of which `SUBNET_1` is a `CloudEngine`? And that `own_subnet_type` defaults to `Application`?

alin-at-dfinity (commented):
All done, LGTM modulo the one bug and the staged rollout.

rs/messaging/src/message_routing/tests.rs
schneiderstefan: They do have access, in fact there is even a special case where threshold keys from the NNS are free. I don't know to what degree they use it though.

rs/messaging/src/routing/stream_builder/tests.rs
schneiderstefan: I spelled it out everywhere.

rs/messaging/src/routing/stream_builder/tests.rs
schneiderstefan: Good point. Done.

rs/messaging/src/routing/stream_builder/tests.rs
schneiderstefan: Added.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: I expanded the variable name.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: Refactored.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: Done in the earlier PR

rs/replicated_state/src/metadata_state/tests.rs
schneiderstefan: Done in the earlier PR

rs/xnet/payload_builder/src/lib.rs
schneiderstefan: Thanks. That seems to have been a leftover from when full_topology still meant something. I fixed it here now, but obviously going forward full_topology will be removed anyway.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: I changed it to EngineNotAllowed and also changed the other one you mentioned to CanisterNotFound

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: It is now rejected.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: Refactored

rs/messaging/src/routing/stream_builder.rs
schneiderstefan: Good point. I added the critical here here too.

rs/messaging/src/routing/stream_builder.rs
schneiderstefan: I added the critical error here too, but I kept the lost cycles accounting in addition.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -1,5 +1,6 @@
 use crate::message_routing::{
-    CRITICAL_ERROR_INDUCT_RESPONSE_FAILED, LatencyMetrics, MessageRoutingMetrics,
+    CRITICAL_ERROR_ENGINE_MESSAGE, CRITICAL_ERROR_INDUCT_RESPONSE_FAILED, LatencyMetrics,
```
alin-at-dfinity: Apologies for the late comment (this was already in the code I reviewed), but could we rename this to something like "illegal engine message"? (I.e. add an "illegal" prefix.)

As is, it makes it sound like any engine message will trigger a critical error.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -512,24 +521,31 @@ impl StreamBuilderImpl {
...
                                 && (rep.deadline == NO_DEADLINE || rep.refund > Cycles::zero()) =>
                         {
+                            error!(
+                                self.log,
+                                "{}: Dropping engine-boundary response (to {}): {:?}",
+                                CRITICAL_ERROR_ENGINE_MESSAGE,
+                                dst_subnet_id,
+                                rep,
+                            );
+                            self.metrics.critical_error_engine_message.inc();
```
alin-at-dfinity, line 551: This just occurred to me: I'm wondering if in the case of a bug (which this would have to be) it wouldn't be better to deliver the response rather than drop it. Particularly if it's a guaranteed delivery response, since the caller would then be stuck forever waiting for it.

You can strip the refund if you want, for good measure.

rs/messaging/src/routing/stream_handler.rs
```diff
@@ -820,52 +821,38 @@ impl StreamHandlerImpl {
...
-                        if is_gr || has_cycles {
+                        // A response that should not exist at the engine boundary: a
+                        // guaranteed-response response, or one carrying cycles. Neither is
+                        // allowed across the boundary, so this can only come from a buggy or
+                        // malicious peer. As with the disallowed requests above, drop it and
+                        // raise a critical error (any cycles are lost); the only difference is
+                        // that a response is dropped with an accept signal rather than rejected,
+                        // since there is no caller left to reject to.
+                        RequestOrResponse::Response(rep)
+                            if rep.deadline == NO_DEADLINE || rep.refund > Cycles::zero() =>
```
alin-at-dfinity, line 854: Same as above, this branch may only result in a successfully inducted illegal response if we have a bug (as there would have to be a matching `Callback`). If so, it may be more reasonable to actually induct the response, particularly a guaranteed delivery response, or the caller may end up hanging forever because of our bug.

alin-at-dfinity (approved):
LGTM, thank you.

rs/messaging/src/routing/stream_builder.rs
schneiderstefan: Renamed

rs/messaging/src/routing/stream_builder.rs
schneiderstefan: I now strip the cycles and still induct guaranteed responses.

rs/messaging/src/routing/stream_handler.rs
schneiderstefan: Ditto. I now strip and let through.

PR #10226, opened 2026-05-14
fix(state-machine-tests): include subnet type in delegation certificates

lwshang:
`get_delegation_for_subnet` was not requesting the `/subnet/<id>/type` path when building the sparse certified-state tree, so the delegation certificate never contained the subnet type leaf. This caused `verify_certified_data_with_cache_for_canister_sig` to see `subnet_type = None` and reject the delegation with "the source subnet <id> is not trusted for delegations".

Add the missing path and a regression test that verifies the type leaf is present and correctly set in the issued delegation certificate.


alin-at-dfinity (approved):
LGTM, thank you.

PR #10228, opened 2026-05-15
feat: [EXC-1783] Prepay potential `OnLowWasmMemory` hook executions

alin-at-dfinity:
When a canister implements an `OnLowWasmMemory` hook, have every message execution prepay for the potential hook execution that it may trigger. This ensures that we can always execute `OnLowWasmMemory` hooks and don't have to hold on to a Ready task for a frozen canister.


alin-at-dfinity:
Closing in favor of #10229.

While this is the Right Thing™ to do, it is a huge change to write, test and validate; and given the very limited potential benefits, it is not worth pursuing now.

PR #10230, opened 2026-05-15
fix: specify VM allocation mode for system tests via environment variable

basvandijk:
Set the `vm_allocation_mode` for system-tests via an environment variable instead of via Rust. 

This is to fix the issue that colocated performance tests would not set the required `performanceOptimizedAllocation` VM allocation mode because the group is created in the wrapper driver which would not set the `vm_allocation_mode`.

Now we specify the optional `vm_allocation_mode` via a parameter of the `system_test` macro which when specified sets the `VM_ALLOCATION_MODE` environment variable. This variable is then read and parsed when creating the group which also works for colocated tests.

### DIFF
--- a/rs/tests/system_tests.bzl
+++ b/rs/tests/system_tests.bzl
@@ -193,6 +193,16 @@ def system_test(
     }
 
     if vm_allocation_mode != None:
+        allowed_vm_allocation_modes = [
+            "minIntraDistanceLoadBalanceAllocation",
+            "performanceOptimizedAllocation",
+            "distributeAcrossDcs",
+        ]
+        if vm_allocation_mode not in allowed_vm_allocation_modes:
+            fail("Invalid vm_allocation_mode {}: must be one of {}".format(
+                repr(vm_allocation_mode),
+                allowed_vm_allocation_modes,
+            ))
         env |= {"VM_ALLOCATION_MODE": vm_allocation_mode}
 
     # RUN_SCRIPT_ICOS_IMAGES:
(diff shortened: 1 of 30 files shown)


rs/tests/system_tests.bzl
```diff
@@ -85,6 +86,12 @@ def system_test(
       logs: Specifies if vector vm for scraping logs should not be spawned.
       exclude_logs: Specifies uvm name patterns to exclude from streaming.
       data: List of files used by the test driver.
+      vm_allocation_mode: Optional VM allocation mode string forwarded to the
+        test driver via the `VM_ALLOCATION_MODE` environment variable. Must
+        match one of the serde rename strings of `VmAllocationMode`, e.g.
+        `"performanceOptimizedAllocation"`,
+        `"minIntraDistanceLoadBalanceAllocation"` or `"distributeAcrossDcs"`.
+        When None it defaults to `"minIntraDistanceLoadBalanceAllocation"`.
```
nmattia, line 94: nit: could actually check this in the code

nmattia (approved):

rs/tests/system_tests.bzl
basvandijk: In the Bazel code you mean? Because it is checked in Rust.

mraszyk (approved):

rs/tests/system_tests.bzl
basvandijk: Done

PR #10231, opened 2026-05-15
test: check for critical errors before replica restart in system tests

mraszyk:
This PR checks for critical errors before replica restart (due to guestOS deployment) in system tests. The motivation for this change is that critical errors after subnet upgrade are only checked in `//rs/tests/consensus/backup:backup_manager_upgrade_test`, but not in `//rs/tests/consensus/upgrade:upgrade_downgrade_nns_subnet_test` which downgrades the subnet after upgrade and thus the critical errors after upgrade are lost following the subsequent replica restart during downgrade.

### DIFF
--- a/rs/tests/driver/src/driver/group.rs
+++ b/rs/tests/driver/src/driver/group.rs
@@ -11,7 +11,7 @@ use crate::driver::{
     task_scheduler::{TaskScheduler, TaskTable},
     test_env_api::{
         FarmBaseUrl, HasFarmUrl, HasGroupSetup, HasTopologySnapshot, IcNodeContainer,
-        ORCHESTRATOR_METRICS_PORT, REPLICA_METRICS_PORT,
+        ORCHESTRATOR_METRICS_PORT, REPLICA_METRICS_PORT, TopologySnapshot,
     },
 };
 use crate::driver::{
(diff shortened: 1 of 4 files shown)


rs/tests/driver/src/driver/group.rs
```diff
@@ -647,6 +647,48 @@ impl SystemTestSubGroup {
...
+
+/// Checks replica and orchestrator error metrics on all subnet nodes, using the
+/// same thresholds as the default `SystemTestGroup` teardown. Call this before
+/// replica restart so that errors accumulated in the current process are not lost.
+pub fn assert_no_critical_errors(env: &TestEnv) {
+    check_metrics_for_nodes(
+        &env.topology_snapshot(),
+        &default_replica_metrics(),
+        &default_orchestrator_metrics(),
+    );
```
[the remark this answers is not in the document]
mraszyk: All tests pass now: let's generalize later if needed.

rs/tests/driver/src/driver/test_env_api.rs
```diff
@@ -869,6 +869,10 @@ pub struct IcNodeSnapshot {
 }
 
 impl IcNodeSnapshot {
+    pub fn test_env(&self) -> &TestEnv {
+        &self.env
+    }
+
```
basvandijk:
```suggestion
```
This is unnecessary since [`IcNodeSnapshot` already implements `HasTestEnv`](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/tests/driver/src/driver/test_env_api.rs?L1584-L1588).

basvandijk (approved):

rs/tests/driver/src/driver/test_env_api.rs
mraszyk: Thanks! Claude (and myself) weren't aware...

eichhorl (approved):

PR #10232, opened 2026-05-15
test(registry): Update `create_subnet` test to set registry's default initial DKG subnet

eichhorl:
This PR updates the `create_subnet` to exercise all the different ways the default initial DKG subnet (which is the subnet responsible for the subnet creation) may be defined:
1. No initial DKG subnet specified in either the registry or the proposal: The creation should be handled by the NNS.
2. A default initialDKG subnet is set in the registry: The creation should be handled by that subnet.
3. An initial DKG subnet is specified in both the registry and the proposal: The creation should be handled by the subnet in the proposal.


daniel-wong-dfinity-org-twin:
added scope to title

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+    // SetupInitialDKG calls.
+    for subnet in topology_snapshot_after_proposals.subnets() {
+        if !original_subnet_ids.contains(&subnet.subnet_id) {
+            continue;
+        }
+        block_on(assert_remote_transcripts_delivered_per_tag(
+            log,
+            &subnet,
+            EXPECTED_TRANSCRIPTS_PER_TAG,
+        ));
```
daniel-wong-dfinity-org-twin: It would be better if we looked at the counters after each proposal (and at the very beginning), not just once at the end. Then, we would be able to attribute each increment to one proposal.

(Alternatively, we could look at the counters after each batch, and at the beginning.)

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+
+/// Submits `count` create-application-subnet proposals (each consuming
+/// `NODES_PER_SUBNET` unassigned nodes) and waits for all of them to execute.
+async fn submit_and_wait_for_create_subnet_proposals(
+    log: &Logger,
+    governance: &Canister<'_>,
+    unassigned_nodes: &mut impl Iterator<Item = NodeId>,
+    version: ReplicaVersion,
+    initial_dkg_subnet_id: Option<SubnetId>,
+    count: usize,
```
daniel-wong-dfinity-org-twin, line 340: Do as you see fit.

Not sure why do > 1 proposal. Seems like extra complexity for minimal gain. I mean, I don't think this is "actually" harmful (hence DAYSF).

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+) {
+    let mut proposal_ids = Vec::with_capacity(count);
```
daniel-wong-dfinity-org-twin: Do as you see fit.

with_capacity seems like pre-mature optimization.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+            "Waiting until subnet {subnet_id} reports {expected_per_tag} delivered remote DKG \
+             transcripts per tag on all {expected_node_count} nodes"
+        ),
+        log,
+        METRIC_TIMEOUT,
+        METRIC_BACKOFF,
+        || async {
+            let values = match metrics.fetch::<u64>().await {
+                Ok(values) => values,
+                Err(err) => bail!("Failed to fetch metrics from subnet {subnet_id}: {err}"),
```
daniel-wong-dfinity-org-twin: Do as you see fit.

> bail

Macros should be avoided, unless they do a significant amount of lifting, because they are magical.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -28,27 +43,40 @@ use ic_system_test_driver::driver::ic::{InternetComputer, Subnet};
...
 use std::time::Duration;
 
-const NNS_PRE_MASTER: usize = 4;
-const APP_PRE_MASTER: usize = 4;
-const DKG_INTERVAL_LENGTH: u64 = 29;
-const APP_SUBNETS: usize = 6;
+const METRIC_TIMEOUT: Duration = Duration::from_secs(120);
+const METRIC_BACKOFF: Duration = Duration::from_secs(5);
+const MR_REGISTRY_VERSION_METRIC: &str = "mr_registry_version";
+const REMOTE_TRANSCRIPTS_METRIC: &str = "consensus_dkg_remote_transcripts_delivered_total";
```
daniel-wong-dfinity-org-twin: Do as you see fit.

The names of the metrics themselves are already meaningful. Therefore, these constants seem to just add indirection.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+        || async {
+            let values = match metrics.fetch::<u64>().await {
+                Ok(values) => values,
+                Err(err) => bail!("Failed to fetch metrics from subnet {subnet_id}: {err}"),
+            };
+            for key in &metric_keys {
+                let Some(per_node) = values.get(key) else {
+                    bail!("Metric {key} not yet exposed on subnet {subnet_id}");
+                };
+                if per_node.len() != expected_node_count {
```
daniel-wong-dfinity-org-twin: > per_node

How about node_metric_values? In general, variable names should be "noun-y"; it should answer "What is this thing?". If it's a collection, that reduces to "What are the elements?", not "What is each element associated with?".

rs/tests/nns/create_subnet_test.rs
```diff
@@ -228,6 +294,164 @@ pub fn test(env: TestEnv) {
...
+            };
+            if per_node.len() != expected_node_count {
+                bail!(
+                    "Subnet {subnet_id}: metric {MR_REGISTRY_VERSION_METRIC} reported by {} out \
+                     of {} nodes",
+                    per_node.len(),
+                    expected_node_count,
+                );
+            }
+            if let Some(behind) = per_node.iter().find(|v| **v < target_version) {
```
daniel-wong-dfinity-org-twin: > behind

suggestion: subnet_version. This is noun-y; it answers "What even IS this thing?" (a version).

daniel-wong-dfinity-org-twin (commented):
Seems like the test has been completely revamped. Another way would be to leave the old simple test, and add a new one. OTOH, these tests are expensive, I guess. If these tests were not slow, I think it would be better to have 1 simple basic test, and a separate more "challenging" one that exercises a specific aspect (e.g. key generation) of whatever operation (in this case, subnet creation) is being tested. Do as you see fit.

rs/tests/nns/create_subnet_test.rs
eichhorl: Good idea, done: https://github.com/dfinity/ic/pull/10232/commits/e2714d9e20addacd23f0482edca2190ff2f2ba6e

rs/tests/nns/create_subnet_test.rs
eichhorl: This is to keep the original behavior of testing multiple subnet creations at the same time

rs/tests/nns/create_subnet_test.rs
eichhorl: No longer applies

rs/tests/nns/create_subnet_test.rs
eichhorl: Using `bail!` seems to be the convention in these `retry_with_msg_async!` loops. Also, inlining it would simply turn it into the `anyhow!` macro, which seems hard to avoid.

rs/tests/nns/create_subnet_test.rs
eichhorl: Removed: https://github.com/dfinity/ic/pull/10232/commits/23f18c5138edd6f92839927e78fc1a03aef2dfa8

rs/tests/nns/create_subnet_test.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10232/commits/ece3950e1fa94085e31dc47ff54758b0631d366c

rs/tests/nns/create_subnet_test.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10232/commits/ece3950e1fa94085e31dc47ff54758b0631d366c

eichhorl:
> Seems like the test has been completely revamped. Another way would be to leave the old simple test, and add a new one. OTOH, these tests are expensive, I guess. If these tests were not slow, I think it would be better to have 1 simple basic test, and a separate more "challenging" one that exercises a specific aspect (e.g. key generation) of whatever operation (in this case, subnet creation) is being tested. Do as you see fit.

That makes sense, however another argument is that we also want to test the interaction between the different cases, i.e. while a default subnet is set, it is still possible to choose a different subnet specifically as part of the create subnet proposal.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -112,19 +109,34 @@ pub fn test(env: TestEnv) {
...
-    ];
+        .filter(|s| s.subnet_id != nns_subnet_id);
+    let default_subnet = other_original_subnets
+        .next()
+        .expect("missing second original subnet for setup_initial_dkg");
+    let third_subnet = other_original_subnets
+        .next()
+        .expect("missing third original subnet for setup_initial_dkg");
+    assert!(
+        other_original_subnets.next().is_none(),
```
daniel-wong-dfinity-org-twin, line 122: This can be expressed as `assert_eq!(observed, None, ...)`. We should use `assert_eq` in favor of `assert` whenever possible, because it allows us to see the observed value "for free" when the assert fails.

rs/tests/nns/create_subnet_test.rs
daniel-wong-dfinity-org-twin: > original behavior

Ah. I didn't notice that. Another reason to maybe do this test separately; make it a little less complicated.

But still, do as you see fit.

rs/tests/nns/create_subnet_test.rs
daniel-wong-dfinity-org-twin: bail is strictly more magical, because a) it uses anyhow!, and b) it actually changes the flow (by generating a return statement), not just generates an expression.

If anyhow was the standard for how we do Errors, it would be different, because "optimize the common case". But it's not the common case -> bail seems like premature optimization to me.

But anyway, the verdict here is not very clear -> do as you see fit (still).

rs/tests/nns/create_subnet_test.rs
daniel-wong-dfinity-org-twin: What about mr_registry_version?

rs/tests/nns/create_subnet_test.rs
```diff
@@ -232,12 +257,12 @@ pub fn test(env: TestEnv) {
...
-        let original_subnet_set = set(&original_subnets);
+        let original_subnet_set = set(&original_subnet_list);
         let final_subnet_set = set(&final_subnets);
-        // check that there are exactly APP_SUBNETS added subnets
+        // check that there are exactly NEW_SUBNETS added subnets
         assert_eq!(
             original_subnet_set.len() + NEW_SUBNETS,
-            final_subnet_set.len(),
+            final_subnet_set.len()
         );
```
daniel-wong-dfinity-org-twin, line 269: Consider adding bread crumbs. What if the assert fails? We want to know not just how many subnets were observed, but maybe something about them, to help diagnose.

rs/tests/nns/create_subnet_test.rs
```diff
@@ -232,12 +257,12 @@ pub fn test(env: TestEnv) {
...
 
-        let original_subnet_set = set(&original_subnets);
+        let original_subnet_set = set(&original_subnet_list);
         let final_subnet_set = set(&final_subnets);
-        // check that there are exactly APP_SUBNETS added subnets
+        // check that there are exactly NEW_SUBNETS added subnets
         assert_eq!(
             original_subnet_set.len() + NEW_SUBNETS,
-            final_subnet_set.len(),
+            final_subnet_set.len()
```
daniel-wong-dfinity-org-twin: Swap the order of the operands, because we do assert_eq!(observed, expected), and in this case, the reference value is original + new, and the observed value is final.

rs/tests/nns/create_subnet_test.rs
eichhorl: `bail!` is the standard in system tests that use `retry_with_msg`, so I prefer to keep it here as well

rs/tests/nns/create_subnet_test.rs
eichhorl: Unfortunately `SubnetSnapshot` implements neither `PartialEq` nor `Debug`, so `assert_eq!(…, None)` doesn't work

rs/tests/nns/create_subnet_test.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10232/commits/3570c6a1aef71372c78f1cbcc418bdc696f14e4d

rs/tests/nns/create_subnet_test.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10232/commits/3570c6a1aef71372c78f1cbcc418bdc696f14e4d

rs/tests/nns/create_subnet_test.rs
daniel-wong-dfinity-org-twin: Not making any suggestions here. Leaving as-is is fine. I'm just marking this as unresolved to avoid the following opinion getting lost.

I don't think that's a strong reason, because you can always define a scope where some pattern is common. In any case, that defeats the purpose: the reason "standard for how we do Errors" matters is that it means that a master programmer who learned Rust and joined DFINITY recently still has a good chance at understanding what's going on. But if you define the scope to be something narrow and specialized, then, that doesn't work anymore. That only works if anyhow is pervasive at DFINITY.

rs/tests/nns/create_subnet_test.rs
daniel-wong-dfinity-org-twin: Do as you see fit.

Why not derive those?

Admittedly, this is beyond the scope of this PR. OTOH, this suggestion is small, simple, and maybe, it should have been done this way originally.

daniel-wong-dfinity-org-twin (approved):

rs/tests/nns/create_subnet_test.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10232/commits/42293371cb7adb74de4f5994a8dd5a64256fb39c

rs/tests/nns/create_subnet_test.rs
eichhorl: > Why not derive those

I don't think it is worth the effort of transitively deriving them on all structs contained in `SubnetSnapshot` just to use `assert_eq!` here. In case of a failure, we wouldn't really care how the additional subnet looks like anyway. Its existance is signal enough and the problem should be very easy to debug given that the "original subnets" are a direct result of the test setup.

rs/tests/nns/create_subnet_test.rs
eichhorl: Resolving this so that the PR can be merged. `anyhow` is extremely common at least in system tests, I personally don't find `bail!` more or less "magical" than the `?` operator for instance.

PR #10234, opened 2026-05-15
fix: Also consider oldest registry version in use by replicated state if no IDKG payload exists

eichhorl:
## Background
Subnet membership changes are enforced on DKG interval boundaries, whenever a new CUP is created. However, even if a node is no longer part of the consensus committee at that point, its participation might still be required as part of other protocols (DKG, threshold signatures, ...).

For that reason, before killing the unassigned replica and closing connections to its peers, the orchestrator uses the latest CUP to calculate the "oldest registry version in use". If the replica is still part of the subnet according to this oldest version, the orchestrator will keep it running until all ongoing protocols are completed.

The oldest registry version is determined based on data in the summary block, and the state at the summary height (which is a checkpoint height). For instance, ECDSA signature request contexts contain a registry version which determines the subnet membership required to fulfil the request. Therefore, this registry version influences the "oldest registry version in use".

## Proposed Changes
Until now, tECDSA and tSchnorr contexts were the only structures in the state influencing the oldest registry version. For that reason, the "oldest registry version in use by replicated state" was only calculated for chain-key subnets running the IDKG protocol.

However, we should also consider the registry version of `SetupInitialDKG` contexts, which similarly determines the membership required to fulfil the request. In the future, the same should be done for HTTP outcalls. These call contexts may exist on ALL subnets. The function `get_oldest_state_registry_version` already considers `SetupInitialDKG` contexts when calculating the registry version, but it is only called if the summary has an IDKG payload.

With this PR, we will also start to consider the oldest state registry version on subnets that do not maintain an IDKG payload.

Note that in order to determine the oldest state registry version, we need to access the state at the checkpoint height. By default, this state is not guaranteed to exist in memory, since a call to `remove_inmemory_states_below` may have purged it already. For that reason, the purger excludes states at summary heights that do not have a CUP yet from this call. Until now, this was however also only done for summaries containing IDKG payloads. 

With this PR, we will protect states at summary heights which do not have a CUP yet from being purged too early, regardless of whether or not it is a chain-key subnet.

### DIFF
--- a/rs/consensus/src/consensus/catchup_package_maker.rs
+++ b/rs/consensus/src/consensus/catchup_package_maker.rs
@@ -208,26 +208,21 @@ impl CatchUpPackageMaker {
                 panic!("Height {height} is not a fully certified height. This should not happen.",);
             }
             Ok(state_hash) => {
-                let summary = start_block.payload.as_ref().as_summary();
-                let registry_version = if summary.idkg.is_some() {
-                    // Should succeed as we already got the hash above
-                    let state = self
-                        .state_manager
-                        .get_state_at(height)
-                        .map_err(|err| {
-                            error!(
-                                self.log,
-                                "Cannot make IDKG CUP at height {}: `get_state_hash_at` \
-                                succeeded but `get_state_at` failed with {}. Will retry",
-                                height,
-                                err,
-                            )
-                        })
-                        .ok()?;
-                    get_oldest_state_registry_version(state.get_ref())
-                } else {
-                    None
-                };
+                // Should succeed as we already got the hash above
+                let state = self
+                    .state_manager
+                    .get_state_at(height)
+                    .map_err(|err| {
+                        error!(
+                            self.log,
+                            "Cannot make CUP at height {}: `get_state_hash_at` \
+                            succeeded but `get_state_at` failed with {}. Will retry",
+                            height,
+                            err,
+                        )
+                    })
+                    .ok()?;
+                let registry_version = get_oldest_state_registry_version(state.get_ref());
                 let content = CatchUpContent::new(
                     HashedBlock::new(ic_types::crypto::crypto_hash, start_block),
                     HashedRandomBeacon::new(ic_types::crypto::crypto_hash, random_beacon),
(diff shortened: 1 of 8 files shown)


rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -208,26 +208,21 @@ impl CatchUpPackageMaker {
...
-                                height,
-                                err,
-                            )
-                        })
-                        .ok()?;
-                    get_oldest_state_registry_version(state.get_ref())
-                } else {
-                    None
-                };
+                // Should succeed as we already got the hash above
```
michael-weigelt, line 211: Why could it not be removed in the meantime?
In the PR description, you mention that "With this PR, we will protect states at summary heights which do not have a CUP yet from being purged too early, ", which presumably would solve that problem. But I don't see any code for that in this PR, should that not be something done in the state manager?

eichhorl: Sorry, that note corresponds to the changes to `get_pending_cup_heights` in `purger.rs`. The connection that is missing in the PR description is that we pass the heights returned by `get_pending_cup_heights` into the `extra_heights` that are protected from `remove_inmemory_states_below`.

Previously we only protected heights of summaries that don't have a CUP yet, if those summaries also contained an IDKG payload. Now we protect heights of _all_ summaries that don't have a CUP yet.

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity, line 514: Would it make sense to also add tests for a state that has:
- No signature requests but a `SetupInitialDkg` request
- Both signature requests and a `SetupInitialDkg` request

Both with & without IDKG payload, even though I feel like a state with signature requests but not having an IDKG payload is not really realistic, i.e. it should not happen in a benign execution of the program.

Same in the validator.

eichhorl: https://github.com/dfinity/ic/pull/10234/commits/e2952ecaf2120963fb92d6bc6c72dda0a5d625c6

pierugo-dfinity: Wonderful thanks!

pierugo-dfinity (approved):

PR #10241, opened 2026-05-18
fix: canister log migration

mraszyk:
This PR fixes canister log (data) migration:
- a new Boolean value `migrated` in `canister.pbuf` (inside `log_memory_store` in `SystemState`) tells if logs have been migrated from the (legacy) field `canister_log` in `SystemState` to the new field `log_memory_store` in `SystemState`:
  - default to false for existing canisters (transparently by protobuf since the field is currently missing);
  - set to the feature flag for new canisters (so that new canisters are not labeled as not migrated once the feature is enabled);
- `fetch_canister_logs` uses the new `log_memory_store` if the feature flag is true and the store is migrated;
- _data migration_: at the very beginning of `SchedulerImpl::execute_round`, for every canister
  - if the feature flag is true and the canister’s log memory store is not migrated => migrate (resize store to 4KiB – which is the maximum size of the legacy `canister_log` in `SystemState` –, migrate records, set `migrated` to true);
  - if the feature flag is false and the canister’s log memory store is migrated (i.e., after a downgrade) => “undo” migration (clear the log memory store by resizing down to 0 and reset `migrated` to false).

The `log_memory_store_feature` flag is used:
- to trigger data migration and rollback in `SchedulerImpl::execute_round`;
- to initialize the `migrated` field for newly created canisters;
- to select between the legacy `canister_log` and new `log_memory_store` when reading logs via `fetch_canister_logs` (this way, flipping the flag immediately falls back to the legacy `canister_log`).

The `migrated` field is used:
- to ensure that data migration is one-off;
- to determine the byte capacity for canister logs during execution;
- to determine if new canister logs should be appended to the new `log_memory_store`;
- to determine if canister logs should be read from the new `log_memory_store`;
- to determine if the new `log_memory_store` can be resized via `update_settings`;
- to determine if metrics should be recorded for the new `log_memory_store`.


rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -402,15 +402,9 @@ impl SystemStateModifications {
...
         }
         if LOG_MEMORY_STORE_FEATURE_ENABLED {
-            let log_memory_store = &mut system_state.log_memory_store;
-            // TODO(DSM-11): cleanup population logic after migration is done.
-            // We need to copy existing canister_log to log_memory_store in order
-            // not to loose any log records until the migration is complete.
-            let old_total = &system_state.canister_log;
-            if log_memory_store.is_empty() && !old_total.is_empty() {
-                log_memory_store.append_delta_log(&mut old_total.clone());
-            }
```
mraszyk, line 412: This migration will only trigger if users resize their log memory store explicitly (via `update_settings`) and then their canister executes a message => we drop and migrate every canister forcibly in `SchedulerImpl::execute_round` (so that we ensure we could drop the legacy code soon without relying on users to migrate on their own).
This migration code is also unsound because the log memory store could be empty after the user resized it down to 0 and then upsized back to some positive size.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -45,9 +45,6 @@ use std::time::Duration;
 /// the maximum message response size.
 #[derive(Debug, ValidateEq)]
 pub struct LogMemoryStore {
-    /// Feature flag for controlling LogMemoryStore enabled.
-    feature_flag: FlagStatus,
```
mraszyk, line 49: Main motivation for removing this: checkpoint loading does not have access to any config and thus the feature flag must be hard-coded which makes testing tricky => do not store the feature flag here, but rather gate calls to the new `log_memory_store` using the config which is available at the call sites.

rs/state_manager/src/tip.rs
```diff
@@ -1273,11 +1272,8 @@ fn serialize_canister_protos_to_checkpoint_readwrite(
             snapshot_visibility: canister_state.system_state.snapshot_visibility.clone(),
             log_memory_limit: canister_state.log_memory_limit(),
             canister_log: canister_state.system_state.canister_log.clone(),
-            next_canister_log_record_idx: if LOG_MEMORY_STORE_FEATURE_ENABLED {
-                canister_state.system_state.log_memory_store.next_idx()
-            } else {
-                canister_state.system_state.canister_log.next_idx()
-            },
+            next_canister_log_record_idx: canister_state.system_state.canister_log.next_idx(),
```
mraszyk, line 1275: - if the feature flag is true and the `migrated` value is true: `next_idx` is the same for both the legacy `canister_log` and the new `log_memory_store` (because every log is appended to both);
- otherwise: then the legacy `canister_log` is the single source of truth anyway.

basvandijk (approved):
rs/tests changes LGTM.

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -394,12 +407,14 @@ impl LogMemoryStore {
...
     pub fn append_delta_log(&mut self, delta_log: &mut CanisterLog) {
-        if self.feature_flag == FlagStatus::Disabled {
-            self.deallocate();
-            return;
-        }
         if delta_log.is_empty() {
-            return; // Don't append if delta is empty.
+            // No records to append, but still carry the monotone index forward.
+            // The canister_log can be empty after uninstalling a canister (uninstall
+            // clears the log records but preserves next_idx). Wrap-around alone will
```
michael-weigelt: Why not reset it? 

I don't understand how the last sentence here explains why to always increase the index.

michael-weigelt (approved):

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
mraszyk: > Why not reset it?

if somebody keeps polling the logs by monotonically increasing a counter, then they should have to reset their requested counter after the canister is uninstalled (which they might not even realize to have happened)

> I don't understand how the last sentence here explains why to always increase the index.

The sentence "Wrap-around alone will never evict all entries because truncation ensures every record fits within the capacity." refers to "uninstall clears the log records but preserves next_idx" and argues why the example scenario talks about uninstalling a canister.

mraszyk: Attempted to clarify here: https://github.com/dfinity/ic/pull/10241/commits/d38616656abde004f0e7a90660695bbe2f0bd8e1

rs/execution_environment/src/canister_logs.rs
```diff
@@ -29,7 +29,10 @@ pub(crate) fn fetch_canister_logs(
     let s = &canister.system_state;
     let canister_log_records = match log_memory_store_feature {
         FlagStatus::Disabled => filter_records(&args, s.canister_log.records())?,
-        FlagStatus::Enabled => s.log_memory_store.records(args.filter),
+        FlagStatus::Enabled if s.log_memory_store.is_migrated() => {
+            s.log_memory_store.records(args.filter)
+        }
+        FlagStatus::Enabled => filter_records(&args, s.canister_log.records())?,
```
schneiderstefan: Is that intentional that you have 2 branches that are exactly the same?

rs/execution_environment/tests/canister_logging.rs
```diff
@@ -2659,3 +2659,354 @@ fn test_log_memory_store_feature_flag_via_execution_test_builder() {
         0,
     );
 }
+
+#[test]
+fn test_log_migration_from_canister_log_to_log_memory_store() {
```
schneiderstefan: Do we have a test for the other direction?

rs/execution_environment/src/canister_logs.rs
mraszyk: The first one has a guard `if s.log_memory_store.is_migrated()`

rs/execution_environment/tests/canister_logging.rs
mraszyk: The name is misleading: the test exercises both upgrade and downgrade.

rs/execution_environment/src/canister_logs.rs
schneiderstefan: I am talking about the Enabled and the Disabled case, both of which have the same `filter_records` code.

rs/execution_environment/tests/canister_logging.rs
mraszyk: https://github.com/dfinity/ic/pull/10241/commits/bbc725ed03f73e2f92f8d963f36774c552ccbdc6

rs/execution_environment/src/canister_logs.rs
mraszyk: Now I see what you mean: I try to avoid `_`, but here it's fine since `FlagStatus` is not really expected to evolve before we drop this entirely in a few weeks after rolling out the feature.
https://github.com/dfinity/ic/pull/10241/commits/b2668e13e826301946c6fa36ab0e385150263b27

schneiderstefan (approved):

rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
michael-weigelt: Thanks!

PR #10242, opened 2026-05-18
feat: CON-1718 Introduce Proposal to set Default Initial DKG Subnet

eichhorl:
## Background
https://github.com/dfinity/ic/pull/9782 introduced the ability to route `SetupInitialDKG` requests (caused by subnet creation/recovery/splitting) to arbitrary subnets, instead of always handling them on NNS.

This was done by extending proposal payloads to create/recover/split/rent a subnet with an optional `SubnetId` field, to which the request should be routed.

## Proposed Changes
The approach above delegates the responsibility of choosing a suitable subnet to the proposer, which is brittle and difficult to test.

Instead, this PR introduces a registry key holding a "default initial DKG subnet" to which all `SetupInitialDKG` requests should be routed, unless specified otherwise by the proposal payload. 

In particular, we:
- Introduce a new proposal to set/unset this new registry key
- Add `ic-admin` support to submit the proposal and query the current default subnet
- Add a registry invariant that the default initial DKG subnet must exist
- Populate the default initial DKG subnet in the replicated state's `NetworkTopology` based on the value in registry

Note that it must still be possible to override the default initial DKG subnet (at least as part of the recovery proposal), for the case that the default DKG subnet is the one being recovered.

## Future Work
Extend the `create_subnet` system test to include a default initial DKG subnet, once the functionality is supported by mainnet NNS canisters.


michael-weigelt (approved):
Core protocol lgtm

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -173,6 +178,41 @@ pub(crate) fn check_subnet_invariants(
...
+    check_default_initial_dkg_subnet_invariant(snapshot)?;
+
+    Ok(())
+}
+
+/// Default initial DKG subnet invariant holds iff:
+///   * The `default_initial_dkg_subnet_id` record does not exist, OR
+///   * The record exists, decodes to a valid `SubnetId`, and that subnet
+///     appears in the subnet list.
+fn check_default_initial_dkg_subnet_invariant(
```
pierugo-dfinity, line 190: General question: do we also want to check that the subnet is a System subnet?

pierugo-dfinity (approved):

rs/registry/canister/src/invariants/subnet.rs
eichhorl: In principal any subnet can perform the remote DKG, so I didn't see a reason why it should be restricted

pierugo-dfinity: Right, just a matter of increased trust maybe

jasonz-dfinity (approved):

PR #10247, opened 2026-05-18
feat: [EXC-1783] No execution overhead on low cycles, take 2

alin-at-dfinity:
[This re-applies commit dc3040c. It was reverted in commit 7602e8d because it introduced a potential infinite loop. That was fixed by 33a65ef, so this is safe to merge.]

When a message execution is skipped due to low cycle balance, don't deduct the message execution overhead from the round instruction limit. The instruction overhead is intended to account for entering and exiting the sandbox, which does not apply here.

This will allow the "execution" of arbitrarily many messages per round on frozen (or close to frozen) canisters, which could drive up round duration. However, there are a lot of other limits in place (ingress messages in blocks, canister call fees, queue sizes) that make this a non-issue. Particularly given that the overhead for skipping a message execution is at least an order of magnitude lower than that of executing one.

### DIFF
--- a/rs/execution_environment/src/scheduler.rs
+++ b/rs/execution_environment/src/scheduler.rs
@@ -222,11 +222,11 @@ impl SchedulerImpl {
                 as_num_instructions(instructions_before - round_limits.instructions);
 
             let messages = match execute_subnet_message_result_type {
-                ExecuteSubnetMessageResultType::Finished => NumMessages::from(1),
-                ExecuteSubnetMessageResultType::Processing => NumMessages::from(0),
-                ExecuteSubnetMessageResultType::Paused => NumMessages::from(0),
+                ExecuteSubnetMessageResultType::Finished => NumMessages::new(1),
+                ExecuteSubnetMessageResultType::Processing => NumMessages::new(0),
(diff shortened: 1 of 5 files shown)


rs/execution_environment/src/scheduler.rs
```diff
@@ -1678,16 +1758,21 @@ fn execute_canisters_on_thread(
...
+                    slices = NumSlices::new(1);
                     executed_canisters.insert(new_canister.canister_id());
                 }
             }
-            measurement_scope.add(round_instructions_executed, NumSlices::from(1), messages);
+            measurement_scope.add(round_instructions_executed, slices, messages);
             total_messages_executed += messages;
+            total_slices_executed += slices;
+            if slices.get() > 0 {
+                // A slice was actually executed, count the execution overhead toward the round limit.
```
michael-weigelt, line 1769: Is every message also considered a slice? Or should messages be checked too?

alin-at-dfinity: A slice in this context is either a slice of a DTS execution or a message that completed execution within a round. I.e. some actual Wasm execution.

michael-weigelt (approved):

PR #10250, opened 2026-05-19
chore: CON-1453 Introduce a type for remote transcript results

eichhorl:
Instead of the verbose `(NiDkgId, CallbackId, Result<NiDkgTranscript, String>)` triple, this PR introduces a dedicated struct for remote NiDKG transcript creation results.

Additionally, we remove all mentions of "early" remote transcripts. After https://github.com/dfinity/ic/pull/10055, "early" transcript creation (i.e. as part of data blocks) is the only way for remote transcripts to be created, since they can no longer be part of summary blocks. Therefore, we can drop the "early" prefix.

Note that, although we change the type of the summary payload, this change is backward/forward compatible (see passing `cup_compatibility_test`). This is because the hash function of the triple is equal to the one of the struct, therefore no migration is necessary.

### DIFF
--- a/rs/consensus/dkg/src/payload_validator.rs
+++ b/rs/consensus/dkg/src/payload_validator.rs
@@ -223,9 +223,9 @@ fn validate_dealings_payload(
         crypto_validate_dealing(crypto, config, message)?;
     }
 
-    // If we have early transcripts, we compare them
+    // If we have remote transcripts, we compare them
     if !dealings.transcripts_for_remote_subnets.is_empty() {
-        let expected_transcripts = payload_builder::create_early_remote_transcripts(
+        let expected_transcripts = payload_builder::create_remote_transcripts(
             pool_reader,
             crypto,
             parent,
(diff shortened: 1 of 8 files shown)


rs/consensus/dkg/src/payload_validator.rs
```diff
@@ -223,9 +223,9 @@ fn validate_dealings_payload(
         crypto_validate_dealing(crypto, config, message)?;
     }
 
-    // If we have early transcripts, we compare them
+    // If we have remote transcripts, we compare them
     if !dealings.transcripts_for_remote_subnets.is_empty() {
```
pierugo-dfinity, line 227:
```suggestion
```

Shall we still keep this `if` now?
We now expect remote transcripts at any data block and I'd assume we'd want to invalidate a blockmaker that has maliciously removed them.

This can also be done in a separate PR since it is out of scope.

pierugo-dfinity (approved):
Nice

rs/consensus/dkg/src/payload_validator.rs
eichhorl: I'll take a look in a follow up 👍

PR #10251, opened 2026-05-19
chore: Introduce a type for remote DKG attempts

eichhorl:
As suggested by @pierugo-dfinity in https://github.com/dfinity/ic/pull/9926#discussion_r3117445553, we introduce a new enum type keeping track of the attempt status of ongoing remote DKG requests.

Previously, this was done by a map holding `u32`, where `0` indicated a completed request, and `n > 0` a request that has been attempted `n` times. This is dangerous, since accidentally incrementing a "completed" request with a count of `0` would lead to the same request being retried again with an attempt count of `1`.

The new enum instead disallows `RemoteDkgAttempts::Attempt(0)` by construction, and requires matching on the enum to access the attempt count.

The new enum serializes to `u32`, as before, and the hash function stays the same. Therefore this change is backward/forward compatible.

Additionally, we rename the map from `initial_dkg_attempts` to `remote_dkg_attempts`, since it also tracks `ReshareChainKey` requests (not just `SetupInitalDkg`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5643,6 +5643,7 @@ dependencies = [
  "ic-registry-proto-data-provider",
  "ic-test-utilities-registry",
  "ic-types",
+ "rand 0.8.5",
  "rustls 0.23.37",
  "sev_guest",
  "sev_guest_testing",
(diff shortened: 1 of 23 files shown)


pierugo-dfinity (approved):
Nice!

PR #10256, opened 2026-05-19
chore: remove some unused deps from the Bazel build

nmattia:
These dependencies are not used (they are also not present in the Cargo build).


basvandijk (approved):
Spring cleaning!

PR #10262, opened 2026-05-20
feat: Move a few more tests away from the blessed versions list

Bownairo:

This moves a few more tests away from depending on the blessed version list, so that we can eventually remove the record entirely.


basvandijk (approved):

michael-weigelt (approved):

frankdavid (approved):
👍🏻 for node

NikolaMilosa (approved):

PR #10264, opened 2026-05-20
fix: IC-1967 Drop HTTP responses if context is no longer part of the state

eichhorl:
The HTTP outcalls client dispatches HTTP request contexts from the replicated state to the HTTP adapter. The adapter does the actual request and returns the response asynchronously at a later point.

In the meantime, the request context may have already disappeared from replicated state, i.e. because the request timed out or it already received enough responses from other replicas.

Before this PR, the response returned from the adapter would be treated as fully replicated by default. This causes the replica to sign and gossip a share despite the missing context.

With this PR, the replica instead drops responses returned by the adapter, for which no context exists anymore.

### DIFF
--- a/rs/https_outcalls/consensus/src/pool_manager.rs
+++ b/rs/https_outcalls/consensus/src/pool_manager.rs
@@ -363,6 +363,19 @@ impl CanisterHttpPoolManagerImpl {
                             .ellipsize(MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES, 90);
                     }
 
+                    // Drop the response if its context is no longer present in the replicated state
+                    // (e.g. the request has timed out or has already been answered by enough other nodes).
+                    let Some(context) = active_contexts.get(&response.id) else {
+                        warn!(
+                            self.log,
+                            "Dropping http response for request ID {}: \
+                             corresponding context is no longer in the replicated state.",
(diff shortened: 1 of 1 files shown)


rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -363,6 +363,19 @@ impl CanisterHttpPoolManagerImpl {
                             .ellipsize(MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES, 90);
                     }
 
+                    // Drop the response if its context is no longer present in the replicated state
+                    // (e.g. the request has timed out or has already been answered by enough other nodes).
+                    let Some(context) = active_contexts.get(&response.id) else {
```
pierugo-dfinity: nit: this could be moved before the previous `if` checking the error message's length to avoid potentially unnecessary work.

pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10264/commits/b17fd5a615cd128f86ad748f95fb50c0211e8a9a

PR #10272, opened 2026-05-21
chore: Remove ExecutionState::canister_root

schneiderstefan:
The canister_root field used to point to the files backing the canister states on disk. It has not been used or populated in a while, with successive state manager improvements (tip thread, LSMT, asynchronous checkpointing) decoupling the replicated state and the execution environment from the file system. The field was also set to "NOT_USED" to prevent any accidental misuse of the field by the execution layer.

This PR removes the field and simplifies the code where possible.


alin-at-dfinity (approved):
Wow, it was even on the replica dashboard. (o:

PR #10276, opened 2026-05-21
perf: Speed up creating_canisters_always_works test

alin-at-dfinity:
Looks like commit 41fdff3 has caused test `creating_canisters_always_works_if_limit_is_set_to_zero` to slow down to 60+ seconds (potentially because of the overhead of encoding and decoding the `CreateCanisterArgs`.

Batch the `CreateCanister` calls to speed the test up to under 1 second.


schneiderstefan (approved):

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -3195,7 +3195,9 @@ fn creating_canisters_always_works_if_limit_is_set_to_zero() {
             .encode(),
```
mraszyk: How about moving the code doing Candid encoding to before the for loop:
```
let bytes = CreateCanisterArgs { ... }.encode();
for i in 0..1_000 {
...
bytes.clone()
...
}
```

alin-at-dfinity: Done, although it doesn't have a measurable effect (test now runs in 0.3s in both guises). Likely it's not the Candid encoding (we still need to decode all 1k arguments, but the `O(N^2)` resulting from `execute_all()` iterating over all canisters, as you observed.

mraszyk (approved):

PR #10280, opened 2026-05-22
feat: [MR-142] Introduce hot/cold-partitioned `CanisterStates`

alin-at-dfinity:
Lays the foundation for splitting `ReplicatedState::canister_states` into "hot" (potentially active) and "cold" (definitely idle) pools, so that per-round operations can completely ignore the long tail of idle canisters.

This PR is intentionally a no-op for the running replica: it only adds the new types and predicates. The integration into `ReplicatedState` and the migration of all consumers follow in subsequent PRs.

Specifically:

  * `CanisterState::is_cold()` — pure predicate that classifies a canister as "definitely idle": no input/output, no heartbeat, not `Stopping`, etc.
  * `CallContextManager::has_unexpired_callbacks()` and the matching `SystemState::has_unexpired_callbacks()` accessor, used by `is_cold`.
  * `CanisterStates`, a hot/cold-partitioned container with eager promotion (mutations land in `hot`) and lazy demotion (via `try_cool`/`try_cool_all`); plus the common map operations, per-pool and merged iterators and bulk mutation.
  * `CanisterStates::validate_strict_split()` for the canonical-partition invariant used in checkpoint validation.
  * `debug_assert_invariants()` runs on every mutating operation in debug builds.


rs/replicated_state/src/canister_states.rs
```diff
@@ -0,0 +1,415 @@
...
+        self.debug_assert_invariants();
+        result
+    }
+
+    /// Retains only the canisters for which the predicate returns true.
+    ///
+    /// Iterates both pools.
+    pub fn retain<F>(&mut self, f: F)
+    where
+        F: Fn(&CanisterId, &Arc<CanisterState>) -> bool,
```
eichhorl: I think this could also be `FnMut` like the other closures in this file
```suggestion
        F: FnMut(&CanisterId, &Arc<CanisterState>) -> bool,
```

alin-at-dfinity: Good point. I was under the impression that the point of `FnMut` is to allow mutation of its arguments; but it's actually about the possibility of mutating captured variables vs potential parallelism.

Done.

rs/replicated_state/src/canister_states/tests.rs
```diff
@@ -0,0 +1,470 @@
...
+
+#[test]
+fn canister_with_pending_task_is_not_cold() {
+    let mut canister = make_canister(1);
+    Arc::make_mut(&mut canister)
+        .system_state
+        .task_queue
+        .enqueue(ExecutionTask::Heartbeat);
+    assert!(!canister.is_cold());
+}
```
eichhorl, line 103: I guess we (AI) could add some more tests of the `is_(not_)cold` variety if we wanted to

rs/replicated_state/src/canister_states.rs
```diff
@@ -0,0 +1,415 @@
...
+        match self.hot.entry(*id) {
+            // In the `hot` pool, return it.
+            Entry::Occupied(entry) => Some(entry.into_mut()),
+
+            Entry::Vacant(entry) => {
+                if let Some(canister) = self.cold.remove(id) {
+                    // Was in the `cold` pool, promote it.
+                    let canister = entry.insert(canister);
+                    // Unfortunately, the borrow checker won't let us do this here.
+                    // self.debug_assert_invariants();
```
eichhorl: If this is important, I guess we could first make the change to the pools, then check the assertions, and then pay for an extra lookup to return the `&mut` from the `hot` pool.

rs/replicated_state/src/canister_states.rs
```diff
@@ -0,0 +1,415 @@
...
+        let to_cool: Vec<CanisterId> = self
+            .hot
+            .iter()
+            .filter(|(_, canister)| canister.is_cold())
+            .map(|(id, _)| *id)
+            .collect();
+        for id in to_cool {
+            let canister = self.hot.remove(&id).unwrap();
+            self.cold.insert(id, canister);
+        }
```
eichhorl: I think we could avoid the extra allocation by using `extract_if`:
```suggestion
        self.cold
            .extend(self.hot.extract_if(.., |_, canister| canister.is_cold()));
```

rs/replicated_state/src/canister_states.rs
```diff
@@ -0,0 +1,415 @@
...
+        for (id, canister) in self.cold.iter_mut() {
+            f(id, canister);
+            if !canister.is_cold() {
+                to_promote.push(*id);
+            }
+        }
+        for id in to_promote {
+            let canister = self.cold.remove(&id).unwrap();
+            self.hot.insert(id, canister);
+        }
```
eichhorl: Also here. One difference is that in case of panics in `is_cold` or `f`, some canisters might have already been promoted (whereas the old code is all-or-nothing). But unless we catch the panic somewhere I don't think it matters.
```suggestion
        self.hot.extend(self.cold.extract_if(.., |id, canister| {
            f(id, canister);
            !canister.is_cold()
        }));
```

rs/replicated_state/src/canister_states.rs
```diff
@@ -0,0 +1,415 @@
...
+    {
+        let mut result = Ok(());
+
+        // Hot pool: short-circuit on the first error.
+        for (id, canister) in self.hot.iter_mut() {
+            if let Err(e) = f(id, canister) {
+                result = Err(e);
+                break;
+            }
+        }
```
eichhorl: Alternatively
```suggestion
        // Hot pool: short-circuit on the first error.
        let mut result = self.hot.iter_mut().try_for_each(|(id, c)| f(id, c));
```

rs/replicated_state/src/canister_states/tests.rs
```diff
@@ -0,0 +1,470 @@
...
+    assert_eq!(states.cold.len(), 1);
+
+    let _ = states.get_mut(&canister.canister_id()).unwrap();
+
+    assert_eq!(states.hot.len(), 1);
+    assert_eq!(states.cold.len(), 0);
+}
+
+#[test]
+fn remove_takes_canister_from_either_pool() {
```
eichhorl: This test is called `remove_takes_canister_from_either_pool` but we actually only test the cold pool

rs/replicated_state/src/canister_states/tests.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    // The canister is still cold by predicate (we didn't mutate it), so try_cool
+    // should demote it.
+    assert!(states.try_cool(&canister.canister_id()));
+    assert_eq!(states.hot.len(), 0);
+    assert_eq!(states.cold.len(), 1);
+}
+
+#[test]
+fn try_cool_leaves_non_cold_in_hot() {
```
eichhorl, line 313: We could also test calling `try_cool` for a canister that is already cool

eichhorl (approved):

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: I tried, but the only way to make it work is to switch from `entry()` to `get_mut()` + `insert()`, which is what I wanted to avoid in the first place.

Reason is, as soon as any bit of code returns a reference returned by `self.entry()`, it means we need to hold a mutable reference on `self` throughout the method. Any other mutable or immutable reference to `self` (including a call to `debug_assert_invariants()`) is off limits.

I did make one simplification, though: I replaced the `if let Some(canister) = self.cold.remove(id)` with `let canister = self.cold.remove(id)?`. That's still something, even if it's just syntactic sugar.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Very nice, thank you.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Done.

If we panic anywhere inside the DSM, the replica crashes. This is by design. So panics are not something we need to worry about in this context.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Done, thanks.

Also managed to combine `extract_if()` with a `RefCell<Response<_, _>>` and a `for` loop that bails out on `Err` to implement an efficient "promote with short-circuit" below.

rs/replicated_state/src/canister_states/tests.rs
alin-at-dfinity: In its final incarnation, the test is called `remove_updates_cold_stats()`. I guess the AI got confused about the naming when I asked it to first create a PR with only the basic `CanisterStates` without stats.

For good measure, renamed it to `remove_takes_canister_from_cold_pool()` and added a separate `remove_takes_canister_from_hot_pool()`.

rs/replicated_state/src/canister_states/tests.rs
alin-at-dfinity: Added, thanks.

rs/replicated_state/src/canister_states/tests.rs
alin-at-dfinity: Added.

PR #10281, opened 2026-05-22
chore: remove unused DeFi rust dependencies

nmattia:
This removes some unused Rust dependencies from the Bazel build.


mbjorkqvist (approved):
Thanks for the cleanup @nmattia!

PR #10286, opened 2026-05-22
feat: [DSM-142] Add `ColdStats` aggregate to `CanisterStates`

alin-at-dfinity:
Maintains a small `ColdStats` aggregate over the canisters in the `cold` pool, updated incrementally on every transition into / out of `cold`. This lets the "touch every canister" aggregate queries — `total_compute_allocation`, `total_canister_memory_usage`, `memory_taken`, `callback_count`, `guaranteed_response_message_memory_taken`, `best_effort_message_memory_taken` — run in `O(|hot|)` instead of `O(|all canisters|)`, which is the primary motivation for the hot/cold split on subnets with a long tail of idle canisters.

The aggregates are derived (not persisted) and are reconstructed by `CanisterStates::new` on checkpoint load. `debug_assert_invariants` ensures every mutating method keeps them in sync, and the `ColdStats` struct stays module-private — callers always reach the totals through the public aggregator methods on `CanisterStates`.

`MemoryTaken`'s fields are bumped from private to `pub(crate)` so that `CanisterStates::memory_taken` can construct the struct directly, keeping `MemoryTaken` in its current home in `replicated_state.rs`. `CanisterStates::memory_taken` itself is `pub(crate)` and will be wired up to `ReplicatedState::memory_taken` in the next PR.


rs/replicated_state/src/canister_states.rs
```diff
@@ -309,18 +458,162 @@ impl CanisterStates {
...
+    pub fn total_compute_allocation(&self) -> u64 {
+        let hot: u64 = self
+            .hot
+            .values()
+            .map(|canister| canister.compute_allocation().as_percent())
+            .sum();
+        hot + self.cold_stats.total_compute_allocation_percent
+    }
+
+    /// Returns the total number of callbacks (responded and unresponded)
```
eichhorl: To me it looks like we only count unresponded callbacks
```suggestion
    /// Returns the total number of unresponded callbacks
```

rs/replicated_state/src/canister_states.rs
```diff
@@ -285,14 +426,22 @@ impl CanisterStates {
...
+        // on `Err`. We always restore the partition / stats for the canister we
         // were visiting when the error occurred, then propagate the error.
         if !result.borrow().is_err() {
-            // Iterator that calls `f` storing its output in `result` and extracts newly hot
-            // canisters.
+            // Iterator that calls `f` storing its output in `result`, extracts newly hot
+            // canisters and updates `cold_stats`.
             let to_promote = self.cold.extract_if(.., |id, canister| {
+                self.cold_stats.sub(canister.as_ref());
                 *result.borrow_mut() = f(id, canister);
```
eichhorl: This is part of the previous PR, but I don't think the `break` in the loop below is enough to short-circuit here. The `extract_if` iterator will only yield an element if the closure returns true (i.e. if the canister is hot after `f`). But if `f` returns an error and the canister stays cold, then we will continue to the next element which might set `result` to `Ok` again.

rs/replicated_state/src/canister_states.rs
```diff
@@ -309,18 +458,162 @@ impl CanisterStates {
...
+        wasm_custom_sections += cold.wasm_custom_sections_memory;
+        canister_history += cold.canister_history_memory;
+
+        MemoryTaken {
+            execution,
+            guaranteed_response_messages,
+            best_effort_messages,
+            wasm_custom_sections,
+            canister_history,
+        }
```
eichhorl: Alternatively, maybe something like the following would work:
```rust
let mut memory_taken = MemoryTaken {
    // Initialize based on cold stats
}
for canister in self.hot {
   // Add stats to memory_taken
}
memory_taken
```

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: AI slop, my apologies.

"Unresponded callbacks" is a historical artifact from when it was computed as number of callbacks minus one, iff there's a paused or aborted execution. That got simplified, but the name stayed.

It then somehow went and confused the AI, who must have thought there exists such a thing as a responded callback.

Cleaned up the doc comment, thanks.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Wow, good catch. Can't even blame the AI on this one, it's my code.

This logic is "only" used to produce an error that would cause us to panic during checkpoint loading. I say "only" because said loading error is not supposed to happen. But if it does happen, we definitely don't want it overwritten by an `Ok(())`.

Added a test and fixed, retaining the short-circuiting behavior.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Indeed, this is much nicer to read. Thanks.

rs/replicated_state/src/canister_states.rs
```diff
@@ -282,45 +423,182 @@ impl CanisterStates {
...
+    ///
+    /// Does **not** include subnet queues. Call `ReplicatedState::memory_taken`
+    /// for the actual subnet-wide memory usage.
+    ///
+    /// `O(|hot canisters|)` thanks to the precomputed cold-pool aggregate.
+    // The only non-test caller of this method is
+    // `ReplicatedState::memory_taken`, which is wired up in the follow-up
+    // integration PR.
+    #[allow(dead_code)]
+    pub(crate) fn memory_taken(&self) -> MemoryTaken {
```
eichhorl, line 584: Is this function `pub(crate)` on purpose? Everything else is `pub`

rs/replicated_state/src/canister_states.rs
```diff
@@ -282,45 +423,182 @@ impl CanisterStates {
...
+                result
             });
-            // Consumes the iterator adding its output to `hot`. Short-circuits on `Err`.
-            for (id, canister) in to_promote {
-                self.hot.insert(id, canister);
-                if result.borrow().is_err() {
-                    break;
-                }
+            // Next canister ID, iff we short-circuited.
+            let next_id = cold_iter_mut.next().map(|(id, _)| *id);
```
eichhorl: Alternatively
```suggestion
            let end = cold_iter_mut.next().map_or(Bound::Unbounded, |(id, _)| Bound::Excluded(*id));
            // ...
            self.cold.extract_if((Bound::Unbounded, end), ...)
            
```

rs/replicated_state/src/canister_states/tests.rs
```diff
@@ -551,6 +651,173 @@ fn retain_drops_canisters_from_both_pools() {
...
+        states.best_effort_message_memory_taken(),
+        expected_best_effort
+    );
+
+    let mt = states.memory_taken();
+    assert_eq!(mt.execution, expected_execution);
+    assert_eq!(mt.guaranteed_response_messages, expected_guaranteed);
+    assert_eq!(mt.best_effort_messages, expected_best_effort);
+    assert_eq!(mt.wasm_custom_sections, expected_wasm_sections);
+    assert_eq!(mt.canister_history, expected_history);
```
eichhorl, line 772: Claude says that in the current setup, these two fields are zero which means the test also passes if they aren't summed up by `memory_taken`. So ideally all stats should be > 0

eichhorl (approved):

rs/replicated_state/src/canister_states.rs
eichhorl: Thanks for the explanation!

rs/replicated_state/src/canister_states.rs
eichhorl: I see, makes sense. Since the error is not supposed to happen, I guess we could also try to simplify this function by not optimizing for the error case, i.e. accept that we will always scan both pools entirely. But I'm also fine with the current implementation.

alin-at-dfinity: It's a public method and you never know what it's going to be used for. Which is why I figured it may as well have an optimal implementation.

Of course, I wasn't expecting something as elaborate as what I ended up with, but since it's here (and considering the above), I'd rather keep it.

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: Yeah, there's a "great" tradition in the Replicated State to expose the same method at different layers, so you can e.g. call `CanisterState::canister_id()` if you have a `CanisterState`; or `SystemState::canister_id()` if what you have is a `SystemState`. Those methods are functionally equivalent.

But there are also cases, such as `push_input()`, which exists for `CanisterState`, `SystemState` and `CanisterQueues`, with increasing layers of validation and side-effects. For such cases, only the top-level method is `pub`, to ensure e.g. that resource bounds are respected.

Because of the caveat that this only covers canisters, not the subnet queue, this is more like the latter, as we don't want someone accidentally calling `CanisterStates::memory_taken()` and assuming it's the same as `ReplicatedState::memory_taken()` (or just mistaking it for the other method).

But having written all this treatise I realize you do have a point, of course. Namely, if we don't want `memory_taken()` to be mistaken as its `ReplicatedState` equivalent, the same applies to `best_effort_message_memory_taken()` and `guaranteed_response_message_memory_taken()`. (`total_compute_allocation()`, `callback_count()` and `total_canister_memory_usage()` are actual canister-only stats, so it's OK for them to be `pub`.)

Also applied `pub(crate)` to `best_effort_message_memory_taken()` and `guaranteed_response_message_memory_taken()`.

rs/replicated_state/src/canister_states/tests.rs
alin-at-dfinity: Done (also for `best_effort_messages`). And rewrote the test so it tests with both hot and cold canisters having / not having relevant bits of the state populated (so we cover both precomputed cold canister stats and computed on-the-fly hot canister stats).

rs/replicated_state/src/canister_states.rs
alin-at-dfinity: I just feel stupid at this point. 🤡

I did look for something exactly like this (in the library code, around `RangeBounds`; and online) and all I found were the various incompatible `Range` types.

Done, thanks for the tip.

rs/replicated_state/src/canister_states.rs
eichhorl: I see, thanks for the explanation!

rs/replicated_state/src/canister_states.rs
eichhorl: > (in the library code, around RangeBounds; and online)

I just asked Claude 😆

PR #10287, opened 2026-05-22
feat: [DSM-142] Use `CanisterStates` in `ReplicatedState`

alin-at-dfinity:
[Context: a3ba27a and 7813995.]

Switches `ReplicatedState::canister_states` from a flat `BTreeMap<CanisterId, Arc<CanisterState>>` to `CanisterStates`, exposing the hot/cold partition to the rest of the system and migrating every caller.

`ReplicatedState` changes:

  * `canister_states` field is now `CanisterStates`.
  * Drop `canisters_iter_mut()`. Round-level callers move to `hot_canisters_iter_mut()` (skips the long tail of cold canisters); bulk callers move to `canisters_for_each_mut` / `canisters_try_for_each_mut`, which iterate every canister and re-establish the partition afterwards.
  * Add `hot_canisters_iter()` for read-only hot-only iteration.
  * Add `repartition_canister_states()`, called from `StateManager::commit_and_certify` after `flush_checkpoint_ops_and_page_maps` to drop canisters that went idle during the round back into `cold` before checkpointing, so that replicas continuing through a checkpoint and replicas (re)starting from it agree on the partition.
  * `take_canister_states` / `put_canister_states` now return / accept `CanisterStates` instead of `BTreeMap`.
  * Aggregator delegations: `total_compute_allocation`, `memory_taken`, `total_canister_memory_usage`, `guaranteed_response_message_memory_taken`, `best_effort_message_memory_taken`, `callback_count` now delegate to `CanisterStates` and run in `O(|hot|)`.

`state_manager`:

  * `commit_and_certify` calls `state.repartition_canister_states()` after `flush_checkpoint_ops_and_page_maps` and before tip handover.
  * `validate_eq_canister_states` calls `CanisterStates::validate_strict_split` on the reference state to verify that the persisted partition matches what `CanisterStates::new` would produce on a fresh load.
  * `flush_checkpoint_ops_and_page_maps` and `switch_to_checkpoint` switch from `canisters_iter_mut` to `canisters_for_each_mut` / `canisters_try_for_each_mut`.
  * Bench: `bench_traversal` likewise.

`execution_environment`:

  * `scheduler.rs`: scheduler hot-only iteration where appropriate (`add_heartbeat_and_global_timer_tasks`, `purge_expired_ingress_messages`, the `ongoing_long_install_code` check); migrate `charge_canisters_for_resource_allocation_and_usage` and the log-memory-store migration loop to `canisters_for_each_mut`.
  * `round_schedule.rs`: `partition_canisters_to_cores` now takes / returns a `CanisterStates`; idle canisters are dropped before the main hot-canister iteration.
  * `query_handler.rs`, `execution_environment.rs`: callers updated.
  * `canister_manager/tests.rs`, scheduler tests (`scheduling.rs`, `metrics.rs`, `dts.rs`, `ecdsa.rs`, `round_schedule/tests.rs`, `test_utilities.rs`, `tests.rs`) updated.
  * `benches/scheduler.rs`: updated.

`canonical_state`:

  * `lazy_tree_conversion.rs`: new `CanisterStatesFork<'_>` that presents a `CanisterStates` as a `LazyFork` over the merged hot+cold pools in `CanisterId` order.

`messaging`:

  * `stream_builder/tests.rs`, `state_machine/tests.rs`, `tests/common/mod.rs`: caller updates.

`replicated_state` queues and system_state:

  * `CanisterQueues` / `SystemState` `local_canisters` parameter type flips from `&BTreeMap<CanisterId, Arc<CanisterState>>` to `&CanisterStates` (no behavioral change; queues only need `contains_key`).

`metrics.rs`:

  * `check_dts` walks `hot_canisters_iter()` (only hot canisters can have non-empty task queues).
  * `check_subnet_memory_usage` switches to `CanisterStates::memory_taken()` for `O(|hot|)` aggregation.

`test_utilities` and `state_tool`:

  * `test_utilities/execution_environment`, `test_utilities/state`, and `state_tool/src/commands/canister_metrics.rs` updated to use the new iteration APIs.


rs/replicated_state/src/replicated_state.rs
```diff
@@ -1179,7 +1168,7 @@ impl ReplicatedState {
 
     /// Garbage collects empty canister and subnet queues.
     pub fn garbage_collect_canister_queues(&mut self) {
-        for (_canister_id, canister) in self.canister_states.iter_mut() {
+        for canister in self.canister_states.hot_values_mut() {
```
eichhorl: Claude complains that cold canisters may still have empty queues which will now no longer be collected

rs/state_machine_tests/src/lib.rs
```diff
@@ -5218,7 +5218,7 @@ impl StateMachine {
             .collect();
         let (_height, mut replicated_state) = self.state_manager.take_tip();
         let mut synthetic_responses = vec![];
-        for canister_state in replicated_state.canisters_iter_mut() {
+        for canister_state in replicated_state.hot_canisters_iter_mut() {
```
eichhorl: Claude complains that a canister blocked on an outstanding remote guaranteed-response call is cold (it holds only an input-slot reservation, with no enqueued messages, tasks or unexpired best-effort callbacks), yet its remote callback still needs rejecting.

rs/replicated_state/src/replicated_state.rs
alin-at-dfinity: Great point, thank you.

This is merely an optimization (we don't write an empty `queues.pbuf` file for a canister with no queues), but it's a really nice to have optimization. And only GC-ing active canisters' queues would have broken it completely.

rs/state_machine_tests/src/lib.rs
alin-at-dfinity: Good point, fixed.

rs/test_utilities/execution_environment/src/lib.rs
```diff
@@ -2206,11 +2206,14 @@ impl ExecutionTest {
...
+        let state = self.state_mut();
+        state.canisters_for_each_mut(|_id, canister_state| {
             let canister_state = Arc::make_mut(canister_state);
-            let es = match canister_state.execution_state.as_mut() {
-                Some(es) => es,
-                None => break,
+            let Some(es) = canister_state.execution_state.as_mut() else {
+                // Nothing to checkpoint for a canister without an execution state.
+                //
+                // Note: This used to be a `break`, but likely `continue` was intended.
```
eichhorl: Doesn't seem that relevant
```suggestion
```

rs/execution_environment/src/scheduler.rs
eichhorl: I guess technically we no longer iterate over all canisters

rs/execution_environment/src/scheduler.rs
```diff
@@ -2138,7 +2137,7 @@ pub fn abort_all_paused_executions(
     log: &ReplicaLogger,
 ) {
     let (canister_states, subnet_schedule) = state.canisters_and_schedule_mut();
-    for canister in canister_states.values_mut() {
+    for canister in canister_states.hot_values_mut() {
         abort_canister(canister, subnet_schedule, exec_env, cost_schedule, log);
     }
```
eichhorl, line 2142: Claude points out that this relies on an invariant where an ingress induction debit greater 0 implies the presence of a paused task. Instead, it suggests to also enforce `debit == 0` as part of `is_cold`

eichhorl (approved):
I double checked the places where we went from "all canisters" to "hot only" against the definition of `is_cold`, and they LGTM.

Should there be any tests for these changes, including the two potential regressions above?

rs/test_utilities/execution_environment/src/lib.rs
alin-at-dfinity: Indeed, this was intended to be converted into a GitHub comment to the reviewer, but I completely forgot about it.

Thanks for the catch, dropped.

rs/execution_environment/src/scheduler.rs
alin-at-dfinity: Updated, thanks.

rs/execution_environment/src/scheduler.rs
alin-at-dfinity: Added.

The whole `ingress_induction_cycles_debit` construct is kind of fragile, relying on all code paths out of a DTS execution applying the debit. But at least this won't make it even more rickety.

alin-at-dfinity (commented):
> I double checked the places where we went from "all canisters" to "hot only" against the definition of is_cold, and they LGTM.
>
> Should there be any tests for these changes, including the two potential regressions above?

I've considered how to ensure that `is_cold()` doesn't fall out of sync with whatever the users of `hot_canisters_iter()` and friends expect from it. I couldn't come up with anything practical. And I believe any issues are likely to come from new uses of these iterators, not so much from changes in semantics of the existing uses. (And if someone does change the semantics, they will be unlikely to update the relevant unit test.)

PR #10288, opened 2026-05-22
refactor: Hoist idle-canister drop in `RoundSchedule::start_iteration`

alin-at-dfinity:
Move the "drop idle canisters with 0-100 AP from the subnet schedule" logic out of the `NextExecution::None` branch of the main per-canister loop and into a dedicated pre-loop at the top of `start_iteration`. Behavior is unchanged: the same set of idle canisters with priorities in the 0-100 AP range get dropped.

Also clarify the doc comment for
`IterationSchedule::partition_canisters_to_cores`.

This is a small standalone refactor extracted from #10287, where the main per-canister loop will switch from iterating all canisters to iterating only hot canisters (at which point hoisting becomes a correctness requirement: cold canisters would otherwise no longer be visited by the main loop and their idle entries would not be dropped).


rs/execution_environment/src/scheduler/round_schedule.rs
```diff
@@ -307,35 +344,22 @@ impl RoundSchedule {
                         if self.long_execution_canisters.contains(canister_id) {
                             return None;
                         }
-                        CanisterRoundState::new(canister, subnet_schedule.get_mut(*canister_id))
+                        CanisterRoundState::new(canister, subnet_schedule.get(canister_id))
```
alin-at-dfinity, line 335: FYI: This replacement (and the similar one right below) result in a change in behavior: canisters no longer get a `SubnetSchedule` entry as soon as they are scheduled by `start_iteration()`.

But this was accidentally left in (during development I had a version of `CanisterRoundState::new()` that actually mutated the provided `CanisterPriority` (don't remember why). And `finish_round()` will call `get_mut()` to create an entry for all scheduled canisters. And no one looks at the `SubnetSchedule` in-between.

rs/execution_environment/src/scheduler/round_schedule.rs
```diff
@@ -277,6 +279,41 @@ impl RoundSchedule {
...
+                        // Canister is idle: drop.
+                        Some(*canister_id)
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<_>>();
+            for canister_id in idle_canisters_to_drop {
+                subnet_schedule.remove(&canister_id);
+            }
```
eichhorl: Could this be simplified by making a single call to `subnet_schedule.retain(...)`?

alin-at-dfinity: Of course. As soon as I hit Enter, AI autocomplete suggested `subnet_schedule.retain(...` :disappointed:

Done. Thank you for the suggestion.

eichhorl (approved):

PR #6761, opened 2025-09-16
chore(XC): remove deprecation from `ic-cdk` in `ic-ckbtc-minter`

gregorydemay:
Follow-up to #6264 to remove the introduced `#![allow(deprecated)]` in the code of the `ic-ckbtc-minter` canister:

1. in [`rs/bitcoin/ckbtc/minter/src/main.rs`](https://github.com/dfinity/ic/blob/3794e29f4a7b0174dd40539262a22ef48bd7d6de/rs/bitcoin/ckbtc/minter/src/main.rs#L1)
2. in [`rs/bitcoin/ckbtc/minter/src/lib.rs`](https://github.com/dfinity/ic/blob/3794e29f4a7b0174dd40539262a22ef48bd7d6de/rs/bitcoin/ckbtc/minter/src/lib.rs#L1)
3. in [`rs/bitcoin/ckbtc/minter/tests/tests.rs`](https://github.com/dfinity/ic/blob/3794e29f4a7b0174dd40539262a22ef48bd7d6de/rs/bitcoin/ckbtc/minter/tests/tests.rs#L1)

The changes in the ckBTC minter API are due to changes in the return type of `get_canister_status`, which is a debug endpoint where backwards-compatibility is not guaranteed.

PR #6264, opened 2025-08-13
feat!(deps): Unify ic-cdk to v0.18.5

lwshang:

PR #10290, opened 2026-05-22
chore(defi): remove deprecated ic-cdk imports in ic-cketh-minter

gregorydemay:
## Summary

Continues [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the `#![allow(deprecated)]` attributes that were introduced in #6264 when `ic-cdk` was upgraded to 0.18, by migrating the affected files off the deprecated `ic_cdk` surface.

This PR handles `ic-cketh-minter`:

- `ic_cdk::api::management_canister::{main, ecdsa, http_request}` → `ic_cdk::management_canister::*` + `EcdsaPublicKeyArgs` / `EcdsaPublicKeyResult` / `SignWithEcdsaArgs` / `CanisterStatusArgs` / `CanisterStatusResult` from `ic_management_canister_types`.
- `ic_cdk::api::call::RejectionCode` → `ic_cdk::call::CallFailed` and `ic_cdk::management_canister::SignCallError`; the existing `Reason::from_reject` mapping is rewritten as `from_call_failed` / `from_sign_call_error` while preserving the same `Reason` variants.
- `ic_cdk::caller()` / `ic_cdk::id()` → `ic_cdk::api::msg_caller()` / `ic_cdk::api::canister_self()`.
- `ic_cdk::api::canister_balance128()` → `ic_cdk::api::canister_cycle_balance()`.
- `ic_cdk::api::stable::stable_size()` → `ic_cdk::stable::stable_size()`.
- `get_canister_status` now returns the modern `CanisterStatusResult`; `cketh_minter.did` is updated to mirror the new fields (the same shape already adopted by ckBTC minter in #6761).

The changes in the ckETH minter API are due to changes in the return type of `get_canister_status`, which is a debug endpoint where backwards-compatibility is not guaranteed.

Follows the pattern set by #6755 (`ic-btc-checker`) and #6761 (`ic-ckbtc-minter`). No behavior change.

### DIFF
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -732,6 +732,8 @@ service : (MinterArg) -> {
     is_address_blocked : (text) -> (bool) query;
 
     // Retrieve the status of the minter canister.
+    //
+    // This is a debug endpoint where backwards-compatibility is not guaranteed.
     get_canister_status : () -> (CanisterStatusResponse);
 
     // Retrieve events from the minter's audit log.
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -7,18 +7,37 @@ type EthereumNetwork = variant {
 
 type Subaccount = blob;
 
+type environment_variable = record {
+    name: text;
+    value: text;
+};
+
 type CanisterStatusResponse = record {
-  query_stats : QueryStats;
   status : CanisterStatusType;
+  ready_for_migration : bool;
+  version : nat64;
   memory_size : nat;
   cycles : nat;
   settings : DefiniteCanisterSettings;
   idle_cycles_burned_per_day : nat;
   module_hash : opt vec nat8;
+  query_stats : QueryStats;
   reserved_cycles : nat;
+  memory_metrics : MemoryMetrics;
 };
 
 type CanisterStatusType = variant { stopped; stopping; running };
+
+type MemoryMetrics = record {
+    wasm_memory_size : nat;
+    stable_memory_size : nat;
+    global_memory_size : nat;
+    wasm_binary_size : nat;
+    custom_sections_size : nat;
+    canister_history_size : nat;
+    wasm_chunk_store_size : nat;
+    snapshots_size : nat;
+};
 
 type DefiniteCanisterSettings = record {
   freezing_threshold : nat;
@@ -28,6 +47,8 @@ type DefiniteCanisterSettings = record {
   reserved_cycles_limit : nat;
   log_visibility: LogVisibility;
   wasm_memory_limit : nat;
+  wasm_memory_threshold : nat;
+  environment_variables : vec environment_variable;
 };
 
 type LogVisibility = variant {
--- a/rs/ethereum/cketh/minter/src/main.rs
+++ b/rs/ethereum/cketh/minter/src/main.rs
@@ -1,4 +1,3 @@
-#![allow(deprecated)]
 use crate::dashboard::DashboardPaginationParameters;
 use candid::Nat;
 use dashboard::DashboardTemplate;
@@ -60,7 +59,7 @@ pub const SEPOLIA_TEST_CHAIN_ID: u64 = 11155111;
 pub const CKETH_LEDGER_TRANSACTION_FEE: Wei = Wei::new(2_000_000_000_000_u128);
 
 fn validate_caller_not_anonymous() -> candid::Principal {
-    let principal = ic_cdk::caller();
+    let principal = ic_cdk::api::msg_caller();
     if principal == candid::Principal::anonymous() {
         panic!("anonymous principal is not allowed");
     }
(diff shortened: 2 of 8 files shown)


rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -7,19 +7,38 @@ type EthereumNetwork = variant {
...
 
+type MemoryMetrics = record {
+    wasm_memory_size : nat;
+    stable_memory_size : nat;
+    global_memory_size : nat;
+    wasm_binary_size : nat;
+    custom_sections_size : nat;
+    canister_history_size : nat;
+    wasm_chunk_store_size : nat;
+    snapshots_size : nat;
```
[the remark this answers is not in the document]
gregorydemay: Not true,. current type comes from ic_cdk.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -575,15 +574,12 @@ async fn add_ckerc20_token(erc20_token: AddCkErc20Token) {
 }
 
 #[update]
-async fn get_canister_status() -> ic_cdk::api::management_canister::main::CanisterStatusResponse {
-    ic_cdk::api::management_canister::main::canister_status(
-        ic_cdk::api::management_canister::main::CanisterIdRecord {
-            canister_id: ic_cdk::id(),
-        },
-    )
+async fn get_canister_status() -> ic_cdk::management_canister::CanisterStatusResult {
```
mbjorkqvist, line 577: I was wondering if it makes sense to return this type directly, rather than doing some translation inside the canister and decouple the response type from the management canister type - the management canister type seems to have (frequent) breaking changes which are not candid-backwards-compatible (e.g., adding and removing non-optional fields). For anyone calling the `ic-cketh-minter` canister, this will be a breaking change that is not necessarily trivial to prepare for and handle. On the other hand, calling this endpoint is not on the regular user flow path, so it's probably not such a big deal, and anyway this is probably something to potentially discuss outside of this PR.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -7,19 +7,38 @@ type EthereumNetwork = variant {
 
 type Subaccount = blob;
 
+type environment_variable = record {
+    name: text;
+    value: text;
+};
```
mbjorkqvist, line 13: nit: The naming of this type is inconsistent with most other types in this file (e.g., `QueryStats`, `MemoryMetrics`), but it matches what's in the ckBTC minter, so I suppose no action needed.

rs/ethereum/cketh/minter/src/management/mod.rs
```diff
@@ -64,17 +64,39 @@ impl fmt::Display for Reason {
...
+                    Ok(code) => Self::InternalError(format!(
+                        "rejection code: {code:?}, rejection message: {message}"
+                    )),
+                    Err(_) => Self::InternalError(format!(
+                        "unrecognized rejection code: {}, rejection message: {message}",
+                        rejected.raw_reject_code()
+                    )),
+                }
+            }
+            CallFailed::InsufficientLiquidCycleBalance(_) => Self::OutOfCycles,
```
mbjorkqvist, line 99: This seems to be new behavior: I couldn't find any codepaths that previously led to an `OutOfCycles` error being returned - I suppose an `InternalError` would have been returned previously? Would this be worth mentioning somewhere, and potentially adding a test?

mbjorkqvist (commented):
Thanks @gregorydemay! Just a few minor comments.

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: Good question. I think we should clearly mark get_canister_status as a debug endpoint. I agree that what you suggest makes sense if backwards-compatibility is a concern.

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: Good catch, I would stay consistent here between ckBTC/ckETH.

rs/ethereum/cketh/minter/src/management/mod.rs
gregorydemay: 🤖 Good catch — you're right that `Reason::OutOfCycles` was effectively dead code before this PR. The new code path comes from the ic-cdk 0.19 `Call` builder, which now does a pre-flight check comparing `canister_liquid_cycle_balance` against `Call::get_cost` and surfaces `CallFailed::InsufficientLiquidCycleBalance` *before* `ic0.call_perform`. The old `call_with_payment128` surface had no such check, so an insufficient-cycles condition would have manifested downstream and ended up mapped to `Reason::InternalError`. Mapping the new variant to `Reason::OutOfCycles` is intentional — it gives a more accurate error message (`"the canister is out of cycles"` vs. a generic `internal error: ...`).

Pushed `622f438` with a unit test for that mapping plus three companions covering the existing reject-code / call-perform-failed / unrecognized-code paths.

mbjorkqvist (approved):

PR #10291, opened 2026-05-22
chore(defi): remove deprecated ic-cdk imports in ic-ledger-suite-orchestrator

gregorydemay:
## Summary

Closes the last of the three remaining crates listed in [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the `#![allow(deprecated)]` attributes introduced in #6264 when `ic-cdk` was upgraded to 0.18, by migrating the affected files off the deprecated `ic_cdk` surface.

This PR handles `ic-ledger-suite-orchestrator`:

- `ic_cdk::api::management_canister::main::{CanisterIdRecord, CanisterStatusResponse, canister_status, start_canister, stop_canister}` → `ic_cdk::management_canister::{CanisterStatusArgs, CanisterStatusResult, StartCanisterArgs, StopCanisterArgs, canister_status, start_canister, stop_canister}`.
- `ic_cdk::api::call::{call, call_with_payment}` and `notify_with_payment128` in `IcCanisterRuntime` are rewritten on top of the `ic_cdk::call::Call` builder (`Call::unbounded_wait(...).with_arg(...).with_cycles(...)` and `.oneway()` for deposit_cycles).
- `ic_cdk::api::call::RejectionCode` → `ic_cdk::call::{CallFailed, OnewayError, RejectCode}`; the existing `Reason::from_reject` mapping is split into `from_call_error` / `from_call_failed` / `from_oneway_error` helpers that share a common `from_reject_code_message` core, preserving the same `Reason` variants.
- `ic_cdk::id()` → `ic_cdk::api::canister_self()`.
- `ic_cdk::api::canister_balance128()` → `ic_cdk::api::canister_cycle_balance()`.
- `ic_cdk::api::stable::stable_size()` → `ic_cdk::stable::stable_size()`.
- `get_canister_status` now returns the modern `CanisterStatusResult`; `ledger_suite_orchestrator.did` is updated to mirror the new fields (the same shape already adopted by ckBTC minter in #6761 and ckETH minter in #10290).

`create_canister` still pays the exact `cycles_for_canister_creation` amount the caller supplies (rather than letting the cdk auto-compute the cost via `create_canister_with_extra_cycles`), so the orchestrator's existing semantics around cycle accounting are unchanged.

Follows the pattern set by #6755 (`ic-btc-checker`) and #6761 (`ic-ckbtc-minter`). No behavior change.

### DIFF
--- a/rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
+++ b/rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
@@ -1,12 +1,12 @@
-#![allow(deprecated)]
 use crate::logs::DEBUG;
 use async_trait::async_trait;
 use candid::{CandidType, Encode, Principal};
-use ic_base_types::PrincipalId;
 use ic_canister_log::log;
-use ic_cdk::api::call::RejectionCode;
-use ic_management_canister_types::{
-    CanisterIdRecord, CanisterInstallMode, CanisterSettings, CreateCanisterArgs, InstallCodeArgs,
+use ic_cdk::call::{Call, CallFailed, OnewayError, RejectCode};
+use ic_cdk::management_canister::{
+    CanisterInstallMode, CanisterSettings, CanisterStatusArgs, CreateCanisterArgs,
+    DepositCyclesArgs, InstallCodeArgs, StartCanisterArgs, StopCanisterArgs, canister_status,
+    install_code, start_canister, stop_canister,
 };
 use serde::de::DeserializeOwned;
 use std::fmt;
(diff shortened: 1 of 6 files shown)


rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
```diff
@@ -206,44 +217,53 @@ impl CanisterRuntime for IcCanisterRuntime {
             "BUG: too many controllers. Expected at most 10, got {}",
             controllers.len()
         );
+        let payment = u128::from(cycles_for_canister_creation);
+        let balance = ic_cdk::api::canister_cycle_balance();
```
mbjorkqvist: nit: I think it would be more accurate to call [`canister_liquid_cycle_balance`](https://docs.rs/ic-cdk/0.18.7/ic_cdk/api/fn.canister_liquid_cycle_balance.html) here, but we would anyway return `OutOfCycles` from the `create_canister` call - the only difference would be that in case `canister_cycle_balance` is enough, but the liquid balance is too low, we'd end up making a call to `create_canister` that we could otherwise have avoided. Perhaps a test for this case could make sense (setting a high freezing threshold would reserve a lot of cycles)?

rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
```diff
@@ -75,16 +75,59 @@ impl fmt::Display for Reason {
...
-    fn from_reject(reject_code: RejectionCode, reject_message: String) -> Self {
+    fn from_call_error(error: ic_cdk::call::Error) -> Self {
+        use ic_cdk::call::Error;
+        match error {
+            Error::CallRejected(rejected) => Self::from_reject_code_message(
+                rejected.reject_code(),
+                rejected.raw_reject_code(),
+                rejected.reject_message().to_string(),
+            ),
+            Error::InsufficientLiquidCycleBalance(_) => Self::OutOfCycles,
```
mbjorkqvist, line 89: Same comment as in [the other PR](https://github.com/dfinity/ic/pull/10290#discussion_r3309556911) - this seems like new behavior, and it could make sense to add a test.

mbjorkqvist (commented):

rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
gregorydemay: 🤖 Good catch — `824644f` switches the pre-flight to `ic_cdk::api::canister_liquid_cycle_balance()` so it matches what `Call::await` will check internally (the variant that becomes `CallFailed::InsufficientLiquidCycleBalance`). Same `Reason::OutOfCycles` end result, but no longer wastes a `Call::await` when liquid balance is below `payment` and the raw cycle balance is above it.

I left the high-freezing-threshold integration test as a follow-up for now: it would need a new `LedgerSuiteOrchestrator::with_high_freezing_threshold` helper in `test_utils` and a dedicated test against `create_canister`, and the unit tests I added in `src/management/tests.rs` already lock in the `InsufficientLiquidCycleBalance → OutOfCycles` mapping. Happy to add the integration test in a follow-up PR if you think it would be valuable.

rs/ethereum/ledger-suite-orchestrator/src/management/mod.rs
gregorydemay: 🤖 Same context as on https://github.com/dfinity/ic/pull/10290#discussion_r3316157301: the mapping is intentional new behaviour from the cdk 0.18 surface — `Reason::OutOfCycles` was a defined-but-never-produced variant before, and routing `InsufficientLiquidCycleBalance` to it yields a more accurate error message than the previous `InternalError` fallback.

`824644f` adds unit tests in `src/management/tests.rs` covering this for all three mapping helpers (`from_call_error`, `from_call_failed`, `from_oneway_error`) plus the reject-code and `CallPerformFailed` paths to lock in the existing shape.

mbjorkqvist (approved):

PR #10292, opened 2026-05-22
feat: Charge a base cost per canister

schneiderstefan:
This PR introduces a base cost charged to each canister, in addition to the existing cost for a canister's memory footprint. The cost is 10k cycles/s, which is roughly 0.026T cycles/month.

The motivation is to correctly account for the fact that canisters cause overhead, even if they have a small memory footprint.


rs/execution_environment/tests/dts.rs
```diff
@@ -467,9 +467,15 @@ fn dts_install_code_with_concurrent_ingress_insufficient_cycles_and_freezing_thr
...
+    // base_per_second_fee also contributes to the freeze threshold.
+    let base_per_second_fee_cycles =
+        config.cycles_account_manager_config.base_per_second_fee * freezing_threshold;
+
     // The initial balance is sufficient to only pay the reservation for installing code
-    // and the compute allocation during the freezing threshold.
-    let initial_balance = max_install_code_cost() + compute_allocation_cycles;
+    // and the freeze threshold (compute allocation + base fee).
+    let initial_balance =
+        max_install_code_cost() + compute_allocation_cycles + base_per_second_fee_cycles;
```
alin-at-dfinity, line 478: Idly wondering: This just occurred to me, after having gone through more than half the PR. Wouldn't it be simpler (at least for the tests not focused on charging) to set the base fee to zero?

alin-at-dfinity (approved):

PR #10208, opened 2026-05-12
feat: [NODE-1956] Update CI/build images to 26.04

Bownairo:
Move the build containers to the new Ubuntu LTS, 26.04.

Previous determinism issues have been addressed with https://github.com/dfinity/ic/pull/10277, https://github.com/dfinity/ic/pull/10303, and https://github.com/dfinity/ic/pull/10331.

PR #10303, opened 2026-05-23
fix: Wrap a few determinism holes

Bownairo:
Following #10277.

Simplify bootloader pinning, and remove the fs wrappers. Pin `zstd` used within the build.

---

In testing across the upgrade from #10208 (where these issues first appeared), I found that an upgraded `zstd` in the build container is responsible for the differences as seen in https://github.com/dfinity/ic/actions/runs/26249865093/job/77264840363.

This is most clear with `//ic-os/setupos/envs/prod:partition-config.tzst`, since it is just an empty partition. The tars contained are the exact same. I've tested this with both the artifacts from CI, and ones built locally on either side of the upgrade.

For example:
```
522daba85ed385ead014b342460482f05be24c951b8dba95a6e71b248041f4b4  A/partition-config/partition-config.tzst
524588ad45eba3076cfd8d67afda09b4d0314e97209517619eef9e8329923653  B/partition-config/partition-config.tzst
71bec378fc2a0bc81b16597cc1446101da44959d0db6142d5b4ad96d1c166533  A/partition-config/partition-config.tar
71bec378fc2a0bc81b16597cc1446101da44959d0db6142d5b4ad96d1c166533  B/partition-config/partition-config.tar
```
Inside the compressed format, the content checksum is the same, so even `zstd` knows the files are the same.

I guess the new `zstd` is one byte worse on compression? 😅
```
wc -c old/partition-config.tzst new/partition-config.tzst
1670 old/partition-config.tzst
1671 new/partition-config.tzst
3341 total
```

### DIFF
--- a/.github/actions/bazel/action.yaml
+++ b/.github/actions/bazel/action.yaml
@@ -158,6 +158,9 @@ runs:
           ^@@rules_rust..crate.crate_index__cranelift-isle            # has a non reproducible isle_tests.rs in its OUT_DIR.
           ^@@rules_rust..crate.crate_index__secp256k1-sys             # has non reproducible object files, like lax_der_parsing.o, in it OUT_DIR.
           ^@@rules_rust..crate.crate_index__sev                       # build.rs depends on the presence of /dev/sev and /dev/sev-guest. See: https://github.com/virtee/sev/issues/315
+          # TODO: fix the following ASAP:
+          ^//ic-os/hostos/envs/prod:rootfs-tree.tar                   # became non-reproducible because of https://github.com/dfinity/ic/pull/10208
+          ^//ic-os/setupos/envs/prod:rootfs-tree.tar                  # idem
         )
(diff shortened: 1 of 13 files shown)


frankdavid (approved):

basvandijk (approved):
Wouldn't it be better to leave mkfs.vfat and mkfs.ext4 hermetic? The fewer deps we use from the env the easier it becomes to move to [remote execution](https://bazel.build/remote/rbe) and the fewer risk of non-determinism.

Bownairo:
> Wouldn't it be better to leave mkfs.vfat and mkfs.ext4 hermetic? The fewer deps we use from the env the easier it becomes to move to [remote execution](https://bazel.build/remote/rbe) and the fewer risk of non-determinism.

I was thinking we tackle just the determinism issues we saw from the upgrade, for now, then we can all meet on what to do with the rest of these tools. There might be more we can do here than wrapping every binary. For example, @frankdavid had a big PR some time ago that dropped many of these binaries entirely.

PR #10305, opened 2026-05-24
feat: [DSM-103] Charge for storage after aborting DTS executions

alin-at-dfinity:
When charging canisters for storage, we skip canisters with paused executions because they should have a constant cycle balance throughout the DTS execution. But we do charge canisters with aborted executions. Swap things around, so that we only charge for storage after having aborted paused executions (whether executions above the limit on regular rounds; or all paused executions on checkpoint rounds).

This ensures that we can charge all canisters on checkpoint rounds (and makes it more likely to charge the average busy canister on regular rounds), making it possible to eventually move away from trying to charge every canister every round just to make sure we succeed.

### DIFF
--- a/rs/execution_environment/tests/dts.rs
+++ b/rs/execution_environment/tests/dts.rs
@@ -1889,7 +1889,7 @@ fn impl_dts_canister_with_update_is_uninstalled_due_to_resource_charges(
     // Enable normal message execution.
     env.set_checkpoints_enabled(false);
 
-    env.await_ingress(update.clone(), 100)
+    env.await_ingress(update, 100)
 }
 
 #[test]
@@ -1908,7 +1908,7 @@ fn dts_canister_with_paused_update_is_not_uninstalled_due_to_resource_charges() 
     let result = impl_dts_canister_with_update_is_uninstalled_due_to_resource_charges(false);
 
     // Canister is only uninstalled after paused update completes successfully.
-    assert_matches!(result, Ok(_));
+    assert_matches!(result, Ok(WasmResult::Reply(_)));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


rs/execution_environment/tests/dts.rs
```diff
@@ -1815,108 +1817,98 @@ fn dts_canister_uninstalled_due_to_resource_charges_with_aborted_updrade() {
...
+        Err(err) if err.code() == ErrorCode::CanisterWasmModuleNotFound
+    );
+}
+
+#[test]
+fn dts_canister_with_paused_update_is_not_uninstalled_due_to_resource_charges() {
+    let result = impl_dts_canister_with_update_is_uninstalled_due_to_resource_charges(false);
+
+    // Canister is only uninstalled after paused update completes successfully.
+    assert_matches!(result, Ok(_));
```
eichhorl:
```suggestion
    assert_matches!(result, Ok(WasmResult::Reply(_)));
```

rs/execution_environment/tests/dts.rs
```diff
@@ -1815,108 +1817,98 @@ fn dts_canister_uninstalled_due_to_resource_charges_with_aborted_updrade() {
...
     // resource usage.
     env.advance_time(Duration::from_secs(10_000_000));
-
     env.tick();
 
     // Enable normal message execution.
     env.set_checkpoints_enabled(false);
 
-    let mut errors = 0;
+    env.await_ingress(update.clone(), 100)
```
eichhorl:
```suggestion
    env.await_ingress(update, 100)
```

eichhorl (approved):

rs/execution_environment/tests/dts.rs
alin-at-dfinity: Done.

rs/execution_environment/tests/dts.rs
alin-at-dfinity: Done, thank you.

PR #10306, opened 2026-05-24
feat: [DSM-103] Lower frequency storage charging

alin-at-dfinity:
Instead of checking every canister every round whether 10 seconds have passed since it was last charged, only do it every N rounds; AND on checkpoint rounds, to ensure that canisters with paused DTS executions get charged at least once per checkpoint interval.


rs/execution_environment/src/scheduler.rs
```diff
@@ -815,14 +815,30 @@ impl SchedulerImpl {
...
+        current_round_type: ExecutionRoundType,
     ) {
+        // Because it is relatively expensive to make every canister mutable just to
+        // check whether 10 seconds have passed since it was last charged, only
+        // charge canisters every 50 rounds and/or on checkpoint rounds.
+        //
+        // The latter ensures that (because we skip canisters with paused
+        // executions), every canister is charged at least once per checkpoint
+        // interval.
+        if !current_round.get().is_multiple_of(50)
```
eichhorl: Maybe we could use a constant for this

rs/execution_environment/src/scheduler/tests/charging.rs
```diff
@@ -58,18 +58,19 @@ fn only_charge_for_allocation_after_specified_duration() {
...
         test.canister_state(canister).system_state.balance().get(),
         initial_cycles
     );
 
     // The time of the current batch is now long enough that allocation charging
-    // should be triggered.
+    // should be triggered. Checkpoint round, to force charging for storage.
     test.set_time(initial_time + 2 * time_between_batches);
-    test.execute_round(ExecutionRoundType::OrdinaryRound);
+    test.execute_round(ExecutionRoundType::CheckpointRound);
```
eichhorl, line 69: Do we also have a test for the inverse, i.e. that canisters will _not_ be charged on ordinary rounds?

eichhorl (approved):

rs/execution_environment/src/scheduler.rs
alin-at-dfinity: Done.

rs/execution_environment/src/scheduler/tests/charging.rs
alin-at-dfinity: Added a test checking that we only charge once every `CHARGE_INTERVAL_ROUNDS` (the new constant).

PR #10307, opened 2026-05-24
perf: Only migrate logs once per process

alin-at-dfinity:
The log_memory_store_feature flag is hardcoded into the replica binary, meaning that logs migration is needed at most once per replica process. Since iterating over all `CanisterStates` is relatively expensive, add a transient field to `ReplicatedState` that tracks whether the (potential) migration has already been completed. Skip the `execute_round()` step altogether if the flag was set earlier.

### DIFF
--- a/rs/execution_environment/src/scheduler.rs
+++ b/rs/execution_environment/src/scheduler.rs
@@ -1112,7 +1112,7 @@ impl Scheduler for SchedulerImpl {
         // The goal is to ensure that we can track the performance of `execute_round` and its individual components.
         let root_measurement_scope = MeasurementScope::root(&self.metrics.round);
 
-        {
+        if !state.metadata.logs_migrated {
             let _timer = self
                 .metrics
                 .round_log_memory_store_migration_duration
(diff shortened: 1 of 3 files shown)


schneiderstefan (commented):
The question is if it is worth doing this to optimize temporary code. Does anybody see any risk here that could cause divergence? We are playing fast and loose with how we usually treat determinism. cc @mraszyk

mraszyk:
> The question is if it is worth doing this to optimize temporary code. Does anybody see any risk here that could cause divergence? We are playing fast and loose with how we usually treat determinism. cc @mraszyk

I proposed this since this should be correct, but I don't have a strong opinion on whether it is worth doing.

schneiderstefan (approved):

alin-at-dfinity:
> We are playing fast and loose with how we usually treat determinism.

Logs migration in either direction is very much intended to be idempotent. Which is why we can afford to do it on every single iteration of `execute_round()`. Because the feature flag is baked into the replica binary and cannot be changed at runtime, this also means that as soon as we did it once, we never actually need to do it again for the lifetime of the process. And, if a replica process restarts from a checkpoint created all canisters' logs have been migrated, it is unnecessary but safe to execute the (conditional) logs migration logic again, just as it would be if we were doing it every `execute_round()`.

PR #10310, opened 2026-05-26
feat: IC-1943 Include refund shares in HTTP outcalls payload

eichhorl:
## Background
Currently, the price for an HTTP outcall is calculated and charged upfront. In the future, we want to implement a "pay-as-you-go" pricing model. This model assigns a fraction (budget) of the total attached cycles to each participating replica (per-replica allowance). 

As the request passes through the target server and the transform function, the HTTP adapter of each replica subtracts cycles from its own budget, based on the amount of resources it consumed (download time, downloaded bytes, transform instructions). In the end, the amount of remaining cycles (refund) is passed back to the replica together with the HTTP response.

## Proposed Changes
With this PR, we let the replica include (and sign) this refund as part of the gossiped response share. 

Before, each replica only signed the response metadata which should be equal across all responses in order to reach consensus. Now, the signed messages may be different even across agreeing replicas, as each of them could sign a different refund share. 

Therefore, we change the structure of the aggregated HTTP response proof. Previously, this was just the metadata of the response and a collection of individual signatures over it. Now, it is the metadata and a collection of individual signatures and refund receipts of each replica. In order to verify the proof, the individual messages consisting of both the metadata and the refund share have to be reconstructed, such that the signature can be verified. 

For that reason we additionally switch signature verification of aggregated responses to use the new batch verification API for multiple messages, similar to what was done in https://github.com/dfinity/ic/pull/10345. As a side effect, this comes with some performance improvements for payload validation:
```
                        time:   [146.72 ms 148.46 ms 150.26 ms]
                        change: [-14.887% -13.555% -12.243%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [225.27 ms 227.44 ms 229.63 ms]
                        change: [-17.525% -16.435% -15.417%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [11.124 ms 11.263 ms 11.406 ms]
                        change: [-61.100% -60.382% -59.642%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [116.12 ms 117.52 ms 118.94 ms]
                        change: [-4.0609% -2.4873% -0.9067%] (p = 0.00 < 0.05)
                        Change within noise threshold.
                        time:   [253.98 ms 256.09 ms 258.22 ms]
                        change: [-3.1565% -2.0740% -0.8815%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

Additionally, during payload validation, we verify that none of the refund shares exceed the per-replica allowance. Note that in the current (legacy) pricing model, nothing is refunded yet, so the allowance, and the returned refund is always 0.


rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -575,12 +541,6 @@ impl CanisterHttpPayloadBuilderImpl {
             }
         };
```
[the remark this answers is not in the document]
eichhorl: done

rs/types/types/src/batch/canister_http.rs
```diff
@@ -249,6 +252,22 @@ impl TryFrom<pb::CanisterHttpResponseWithConsensus> for CanisterHttpResponseWith
...
+                signature.payment_receipt,
+                "CanisterHttpResponseSignature::payment_receipt",
+            )?;
+            signatures.insert(
+                signer,
+                CanisterHttpResponseSignature {
+                    payment_receipt,
+                    signature: BasicSigOf::new(BasicSig(signature.signature)),
+                },
+            );
```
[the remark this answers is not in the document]
eichhorl: This seems to be convention across many types, can take a look in a follow up.

rs/types/types/src/canister_http.rs
```diff
@@ -1081,15 +1081,117 @@ impl CountBytes for CanisterHttpResponseMetadata {
...
+}
+
+impl CanisterHttpResponseProof {
+    pub fn registry_version(&self) -> RegistryVersion {
+        self.metadata.registry_version
+    }
+}
+
+/// A signature share of [`CanisterHttpResponseReceiptShare`].
+pub type CanisterHttpResponseShare = BasicSigned<CanisterHttpResponseReceiptShare>;
```
pierugo-dfinity: Just my two cents.

Should `CanisterHttpResponseReceiptShare` have `Share` in its name even though it does not contain anything identifying the replica that issued that share? `Share` is already in `CanisterHttpResponseShare`, which makes sense since it contains the signature. Maybe `CanisterHttpResponseReceiptShare` could be renamed to `CanisterHttpResponseContent`?

This is just a remark mostly for consistency purposes with other artifact names. F.ex. `type NotarizationShare = Signed<NotarizationContent, ...>`, same for finalization shares, CUP shares, RB shares, RT shares, IDKG Complaints, IDKG openings.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -1475,31 +1618,112 @@ pub(crate) fn metadata_to_share_with_signature(
...
+            metadata: metadata.clone(),
+            payment_receipt: receipt_exceeding_allowance(),
+        },
+        node_test_id(signer_node),
+    )
+}
+
+/// Builds an aggregated-proof signature entry whose payment receipt claims a
+/// refund exceeding the default per-replica allowance.
+fn proof_signature_with_excess_refund() -> CanisterHttpResponseSignature {
```
pierugo-dfinity: nit: it looks like this is always called by inserting it into a proof's `signatures` field. You could follow a similar signature as `add_signer_to_proof` and directly mutate a given proof's `signatures` inside this function for consistency.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -2342,8 +2566,8 @@ fn flexible_invalid_content_size_mismatch() {
     setup_test_with_flexible_context(4, callback_id, committee, 1, 4, |payload_builder, _pool| {
         let mut entry = flexible_response(42, 0, b"data");
         let expected_size = entry.response.content.count_bytes() as u32;
-        entry.proof.content.content_size = expected_size.wrapping_add(1);
-        let wrong_size = entry.proof.content.content_size;
+        entry.proof.content.metadata.content_size = expected_size.wrapping_add(1);
+        let wrong_size = entry.proof.content.content_size();
```
pierugo-dfinity: nit: I think it makes sense to have the newly introduced methods to refer to fields but, as a reader, here and below, we now have an additional indirection to understand what they refer to. Said differently, it is now not directly clear anymore that the `wrong_size = entry.proof.content.content_size()` refers to the just-modified `entry.proof.content.metadata.content_size`.

rs/types/types/src/batch/canister_http.rs
```diff
@@ -611,19 +631,26 @@ mod tests {
     /// works correctly, both with and without a full response.
     #[test]
     fn canister_http_response_artifact_conversion() {
+        use ic_types_cycles::Cycles;
```
pierugo-dfinity: I would move that import to the top of the module

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -3389,15 +3458,127 @@ pub mod test {
...
+                    replica_config,
+                    crypto,
+                    state_manager,
+                    registry,
+                    ..
+                } = dependencies(pool_config.clone(), 5);
+                let mut shim_mock = MockNonBlockingChannel::<CanisterHttpRequest>::new();
+                shim_mock
+                    .expect_try_receive()
+                    .return_const(Err(TryReceiveError::Empty));
```
```suggestion
```
Is this needed?

rs/types/types/src/crypto/hash/tests.rs
```diff
@@ -1094,8 +1096,14 @@ mod crypto_hash_stability {
             registry_version: RegistryVersion::from(1),
             replica_version: ReplicaVersion::default(),
         };
+        let receipt_share = CanisterHttpResponseReceiptShare {
+            metadata,
+            payment_receipt: CanisterHttpPaymentReceipt {
+                refund: Cycles::from(42_u64),
```
pierugo-dfinity: nit
```suggestion
                refund: Cycles::new(42),
```

pierugo-dfinity (approved):
Do you think it would be feasible to implement hashes-in-blocks for `CanisterHttp` artifacts in the long term?

rs/types/types/src/canister_http.rs
eichhorl: That's fair, unfortunately `CanisterHttpResponseContent` is already taken by the actual response _content_ (i.e. the bytes that the user requested), whereas here we only sign the metadata (the hash of the response) and the payment receipt. Perhaps we could just drop `Share` and stick with `CanisterHttpResponseReceipt`, WDYT?

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Good idea, done

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Reverted for clarity 👍

rs/types/types/src/batch/canister_http.rs
eichhorl: Good catch

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Nope, thanks

eichhorl:
> Do you think it would be feasible to implement hashes-in-blocks for `CanisterHttp` artifacts in the long term?

Unlike ingress and IDKG the `CanisterHttpPayload` is stored in the payload as plain bytes, which makes this a bit more annoying, but it definitely makes sense.

rs/types/types/src/canister_http.rs
pierugo-dfinity: Ah ok unfortunate. Yes, `CanisterHttpResponseReceipt` works 👍

rs/https_outcalls/consensus/src/payload_builder/tests.rs
pierugo-dfinity: There are a few other places where this is the case. I think we can keep using the methods in production code, but maybe not in tests in places where we mutate those fields

rs/types/types/src/canister_http.rs
eichhorl: Done

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Thanks, I hope I got all of them now

pierugo-dfinity: Nice, thank you!

rs/types/types/src/canister_http.rs
pierugo-dfinity: Nice

PR #10314, opened 2026-05-26
fix(consensus): create CUP shares without waiting for finalized tip to point to CUP height

pierugo-dfinity:


pierugo-dfinity:
Closing in favour of #10347

PR #10318, opened 2026-05-27
chore: remove deprecated ic-cdk imports in ic-icp-archive

gregorydemay:
## Summary

Extends [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the file-level `#![allow(deprecated)]` attribute introduced in #6264.

This PR handles `ic-icp-archive`:

- `ic_cdk::api::call::{arg_data_raw, reply, reply_raw}` → `ic_cdk::api::{msg_arg_data, msg_reply}`.
- `ic_cdk::api::caller` → `ic_cdk::api::msg_caller`.
- `ic_cdk::api::print` → `ic_cdk::api::debug_print`.
- `ic_cdk::api::stable::stable_size` → `ic_cdk::stable::stable_size`.

The few `reply(tuple)` sites are rewritten as `msg_reply(candid::Encode!(...).unwrap())` so the on-wire Candid output is unchanged. No behavior change.

Follows the pattern set by #6755 (`ic-btc-checker`), #6761 (`ic-ckbtc-minter`), #10289, #10290, #10291, and #10317.

[DEFI-2304]: https://dfinity.atlassian.net/browse/DEFI-2304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10289, opened 2026-05-22
chore(defi): remove deprecated ic-cdk imports in ic-bitcoin-canister-mock

gregorydemay:
## Summary

Continues [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the file-level `#![allow(deprecated)]` attributes that were introduced in #6264 when `ic-cdk` was upgraded to 0.18, by migrating the affected files off the deprecated `ic_cdk::api::management_canister::*` surface.

This PR handles `ic-bitcoin-canister-mock`. The only deprecated APIs it referenced were the `BitcoinNetwork` and `SendTransactionRequest` types; both now live at `ic_cdk::bitcoin_canister::{Network, SendTransactionRequest}`. No behavior change.

Follows the same pattern as #6755 (`ic-btc-checker`) and #6761 (`ic-ckbtc-minter`).

[DEFI-2304]: https://dfinity.atlassian.net/browse/DEFI-2304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10320, opened 2026-05-27
chore: remove deprecated ic-cdk imports in icp-ledger

gregorydemay:
## Summary

Extends [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the file-level `#![allow(deprecated)]` attribute introduced in #6264.

This PR handles `icp-ledger`:

- `ic_cdk::api::call::{arg_data_raw, reply, reply_raw}` → `ic_cdk::api::{msg_arg_data, msg_reply}`.
- `ic_cdk::api::caller` → `ic_cdk::api::msg_caller`.
- `ic_cdk::api::print` → `ic_cdk::api::debug_print`.
- `ic_cdk::api::set_certified_data` → `ic_cdk::api::certified_data_set`.
- `ic_cdk::api::stable::StableReader` → `ic_cdk::stable::StableReader`.
- `ic_cdk::api::stable::stable_size` → `ic_cdk::stable::stable_size`.

Imports use `as` aliases so the existing identifier-based call sites (`caller()`, `arg_data_raw()`, `reply_raw(&res)`, `print(...)`) are preserved — that keeps the diff focused on the import line. The single fully qualified `ic_cdk::api::call::reply((result,))` site is rewritten as `reply_raw(Encode!(&result).unwrap())` to preserve the on-wire Candid output. No behavior change.

Follows the pattern set by #6755 (`ic-btc-checker`), #6761 (`ic-ckbtc-minter`), #10289, #10290, #10291, #10317, #10318 and #10319.


rs/ledger_suite/icp/ledger/src/main.rs
```diff
@@ -157,7 +156,7 @@ fn init(
         }
     }
     #[cfg(not(feature = "canbench-rs"))]
-    set_certified_data(
+    certified_data_set(
```
mbjorkqvist, line 159: nit: The line below could be `LEDGER` (without the `&` borrow). Clippy doesn't catch the `needless_borrows_for_generic_args` since it runs with `--all-features`.

mbjorkqvist (approved):

rs/ledger_suite/icp/ledger/src/main.rs
gregorydemay: 🤖 Good catch — fixed in `3d2ba73bb2` (dropped the `&`, matching the other two `certified_data_set` call sites in this file). Resolving.

PR #10322, opened 2026-05-27
chore: remove deprecated ic-cdk imports in ic-icp-test-ledger

gregorydemay:
## Summary

Extends [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the file-level `#![allow(deprecated)]` attribute. For `ic-icp-test-ledger` the attribute was added later by #8490, but it has the same root cause as the rest of the DEFI-2304 surface: the canister was templated from another ledger canister that still used the deprecated `ic_cdk::api::call::{arg_data_raw, reply_raw}` imports.

- `ic_cdk::api::call::{arg_data_raw, reply_raw}` → `ic_cdk::api::{msg_arg_data, msg_reply}` (imported with `as` aliases so call sites stay untouched).

No behavior change.


mbjorkqvist (approved):
Thanks @gregorydemay!

PR #10323, opened 2026-05-27
chore: remove deprecated ic-cdk imports in ic-sender-canister

gregorydemay:
## Summary

Extends [DEFI-2304](https://dfinity.atlassian.net/browse/DEFI-2304) — removing the file-level `#![allow(deprecated)]` attributes introduced in #6264.

This PR handles `ic-sender-canister`:

- `ic_cdk::api::call::call_raw128(to, &method, arg, payment).await` → `ic_cdk::call::Call::unbounded_wait(to, &method).take_raw_args(arg).with_cycles(payment).await` + `Response::into_bytes` to recover the raw reply bytes.
- `ic_cdk::api::call::RejectionCode` → a local `RejectionCode` enum that mirrors the 7 variants previously exposed by the deprecated cdk type. The Candid interface declared in `sender.did` is unchanged.
- The new `CallFailed` error is mapped back to `(RejectionCode, String)` so the public `SendError = (RejectionCode, String)` type alias keeps the same wire shape.

No behavior change.

### DIFF
--- a/rs/rosetta-api/icp/test_utils/sender_canister/src/lib.rs
+++ b/rs/rosetta-api/icp/test_utils/sender_canister/src/lib.rs
@@ -1,6 +1,4 @@
-#![allow(deprecated)]
 use candid::{CandidType, Principal};
-use ic_cdk::api::call::RejectionCode;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


mbjorkqvist (approved):

PR #10329, opened 2026-05-27
fix: Always write state manager metadata together

schneiderstefan:
This commit fixes a potential race condition, where during a checkpoint round we first populated `certifications_metadata`, before (with a separate lock) writing `states_metadata`. This made it possible to observe a state where we have the former but not the latter.

With this change they are written inside the same lock, so that they are observable at the same time.

### DIFF
--- a/rs/state_manager/src/lib.rs
+++ b/rs/state_manager/src/lib.rs
@@ -1399,7 +1399,13 @@ impl StateManagerImpl {
             metrics.clone(),
             malicious_flags.clone(),
         );
-        let (_hash_thread_handle, hash_channel) = spawn_hash_thread(metrics.clone(), log.clone());
+        let persist_metadata_guard = Arc::new(Mutex::new(()));
+        let (_hash_thread_handle, hash_channel) = spawn_hash_thread(
+            metrics.clone(),
+            log.clone(),
+            state_layout.clone(),
+            persist_metadata_guard.clone(),
+        );
 
         let starting_time = Instant::now();
         let loaded_states_metadata =
@@ -1612,8 +1618,6 @@ impl StateManagerImpl {
             tip_height: tip_height_and_state.0,
             tip: Some(tip_height_and_state.1),
         }));
-
-        let persist_metadata_guard = Arc::new(Mutex::new(()));
 
         let deallocator_thread =
             DeallocatorThread::new("StateDeallocator", Duration::from_millis(1));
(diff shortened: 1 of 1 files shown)


rs/state_manager/src/lib.rs
```diff
@@ -3649,6 +3668,8 @@ enum HashRequest {
 fn spawn_hash_thread(
     metrics: StateManagerMetrics,
     log: ReplicaLogger,
+    state_layout: StateLayout,
+    persist_metadata_guard: Arc<Mutex<()>>,
 ) -> (JoinOnDrop<()>, Sender<HashRequest>) {
```
alin-at-dfinity, line 3673: Nit: Google's style guides (which I've found useful and I'm still trying to follow) suggest putting metrics and logs arguments last. And, more generally, ordering arguments in order of their importance / relevance.

pierugo-dfinity:
Is there any (low-effort) regression test we could add?

rs/state_manager/src/lib.rs
```diff
@@ -3719,6 +3740,17 @@ fn spawn_hash_thread(
...
+                            // For checkpointing heights, insert `states_metadata` under the same
+                            // write lock as `certifications_metadata`, so the two maps are always
+                            // consistent. We use `or_insert` so that an existing entry (e.g.
+                            // populated by state sync) is preserved.
+                            if let Some(state_metadata) = state_metadata {
+                                states
+                                    .states_metadata
+                                    .entry(height)
+                                    .or_insert(state_metadata);
+                            }
```
alin-at-dfinity, line 3752: Clueless question: Is the order in which we do these two operations important or is it sufficient for both to be done while holding the lock?

rs/state_manager/src/lib.rs
```diff
@@ -3767,6 +3799,20 @@ fn spawn_hash_thread(
                                     );
                                 }
                             }
+
+                            // For checkpointing heights, persist `states_metadata` to disk while
+                            // still holding (and then releasing) the same write lock used to insert
+                            // the in-memory entry above. This keeps the persisted file in lockstep
+                            // with the in-memory state.
```
alin-at-dfinity, line 3806: Clueless question: `release_lock_and_persist_metadata()` only persists the metadata after having released the lock. At first sight, this does not seem to match the "keeps the persisted file in lockstep with the in-memory state" claim.

rs/state_manager/src/lib.rs
schneiderstefan: Done.

rs/state_manager/src/lib.rs
schneiderstefan: It does not matter. While we hold a write lock, nobody can hold a read lock, and hence cannot observe the inconsistent state.

rs/state_manager/src/lib.rs
schneiderstefan: Well the idea is that whenever we make a change we kick of the persisting, so that at least eventually they are consistent again. There could absolutely be timing issues, but keeping the log during the IO caused big problems in the past (namely that all other threads were blocked waiting for the lock).

schneiderstefan:
> Is there any (low-effort) regression test we could add?

I tried writing a test around the idea of doing c&c, while another thread bombards it with `get_hash_at`. Despite my best efforts I didn't manage to make it work, or more precisely, I didn't manage to make it fail pre-fix. I think to actually trigger the race-condidtion, you need both bad luck and significant load that exceeds what's reasonable in a unit test.

eichhorl:
> Is there any (low-effort) regression test we could add?

Maybe we could add `This should not happen` to the [unallowed log patterns](https://sourcegraph.com/r/github.com/dfinity/ic@13a59c105520b81acce75fe3369fc32830086b34/-/blob/rs/tests/driver/src/driver/group.rs?L747-761), but there seems to be another common occurrence in an NNS canister, see [here](https://kibana.testnet.dfinity.network/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:now-8w,to:now))&_a=(columns:!(MESSAGE,ic_subnet),filters:!(),hideChart:!f,index:testnet-vector-push,interval:auto,query:(language:kuery,query:'%22This%20should%20not%20happen%22'),rowsPerPage:500,sort:!(!(timestamp,desc)))) (which could be excluded).

Edit: But actually panics are already unallowed 🤔 

Edit 2: Ah, maybe these panics never happened on master

alin-at-dfinity (approved):

PR #10331, opened 2026-05-27
feat: Strip timestamps from base images

Bownairo:
From the fallout of #10208

The new podman seems to be able to wrap more timestamps coming from base images (back to epoch), which leads to cache pollution across the upgrade to 26.04. To avoid this, we clear all timestamps before creating cached artifacts, and make some first steps to removing timestamps from the base images entirely.

### DIFF
--- a/toolchains/sysimage/build_container_filesystem_tar.py
+++ b/toolchains/sysimage/build_container_filesystem_tar.py
@@ -78,7 +78,15 @@ def build_container(
         # Override the first FROM statement - grabs it from local cache
         cmd += f" --from {base_image_override.image_tag}"
 
-    # Set timestamp for all files for determinism
+    # Set timestamp for all new files for determinism
+    # NOTE: This does not touch base layers. As base images are currently built
+    # by Docker, not podman, they leak timestamps. (This doesn't really matter,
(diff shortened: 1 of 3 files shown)


basvandijk (approved):

toolchains/sysimage/build_container_filesystem_tar.py
```diff
@@ -78,7 +78,15 @@ def build_container(
         # Override the first FROM statement - grabs it from local cache
         cmd += f" --from {base_image_override.image_tag}"
 
-    # Set timestamp for all files for determinism
+    # Set timestamp for all new files for determinism
+    # NOTE: This does not touch base layers. As base images are currently built
+    # by Docker, not podman, they leak timestamps. (This doesn't really matter,
```
Bownairo, line 83: `container-base-images.yml` does not go through the same build path as `build_container_base_image.py`, so it's not quite as hermetic, and might give different results than local builds. The whole thing isn't reproducible anyway, which is why we use pinned containers, but it might be nice to get them on the same build path.

It's probably just as much work to remove the base images entirely, which would be a better route.

r-birkner (approved):

PR #10335, opened 2026-05-28
refactor: IC-1942 Return payment receipt from HTTP outcalls adapter

eichhorl:
Currently, the price for an HTTP outcall is calculated and charged upfront. In the future, we want to implement a "pay-as-you-go" pricing model. This model assigns a fraction (budget) of the total attached cycles to each participating replica (per-replica allowance). This is captured by the `BudgetTracker`, which is instantiated by the adapter for each new HTTP request that is made.

As the request passes through the target server and the transform function, the adapter subtracts cycles from this budget, based on the amount of resources it consumed (download time, downloaded bytes, transform instructions).

In the end, the amount of remaining cycles (refund) is passed back to the replica together with the HTTP response.

This PR adds the struct representing this payment receipt, and a function of the budget tracker to create the receipt.

In the current (legacy) pricing model, nothing is refunded yet, so the returned refund is always 0. Furthermore, the receipt is still ignored by the replica.

### DIFF
--- a/rs/https_outcalls/client/src/client.rs
+++ b/rs/https_outcalls/client/src/client.rs
@@ -126,7 +126,6 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
         // Spawn an async task that sends the canister http request to the adapter and awaits the response.
         // After receiving the response from the adapter an optional transform is applied by doing an upcall to execution.
         // Once final response is available send the response over to the channel making it available to the client.
-
         self.rt_handle.spawn(async move {
             let CanisterHttpRequest {
                 id: request_id,
--- a/rs/https_outcalls/client/src/client.rs
+++ b/rs/https_outcalls/client/src/client.rs
@@ -186,7 +186,7 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
                     request_headers,
                     request_body,
                     socks_proxy_addrs,
-                    &mut budget,
+                    &mut *budget,
                 )
                 .await?;
 
@@ -207,7 +207,7 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
                 let transformed_payload = match &request_transform {
                     Some(transform) => {
                         let (transform_result, instruction_count) = transform_adapter_response(
-                            &mut budget,
+                            &mut *budget,
                             query_handler,
                             canister_http_payload,
                             request_sender,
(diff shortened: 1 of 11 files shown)


rs/https_outcalls/client/src/client.rs
```diff
@@ -394,6 +309,124 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
...
+/// the number of bytes downloaded as reported by the adapter, and the
+/// elapsed wall-clock duration as measured by the client.
+async fn execute_http_request(
+    adapter_client: &mut HttpsOutcallsServiceClient<Channel>,
+    url: String,
+    http_method: CanisterHttpMethod,
+    headers: Vec<CanisterHttpHeader>,
+    body: Option<Vec<u8>>,
+    socks_proxy_addrs: Vec<String>,
+    budget: &mut Box<dyn BudgetTracker>,
```
pierugo-dfinity:
```suggestion
    budget: &mut dyn BudgetTracker,
```
I think we can simplify this

rs/https_outcalls/client/src/client.rs
```diff
@@ -394,6 +309,124 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
...
+) -> Result<(HttpsOutcallResponse, NumBytes, Duration), (RejectCode, String)> {
```
pierugo-dfinity: Would it make sense to return a `CanisterHttpReject` directly here?

```suggestion
) -> Result<(HttpsOutcallResponse, NumBytes, Duration), CanisterHttpReject> {
```
Same in `validate_response`

rs/https_outcalls/client/src/client.rs
```diff
@@ -394,6 +309,124 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
...
+            })
+            .collect(),
+        body: body.unwrap_or_default(),
+        socks_proxy_addrs,
+    };
+
+    let (grpc_result, elapsed) =
+        timeout_with_capped_metric(max_response_time, adapter_client.https_outcall(proto_req))
+            .await
+            .map_err(|_| (RejectCode::SysTransient, "Deadline Exceeded".to_string()))?;
```
pierugo-dfinity: Did this error become `SysTransient` on purpose?

pierugo-dfinity (approved):

rs/https_outcalls/client/src/client.rs
eichhorl: Nice, thanks: https://github.com/dfinity/ic/pull/10335/commits/8c72ecf7e7e20cd7d057ec58683f20147c70d9d3

rs/https_outcalls/client/src/client.rs
eichhorl: The footprint was a bit larger than expected, but I think it makes sense: https://github.com/dfinity/ic/pull/10335/commits/3491506e32a1a636da59299ae8680634f8ce4c6a

rs/https_outcalls/client/src/client.rs
eichhorl: I think it makes sense for the http timeout to be transient

PR #10339, opened 2026-05-28
feat(crypto): Introduce batched signature verification with multiple messages

eichhorl:
This PR adds a new function `BasicSigVerifier::verify_basic_sig_batch_multi_msg` that batch-verifies basic signatures over potentially different messages. This complements the existing `verify_basic_sig_batch` (which requires all signatures to be over the same message). 

Note that the existing function is a special case, and in general it is not necessary for the messages to be equal in order to benefit from batching. Verifying batches with different messages is already supported and implemented by the underlying crypto primitive. 

In contrast to verifying all signatures one-by-one, batching N signature verifications replaces N independent elliptic-curve operations with one combined operation whose cost grows sublinearly in N, so the per-signature cost shrinks with larger batches.

## Motivation
The motivation for this is mainly HTTP outcalls: In the current happy case (fully replicated outcall, where all replicas derive the same HTTP response), the message signed by each replica is indeed the same. With the new pricing however, each replica will additionally sign the amount of cycles they want to return to the caller, which may be a different value for each of them.

Additionally, there are cases where the signed HTTP responses aren't equal already today, i.e. divergence responses and flexible outcalls responses. Signatures on these responses are currently verified individually, and could similarly be optimized by using the new function.

## Testing
We add the same unit test coverage that already exists for `verify_basic_sig_batch`.

### DIFF
--- a/rs/crypto/src/sign/basic_sig.rs
+++ b/rs/crypto/src/sign/basic_sig.rs
@@ -59,13 +59,41 @@ impl BasicSigVerifierInternal {
                 message: "Empty BasicSignatureBatch. At least one signature should be included in the batch.".to_string(),
             });
         };
+        let inputs: Vec<_> = signatures
+            .signatures_map
+            .iter()
+            .map(|(signer, signature)| (*signer, signature, message))
+            .collect();
+        Self::verify_basic_sigs_batch(csprng, registry, &inputs, registry_version)
+    }
 
-        let message = message.as_signed_bytes();
-        let mut msgs = Vec::with_capacity(signatures.signatures_map.len());
-        let mut sigs = Vec::with_capacity(signatures.signatures_map.len());
-        let mut keys = Vec::with_capacity(signatures.signatures_map.len());
+    /// Verifies a batch of basic signatures on potentially different messages.
+    ///
+    /// Unlike `verify_basic_sig_batch`, the entries in `inputs` may have
+    /// distinct messages, and the same `NodeId` may appear more than once.
+    pub fn verify_basic_sigs_batch<H: Signable, R: CryptoComponentRng>(
+        csprng: &CspRwLock<R>,
+        registry: &dyn RegistryClient,
+        inputs: &[(NodeId, &BasicSigOf<H>, &H)],
+        registry_version: RegistryVersion,
+    ) -> CryptoResult<()> {
+        if inputs.is_empty() {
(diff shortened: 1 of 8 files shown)


rs/crypto/src/sign/basic_sig.rs
```diff
@@ -59,15 +59,67 @@ impl BasicSigVerifierInternal {
...
+                message:
+                    "Empty signature batch. At least one signature should be included in the batch."
+                        .to_string(),
+            });
+        };
+        // Serialize each message once and keep the owned buffers alive so we
+        // can borrow them into the helper's input iterator.
+        let entries: Vec<(NodeId, &BasicSigOf<H>, Vec<u8>)> = inputs
+            .iter()
+            .map(|(signer, signature, message)| (*signer, *signature, message.as_signed_bytes()))
```
alin-at-dfinity, line 97: Nothing to do with your change, but cloning a message just to prepend a domain prefix looks wasteful. All we do with it afterwards is feed it to a hasher. We could just as well feed the hasher first the domain, then the message, from two separate slices.

Or am I missing something?

alin-at-dfinity (approved):

rs/crypto/src/sign/basic_sig.rs
eichhorl: This is a good point. Unfortunately the hasher isn't the only consumer of the `as_signed_bytes()` output. Some consumers require the data to be concatenated, and some others require it to be sent to the vault process, for which it needs to be owned. So changing this (only where possible) is a bit more work, and luckily most of the data that we sign is quite small, so I wouldn't prioritize it for now.

However, this allowed me find a different improvement which should be an easier win:

Currently there are two kinds of messages that we sign:
1. Structured data which we serialize on the fly (e.g. consensus artifacts via serde_cbor, or protobuf).
2. Data that we already hold as raw bytes (e.g. message IDs).

In both cases we currently produce a separate owned vec (a fresh serialization for 1. and a clone for 2.), which is then appended onto the domain prefix. 

Instead, we should be able to replace `fn as_signed_bytes_without_domain_separator(&self) -> Vec<u8>` with `fn write_signed_bytes_without_domain_separator(&self, bytes: &mut Vec<u8>)` which appends/serializes the message bytes directly into the buffer that already holds the domain separator.

I'll take a look at that in a different PR.

PR #10342, opened 2026-05-28
feat: Add tee-recovery to HostOS

frankdavid:
TEE recovery builds on the foundation that we created for NNS recoveries. There are a few differences:
- No recovery packages are used, only an alternative GuestOS image is written.
- The GuestOS's var partition is not wiped.
- The same GuestOS boot alternative must be used where a (broken) GuestOS is deployed that has access to the encrypted partitions.

This change adds a new `tee-recovery` command to the limited console which does not require the recovery package hash. Furthermore, a new field is added for TEE recovery which allows selecting the target boot alternative. This field's default value is current for TEE recoveries and opposite for NNS recoveries.

image

### DIFF
--- a/ic-os/components/hostos/misc/limited-console
+++ b/ic-os/components/hostos/misc/limited-console
@@ -24,10 +24,47 @@ show_menu() {
     echo "  routes                  - Show routing table"
     echo "  guestos-alternative     - Show or swap the GuestOS boot alternative (A/B)"
     echo "  manual-recovery         - Manual node recovery (ONLY FOR EMERGENCY USE)"
+    echo "  tee-recovery            - Install alternative GuestOS in TEE mode"
     echo "  rbash-console           - Drop into restricted bash console"
     echo "  clear                   - Clear the console"
     echo "  exit                    - Exit console"
     echo ""
+}
+
+run_manual_recovery() {
+    local recovery_mode="${1:-nns}"
+
+    echo "=== Manual Node Recovery ==="
+    echo "This will perform a manual node recovery. Do not attempt this unless you are certain it is appropriate."
+    echo "Manual node recovery should only occur in the event of a critical IC failure."
+    echo "In such cases, the process will be coordinated by reputable IC community leaders"
+    echo "and actively discussed publicly with node providers."
+    echo ""
+    echo "Are you sure you want to proceed? (y/n)"
+    read -r confirm
+    if [ "$confirm" == "y" ]; then
+        # Launch the recovery TUI tool via hostos_tool
+
+        # Fix for serial/remote consoles where window size may be missing or tiny
+        if ! size=$(stty size 2>/dev/null); then
+            echo "Error: could not determine terminal size. The manual-recovery TUI requires a TTY."
+            echo "Please see \"Manual Recovery Fallback\" in the Manual Node Recovery Guide for further direction."
+            exit 1
+        fi
+
+        read -r rows cols <<< "$size"
+        # Treat anything smaller than the Manual Recovery TUI minimum as unusable
+        min_rows=15
+        min_cols=10
+        if [ "${rows:-0}" -lt "$min_rows" ] || [ "${cols:-0}" -lt "$min_cols" ]; then
+            echo "Terminal size (${rows}x${cols}) is too small or invalid, setting to 24x80"
+            stty rows 24 cols 80
+            sleep 5 # Give the user a chance to see the message
+        fi
+
+        cmd=(/opt/ic/bin/hostos_tool manual-recovery "$recovery_mode")
+        "${cmd[@]}"
+    fi
 }
 
 # Command dispatcher
@@ -136,36 +173,10 @@ execute_command() {
             esac
             ;;
         "manual-recovery")
-            echo "=== Manual Node Recovery ==="
-            echo "This will perform a manual node recovery. Do not attempt this unless you are certain it is appropriate."
-            echo "Manual node recovery should only occur in the event of a critical IC failure."
-            echo "In such cases, the process will be coordinated by reputable IC community leaders"
-            echo "and actively discussed publicly with node providers."
-            echo ""
-            echo "Are you sure you want to proceed? (y/n)"
-            read -r confirm
-            if [ "$confirm" == "y" ]; then
-                # Launch the NNS recovery TUI tool via hostos_tool
-
-                # Fix for serial/remote consoles where window size may be missing or tiny
-                if ! size=$(stty size 2>/dev/null); then
-                    echo "Error: could not determine terminal size. The manual-recovery TUI requires a TTY."
-                    echo "Please see \"Manual Recovery Fallback\" in the Manual Node Recovery Guide for further direction."
-                    exit 1
-                fi
-
-                read -r rows cols <<< "$size"
-                # Treat anything smaller than the Manual Recovery TUI minimum as unusable
-                min_rows=15
-                min_cols=10
-                if [ "${rows:-0}" -lt "$min_rows" ] || [ "${cols:-0}" -lt "$min_cols" ]; then
-                    echo "Terminal size (${rows}x${cols}) is too small or invalid, setting to 24x80"
-                    stty rows 24 cols 80
-                    sleep 5 # Give the user a chance to see the message
-                fi
-
-                /opt/ic/bin/hostos_tool manual-recovery
-            fi
+            run_manual_recovery nns
+            ;;
+        "tee-recovery")
+            run_manual_recovery tee
             ;;
         "rbash-console")
             echo "=== Dropping into Restricted Bash Console ==="
--- a/ic-os/components/hostos/misc/limited-console
+++ b/ic-os/components/hostos/misc/limited-console
@@ -23,8 +23,8 @@ show_menu() {
     echo "  ping-gateway            - Test connectivity to IPv6 gateway"
     echo "  routes                  - Show routing table"
     echo "  guestos-alternative     - Show or swap the GuestOS boot alternative (A/B)"
-    echo "  manual-recovery         - Manual node recovery (ONLY FOR EMERGENCY USE)"
-    echo "  tee-recovery            - Install alternative GuestOS in TEE mode"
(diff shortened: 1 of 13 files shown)


alin-at-dfinity (approved):

rs/ic_os/os_tools/manual_guestos_recovery/src/lib.rs
```diff
@@ -891,26 +1057,41 @@ mod tests {
...
         FailureState {
-            params: RecoveryParams::default(),
+            params: default_recovery_params(),
             logs: vec!["test log".to_string()],
             exit_status: status,
             error_messages: vec!["test error".to_string()],
         }
     }
 
+    fn default_recovery_params() -> RecoveryParams {
```
Bownairo: Why move this out of `Default`?

Bownairo (approved):
I very quickly looked through the UI code, but assuming the UI code ends up with what's in the screenshot, LGTM.

ic-os/components/hostos/misc/limited-console
```diff
@@ -24,12 +24,49 @@ show_menu() {
     echo "  routes                  - Show routing table"
     echo "  guestos-alternative     - Show or swap the GuestOS boot alternative (A/B)"
     echo "  manual-recovery         - Manual node recovery (ONLY FOR EMERGENCY USE)"
```
pierugo-dfinity: Should we rename this to `nns-recovery`?

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-launcher.sh
```diff
@@ -20,17 +22,35 @@ validate_argument() {
...
+        mode=*)
+            mode="${arg#*=}"
+            ;;
+        target-boot-alternative=*)
+            has_target_boot_alternative=true
+            ;;
+    esac
 done
 
+if [[ ("$mode" == "prep" || "$mode" == "run") && "$has_target_boot_alternative" != true ]]; then
```
pierugo-dfinity: Isn't that already checked in `guestos-recovery-upgrader.sh`? If we also want to do validation at this stage, should we then also check that a `version` is given in those modes for completeness?

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
```diff
@@ -414,39 +492,65 @@ main() {
...
 
     prepare_guestos_upgrade
 
     extract_upgrade "$STAGE_DIR" "$extract_dir"
 
-    log_message "Stopping guestos.service for manual upgrade"
-    systemctl stop guestos.service
-    log_message "GuestOS service stopped"
+    if ! recovery_hash_enabled; then
+        stop_guestos_service "Stopping guestos.service before overwriting the currently running GuestOS alternative"
```
pierugo-dfinity: Technically, wouldn't this log be misleading if the target boot alternative was chosen to not be the currently running alternative but the other one?

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
```diff
@@ -370,23 +431,40 @@ guestos_upgrade_cleanup() {
...
+
+    if [ "$RESTART_GUESTOS_SERVICE_ON_EXIT" = "true" ] && [ "$INSTALL_STARTED" != "true" ]; then
+        log_message "Restarting guestos.service during cleanup after early exit"
+        if systemctl start guestos.service; then
+            log_message "GuestOS service restarted during cleanup"
+            GUESTOS_SERVICE_STOPPED=false
+            RESTART_GUESTOS_SERVICE_ON_EXIT=false
+        else
+            log_message "WARNING: Failed to restart guestos.service during cleanup"
+        fi
```
pierugo-dfinity: nit: would it make sense to re-use function `restart_guestos_service` here?

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
```diff
@@ -73,6 +88,28 @@ log_message() {
...
+
+    log_message "$reason"
+    systemctl stop guestos.service
+    GUESTOS_SERVICE_STOPPED=true
+    RESTART_GUESTOS_SERVICE_ON_EXIT=true
+    log_message "GuestOS service stopped"
+}
+
+restart_guestos_service() {
+    local reason="$1"
```
pierugo-dfinity: Similarly, do we want to check that `$GUESTOS_SERVICE_STOPPED` is `true` before continuing? Otherwise this would be a no-op.
Then we could argue that it's not necessary precisely because it is a no-op. But in that case, then why check that `GUESTOS_SERIVCE_STOPPED` is `false` in `stop_guestos_service`? Could we remove the check there in that case?
I am not entirely sure why we need that variable in the first place actually.

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
```diff
@@ -73,6 +88,28 @@ log_message() {
...
+    fi
+
+    log_message "$reason"
+    systemctl stop guestos.service
+    GUESTOS_SERVICE_STOPPED=true
+    RESTART_GUESTOS_SERVICE_ON_EXIT=true
+    log_message "GuestOS service stopped"
+}
+
+restart_guestos_service() {
```
pierugo-dfinity: Would it be more correct to name this function `start_guestos_service`?

rs/ic_os/os_tools/manual_guestos_recovery/src/lib.rs
```diff
@@ -1,11 +1,13 @@
-//! Interactive TUI application for manual GuestOS recovery. Prompts the operator
-//! for a recovery version and hash prefix, downloads and verifies recovery
-//! artifacts, and installs the recovery GuestOS image.
+//! Interactive TUI application for manual GuestOS recovery.
+//! NNS mode prompts for a recovery version, recovery hash prefix, and target boot alternative.
```
pierugo-dfinity:
```suggestion
//! NNS mode prompts for a recovery version and recovery hash prefix.
```

rs/ic_os/os_tools/manual_guestos_recovery/src/lib.rs
```diff
@@ -1211,9 +1489,35 @@ mod tests {
                 ));
             }
 
+            #[test]
+            fn tab_navigates_through_tee_mode_without_recovery_hash_field() {
+                let mut app = create_test_app(RecoveryMode::Tee);
+
```
pierugo-dfinity:
```suggestion
                assert!(matches!(
                    app.get_state(),
                    Some(AppState::Input(s)) if s.current_field() == Field::Version
                ));
```

rs/ic_os/os_tools/manual_guestos_recovery/src/lib.rs
```diff
@@ -1211,9 +1489,35 @@ mod tests {
...
+                    Some(AppState::Input(s)) if s.current_field() == Field::CheckArtifactsButton
+                ));
+
+                app.handle_event(Event::Key(key_event(KeyCode::Tab)))
+                    .unwrap();
+                assert!(matches!(
+                    app.get_state(),
+                    Some(AppState::Input(s)) if s.current_field() == Field::ExitButton
+                ));
+            }
```
pierugo-dfinity, line 1525:
```suggestion
                app.handle_event(Event::Key(key_event(KeyCode::Tab)))
                    .unwrap();
                assert!(matches!(
                    app.get_state(),
                    Some(AppState::Input(s)) if s.current_field() == Field::Version
                ));
            }
```

rs/ic_os/os_tools/manual_guestos_recovery/src/recovery_utils.rs
```diff
@@ -33,33 +34,99 @@ pub fn build_recovery_upgrader_command(mode: &str, args: &[String]) -> RecoveryU
...
 
+fn maybe_add_wipe_var_partition(args: &mut Vec<String>, wipe_var_partition: bool) {
+    if wipe_var_partition {
+        args.push("wipe-var-partition".to_string());
+    }
+}
+
 pub fn build_recovery_upgrader_prep_command(
     version: &str,
+    target_boot_alternative: Option<BootAlternative>,
```
pierugo-dfinity:
```suggestion
    target_boot_alternative: BootAlternative,
```
It looks like all callers use `Some(...)`?

rs/ic_os/os_tools/manual_guestos_recovery/src/lib.rs
frankdavid: According to the docs, `Default values are often some kind of initial value, identity value, or anything else that may make sense as a default.`

In the case of `RecoveryParams`, I don't think there is a natural default value that applies to all use-cases so I'd rather not implement `Default`. It's still useful in tests, so I added a helper function for tests where we can use some default value without having to be 'scientifically correct'.

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
frankdavid: Simplified it

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
frankdavid: Yeah, the check is not really necessary, removed.

ic-os/components/misc/guestos-recovery/guestos-recovery-upgrader/guestos-recovery-upgrader.sh
frankdavid: I had to change the logic of start_guestos_service a little bit, now we can reuse it.

PR #10344, opened 2026-05-29
fix: II init args in PocketIC

mraszyk:


rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -2180,6 +2201,8 @@ impl PocketIcSubnets {
                 backend_canister_id: Some(IDENTITY_CANISTER_ID.get().0),
                 backend_origin: None,
                 sso_discoverable_domains: None,
+                dnssec_config: None, // DIFFERENT FROM ICP MAINNET
+                doh_config: None,    // DIFFERENT FROM ICP MAINNET
```
mraszyk, line 2205: @aterga Is it okay to use `None` for both fields for local development in PocketIC?

aterga: Good question. 

Short-term, it should be safe to omit these configs, but users won't be able to test email recovery flows locally. 

Conversely, testing email recovery flows would require a few other components (e.g., mock SMTP Gateway; DNSSEC or DoH roots of trust) which are missing in PocketIc, so it's not just a matter of specifying these configs.

mraszyk: > users won't be able to test email recovery flows locally

I suppose that's fine or do you see a use case for that?

aterga: That's fine for now.

Note that the SMTP Gateway is a feature that's going live next week (it will work for all canisters, not just II), we should discuss how to support it in PocketIc.

alin-at-dfinity (approved):

PR #10345, opened 2026-05-29
perf: Verify diverging and flexible HTTPS outcalls responses in batches

eichhorl:
https://github.com/dfinity/ic/pull/10339 introduced a function to verify batches of signatures over multiple messages, which has better performance than verifying signatures one-by-one.

This PR switches the verification of HTTP outcall's divergence and flexible responses over to this new function. Previously, signatures over these responses were validated individually. This is done by collecting all signatures in a local vector, as the payload validation progresses. The signatures are then verified together as a single batch at the end of the payload validation function.

Note that for now, (non-)replicated outcalls are still verified separately via batch verification of signatures _over the same message_. Once the new pricing implementation progresses, these messages will no longer be equal, at which point they will also have to be verified using the new API.

Benchmark changes:
```
                        time:   [170.05 ms 171.74 ms 173.48 ms]
                        change: [-42.232% -41.520% -40.842%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [269.79 ms 272.17 ms 274.62 ms]
                        change: [-2.4546% -1.3239% -0.0543%] (p = 0.04 < 0.05)
                        Change within noise threshold.
                        time:   [28.067 ms 28.429 ms 28.801 ms]
                        change: [-0.0407% +1.7347% +3.4815%] (p = 0.05 > 0.05)
                        No change in performance detected.
                        time:   [119.18 ms 120.51 ms 121.91 ms]
                        change: [-62.752% -62.200% -61.596%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [259.30 ms 261.52 ms 263.82 ms]
                        change: [-54.204% -53.664% -53.124%] (p = 0.00 < 0.05)
                        Performance has improved.
```
As expected, the results for (non-)replicated outcalls are still the same, while verification times of other outcalls responses was roughly cut in half.


rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -174,13 +178,34 @@ pub(crate) fn validate_response_share(
...
+);
+
+/// Maps response shares to the `(signer, signature, message)` inputs consumed by
+/// [`BasicSigVerifier::verify_basic_sig_batch_multi_msg`].
+pub(crate) fn response_share_sig_inputs<'a, I>(
+    shares: I,
+) -> impl Iterator<Item = ResponseShareSigInput<'a>>
+where
+    I: IntoIterator<Item = &'a CanisterHttpResponseShare>,
+{
```
pierugo-dfinity, line 199: Any particular reason why mapping an iterator to another and not for example implementing `From<&'a CanisterHttpResponseShare> for ResponseShareSigInput<'a>` to stay more generic?

pierugo-dfinity (approved):
Very nice! Are there other artifacts that could benefit from this?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: We would have to turn `ResponseShareSigInput` into a real struct instead of an alias in order to implement `From`, but that's not what the crypto API expects. 

Mapping from `IntoIterator` to `Iterator<ResponseShareSigInput>` is nice because we can call it as `response_share_sig_inputs(&response.shares)` (which AI [missed](https://github.com/dfinity/ic/pull/10345/commits/a5700828d5f45d39f167dfea92fd057f1efbf48d)) or `response_share_sig_inputs(reject_responses.iter().map(|r| &r.proof))`, and so on.

eichhorl:
> Very nice! Are there other artifacts that could benefit from this?

The biggest one is probably verification of support shares in IDKG, but unlike here, that isn't really the bottleneck anymore right now. One disadvantage is that if the batch verification fails, we don't know which of the shares are actually invalid, so we would have to validate them again one-by-one to find out. Definitely something to keep in mind and investigate further though

PR #10346, opened 2026-05-29
fix: deflake //rs/dogecoin/ckdoge/minter:integration_tests

gregorydemay:
## Purpose

The ckDOGE withdrawal integration tests (`should_resubmit_transaction` and `should_resubmit_transaction_when_many_utxos`) are flaky: they intermittently fail with a withdrawal-fee mismatch between the `estimate_withdrawal_fee` query and the fee in the `SentDogeTransaction` event (e.g. `dogecoin_fee: 1` vs `376000000`).

## Root cause

`estimate_withdrawal_fee` reflects the minter's last refreshed median fee percentiles. That value starts at the initialization default and is only updated once the minter's periodic refresh task runs successfully after the dogecoin canister is synced. The withdrawal flow captures the estimate before the withdrawal's own submission refreshes the median, so on the first withdrawal of a setup the estimate races the refresh — passing when a refresh already landed, failing otherwise.

## Fix

Add a mandatory `minter_await_fee_refresh` step at the start of the withdrawal flow that drives the fee-percentile refresh and waits until the estimate stabilizes, before the withdrawal request is submitted (kept before submission so it cannot trigger an early transaction). The flow's type-state makes the step impossible to skip for any caller reaching `expect_withdrawal_request_accepted`.

Successfully tried out 100 times with
```
bazel test //rs/dogecoin/ckdoge/minter:integration_tests --runs_per_test=100
```

Supersedes #10336.


rs/dogecoin/ckdoge/test_utils/src/flow/withdrawal.rs
```diff
@@ -38,6 +39,38 @@ impl<S> WithdrawalFlowStart<S> {
...
+            let current = minter.estimate_withdrawal_fee(withdrawal_amount);
+            if current == previous {
+                return WithdrawalFlowApproval { setup: self.setup };
+            }
+            previous = current;
+        }
+        panic!(
+            "BUG: fee estimate did not stabilize after {max_refreshes} refreshes; last estimate: {previous:?}"
+        );
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch — fixed in f0000cd89c. Comparing fee estimate *values* can't tell an already-fresh estimate apart from one whose refresh hasn't succeeded yet, so a stale estimate could slip through.

The refresh helper (now `MinterCanister::await_fee_refresh`) no longer compares fee values. It records the number of successful `[estimate_fee_per_vbyte]: update median fee per vbyte` log entries, advances time past the refresh interval, then ticks until that count increases — i.e. until there is concrete evidence a `RefreshFeePercentiles` task succeeded. If none occurs within the tick budget it panics rather than proceeding with stale fees.

rs/dogecoin/ckdoge/test_utils/src/lib.rs
```diff
@@ -186,6 +188,18 @@ impl Setup {
...
 
+        // After minter install, setup_tasks() schedules ProcessLogic and RefreshFeePercentiles at
+        // T0. If left unfired, RefreshFeePercentiles would fire during the first DogecoinSyncGuard
+        // tick in mine_blocks, sending a fee-percentile XNet call to the dogecoin canister while it
+        // is syncing new blocks. This puts the canister in a "not fully synced" state that prevents
+        // DogecoinSyncGuard from observing the target block height, causing a timeout.
+        // Tick 1 fires ProcessLogic; tick 2 fires RefreshFeePercentiles and starts the XNet call;
+        // the remaining ticks let the XNet round-trip complete. After this the timer is at T0+360s,
+        // far beyond any DogecoinSyncGuard window.
+        for _ in 0..20 {
```
mbjorkqvist: 🤖 Minor: this 8-line comment + `for _ in 0..20 { env.tick() }` loop is duplicated almost verbatim in `minter.rs` `upgrade()` (post-upgrade drain). Consider extracting a shared helper (e.g. `drain_startup_tasks(env)`) so the explanation lives in one place and the magic `20` can't drift between the two sites.

rs/dogecoin/ckdoge/test_utils/src/minter.rs
```diff
@@ -351,8 +373,16 @@ impl MinterCanister {
...
-        // run immediate tasks after upgrade, like refreshing fee percentiles.
-        self.env.tick();
+        // post_upgrade calls setup_tasks() which resets both ProcessLogic and RefreshFeePercentiles
+        // to T0. Tick 1 fires ProcessLogic; tick 2 fires RefreshFeePercentiles and starts its XNet
+        // call to the dogecoin canister; the remaining ticks let the round-trip complete so the
+        // timer is pushed to T0+360s before control returns. Without this, RefreshFeePercentiles
+        // would fire during the first DogecoinSyncGuard tick in a subsequent mine_blocks call,
+        // sending the fee-percentile request while the dogecoin canister is syncing blocks, which
+        // puts it in a "not fully synced" state that causes DogecoinSyncGuard to time out.
+        for _ in 0..20 {
```
mbjorkqvist: 🤖 Minor: same drain loop + comment block as in `lib.rs` after minter install — candidate for a shared helper to avoid the two copies drifting.

rs/dogecoin/ckdoge/test_utils/src/lib.rs
```diff
@@ -48,6 +48,8 @@ pub const LEDGER_TRANSFER_FEE: u64 = DOGE / 100;
 const MAX_TIME_IN_QUEUE: Duration = Duration::from_secs(10);
 pub const MIN_CONFIRMATIONS: u32 = 60;
 pub const BLOCK_TIME: Duration = Duration::from_secs(60);
+/// Must be at least the minter's `refresh_fee_percentiles_frequency`.
+pub const FEE_PERCENTILES_REFRESH_INTERVAL: Duration = Duration::from_secs(360);
```
mbjorkqvist, line 56: 🤖 Minor: `360` duplicates the minter's `refresh_fee_percentiles_frequency` (`SIX_MINUTES` in `minter/src/lib.rs:68`) with no compile-time link. The doc comment notes the coupling, but if the minter ever changes that frequency this constant silently becomes too small and the flake returns. Worth a stronger note, or ideally a shared source of truth.

rs/dogecoin/ckdoge/test_utils/src/minter.rs
```diff
@@ -163,6 +163,28 @@ impl MinterCanister {
         Decode!(&call_result, Result<WithdrawalFee, EstimateWithdrawalFeeError>).unwrap()
     }
 
+    pub fn await_fee_refresh(&self) {
```
mbjorkqvist, line 168: 🤖 Minor: if `get_current_fee_percentiles` succeeds but `estimate_median_fee` returns `None` (or the XNet call errors), the `update median fee per vbyte` log line is never emitted, so this exhausts all 100 ticks and panics with `did not refresh fee percentiles` — slightly misleading, since the task *did* run, it just couldn't compute a median. Not blocking (the canister is expected synced here, and the `dbg!(get_logs())` dump mitigates it); just noting the message could mislead.

mbjorkqvist (approved):
Thanks @gregorydemay! Claude found some nits, but overall LGTM!

rs/dogecoin/ckdoge/test_utils/src/lib.rs
gregorydemay: 🤖 Done in 59fc26b60a. Extracted a shared `drain_startup_tasks(env)` helper (in `lib.rs`) that holds the explanation in one place and puts the magic `20` behind a named `STARTUP_DRAIN_TICKS` const. Both `Setup::new` and `upgrade()` now call it, so the two copies can no longer drift.

rs/dogecoin/ckdoge/test_utils/src/minter.rs
gregorydemay: 🤖 Done in 59fc26b60a — `upgrade()` now calls the shared `drain_startup_tasks(&self.env)` helper instead of its own copy of the loop and comment.

rs/dogecoin/ckdoge/test_utils/src/lib.rs
gregorydemay: 🤖 Done in 59fc26b60a. `SIX_MINUTES` is a method-local `const` inside the trait impl, so there is no clean way to share a single source of truth without modifying productive code. Instead I strengthened the doc comment to name `refresh_fee_percentiles_frequency` and the file it lives in, and to spell out that the flake returns if the minter ever lowers that frequency below this interval.

rs/dogecoin/ckdoge/test_utils/src/minter.rs
gregorydemay: 🤖 Done in 59fc26b60a. Reworded the panic to "did not observe a successful fee-percentile refresh within {max_ticks} ticks (the RefreshFeePercentiles task may have run but failed to compute a median fee)", so it no longer claims the task did not run.

PR #10348, opened 2026-05-29
chore: Exclude debug logs in guest_disk

frankdavid:
In order to get rid of the unnecessary log lines regarding mutex locks/unlocks.


Bownairo (approved):
Goodbye scary red text 👋

PR #10351, opened 2026-05-29
perf: Directly serialize signed bytes into `Vec` holding domain separator

eichhorl:
Currently there are two kinds of messages that are signed by the replica:
1. Structured data which is serialized on the fly (e.g. consensus artifacts via serde_cbor, or protobuf).
2. Data that is already present as raw bytes (e.g. message IDs).

In both cases, the function `as_signed_bytes_without_domain_separator` currently produces a newly owned vec (a fresh serialization for 1. and a clone for 2.), which is then appended onto a different vector holding the domain prefix. 

Instead, this PR replaces `fn as_signed_bytes_without_domain_separator(&self) -> Vec<u8>` with `fn write_signed_bytes_without_domain_separator(&self, bytes: &mut Vec<u8>)` which appends/serializes the message bytes directly into the buffer that already holds the domain separator. This saves one clone/allocation of the data to be signed/verified in both cases.

Additionally we remove a few other unnecessary allocations.

### DIFF
--- a/rs/types/types/src/messages/webauthn.rs
+++ b/rs/types/types/src/messages/webauthn.rs
@@ -76,8 +76,8 @@ pub struct WebAuthnEnvelope {
 }
 
 impl WebAuthnEnvelope {
-    pub fn challenge(&self) -> Vec<u8> {
-        self.challenge.clone()
+    pub fn challenge(&self) -> &[u8] {
+        &self.challenge
     }
 }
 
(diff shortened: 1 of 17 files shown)


rs/types/types/src/messages/webauthn.rs
```diff
@@ -97,7 +97,7 @@ impl TryFrom<&WebAuthnSignature> for WebAuthnEnvelope {
         };
 
         let mut signed_bytes = signature.authenticator_data.0.clone();
-        signed_bytes.append(&mut Sha256::hash(&signature.client_data_json.0.clone()[..]).to_vec());
+        signed_bytes.extend_from_slice(&Sha256::hash(&signature.client_data_json.0));
```
alin-at-dfinity, line 100: 😮

alin-at-dfinity (approved):
Very nice, thank you!

PR #10356, opened 2026-06-01
test: Deflake `//rs/tests/networking/firewall:firewall_correctness_test`

eichhorl:
It always fails with:
```
Task firewall_correctness_test        FAILED  in 185.20s
     called `Result::unwrap()` on an `Err` value: Func="check if latest registry version >= 2 [rs/tests/driver/src/driver/test_env_api.rs:682]" timed out after 180.08830969s on attempt 26.
      Last error: Registry Transport Error

     Caused by:
         Registry Canister Error. Msg: An unknown error occurred in the registry canister: Failed to query get_certified_changes_since on canister rwlgt-iiaaa-aaaaa-aaaaa-cai: HttpClient: Request timed out for http://[2602:fb2b:110:10:502c:96ff:fec1:7da]:8080/api/v2/canister/rwlgt-iiaaa-aaaaa-aaaaa-cai/query: Elapsed(())
```

Most likely there is a race condition between accessing port 8080 to update the topology snapshot and blocking that same port. However we don't need to update the topology snapshot, since we await the new registry version based on metrics afterwards.

### DIFF
--- a/rs/tests/networking/firewall/firewall_correctness_test.rs
+++ b/rs/tests/networking/firewall/firewall_correctness_test.rs
@@ -123,19 +123,13 @@ pub fn firewall_correctness_test(env: TestEnv) {
             &logger,
         )
         .await;
-        let new_registry_version = env
-            .topology_snapshot()
-            .block_for_newer_registry_version()
-            .await
-            .unwrap()
-            .get_registry_version();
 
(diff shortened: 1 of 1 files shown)


pierugo-dfinity (approved):
Thanks for fixing this!

PR #10363, opened 2026-06-01
feat: install all dockerfile tools as root

nmattia:
The removes any mentions of extra users `ubuntu` and `buildifier` in the Dockerfile aside for the actual user creation. All tools are now installed as root, though still usable by `ubuntu` and `buildifier` when containers are run.


nmattia:
Note: we go through a lot of hoops to make `ubuntu` and `buildifier` available, maybe we should just forget about that and run everything as root or settle on one UID (probably 1001 to align with GHA and namespace).

tmu0 (approved):

mbjorkqvist (approved):

basvandijk (approved):

PR #10364, opened 2026-06-01
test: make `FakeStateManager` return a valid height witness

pierugo-dfinity:
This PR makes `FakeStateManager` return a valid height witness in its implementation of `list_state_hashes_to_certify`. This change is driven by an upcoming new consensus integration test which depends on the certified height increasing. Though as of today, nodes in the integration test keeps [invalidating](https://github.com/dfinity/ic/blob/8dc9fa9ad19c79f9d20d757fe95c306fc66389be/rs/consensus/certification/src/certifier.rs#L604) certification shares (preventing the certified height to increase) because the height witness is not valid.

To do so, the `FakeStateManager` now calls the [same functions](https://github.com/dfinity/ic/blob/8dc9fa9ad19c79f9d20d757fe95c306fc66389be/rs/state_manager/src/lib.rs#L1914-L1927) as `StateManagerImpl` (i.e. `replicated_state_as_lazy_tree`, `hash_lazy_tree`, `compute_state_height_witness`) to compute the witness instead of using a `::new_for_testing`.
The common part of the code was thus moved to the `ic_canonical_state` crate.


rs/state_manager/src/lib.rs
```diff
@@ -25,8 +25,8 @@ use crossbeam_channel::{Sender, bounded, unbounded};
 use ic_canonical_state::lazy_tree_conversion::replicated_state_as_lazy_tree;
 use ic_canonical_state_tree_hash::{
     hash_tree::{HashTree, HashTreeError, hash_lazy_tree},
-    lazy_tree::LazyTree,
-    lazy_tree::materialize::materialize_partial,
+    lazy_tree::{LazyTree, materialize::materialize_partial},
+    witness::compute_state_height_witness,
```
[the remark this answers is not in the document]
pierugo-dfinity: Faster than CI 👀

rs/canonical_state/tree_hash/src/witness.rs
```diff
@@ -0,0 +1,15 @@
+use crate::{
+    hash_tree::HashTree,
+    lazy_tree::{LazyTree, materialize::materialize_partial},
+};
+use ic_crypto_tree_hash::{Witness, sparse_labeled_tree_from_paths};
+
+pub fn compute_state_height_witness(lazy_tree: &LazyTree, hash_tree: &HashTree) -> Witness {
+    let paths = vec![vec!["metadata".into(), "height".into()].into()];
```
eichhorl: To me, this crate seems more like a library which shouldn't need to know about the state layout. Maybe a more natural place could be in `rs/canonical_state/src`, which already has constants for the metadata and height labels?

eichhorl (approved):

rs/canonical_state/tree_hash/src/witness.rs
pierugo-dfinity: Makes sense yes! Hadn't considered that

schneiderstefan (approved):

PR #10365, opened 2026-06-01
fix: CON-1383 CON-1401 Abort download of stripped messages if surrounding block becomes unwanted

eichhorl:
Block proposals are transmitted/downloaded in stages:
1. An advert is received containing the block id (height and hash). This ID is passed to the bouncer.
2. Once the bouncer returns `Wanted`, the _stripped_ proposal is downloaded (ingress and IDKG dealings are replaced with hashes). If the bouncer returns `Unwanted` before the download finishes, the advert is dropped instead.
3. The stripped proposal is assembled by collecting the missing ingress messages and IDKG dealings from the local pools or from peers.

Until now, we would continue to assemble the stripped proposal, even if the bouncer switched to `Unwanted` in the mean time. 

With this PR, we instead abort the assembly once the surrounding block becomes unwanted, for instance because the finalized height already advanced past it.

### DIFF
--- a/rs/p2p/artifact_downloader/Cargo.toml
+++ b/rs/p2p/artifact_downloader/Cargo.toml
@@ -29,6 +29,7 @@ tracing = { workspace = true }
 [dev-dependencies]
 assert_matches = { workspace = true }
 criterion = { workspace = true }
+tokio = { workspace = true, features = ["test-util"] }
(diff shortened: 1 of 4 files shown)


pierugo-dfinity (approved):
Nice

PR #10368, opened 2026-06-02
feat: support docker as container runtime in container-run.sh

nmattia:
Add an optional `docker` backend to ci/container/container-run.sh, selected via the CONTAINER_RUNTIME env var (defaults to `podman`, so existing behaviour is unchanged). The podman code paths are left identical with master.

When CONTAINER_RUNTIME is 'docker':
  - run unprivileged (no `sudo`) with the minimal isolation flags the nested rootless podman used by the IC-OS build needs: --device /dev/fuse, unconfined seccomp/apparmor, disabled labelling, systempaths=unconfined, --cap-add SYS_ADMIN and --network=host (much narrower than podman's --privileged);
  - use `image inspect` instead of podman's `image exists`;
  - treat /.dockerenv as a nested-container marker;
  - source the ict_testnets dir from the cache dir rather than host /tmp (some namespaced docker daemons can't bind-mount the host's /tmp).

A small entrypoint shim (ci/container/docker-init.sh) opens up /dev/fuse (docker exposes it 0600 root:root) and ensures /tmp/zig-cache exists, via the image's passwordless sudo.

### DIFF
--- a/ci/container/container-run.sh
+++ b/ci/container/container-run.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -eEuo pipefail
 
-## This script only supports podman as container runtime
+## Supports two container runtimes, selected via the CONTAINER_RUNTIME env var:
+## 'podman' (default, rootful & privileged) and 'docker' (using docker daemon).
 
 eprintln() {
     echo "$@" >&2
@@ -14,6 +15,14 @@ warn() {
     eprintln "$@"
     tput -T xterm sgr0 >&2
 }
+
+# Container runtime to use: 'podman' (default) or 'docker'.
+RUNTIME="${CONTAINER_RUNTIME:-podman}"
+if [ "$RUNTIME" != podman ] && [ "$RUNTIME" != docker ]; then
+    eprintln "Unsupported CONTAINER_RUNTIME '$RUNTIME' (expected 'podman' or 'docker')."
+    exit 1
+fi
+eprintln "Using container runtime '$RUNTIME'"
 
 usage() {
     cat <<EOF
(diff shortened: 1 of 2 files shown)


ci/container/docker-init.sh
nmattia, line 1: Would not be a bad idea to also have an entrypoint for podman where we could e.g. set up the bazel completion (instead of the current hack)

ci/container/container-run.sh
```diff
@@ -250,8 +300,8 @@ if [ -f "$HOME/.container-run.conf" ]; then
     # when it is in use.
     warn "Sourcing user's ~/.container-run.conf"
     source "$HOME/.container-run.conf"
-    PODMAN_RUN_ARGS+=("${PODMAN_RUN_USR_ARGS[@]}")
+    RUNTIME_RUN_ARGS+=("${PODMAN_RUN_USR_ARGS[@]}")
 fi
 
 set -x
-exec "${CONTAINER_CMD[@]}" run "${PODMAN_RUN_ARGS[@]}" -w "$WORKDIR" "$IMAGE" "${cmd[@]}"
+exec "${CONTAINER_CMD[@]}" run "${RUNTIME_RUN_ARGS[@]}" -w "$WORKDIR" "$IMAGE" "${cmd[@]}"
```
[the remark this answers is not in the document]
nmattia: Will do separately.

ci/container/container-run.sh
```diff
@@ -32,7 +41,7 @@ To run a different shell or command, pass it as arguments, e.g.:
 EOF
 }
 
-if [ -e /run/.containerenv ]; then
+if [ -e /run/.containerenv ] || [ -e /.dockerenv ]; then
```
nmattia:
```suggestion
if [ -e /run/.containerenv ]; then
```

This is breaking the repro checks, so let's just remove it for the docker case.

basvandijk (approved):

PR #10369, opened 2026-06-02
feat: build mkfs.ext4 (e2fsprogs) from source and use it in ext4_image

nmattia:
Build the mke2fs binary (mkfs.ext4) from the e2fsprogs sources and expose it as //:mkfs.ext4, then use it in the ext4_image rule and build_ext4_image.py instead of the host's /usr/sbin/mkfs.ext4.


basvandijk (approved):

third_party/BUILD.e2fsprogs.bazel
```diff
@@ -0,0 +1,61 @@
+"""
+e2fsprogs, built from source. The bundled libuuid/libblkid are enabled so we
+don't link against system copies. The resulting mke2fs binary is currently
+dynamically linked against libc/libpthread because the hermetic glibc toolchain
+doesn't ship static libc archives -- `-static` is passed but silently ignored
+by the linker. Fully-static linkage via the musl variant of the toolchain is a
+follow-up.
```
nmattia:
```suggestion
don't link against system copies. 
```

frankdavid (approved):

PR #10372, opened 2026-06-02
feat: [NODE-1962] Create baseimage for bootloader

Bownairo:
Create a new baseimage for the bootloader, and use this instead of ubuntu snapshots for deterministic builds.

We've found that snapshots have a lower ratelimit, and so can fail builds on CI. For now, we move the bootloader to a baseimage like we have for other IC-OS builds, but going forward we should see about moving all of these to mirrored snapshots, or rework the bootloader build away from containers (we only run a single command).

### DIFF
--- a/ic-os/bootloader/context/docker-base
+++ b/ic-os/bootloader/context/docker-base
@@ -1,0 +1,1 @@
+ghcr.io/dfinity/bootloader-base-branch@sha256:28146eac51aa807d809b66f9b9152d2965bd1ff4633ce469bde58d4dbc4386fa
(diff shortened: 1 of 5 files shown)


ic-os/bootloader/context/docker-base
```diff
@@ -0,0 +1 @@
+ghcr.io/dfinity/bootloader-base-branch@sha256:28146eac51aa807d809b66f9b9152d2965bd1ff4633ce469bde58d4dbc4386fa
```
Bownairo, line 1: I'm using a tag built off of the branch for now, but we can very quickly move it to a `master` built image once this merges.

I can also split the PR up (images should not be "trusted" if they are built off `master`), if we would prefer to not use `-branch` images at all.

basvandijk (approved):

blind-oracle (approved):

ic-os/bootloader/context/docker-base
Bownairo: Tada! https://github.com/dfinity/ic/pull/10381

PR #10374, opened 2026-06-03
fix: persist log_memory_store_persistent_next_idx

mraszyk:
This PR fixes the case of producing a canister log with zero log memory limit set on the canister: this case used to trigger
```
ERRO s:/n:/ic_state_manager/checkpoint state_manager_replicated_state_altered_after_checkpoint: Replicated state altered: system_state.log_memory_store.persistent_next_idx
```
because `persistent_next_idx` was not persisted in the checkpoint separately. When the checkpoint was loaded, `persistent_next_idx` was set based on the legacy `next_idx` which is different for a zero log memory limit (the legacy `next_idx` keeps increasing as canister logs are produced because the legacy store always has the default 4KiB capacity).


alin-at-dfinity (approved):
LGTM

PR #10378, opened 2026-06-03
feat(outcalls): add `PATCH` to `http_request`, rejected until rollout

ggreif:
## What

Adds the `PATCH` HTTP method to canister HTTPS outcalls (`http_request`), following the `PUT`/`DELETE` rollout in #8715 / #8717.

`PATCH` variants are added across every HTTP-method enum and conversion in the outcall path:

- `ic_management_canister_types_private::HttpMethod` and the public `ic-management-canister-types` `HttpMethod` (+ `tests/ic.did` candid + fixture)
- `CanisterHttpMethod` (+ the `state.metadata.v1` protobuf enum, its `.proto`, and the generated Rust)
- the `https_outcalls` service `.proto`, the adapter method mapping (`rpc_server.rs`), and the client (`client.rs`)
- the pocket-ic REST `CanisterHttpMethod` and the pocket-ic-server mapping

## Rejected in the execution layer until rollout

Per review, and following the staged approach from #8715: this PR **plumbs** `PATCH` but does **not** enable it on replicated subnets yet. `generate_from_args` and `generate_from_flexible_args` reject `PATCH` outright with a new `CanisterHttpRequestContextError::HttpMethodNotYetSupported` (*"The PATCH HTTP method is not yet supported."*), independent of `is_replicated`. This guarantees no `PATCH` `http_method` enters replicated state before the variant is understood by all replicas (avoiding cross-version deserialization / replay divergence during rollout).

A follow-up PR removes the guard once the rollout is complete, at which point `PATCH` behaves like `PUT`/`DELETE` — permitted only for non-replicated / deterministic-response-count requests, since `PATCH` is a non-idempotent mutation (RFC 5789 §2).

## Tests

- `canister_http.rs`: `patch_is_rejected_until_rollout` asserts `PATCH` is rejected via both entry points; `PATCH` is dropped from the `PUT`/`DELETE` cases in `put_delete_requires_non_replicated` and `flexible_methods_require_deterministic_response_counts`; the enum-stability assertion stays `[1, 2, 3, 4, 5, 6]`.
- `https_outcalls/adapter/tests/server_test.rs`: `test_canister_http_server_patch` + a `/patch` route on the test server (the adapter mapping is below the execution-layer guard, so it is exercised directly).
- `rs/tests/networking/canister_http_correctness_test.rs`: `test_patch_rejected` asserts a `PATCH` outcall is rejected.
- `httpbin-rs`: `/anything` gains a `PATCH` route.

## Notes for reviewers

- Staged rollout per @eichhorl: `PATCH` is plumbed but hard-rejected in the execution layer; the enabling PR follows once the rollout is done.
- Built/validated by CI rather than locally.

Spec/docs companion: dfinity/portal#6245. Requested in dfinity/portal#6244.


basvandijk (dismissed):

ggreif:
@basvandijk, sorry Bas, I killed your review. Anything needs to be done on my side?

basvandijk (dismissed):

basvandijk:
> @basvandijk, sorry Bas, I killed your review. Anything needs to be done on my side?

No worries. For external PRs we require a review for the last commit because it would be too easy for an attacker to sneak in a trojan otherwise.

basvandijk (dismissed):

packages/ic-management-canister-types/src/lib.rs
```diff
@@ -911,6 +911,9 @@ pub enum HttpMethod {
     /// DELETE
     #[serde(rename = "delete")]
     DELETE,
+    /// PATCH
+    #[serde(rename = "patch")]
+    PATCH,
```
eichhorl, line 916: I think there is also a `tests/ic.did` file and a changelog that we should update for this crate

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1104,6 +1104,7 @@ pub enum CanisterHttpMethod {
     HEAD,
     PUT,
     DELETE,
+    PATCH,
```
eichhorl, line 1107: Maybe we can also update the changelog of this crate

eichhorl (commented):
LGTM, however, for compatibility reasons we should follow the same approach that was previously taken by https://github.com/dfinity/ic/pull/8715 and https://github.com/dfinity/ic/pull/8717.

Ideally, this PR adds support for the new method, but still rejects PATCH http requests in the execution layer, such that they don't enter the replicated state yet.

Once support is rolled out to all subnets, I can then make the PR to enable the new method.

ggreif:
@eichhorl thanks for the review! Before I implement the staged-rollout change, one clarification on what "reject in the execution layer" should mean here.

PATCH is already gated exactly like PUT/DELETE in `generate_from_args` (`rs/types/types/src/canister_http.rs`): all three are rejected in **replicated** mode (`DeterministicResponseCountRequired`) and only accepted with `is_replicated = false`. So PATCH already cannot enter replicated state today — it's limited to non-replicated outcalls, same as PUT/DELETE.

Do you want PATCH rejected **entirely** (non-replicated too) until the rollout lands — treating it as fully not-yet-enabled, like #8715 did before its enabling PR — or does the existing non-replicated-only gate already satisfy "doesn't enter the replicated state yet"?

One scope note: `generate_from_args` is the single shared entry point (the replica's execution layer *and* PocketIC both reach it via `execution_environment.rs:1260`), so a blanket rejection there would also disable **non-replicated** PATCH under PocketIC. We're relying on non-replicated PATCH in PocketIC to dogfood a generated client, so if a full rejection is needed for the rollout I'd want to scope it so non-replicated / PocketIC stays usable. Happy to implement whichever you prefer — just want the boundary right before I push.

eichhorl:
@ggreif Note the name clash between (non-)replicated outcalls and the replicated state. All HTTP contexts are stored in the replicated state regardless of wether or not they are (non-)replicated outcalls.

For that reason I would suggest to follow the same approach that was previously taken by https://github.com/dfinity/ic/pull/8715 and https://github.com/dfinity/ic/pull/8717 and reject all PATCH calls until the support is rolled out to all subnets, at which point the new method may be enabled.

ggreif:
@eichhorl thanks — both threads addressed:

**Staged rollout.** `generate_from_args` and `generate_from_flexible_args` now reject `PATCH` outright with a new `CanisterHttpRequestContextError::HttpMethodNotYetSupported` (*"The PATCH HTTP method is not yet supported."*), independent of `is_replicated` — so no `PATCH` `http_method` can enter replicated state until the follow-up enabling PR after rollout, matching the #8715/#8717 staging. (`26e3ae7d1b`)

**`ic-management-canister-types` + `pocket-ic`.** Added `patch` to the `http_request` method variant in `packages/ic-management-canister-types/tests/ic.did`, and a changelog entry to both `packages/ic-management-canister-types/CHANGELOG.md` and `packages/pocket-ic/CHANGELOG.md` (noting the variant is plumbed but not yet enabled). (`de0675b006`)

The PR title/description are updated to describe the staged approach. The PATCH system test now asserts rejection (`test_patch_rejected`) and a unit test (`patch_is_rejected_until_rollout`) covers both entry points.

Ready for another look. @basvandijk — re-review on the latest commit whenever you get a chance.

basvandijk (dismissed):

ggreif:
@eichhorl exactly — thanks for spelling out the naming clash. Agreed: since every HTTP context is stored in the replicated state regardless of whether the *outcall* is (non-)replicated, gating on `is_replicated` wouldn't be sufficient. (Our comments just crossed.) The pushed change does exactly what you describe — `26e3ae7d1b` rejects all PATCH calls in both `generate_from_args` and `generate_from_flexible_args`, independent of `is_replicated`, so no PATCH context can enter the replicated state until the enabling PR after rollout.

eichhorl (dismissed):

ggreif:
The test has failed for unrelated reasons. Can you rerun them, please @eichhorl (or @basvandijk)?

eichhorl:
@ggreif I restarted them, but it seems CI is not having a good day. This failure seems to be persistent however, so we probably have to switch the numbers in this estimate:
```
thread 'load_metrics_e2e_test' (17) panicked at rs/recovery/subnet_splitting/tests/integration_tests.rs:182:9:
assertion `left == right` failed
  left: Estimates { source: 19, destination: 20 }
 right: Estimates { source: 20, destination: 19 }
 ```

**Comment by ggreif** [comment]
> CI is not having a good day

Yes, a second run failed the same (or similar) way. I rebased, let's see...

**Review by eichhorl — DISMISSED** [review]

**Comment by ggreif** [comment]
@eichhorl @basvandijk there is no build scheduled... Something is amiss.

**Comment by basvandijk** [comment]
> there is no build scheduled... Something is amiss.

Nothing amiss. Builds for PRs from forks have to be scheduled manually for security reasons (think an attacker injecting a trojan in a CI job which will then run on our infrastructure). I'll schedule another one ...

**Review by basvandijk — DISMISSED** [review]

**Comment by ggreif** [comment]
> Run on [f99fe64](https://github.com/dfinity/ic/commit/f99fe640cad7df97f943ee2a3cd4c5ebcfbd1927) URL: https://github.com/dfinity/ic/actions/runs/27006751829

I am starting to suspect that the damage is caused by my changes. The breakage seems symmetrical, so maybe it is not breakage per se, but just subtly changed behaviour? @basvandijk @eichhorl

You probably have a better AI to diagnose than me...

**Comment by basvandijk** [comment]
> I am starting to suspect that the damage is caused by my changes.

No it's probably the flakiness we have been suffering from for the last few days. We know why it's happening:
Image downloads take too long because of some change in the networking between our DCs.
We also have a plan to address it which @dfinityianblenke is currently working on.

**Comment by eichhorl** [comment]
@ggreif I do think the (benign) failure in the subnet splitting test is caused by this PR, and we should fix it by updating the estimates of that test according to what the asserts expect: https://github.com/dfinity/ic/pull/10378#issuecomment-4624027005

Unfortunately I don't think I can push to your fork

**Comment by ggreif** [comment]
> Unfortunately I don't think I can push to your fork

@eichhorl @basvandijk I pushed a 2-line fix.

**Review by basvandijk — DISMISSED** [review]

**Comment by eichhorl** [comment]
@ggreif Thanks Gabor, running the test locally reveals that other asserts after the one that was fixed will fail as well. Here is the complete diff:
```diff
diff --git a/rs/recovery/subnet_splitting/tests/integration_tests.rs b/rs/recovery/subnet_splitting/tests/integration_tests.rs
index 9dc9aadacd..033c48ba2c 100644
--- a/rs/recovery/subnet_splitting/tests/integration_tests.rs
+++ b/rs/recovery/subnet_splitting/tests/integration_tests.rs
@@ -196,8 +196,8 @@ fn load_metrics_e2e_test() {
         assert_eq!(
             local_subnet_messages_executed_upper_bound,
             Estimates {
-                source: 15,
-                destination: 13,
+                source: 13,
+                destination: 15,
             }
         );
         assert_eq!(
@@ -210,8 +210,8 @@ fn load_metrics_e2e_test() {
         assert_eq!(
             heartbeats_and_global_timers_executed,
             Estimates {
-                source: 276,
-                destination: 418,
+                source: 418,
+                destination: 276,
             }
         );
         // Check if the split finder found a split satisfying the load constraints
```

ggreif:
> @ggreif Thanks Gabor, running the test locally reveals that other asserts after the one that was fixed will fail as well. Here is the complete diff:

@eichhorl @basvandijk the combined fix is pushed!

eichhorl (approved):

basvandijk (approved):

ggreif:
All green! 🍀🪲
@eichhorl @basvandijk

ggreif:
> Once support is rolled out to all subnets, I can then make the PR to enable the new method.

@eichhorl How is the rollout going?

eichhorl:
> > Once support is rolled out to all subnets, I can then make the PR to enable the new method.
> 
> @eichhorl How is the rollout going?

@ggreif the change of this PR is being rolled out this week. https://github.com/dfinity/ic/pull/10453 (which enables the feature) was merged yesterday, and should be rolled out to all subnets next week.

PR #10379, opened 2026-06-03
fix: uninstall_code does not deallocate canister log

mraszyk:
This PR fixes the `uninstall_code` endpoint of the management canister to only clear the canister log records instead of fully deallocating the canister log (which is now only the case if the canister runs out of cycles). This is analogous to compute/memory allocation that are also not changed by the `uninstall_code` endpoint of the management canister, but only if the canister runs out of cycles.

### DIFF
--- a/rs/execution_environment/src/canister_manager/tests.rs
+++ b/rs/execution_environment/src/canister_manager/tests.rs
@@ -4941,16 +4941,6 @@ fn uninstall_code_on_empty_canister_updates_subnet_available_memory() {
     // Assert that canister history memory was non empty.
     let initial_canister_history_memory_usage = canister_history_memory_usage(&mut test);
     assert_gt!(initial_canister_history_memory_usage, 0);
-    let initial_log_memory_store_memory_usage = test
-        .canister_state(canister_id)
-        .log_memory_store_memory_usage()
-        .get();
-    if LOG_MEMORY_STORE_FEATURE_ENABLED {
-        // Assert that canister log memory store memory was non empty.
-        assert_gt!(initial_log_memory_store_memory_usage, 0);
-    } else {
-        assert_eq!(initial_log_memory_store_memory_usage, 0);
-    }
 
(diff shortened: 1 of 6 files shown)


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -4941,16 +4941,6 @@ fn uninstall_code_on_empty_canister_updates_subnet_available_memory() {
     // Assert that canister history memory was non empty.
     let initial_canister_history_memory_usage = canister_history_memory_usage(&mut test);
     assert_gt!(initial_canister_history_memory_usage, 0);
-    let initial_log_memory_store_memory_usage = test
```
schneiderstefan, line 4944: Would it make sense to just assert the opposite, rather than removing the asserts altogether?

rs/execution_environment/src/scheduler.rs
```diff
@@ -891,6 +891,7 @@ impl SchedulerImpl {
                 canister.system_state.compute_allocation = ComputeAllocation::zero();
                 canister.system_state.memory_allocation = MemoryAllocation::default();
                 canister.system_state.clear_canister_history();
+                canister.remove_log();
```
schneiderstefan, line 894: Why was this line not needed before?

schneiderstefan (approved):

rs/execution_environment/src/scheduler.rs
mraszyk: because `uninstall_canister` (called a few lines above) deallocated the new log memory store before; now `uninstall_canister` only clears the log records, but does not deallocate the new log memory store.

rs/execution_environment/src/canister_manager/tests.rs
mraszyk: We could assert that the log memory store memory usage did not change, but there's a dedicated test `test_canister_uninstall_code_clears_logs` for this scenario now.

PR #10384, opened 2026-06-04
fix: deflake //rs/tests/consensus/orchestrator:unstuck_subnet_test

basvandijk:
## What

Deflakes `//rs/tests/consensus/orchestrator:unstuck_subnet_test`, which was
frequently failing on master with the `test` task timing out.

## Root causes

There were **two independent** root causes:

### 1. `FileHashMismatchError` detection race

The test verified the failing upgrade by *following* each node's journal and
waiting for a **new** `FileHashMismatchError` line:

```rust
JournalStreamer::new(session).follow().max_lines(1)
    .contains("FileHashMismatchError")
```

The orchestrator only logs that error **after** downloading the whole
multi-hundred-MB GuestOS image and noticing the hash mismatch, and it retries
the download with an **exponential backoff** (gaps grow to several minutes). If
the stream started during a backoff gap — or after the error had already been
logged — `follow()` would block waiting for the *next* occurrence, frequently
past the test timeout.

Fix: poll the whole journal for an *already-logged* occurrence with a retry
loop instead of following it.

### 2. Too-tight test-driver timeouts

The `test` task ran under the default 10-minute per-test/overall timeouts, but
the test internally waits up to 30 minutes for the upgrade and downloads the
large GuestOS image multiple times (the failing upgrade, the manual
re-download, and the successful upgrade). That image download has a **high
variance** and can occasionally take well over 10 minutes, so the driver killed
the task mid-run.

Fix: raise the per-test/overall timeouts and set `test_timeout = "eternal"` at
the Bazel level (matching the existing `rejoin_test` precedent in
`rs/tests/message_routing/BUILD.bazel`).

## Validation

Validated locally with `bazel test --runs_per_test=3 --jobs=1`: **3/3 passing**
(max = 1188.6s, min = 440.8s, avg = 695.9s). The slowest run exercised the
slow-download path that the previous, tighter caps were killing.

---

Created following `.claude/skills/fix-flaky-tests/SKILL.md`.

### DIFF
--- a/rs/tests/consensus/orchestrator/BUILD.bazel
+++ b/rs/tests/consensus/orchestrator/BUILD.bazel
@@ -108,6 +108,7 @@ system_test_nns(
     tags = [
         "long_test",
     ],
+    test_timeout = "eternal",
     runtime_deps = MESSAGE_CANISTER_RUNTIME_DEPS,
     deps = [
         "//rs/registry/subnet_type",
--- a/rs/tests/consensus/orchestrator/unstuck_subnet_test.rs
+++ b/rs/tests/consensus/orchestrator/unstuck_subnet_test.rs
@@ -90,16 +90,41 @@ fn test(test_env: TestEnv) {
     ));
     info!(logger, "Upgrade started");
 
+    // The orchestrator retries the failing image download with an exponential backoff, so it logs
+    // `FileHashMismatchError` with ever growing gaps. We therefore poll the whole journal for an
+    // already-logged occurrence of the error instead of following the journal and waiting for a new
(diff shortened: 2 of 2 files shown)


pierugo-dfinity:
This adds quite some complexity without really fixing the root cause... I feel like it's a test that actually gives good signal that something is wrong with the image download if it fails. Deflaking it this way feels like hiding the problem imo.

Also, I think a 1-hour timeout looks *very* generous for this test

basvandijk:
Agreed. I'm seeing that a lot of the other flaky tests are failing because Farm hosts have trouble downloading images. So I first would like to understand better what's going wrong there before fixing individual tests.

PR #10387, opened 2026-06-04
feat(registry-canister): get_subnet endpoint

mraszyk:
This PR introduces a `get_subnet` endpoint on the NNS registry canister that returns the SubnetRecord for a given subnet ID. The immediate motivation for this change is to enable the migration canister check if a subnet is a cloud engine (by retrieving the subnet type from the corresponding subnet record), but we implement a generic `get_subnet` endpoint in this PR to cover future use cases, too.

The sizes of subnet records on mainnet vary between 732 bytes and 2103 bytes and take between ~100k and ~150k instructions to produce.

### DIFF
--- a/rs/registry/canister/src/get_subnet.rs
+++ b/rs/registry/canister/src/get_subnet.rs
@@ -58,8 +58,9 @@ impl TryFrom<SubnetRecordPb> for SubnetRecord {
                 .unwrap_or(CanisterCyclesCostSchedulePb::Unspecified);
         let canister_cycles_cost_schedule = match cost_schedule_pb {
             CanisterCyclesCostSchedulePb::Free => CanisterCyclesCostSchedule::Free,
-            CanisterCyclesCostSchedulePb::Normal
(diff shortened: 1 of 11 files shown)


rs/registry/canister/src/get_subnet.rs
```diff
@@ -0,0 +1,128 @@
...
+            subnet_admins,
+            recalled_replica_version_ids: pb.recalled_replica_version_ids,
+            resource_limits: pb.resource_limits,
+        })
+    }
+}
+
+impl Registry {
+    pub fn get_subnet_record(&self, request: GetSubnetRequest) -> Result<SubnetRecord, String> {
+        let subnet_id = request
```
schneiderstefan, line 113: More a question than anything, but does this type of registry endpoint usually have a rate limit or other protections?

mraszyk: The new endpoint is analogous to the existing endpoint `get_subnet_for_canister` in terms of security measures.

r-birkner (approved):

eichhorl (approved):

bjoernek (approved):

PR #10391, opened 2026-06-04
feat: [DSM-143] Atomic outbound `Request` enqueuing and `Callback` registration

alin-at-dfinity:
Encapsulate `Request` enqueueing and `Callback` registration within `SystemState`. We already have `pop_output()` returning a `(Response, Callback)` tuple. This change addresses the other end: atomic `Callback` creation while enqueuing an outbound `Request`.

There is still a gap in `unregister_callback()`, which is public to allow Execution to unregister unreplicated (only) query callbacks. Other than that, `Callbacks` are now atomically registered and unregistered when `Requests` are enqueued and `Responses` are dequeued.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
@@ -205,7 +205,7 @@ impl RequestInPrep {
 
 /// Turns a `RequestInPrep` into an `OutputRequest`.
 #[allow(clippy::too_many_arguments)]
-pub(crate) fn into_request(
+pub(crate) fn into_output_request(
     RequestInPrep {
         sender,
         callee,
--- a/rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
@@ -2,14 +2,15 @@ use super::{sandbox_safe_system_state::SandboxSafeSystemState, valid_subslice};
 use ic_base_types::InternalAddress;
 use ic_interfaces::execution_environment::{HypervisorError, HypervisorResult};
 use ic_logger::ReplicaLogger;
+use ic_replicated_state::OutputRequest;
 use ic_types::Time;
 use ic_types::{
     CanisterId, NumBytes, PrincipalId,
-    messages::{CallContextId, NO_DEADLINE, Request},
-    methods::{Callback, WasmClosure},
+    messages::{CallContextId, NO_DEADLINE},
+    methods::WasmClosure,
     time::CoarseTime,
 };
-use ic_types_cycles::{CompoundCycles, Cycles, Instructions, RequestAndResponseTransmission};
+use ic_types_cycles::Cycles;
 use ic_wasm_types::doc_ref;
 use serde::{Deserialize, Serialize};
 use std::{convert::TryFrom, time::Duration};
@@ -202,13 +203,7 @@ impl RequestInPrep {
     }
 }
 
-pub(crate) struct RequestWithPrepayment {
-    pub request: Request,
-    pub prepayment_for_response_execution: CompoundCycles<Instructions>,
-    pub prepayment_for_call_transmission: CompoundCycles<RequestAndResponseTransmission>,
-}
-
-/// Turns a `RequestInPrep` into a `Request`.
+/// Turns a `RequestInPrep` into an `OutputRequest`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn into_request(
     RequestInPrep {
@@ -228,7 +223,7 @@ pub(crate) fn into_request(
     sandbox_safe_system_state: &mut SandboxSafeSystemState,
     _logger: &ReplicaLogger,
     time: Time,
-) -> HypervisorResult<RequestWithPrepayment> {
+) -> HypervisorResult<OutputRequest> {
     let destination_canister = CanisterId::unchecked_from_principal(callee);
 
     let payload_size = (method_name.len() + method_payload.len()) as u64;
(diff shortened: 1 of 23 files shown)


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -69,6 +72,131 @@ lazy_static! {
 /// Maximum number of canister changes stored in the canister history.
 pub const MAX_CANISTER_HISTORY_CHANGES: u64 = 20;
 
+/// A bundle of outgoing inter-canister `Request and matching `Callback`.
```
eichhorl:
```suggestion
/// A bundle of outgoing inter-canister `Request` and matching `Callback`.
```

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2546,4 +2688,104 @@ pub mod testing {
...
         };
     }
+
+    /// Builder for `OutputRequest`, for use in tests.
+    pub struct OutputRequestBuilder {
+        request: OutputRequest,
+    }
+
+    impl Default for OutputRequestBuilder {
+        /// Creates an `OutputRequestBuilder`` with default values.
```
eichhorl:
```suggestion
        /// Creates an `OutputRequestBuilder` with default values.
```

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2546,4 +2688,104 @@ pub mod testing {
...
+                    on_reply: ic_types::methods::WasmClosure::new(0, 0),
+                    on_reject: ic_types::methods::WasmClosure::new(0, 0),
+                    on_cleanup: None,
+                },
+            }
+        }
+    }
+
+    impl OutputRequestBuilder {
+        /// Creates a new `RequestBuilder`.
```
eichhorl:
```suggestion
        /// Creates a new `OutputRequestBuilder`.
```

rs/replicated_state/tests/system_state.rs
```diff
@@ -365,21 +359,15 @@ fn induct_messages_to_self_memory_limit_test_impl(
         Some(CanisterMessage::Response { response, callback }),
         fixture.pop_input(),
     );
-    assert_eq!(Some(CanisterMessage::Request(request)), fixture.pop_input(),);
+    assert_matches!(fixture.pop_input(), Some(CanisterMessage::Request(_)));
```
eichhorl, line 375: A lot of the assertions in this file were weakened like this, maybe `issue_outbound_call` could also return the `CallbackId` which we could check instead?

rs/replicated_state/tests/system_state.rs
```diff
@@ -247,11 +240,12 @@ fn induct_messages_to_self_in_running_status_works() {
 
 #[test]
 fn induct_messages_to_self_in_stopped_status_does_not_work() {
-    let mut fixture = SystemStateFixture::stopped();
-
+    let mut fixture = SystemStateFixture::running();
     fixture
         .push_output_request(default_request_to_self())
         .unwrap();
```
eichhorl, line 259: Should there be a test that pushing a request now fails for a stopped canister?

rs/replicated_state/tests/system_state.rs
```diff
@@ -260,11 +254,12 @@ fn induct_messages_to_self_in_stopped_status_does_not_work() {
...
 #[test]
 fn induct_messages_to_self_in_stopping_status_does_not_work() {
-    let mut fixture = SystemStateFixture::stopping();
-
+    let mut fixture = SystemStateFixture::running();
     fixture
         .push_output_request(default_request_to_self())
         .unwrap();
+
+    fixture.set_stopping();
```
eichhorl, line 275: IIUC this would have actually worked without change?

rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
```diff
@@ -202,13 +203,7 @@ impl RequestInPrep {
...
-pub(crate) struct RequestWithPrepayment {
-    pub request: Request,
-    pub prepayment_for_response_execution: CompoundCycles<Instructions>,
-    pub prepayment_for_call_transmission: CompoundCycles<RequestAndResponseTransmission>,
-}
-
-/// Turns a `RequestInPrep` into a `Request`.
+/// Turns a `RequestInPrep` into an `OutputRequest`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn into_request(
```
eichhorl: Maybe
```suggestion
pub(crate) fn into_output_request(
```
to avoid the name clash with `OutputRequest::into_request`

rs/embedders/tests/sandbox_safe_system_state.rs
```diff
@@ -190,24 +178,22 @@ fn push_output_request_succeeds_with_enough_cycles() {
...
+
+    request.prepayment_for_response_execution = cycles_account_manager
         .prepayment_for_response_execution(
             SMALL_APP_SUBNET_MAX_SIZE,
             cost_schedule,
             WASM_EXECUTION_MODE,
         );
-    let prepayment_for_response_transmission = cycles_account_manager
+    request.prepayment_for_response_transmission = cycles_account_manager
         .prepayment_for_response_transmission(SMALL_APP_SUBNET_MAX_SIZE, cost_schedule);
```
eichhorl, line 190: Should we also set `prepayment_for_call_transmission`?

eichhorl (approved):

rs/embedders/src/wasmtime_embedder/system_api/request_in_prep.rs
alin-at-dfinity: Done.

rs/embedders/tests/sandbox_safe_system_state.rs
alin-at-dfinity: Done.

rs/replicated_state/tests/system_state.rs
alin-at-dfinity: Yeah, it looks like it would have. `Stopping` canisters are still allowed to make downstream calls, just not receive incoming calls (including from themselves).

rs/replicated_state/tests/system_state.rs
alin-at-dfinity: I was considering that at some point, while refactoring one of the tests (I don't remember which). But then I found a cleaner solution for that specific test; and the other tests either just enqueue the one request; or a request and a response; or two requests, which we expect to be popped in order.

I.e. I don't think it's necessary for the tests we currently have.

rs/replicated_state/tests/system_state.rs
alin-at-dfinity: Great idea, done.

PR #10394, opened 2026-06-04
feat: Allow engine controller canister to create and delete subnets

r-birkner:
Add ENGINE_CONTROLLER_CANISTER_ID (`si2b5-pyaaa-aaaaa-aaaja-cai`) and authorize it, alongside governance, to call the registry create_subnet and delete_subnet methods.

### DIFF
--- a/rs/nns/constants/src/lib.rs
+++ b/rs/nns/constants/src/lib.rs
@@ -178,7 +178,7 @@ pub const SNS_AGGREGATOR_CANISTER_ID: CanisterId =
 ///
 /// As of May 2024, it looks like this is only used by (a whole bunch of) tests, mostly as the
 /// argument to send_whitelist.
-pub const ALL_NNS_CANISTER_IDS: [&CanisterId; 18] = [
+pub const ALL_NNS_CANISTER_IDS: [&CanisterId; 19] = [
     &REGISTRY_CANISTER_ID,
     &GOVERNANCE_CANISTER_ID,
     &LEDGER_CANISTER_ID,
@@ -197,9 +197,10 @@ pub const ALL_NNS_CANISTER_IDS: [&CanisterId; 18] = [
     &ICP_LEDGER_ARCHIVE_3_CANISTER_ID,
     &NODE_REWARDS_CANISTER_ID,
     &MIGRATION_CANISTER_ID,
+    &ENGINE_CONTROLLER_CANISTER_ID,
 ];
 
-pub const PROTOCOL_CANISTER_IDS: [&CanisterId; 23] = [
+pub const PROTOCOL_CANISTER_IDS: [&CanisterId; 24] = [
     &REGISTRY_CANISTER_ID,
     &GOVERNANCE_CANISTER_ID,
     &LEDGER_CANISTER_ID,
@@ -223,6 +224,7 @@ pub const PROTOCOL_CANISTER_IDS: [&CanisterId; 23] = [
     &DOGECOIN_CANISTER_ID,
     &BITCOIN_WATCHDOG_CANISTER_ID,
     &DOGECOIN_WATCHDOG_CANISTER_ID,
+    &ENGINE_CONTROLLER_CANISTER_ID,
 ];
 
 /// The current value is 4 GiB, s.t. the SNS governance canister never hits the soft memory limit.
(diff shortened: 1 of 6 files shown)


NikolaMilosa (approved):

rs/registry/canister/tests/create_subnet.rs
```diff
@@ -150,6 +152,17 @@ fn test_a_canister_other_than_the_governance_canister_cannot_create_a_subnet() {
 
 #[tokio::test]
 async fn test_accepted_proposal_mutates_the_registry_some_subnets_present() {
```
mraszyk: maybe name the test `test_governance_canister_can_create_a_subnet`

rs/registry/canister/tests/delete_subnet.rs
```diff
@@ -0,0 +1,129 @@
...
+            .await
+            .unwrap_or_else(|err| {
+                panic!("delete_subnet call by authorized caller {caller} was unexpectedly rejected: {err:?}")
+            });
+
+        let result = Decode!(&response, Result<(), String>).unwrap();
+        assert_eq!(
+            result,
+            Err("Only CloudEngines may be deleted".to_string()),
+            "caller {caller} should pass authorization and reach the business logic"
```
mraszyk: How about making the deleted subnet type of CloudEngine so that we succeed here?

rs/registry/canister/tests/delete_subnet.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    // A principal that is neither governance nor the engine controller.
+    let unauthorized_caller = PrincipalId::new_user_test_id(1);
+    assert_ne!(unauthorized_caller, GOVERNANCE_CANISTER_ID.get());
+    assert_ne!(unauthorized_caller, ENGINE_CONTROLLER_CANISTER_ID.get());
+
+    let payload = DeleteSubnetPayload { subnet_id };
+
+    let response = pocket_ic
+        .update_call(
```
mraszyk, line 89: this is still an ingress message, we could harden the test by proxing the message through a canister so that it is really an inter-canister call (this would capture bugs such as moving the access control to `inspect_message` which would pass this test, but still be insecure)

mraszyk: see test_a_canister_other_than_the_governance_canister_cannot_create_a_subnet

rs/nns/constants/src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub const ICP_LEDGER_ARCHIVE_2_CANISTER_INDEX_IN_NNS_SUBNET: u64 = 14;
 pub const ICP_LEDGER_ARCHIVE_3_CANISTER_INDEX_IN_NNS_SUBNET: u64 = 15;
 pub const NODE_REWARDS_CANISTER_INDEX_IN_NNS_SUBNET: u64 = 16;
 pub const MIGRATION_CANISTER_INDEX_IN_NNS_SUBNET: u64 = 17;
+pub const ENGINE_CONTROLLER_CANISTER_INDEX_IN_NNS_SUBNET: u64 = 18;
```
mraszyk, line 36: The canister does not seem to exist yet, so we should only merge this PR once it does (I believe).

rs/registry/canister/tests/delete_subnet.rs
mraszyk: I'd suggest to test (1) anonymous as well as PrincipalId::new_user_test_id via ingress message (`pic.update_call`), (2) inter-canister call by proxying through universal canister

rs/nns/constants/src/lib.rs
Dfinity-Bjoern: Why? We will be creating that canister through a proposal. We plan to create both proposals today.

mraszyk: e.g., if the ICP ledger canister spawns an archive in the meantime, then this canister ID will not match, I'm not saying it's a security risk, but something to keep in mind

rs/registry/canister/tests/delete_subnet.rs
r-birkner: I added both a test making sure that non-cloud engine subnets cannot be deleted (even if authorized) and a test making sure authorized principals can delete a cloud engine subnet.

rs/nns/constants/src/lib.rs
Dfinity-Bjoern: I checked with the ledgers team and the risk for this is negligible. Let us accept this risk.

Thanks @mraszyk for pointing it out, though!

Dfinity-Bjoern (approved):

PR #10396, opened 2026-06-05
feat: minimal engine controller canister

NikolaMilosa:
This PR implements a minimal canister that will allow on demand creating and deleting of Cloud Engines.


rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+    Ok(caller)
+}
+
+#[update]
+async fn create_engine(args: CreateEngineArgs) -> Result<(), String> {
+    let caller = ensure_authorized()?;
+
+    // Validate node list.
+    if args.node_ids.len() != REQUIRED_NODE_COUNT {
```
r-birkner: Let's drop this step altogether or do
```suggestion
    if args.node_ids.len() < REQUIRED_NODE_COUNT {
```
such that there is some flexibility to also create larger cloud engines.

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,189 @@
...
+        return Err(format!(
+            "Expected exactly {REQUIRED_NODE_COUNT} node ids, got {}",
```
r-birkner: Just aligning it with the above comment
```suggestion
            "Expected at least {REQUIRED_NODE_COUNT} node ids, got {}",
```

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,189 @@
...
+        max_ingress_messages_per_block: 1000,
+        unit_delay_millis: 1000,
+        canister_cycles_cost_schedule: Some(CanisterCyclesCostSchedule::Free),
+        features: SubnetFeatures {
+            canister_sandboxing: false,
+            http_requests: true,
+            sev_enabled: Some(false),
+        },
+        // Maybe this can be used to know subnet id ahead of time?
+        // subnet_id_override: todo!(),
```
r-birkner: I think this is for testing (e.g., I want to create a subnet with the same ID as on mainnet)

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,189 @@
+//! The engine controller canister.
+//!
+//! This canister provides a thin user-facing API on top of the registry
+//! canister's `create_subnet` / `delete_subnet` endpoints. Only a single,
+//! hard-coded authorized principal may invoke its methods.
```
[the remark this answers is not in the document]
NikolaMilosa: Generally, we intend to not change this for a network. This was made configurable for deploying to non-mainnet setups where we don't have access to the production key.

rs/engine_controller/tests/tests.rs
```diff
@@ -0,0 +1,424 @@
...
+            replica_version_id: test_replica_version(),
+        },
+    )
+    .await
+    .unwrap_err();
+    assert!(err.contains("Duplicate node id"), "unexpected error: {err}");
+}
+
+#[tokio::test]
+async fn init_arg_overrides_authorized_caller_and_survives_upgrade() {
```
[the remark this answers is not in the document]
NikolaMilosa: Probably won't be used like that.

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+pub struct EngineControllerInitArgs {
+    /// If `Some`, replaces the default authorized caller; if `None`, the
+    /// default is kept.
+    pub authorized_caller: Option<Principal>,
+}
+
+fn apply_init_args() {
+    // Accept either an empty argument (Candid cannot encode "no arg" for
+    // `opt T` as zero bytes, but installs/upgrades commonly omit the arg
+    // entirely) or a Candid-encoded `opt EngineControllerInitArgs`.
```
mraszyk: Specifying an empty vector of bytes is treated the same as `()` which is compatible with `(opt T)` so there should not be an issue. Could you please fix this in Candid, @lwshang ?

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+    Ok(caller)
+}
+
+#[update]
+async fn create_engine(args: CreateEngineArgs) -> Result<(), String> {
+    let caller = ensure_authorized()?;
+
+    // Validate node list.
+    if args.node_ids.len() < REQUIRED_NODE_COUNT {
```
mraszyk, line 130: this allow any high number of nodes, contradicting the above comment: "Exactly this many nodes are expected when creating a new engine"

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+            return Err(format!("Duplicate node id supplied: {n}"));
+        }
+    }
+
+    // Make sure the caller is part of the subnet admins.
+    let mut subnet_admins: Vec<PrincipalId> =
+        args.subnet_admins.into_iter().map(PrincipalId).collect();
+    let caller_pid = PrincipalId(caller);
+    if !subnet_admins.contains(&caller_pid) {
+        subnet_admins.push(caller_pid);
```
mraszyk, line 148: Shouldn't we reject the input instead?

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+        dkg_interval_length: 499,
+        dkg_dealings_per_block: 1,
+        initial_notary_delay_millis: 300,
+        max_block_payload_size: 4 * 1024 * 1024, // 4 MiB
+        max_ingress_bytes_per_message: 2 * 1024 * 1024, // 2 MiB
+        max_ingress_messages_per_block: 1000,
+        unit_delay_millis: 1000,
+        canister_cycles_cost_schedule: Some(CanisterCyclesCostSchedule::Free),
+        features: SubnetFeatures {
+            canister_sandboxing: false,
```
mraszyk, line 174: Is this really what other subnets are using at the moment? Canister sandboxing is always used so if we set `false` here, then we're likely hitting dead code anyway.

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+            http_requests: true,
+            sev_enabled: Some(false),
+        },
+        ..Default::default()
+    };
+
+    let response: Result<NewSubnet, String> =
+        Call::bounded_wait(REGISTRY_CANISTER_ID.into(), "create_subnet")
```
mraszyk: using a bounded-wait call for a non-idempotent action is risky because a timeout will prevent the caller from knowing if the action failed (and what reason caused the failure) or succeeded; since both caller (this canister) and callee (the registry canister) are on the NNS subnet, using an unbounded-wait call should be fine in this case

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,187 @@
...
+#[update]
+async fn delete_engine(args: DeleteEngineArgs) -> Result<(), String> {
+    ensure_authorized()?;
+
+    let payload = DeleteSubnetPayload {
+        subnet_id: args.subnet_id,
+    };
+
+    let response: Result<(), String> =
+        Call::bounded_wait(REGISTRY_CANISTER_ID.into(), "delete_subnet")
```
mraszyk: using a bounded-wait call for a non-idempotent action is risky because a timeout will prevent the caller from knowing if the action failed (and what reason caused the failure) or succeeded; since both caller (this canister) and callee (the registry canister) are on the NNS subnet, using an unbounded-wait call should be fine in this case

r-birkner (approved):

rs/engine_controller/engine_controller.did
```diff
@@ -0,0 +1,20 @@
...
+type Result = variant { Ok; Err : text };
+
+type EngineControllerInitArgs = record {
+  authorized_caller : opt principal;
+};
+
+service : (opt EngineControllerInitArgs) -> {
+  create_engine : (CreateEngineArgs) -> (Result);
+  delete_engine : (DeleteEngineArgs) -> (Result);
+}
```
mraszyk, line 21: This should be covered by a test such as the one in rs/nns/governance/canister/tests.rs

rs/engine_controller/tests/tests.rs
```diff
@@ -0,0 +1,428 @@
...
+        Some(CanisterSettings {
+            controllers: Some(vec![controller]),
+            ..Default::default()
+        }),
+        canister_id.into(),
+    )
+    .await
+    .unwrap();
+    pic.add_cycles(canister_id.into(), 1_000_000_000_000_000)
+        .await;
```
mraszyk: Why are we adding cycles for NNS canisters?

rs/engine_controller/tests/tests.rs
```diff
@@ -0,0 +1,428 @@
...
+    .await;
+
+    let engine_controller_wasm =
+        Project::cargo_bin_maybe_from_env("engine-controller-canister", &[]);
+    install_at(
+        &pic,
+        ENGINE_CONTROLLER_CANISTER_ID,
+        engine_controller_wasm.bytes(),
+        vec![],
+        Principal::anonymous(),
```
mraszyk: should be ROOT_CANISTER_ID

rs/engine_controller/canister/canister.rs
mraszyk: although here, delete is actually idempotent...

rs/engine_controller/canister/canister.rs
NikolaMilosa: Yes, we will manage that in another place for now, this is just a lower bound. I will fix the comment.

rs/engine_controller/canister/canister.rs
NikolaMilosa: This is then non-actionable on my side right now right?

rs/engine_controller/canister/canister.rs
NikolaMilosa: This is just a safeguard because we still need to always have us as the subnet admin in case something goes bad. As we have it in the dashboard, this is always skipped.

rs/engine_controller/canister/canister.rs
NikolaMilosa: I was looking exactly for the feedback on that! If you have enough context can you suggest what should I put there?

rs/engine_controller/tests/tests.rs
NikolaMilosa: AI did it, I wasn't looking into it. You are right we don't need it.

rs/engine_controller/canister/canister.rs
mraszyk: According to `ic-admin get-topology`, all subnets have `canister_sandboxing: false` so we're fine here and I just took a note to clean up the dead code.

rs/engine_controller/engine_controller.did
NikolaMilosa: Corrected it.

mraszyk: I missed the test - I'd really recommend it since otherwise callers might be confused about the canister's interface (if they pull it from this file, such as dfx does it).

mraszyk (approved):
Approving to unblock, but highly recommend adding a test of the candid interface.

mraszyk (approved):
LGTM Thanks!

rs/engine_controller/canister/canister.rs
```diff
@@ -0,0 +1,176 @@
...
+        max_ingress_bytes_per_message: 2 * 1024 * 1024, // 2 MiB
+        max_ingress_messages_per_block: 1000,
+        unit_delay_millis: 1000,
+        canister_cycles_cost_schedule: Some(CanisterCyclesCostSchedule::Free),
+        features: SubnetFeatures {
+            canister_sandboxing: false,
+            http_requests: true,
+            sev_enabled: Some(false),
+        },
+        ..Default::default()
```
eichhorl, line 178: Note that setting `initial_dkg_subnet_id` to `None` here means that the cloud engine will be created by the NNS, which is something we would like to avoid. https://github.com/dfinity/ic/pull/10242 introduced a proposal to set the default initial DKG subnet to something other than NNS (at which point leaving it `None` here would be fine), but that change has not been deployed to the mainnet registry, so we can't make such a proposal yet.

rs/engine_controller/BUILD.bazel
```diff
@@ -0,0 +1,71 @@
...
+    # Keep sorted.
+    "//rs/nns/constants",
+    "//rs/protobuf",
+    "//rs/registry/canister",
+    "//rs/registry/subnet_type",
+    "//rs/types/base_types",
+    "@crate_index//:candid",
+    "@crate_index//:ic-cdk",
+    "@crate_index//:serde",
+]
```
basvandijk: We're moving away from defining dependencies in variables because it's too easy to introduce unused dependencies that way. Consider just inlining them.

basvandijk (approved):

rs/engine_controller/canister/canister.rs
NikolaMilosa: Addressed it! We decided to use `fuqsr`

eichhorl (approved):

PR #10398, opened 2026-06-05
feat: new canister settings minimum_incoming_canister_call_cycles

mraszyk:
This PR introduces a new canister setting `minimum_incoming_canister_call_cycles` specifying the minimum amount of cycles that must be available as transferred cycles in an incoming inter-canister call from a different canister (i.e., ignoring ingress messages: `canister_inspect_message` hook can be used to filter them, albeit only via non-replicated execution).


packages/ic-management-canister-types/src/lib.rs
```diff
@@ -161,6 +161,14 @@ pub struct CanisterSettings {
     ///
     /// Default value: `null` (i.e., no environment variables provided).
     pub environment_variables: Option<Vec<EnvironmentVariable>>,
+    /// Indicates the minimum number of cycles required for an incoming message
+    /// from another canister. Messages with fewer cycles are rejected with a
+    /// `CanisterError`. Ingress messages are not affected.
```
alin-at-dfinity: > Ingress messages are not affected.

I was thinking this kind of defeats the purpose. Took me a while to realize that there's [whatever the single-replica ingress message filtering mechanism is called]. Maybe mention that, just in case someone is looking at the implementation, trying to figure out how to deal with call spam.

packages/ic-management-canister-types/src/lib.rs
```diff
@@ -161,6 +161,14 @@ pub struct CanisterSettings {
...
+    ///
+    /// Must be a number between 0 and 2<sup>128</sup>-1, inclusively.
+    ///
+    /// Default value: `0` (i.e., no minimum enforced).
+    pub minimum_msg_cycles_available: Option<Nat>,
 }
```
alin-at-dfinity, line 174: When I first read the name I thought this was some sort of freezing threshold for cycles reserved for outgoing calls. Maybe we should call it something like "minimum [call] payment"? Or, if you think "pauyment" is only used in the implementation, not the spec, and is thus potentially confusing, then "attached cycles"?

"Available" sounds too much like "balance" or "reservation".

alin-at-dfinity (commented):
Dropping this one comment early. In case you agree with it, it will require a lot of renaming and it's probably best to get it out of the way.

packages/ic-management-canister-types/src/lib.rs
mraszyk: The motivation for this name was the system API `ic0.msg_cycles_available` which returns how many cycles have been transferred by the caller (and not yet accepted by the callee) so the new setting is a lower bound for the value of that system API at the very beginning of the call.

But I see your confusion: how does `minimum_incoming_call_cycles` or `minimum_canister_call_cycles` sound? (Focus on either incoming/outgoing and ingress/inter-canister, respectively, the combined `minimum_incoming_canister_call_cycles` is the most precise, but pretty long.)

rs/types/management_canister_types/src/lib.rs
```diff
@@ -1402,9 +1404,11 @@ pub struct DefiniteCanisterSettingsArgs {
     wasm_memory_limit: candid::Nat,
     wasm_memory_threshold: candid::Nat,
     environment_variables: Vec<EnvironmentVariable>,
+    minimum_msg_cycles_available: candid::Nat,
```
alin-at-dfinity: Personal opinion: I would put this somewhere next to the other fields that have to do with cycles, rather than tack it onto the end. Same with `SystemState` / `CanisterStateBits` / `CanisterStateBits` proto.

alin-at-dfinity (commented):
Only one more nitpick, otherwise LGTM.

packages/ic-management-canister-types/src/lib.rs
alin-at-dfinity: `minimum_canister_call_cycles` sounds more obvious.

I understand now where it comes from, but I suppose `msg_cycles_available` is both somewhat more specific and somewhat unfortunate too. (o:

daniel-wong-dfinity-org-twin (approved):
Approving only for the one Governance-owned file (under nervous_system).

packages/ic-management-canister-types/src/lib.rs
```diff
@@ -126,6 +126,15 @@ pub struct CanisterSettings {
     ///
     /// Default value: `5_000_000_000_000` (5 trillion cycles).
     pub reserved_cycles_limit: Option<Nat>,
+    /// Indicates the minimum number of cycles required for an incoming message
+    /// from another canister. Messages with fewer cycles are rejected with a
```
alin-at-dfinity:
```suggestion
    /// Indicates the minimum number of cycles required for an incoming call
    /// from another canister. Calls with fewer cycles are rejected with a
```

packages/ic-management-canister-types/src/lib.rs
```diff
@@ -182,6 +191,8 @@ pub struct DefiniteCanisterSettings {
     pub freezing_threshold: Nat,
     /// Upper limit on [`CanisterStatusResult::reserved_cycles`] of the canister.
     pub reserved_cycles_limit: Nat,
+    /// Minimum number of cycles required for an incoming message from another canister.
```
alin-at-dfinity:
```suggestion
    /// Minimum number of cycles required for an incoming call from another canister.
```

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -557,6 +557,13 @@ pub struct SystemState {
     /// fail if `reserved_balance + N` exceeds this limit if the limit is set.
     reserved_balance_limit: Option<Cycles>,
 
+    /// Minimum number of cycles required for an incoming message from another canister.
+    /// Messages with fewer cycles are rejected with a CanisterError at no cycles cost to the callee.
```
alin-at-dfinity:
```suggestion
    /// Minimum number of cycles required for an incoming call from another canister.
    /// Calls with fewer cycles are rejected with a CanisterError at no cycles cost to the callee.
```

rs/execution_environment/src/execution_environment.rs
```diff
@@ -2269,6 +2269,34 @@ impl ExecutionEnvironment {
                 );
             }
             CanisterMessageOrTask::Message(CanisterMessage::Request(request)) => {
+                let min_cycles = canister.system_state.minimum_incoming_canister_call_cycles;
+                if request.payment < min_cycles {
```
alin-at-dfinity: The doc comment (and common sense) say "incoming message/call from another canister". But this seems to apply uniformly to all canisters including self.

Am I missing something?

alin-at-dfinity (approved):
LGTM, modulo the self-calls comment.

rs/execution_environment/src/execution_environment.rs
mraszyk: Good point, thanks! https://github.com/dfinity/ic/pull/10398/commits/9be1e02141e1f192c9bdf7fd6c49454090044472

PR #10407, opened 2026-06-08
feat: subnet deletion in PocketIC

mraszyk:
This PR adds a PocketIC operation to delete a subnet:
- the subnet's `StateMachine` is properly dropped from the PocketIC state (including state directory removal);
- the registry is updated accordingly (including registry changes in the registry canister if the registry canister has been bootstrapped by PocketIC, i.e., not manually).

Note. This PR does not implement subnet deletion in DSM! Stale artifacts (such as stale streams and hanging unbounded-wait calls) might persist.

### DIFF
--- a/packages/pocket-ic/src/common/rest.rs
+++ b/packages/pocket-ic/src/common/rest.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use strum::IntoEnumIterator;
 use strum_macros::EnumIter;
 
 pub type InstanceId = usize;
@@ -482,6 +483,23 @@ pub enum SubnetKind {
     VerifiedApplication,
 }
 
+impl SubnetKind {
+    pub fn is_named(self) -> bool {
+        match self {
+            SubnetKind::NNS
+            | SubnetKind::SNS
+            | SubnetKind::II
+            | SubnetKind::Fiduciary
+            | SubnetKind::Bitcoin
+            | SubnetKind::TestThresholdKeys => true,
+            SubnetKind::Application
+            | SubnetKind::CloudEngine
+            | SubnetKind::System
+            | SubnetKind::VerifiedApplication => false,
+        }
+    }
+}
+
 /// This represents which named subnets the user wants to create, and how
 /// many of the general app/system subnets, which are indistinguishable.
 #[derive(Debug, Clone, Eq, Hash, PartialEq, Serialize, Deserialize, Default, JsonSchema)]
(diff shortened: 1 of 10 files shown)


packages/pocket-ic/src/common/rest.rs
```diff
@@ -805,25 +822,37 @@ impl SubnetStateConfig {
...
-            (self.test_threshold_keys.clone(), TestThresholdKeys),
-        ]
-        .into_iter()
-        .filter(|(mb, _)| mb.is_some())
-        .map(|(mb, kind)| {
-            let spec = mb.unwrap();
-            (kind, spec.get_state_path(), spec.get_instruction_config())
-        })
-        .collect()
+        [NNS, SNS, II, Fiduciary, Bitcoin, TestThresholdKeys]
```
alin-at-dfinity: Idea, feel free to ignore: Maybe a `NamedSubnet` enum and an `impl TryInto<NamedSubnet> for SubnetKind` would solve both filtering and ordering; plus ensure that this particular list can never fall out of sync with `SubnetKind::is_named()`.

packages/pocket-ic/tests/tests.rs
```diff
@@ -3631,3 +3631,185 @@ fn cloud_engine_default_effective_canister_id() {
...
+        )
+        .expect("inter-canister call via canister_1 should succeed");
+    assert_eq!(
+        Decode!(&reply, u32).unwrap(),
+        RejectCode::DestinationInvalid as u32
+    );
+}
+
+#[test]
+fn test_delete_default_effective_canister_id_subnet_fails() {
```
alin-at-dfinity, line 3711: Could you please add a doc (or inline) comment explaining why this is so?

I imagine it's a sanity thing (i.e. so as to not break PocketIC), but I actually have no idea.

rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -2776,6 +2877,26 @@ pub struct PocketIc {
...
+            Ok(sm) => {
+                sm.drop();
+                break;
+            }
+            Err(sm) => {
+                state_machine = Some(sm);
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            }
+        }
+        if start.elapsed() > std::time::Duration::from_secs(5 * 60) {
```
alin-at-dfinity: Random thought: Do we really want to wait 5 minutes per subnet?

(We could also set a deadline -- e.g. initially 5 minutes from "now" -- and pass that deadline as an argument, so e.g. `PocketIc::drop()` would only wait at most 5 minutes for all subnets.)

alin-at-dfinity (approved):

PR #10408, opened 2026-06-08
chore(P2P): Only increment error metric if transport infos of all registry versions are empty

eichhorl:
The peer manager maintains connections to all peers that are in the subnet record in any registry version between the "oldest version in use" and the latest locally available registry version.

The "oldest registry version in use" is determined by the CUP. In case of a subnet creation, this CUP is a registry CUP that contains the registry version just _before_ the subnet was created (see [here](https://sourcegraph.com/r/github.com/dfinity/ic@8cf338f5ef95432dc61233df889d98837d780466/-/blob/rs/registry/canister/src/mutations/do_create_subnet.rs?L63)).

This means that when the peer manager attempts to make connections to peers according to the "oldest registry version in use", the subnet doesn't actually exist yet, which then triggers the `empty_list_of_node_records` error. This is fine however, because it will find the correct peers in the subsequent registry version (which is guaranteed to exist, otherwise we would not have the registry CUP).

To improve the error signal of this metric, we change it such that it is only incremented if _all_ registry versions do not contain transport infos, meaning there are no peers shared with the new subnet topology.

### DIFF
--- a/rs/p2p/peer_manager/src/lib.rs
+++ b/rs/p2p/peer_manager/src/lib.rs
@@ -135,11 +135,6 @@ impl PeerManager {
                         self.log,
                         "Got transport infos but it's empty. Registry version {version}"
                     );
-                    self.metrics
-                        .topology_watcher_errors
-                        .with_label_values(&["empty_list_of_node_records"])
-                        .inc();
-
                     Vec::new()
                 }
                 Err(err) => {
@@ -192,6 +187,13 @@ impl PeerManager {
                     }
                 }
             }
+        }
+
+        if subnet_nodes.is_empty() {
+            self.metrics
+                .topology_watcher_errors
+                .with_label_values(&["empty_subnet_nodes"])
+                .inc();
         }
 
         SubnetTopology::new(


pierugo-dfinity:
This works well. Alternatively, I had thought of making an exception in `CatchUpPackage::get_oldest_registry_version_in_use` which would return the block's validation context's registry version instead of the summary's registry version _in case cup.height == 0_, because it is the registry version at which the CUP [was added to the registry](https://sourcegraph.com/r/github.com/dfinity/ic@8cf338f5ef95432dc61233df889d98837d780466/-/blob/rs/consensus/cup_utils/src/lib.rs?L187) (which is the same that added the subnet record). 

I agree that this would introduce some complexity which could be avoided with your simpler change. But implementing it this way also makes more sense functionality-wise since the consumers of `get_oldest_registry_version_in_use` are all interested in membership (P2P, orchestrator for detecting unassignment, firewall), so they all expect to see a `SubnetRecord` at that registry version (in practice not the firewall, but it still does conceptually).

WDYT?

alin-at-dfinity (approved):

alin-at-dfinity (commented):
Damn, only noticed Pierugo's comment after I hit Approve. My apologies, I thought I was just being helpful.

eichhorl:
> making an exception in `CatchUpPackage::get_oldest_registry_version_in_use` which would return the block's validation context's registry version instead of the summary's registry version _in case cup.height == 0_,

I think I prefer to keep it as is, because introducing the special case leaks information about the callers into the function and breaks the documented contract of returning "the oldest version in use". The other callers around unassignment and firewalls are already written defensively, and do not assume that "the oldest version in use" must necessarily relate to an "oldest membership to a subnet version" (which in my opinion is the problem this PR is trying to fix for P2P).

Pace222 (approved):
Alright, that makes sense to me! I agree that my workaround feels hacky

pierugo-dfinity (approved):
With my proper account :D

PR #10410, opened 2026-06-08
chore: refactor test_archiving_in_chunks_returns_disjoint_block_range_locations to not depend on canister performance

mraszyk:
This PR refactors the test `test_archiving_in_chunks_returns_disjoint_block_range_locations` to not depend on canister (WASM) performance (which changes in an upcoming PR bumping Rust compiler version to 1.95.0).

This goal is achieved by setting round instruction limits so that only a single canister message can be executed per round:
```
            let max_instructions_per_slice = std::cmp::max(
                self.config.max_instructions_per_slice,
                self.config.max_instructions_per_install_code_slice,
            );
            let round_instructions = as_round_instructions(self.config.max_instructions_per_round)
                - as_round_instructions(max_instructions_per_slice)
                + RoundInstructions::new(1);
```
in the scheduler yields `round_instructions` equal to 1 in the corresponding test (which means that the round instructions are certainly exhausted after a single message execution). There can still be multiple management canister messages per round, but this is not an issue for the test's robustness against canister (WASM) performance.

The only risk in this PR is its sensitivity to scheduler changes.

### DIFF
--- a/rs/ledger_suite/tests/sm-tests/src/lib.rs
+++ b/rs/ledger_suite/tests/sm-tests/src/lib.rs
@@ -5081,6 +5081,8 @@ pub mod archiving {
 
         // Install a ledger with a lot of initial balances
         let mut subnet_config = SubnetConfig::new(SubnetType::System, SubnetSecurity::None);
+        // Set max_instructions_per_round to max(max_instructions_per_slice, max_instructions_per_install_code_slice)
+        // to ensure that at most one canister message can be executed per round.
         subnet_config.scheduler_config.max_instructions_per_round = subnet_config
             .scheduler_config
             .max_instructions_per_slice
--- a/rs/ledger_suite/tests/sm-tests/src/lib.rs
+++ b/rs/ledger_suite/tests/sm-tests/src/lib.rs
@@ -4676,7 +4676,6 @@ pub mod archiving {
     use ic_ledger_canister_core::range_utils;
     use ic_ledger_suite_state_machine_helpers::{get_logs, icrc3_get_blocks};
     use ic_state_machine_tests::StateMachineBuilder;
-    use ic_types::NumInstructions;
     use ic_types::ingress::{IngressState, IngressStatus};
     use ic_types::messages::MessageId;
     use icp_ledger::{GetEncodedBlocksResult, QueryEncodedBlocksResponse};
--- a/rs/ledger_suite/tests/sm-tests/src/lib.rs
+++ b/rs/ledger_suite/tests/sm-tests/src/lib.rs
@@ -4676,6 +4676,7 @@ pub mod archiving {
     use ic_ledger_canister_core::range_utils;
     use ic_ledger_suite_state_machine_helpers::{get_logs, icrc3_get_blocks};
     use ic_state_machine_tests::StateMachineBuilder;
+    use ic_types::NumInstructions;
     use ic_types::ingress::{IngressState, IngressStatus};
     use ic_types::messages::MessageId;
     use icp_ledger::{GetEncodedBlocksResult, QueryEncodedBlocksResponse};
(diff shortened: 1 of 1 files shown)


rs/ledger_suite/tests/sm-tests/src/lib.rs
```diff
@@ -5134,16 +5149,13 @@ pub mod archiving {
...
-            1,
-        );
-        assert!(
-            !get_blocks_res.blocks.is_empty(),
-            "archive should contain at least one block"
-        );
+        let archive_canister_id =
+            CanisterId::unchecked_from_principal(PrincipalId::from(*archive_id));
+        let mut get_blocks_res = archive_get_blocks_fn(&env, archive_canister_id, 0, 1);
+        while get_blocks_res.blocks.is_empty() {
```
mraszyk, line 5147: it takes (a few) more round(s) to actually archive the block after the archive is created now that we only process one canister message per round

rs/ledger_suite/tests/sm-tests/src/lib.rs
```diff
@@ -4676,6 +4676,7 @@ pub mod archiving {
     use ic_ledger_canister_core::range_utils;
     use ic_ledger_suite_state_machine_helpers::{get_logs, icrc3_get_blocks};
     use ic_state_machine_tests::StateMachineBuilder;
+    use ic_types::NumInstructions;
```
mraszyk:
```suggestion
```

rs/ledger_suite/tests/sm-tests/src/lib.rs
```diff
@@ -5134,16 +5150,13 @@ pub mod archiving {
...
-            !get_blocks_res.blocks.is_empty(),
-            "archive should contain at least one block"
-        );
+        let archive_canister_id =
+            CanisterId::unchecked_from_principal(PrincipalId::from(*archive_id));
+        let mut get_blocks_res = archive_get_blocks_fn(&env, archive_canister_id, 0, 1);
+        while get_blocks_res.blocks.is_empty() {
+            env.tick();
+            get_blocks_res = archive_get_blocks_fn(&env, archive_canister_id, 0, 1);
+        }
```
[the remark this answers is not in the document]
mraszyk: I think the loop is fine as-is since there's a similar loop before waiting until an archive canister is created.

mbjorkqvist, line 5150: Since we add the loop here, we need to move the check on lines 5139-5146:
```
        // Verify that the ledger reports block `0` to be present only in the ledger
        let get_blocks_res = get_blocks_fn(&env, ledger_id, 0, 1);
        assert!(
            !ledger_reports_first_block_in_two_places(0, &get_blocks_res),
            "get_blocks_res: {get_blocks_res:?}"
        );
        // Verify that the ledger response contained no archive info.
        assert!(get_blocks_res.archived_ranges.is_empty());
```
 to after the loop, since we want to verify that:

- Block `0` already exists in the archive (when we query the archive directly).
- The ledger still reports block `0` as present only in the ledger, since the archiving operation has not yet completed in the ledger.

mbjorkqvist (approved):
Apart from the need to move the ledger check to after the newly-introduced loop, this LGTM, thanks @mraszyk!

rs/ledger_suite/tests/sm-tests/src/lib.rs
mraszyk: Good catch! Could you please double check my fix: https://github.com/dfinity/ic/pull/10410/commits/2841ba8688149e9cb8743acfb99adab4957a7a31

mbjorkqvist: Thanks, LGTM!

PR #10411, opened 2026-06-08
chore: Drop `SystemMetadata::full_topology`

alin-at-dfinity:
Since now all subnets (including engines) always "see" each other, there is no need for a separate `full_topology` to be certified by the NNS. We can fall back on the regular `topology` and drop `full_topology` altogether.

This is to only be merged into `master` once #10215 has been deployed, to avoid any forward compatibility issues (rolling back from a replica without `full_topology` to one depending on `full_topology` would break delegations).


schneiderstefan:
I don't think you can just drop it like that because of upgrades and downgrades. I initially thought about including this in my PR, but it's a bit trickier. During downgrades, the old version would expect the full topology to still be there. During upgrades, you also want to load the full topology still, otherwise you have no full topology in the first round after the upgrade.

alin-at-dfinity:
Closed in favor of #11216.

PR #10420, opened 2026-06-09
fix: Also reduce read_ahead for dm-* devices

alin-at-dfinity:
Looks like it was actually dm-2 (store-crypt) and/or dm-4 (store-shared--data, more likely) suffering from read amplification, not vda. They're probably all the same (virtual?) device underneath, which is why reads appear to be spiking on all 3 at once. But it's likely the shared data partition that is actually being pounded.

Reduce the read-ahead for all these devices similar to what was done for vda, just to be safe.

### DIFF
--- a/ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
+++ b/ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
@@ -1,10 +1,5 @@
 # Block device read_ahead configuration
 #
 
-block/dm-1/queue/read_ahead_kb = 128
-block/dm-2/queue/read_ahead_kb = 128
-block/dm-3/queue/read_ahead_kb = 128
-block/dm-4/queue/read_ahead_kb = 128
-block/dm-5/queue/read_ahead_kb = 128
-block/dm-6/queue/read_ahead_kb = 128
+block/dm-*/queue/read_ahead_kb = 128
 block/vda/queue/read_ahead_kb = 128
--- a/ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
+++ b/ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
@@ -1,4 +1,10 @@
 # Block device read_ahead configuration
 #
 
+block/dm-1/queue/read_ahead_kb = 128
+block/dm-2/queue/read_ahead_kb = 128
+block/dm-3/queue/read_ahead_kb = 128
+block/dm-4/queue/read_ahead_kb = 128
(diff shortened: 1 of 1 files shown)


ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
```diff
@@ -1,4 +1,10 @@
 # Block device read_ahead configuration
 #
 
+block/dm-1/queue/read_ahead_kb = 128
```
Bownairo: We should be able to glob to catch all of these, but I don’t mind being explicit and listing them out. It’s likely if these crypt devices have the issue, it’s the same for var, too, but as you said, data is probably the one causing most of the issues for us.

Bownairo (approved):

ic-os/components/guestos/misc/sysfs.d/read_ahead.conf
alin-at-dfinity: As commented elsewhere, this was pure copy-pasta on my part. I had no clue what was possible.

Do you mean I could just write something like
```suggestion
block/dm-*/queue/read_ahead_kb = 128
```
?

Bownairo: Exactly (https://manpages.ubuntu.com/manpages/questing/man5/sysfs.conf.5.html).

But I’m not sure, maybe we just lower it for data? What I can’t understand is, why would they make this change unless they expected it to be safe for most users? If it’s on us because of our large random reads, maybe it _should_ only be data?

I think either is safe regardless, and we can revisit with less urgency.

alin-at-dfinity: Technically we should only lower it for `store-shared--data` (`dm-4`), but I messed it up once before, so I'd rather be safe than sorry. Drives other than `data` don't see a lot of reads anyway, so it won't make any difference for them.

> What I can’t understand is, why would they make this change unless they expected it to be safe for most users?

I suppose it is an improvement for most use cases. But our use case goes entirely against it, by reading exactly every 10th MB (0, 10, 20, ...); then switching to the next 10th (1, 11, 21, ...). And because on this particular subnet (only) the state is significantly larger than the RAM, by the end of every iteration we will have read 80% of the state and already invalidated what we read at the beginning of the iteration; so we have to start reading everything back again.

We also discussed hashing contiguous 1 MB chunks instead of every 10th. In that case, read-ahead would work in our favor. But it's not a trivial change. And no one got started on it yet, anyway.

Switched to a wildcard.

PR #10426, opened 2026-06-10
feat: [DSM-142] Improve `process_batch()` phase duration coverage

alin-at-dfinity:
Following the introduction of structured `CanisterStates` (hot vs cold canisters), almost half of the remaining DSM overhead is not covered by `mr_process_batch_phase_duration_seconds`, so it can only be bundled under "other" (the difference to `mr_process_batch_duration_seconds`).

Explicitly track `read_registry()` duration (including blockmaker metrics updates, likely insignificant); and schedule/queues garbage collection plus (heavy) `canister_state_bytes` computation.

### DIFF
--- a/rs/messaging/src/state_machine.rs
+++ b/rs/messaging/src/state_machine.rs
@@ -15,7 +15,6 @@ use ic_registry_subnet_features::SubnetFeatures;
 use ic_replicated_state::{NetworkTopology, ReplicatedState};
 use ic_types::batch::{Batch, BatchContent};
 use ic_types::{ExecutionRound, NumBytes, SubnetId};
-use std::time::Instant;
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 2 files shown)


rs/messaging/src/state_machine.rs
```diff
@@ -69,15 +68,6 @@ impl StateMachineImpl {
...
     }
 
-    /// Adds an observation to the `METRIC_PROCESS_BATCH_PHASE_DURATION`
-    /// histogram for the given phase.
-    fn observe_phase_duration(&self, phase: &str, since: &Instant) {
-        self.metrics
-            .process_batch_phase_duration
-            .with_label_values(&[phase])
-            .observe(since.elapsed().as_secs_f64());
-    }
```
eichhorl, line 79: Does this mean `process_batch_phase_duration` can be private now?

eichhorl (approved):

rs/messaging/src/state_machine.rs
alin-at-dfinity: It does. I also made most of the other metrics private too.

PR #10427, opened 2026-06-10
chore: migration canister rejects canister migration from/to cloud engine

mraszyk:
This PR extends the migration canister to reject canister migration requests in which the migrated or replaced canister is on a cloud engine. Currently, such a canister migration is rejected because xnet calls to a cloud engine are not allowed. This PR makes the requirement explicit so that canister migration from/to a cloud engine keeps being rejected once xnet calls to a cloud engine are allowed.

### DIFF
--- a/rs/migration_canister/src/validation.rs
+++ b/rs/migration_canister/src/validation.rs
@@ -77,21 +77,25 @@ pub async fn validate_request(
         return Err(ValidationError::SameSubnet(Reserved));
     }
 
-    // 2. Are the migrated and replaced canisters on cloud engine subnets?
-    // It is safe to perform this check before acquiring locks because the fact that
-    // neither the migrated nor the replaced canister is on a cloud engine subnet
-    // cannot change later.
-    let migrated_canister_subnet = get_subnet_for_canister(migrated_canister).await?;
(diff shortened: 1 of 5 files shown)


rs/migration_canister/src/validation.rs
```diff
@@ -72,21 +72,39 @@ pub async fn validate_request(
...
+    if get_subnet(migrated_canister_subnet).await? == SubnetType::CloudEngine {
+        return Err(ValidationError::CloudEngineSubnet {
+            subnet: migrated_canister_subnet,
+        });
+    }
+    if get_subnet(replaced_canister_subnet).await? == SubnetType::CloudEngine {
+        return Err(ValidationError::CloudEngineSubnet {
+            subnet: replaced_canister_subnet,
+        });
+    }
```
mraszyk: These checks are performed before canister_status calls (so that we get `ValidationError::CloudEngineSubnet`), but this means also before locks are acquired (locks are acquired after canister_status checks so that an attacker cannot keep the locks acquired and prevent legit users from proceeding). Hence, they would have to be repeated after the locks are acquired. Alternatively, we can wait with this PR until xnet to engines is allowed and move the checks after canister_status calls. CC @schneiderstefan

schneiderstefan (approved):

PR #10429, opened 2026-06-10
fix(NODE-1958): Make cloud GuestOS great again

blind-oracle:
It seems that Ubuntu 26 upgrade changed the way the network is initialized by `systemd` which breaks our intermediate DHCP step. It's hard to debug especially in combination with SELinux policies.

* Use `dhcpcd` instead of `systemd-networkd` to obtain a temporary DHCP address. This avoids problems with new Ubuntu 26 init order (`systemd-networkd` is started now by `systemd` early it seems and if we launch a 2nd instance - it falls apart).
* Remove `console=tty0` from kernel boot args. It was added to support logging on serial-port-less clouds, but the problem is that when both `ttyX` and `ttySX` are present - the kernel makes the normal (graphical / ttyX) console the main one (regardless of their order) and serial console doesn't get any recovery shell or proper logging. We can try later to improve that somehow to get the best of the both worlds, though it's not easy. Also see https://github.com/systemd/systemd/issues/9899 about that.
* Alter the `icos_logging` so that it logs to both `stderr` and `journald` by default to ease debugging
* Remove timeout for `init-config` service since it anyway is critical and it makes not much sense to kill it if it hangs around longer
* Add timeouts to cloud-related HTTP calls, add more verbose logging

### DIFF
--- a/rs/ic_os/logging/src/lib.rs
+++ b/rs/ic_os/logging/src/lib.rs
@@ -7,8 +7,9 @@ use tracing_subscriber::{
     Layer,
     filter::LevelFilter,
     fmt::{
-        FmtContext, format,
+        FmtContext,
         format::{FormatEvent, FormatFields, Writer},
+        writer::MakeWriterExt,
     },
     layer::SubscriberExt,
     registry::LookupSpan,
@@ -65,22 +66,27 @@ fn syslog_identifier() -> String {
     syslog_identifier_from_arg0(arg0.as_deref())
 }
 
-/// Initialize tracing, using journald if available and falling back to stderr.
+/// Initialize tracing, using journald+stderr if journald is available and falling back to just stderr.
+/// Logs with TRACE level.
 pub fn init_logging() {
     init_logging_with_level(Level::TRACE);
 }
 
-/// Initialize tracing, using journald if available and falling back to stderr.
+/// Initialize tracing, using journald+stderr if journald is available and falling back to just stderr.
 pub fn init_logging_with_level(max_level: Level) {
     let level_filter = LevelFilter::from_level(max_level);
 
     match tracing_journald::layer() {
         Ok(layer) => tracing_subscriber::registry()
             .with(layer.with_filter(level_filter))
+            .with(
+                tracing_subscriber::fmt::layer()
+                    .with_writer(std::io::stderr.with_max_level(max_level)),
+            )
             .init(),
         Err(_) => tracing_subscriber::fmt()
             .with_writer(std::io::stderr)
-            .with_max_level(max_level)
+            .with_max_level(level_filter)
             .init(),
     }
 }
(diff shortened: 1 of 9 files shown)


rs/ic_os/logging/src/lib.rs
```diff
@@ -65,22 +65,28 @@ fn syslog_identifier() -> String {
...
 
-/// Initialize tracing, using journald if available and falling back to stderr.
+/// Initialize tracing, using journald+stderr if journald is available and falling back to just stderr.
 pub fn init_logging_with_level(max_level: Level) {
     let level_filter = LevelFilter::from_level(max_level);
 
     match tracing_journald::layer() {
         Ok(layer) => tracing_subscriber::registry()
             .with(layer.with_filter(level_filter))
+            .with(
```
frankdavid, line 81: Do you want to do it here instead of configuring the services with `journal+console` in systemd?

blind-oracle: Hard to say what's best. It's a CLI app and I think if e.g. I run it manually I'd expect it to output something to the console. Maybe it should just write to stderr (and not journald) and systemd should do the job of logging it..

Bownairo:
> Remove `console=tty0` from kernel boot args. It was added to support logging on serial-port-less clouds, but the problem is that when both `ttyX` and `ttySX` are present - the kernel makes the normal (graphical / ttyX) console the main one (regardless of their order) and serial console doesn't get any recovery shell or proper logging. We can try later to improve that somehow to get the best of the both worlds, though it's not easy.

~Does this mean the clouds still don't have output, and the arg essentially did nothing?~
Edit: Ah no, for clouds that have both, but connect the serial console to the UI, we would see nothing. Right?

In a few places, we use [output-wrapper.sh](https://github.com/dfinity/ic/blob/abf88f2283940c12b2974b0fd19ceb2bf485ef8f/ic-os/components/misc/output-wrapper.sh) to throw some extra logs around. Maybe useful on some of these units? (Even if no startup logs.)

Bownairo (approved):

blind-oracle:
> > Remove `console=tty0` from kernel boot args. It was added to support logging on serial-port-less clouds, but the problem is that when both `ttyX` and `ttySX` are present - the kernel makes the normal (graphical / ttyX) console the main one (regardless of their order) and serial console doesn't get any recovery shell or proper logging. We can try later to improve that somehow to get the best of the both worlds, though it's not easy.
> 
> ~Does this mean the clouds still don't have output, and the arg essentially did nothing?~ Edit: Ah no, for clouds that have both, but connect the serial console to the UI, we would see nothing. Right?
> 
> In a few places, we use [output-wrapper.sh](https://github.com/dfinity/ic/blob/abf88f2283940c12b2974b0fd19ceb2bf485ef8f/ic-os/components/misc/output-wrapper.sh) to throw some extra logs around. Maybe useful on some of these units? (Even if no startup logs.)

We have two kinda types of clouds - one with *only* serial console (major ones like AWS/GCP/Azure) and one with *only* graphical one (some Pakistani OpenStack etc). Initially we only worked with serial ones, but later this Pakistani thing arose and we had to have graphical console. There are probably clouds with both consoles, but I haven't seen'em.

And since kernel prefers graphical one - we only see limited boot logs (kernel kmsg & up to a certain point in systemd init) on the clouds with a serial console.

PR #10431, opened 2026-06-10
feat: allow engine-controller canister to update cloud engines directly

NikolaMilosa:
The users should be able to control their own settings for the engines they use. For now we are scoping that to only `subnet_admins` and `replica_version_id` but in the future we might want to allow more things.

### DIFF
--- a/rs/engine_controller/canister/canister.rs
+++ b/rs/engine_controller/canister/canister.rs
@@ -7,7 +7,8 @@ use candid::Principal;
 use ic_base_types::{NodeId, PrincipalId, SubnetId};
 use ic_cdk::{api::msg_caller, call::Call, init, post_upgrade, println, update};
 use ic_engine_controller::{
-    CreateEngineArgs, DeleteEngineArgs, EngineControllerInitArgs, NewSubnet,
+    CreateEngineArgs, DeleteEngineArgs, DeployGuestosToAllSubnetNodesPayload,
+    EngineControllerInitArgs, NewSubnet, UpdateSubnetPayload,
 };
 use ic_nns_constants::REGISTRY_CANISTER_ID;
 use ic_protobuf::registry::subnet::v1::SubnetFeatures;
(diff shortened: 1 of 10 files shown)


pietrodimarco-dfinity (approved):

rs/registry/canister/src/mutations/do_update_subnet.rs
```diff
@@ -1703,6 +1718,93 @@ mod tests {
...
         payload.subnet_admins = Some(vec![*TEST_USER1_PRINCIPAL]);
 
-        registry.do_update_subnet(payload);
+        registry.do_update_subnet(GOVERNANCE_CANISTER_ID.get(), payload);
+    }
+
+    /// Builds a registry that already has a `CloudEngine` subnet, suitable
+    /// for exercising the engine-controller permission checks on
+    /// `do_update_subnet`.
+    fn make_registry_with_cloud_engine_subnet() -> (Registry, SubnetId) {
```
r-birkner, line 1836: In the test helpers, we already have such a method: https://github.com/dfinity/ic/blob/master/rs/registry/canister/tests/common/test_helpers.rs#L216
You might want to use that.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -48,6 +67,109 @@ impl Registry {
...
+    use ic_protobuf::registry::subnet::v1::{
+        CanisterCyclesCostSchedule, SubnetType as SubnetTypePb,
+    };
+    use ic_registry_subnet_type::SubnetType;
+    use ic_test_utilities_types::ids::subnet_test_id;
+    use ic_types::ReplicaVersion;
+    use maplit::btreemap;
+
+    /// Creates a registry with a single subnet of the given `subnet_type`.
+    fn make_registry_with_subnet(subnet_type: SubnetType) -> (Registry, SubnetId) {
```
r-birkner: Here, you could probably use the test_helper method:
https://github.com/dfinity/ic/blob/master/rs/registry/canister/tests/common/test_helpers.rs#L216 it creates a registry with a system and a cloud engine subnet

rs/engine_controller/canister/canister.rs
```diff
@@ -186,6 +187,180 @@ async fn delete_engine(args: DeleteEngineArgs) -> Result<(), String> {
...
+    let mut admins = admins;
+    if !admins.contains(&super_admin) {
+        admins.push(super_admin);
+    }
+    admins
+}
+
+/// Proxies to the registry's `update_subnet` endpoint. Only `subnet_admins`
+/// may be updated through this path; every other field must be left at its
+/// default value (`None` / `false`) or the call is rejected.
```
r-birkner: Do I remember correctly, that when something is unset in the UpdateSubnetPayload, it just means that the value in the registry is not changed, right?

NikolaMilosa: Yes, they basically update just what is provided.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
NikolaMilosa: But those are integration test level things? I will add the same thing in the shared crate but I cannot use those in unit tests here

rs/registry/canister/src/mutations/do_update_subnet.rs
NikolaMilosa: Those are not accessible here because they are for integration tests 😟

rs/registry/canister/src/mutations/do_update_subnet.rs
```diff
@@ -22,17 +23,36 @@ use std::collections::HashSet;
...
+/// This method is called by:
+///   * the governance canister, after a proposal for updating a subnet has
+///     been accepted (no scope restriction); and
+///   * the engine controller canister, which may only target CloudEngine
+///     subnets and is further restricted to a small subset of fields (see
+///     [`ensure_engine_controller_payload_scope`]).
 impl Registry {
-    pub fn do_update_subnet(&mut self, payload: UpdateSubnetPayload) {
-        println!("{}do_update_subnet: {:?}", LOG_PREFIX, payload);
+    pub fn do_update_subnet(&mut self, caller: PrincipalId, payload: UpdateSubnetPayload) {
```
schneiderstefan, line 33: Is there rate limiting in place, ideally in the registry canister itself? It could be dangerous if the end user is allowed to make changes and can do as many as they want.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -15,14 +16,32 @@ use serde::Serialize;
 impl Registry {
     pub fn do_deploy_guestos_to_all_subnet_nodes(
```
schneiderstefan, line 17: Note that any pair of blessed versions might be incompatible at any time. So it's easy to shoot yourself in the foot if we give too many options to the end users.

schneiderstefan (commented):
Implementation-wise LGTM, but some high-level concerns.

rs/registry/canister/src/mutations/do_update_subnet.rs
NikolaMilosa: For transparency: at the moment we aren't even allowing this call, we need this for something completely different that isn't user facing. So users for now cannot do anything with this specifically and won't be able for some time still until we figure out everything. And we won't be constantly changing any of these values. I expect that per engine it could be 2-3 times per month. Basically we want to be able to halt a subnet if an engine payment is overdue. This will eventually go into this canister (the whole accounting logic), but in the interest of time we have to take it bit by bit

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
NikolaMilosa: Absolutely! Again this is not exposed to the end users. This is still heavily controlled by us. I highly doubt that any user will ever care about that so this will most likely go away.

pietrodimarco-dfinity (approved):

r-birkner (approved):

bjoernek (approved):

PR #10432, opened 2026-06-10
feat: Introduce EngineNotAllowed signals

schneiderstefan:
This commit introduces XNet reject signals that will fire when messages are sent to/from engines that are not allowed. The current PR only introduces the ability to read these signals, without writing them themselves. The actual implementation of XNet with engines will come in a follow-up, and will start sending these signals.


rs/canonical_state/src/encoding/tests/compatibility.rs
```diff
@@ -166,6 +166,92 @@ fn canonical_encoding_stream_header() {
...
+/// ```
+///
+/// Relative to the `V25` encoding, the reject signals map grows from `A7`
+/// (map(7)) to `A8` (map(8)) and gains the trailing field:
+///
+/// ```text
+///    07    # field_index(RejectSignals::engine_not_allowed_deltas)
+///    81    # array(1)
+///       08 # unsigned(8)   (`signals_end` 256 - 248)
+/// ```
```
alin-at-dfinity: Nit: All the other test descriptions include the full binary representation.

rs/canonical_state/tests/compatibility.rs
alin-at-dfinity, line 146: You should also import `StreamHeaderV25` and add it here.

And import `StreamHeader` as `StreamHeaderV26` instead of `StreamHeader21` now.

rs/canonical_state/src/encoding/tests/compatibility.rs
```diff
@@ -166,6 +166,92 @@ fn canonical_encoding_stream_header() {
...
+        canister_stopping_deltas: vec![3],
+        queue_full_deltas: vec![5],
+        out_of_memory_deltas: vec![4],
+        unknown_deltas: vec![1],
+    };
+
+    // The V26 `RejectSignals` with no engine signals encodes identically to V25.
+    let new_bytes = serde_cbor::ser::to_vec_packed(&new).unwrap();
+    let old_bytes = serde_cbor::ser::to_vec_packed(&old).unwrap();
+    assert_eq!(new_bytes, old_bytes);
```
alin-at-dfinity: The existing `canonical_encoding_stream_header()` already tests that all supported certification versions encode to the exact same bytes.

rs/canonical_state/src/encoding/tests/compatibility.rs
```diff
@@ -166,6 +166,92 @@ fn canonical_encoding_stream_header() {
...
+
+    // A V25 encoding still decodes into the new type, with no engine signals.
+    let decoded: types::RejectSignals = serde_cbor::de::from_slice(&old_bytes).unwrap();
+    assert!(decoded.engine_not_allowed_deltas.is_empty());
```
alin-at-dfinity: As for the decoding part, that is covered by `stream_header_roundtrip_encoding()` in `rs/canonical_state/tests.compatibility.rs`. (If you also add `StreamHeaderV25` to the list of `StreamHeader` encodings.)

I.e. this test case as a whole can be dropped. It's not in the right location anyway. (o;

rs/canonical_state/src/encoding/old_types.rs
```diff
@@ -513,3 +513,22 @@ impl TryFrom<StreamHeaderV19> for ic_types::xnet::StreamHeader {
         ))
     }
 }
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct RejectSignalsV25 {
```
alin-at-dfinity, line 524: You'll also have to copy over the `From` / `TryFrom` implementations (without your current changes) for `RejectSignalsV25`. Otherwise the type cannot be used by encoding / decoding tests.

rs/canonical_state/src/encoding/tests/conversion.rs
```diff
@@ -53,6 +53,38 @@ fn roundtrip_conversion_stream_header() {
     }
 }
 
+/// `EngineNotAllowed` reject signals must not be *encoded* below V26, but the decode path is
+/// version-agnostic and must read them back. `MAX_SUPPORTED_CERTIFICATION_VERSION` is V26.
+#[test]
+fn roundtrip_conversion_stream_header_with_engine_not_allowed_signal() {
```
alin-at-dfinity: Same as with `reject_signals_v26_compatible_with_v25()`, this should be covered by `stream_header_roundtrip_encoding()` (if `STREAM_HEADER_ENCODINGS` is updated accordingly).

So this test can be dropped.

rs/state_manager/src/tree_hash.rs
```diff
@@ -392,6 +392,9 @@ mod tests {
             "07797459A2F82D6F64628C0668C5BDB7F83447680DDB178208A40C2256409E8D",
             "F80B2659485C03F68935F214E4CB5D8CCAC02913DCA88E913C4B497F2120DA50",
             "416172D9AFD573236F1CDE2459756736EEB25028D64FB8D7192AAF33AFC0DA6F",
+            // V26 adds `EngineNotAllowed` reject signals; absent any such signal the encoding is
+            // identical to V25, so the hash matches.
+            "416172D9AFD573236F1CDE2459756736EEB25028D64FB8D7192AAF33AFC0DA6F",
```
alin-at-dfinity: You should instead add a `if version >= V26 { stream.push_reject_signal(EngineNotAllowed) }`. Then the hash will change.

rs/test_utilities/state/src/lib.rs
```diff
@@ -951,7 +952,15 @@ prop_compose! {
...
+            // `EngineNotAllowed` is only encodable from V26; these streams are certified at
+            // `CURRENT_CERTIFICATION_VERSION`.
+            RejectReason::all()
+                .into_iter()
+                .filter(|reason| {
+                    *reason != RejectReason::EngineNotAllowed
+                        || CURRENT_CERTIFICATION_VERSION >= CertificationVersion::V26
+                })
+                .collect(),
         )
```
alin-at-dfinity: Because `arb_stream()` is used by `rs/canonical_state/tests/compatibility.rs`, it should be changed to take a `CertificationVersion` argument and decide whether to include `EngineNotAllowed` based on that.

rs/canonical_state/src/encoding/old_types.rs
schneiderstefan: Done (try_from_deltas_v25), but please double check if that's what you had in mind.

rs/canonical_state/src/encoding/tests/compatibility.rs
schneiderstefan: Done.

rs/canonical_state/src/encoding/tests/compatibility.rs
schneiderstefan: Test removed.

rs/canonical_state/src/encoding/tests/compatibility.rs
schneiderstefan: Test removed.

rs/canonical_state/src/encoding/tests/conversion.rs
schneiderstefan: Test removed.

rs/state_manager/src/tree_hash.rs
schneiderstefan: Done.

rs/canonical_state/tests/compatibility.rs
schneiderstefan: Done.

rs/test_utilities/state/src/lib.rs
schneiderstefan: Done. Could you please double check if I pass in the correct version at all call sites?

rs/canonical_state/src/encoding/old_types.rs
```diff
@@ -532,3 +537,192 @@ pub struct RejectSignalsV25 {
...
+
+        RejectSignalsV25 {
+            canister_migrating_deltas: deltas_for(RejectReason::CanisterMigrating),
+            canister_not_found_deltas: deltas_for(RejectReason::CanisterNotFound),
+            canister_stopped_deltas: deltas_for(RejectReason::CanisterStopped),
+            canister_stopping_deltas: deltas_for(RejectReason::CanisterStopping),
+            queue_full_deltas: deltas_for(RejectReason::QueueFull),
+            out_of_memory_deltas: deltas_for(RejectReason::OutOfMemory),
+            unknown_deltas: deltas_for(RejectReason::Unknown),
+        }
```
alin-at-dfinity, line 601: As a sanity check, we should have an explicit assert that we are never asked to encode `EngineNotAllowed` signals. Something like:

```rust
assert_eq(deltas_for(RejectReason::EngineNotAllowed), vec![]);
```

rs/canonical_state/tests/compatibility.rs
```diff
@@ -103,9 +103,18 @@ pub(crate) fn arb_invalid_versioned_stream_header(
 ) -> impl Strategy<Value = (StreamHeader, RangeInclusive<CertificationVersion>)> {
     prop_oneof![
         // Encoding a stream header with reject signal flavors other than `CanisterMigrating`
-        // before certification version 19 should panic.
+        // before certification version 19 should panic. `EngineNotAllowed` is excluded as it is
+        // only encodable from V26.
```
alin-at-dfinity, line 107: You could have left it in. It's supposed to be invalid anyway.

No need to change it back, though.

alin-at-dfinity (approved):
LGTM, thank you.

rs/canonical_state/src/encoding/old_types.rs
schneiderstefan: Added. Everything still passes.

PR #10439, opened 2026-06-11
feat: hoist dosfstools and mtools out of Dockerfile

nmattia:
This moves two further IC-OS tools, dosfstools and mtools, out of the Dockerfile and into the Bazel build. This is similar to what we did with e2fsprogs. This makes the build more portable.

See also: https://github.com/dfinity/ic/pull/10369


basvandijk (approved):
Nice. LGTM!

Bownairo (approved):
👌🎊

PR #10440, opened 2026-06-11
chore(Core): Adjust instruction cost for accessing pages. 

michael-weigelt:
A follow-up PR will refactor and unify the logic and constants. 

DO NOT MERGE.


rs/memory_tracker/src/deterministic.rs
```diff
@@ -353,7 +355,7 @@ impl DeterministicMemoryTracker {
         }
 
         // Charge 1 instruction per OS page accessed.
```
alin-at-dfinity: It's no longer 1 instruction.

rs/memory_tracker/src/deterministic.rs
```diff
@@ -364,7 +366,7 @@ impl DeterministicMemoryTracker {
         // Charge 1 instruction per OS page dirtied.
```
alin-at-dfinity: Ditto.

rs/memory_tracker/src/deterministic.rs
```diff
@@ -83,6 +83,8 @@ use crate::checksum;
 #[cfg(feature = "sigsegv_handler_checksum")]
 use std::ops::BitXor;
 
+const INSTRUCTION_OVERHEAD_PER_OS_PAGE: u64 = 5_000;
```
alin-at-dfinity: Can't you import `DEFAULT_DIRTY_PAGE_OVERHEAD` instead?

michael-weigelt: Instead of depending on the default constant (which could be overwritten in the config) I now thread through the config value instead.

michael-weigelt: 29588429f9

PR #10444, opened 2026-06-11
fix: disable canister log memory store feature

mraszyk:
Reverts https://github.com/dfinity/ic/pull/10160


schneiderstefan:
I think it would be clearer to follow if it mentioned that this is a revert of the commit that set it to true.

mraszyk:
> I think it would be clearer to follow if it mentioned that this is a revert of the commit that set it to true.

Updated the PR description; would you also like a code comment?

schneiderstefan (approved):
No code comment needed. I was mainly thinking of external reviewers, who would see both the enabling and the disabling as two separate PRs and might be confused.

PR #10448, opened 2026-06-11
feat: allow unlimited number of gen4 nodes

NikolaMilosa:
This pull request allows node providers to add more type4.x nodes (excluding type4.5 for now see #10393). 

This is needed to allow for provisioning on-demand nodes in the cloud and registering them with the network.

### DIFF
--- a/rs/registry/canister/src/mutations/node_management/do_add_node.rs
+++ b/rs/registry/canister/src/mutations/node_management/do_add_node.rs
@@ -136,7 +136,7 @@ impl Registry {
                 "{LOG_PREFIX}do_add_node: Node reward type is required."
             ))?;
 
-            // TODO(NNS1-XXXX): Re-enable the max_rewardable_nodes quota check for
+            // TODO(CLO-15): Re-enable the max_rewardable_nodes quota check for
             // type4.1-type4.4 once we no longer treat rewards of type4.5 as type1.1.
             // For now, node providers may deploy arbitrarily many nodes of these
             // types without being constrained by `max_rewardable_nodes`. Type4.5 is
@@ -1141,7 +1141,7 @@ mod tests {
 
     #[test]
     fn should_allow_arbitrarily_many_nodes_of_type4dot1_through_type4dot4() {
-        // TODO(NNS1-XXXX): Remove this test once we re-enable the
+        // TODO(CLO-15): Remove this test once we re-enable the
         // max_rewardable_nodes quota check for type4.1-type4.4. See related TODO
         // in `do_add_node_`.
         for node_reward_type in [
--- a/rs/registry/canister/src/mutations/node_management/do_add_node.rs
+++ b/rs/registry/canister/src/mutations/node_management/do_add_node.rs
@@ -136,12 +136,15 @@ impl Registry {
                 "{LOG_PREFIX}do_add_node: Node reward type is required."
             ))?;
 
-            // TODO(CLO-15): Re-enable the max_rewardable_nodes quota check for
-            // type4.1-type4.4 once we no longer treat rewards of type4.5 as type1.1.
-            // For now, node providers may deploy arbitrarily many nodes of these
-            // types without being constrained by `max_rewardable_nodes`. Type4.5 is
-            // explicitly excluded from this exemption.
-            let bypass_max_rewardable_nodes_check = matches!(
+            // TODO(CLO-15): Remove the type4.1-type4.4 special case below once
+            // we no longer treat rewards of type4.5 as type1.1. For now, node
+            // providers must be able to deploy arbitrarily many nodes of these
+            // types without being constrained by `max_rewardable_nodes`. Rather
+            // than branching around the check, we model this as an effectively
+            // unbounded quota so the same code path is exercised for every
+            // reward type. Type4.5 is explicitly excluded from this exemption.
+            const EXCESSIVE_NUMBER_OF_TYPE_4_NODES: u32 = 1_000;
+            let is_unbounded_type4 = matches!(
                 node_reward_type,
                 NodeRewardType::Type4dot1
                     | NodeRewardType::Type4dot2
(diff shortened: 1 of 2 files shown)


rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,27 +136,42 @@ impl Registry {
...
+            // For now, node providers may deploy arbitrarily many nodes of these
+            // types without being constrained by `max_rewardable_nodes`. Type4.5 is
+            // explicitly excluded from this exemption.
+            let bypass_max_rewardable_nodes_check = matches!(
+                node_reward_type,
+                NodeRewardType::Type4dot1
+                    | NodeRewardType::Type4dot2
+                    | NodeRewardType::Type4dot3
+                    | NodeRewardType::Type4dot4
+            );
```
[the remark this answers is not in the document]
NikolaMilosa: We don't want to do it for type4 nodes. They were a one-off and won't be used for this feature.

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,27 +136,42 @@ impl Registry {
...
+
+            if !bypass_max_rewardable_nodes_check {
+                let max_rewardable_nodes_same_type = *node_operator_record
+                    .max_rewardable_nodes
+                    .get(&(node_reward_type.to_string()))
+                    .ok_or(format!("{LOG_PREFIX}do_add_node: Node Operator does not have rewardable nodes for {node_reward_type}"))?;
```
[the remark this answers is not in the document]
NikolaMilosa: On the reward canister side, these aren't considered anyways. So if they exist they are skipped.

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,27 +136,42 @@ impl Registry {
...
-                    Number of nodes in the registry with {node_reward_type} type = {num_in_registry_same_type},\
-                    Number of removed nodes with same IP and same type = {num_removed_same_ip_same_type},\
-                    {node_reward_type} quota = {max_rewardable_nodes_same_type}"
-                ));
+            // TODO(CLO-15): Re-enable the max_rewardable_nodes quota check for
+            // type4.1-type4.4 once we no longer treat rewards of type4.5 as type1.1.
+            // For now, node providers may deploy arbitrarily many nodes of these
+            // types without being constrained by `max_rewardable_nodes`. Type4.5 is
+            // explicitly excluded from this exemption.
+            let bypass_max_rewardable_nodes_check = matches!(
```
daniel-wong-dfinity-org-twin: Do as you see fit.

Alternatively, you can just say that there must be < EXCESSIVE_NUMBER_OF_TYPE_4_NODES type 4.x nodes. That way, they do not need to be treated specially.

When it is possible to not treat something specially, it is better to treat it like everything else. Since the improvement here is marginal, I leave it up to you.

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,27 +136,42 @@ impl Registry {
...
+
+                let num_in_registry_same_type = get_node_operator_nodes(self, caller_id)
+                    .into_iter()
+                    .filter_map(|node| node.node_reward_type)
+                    .filter(|t| t == &(node_reward_type as i32))
+                    .count() as u32;
+
+                // Validate node operator's max_rewardable_nodes quota
+                if max_rewardable_nodes_same_type
+                    <= num_in_registry_same_type.saturating_sub(num_removed_same_ip_same_type)
```
daniel-wong-dfinity-org-twin: I know it was like this before already, but why `<=` ?

Suggestion:

```suggestion
                let num_remaining_nodes = num_in_registry_same_type.saturating_sub(num_removed_same_ip_same_type);
                if num_remaining_nodes > max_rewardable_nodes_same_type {
```

Note that my suggestion is not equivalent, because `>` is used. Seems like the original code has an off by one bug? Equal to max should be fine, right??

The most important change in my suggestion is swapping the order of operands, putting the reference value on the right hand side of the comparison operator, because we pretty much always do it that way.

The dummy variable in my suggestion is less important. It helps because multi-line `if` expressions should be avoided (and in this case, it can be avoided pretty easily), but this is not something that we particularly follow; hence, lower importance.

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -1124,6 +1139,67 @@ mod tests {
...
+                registry_add_node_operator_for_node(&mut registry, node_ids[0], btreemap! {});
+
+            // Adding several nodes of this type should all succeed even
+            // though no quota is configured.
+            for i in 0..3_u8 {
+                let (payload, _) = prepare_add_node_payload(10 + i, node_reward_type);
+                registry
+                    .do_add_node_(payload, node_operator_id, now_system_time())
+                    .unwrap_or_else(|e| {
+                        panic!("do_add_node_ failed for {node_reward_type} on iteration {i}: {e}")
```
daniel-wong-dfinity-org-twin, line 1194: Maybe, comment here that the next test shows that for other node reward types, this would fail. Otherwise, even if this test passes, it could be simply due to panic NEVER occurs, regardless of node reward type.

rs/registry/canister/unreleased_changelog.md
```diff
@@ -24,6 +24,11 @@ on the process that this file is part of, see
 
 ## Changed
 
+* Temporarily bypass the `max_rewardable_nodes` quota check in `add_node` for
+  node reward types `type4.1` through `type4.4`, allowing node providers to
+  register an arbitrary number of such nodes. `type4.5` is explicitly excluded
+  and still subject to the quota. This is a temporary measure until rewards of
+  `type4.5` are no longer treated as `type1.1` (see CLO-15).
```
daniel-wong-dfinity-org-twin: Can you maybe explain the motivation?

daniel-wong-dfinity-org-twin (commented):
Seems like, this could cause problems space problems in the Registry canister... I mean, an alternative is just to have a very large (but finite!) limit, right??

rs/registry/canister/src/mutations/node_management/do_add_node.rs
NikolaMilosa: I can address this, but this is the question for @pietrodimarco-dfinity, I assume that this isn't an array and is actually how many nodes you have vs how many you are allowed. If you have 5 and you are allowed 5 it is fine, right?

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,42 +149,43 @@ impl Registry {
...
+            let num_in_registry_same_type = get_node_operator_nodes(self, caller_id)
+                .into_iter()
+                .filter_map(|node| node.node_reward_type)
+                .filter(|t| t == &(node_reward_type as i32))
+                .count() as u32;
+
+            // Validate node operator's max_rewardable_nodes quota.
+            let num_remaining_nodes =
+                num_in_registry_same_type.saturating_sub(num_removed_same_ip_same_type);
+            if num_remaining_nodes >= max_rewardable_nodes_same_type {
```
daniel-wong-dfinity-org-twin, line 186: did you get an answer on whether == max is bad? Seems like the original code (where == max was considered bad) was buggy. If it is bad, I'd like to know why, but I'm not going to block on this, because at least we are not regressing; if it was a bug before, at least we are not introducing a NEW bug.

It is probably worth explaining _in comments_ why == max is considered bad, because that is highly unintuitive.

rs/registry/canister/unreleased_changelog.md
```diff
@@ -25,10 +25,17 @@ on the process that this file is part of, see
 ## Changed
 
 * Temporarily bypass the `max_rewardable_nodes` quota check in `add_node` for
-  node reward types `type4.1` through `type4.4`, allowing node providers to
-  register an arbitrary number of such nodes. `type4.5` is explicitly excluded
-  and still subject to the quota. This is a temporary measure until rewards of
-  `type4.5` are no longer treated as `type1.1` (see CLO-15).
+  node reward types `type4.1` through `type4.4` (implemented by treating their
+  effective quota as unbounded), allowing node providers to register an
+  arbitrary number of such nodes. `type4.5` is explicitly excluded and remains
```
daniel-wong-dfinity-org-twin: This was maybe the original intent, but we now have a cap.

rs/registry/canister/src/mutations/node_management/do_add_node.rs
```diff
@@ -136,42 +149,43 @@ impl Registry {
...
-            // TODO(CLO-15): Re-enable the max_rewardable_nodes quota check for
-            // type4.1-type4.4 once we no longer treat rewards of type4.5 as type1.1.
-            // For now, node providers may deploy arbitrarily many nodes of these
-            // types without being constrained by `max_rewardable_nodes`. Type4.5 is
+            // See `EXCESSIVE_NUMBER_OF_TYPE_4_NODES` at the top of this file for
+            // why type4.1-type4.4 are handled via a sentinel quota rather than
+            // the per-operator `max_rewardable_nodes` map. Type4.5 is
             // explicitly excluded from this exemption.
-            let bypass_max_rewardable_nodes_check = matches!(
+            let is_unbounded_type4 = matches!(
```
daniel-wong-dfinity-org-twin: "unbounded" is a misnomer now that we have EXCESSIVE_NUMBER_OF_TYPE_4_NODES.

Suggestion: is_permissionless_type_4_x. IDK. This might make people angry.

If this code is temporary anyway, then, don't sweat it too much.

daniel-wong-dfinity-org-twin (approved):

PR #10449, opened 2026-06-11
feat: reject update calls for delegations with queries-only permissions

aterga:
# Queries-only permissions on delegations

**This is the consolidated read-only-sessions feature** (the previously stacked alternatives #10447/#10452, which carried the restriction in certified `sender_info` attributes, were closed in favor of this more lightweight approach).

## What it does

Request delegations gain an optional `permissions` field, mirroring the existing `targets` field. It is modeled as a typed enum `DelegationPermissions` whose variants are encoded on the wire (and in the representation-independent hash) as text:

* `permissions` **absent** or **`"all"`** (`DelegationPermissions::All`) → the sender can execute all functions — queries, replicated queries, and updates (today's semantics; byte-for-byte identical hash for existing delegations without the field).
* **`"queries"`** (`DelegationPermissions::Queries`) → the sender can only execute queries: requests to `/call` endpoints (updates *and* replicated queries) authenticated through such a chain are rejected during ingress validation with the new `RequestValidationError::UpdateCallNotPermittedByDelegation`; query and `read_state` requests remain permitted.
* **Any other value** → because the field is a closed enum, an unsupported value fails to deserialize, so the request is **rejected when it is decoded** (before validation). There is no dedicated validator error for it.

A restriction anywhere in the chain applies to the whole chain (like `targets`, restrictions only accumulate). The matching interface-spec change (dfinity/developer-docs#292) models the field as the union `Queries | All | Unrestricted`, which this enum mirrors exactly.

## Why this can't be bypassed, and why it's backward compatible

* The field is part of the delegation's representation-independent hash, i.e. covered by the delegation signature. A dapp holding the session key **cannot strip the field** — doing so changes the hash and invalidates the signature.
* **Old replicas fail closed**: they parse the delegation without the unknown field (serde ignores unknown map keys), recompute the hash without it, and signature verification fails. A restricted delegation is unusable on un-upgraded replicas rather than silently unrestricted. Rollout: replicas first, issuers (Internet Identity) after.
* Delegations without the field are completely unaffected, so existing agents, dapps, and issuers need no changes. Agents only need to round-trip the new field once they want to carry restricted delegations; an agent that drops it fails closed.
* The enforcement point is `HttpRequestVerifier<SignedIngressContent>` — the same verifier used by the ingress manager for block validation, so a malicious boundary node or replica cannot smuggle a restricted update call into a block.

## Rationale

Enables issuers like Internet Identity to sign **read-only delegations**: sessions that can read on the user's behalf (query calls) but cannot change state (update calls), enforced by the protocol regardless of the client's cooperation. II sets this field from a "Read-only mode" checkbox in the authorize flow (II-side changes prepared separately).

Historical precedent: the interface spec already extended the delegation map with an optional restriction field once — the `senders` field (added Dec 2021, never implemented by the replica, removed in [interface-spec#246](https://github.com/dfinity/interface-spec/pull/246)). This PR follows the same formal pattern (`verify_delegations` accumulation), but implementation-first.

## Changes

* `ic-types`: typed `DelegationPermissions { Queries, All }` enum (serialized as `"queries"`/`"all"`); optional `Delegation.permissions` field included in the signed bytes (`hash_of_map` key `"permissions"`); chainable `with_targets`/`with_permissions` builder methods (replacing the former `new_with_targets`/`new_with_permissions` constructors so a delegation can carry both).
* `ic-validator`: `DelegationRestrictions` (targets + queries-only) threaded through delegation-chain validation; update-call rejection via `RequestValidationError::UpdateCallNotPermittedByDelegation`. The `/call` path uses the shared `validate_request_content` (no special-casing) and checks the restriction afterwards.
* `ic-validator-ingress-message`: the new error variant mirrored.
* `ic-validator-http-request-test-utils`: `delegate_to_with_permissions` chain-builder support.
* Tests: signed-bytes golden test for the new field; CBOR encoding + round-trip tests; a decode-rejection test (`delegation_permissions_rejects_unsupported_value`) covering unsupported values including case/whitespace variants; validator-level e2e tests (update-rejected / query-and-read_state-permitted / `"all"`-permitted / restriction-mid-chain / combined targets+permissions); and a system test (`requests_with_delegation_permissions` in `rs/tests/crypto/ingress_verification_test.rs`, tagged `long_test`) exercising the full HTTP path against a real replica.

## Caveats / follow-ups

* Calls to query methods submitted via the `/call` path (replicated queries) are rejected for `"queries"` delegations by design — the `"all"` vocabulary makes this explicit.
* Interface-spec change in dfinity/developer-docs#292; II-side issuance and agent (`@icp-sdk/core`) round-tripping of the field prepared/pending.

`cargo test -p ic-types -p ic-validator -p ic-validator-ingress-message` passes and clippy is clean; downstream consumers (`ic-http-endpoints-public`, `ic-ingress-manager`, `ic-canister-client`, `ic-state-machine-tests`) and both affected system-test binaries compile.


rs/validator/ingress_message/tests/validate_request.rs
```diff
@@ -2175,6 +2175,142 @@ mod sender_info {
     }
 }
 
+mod delegation_permissions {
+    use super::*;
+    use ic_validator_http_request_test_utils::DelegationChain;
+    use ic_validator_ingress_message::AuthenticationError::UnsupportedDelegationPermissions;
+
+    #[test]
+    fn should_reject_update_call_when_delegation_restricts_to_queries() {
```
eichhorl, line 2184: Maybe we could similarly extend system tests around ingress validation

eichhorl: I don't think this has been addressed

rs/types/types/src/messages/http.rs
```diff
@@ -544,6 +544,7 @@ pub struct Delegation {
     pubkey: Blob,
     expiration: Time,
     targets: Option<Vec<Blob>>,
+    permissions: Option<String>,
```
mraszyk: A discussion about the exact type is still ongoing: https://github.com/dfinity/developer-docs/pull/292/files#r3410163114

rs/validator/ingress_message/tests/validate_request.rs
```diff
@@ -2175,6 +2175,142 @@ mod sender_info {
...
+    }
+
+    #[test]
+    fn should_reject_all_request_types_when_delegation_permissions_unsupported() {
+        let rng = &mut reproducible_rng();
+        let verifier = verifier_at_time(CURRENT_TIME).build();
+        // Note that "updates" is deliberately unsupported: the supported
+        // vocabulary is "queries" (queries only) and "all" (queries,
+        // replicated queries, and updates).
+        for unsupported in ["writes", "updates", ""] {
```
eichhorl: Maybe we could also add capitalization and whitespace variants, i.e. `"QUERIES", "queries ", ...`

rs/validator/ingress_message/src/lib.rs
```diff
@@ -46,6 +46,8 @@ pub use internal::TimeProvider;
 /// * [`RequestValidationError::CanisterNotInDelegationTargets`]: if the request targets a canister
 ///   that is not authorized in one of the delegations.
 /// * [`RequestValidationError::InvalidSenderInfo`]: if sender info is provided but invalid.
+/// * [`RequestValidationError::UpdateCallNotPermittedByDelegation`]: if the request is an
+///   update call but a delegation restricts the sender to query calls.
```
eichhorl:
```suggestion
/// * [`RequestValidationError::UpdateCallNotPermittedByDelegation`]: if the request is submitted
///   to `/call` (an update call or replicated query) but a delegation restricts the sender to query calls.
```

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -677,6 +678,190 @@ pub fn requests_with_delegations_with_targets(env: TestEnv) {
...
+                        }
+                        Outcome::Invalid { err } => {
+                            assert_eq!(
+                                response.status(),
+                                400,
+                                "Scenario {} (update) using v{api_ver} unexpectedly succeeded",
+                                scenario.note
+                            );
+                            response.expect_text_error(err);
+                        }
```
pierugo-dfinity: nit: does this work?
```suggestion
                        Outcome::QueriesOnly { err } | Outcome::Invalid { err } => {
                            assert_eq!(
                                response.status(),
                                400,
                                "Scenario {} (update) using v{api_ver} unexpectedly succeeded",
                                scenario.note
                            );
                            response.expect_text_error(err);
                        }
```

rs/types/types/src/messages/http.rs
```diff
@@ -560,6 +562,16 @@ impl Delegation {
...
+    }
+
+    pub fn new_with_permissions(pubkey: Vec<u8>, expiration: Time, permissions: String) -> Self {
+        Self {
+            pubkey: Blob(pubkey),
+            expiration,
+            targets: None,
+            permissions: Some(permissions),
         }
     }
```
pierugo-dfinity: Would it be a lot of work to replace `new_with_targets` and `new_with_permissions` with `with_targets` and `with_permissions`, following a builder pattern? Such that we could chain the two calls in the future if we'd like to build delegations with both targets and permissions.

rs/validator/src/ingress_validation/tests.rs
```diff
@@ -207,10 +207,97 @@ fn plain_authentication_with_one_scoped_delegation() {
...
+        );
+    }
+
+    #[test]
+    fn unsupported_permission_is_rejected() {
+        assert_matches!(
+            validate_single_delegation(Some("writes")),
+            Err(InvalidDelegation(UnsupportedDelegationPermissions(value))) if value == "writes"
+        );
+        // The previously-used "updates" value is now unsupported.
```
eichhorl: What does "previously-used" mean here?

pierugo-dfinity: Were we using "updates" previously?

rs/validator/src/ingress_validation.rs
```diff
@@ -60,6 +60,35 @@ const MAXIMUM_NUMBER_OF_PATHS: usize = 1_000;
...
+const DELEGATION_PERMISSIONS_ALL: &str = "all";
+
+/// Restrictions that a chain of delegations places on the sender.
+#[derive(Debug)]
+struct DelegationRestrictions {
+    /// The set of canister IDs that are common to all delegations' `targets`.
+    targets: CanisterIdSet,
+    /// Whether some delegation in the chain restricts the sender to query
+    /// calls via the `permissions` field.
+    queries_only: bool,
```
pierugo-dfinity, line 70: Do we plan to add more permissions in the future? Having a simple boolean here does not look very extendable. For example this could be a vector of all allowed endpoints instead?

rs/validator/src/ingress_validation.rs
```diff
@@ -115,14 +147,29 @@ where
         root_of_trust_provider: &R,
     ) -> Result<CanisterIdSet, RequestValidationError> {
         validate_ingress_expiry(request, current_time)?;
-        let delegation_targets = validate_request_content(
```
eichhorl, line 118: Would it be easier to add another argument to `validate_request_content` instead of inlining it here?

rs/validator/src/ingress_validation.rs
pierugo-dfinity, line 134: This function now exceptionally duplicates `validate_request_content` motivated by saving the work of validating a potentially expensive canister signature in `validate_sender_info`. Though, couldn't a malicious request already trigger this by reaching out to the query or read_state endpoint?

rs/validator/ingress_message/tests/validate_request.rs
```diff
@@ -2175,6 +2175,142 @@ mod sender_info {
...
+    }
+
+    #[test]
+    fn should_reject_update_call_when_any_delegation_in_chain_restricts_to_queries() {
+        let rng = &mut reproducible_rng();
+        let verifier = verifier_at_time(CURRENT_TIME).build();
+        // The restriction sits in the middle of the chain; subsequent
+        // unrestricted delegations must not lift it.
+        let chain = DelegationChain::rooted_at(random_user_key_pair(rng))
+            .delegate_to_with_permissions(random_user_key_pair(rng), CURRENT_TIME, "queries")
```
eichhorl: Should there be a test that does both `delegate_to_with_permissions` and `delegate_to_with_targets`?

rs/validator/ingress_message/tests/validate_request.rs
aterga: A system test was added in `rs/tests/crypto/ingress_verification_test.rs`: `requests_with_delegation_permissions` (registered in the suite next to `requests_with_delegations_with_targets`). It drives real signed delegation chains against a replica over HTTP and checks that `"queries"` rejects update calls across all `/call` API versions while permitting query and `read_state`, `"all"` is accepted, and unsupported values are rejected for all request kinds, including the whole-chain case. (It's tagged `long_test`, so it runs in the system-test pipeline rather than standard PR CI.)

---

rs/validator/src/ingress_validation.rs
aterga: Removed the duplication entirely (5c5d850): the `/call` path now calls the shared `validate_request_content` like the query/read_state paths and checks `queries_only` afterwards. This also resolves the related concern about the short-circuit — as @pierugo-dfinity noted, short-circuiting before `validate_sender_info` didn't actually prevent DoS, since a request can trigger the same `sender_info` canister-signature verification via the query/read_state endpoints anyway. So the short-circuit bought nothing and is gone.

---

rs/validator/src/ingress_validation/tests.rs
aterga: Sorry, that was misleading — `"updates"` was never shipped or used on mainnet. It was just the value used for "unrestricted" in an earlier revision of *this* PR, before we settled on `"queries"`/`"all"`. I've reworded the comment (5c5d850) to drop "previously-used" and simply note that `"updates"` is a plausible-looking but unsupported value tested to confirm it's rejected.

---

aterga: No — `"updates"` was never used on mainnet. It was the "unrestricted" value in an earlier revision of this PR before we switched to `"queries"`/`"all"`. It's kept here only as a representative unsupported value to confirm it's rejected; reworded the comment in 5c5d850 to make that clear.

---

rs/types/types/src/messages/http.rs
aterga: Resolved in 98c7215: now that dfinity/developer-docs#292 settled the field on the union `Queries | All | Unrestricted`, the implementation models it as a typed enum `DelegationPermissions { Queries, All }` (field `Option<DelegationPermissions>`), serialized on the wire and in the representation-independent hash as the text `"queries"`/`"all"`. Unsupported values are now rejected at decoding time rather than via a runtime validator check.

---

rs/validator/ingress_message/tests/validate_request.rs
aterga: Good idea — those variants are now covered, just at a different layer. After switching `permissions` to a typed enum (98c7215), an unsupported value can no longer be constructed through the typed builder; it's rejected when the request is CBOR-decoded. So this e2e test was removed and the case/whitespace coverage moved to `delegation_permissions_rejects_unsupported_value` in `rs/types/types/src/messages/http/tests.rs`, which asserts that `"writes"`, `"updates"`, `""`, `"QUERIES"`, `"queries "`, `" queries"`, `"All"`, and `"all "` all fail to deserialize.

---

rs/validator/src/ingress_validation.rs
aterga: The `queries_only: bool` mirrors the spec's current closed set of values (`Queries | All`, per dfinity/developer-docs#292) — there are exactly two, and `All` is the unrestricted default, so a single "is the sender restricted to queries" bit captures it without loss. If the spec later introduces more permission kinds, the wire enum `DelegationPermissions` and this internal flag would be extended together at that point; generalizing now (e.g. to a set of allowed endpoints) would be speculative ahead of the spec defining what those values mean. Happy to revisit if the team wants to plan for additional permissions.

---

rs/types/types/src/messages/http.rs
aterga: Done in 5c5d850 — replaced `new_with_targets`/`new_with_permissions` with chainable `with_targets`/`with_permissions` on `Delegation`, so the two can be combined (`Delegation::new(pk, exp).with_targets(..).with_permissions(..)`). Updated all call sites, including the pre-existing `request_auth_malicious_replica_test`.

---

rs/tests/crypto/ingress_verification_test.rs
aterga: Applied the arm merge in 5c5d850. Note it changed again in 98c7215: when `permissions` became a typed enum, unsupported values are rejected at decode time, so the `Outcome::Invalid` case was removed entirely. The update-call match is now just `Outcome::All => expect_update_ok` and `Outcome::QueriesOnly { err } => expect rejection`.

---

rs/validator/ingress_message/tests/validate_request.rs
aterga: Added in 5c5d850: `should_apply_both_targets_and_permissions_restrictions_in_a_chain` (in the e2e `delegation_permissions` module) builds a chain with one `delegate_to_with_targets` link and one `delegate_to_with_permissions("queries")` link, and checks that an update to an in-target canister is rejected by the queries restriction, a query to that canister succeeds, and a query to an out-of-target canister is rejected by the targets restriction.

---

rs/validator/src/ingress_validation.rs
aterga: Good point — and you're right that the short-circuit wasn't buying anything (the query/read_state endpoints validate `sender_info` unconditionally, so a malicious request could trigger the same canister-signature work there). So in 5c5d850 I removed the duplication entirely: the `/call` path now calls the shared `validate_request_content` like the other paths and checks `queries_only` afterwards, with no short-circuit.

---

rs/validator/src/ingress_validation/tests.rs
```diff
@@ -207,10 +207,89 @@ fn plain_authentication_with_one_scoped_delegation() {
...
+        assert_matches!(
+            validate_single_delegation(None),
+            Ok(restrictions) if !restrictions.queries_only
+        );
+    }
+
+    // Unsupported `permissions` values cannot reach the validator: the field
+    // is a closed enum, so unknown values are rejected when the request is
+    // decoded. That rejection is covered by a CBOR-decoding test in ic-types
+    // (`delegation_permissions_rejects_unsupported_value`).
```
eichhorl: Seems redundant
```suggestion
```

aterga: Removed in b988ba6.

---

rs/types/types/src/messages/http.rs
```diff
@@ -536,6 +536,35 @@ impl From<CanisterIdError> for HttpRequestError {
...
 }
 
+/// The kinds of calls a delegation permits, as defined in
+/// `<https://internetcomputer.org/docs/current/references/ic-interface-spec#authentication>`.
+/// An absent field on the [`Delegation`] is unrestricted, equivalent to
+/// [`DelegationPermissions::All`]. Unsupported values are rejected when
+/// decoding the request.
```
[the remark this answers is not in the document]
aterga: Good catch — the description was stale after the typed-enum and builder refactors. I've updated it: it now describes the `DelegationPermissions` enum, notes that unsupported values are rejected at request-decoding time (no `UnsupportedDelegationPermissions` validator error — that variant was removed), and reflects the `with_targets`/`with_permissions` builder, the system test, and the dfinity/developer-docs#292 spec reference.

---

rs/validator/src/ingress_validation.rs
```diff
@@ -115,14 +138,20 @@ where
...
-        validate_request_target(request, &delegation_targets)?;
-        Ok(delegation_targets)
+        // A request to a `/call` endpoint (an update call or a replicated
+        // query) is rejected if a delegation in the chain restricts the
+        // sender to query calls.
+        if restrictions.queries_only {
+            return Err(UpdateCallNotPermittedByDelegation);
+        }
+        validate_request_target(request, &restrictions.targets)?;
+        Ok(restrictions.targets)
```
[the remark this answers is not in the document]
aterga: This was considered and deliberately not done (see the resolved threads above with @pierugo-dfinity / @eichhorl). The short-circuit doesn't actually reduce the DoS surface: the same `sender_info` canister-signature verification runs unconditionally on the query (and read_state) endpoints via the same `validate_request_content`, and those don't reject queries-restricted senders. So an attacker can force the identical work — in fact strictly *more* work, since the query then proceeds — by sending the same large `sender_info` to `/query`. Gating only `/call` therefore buys nothing while reintroducing the per-path duplication that was just removed. If we ever did want to skip it, the clean way (per eichhorl) would be a parameter to `validate_request_content` rather than inlining — but it isn't worth it here.

---

rs/types/types/src/messages/http.rs
```diff
@@ -552,15 +582,22 @@ impl Delegation {
...
-            pubkey: Blob(pubkey),
-            expiration,
-            targets: Some(targets.iter().map(|c| Blob(c.get().to_vec())).collect()),
-        }
+    /// Restricts the delegation to the given canister targets. Chainable
+    /// with [`Self::with_permissions`] to build a delegation carrying both.
+    pub fn with_targets(mut self, targets: Vec<CanisterId>) -> Self {
+        self.targets = Some(targets.iter().map(|c| Blob(c.get().to_vec())).collect());
+        self
+    }
```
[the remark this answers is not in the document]
aterga: The replacement was explicitly requested by @pierugo-dfinity (the thread above) so that `targets` and `permissions` can be combined on one delegation, and all in-repo call sites were updated. `ic-types` is an internal replica crate within this monorepo — it isn't published to crates.io as a semver-stable public API (the agent-facing delegation types live elsewhere, e.g. `@icp-sdk/core`), so there are no external downstream consumers of this `pub fn` to break. Keeping a deprecated wrapper would contradict the reviewer request and leave a second constructor for the same thing, so I'd lean against it unless there's a concrete external consumer to preserve.

---

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -677,6 +679,149 @@ pub fn requests_with_delegations_with_targets(env: TestEnv) {
...
+                All,
+            }
+
+            struct PermissionsTest {
+                note: &'static str,
+                // The `permissions` value of each delegation in the chain
+                // (`None` means the field is omitted for that delegation).
+                // Unsupported values cannot be represented (the field is a
+                // closed enum) and are rejected when the request is decoded;
+                // that path is covered by ic-types CBOR-decoding tests.
```
pierugo-dfinity:
```suggestion
```
I don't think this comment is really relevant here. It looks like an artifact of a Claude's refactor of code that never left this PR.

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -677,6 +679,149 @@ pub fn requests_with_delegations_with_targets(env: TestEnv) {
...
+                for &api_ver in ALL_QUERY_API_VERSIONS {
+                    let response = perform_query_call_with_delegations(
+                        api_ver,
+                        &test_info,
+                        sender,
+                        signer,
+                        &delegations,
+                    )
+                    .await;
+                    // Both remaining outcomes permit query calls.
```
pierugo-dfinity: What is "Both", what is "remaining"?

rs/tests/crypto/ingress_verification_test.rs
```diff
@@ -677,6 +679,149 @@ pub fn requests_with_delegations_with_targets(env: TestEnv) {
...
+                for &api_ver in ALL_READ_STATE_API_VERSIONS {
+                    let response = perform_read_state_call_with_delegations(
+                        api_ver,
+                        &test_info,
+                        sender,
+                        signer,
+                        &delegations,
+                    )
+                    .await;
+                    // Both remaining outcomes permit read_state requests.
```
pierugo-dfinity: ditto

rs/types/types/src/messages/http.rs
```diff
@@ -536,6 +536,35 @@ impl From<CanisterIdError> for HttpRequestError {
     }
 }
 
+/// The kinds of calls a delegation permits, as defined in
+/// `<https://internetcomputer.org/docs/current/references/ic-interface-spec#authentication>`.
```
pierugo-dfinity: Looks like anchor `#authentication` does not exist

rs/types/types/src/messages/http.rs
pierugo-dfinity:
```suggestion
/// [`DelegationPermissions::All`].
```
I don't think this is really relevant

eichhorl (approved):

rs/types/types/src/messages/http.rs
```diff
@@ -552,15 +582,22 @@ impl Delegation {
...
     }
 
-    pub fn new_with_targets(pubkey: Vec<u8>, expiration: Time, targets: Vec<CanisterId>) -> Self {
-        Self {
-            pubkey: Blob(pubkey),
-            expiration,
-            targets: Some(targets.iter().map(|c| Blob(c.get().to_vec())).collect()),
-        }
+    /// Restricts the delegation to the given canister targets. Chainable
+    /// with [`Self::with_permissions`] to build a delegation carrying both.
```
pierugo-dfinity:
```suggestion
    /// Restricts the delegation to the given canister targets.
```

rs/types/types/src/messages/http.rs
```diff
@@ -552,15 +582,22 @@ impl Delegation {
...
+    pub fn with_targets(mut self, targets: Vec<CanisterId>) -> Self {
+        self.targets = Some(targets.iter().map(|c| Blob(c.get().to_vec())).collect());
+        self
+    }
+
+    /// Restricts the kinds of calls the delegation permits. Chainable with
+    /// [`Self::with_targets`] to build a delegation carrying both.
```
pierugo-dfinity:
```suggestion
    /// Restricts the kinds of calls the delegation permits.
```

pierugo-dfinity (approved):
Mostly me ranting about Claude's comments

aterga:
Thanks @pierugo-dfinity — addressed all of these in 0ecc183:

- Trimmed the verbose doc comments on `DelegationPermissions`, `with_targets`, and `with_permissions` to the concise one-liners you suggested.
- Dropped the broken `#authentication` anchor (the link now points at the spec page).
- Removed the "Unsupported values…" comment in the system test (artifact of an earlier refactor, no longer relevant — unsupported values are a closed-enum decode concern covered elsewhere).
- Rewrote the confusing "Both remaining outcomes…" comments to "Query calls / read_state requests are permitted regardless of the outcome."

---

rs/types/types/src/messages/http.rs
```diff
@@ -536,6 +536,34 @@ impl From<CanisterIdError> for HttpRequestError {
     }
 }
 
+/// The kinds of calls a delegation permits, as defined in
+/// `<https://internetcomputer.org/docs/current/references/ic-interface-spec>`.
```
pierugo-dfinity:
```suggestion
/// `<https://docs.internetcomputer.org/references/ic-interface-spec/https-interface/#authentication>`.
```
Maybe this is the link you're looking for.
Although I agree that this is not the only link that is broken since we migrated the website, I think it is best to have correct links at least for newly-added code.

aterga: Applied your suggested link verbatim in 937ddbe — the doc comment now points at `https://docs.internetcomputer.org/references/ic-interface-spec/https-interface/#authentication`. Agreed on keeping at least newly-added links correct. Thanks!

---

PR #10451, opened 2026-06-12
fix: process query stats during idle epochs in PocketIC

mraszyk:
This PR fixes the query stats feature in PocketIC to process empty query stats during idle epochs to unblock processing of (non-empty) stats from previous epochs. The motivation for this change is to provide query stats in local testing (PocketIC) without need to make query calls continuously to have non-empty stats for every epoch (there's no effective change beyond new types in production, i.e., production code paths do not trigger processing of empty query stats).

### DIFF
--- a/rs/execution_environment/src/lib.rs
+++ b/rs/execution_environment/src/lib.rs
@@ -39,7 +39,7 @@ use ic_interfaces::execution_environment::{
 use ic_interfaces_state_manager::StateReader;
 use ic_logger::ReplicaLogger;
 use ic_metrics::MetricsRegistry;
-use ic_query_stats::QueryStatsPayloadBuilderParams;
+use ic_query_stats::{QueryStatsCollector, QueryStatsPayloadBuilderParams};
 use ic_registry_subnet_type::SubnetType;
 use ic_replicated_state::page_map::PageAllocatorFileDescriptor;
 use ic_replicated_state::{CallOrigin, NetworkTopology, ReplicatedState};
@@ -132,6 +132,7 @@ impl ExecutionServices {
             query_stats_payload_builder,
             cycles_account_manager,
             execution_environment,
+            query_stats_collector,
         ) = setup_execution_helper(
             logger.clone(),
             metrics_registry,
@@ -180,6 +181,7 @@ impl ExecutionServices {
             config.rate_limiting_of_instructions,
             config.log_memory_store_feature,
             Arc::clone(&fd_factory),
+            query_stats_collector,
         ));
 
         Self {
@@ -228,6 +230,7 @@ pub struct ExecutionServicesForTesting {
     pub query_stats_payload_builder: QueryStatsPayloadBuilderParams,
     pub cycles_account_manager: Arc<CyclesAccountManager>,
     pub execution_environment: Arc<ExecutionEnvironment>,
+    pub local_query_execution_stats: Arc<QueryStatsCollector>,
 }
 
 impl ExecutionServicesForTesting {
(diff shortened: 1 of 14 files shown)


rs/execution_environment/src/lib.rs
```diff
@@ -228,6 +230,7 @@ pub struct ExecutionServicesForTesting {
     pub query_stats_payload_builder: QueryStatsPayloadBuilderParams,
     pub cycles_account_manager: Arc<CyclesAccountManager>,
     pub execution_environment: Arc<ExecutionEnvironment>,
+    pub local_query_execution_stats: Arc<QueryStatsCollector>,
```
alin-at-dfinity: Nit: Most style guides advise against simply tacking on new fields/arguments to the end of the list when there's a structurally more appropriate place for them. Chronological order is not all that useful in the long run.

In this case, there are already a couple of `query_*` fields, I would have placed it next to them.

rs/query_stats/src/payload_builder.rs
```diff
@@ -304,13 +307,15 @@ impl QueryStatsPayloadBuilderImpl {
     ///
     /// This function also returns an error, if we are requesting data on an epoch,
     /// that has been already aggregated.
+    /// Returns `(previous_canister_ids, has_submitted)` where `has_submitted` is true if this
+    /// node has already submitted any report (even empty) for the given epoch.
```
alin-at-dfinity, line 312: Ultimate nitpicking: Either leave an empty line if this is supposed to be a new paragraph; or else continue on the previous line. The generated doc comment will completely ignore the newline.

rs/query_stats/src/payload_builder.rs
```diff
@@ -363,41 +368,40 @@ impl QueryStatsPayloadBuilderImpl {
...
+                            "Failed to deserialize past payload, this is a bug"
+                        );
+                    })
+                    .ok()
+                    .flatten()
+            })
+            .filter(|stats| stats.epoch == epoch && stats.proposer == node_id)
+            .collect();
+
+        let has_submitted_in_past = !matching_past_payloads.is_empty();
```
alin-at-dfinity: You could have also just added an `.inspect(|_| { has_submitted_in_past = true })` to the existing code. No need to materialize a `Vec` only to set a flag.

rs/execution_environment/src/scheduler.rs
```diff
@@ -1133,6 +1137,11 @@ impl Scheduler for SchedulerImpl {
         // The goal is to ensure that we can track the performance of `execute_round` and its individual components.
         let root_measurement_scope = MeasurementScope::root(&self.metrics.round);
 
+        // Advance the query stats epoch unconditionally so that the previous epoch's
+        // stats are flushed to the payload builder even when no queries arrive.
+        self.local_query_execution_stats
+            .set_epoch_from_height(Height::from(current_round.get()));
```
alin-at-dfinity: Doesn't this lead to a race condition between query handler (using the latest certified state) and replicated execution (using a not yet certified state)?

I see in `QueryStatsCollector::set_epoch()` that we are [somewhat relaxed about this](https://github.com/dfinity/ic/blob/3ceb4bae2eecb70fd6dd61e5865cfd0e8e14caed/rs/query_stats/src/lib.rs#L75-L83), but we do end up going back and forth between epochs if `set_epoch_from_height()` is called with the "right" heights. And that comment (obviously) only mentions races between queries, not races between non-replicated queries and replicated execution.

alin-at-dfinity (approved):

rs/execution_environment/src/scheduler.rs
mraszyk: Good point: I've changed this PR to only produce empty query stats in PocketIC (i.e., not on production code paths).

alin-at-dfinity (approved):

PR #10452, opened 2026-06-12
feat: delegations can require calls to carry a signed sender_info

aterga:
# Delegations can require calls to carry a signed `sender_info`

This is **step 1 of combining the two read-only-delegation experiments** (#10447: `sender_info` attribute checking, #10449: delegation-level permissions): request delegations gain **one bit** — an optional `sender_info_required` field, covered by the delegation signature — that forces update calls authenticated through the chain to carry a (signed and verified) `sender_info`.

## Why

The `sender_info` attribute mechanism of #10447 enforces `implicit:permissions = "queries"` *when the attributes are present* — but the party we want to constrain (the dapp holding the session key) decides whether to attach them, so it can lift the restriction by omitting the blob. This bit closes that hole while keeping all permission *semantics* in the certified attributes:

* The **delegation** (which the dapp cannot alter — the field is part of the representation-independent hash, so stripping it invalidates the signature) says: *"every update call must present the certified session attributes."*
* The **attributes** (`sender_info`, signed by the same issuer under `ic-sender-info`) say what the session may do, e.g. `implicit:permissions = "queries"` → update calls rejected per #10447.

Together: an issuer like Internet Identity signs a read-only delegation chain whose update calls must carry attributes that in turn forbid update calls — protocol-enforced, regardless of client cooperation. #10447 is being rebased on top of this PR so the stack forms the complete combined design.

## Semantics

* Field absent or `false` → today's behavior, byte-for-byte identical hashes for existing delegations; nothing in the ecosystem changes.
* `sender_info_required = true` anywhere in the chain → update calls (`/call` path) without a `sender_info` are rejected with the new `RequestValidationError::SenderInfoRequiredByDelegation`; calls *with* one proceed through the existing signature verification of the blob. Requirements only accumulate along the chain (like `targets`).
* Query and read_state requests are unaffected: read_state envelopes cannot carry a `sender_info` at all, and queries are not restricted by the permissions attribute anyway.
* The presence check runs after envelope/delegation signature validation but **before** the (expensive) canister-signature verification of the `sender_info` itself.
* Backward compatibility mirrors #10449's analysis: old replicas recompute the delegation hash without the unknown field, so restricted delegations **fail closed** on un-upgraded replicas (unusable rather than unrestricted); rollout is replicas first, issuers after.

## Changes

* `ic-types`: `Delegation.sender_info_required` (optional bool; hashed as nat 0/1 under the key `sender_info_required`), constructor, accessor; CBOR encoding tests.
* `ic-validator`: `DelegationRestrictions` threaded through delegation-chain validation; presence enforcement on the call path; new error variant mirrored in `ic-validator-ingress-message`.
* `ic-validator-http-request-test-utils`: `delegate_to_with_sender_info_required` chain-builder support.
* Tests: signed-bytes golden test; e2e tests covering update-without-sender_info rejected / update-with-valid-sender_info accepted (chain rooted at the same canister signer that signed the attributes, mirroring the II setup) / query unaffected / `false` unrestricted / requirement in the middle of a chain.

## Follow-ups

* #10447 rebased on top of this branch (attribute semantics + this bit = the combined POC).
* Interface-spec update (delegation map field + formal conditions) for dfinity/portal, and II-side issuance of the bit from the same "Read-only mode" checkbox.

`cargo test -p ic-types -p ic-validator -p ic-validator-ingress-message` passes and clippy is clean. Bazel was not available in the development container, so CI should confirm the Bazel build.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQNgPJKgxWohrJ6owwJBzz)_

### DIFF
--- a/rs/types/types/src/messages/http.rs
+++ b/rs/types/types/src/messages/http.rs
@@ -544,6 +544,7 @@ pub struct Delegation {
     pubkey: Blob,
     expiration: Time,
     targets: Option<Vec<Blob>>,
+    sender_info_required: Option<bool>,
 }
 
 impl Delegation {
@@ -552,6 +553,7 @@ impl Delegation {
             pubkey: Blob(pubkey),
             expiration,
             targets: None,
+            sender_info_required: None,
         }
     }
 
@@ -560,6 +562,20 @@ impl Delegation {
             pubkey: Blob(pubkey),
             expiration,
             targets: Some(targets.iter().map(|c| Blob(c.get().to_vec())).collect()),
+            sender_info_required: None,
+        }
+    }
+
+    pub fn new_with_sender_info_required(
+        pubkey: Vec<u8>,
+        expiration: Time,
+        sender_info_required: bool,
+    ) -> Self {
+        Self {
+            pubkey: Blob(pubkey),
+            expiration,
+            targets: None,
+            sender_info_required: Some(sender_info_required),
         }
     }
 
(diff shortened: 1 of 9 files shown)


aterga:
Closing in favor of the more lightweight approach in #10449 (delegations carry the `queries`/`all` permissions directly), which makes the sender_info-presence bit unnecessary for the read-only-sessions POC.

---

PR #10453, opened 2026-06-12
feat: Enable `PATCH` for HTTP outcalls

eichhorl:
https://github.com/dfinity/ic/pull/10378 introduced support for the `PATCH` method to the HTTP outcalls implementation. For backward compatibility reasons, the new method remained disabled such that requests specifying it are still rejected.

The change above is being rolled out this week as part of release [142256](https://dashboard.internetcomputer.org/proposal/142256).

Therefore, this PR enables the `PATCH` method such that it may be used in HTTP outcall requests.

For backward compatibility reasons, note that subnets upgraded to the version of this PR should not directly be downgraded to versions earlier than  release [142256](https://dashboard.internetcomputer.org/proposal/142256) mentioned above.

This reverts commit 26e3ae7d1b4cb5682edcaa4d828baa27fc7943e6.

### DIFF
--- a/rs/tests/networking/canister_http_correctness_test.rs
+++ b/rs/tests/networking/canister_http_correctness_test.rs
@@ -128,7 +128,8 @@ fn main() -> Result<()> {
                 .add_test(systest!(test_put_without_non_replicated_rejected))
                 .add_test(systest!(test_delete_call))
                 .add_test(systest!(test_delete_without_non_replicated_rejected))
-                .add_test(systest!(test_patch_rejected))
+                .add_test(systest!(test_patch_call))
+                .add_test(systest!(test_patch_without_non_replicated_rejected))
                 .add_test(systest!(test_max_possible_request_size))
                 .add_test(systest!(test_max_possible_request_size_exceeded))
                 // This section tests the request headers limits scenarios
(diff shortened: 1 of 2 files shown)


mraszyk:
For my own understanding: why does this require a staged roll-out? I.e., what could go wrong if that feature was implemented and enabled at the same time?

eichhorl:
> For my own understanding: why does this require a staged roll-out? I.e., what could go wrong if that feature was implemented and enabled at the same time?

@mraszyk The change introduces a new variant to the `HttpMethod` protobuf enum. The HTTP outcalls request context (including that enum) is stored in replicated state. If a `PATCH` request were made and stored in the state just before downgrading to a version that doesn't know the new variant yet, then the state would no longer be deserializable after the downgrade.

mraszyk:
Thanks a lot for the explanation! I suppose the issue comes from the fact that the deserialization of the enum in ReplicatedState does not handle failure gracefully, e.g., by rejecting the request, right?

eichhorl:
> Thanks a lot for the explanation! I suppose the issue comes from the fact that the deserialization of the enum in ReplicatedState does not handle failure gracefully, e.g., by rejecting the request, right?

Yes, we panic if the state can't be deserialized: https://github.com/dfinity/ic/blob/78365d79b6f9e25799a08b7b44e0e8bcd9c4f578/rs/state_manager/src/lib.rs#L1529-L1537

pierugo-dfinity (approved):

PR #10456, opened 2026-06-12
fix: pin Farm client to HTTP/1.1

basvandijk:
## Problem

Farm's web server (warp-tls) advertises HTTP/2 `SETTINGS_MAX_CONCURRENT_STREAMS=64` and tears down the **whole connection** (`RST_STREAM` + `GOAWAY \"exceeds max concurrent\"`) when a burst of multiplexed streams exceeds that limit — strict behavior vs. RFC 9113 §5.1.2, which prescribes rejecting only the offending stream.

The blocking reqwest client used by the test driver multiplexes all parallel requests over a **single** HTTP/2 connection. When a test starts many VMs at once (e.g. the 120-node `xnet_slo_120_subnets_staging_test`), the request burst outruns the server's SETTINGS frame on high-RTT paths and Farm kills the connection, failing all in-flight requests at once:

```
ERRO ... sending a request to Farm failed: reqwest::Error { kind: Request, ...,
source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Http2,
Error { kind: GoAway(b\"exceeds max concurrent\", REFUSED_STREAM, Remote) })) }
WARN ... Starting VM failed with: Retried too many times: sending a request to Farm
```

→ setup fails with **zero** VMs started. Low-RTT environments usually win the race, which is why this mostly bites when running system tests from devenvs.

Diagnosis detail: all failures land in the same millisecond (one shared connection dying), and `ss` shows only a single connection to Farm during a 120-VM burst.

## Fix

- **Force HTTP/1.1** on the Farm client: one TCP connection per concurrent request, no stream-concurrency limit to violate.

## Verification

Ran `//rs/tests/message_routing/xnet:xnet_slo_120_subnets_staging_test` (120 single-node subnets) from a dev container:

- Without this change: setup fails — all 120 VM-start requests die with the GoAway error above, 0 VMs started.
- With this change: setup starts all 120 VMs reliably and the test runs to completion (passed in combination with #10455).


rs/tests/driver/src/driver/farm.rs
```diff
@@ -43,6 +44,52 @@ const TIMEOUT_SETTINGS: TimeoutSettings = TimeoutSettings {
 const FARM_VM_CREATED_EVENT_NAME: &str = "farm_vm_created_event";
 const VM_CONSOLE_LINK_CREATED_EVENT_NAME: &str = "vm_console_link_created_event";
 
+/// Cap on concurrent `upload_file` calls per `Farm` instance (clones share the
+/// HTTP client and this semaphore). Farm's web server advertises HTTP/2
+/// SETTINGS_MAX_CONCURRENT_STREAMS=64 and tears down the whole connection if a
+/// burst of multiplexed streams exceeds that limit, so stay well below it,
+/// leaving headroom for other requests sharing the client (e.g. the keepalive
+/// task). Under HTTP/1.1 this bounds the number of parallel TCP connections
+/// instead.
```
nmattia: I'm a bit confused by this comment. I see below that we now enforce HTTP1.1, so what do we mean by "..., so stay well below [the HTTP2 max streams]"?

rs/tests/driver/src/driver/farm.rs
```diff
@@ -43,6 +44,52 @@ const TIMEOUT_SETTINGS: TimeoutSettings = TimeoutSettings {
...
+    fn new(permits: usize) -> Self {
+        Self {
+            permits: Mutex::new(permits),
+            cvar: Condvar::new(),
+        }
+    }
+
+    /// Blocks until a permit is free. The permit is held until the returned
+    /// guard is dropped (including on panic).
+    fn acquire(&self) -> UploadPermit<'_> {
```
nmattia: is this ever called within tokio? this will be blocking the whole thread, correct?

nmattia (commented):
Nice! couple questions to help me understand better

rs/tests/driver/src/driver/farm.rs
basvandijk: I've removed the cap on concurrent uploads and just do the HTTP/1.1 change.

nmattia (approved):

PR #10457, opened 2026-06-12
feat: Create variants that can deploy multiple copies of GuestOS to a single node

Bownairo:
Automatically configures multiple nodes for variants:
- `type4.0` (120 nodes, 2 overcommitted vCPUs, 4GB memory)
- `type4.1`
- `type4.2`
- `type4.3`
- `type4.4`


ic-os/components/setupos/install-hostos.sh
```diff
@@ -96,9 +97,30 @@ function resize_partition() {
         log_and_halt_installation_on_error "${?}" "Unable to include PV '/dev/${drive}' in VG."
     done
 
-    # Extend GuestOS LV to fill VG space
-    lvextend -i "${count}" --type striped -l +100%FREE /dev/hostlvm/guestos >/dev/null 2>&1
-    log_and_halt_installation_on_error "${?}" "Unable to extend logical volume: /dev/hostlvm/guestos"
+    local node_reward_type=$(get_config_value '.icos_settings.node_reward_type')
+
+    # Configure multiple GuestOS if type4.X
+    # TODO: Configure differently for each variant
```
Bownairo: Here we need to handle the different variants from SetupOS, and we still need to chunk up the memory/vcpus in HostOS (I hardcode lower values locally), and probably some work on the activation of the VM units.

ic-os/components/hostos/guestos/guestos@.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
```
blind-oracle, line 1: Probably you can use the systemd's unit templating, like `guestos@.service` and then inside just use `%i` as a placeholder, e.g.
```
[Unit]
Description=Manage GuestOS %i virtual machine
# Not Requires, otherwise guestos will restart when libvirtd restarts
Wants=libvirtd.service
After=libvirtd.service

RequiresMountsFor=/var

[Service]
Type=notify
ExecStartPre=/bin/sh -c 'echo -e "\nStarting GuestOS %i service...\n\n" | tee /dev/tty1 > /dev/ttyS0'
ExecStartPre=/opt/ic/bin/detect-first-boot.sh
ExecStart=/opt/ic/bin/guest_vm_runner run --slot %i
ExecStartPost=/opt/ic/bin/manageboot.sh hostos confirm
Restart=always
RestartSec=60
TimeoutStopSec=130s
```

```
systemctl start guestos@1
systemctl start guestos@2
systemctl start guestos@3
...
```

This is probably easier to scale up/down too w/o hardcoding.

Bownairo: 🙌 🙌 🙌
Great idea, hopefully I can pull that together with a generator without too much trouble.

Bownairo:
This branch might get a little messier before I clean it up, while I wire everything together, the last "working" commit for the moment is 729ab5ce5ad397f6135de45a006deab0d488696a.

ic-os/components/multi-hostos/start-guestos.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script dynamically starts "guestos@" services to create the right number of VMs per node type
+
+source /opt/ic/bin/config.sh
+
+node_reward_type=$(get_config_value '.icos_settings.node_reward_type')
+
+if [[ $node_reward_type =~ ^type4.1$ ]]; then
```
blind-oracle: Just set the COUNT in the `if`-s and later run some loop to start them?

```
for SLOT in {0..$((COUNT-1)))
do
    systemctl start guestos@${SLOT}
done
```

(out of my head, but smth like this)

frankdavid: +1

We could even merge all the systemctl starts into the same command then systemd will launch them in parallel. Something like `eval systemctl start guestos@{0..$((COUNT-1))}.service`

frankdavid:
Don't forget to update export-guestos-serial-logs.sh

Bownairo:
This now builds on the existing variants, `dev` and `prod`.

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
```diff
@@ -1,10 +1,35 @@
...
+        s=""
+    else
+        s=$i
+    fi
+
+    tail -F "/var/log/libvirt/qemu/guestos-serial$s.log" | sed --unbuffered 's/\x1b\[[0-9;]*[a-zA-Z]//g; s/\[[0-9]\+;[0-9;]*m//g' | systemd-cat -t "guestos-serial$s" -p info &
+done
+
+# And the upgrade VM
 tail -F /var/log/libvirt/qemu/upgrade-guestos-serial.log | sed --unbuffered 's/\x1b\[[0-9;]*[a-zA-Z]//g; s/\[[0-9]\+;[0-9;]*m//g' | systemd-cat -t upgrade-guestos-serial -p info &
```
frankdavid: I think we should have N upgrade GuestOS-s. For example, if multiple GuestVM-s start upgrading at once, each will need a separate Upgrade VM.

ic-os/components/hostos/guestos/start-guestos.sh
```diff
@@ -0,0 +1,25 @@
...
+    type4.2) COUNT=8 ;;
+    type4.3) COUNT=4 ;;
+    type4.4) COUNT=2 ;;
+    *) COUNT=1 ;;
+esac
+
+# If not a type4 node, fall to the default GuestOS flow
+if ((COUNT == 1)); then
+    systemctl start guestos.service || true
+    exit 0
```
frankdavid, line 22: Why are you masking the error? (same below)

rs/ic_os/networking/deterministic_ips/src/lib.rs
```diff
@@ -81,6 +81,27 @@ pub fn calculate_deterministic_mac(
     [0x6a, index, hash[0], hash[1], hash[2], hash[3]].into()
 }
 
+pub fn calculate_multi_deterministic_mac(
+    mgmt_mac: &MacAddr6,
+    deployment_environment: DeploymentEnvironment,
+    slot: u8,
```
frankdavid: I think it wouldn't be a bad idea to also include the node type here, it'd be necessary to support multiple upgrade guestos-s.

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -31,6 +33,8 @@ pub struct GuestOSTemplateProps {
     pub console_log_path: String,
     pub vm_memory: u32,
     pub nr_of_vcpus: u32,
+    pub topology: Topology,
+    pub cid: Option<u32>,
```
frankdavid: Could you add a comment or a prefix like `vsock_cid`?

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -178,30 +204,64 @@ pub(crate) fn vm_resources(_config: &HostOSConfig) -> (String, u32, u32) {
     ("kvm".to_string(), DEFAULT_VM_MEMORY_GIB, DEFAULT_VM_VCPUS)
 }
 
-pub fn vm_domain_name(guest_vm_type: GuestVMType) -> &'static str {
+fn split_resources_for_type_4(
+    config: &HostOSConfig,
+    memory: u32,
+    vcpus: u32,
+) -> (u32, u32, Topology) {
```
frankdavid: Can you add a method comment and say what the u32 values are? (or alternatively you can also add the comments inline like 
```
) -> (/*memory*/u32, /*vcpus*/u32...)
```
if the rust formatter is happy with it.

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -178,30 +204,64 @@ pub(crate) fn vm_resources(_config: &HostOSConfig) -> (String, u32, u32) {
...
+            nr_of_sockets: 2,
+            nr_of_cores: vcpus / 4,
+            nr_of_threads: 2,
+        },
+    };
+
+    (memory, vcpus, topology)
+}
+
+pub fn vm_domain_name(guest_vm_type: GuestVMType, slot: Option<u8>) -> String {
```
frankdavid: same comments here as above regarding N upgrade VMs

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -178,30 +204,64 @@ pub(crate) fn vm_resources(_config: &HostOSConfig) -> (String, u32, u32) {
...
     match guest_vm_type {
-        GuestVMType::Default => Path::new(DEFAULT_SERIAL_LOG_PATH),
-        GuestVMType::Upgrade => Path::new(UPGRADE_SERIAL_LOG_PATH),
+        GuestVMType::Default => PathBuf::from(format!("{DEFAULT_SERIAL_LOG_PATH}{slot_suffix}")),
+        GuestVMType::Upgrade => PathBuf::from(UPGRADE_SERIAL_LOG_PATH.to_string()),
     }
 }
 
 #[cfg(all(test, not(feature = "skip_default_tests")))]
 mod tests {
```
frankdavid, line 280: Don't you want to add some tests for the new logic?

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -1053,6 +1063,7 @@ mod tests {
                     .unwrap(),
                 ),
                 guest_vm_type,
+                guest_vm_slot: None,
```
frankdavid: Maybe 1-2 test cases here?

rs/ic_os/os_tools/guest_vm_runner/templates/guestos_vm_template.xml
```diff
@@ -191,7 +191,11 @@
       <alias name='console1'/>
     </console>
     <vsock model='virtio'>
+  {%- if cid.is_some() %}
```
frankdavid: optional nit: you can use `if let Some(cid) = cid` even in the template

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -22,6 +22,8 @@ const DEFAULT_VM_MEMORY_GIB: u32 = 480;
 const DEFAULT_VM_VCPUS: u32 = 64;
 const UPGRADE_VM_MEMORY_GIB: u32 = 4;
 
+const VIR_VSOCK_GUEST_CID_MIN: u32 = 3;
```
frankdavid, line 27: Can you add a note that this must be the same as in the vsock code?

rs/ic_os/vsock/vsock_lib/src/host/server.rs
```diff
@@ -3,6 +3,7 @@ use crate::protocol::{Request, Response, parse_request};
 use std::io::{Error, ErrorKind, Read, Result, Write};
 use vsock::{VMADDR_CID_ANY, VsockAddr, VsockListener, VsockStream};
 
+const VIR_VSOCK_GUEST_CID_MIN: u32 = 3;
```
frankdavid, line 8: Can you add a note that this must be the same as in the guest runner?

ic-os/components/monitoring/hostos/monitor-guestos.service
frankdavid, line 8: You may have to update the logic in `monitor-guestos.sh`.

ic-os/components/hostos/guestos/guestos@.service
frankdavid, line 1: Can we deduplicate this and `guestos.service`?

ic-os/components/hostos/guestos/start-guestos.sh
Bownairo: Sometimes the services fail initially, mostly due to trying to schedule so many VMs at once. The restart logic gets them there in a few attempts, and I didn't want to have this service go down if one of them didn't make it. I think skipping the check instead is a better fit, what do you think?

rs/ic_os/os_tools/guest_vm_runner/templates/guestos_vm_template.xml
Bownairo: Very cool!

ic-os/components/hostos/guestos/guestos@.service
Bownairo: I'm still trying to work out a good way to get the `slot` information across without disrupting the default flow too much. (Right now passing `slot` also switches to the new IP format.)

I'm thinking maybe `--slot 0` can be for the default case, and multi VMs could always use 1-n? I don't think there's a good way to pass this info to the service start 😕.

ic-os/components/hostos/guestos/start-guestos.sh
frankdavid: Could you add this as a comment? If we know at which point the guestos fails to start, it'd be nice to handle the failure at a lower level (at the exact point where it happens) - log+retry or even better: fix the code such that multiple instances don't cause problems.

rs/ic_os/networking/deterministic_ips/src/lib.rs
```diff
@@ -214,4 +217,75 @@ mod test {
...
+            DeploymentEnvironment::Mainnet,
+            NodeType::UpgradeGuestOS,
+            241,
+        );
+        let slaac = mac.calculate_slaac(prefix).unwrap();
+        assert_eq!(slaac, expected_ip);
+    }
+
+    #[test]
+    fn multi_slaac_similarities() {
```
frankdavid, line 241: Could you add a test that ensures that multi mac with slot = 0 == single mac? (at least I'd hope that's the case)

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -1083,7 +1087,7 @@ mod tests {
                 systemd_notifier,
                 termination_token,
                 libvirt_connection: self.libvirt_connection.clone(),
-                vm_domain_name: vm_domain_name(guest_vm_type, None).to_string(),
+                vm_domain_name: vm_domain_name(guest_vm_type, slot).to_string(),
```
frankdavid: `vm_domain_name` already returns `String` so `to_string()` is unnecessary

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -1039,7 +1039,11 @@ mod tests {
         /// Starts a VM service in the background.
         /// This roughly corresponds to invoking `run_guest_vm()` in prod code.
         /// The returned instance can be used to interact with the newly started service.
-        fn start_service(&self, guest_vm_type: GuestVMType) -> TestServiceInstance {
+        fn start_service(
+            &self,
+            guest_vm_type: GuestVMType,
+            slot: Option<u8>,
```
frankdavid: Here and elsewhere: do you think it's possible to declare slot as never empty? On "normal" nodes with a single Guest VM, it'd just take slot 0. Then we wouldn't need to differentiate between None and Some(0).

frankdavid: So even the systemd service name could be guestos0 just like the first guestos on a multi-guestos system.

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -1250,10 +1254,36 @@ mod tests {
...
+
+        let mut service1 = fixture.start_service(GuestVMType::Default, Some(0));
+        service1.wait_for_systemd_ready().await;
+
+        let mut service2 = fixture.start_service(GuestVMType::Default, Some(64));
+        service2.wait_for_systemd_ready().await;
+
+        // Assert that both VMs are running
+        service1.check_vm_running().unwrap();
+        service2.check_vm_running().unwrap();
```
frankdavid, line 1284: Consider adding a check that they use their own disk.

ic-os/components/hostos/guestos/guestos@.service
Bownairo: I took a shot at this, what do you think?

rs/ic_os/networking/deterministic_ips/src/lib.rs
Bownairo: This wasn't the case, but now with how I've re-worked things, it's part of the type.

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
Bownairo: This is now wrapped up with the new type.

basvandijk (approved):

ic-os/testing/reload_icos.sh
frankdavid, line 1: Could you add a check at the beginning that the host only has 1 GuestOS? The behavior may be unexpected when there are multiple GuestOS-s, I think it's better to just not support it and let the user know the error until we support it properly. You could also add a TODO to support multi-hosts in the quick deployment tool. We'd have to add a parameter for the guestos slot.

rs/ic_os/config/types/src/lib.rs
```diff
@@ -343,6 +344,30 @@ pub struct FixedIpv6Config {
...
+            0 => VmSlot::Plain,
+            v => VmSlot::Multi(NonZeroU8::new(v).unwrap()),
+        }
+    }
+}
+
+impl fmt::Display for VmSlot {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            VmSlot::Plain => write!(f, ""),
```
frankdavid: nit: it's a bit weird to have empty output for Display

rs/ic_os/config/types/src/lib.rs
```diff
@@ -343,6 +344,30 @@ pub struct FixedIpv6Config {
     pub gateway: Ipv6Addr,
 }
 
+#[derive(Copy, Clone)]
+pub enum VmSlot {
```
frankdavid, line 348: Is there a benefit in special-casing single guestos ("plain") machines?

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -168,6 +161,11 @@ pub fn generate_vm_config(
 
     metrics.set_hugepages_enabled(guest_vm_type, use_hugepages);
 
+    let vsock_cid = match guest_vm_slot {
+        VmSlot::Plain => None,
+        VmSlot::Multi(v) => Some(Into::<u8>::into(v) as u32 + VIR_VSOCK_GUEST_CID_MIN),
```
frankdavid: nit: probably easier to just call `v.get()` instead of `Into::<u8>::into(v) `

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -241,27 +239,28 @@ fn split_resources_for_type_4(
     (memory, vcpus, topology)
 }
 
-pub fn vm_domain_name(guest_vm_type: GuestVMType, slot: Option<u8>) -> String {
-    let slot_suffix = slot.map(|v| v.to_string()).unwrap_or_default();
+pub fn vm_domain_name(guest_vm_type: GuestVMType, slot: VmSlot) -> String {
     match guest_vm_type {
-        GuestVMType::Default => format!("{DEFAULT_GUEST_VM_DOMAIN_NAME}{slot_suffix}"),
-        GuestVMType::Upgrade => format!("{UPGRADE_GUEST_VM_DOMAIN_NAME}{slot_suffix}"),
+        GuestVMType::Default => format!("{DEFAULT_GUEST_VM_DOMAIN_NAME}{slot}"),
```
frankdavid: nit: Using `Display` for this feels somewhat brittle and highly coupled (see my comment above about producing empty string for plain slots)

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -464,17 +466,17 @@ impl GuestVmService {
         )
         .context("Could not initialize SEV certificate provider")?;
 
+        let device_string = format!("{GUESTOS_DEVICE}{guest_vm_slot}");
```
frankdavid: nit: since now we construct the device as a local, we could construct a `PathBuf` directly (this was not possible before with the `GUESTOS_DEVICE` constant because `Path::new` is not `const`).

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -993,6 +996,16 @@ mod tests {
                 .to_string()
         }
 
+        fn get_disk_path(&self) -> String {
+            let domain = self.get_domain();
+            let vm_config = domain.get_xml_desc(0).unwrap();
+            Regex::new("<source dev='([^']+)'/>")
+                .unwrap()
+                .captures(&dbg!(vm_config))
```
frankdavid: do you need this `dbg!` here?

rs/ic_os/config/types/src/lib.rs
Bownairo: Agreed, I picked a different fn, `to_suffix`, to use instead.

rs/ic_os/config/types/src/lib.rs
Bownairo: I have it so that a "plain" machines (maybe "standalone"? I don't really like the "plain".) still use the normal MAC/IP addressing, with the `68` prefix. For multi machines we use `78` with the extra slot space. I wanted it so that all existing machines don't change address, and all cloud engine machines on a single node share the same MAC/IP format.

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
Bownairo: Is this what you had in mind?

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
Bownairo: No, thank you 😅.

frankdavid (approved):

PR #10253, opened 2026-05-19
feat: support running system-tests locally instead of via Farm

basvandijk:
# What?

Provide a "local" variant for each system-test which deploys testnets inside the local bazel sandbox (without `sudo` and without `requires-network`) instead of using the external, proprietary and VPN-only Farm web-service which deploys testnets to DFINITY's DCs.

To run a system-test locally suffix its name with `_local` for example:

```
bazel test //rs/tests/execution:general_execution_test_local --test_output=streamed --cache_test_results=no
```

The `_local` system-tests currently don't run as part of the `bazel-test-all` job since they reserve a high number of CPUs from bazel causing them to spend most of their time acquiring resources from the single 64-CPU GitHub runner. In total it takes over 3 hours to run them which is obviously too long for PRs. The plan is to run all tests on a Bazel Remote Execution cluster such that they parallelise over many 64-CPU machines. This is WIP. For the moment we just run the single `//rs/tests/idx:basic_health_test_local` in `bazel-test-all` as a smoke-test such that the `local_backend` code gets exercised. (See [BuildBuddy](https://dash.dm1-idx1.dfinity.network/invocation/64faba9b-6a51-4996-88b8-d4860cadeea0?target=%2F%2Frs%2Ftests%2Fidx%3Abasic_health_test_local&targetStatus=5)). The remaining jobs will temporarily run in the `local-system-tests` job of the nightly `schedule-daily` workflow.

# Why?

* More hermetic and isolated tests which don't suffer from networking issues. No more image downloads that time out, IPv4 address exhaustion, packet drops between our DCs, etc..
* More efficient server usage since we don't need as many dedicated Farm hosts. We could run all tests on a Bazel Remote Execution cluster meaning that a builder will not only be used for hosting VMs but also for running other builds / tests.
* Support running most of CI on 3rd party services like [Namespace.so](https://namespace.so/) which means less self-hosted hardware to support.
* Allow external (non-DFINITY) devs to run system-tests since no proprietary infrastructure like Farm is needed anymore.
* Easier debugging of tests since all logs (journal and console) are streamed to the test log. So the DFINITY-VPN-only ElasticSearch, Kibana and Farm-console feature are no longer required.

# Status?

166 non farm-only tests are [passing with the local backed](https://dash.dm1-idx1.dfinity.network/invocation/86a57328-6224-4758-b4e6-855e77e3c8fc#targets)!

[...]

PR #10459, opened 2026-06-14
fix: deflake //rs/tests/cross_chain:ic_xc_cketh_test by compiling contracts offline

basvandijk:
## Problem

`//rs/tests/cross_chain:ic_xc_cketh_test` and `…:ic_xc_cketh_test_head_nns` are
flaky in data centers with unreliable IPv4 connectivity (e.g. `dm1`), panicking
in `deploy_smart_contract`:

```
thread 'main' panicked at rs/tests/cross_chain/ic_xc_cketh_test.rs:
called `Result::unwrap()` on an `Err` value: block_on_bash_script: exit_status = 1. Output:
  Err: Error: error sending request for url (https://binaries.soliditylang.org/linux-amd64/list.json)
  - Error #1: tcp connect error: Network unreachable (os error 101)
```

## Root cause

The `foundry` container ships `forge`/`cast`/`anvil` but **no `solc`**. When `forge create` compiles a helper contract, it downloads the Solidity compiler from `binaries.soliditylang.org` at test runtime (first the release index `list.json`, then the `solc` binary). That is why the foundry UVM was created with `.enable_ipv4()`. Whenever the UVM's IPv4 connectivity is unavailable, the download fails and the test flakes.

This is a third, distinct source of flakiness from the two fixed in #9209.

This is also [preventing](https://github.com/dfinity/ic/pull/10253/changes#r3387217110.) these tests to run on the local system-test backend which doesn't have external network access.

## Fix

Vendor the required `solc` compilers as pinned, hash-verified Bazel `http_file` dependencies, ship them in the ckETH UVM config image, bind-mount them into the foundry container, and pass them to `forge create --use <path>`. Compilation now happens fully offline and the `.enable_ipv4()` dependency is removed.

- **MODULE.bazel**: add the `solc` and `solc_eth_deposit_helper` `http_file` deps (sha256 taken from `list.json`). The compilers are named by role rather than version, so a future compiler bump only touches the `http_file` URL and sha256.
- **rs/tests/cross_chain/BUILD.bazel**: add both binaries to the `cketh_uvm_config_image` srcmap (→ `/config/solc` and `/config/solc_eth_deposit_helper`).
- **rs/tests/cross_chain/ic_xc_cketh_test.rs**: drop `.enable_ipv4()`; copy the compilers to a writable path + `chmod +x` (vfat drops the exec bit); pass the compiler matching each contract's `pragma` to `forge create --use`.

This also removes a source of non-determinism: contracts declaring `pragma solidity ^0.8.20` previously let `svm` pick the latest 0.8.x available at runtime; they are now pinned to exactly 0.8.20. `EthDepositHelper.sol` stays on 0.8.18 to remain faithful to the [deployed mainnet helper contract](https://etherscan.io/address/0x7574eB42cA208A4f6960ECCAfDF186D627dCC175), so it gets its own `solc_eth_deposit_helper` compiler.

## Verification

Ran on `dm1` (the flaky environment) with caching disabled:

```
bazel test //rs/tests/cross_chain:ic_xc_cketh_test --cache_test_results=no
```

```
============== Summary for //rs/tests/cross_chain:ic_xc_cketh_test ==============
Task setup                            PASSED  in 118.22s
Task ic_xc_cketh_test                 PASSED  in 225.80s
Task assert_no_metrics_errors         PASSED  in   3.96s
Task assert_no_unallowed_log_patterns PASSED  in   0.85s
//rs/tests/cross_chain:ic_xc_cketh_test    PASSED in 353.9s
```

The foundry UVM came up without IPv4, `forge create --use` compiled and deployed all four contracts from the vendored compilers, and every ETH/ERC-20/subaccount deposit flow passed — no network fetch to `binaries.soliditylang.org`.


gregorydemay:
@basvandijk Thanks for that PR! I tried twice to run in the devenv container (also after a bazel clean)
```
bazel test //rs/tests/cross_chain:ic_xc_cketh_test --cache_test_results=no
```
but I'm getting some errors
```
========================================== Summary for //rs/tests/cross_chain:ic_xc_cketh_test ==========================================
Task assert_no_metrics_errors         PASSED  in   0.04s   -- Exited with code 0.
Task assert_no_unallowed_log_patterns PASSED  in   0.46s   -- Exited with code 0.
Task setup                            FAILED  in 511.64s   -- a scoped thread panicked
Task ic_xc_cketh_test                 FAILED  in   0.05s   -- Failed to find SetupResult attribute after setup. Cancelling test function.
=========================================================================================================================================
```

gregorydemay (approved):

rs/tests/cross_chain/ic_xc_cketh_test.rs
```diff
@@ -49,6 +49,13 @@ use std::time::Duration;
 const FOUNDRY_VM_NAME: &str = "foundry";
 const DOCKER_NETWORK_NAME: &str = "ethereum";
 const FOUNDRY_PORT: u16 = 8545;
+/// File names of the `solc` compilers vendored via Bazel (see `MODULE.bazel`)
+/// and shipped in the foundry UVM config image. They are used by
+/// `deploy_smart_contract` so that `forge` compiles the helper smart contracts
+/// offline instead of downloading `solc` at runtime. The version of each
+/// compiler matches the `pragma solidity` of the contracts it compiles.
+const SOLC_V0_8_18: &str = "solc-0.8.18";
+const SOLC_V0_8_20: &str = "solc-0.8.20";
```
nmattia: nit: would remove the version from the variable names to avoid having to modify it everywhere if ever updated

nmattia (approved):

rs/tests/cross_chain/ic_xc_cketh_test.rs
basvandijk: Good point. Done.

PR #10460, opened 2026-06-14
fix: deflake //rs/tests/consensus/orchestrator:ssh_access_to_nodes_test

basvandijk:
## Summary

Deflakes `//rs/tests/consensus/orchestrator:ssh_access_to_nodes_test`, whose subtest `node_does_not_remove_keys_on_restart` was flaky.

## Root cause

All recent flaky runs failed identically: a panic at `rs/tests/consensus/utils/src/ssh_access.rs:101` (the `.unwrap()` in `assert_authentication_works`) with:

```
[Session(-43)] Failed getting banner
```

`Session(-43)` is `LIBSSH2_ERROR_SOCKET_RECV` and *"Failed getting banner"* means the SSH **transport** handshake (`session.handshake()`) failed **before any authentication took place**. This is a transient transport-level hiccup that occurs when the node is briefly unresponsive — in particular right after `sudo systemctl restart ic-replica`, which is exactly what this subtest does before repeatedly asserting (in two polling loops) that SSH access still works.

The SSH keys were present the entire time (that is precisely what the subtest verifies), so this was a **false failure**: a transport hiccup, not a missing key. Because `assert_authentication_works` did a single connect + handshake + auth with **no retry**, one transient transport failure was enough to panic the whole test.

This matches all 4 flaky runs observed in the last week, which always:
- failed in `node_does_not_remove_keys_on_restart`,
- panicked at `ssh_access.rs:101` with `Failed getting banner`,
- did so during the post-restart "still accept connections" polling loops.

## Fix

Split `SshSession::login` into two steps:
- `connect` — TCP connect + SSH transport handshake (the transport layer).
- `authenticate` — user authentication (reflects whether the key actually grants access).

`assert_authentication_works` now retries only the transport `connect` step for a bounded time (the existing `SSH_ACCESS_TIMEOUT` of 30s with `SSH_ACCESS_BACKOFF` of 5s), then authenticates **exactly once**. Transient transport failures are absorbed, while a genuine loss of access (e.g. a key being incorrectly removed) still fails immediately, because authentication is not retried. The public `login` signature is unchanged, so the other callers are unaffected.

## Verification

`bazel test --test_output=errors --runs_per_test=3 --jobs=3 --cache_test_results=no //rs/tests/consensus/orchestrator:ssh_access_to_nodes_test`

2 of 3 runs passed fully, including `node_does_not_remove_keys_on_restart`, and the `Failed getting banner` panic did not recur. The 1 remaining failure was an unrelated `Retried too many times: sending a request to Farm` infrastructure timeout during setup, which is explicitly ignored per the flaky-test guide.

---

### DIFF
--- a/rs/tests/consensus/utils/src/ssh_access.rs
+++ b/rs/tests/consensus/utils/src/ssh_access.rs
@@ -80,21 +80,35 @@ impl Default for SshSession {
 }
 
 impl SshSession {
-    /// Opens a TCP connection to the node and performs the SSH transport handshake
-    /// (protocol-version banner exchange and key exchange). This is the transport
-    /// layer, established before any authentication takes place. Failures here (e.g.
-    /// "Failed getting banner" when the node is briefly unresponsive right after a
-    /// restart) are transient and unrelated to whether a key grants access.
-    fn connect(&mut self, ip: &IpAddr) -> Result<(), String> {
+    pub fn login(&mut self, ip: &IpAddr, username: &str, mean: &AuthMean) -> Result<(), String> {
+        // The SSH transport handshake (TCP connect + protocol banner and key
+        // exchange) is retried for a bounded time because it can fail transiently
+        // (e.g. with "Failed getting banner") when the node is briefly unresponsive,
+        // for instance right after the replica/orchestrator is restarted. Such
+        // transport-level failures are unrelated to whether the key grants access,
+        // which is what the callers actually test. The authentication step below is
+        // not retried, so an actual loss of access (e.g. a key being incorrectly
+        // removed) is still reported immediately.
         let ip_str = format!("[{ip}]:22");
-        let tcp = TcpStream::connect(ip_str).map_err(|err| err.to_string())?;
-        self.session.set_tcp_stream(tcp);
-        self.session.handshake().map_err(|err| err.to_string())
-    }
+        let start = std::time::Instant::now();
+        loop {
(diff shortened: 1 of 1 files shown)


pierugo-dfinity (commented):
I looked at callers of `login` and they all interpret an `Err` as being a failed authentication. I think it would make more sense to put the retry logic inside `login` and keep `assert_authentication_works` unchanged.
In that case, for simplicity we also wouldn't need to extract the two steps into functions `connect`/`authenticate` but keep it like today (just with a retry logic around the connection step).

basvandijk:
Good call, thanks! I moved the bounded transport-handshake retry inside `login` and reverted the `connect`/`authenticate` split, so `assert_authentication_works` is unchanged again. Since every caller of `login` treats an `Err` as a failed authentication, retrying only the transient transport step (TCP connect + handshake) inside `login` now benefits all callers — including `assert_authentication_fails`, which addresses the related point the Copilot reviewer raised. Authentication is still attempted only once, so a genuine loss of access is reported immediately.

Re-verified with `bazel test --runs_per_test=3 --jobs=3 --cache_test_results=no //rs/tests/consensus/orchestrator:ssh_access_to_nodes_test`: all 3 runs passed, with `node_does_not_remove_keys_on_restart` succeeding each time.

pierugo-dfinity (approved):
Thanks Bas! This is what I had in mind.

PR #10465, opened 2026-06-15
chore: harden assert_no_snapshot and get_registry_version in migration canister

mraszyk:
This PR hardens the following checks to work after subnet deletion:
- asserting no snapshots succeeds if the call returns `DestinationInvalid` (if the canister or subnet was deleted, then the canister has no snapshots trivially);
- waiting until a subnet is at a certain registry version only makes sense as long as the subnet exists.

### DIFF
--- a/rs/migration_canister/src/external_interfaces/management.rs
+++ b/rs/migration_canister/src/external_interfaces/management.rs
@@ -273,22 +273,30 @@ pub async fn assert_no_snapshots(canister_id: Principal) -> ProcessingResult<(),
                 ))
             }
         }
-        Err(e) => {
-            println!(
-                "Call `list_canister_snapshots` for {} failed: {:?}",
-                canister_id, e
-            );
-            match e {
-                CallError::CallRejected(e) => {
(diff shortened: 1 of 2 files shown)


rs/migration_canister/src/external_interfaces/management.rs
```diff
@@ -273,13 +273,30 @@ pub async fn assert_no_snapshots(canister_id: Principal) -> ProcessingResult<(),
...
+        Err(e) => match e {
+            CallError::CallRejected(e) => {
+                if e.reject_code() == Ok(RejectCode::DestinationInvalid) {
+                    println!(
+                        "Call `list_canister_snapshots` for {} returned DestinationInvalid, treating as success",
+                        canister_id
+                    );
+                    ProcessingResult::Success(())
+                } else {
+                    println!(
```
schneiderstefan: Can this be written nicer? Two of the 3 arms are exactly the same.

mraszyk: Good point! Done.

schneiderstefan (approved):

PR #10466, opened 2026-06-15
chore(DSM-121): add critical_error_sync_call_unknown_certificate_status

mraszyk:
This PR introduces a critical error for the case of the synchronous update call HTTP handler trying to return a certificate in which the update call status is unknown. This case has not happened in prod for ~10 days by now ([kibana](https://kibana.mainnet.dfinity.network/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:now-30d%2Fd,to:now))&_a=(columns:!(),filters:!(),hideChart:!f,index:ic-logs-mainnet,interval:auto,query:(language:kuery,query:'%22Unknown%20status%20of%20call%22'),sort:!(!(timestamp,desc))))) so it is safe to assume that the underlying issue has been fixed and raise a critical error going forward.

### DIFF
--- a/rs/http_endpoints/public/src/call/call_sync.rs
+++ b/rs/http_endpoints/public/src/call/call_sync.rs
@@ -350,8 +350,8 @@ async fn call_sync(
         error!(
             every_n_seconds => LOG_EVERY_N_SECONDS,
             log,
-            "Unknown status of call {} in the certificate at height {}. This is a critical error: {}",
-            message_id, certification.height, CRITICAL_ERROR_SYNC_CALL_UNKNOWN_CERTIFICATE_STATUS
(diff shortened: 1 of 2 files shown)


rs/http_endpoints/public/src/call/call_sync.rs
```diff
@@ -345,14 +346,16 @@ async fn call_sync(
         .with_label_values(&[status_label])
         .inc();
 
-    // TODO(DSM-121): ensure that `ParsedMessageStatus::Unknown` never occurs
```
schneiderstefan, line 348: Nit: You should put mention this ticket number in the title so that the PR and the ticket get automatically linked.

schneiderstefan (approved):

PR #10467, opened 2026-06-15
feat: subnet deletion in DSM

mraszyk:
This PR implements subnet deletion in DSM: when the deleted (remote) subnet is gone from the `subnet_list` and `routing_table` registry keys, the (local) subnet
- produces synthetic reject responses (with the reject message "Canister has been uninstalled" and the corresponding reject code) for all outstanding callbacks such that (1) their respondent is on the remote subnet and (2) the corresponding request is no longer in an output queue/the corresponding response is not in an input queue;
- and drops the stream to the remote subnet (implemented in a previous [PR](https://github.com/dfinity/ic/pull/10145));
- bounded-wait/unbounded-wait responses to the remote subnet are dropped (unbounded-wait responses still raise a critical error) and their (lost) cycles are observed in metrics (output queue responses only, cycles lost in dropped stream responses are not tracked since the response might have been pulled and processed by the remote subnet before deletion and not signaled);
- tests cover canister/subnet bounded-wait/unbounded-wait requests/responses that are in output queues/streams.


rs/messaging/src/routing/stream_builder.rs
```diff
@@ -651,34 +658,60 @@ impl StreamBuilderImpl {
...
+                own_cost_schedule,
+            ));
+        }
+
         for req in requests_to_reject {
+            // A response may already be enqueued if `critical_error_infinite_loops` was raised in
+            // a previous `build_streams()` call, causing it to exit early and leave some output
+            // messages to a deleted subnet unprocessed; `generate_reject_responses_for_deleted_subnets`
+            // subsequently processed such a message and enqueued a response.
+            // Skip to avoid a duplicate response.
```
alin-at-dfinity: I don't think this is a realistic case: David added that validation and critical error, but it basically guards (in production code) against bugs in `OutputIterator`. `OutputIterator::next()` will only return `Some(x)` if it has popped that `x` from one output queue or other.

I believe we'd be better off dropping the check and critical error instead and just assume that, as with any other iterator, this one will eventually terminate.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -651,34 +658,60 @@ impl StreamBuilderImpl {
...
-                                    "{}: Discarding response, destination not found: {:?}",
-                                    CRITICAL_ERROR_RESPONSE_DESTINATION_NOT_FOUND,
+                                    "Discarding unbounded-wait response, destination not found: {:?}",
                                     rep
                                 );
-                                self.metrics
-                                    .critical_error_response_destination_not_found
-                                    .inc();
                             }
+                            dropped_response_cycles += rep.refund;
```
alin-at-dfinity: Why are we downgrading this from a critical error to a warning? I suppose that one could argue either way when it comes to an actually deleted subnet (Should we have deleted it with pending guaranteed response calls?).

But what if this happens due to a bug? E.g. the response was supposed to go to A but somehow it ended up sent to a nonexistent B. We'd need to at least weigh that against the potential alert spam resulting from deleting a subnet with pending guaranteed response calls.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -876,7 +877,7 @@ impl ReplicatedState {
         self.metadata.streams.get(destination_subnet_id)
     }
 
-    /// Discards streams to subnets no longer present in the network topology.
+    /// Discards outgoing streams targeting subnets no longer present in the network topology.
```
alin-at-dfinity: Nit: There are only outgoing streams. The incoming ones are "stream slices" and we never actually see them as streams, it's just the XNet payload builder that deals with them as actual streams. Beyond that, they're just individual messages we are inducting.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -894,9 +895,100 @@ impl ReplicatedState {
...
     }
 
+    /// Generates synthetic reject responses for callbacks to deleted subnets.
+    ///
+    /// Should be called after `build_streams()`: `build_streams()` unconditionally
+    /// rejects any request in an output queue with no route (e.g. destined for a
+    /// deleted subnet), ensuring such callbacks already have an enqueued response
+    /// and are skipped by this function. Calling before `build_streams()` is not
+    /// incorrect, but may result in a less precise error message for those callbacks
+    /// (the reject from `build_streams()` contains `DestinationInvalid` which is more precise).
```
alin-at-dfinity: This is not accurate: `build_streams()` would produce a critical error if it was called after this and tried to produce what would be a second reject response for a guaranteed-response call. While we silently ignore duplicate best-effort responses (as we may have already generated a timeout response locally), there is currently no circumstance under which we should encounter duplicate (or multiple) responses to guaranteed-response calls.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -894,9 +895,100 @@ impl ReplicatedState {
...
+    pub fn generate_reject_responses_for_deleted_subnets(&mut self) -> Vec<StateError> {
```
alin-at-dfinity: This looks like routing logic. I.e. it should live somewhere under `rs/messaging`, not in `rs/replicated_state`.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -894,9 +895,100 @@ impl ReplicatedState {
...
+        let routing_table = self.routing_table();
+        // Collect subnet IDs to also skip callbacks to management canisters of other subnets
+        // (where respondent = subnet_id, which has no entry in the routing table).
+        let subnet_ids: BTreeSet<SubnetId> = self
+            .metadata
+            .network_topology
+            .subnets()
+            .keys()
+            .cloned()
+            .collect();
```
alin-at-dfinity: You could use `NetworkTopology::route()` instead of `RoutingTable::lookup_entry()` below. It deals transparently with subnet IDs.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -894,9 +895,100 @@ impl ReplicatedState {
...
+                    && !canister
+                        .system_state
+                        .queues()
+                        .has_enqueued_response(callback_id)
+                {
+                    rejects.entry(*canister_id).or_default().push((
+                        *callback_id,
+                        respondent,
+                        callback.deadline,
+                    ));
```
alin-at-dfinity: Nit: You could already produce the reject responses here, rather than taking the indirect route of storing the relevant bits in a tuple first.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -894,9 +895,100 @@ impl ReplicatedState {
...
+            .keys()
+            .cloned()
+            .collect();
+        let own_subnet_type = self.metadata.own_subnet_type;
+
+        // Collect per-canister (callback_id, respondent, deadline) tuples where
+        // the respondent is on a deleted subnet and no response has been enqueued yet.
+        // Grouped by canister_id so each canister is removed and reinserted exactly once.
+        let mut rejects: BTreeMap<CanisterId, Vec<(CallbackId, CanisterId, CoarseTime)>> =
+            BTreeMap::new();
```
alin-at-dfinity: This code will almost never be executed, so I don't think we need to worry about efficiency. If you simply do `ReplicatedState::push_input()` instead, you won't need to remove and replace canister states and explicitly group responses by canister. Just collect all reject responses into a flat `Vec`, then enqueue them in the simplest way possible.

rs/messaging/src/state_machine.rs
```diff
@@ -264,6 +264,45 @@ impl StateMachine for StateMachineImpl {
...
+            }
+            // Only update the subnet list if all responses were successfully inducted.
+            // On error, retry next round (already-enqueued responses are filtered by
+            // `has_enqueued_response`, so only the relevant callbacks are retried).
+            if errors.is_empty() {
+                state_after_stream_builder
+                    .metadata
+                    .subnet_ids_at_last_reject_generation = Some(current_subnet_ids);
+            }
+        }
```
alin-at-dfinity: Personal preference. I'd put all of this logic inside `generate_reject_responses_for_deleted_subnets()` (or some wrapper, if you prefer) to keep `StateMachineImpl::execute_round()` at least vaguely concise and readable.

It also needs to be covered by a timer, e.g. by moving it before the `message_routing_timer.observe_duration()` just above (it does have to do with routing messages more than anything).

rs/replicated_state/src/metadata_state.rs
```diff
@@ -193,6 +193,14 @@ pub struct SystemMetadata {
...
     #[validate_eq(Ignore)]
     pub logs_migrated: bool,
+
+    /// The subnet list from network topology that was last used by
+    /// `generate_reject_responses_for_deleted_subnets()`. The function is
+    /// skipped if the subnet list has not changed since the last call.
+    ///
+    /// Transient: reset to `None` on checkpoint load.
+    #[validate_eq(Ignore)]
+    pub subnet_ids_at_last_reject_generation: Option<Vec<SubnetId>>,
```
alin-at-dfinity, line 203: Thinking out loud: if we implement the reuse of `NetworkTopology` across successive states using the same registry version, we could integrate the detection of deleted subnets into that and set a plain boolean flag.

rs/messaging/src/state_machine/tests.rs
```diff
@@ -276,49 +278,251 @@ fn state_machine_handles_messages_to_deleted_subnet() {
...
+        let stream = streams.entry(SUBNET_2).or_default();
+        // Regular requests: local_canister → remote_canister.
+        stream.push(StreamMessage::Request(Arc::new(
+            RequestBuilder::new()
+                .sender(local_canister_id)
+                .receiver(remote_canister_id)
+                .payment(stream_cycles)
+                .deadline(deadline)
+                .build(),
+        )));
```
alin-at-dfinity: Wouldn't it be cleaner to pop the requests from the local canister's output queue and push them onto the stream? As in: enqueue the first 4 requests into output queue; consume the output queue, enqueuing everything into the stream; enqueue the 2 already delivered requests into the output queue; consume them; enqueue the rest.

rs/messaging/src/state_machine/tests.rs
```diff
@@ -276,49 +278,251 @@ fn state_machine_handles_messages_to_deleted_subnet() {
...
+    // causing the stream builder to generate a reject for output requests.
+    let remote_canister_id = CANISTER_RANGE_B.start;
+    let subnet_as_canister_id = CanisterId::from(SUBNET_2);
+    let deadline = CoarseTime::from_secs_since_unix_epoch(u32::MAX);
+    // Cycles attached to output-queue requests: refunded in reject responses.
+    let req_payment = Cycles::new(1_000_000);
+    // Cycles attached to output-queue responses: observed in DroppedMessages metric.
+    let resp_refund = Cycles::new(2_000_000);
+    // Cycles attached to stream messages: silently dropped, NOT observed in any metric.
+    let stream_cycles = Cycles::new(3_000_000);
```
alin-at-dfinity: Nit: You could make these differ by a couple of orders of magnitude, so it becomes impossible for e.g. 2 request payments to be interpreted as one response refund.

alin-at-dfinity (approved):
Only nitpicks, LGTM.

rs/messaging/src/routing/stream_builder.rs
mraszyk: In every loop iteration, we do `output_iter.exclude_queue()` or `output_iter.next().unwrap()` so the loop indeed always terminates: I guess the extra guard might be to protect against a new code path that does neither `output_iter.exclude_queue()` nor `output_iter.next().unwrap()` and thus make the loop infinite.

Are you suggesting to also drop
```
            let output_size = output_iter.size();
            debug_assert!(output_size < last_output_size);
            if output_size >= last_output_size {
                error!(
                    self.log,
                    "{}: Infinite loop detected in StreamBuilder::build_streams @{}.",
                    CRITICAL_ERROR_INFINITE_LOOP,
                    output_size
                );
                self.metrics.critical_error_infinite_loops.inc();
                break;
            }
            last_output_size = output_size;
```
that was not introduced by this PR?

rs/replicated_state/src/replicated_state.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/fd0631ccb3e015705d8a67d2d9c63fce825e2bdb

rs/replicated_state/src/replicated_state.rs
mraszyk: Good point: https://github.com/dfinity/ic/pull/10467/commits/8b33b4de4c25738284727a28270b28b1e937f7c3

rs/replicated_state/src/replicated_state.rs
mraszyk: Good point: https://github.com/dfinity/ic/pull/10467/commits/cba593297dc569a9cec26809b198bb78740fe117

rs/messaging/src/state_machine/tests.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/b4196758b0e29dd5f72ff90df2db1d08b09f88fb

rs/replicated_state/src/replicated_state.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/10467/commits/23c1b52705fe73c85c5c3a9a060891493e61a271

rs/replicated_state/src/replicated_state.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/10467/commits/23c1b52705fe73c85c5c3a9a060891493e61a271

rs/messaging/src/state_machine.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/1f3a28a5253068298f10c9bc6eb32285bff73466

rs/replicated_state/src/replicated_state.rs
mraszyk: And https://github.com/dfinity/ic/pull/10467/commits/8188d5887dc5b0a57d601b2459b5dd5c11771704

rs/messaging/src/state_machine/tests.rs
mraszyk: Good point: https://github.com/dfinity/ic/pull/10467/commits/1f71e91fd1d7cc6d48bbaff0d18a409ba774a4b7

rs/replicated_state/src/replicated_state.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/10467/commits/faee6fbb929fd256769e945c2c339a1aa0473294

rs/messaging/src/routing/stream_builder.rs
mraszyk: > Should we have deleted it with pending guaranteed response calls?

Ensuring that the deleted subnet has no pending guaranteed response calls would complicate the design of subnet deletion substantially: the subnet to be deleted would have to be marked as "preparing for deletion" first so that it rejects all outgoing calls (otherwise, there might always be some pending calls) and then we'd have to wait (potentially indefinitely) until all pending calls are replied. This resembles the flow of stopping a canister and would likely carry over its pitfalls to subnet deletion.

> We'd need to at least weigh that against the potential alert spam resulting from deleting a subnet with pending guaranteed response calls.

Fair point. Here are the options I could think of:
- silence the critical error completely (this PR): we could miss a bug in this case;
- keep the critical error as is: we could get spammy alerts after subnet deletion;
- keep a list of deleted subnets (e.g., derived from the registry) and silence the critical error for responses to such subnets: this introduces quite some extra code complexity.

The spammy alerts can be ignored if replica logs confirm that all violations target the deleted subnet (manual check needed).

rs/replicated_state/src/metadata_state.rs
mraszyk: Makes sense!

rs/messaging/src/routing/stream_builder.rs
alin-at-dfinity: Pretty much, yeah. IMHO this is an unnecessary check that necessarily (since there's an explicit `break` that might technically cause us to not process all outputs) requires all downstream code (such as the code you're adding) to deal with that eventuality.

If we treated this as just another finite iterator and did not second-guess ourselves, then no extra downstream error handling would be necessary. Just rely on the `OutputIterator` unit tests to ensure its correct behavior.

I'm fine with not doing this in this PR, if you feel it doesn't fit. But I would do it eventually.

rs/messaging/src/routing/stream_builder.rs
alin-at-dfinity: I would go with the spammy alerts. As you point out, we are (by necessity, but still) doing a bit of a sloppy job of deleting subnets. So it's only fair that, if we do delete a subnet with open guaranteed response call contexts, we might have to deal with a bit of manual work.

All of the alert spam would be concentrated around the actual subnet deletion. It's not like we might get alerted days after the fact one callback at a time, or something. One could even silence the particular alert as part of the preparation for deleting the subnet and investigate (or not) the logs afterward.

To me this feels like a reasonable trade-off.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -641,7 +643,12 @@ impl StreamBuilderImpl {
                     };
                 }
 
-                // Destination subnet not found.
+                // Destination subnet not found: process the message immediately;
+                // in particular, generate a reject response with the more precise
+                // `RejectCode::DestinationInvalid` for requests
+                // so that `generate_reject_responses_for_deleted_subnets()`
+                // does not generate a reject response with the less precise
+                // `RejectCode::CanisterReject`.
```
alin-at-dfinity: This is also no longer accurate (or rather, complete) considering this earlier [comment](https://github.com/dfinity/ic/pull/10467#discussion_r3498897265).

It is important that `build_streams()` first gets to reject any guaranteed response calls before `generate_reject_responses_for_deleted_subnets()` rejects the corresponding callback, as `build_streams()` does not expect to deal with already enqueued responses (nor should it).

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -872,3 +905,88 @@ impl StreamBuilder for StreamBuilderImpl {
...
+
+/// Generates synthetic reject responses for callbacks to deleted subnets.
+///
+/// Must be called after `build_streams()`: `build_streams()` unconditionally
+/// rejects any request in an output queue with no route (e.g. destined for a
+/// deleted subnet), ensuring such callbacks already have an enqueued response
+/// and are skipped by this function. Calling before `build_streams()`
+/// would result in a critical error in `build_streams()` if `build_streams()`
+/// tried to reject an unbounded-wait request in an output queue to a deleted subnet
+/// for which a reject response has already been generated.
```
alin-at-dfinity:
```suggestion
/// Must be called after `build_streams()`: `build_streams()` unconditionally
/// rejects any request in an output queue with no route (e.g. destined for a
/// deleted subnet), never expecting to deal with duplicate responses. Producing
/// a reject response for an unbounded-wait callback before `build_streams()` has
/// had a chance to produce a reject response for the corresponding request would
/// trigger a critical error in `build_streams()`.
```

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -872,3 +905,88 @@ impl StreamBuilder for StreamBuilderImpl {
...
+                    // Use the same reject code and message as canister uninstallation, but do
+                    // not refund cycles here: the deleted subnet may have partially executed
+                    // the request and consumed some or all of the payment/refund cycles.
+                    response_payload: Payload::Reject(
+                        RejectContext::new_with_message_length_limit(
+                            RejectCode::CanisterReject,
+                            "Canister has been uninstalled.",
+                            MR_SYNTHETIC_REJECT_MESSAGE_MAX_LEN,
+                        ),
+                    ),
```
alin-at-dfinity, line 926: Just curious, why not produce the same `DestinationInvalid` as `build_streams()`?

rs/replicated_state/src/replicated_state.rs
```diff
@@ -36,7 +36,7 @@ use ic_types::{
     time::CoarseTime,
 };
 use ic_types_cycles::{
-    CanisterCyclesCostSchedule, CompoundCycles, CyclesAccountManagerSubnetConfig,
+    CanisterCyclesCostSchedule, CompoundCycles, Cycles, CyclesAccountManagerSubnetConfig,
```
alin-at-dfinity: Nit: Is this still necessary?

alin-at-dfinity (approved):
LGTM, thank you.

rs/messaging/src/routing/stream_builder.rs
mraszyk: `DestinationInvalid` means that the request was not processed (not even partially), but in this case, the request could have been processed partially (because the other subnet pulled the message from the stream, but did not yet signal that back).

rs/replicated_state/src/replicated_state.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/1fbff8c9c798205570f1e0de25555862d2beebb7

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/fd8d1a269b35d1a204047a29053f37c04c82d75e

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/10467/commits/9744b25a931afd3b042772ec3980b438c22218ca

rs/messaging/src/routing/stream_builder.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/10467/commits/b575385d6ad54e7b4c0da6050929c48b1c70d04f

rs/messaging/src/routing/stream_builder.rs
mraszyk: Dropped all the extra complexity (in a separate [PR](https://github.com/dfinity/ic/pull/10623) and this [commit](https://github.com/dfinity/ic/pull/10467/commits/a6c6cfa74f78508f3f39e2a6be66ff8980eb3e6c)).

PR #10469, opened 2026-06-15
feat: make reference subnet size in CyclesAccountManager configurable at runtime

mraszyk:
This PR makes the reference subnet size in CyclesAccountManager configurable at runtime so that a subnet's security can be changed from `None` to SEV-SNP and this change is properly picked up from the registry and reflected in cycles scaling at runtime.

The implementation proceeds by moving the reference subnet size from the (static) `CyclesAccountManagerConfig` to the (dynamic) `CyclesAccountManagerSubnetConfig` derived via `NetworkTopology` in `ReplicatedState` at runtime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18461,6 +18461,7 @@ dependencies = [
  "ic-canonical-state",
  "ic-cdk",
  "ic-certification 0.9.0",
+ "ic-config",
  "ic-crypto-tree-hash",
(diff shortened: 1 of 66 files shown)


schneiderstefan (approved):

gregorydemay (approved):
DEFI changes LGTM!

daniel-wong-dfinity-org-twin (approved):
Approving only for the 4 Governance-owned files:

PR #10470, opened 2026-06-15
fix(registry): support canister migration from/to deleted subnet in registry

mraszyk:
This PR supports canister migration from/to deleted subnet in the registry by:
- adding a test that canister migration from a deleted (source) subnet succeeds (the implementation already supports this case so we just add a test to prevent a regression);
- skipping updating the deleted target subnet in the routing table (the canister migration is still successful since the outcome should be as if the migration completed first and the subnet was only deleted later) and adding a test covering this case.

The motivation for this change is to ensure that the state machine in the migration canister does not get stuck if a subnet is deleted while a canister migration is ongoing. In such a case, the canister migration is supposed to complete with the same outcome as if the canister migration completed before subnet deletion (which is infeasible to always ensure in practice though, e.g., if the deleted subnet was stuck).


mraszyk:
I believe this PR is not worth an entry in the changelog...

rs/registry/canister/src/mutations/do_migrate_canisters.rs
```diff
@@ -39,8 +39,9 @@ impl Registry {
             })
             .collect::<Result<Vec<_>, _>>()?;
 
+        // Intentionally do not validate that the target subnet exists: the migration orchestrator
+        // may still attempt to migrate canisters to/from a subnet that has been deleted.
```
daniel-wong-dfinity-org-twin: migrating to a deleted subnet is confusing (even migrating from a deleted canister is confusing, but less). Can you elaborate (in this comment)? Like, what is the eventual result?

rs/registry/canister/src/mutations/do_migrate_canisters.rs
```diff
@@ -154,6 +165,151 @@ mod test {
...
+            .unwrap();
+        assert_eq!(updated_routing_table, expected_routing_table);
+    }
+
+    #[test]
+    fn test_migrate_canisters_succeeds_if_target_subnet_does_not_exist() {
+        // Verify that migrate_canisters succeeds when the target subnet does not exist
+        // (has no key in subnet_list and no ranges in routing_table).
+        //
+        // We simulate this by using a subnet ID that is not in the routing table to begin with.
```
daniel-wong-dfinity-org-twin: This describes the whole test, right? Such comments should appear BEFORE the test, just like in all other cases: when a (non-inline) comment talks about a thing, the comment appears before that thing (in this case, the test), not within it.

Looks like in the case of tests (e.g. test_basic_migrate_canisters), such comments are not triple slash, even though tests are technically functions.

rs/registry/canister/src/mutations/do_migrate_canisters.rs
```diff
@@ -154,6 +165,151 @@ mod test {
...
+                    end: CanisterId::from_u64(255),
+                },
+                system_subnet_id.into(),
+            )
+            .unwrap();
+        assert_eq!(updated_routing_table, expected_routing_table);
+    }
+
+    #[test]
+    fn test_migrate_canisters_succeeds_if_target_subnet_does_not_exist() {
```
daniel-wong-dfinity-org-twin, line 262: (We discussed in this in Slack, but just so that this lives in the record of GitHub...)

It seems like such requests should be blocked, not result in canister deletion, because that would be a surprising data loss, the worst kind of surprise.

daniel-wong-dfinity-org-twin:
I would say this is a fix, not chore. Why: IIUC, people are already allowed to migrate when deletion is happening, but if they do that, the system ends up in an inconsistent state. Whereas, this change makes it so that such migrations do NOT result in inconsistent state.

Again, chore is for if you don't really care if this gets into production in a timely manner. But it seems to me, that you do care. If you care, don't use chore, because we more or less ignore chore PRs when considering whether an upgrade should be done.

Also, scope to registry.

rs/registry/canister/src/mutations/do_migrate_canisters.rs
mraszyk: https://github.com/dfinity/ic/pull/10470/commits/4f6b9bf905531a31f85516b944a15aa78f214c15

rs/registry/canister/src/mutations/do_migrate_canisters.rs
mraszyk: https://github.com/dfinity/ic/pull/10470/commits/183afe325d0ebc9bfe1c117c3bebbb4f8f2af836

rs/registry/canister/src/mutations/do_migrate_canisters.rs
mraszyk: Subnet deletion is (currently) a manual process and we don't expect to actually delete a subnet while a canister migration is ongoing (instead we expect to disable the canister migration API first and only delete a subnet once there's no ongoing canister migration). However, e.g., if a subnet is permanently stuck, it might be impossible to successfully complete an ongoing canister migration and in this case, we'd like to still be able to delete such a subnet. This leaves us with two options:
- adjust canister migration so that it can gracefully abort canister migration to a subnet that is stuck/deleted (that would very likely result in substantial engineering overhead and risk);
- adjust canister migration so that performing canister migration from/to a stuck/deleted subnet (transparently) behaves as if canister migration completed before the subnet got stuck/deleted.

The second option requires less engineering overhead which seems like a good trade-off given that it is rather unlikely that we'll need this in practice.

daniel-wong-dfinity-org-twin: Ah. I didn't know all of that. It would be good for that information to be in in the comments on lines do_migrate_canister.rs:42-46.

rs/registry/canister/src/mutations/do_migrate_canisters.rs
daniel-wong-dfinity-org-twin: Thanks.

Interesting that you removed that migrating from a deleted subnet is allowed. What happens then? What happens if a migration starts, but while that's going on, the original subnet is deleted? I guess the deletion wins, right? This seems worth documenting.

> i.e., the canister should be removed from the source subnet's canister ranges and not added to the (deleted) target subnet's canister ranges.

So, basically, the canister gets deleted, right? I think this fact should be stated more directly.

mraszyk: > Interesting that you removed that migrating from a deleted subnet is allowed.

It is allowed, but it does not require any changes to the current implementation (this PR only covers that by a new test to prevent regressions). The target subnet is not the source subnet and thus talking about this case in this comment does not make sense.

> What happens then?

The canister migration succeeds and the routing table lists the canister as belonging to the target subnet which is desired.

> What happens if a migration starts, but while that's going on, the original subnet is deleted?

Independently of the interleaving (data race), the final outcome is as if the canister migration completed first and then the subnet was deleted.

> > i.e., the canister should be removed from the source subnet's canister ranges and not added to the (deleted) target subnet's canister ranges.
> So, basically, the canister gets deleted, right?

correct

rs/registry/canister/src/mutations/do_migrate_canisters.rs
mraszyk: https://github.com/dfinity/ic/pull/10470/commits/ba435704ff0698225abcf27f8e8fc25ed07fbfaa

rs/registry/canister/src/mutations/do_migrate_canisters.rs
mraszyk: https://github.com/dfinity/ic/pull/10470/commits/ba435704ff0698225abcf27f8e8fc25ed07fbfaa

rs/registry/canister/src/mutations/routing_table.rs
```diff
@@ -296,22 +296,34 @@ impl Registry {
...
-                routing_table.assign_canister(canister_id, subnet_id);
+                routing_table.assign_canister(canister_id, target_subnet_id);
             }
-            if !target_in_routing_table {
-                // Target subnet has no canister ranges in the routing table: remove
-                // any canister ranges just assigned to it, keeping source subnets split.
-                routing_table.remove_subnet(subnet_id);
+            if !target_subnet_in_routing_table {
+                // Target subnet has no canister ranges in the routing table (e.g., it has
+                // been deleted): remove any canister ranges just assigned to it.
```
daniel-wong-dfinity-org-twin: Any other reason that a subnet would not be in the routing table? I mean, if deleted is the only reason, then, I can understand why remove the subnet. Otherwise, it is a bit confusing why remove immediately after add. Seems like it might be wrong??

In any case, seems to me, if target_subnet_in_routing_table is true, then, you can instead just skip the assign_canister_calls (on lines 320-322), rather than remove after the fact, right? I think skipping is less confusing than adding and then immediately removing.

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/src/mutations/routing_table.rs
mraszyk: If the target subnet is not in the routing table, we still need to remove the canisters from the source subnet's ranges which is done by assign_canister at the moment. I agree that the flow
```
for canister_id in canister_ids {
    if target_subnet_in_routing_table {
                routing_table.assign_canister(canister_id, target_subnet_id);
   } else {
       routing_table.unassign_canister(canister_id);
   }
}
```
might be better readable, but there's currently no method unassign_canister. Would you prefer that I add such a method `RoutingTable::unassign_canister`?

mraszyk: Done: https://github.com/dfinity/ic/pull/10470/commits/2d01883f03133bf08d83e69734b62b665a85ce0d

daniel-wong-dfinity-org-twin: 👍

PR #10413, opened 2026-06-08
feat: bump bazel version to 9.1.1

nmattia:
This bumps our bazel version to the latest available version. Some dependencies had to be bumped for 9.X support, but overall very little changed.

PR #10479, opened 2026-06-15
fix: Fix python bazel wrapper after upgrade

Bownairo:
After https://github.com/dfinity/ic/pull/10413 upgraded `rules_python`, nightly bare metal jobs started failing. This updates how we bazel wrap our tools, to match the new environment format.


frankdavid (approved):
Thanks for fixing it!

PR #10487, opened 2026-06-16
test(systest): eliminate inter-DC network traffic

basvandijk:
## Why

Inter-DC network traffic should be eliminated to ensure networking issues no longer affect system-test stability. There are two primary flows of networking traffic relevant to system-tests:

1. test <-> testnet
2. IC-OS image web-server -> testnet / Farm host

To partially address 1 and 2, 557d7278dcbb0305411c6536645c3a32b4ec64b6 already forced testnet allocation to the DC the *github runner* was hosted in. We would like to have similar behaviour for manually run system-tests from devenvs.

To address 2 we need to ensure images are always served from the cluster that is running the test.

## What

`rs/tests/upload_systest_dep.sh` no longer queries the redirect server (`https://artifacts.idx.dfinity.network`). Instead it checks the **local cluster's** bazel-remote cache directly and uploads the dependency there if it is missing.

To determine the cluster, the script `rs/tests/run_systest.sh` will query the in-cluster K8s API server certificate SAN and extract the cluster name from `api.<cluster>.dfinity.network`.

It will then also set the `DC` environment variable based on that cluster (if it hasn't been set already). This variable determines where the testnet will be allocated to. This means `bazel/workspace_status.sh` no longer has to define that variable.


rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,88 @@
...
+}
+
+# Determines the name of the local cluster (e.g. "zh1-idx1"), used to build the
+# download URL served at artifacts.<cluster>.dfinity.network.
+resolve_cluster() {
+    # Allow an explicit override (e.g. for manual runs or unusual environments).
+    if [ -n "${SYSTEST_UPLOAD_CLUSTER:-}" ]; then
+        echo "$SYSTEST_UPLOAD_CLUSTER"
+        return
     fi
```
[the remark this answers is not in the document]
basvandijk: Done in 2dccaa2520. `resolve_cluster` now validates both the `SYSTEST_UPLOAD_CLUSTER` override and the auto-detected name against `^[a-z0-9]([a-z0-9-]*[a-z0-9])?$` before it is interpolated into the URL, rejecting whitespace, `/`, `@`, etc. with a clear error.

rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,88 @@
...
 
-    local result_url
-    result_url=$(cut -d' ' -f2 <<<"$result")
-    if [ -z "$result_url" ]; then
-        echo "Looking up dependency '$1' did not return a URL, got: '$result'" >&2
+    # Otherwise auto-detect from the in-cluster Kubernetes API server certificate,
+    # which carries a SAN of the form 'api.<cluster>.dfinity.network'. This works
+    # both on CI runners and in devenvs, and yields the exact cluster name.
+    local cluster=""
+    cluster=$(openssl s_client -connect kubernetes.default.svc:443 </dev/null 2>/dev/null \
```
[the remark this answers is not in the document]
basvandijk: Done in 2dccaa2520. The probe is now wrapped in `timeout 15`, so DNS/network stalls fail fast instead of hanging CI.

rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,88 @@
...
+        | openssl x509 -noout -text 2>/dev/null \
```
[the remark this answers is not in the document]
basvandijk: Done in 2dccaa2520. Added an explicit `command -v openssl` check that exits with an actionable message (suggesting `SYSTEST_UPLOAD_CLUSTER`) before attempting detection.

rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,88 @@
...
+        | grep -oE 'api\.[a-z0-9][a-z0-9-]*\.dfinity\.network' \
+        | head -n1 \
+        | sed -E 's/^api\.(.*)\.dfinity\.network$/\1/') || true
```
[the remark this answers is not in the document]
basvandijk: Switched to `grep -m1` and dropped `head` in 2dccaa2520. I kept a (now documented) `|| true`, but for a different reason than SIGPIPE: `openssl s_client` exit status is unreliable here (it can reflect verification of the internal CA rather than connectivity), so under `pipefail` the pipeline may be non-zero even on success. Real failures are no longer masked because the extracted name is now validated against the cluster-name pattern and a mismatch/empty result produces a clear "could not determine the local cluster name" error.

rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,106 @@
...
 
-    if [ "$result_code" != "307" ]; then
-        echo "Expected 404 or 307 when looking up dependency '$1', got '$result_code'" >&2
+    # Otherwise auto-detect from the in-cluster Kubernetes API server certificate,
+    # which carries a SAN of the form 'api.<cluster>.dfinity.network'. This works
+    # both on CI runners and in devenvs, and yields the exact cluster name.
+    if ! command -v openssl >/dev/null; then
+        echo "openssl not found; cannot auto-detect the cluster name, set SYSTEST_UPLOAD_CLUSTER explicitly" >&2
         exit 1
     fi
```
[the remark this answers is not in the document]
basvandijk: Let's just assume coreutils like `timeout` are available.

rs/tests/upload_systest_dep.sh
```diff
@@ -2,66 +2,106 @@
...
+        *)
+            echo "Unexpected HTTP code '$code' when looking up dependency '$key' at '$url'" >&2
+            exit 1
+            ;;
+    esac
+}
+
+# Determines the name of the local cluster (e.g. "zh1-idx1"), used to build the
+# download URL served at artifacts.<cluster>.dfinity.network.
+resolve_cluster() {
```
nmattia: since this script is called from `run_systest.sh`, and since `run_systest.sh` also tries to infer the DC, and since the two results should be aligned (AFAIU), how about taking the cluster as an input and let the caller specify the DC/cluster? or is there a reason why in one case we look it up from the Farm metadata, and once from the k8s records?

rs/tests/testnets/nns_recovery.rs
```diff
@@ -5,7 +5,9 @@
 // The bazel target `mainnet_nns_recovery` also uses this testnet by setting the
 // `USE_MAINNET_STATE` environment variable to true, which makes the testnet use mainnet state.
 //
-// You can pass `--set-required-host-features=dmz` to make the testnet open to the Internet.
+// You can pass `--set-required-host-features=dmz` to make the testnet open to the Internet. If you do also pass:
+// `--test_env=DC=zh1-dmz` to force the testnet to be deployed to that DC.
```
mraszyk: The sentence starting with "if" reads a bit weird.

basvandijk: Agreed. I added a comma. Does that read better?

rs/tests/upload_systest_dep.sh
basvandijk: Done in https://github.com/dfinity/ic/pull/10487/changes/b627499df04bab6d918ec10ebac428cda0c0e36d.

rs/tests/testnets/nns_recovery.rs
mraszyk: Thank you! Now it makes sense. I didn't want to be picky, but instead of (If you do,)(also pass) I parsed it as (If you do also pass) treating "do" as emphasis and then it didn't make sense to me.

mraszyk (approved):

rs/tests/testnets/nns_recovery.rs
basvandijk: Yeah I had the same when I read it back. Thanks!

NikolaMilosa (approved):

nmattia (approved):
Nice

PR #10489, opened 2026-06-16
feat: Add delivered HTTP call contexts to subnet call context manager 

eichhorl:
Currently, all HTTP request contexts are stored in the same collection of the subnet call context manager. They are removed once consensus delivers a response to the canister.

In the future, there will be a second "asynchronous refund" phase as part of the context's life cycle: After consensus delivers the HTTP response and an _initial_ refund, consensus may continue to deliver _asynchronous_ refunds, that correspond to shares which didn't make it into the original response.

To do this, the context needs to persist even after a response was delivered.

This PR proposes to support this by introducing a second collection, which will hold such delivered contexts. Whenever a response is delivered, the corresponding context is moved into the new collection.

An alternative would have been to mark the context as "delivered", e.g. by setting a boolean flag. However, introducing a second collection has some natural advantages:
1. The size of the collection holding undelivered contexts is limited. Including delivered contexts in the same collection would therefore adversely affect throughput.
2. Some places will only want to look at undelivered contexts (I.e. response share aggregation & validation).
3. Some places will only want to look at delivered contexts (I.e. async refund creation & validation)
4. Some places will want to look at both collections (I.e share validation & purging), for which we can simply take the union of both collections.

Note that for compatibility reasons, the new collection is still unused.

### DIFF
--- a/rs/protobuf/def/state/metadata/v1/metadata.proto
+++ b/rs/protobuf/def/state/metadata/v1/metadata.proto
@@ -306,6 +306,7 @@ message SubnetCallContextManager {
   repeated ReshareChainKeyContextTree reshare_chain_key_contexts = 17;
   repeated SignWithThresholdContextTree sign_with_threshold_contexts = 18;
   repeated PreSignatureStashTree pre_signature_stashes = 19;
+  repeated CanisterHttpRequestContextTree delivered_canister_http_request_contexts = 20;
 }
 
 message SubnetMetrics {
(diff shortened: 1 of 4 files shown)


schneiderstefan (approved):
This PR LGTM, as it's just some protobuf to/froms, but I have some high-level questions on where this is going.

At what point do we remove items from this list? Is there a bound on how much it can grow and/or can a malicious replica force that these entries are kept for an unreasonable amount of time?

eichhorl:
> This PR LGTM, as it's just some protobuf to/froms, but I have some high-level questions on where this is going.
> 
> At what point do we remove items from this list? Is there a bound on how much it can grow and/or can a malicious replica force that these entries are kept for an unreasonable amount of time?

@schneiderstefan the asynchonous refund phase will last for one minute, after which the context will be removed entirely. Replicas that haven't submitted a refund share until then are assumed to be offline/behind and their entire allowance will be refunded.

Currently we target 100 HTTP requests per second, which would lead to 6000 contexts over one minute: https://github.com/dfinity/ic/blob/bfcd00b670601d1d16f40dc59874ade30c779b6b/rs/config/src/execution_environment.rs#L201-L204

PR #10490, opened 2026-06-16
feat: hoist e2fsdroid (from platform-tools) from Dockerfile to Bazel

nmattia:
Hoist e2fsdroid out of the Dockerfile and into the Bazel build, like we did for e2fsprogs (mke2fs), dosfstools and mtools.

This will still pull the prebuilt binary; we don't build it from source (yet).

Instead, this pins the platform-tools archive directly via http_archive and exposes just the prebuilt e2fsdroid binary as //:e2fsdroid. Wire it into the ext4_image rule and build_ext4_image.py the same way as //:mkfs.ext4, so the image build no longer relies on e2fsdroid being on PATH.


nmattia:
Requires https://github.com/dfinity/ic/pull/10488 to be merged first

basvandijk (approved):

mbjorkqvist (approved):

Bownairo (approved):

tmu0 (approved):

PR #10492, opened 2026-06-16
feat(registry): make input fields in SubnetFeatures of type opt bool in registry canister

mraszyk:
This PR makes the input fields in `SubnetFeatures` of type `opt bool` (instead of `bool`) in the registry canister. The motivation for this change is to allow dropping unused fields (such as `canister_sandboxing`): as long as the registry canister expects `bool` for some field, dropping that field would break tooling and tests targetting the mainnet version of the registry canister which still has that field of type `bool`.


rs/registry/canister/canister/registry.did
```diff
@@ -86,9 +86,15 @@ type ResourceLimits = record {
   maximum_state_delta : opt nat64;
 };
 
+type SubnetFeaturesInput = record {
```
daniel-wong-dfinity-org-twin: A reader is not going to readily understand what "Input" is supposed to indicate. OTOH, I don't have any better ideas.

rs/registry/canister/src/mutations/do_create_subnet.rs
```diff
@@ -325,6 +327,96 @@ pub struct CreateSubnetPayload {
     pub gossip_retransmission_request_ms: u32,
 }
 
+/// Input type for `create_subnet` canister method.
+/// Same as `CreateSubnetPayload` but with `SubnetFeaturesInput` (all `Option<bool>`) for `features`.
+#[derive(Clone, Eq, PartialEq, Debug, Default, CandidType, Deserialize, Serialize)]
+pub struct CreateSubnetArg {
```
daniel-wong-dfinity-org-twin: This a violent departure from our very consistent pattern: the input type for a method named dance_the_foxtrot method is called DanceTheFoxtrotPayload. Why does this super similar type need to be introduced?? Seems like you can simply change CreateSubnetPayload. This seems like a massive DRY violation. I really hope we can avoid this.

daniel-wong-dfinity-org-twin (commented):
**Aside**: I did not know you can do this in Candid. Maybe, I used to know, but to avoid trauma, actively forgot it. IIUC, if someone sends true, but you want an opt bool, then decoding treats the true just like opt true (and likewise for false)... And yet, you cannot go the other way; if someone send you opt true, but all you want is a bool, you will NOT get a true, but rather, you will just fail to decode (and ditto for opt false)... This is why I hate Candid. I mean, I can understand why there is no "natural" value when you receive a null, but you want some other (non-opt) type. Well, Protocol Buffers has no problem with this, because every type has a "default" value, and that's what you end up with if you receive nothing when you want something... Anyway, this wacky inconsistency is tangential to your PR (hence "aside")...

rs/registry/canister/src/mutations/do_create_subnet.rs
mraszyk: The reason for a new type is that the original type is still used in tooling (e.g., ic-admin) and tests that target the mainnet version of the registry canister and thus must use the original type.

rs/registry/canister/canister/registry.did
mraszyk: I could use `SubnetFeaturesV2` if that sounds better.

daniel-wong-dfinity-org-twin:
This is a feat, not a chore, because it enables new calls. More precisely, it allows callers to omit a couple of fields.

chore means that you do not care if this gets deployed in a timely way.

daniel-wong-dfinity-org-twin:
Conventional Commit prefix should have scope. Here, the choice is pretty obvious: registry.

rs/registry/canister/canister/registry.did
mraszyk: https://github.com/dfinity/ic/pull/10492/commits/f4061ff03e790e6243606f283c8366d5105ff37f

rs/registry/canister/src/mutations/do_create_subnet.rs
mraszyk: Renamed to V2: https://github.com/dfinity/ic/pull/10492/commits/e688a3997ac68b889fcb3b969c3c3548e894d2eb

rs/registry/canister/canister/registry.did
daniel-wong-dfinity-org-twin: V2 is wrong, because it does not supersede SubnetFeatures; SubnetFeatures is still current.

Idea: RequestedSubnetFeatures. When requesting features, it is natural that you need not explicitly mention ALL features (only the ones you care about).

Differences between my suggestion and "Input":

1. SubnetFeatures is the base idea, so it should remain at the end, not be suffixed.
2. At least in my experience, "request" is used more specifically in the context of RPCs. Whereas, "input" could be a command line flag or a function argument.

Ideally, we would also rename the existing SubnetFeatures to something like InEffectSubnetFeatures, but that is tangential, and disruptive. Similarly, it is natural that regardless of past requests, it must be that ALL features are either enabled or disabled.

rs/registry/canister/canister/registry.did
```diff
@@ -86,15 +86,15 @@ type ResourceLimits = record {
...
 
-type SubnetFeaturesInput = record {
+type SubnetFeaturesV2 = record {
   canister_sandboxing : opt bool;
   http_requests : opt bool;
   sev_enabled : opt bool;
 };
 
-type CreateSubnetPayload = record {
+type CreateSubnetPayloadV2 = record {
```
daniel-wong-dfinity-org-twin, line 95: Please, do not append "V2". Why not: the new version is "just an evolution" of the old one in a precise and technical way: a person who sends a request based on the previous version will still get the same/desired effect. In such cases, the name is generally kept the same, and there is no special reason to depart from that in this case.

rs/registry/canister/canister/registry.did
```diff
@@ -499,10 +500,10 @@ type UpdateSshReadOnlyAccessForAllUnassignedNodesPayload = record {
   ssh_readonly_keys : vec text;
 };
 
-type UpdateSubnetPayload = record {
+type UpdateSubnetPayloadV2 = record {
```
daniel-wong-dfinity-org-twin, line 503: Ditto.

rs/registry/canister/src/mutations/do_create_subnet.rs
daniel-wong-dfinity-org-twin: > tests that target the mainnet version

This can be addressed by letting such tests (and ic-admin?) use a legacy definition, which can be deleted once sufficiently new code is in production. Ideally, this legacy definition is sequestered in some other file, since it is not needed by the main implementation; rather, it is only needed by code that wants to ENCODE using the old definition in order to make canister method calls.

With Arg + Payload, it looks like the two are supposed to both exist indefinitely. But their roles are more or less the same, and so, we should minimize having both of them. If the overlap temporarily, that's ok.

mraszyk: > This can be addressed by letting such tests (and ic-admin?) use a legacy definition, which can be deleted once sufficiently new code is in production.

This is indeed the idea: the canister uses V2 and every tool uses the old version for now.

> Ideally, this legacy definition is sequestered in some other file, since it is not needed by the main implementation

which file would you suggest? and what you suggest to use the same type name and only distinguish via the crate (qualification) or renaming (use ...::CreateSubnetPayload as CreateSubnetPayloadV2)?

rs/registry/canister/canister/registry.did
mraszyk: > Idea: RequestedSubnetFeatures. When requesting features, it is natural that you need not explicitly mention ALL features (only the ones you care about).

The issue with this suggestion is that `http_requests` are enabled by default so leaving `None` in the payload would default to `true` (and this makes sense in my opinion to prevent surprises).

PR #10495, opened 2026-06-16
feat: Drop final active blessed version components

Bownairo:

This removes the tracking of this list from the registry canister, and updates the last few components in sync. All uses have been cleared in this repo, and DRE tools have been updated as well.

What remains is cleaning up this record from the live registry, and a few miscellaneous followups.


basvandijk (approved):

schneiderstefan (approved):

daniel-wong-dfinity-org-twin (approved):
TBH, I do not well understand what's going on here, but since this is mostly deletions, it seems harmless (hence approve). Ideally though, the title and description would better explain what we are getting rid of (and maybe also, why).

PR #10497, opened 2026-06-17
chore: unify optional Prometheus data download in performance tests

basvandijk:
## Why

Each performance test implemented its own logic to optionally download the Prometheus data directory of the `PrometheusVm` to the local test directory. The triggers were inconsistent:

- `consensus_performance` and `tecdsa_performance_test_template` gated on the `DOWNLOAD_P8S_DATA` env variable.
- `p2p_performance_test` and `fill_execution_rounds_workload` always downloaded via a `const DOWNLOAD_PROMETHEUS_DATA = true`.
- `cloner_canister_workload` downloaded unconditionally.

The pre-download cooldown sleeps also differed (300s / 100s / 1h / none) and were hard-coded in the test bodies.

## What

- Add a single teardown to `SystemTestGroup` (in `group.rs`) that is registered automatically when `DOWNLOAD_P8S_DATA` is set to `true` or `1`, mirroring how `consensus_performance` worked before. It optionally waits `DOWNLOAD_P8S_DATA_COOLDOWN_SECS` seconds (default `0`) so Prometheus has time to scrape the final metrics, then calls `env.download_prometheus_data_dir_if_exists()`.
- Remove the bespoke download logic (and the now-unused `HasPrometheus` imports) from `consensus_performance`, `tecdsa_performance_test_template`, `p2p_performance_test`, `fill_execution_rounds_workload` and `cloner_canister_workload`.
- Move the per-test cooldowns into the corresponding Bazel targets via `DOWNLOAD_P8S_DATA_COOLDOWN_SECS` (tecdsa: `300`, p2p: `3600`, fill_execution_rounds: `100`).

## Behavioural notes

- `p2p_performance_test`, `fill_execution_rounds_workload` and `cloner_canister_workload` now only download when `DOWNLOAD_P8S_DATA` is set, instead of always/unconditionally.
- `tecdsa_performance_test_template` now always runs its success-rate assertion (previously it was skipped when downloading p8s data). The teardown still downloads even if the assertion fails.
- Removed the redundant `TESTNET_LIFETIME_AFTER_SETUP` keepalive sleep from `cloner_canister_workload`.


rs/tests/driver/src/driver/group.rs
```diff
@@ -689,6 +690,51 @@ pub fn assert_no_critical_errors(env: &TestEnv) {
...
+                "Waiting {cooldown_secs} seconds before downloading the prometheus data \
+                such that prometheus has time to scrape the final metrics ..."
+            );
+            std::thread::sleep(Duration::from_secs(cooldown_secs));
+        }
+        env.download_prometheus_data_dir_if_exists();
+        env.emit_report(String::from(
+            "Downloaded prometheus data to 'prometheus-data-dir.tar.zst' in the test output \
+            directory. You can now use `rs/tests/run-p8s.sh` script to play with the metrics",
+        ));
```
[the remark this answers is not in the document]
basvandijk: Good catch. Fixed in 634d713e84: `download_prometheus_data_dir_if_exists` now returns a `bool` indicating whether it actually downloaded the data directory. When no PrometheusVm was deployed it returns `false`, and the teardown emits a report explaining that nothing was downloaded and that metrics need to be enabled for the test.

schneiderstefan (approved):

nmattia (approved):

rs/tests/driver/src/driver/group.rs
```diff
@@ -689,6 +690,58 @@ pub fn assert_no_critical_errors(env: &TestEnv) {
     );
 }
 
+/// Environment variable that, when set to `true` or `1`, makes every
+/// `SystemTestGroup` automatically download the Prometheus data directory of the
+/// `PrometheusVm` (if one was deployed) to the local test directory on teardown.
+const DOWNLOAD_P8S_DATA_ENV: &str = "DOWNLOAD_P8S_DATA";
```
nmattia: nit: this just creates one level of indirection (don't think the env var is used elsewhere)

basvandijk: I'll inline it.

PR #10498, opened 2026-06-17
docs: add missing firewall rule scope in `ic-admin` docs

pierugo-dfinity:
This tiny PR adds `api_boundary_nodes` as possible scope when fetching firewall rules in `ic-admin`'s documentation (see [source](https://sourcegraph.com/r/github.com/dfinity/ic@58168c8e1c6373a5e439729af87ddebebe369622/-/blob/rs/registry/keys/src/lib.rs?L192)).


daniel-wong-dfinity-org-twin (approved):
I am not sure if this change is correct, but it is "harmless" (since it's "just" a comment change). Therefore, I'm approving on the assumption that you know what you are doing.

PS: It is more likely that I will see your code review request in a timely manner if you send a message to the #eng-nns Slack channel. Thanks.

PR #10499, opened 2026-06-17
refactor(orchestrator): prepare running `ic-gateway` as a side-car to the replica in cloud engines

pierugo-dfinity:
Cloud engines should be self-contained, but today their nodes rely on external routing to reach the replica. This PR prepares running `ic-gateway` as a side-car alongside the replica on each node, so the gateway terminates external traffic locally and proxies it to the replica.
Launching `ic-gateway` is still gated behind a flag and only happens on cloud engines, so there's no behavioral change for now. To enable the functionality, it should be enough to flip that flag.

### Refactoring the orchestrator's process manager
The orchestrator's process management was written around a single process (the replica), with a parallel, partially-duplicated path for `ic-boundary`. Adding a third managed process the same way would have meant copying that logic again.
Instead, process management is now generic. The upgrade loop no longer knows what it's running — it just starts and stops a manager that owns the set of node processes. Adding a future process is a matter of describing how it's built and registering it; no changes to the upgrade loop.

**Scope**: this refactor is confined to the upgrade loop, i.e. processes whose lifetime is tied to a subnet's. Unassigned nodes and API BNs are unaffected.

### Testing
A new system test `cloud_engine_ic_gateway_test` (now still `manual` because the feature flag is disabled) hits `api/v2/status` on port 80 of each node, confirming `ic-gateway` is up and correctly proxying to the replica.
Moreover, new unit tests were added to test the existing `ic-boundary` logic that restarts the process on domain name changes.

<details>
<summary>Implementation details for reviewers</summary>

- The upgrade loop now holds a `MultipleProcessesManager` and calls `start_all`/`stop_all`. It's process-agnostic, except for `stop_replica` which is (arguably, see [comment](https://github.com/dfinity/ic/pull/10499#discussion_r3450746627)) still needed during recoveries.
- All per-process logic (including `ic-boundary`) moved into `processes.rs`. Each `Process` declares how it's built via `build(&Self::Config, Self::Args)`, separating static config from dynamic arguments that can change across the orchestrator's lifetime, likely derived from the registry.
- `ProcessManager<Process>` centralizes the logging, metrics, and `OrchestratorResult` handling that was previously duplicated. `MultipleProcessesManager` holds several of these and decides what to start/stop (e.g. `ic-gateway` only on cloud engines).
- `ic-boundary` is an exception as it has some additional logic when the node's domain changes. Keep that logic by introducing `IcBoundaryManager`, a wrapper over `ProcessManager<IcBoundaryProcess>` exposing `ensure_ic_boundary_running_and_restarted_on_domain_change`.

**To add a new process**: define `NewProcessConfig` and `NewProcess` implementing `Process`, add a `ProcessManager<NewProcess>` field to `MultipleProcessesManager`, and wire it into the start/stop methods.
</details>


rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,508 @@
...
+        let inner = ProcessManager::new(config, metrics, logger.clone());
+        Self {
+            inner,
+            registry,
+            current_domain_name: None,
+            logger,
+        }
+    }
+
+    // TODO: consider returning an error when thing fail instead of logging
```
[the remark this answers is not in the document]
pierugo-dfinity: Done

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,508 @@
...
+// the node's configuration in the registry.
+// ---------------------------------------------------------------------------
+
+/// Whether the orchestrator is currently allowed to actually launch
+/// `ic-gateway`. CloudEngine nodes *should* run `ic-gateway` (see
+/// [`should_run_ic_gateway`]), but the launch is gated off for now while the
+/// rollout is being prepared. To trigger it later, flip this to `true` (and
+/// re-enable the `cloud_engine_ic_gateway_test` system test by removing its
+/// `manual` tag in `rs/tests/consensus/orchestrator/BUILD.bazel`).
+const IC_GATEWAY_LAUNCH_ENABLED: bool = false;
```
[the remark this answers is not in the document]
pierugo-dfinity: Done

rs/tests/driver/src/driver/test_env_api.rs
```diff
@@ -1138,9 +1138,12 @@ impl IcNodeSnapshot {
...
             for (name, value) in metrics {
+                // Assume the metrics to check are prefix-free. This allows to specify a metric name
+                // prefix to check all metrics with that prefix.
                 let max_value = metrics_to_check
-                    .get(name.split('(').next().unwrap())
-                    .copied()
+                    .iter()
+                    .find(|(metric_name, _)| name.starts_with(**metric_name))
+                    .map(|(_, max_value)| *max_value)
                     .unwrap_or_default();
```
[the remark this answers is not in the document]
pierugo-dfinity: This is precisely why the documentation of `default_[replica/orchestrator]_metrics` mentions that the metrics should be prefix-free. Though I'm happy to revisit if the owners (IDX) think otherwise.

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,508 @@
...
+                | Some(SubnetType::System)
+                | Some(SubnetType::VerifiedApplication) => {
+                    self.ic_gateway_manager.stop()?;
+                }
+                Some(SubnetType::CloudEngine) => {
+                    self.ic_gateway_manager.ensure_running(replica_version)?;
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
pierugo-dfinity: The feature flag is there to preserve the current behaviour of the orchestrator: manage only the replica and be agnostic of the new `ic-gateway`. Also, the proposed change would have no impact because calling `stop` without having called `start` before is a no-op. We call it for non-cloud engine nodes just as a sanity check.

> a rollout toggle

This would reboot the VM, i.e. kill all processes.

> a manual start

In that case, this would not help either, because `stop` only cares about what **the orchestrator** started, by keeping track of the PID on start. If it was started manually, `stop` would not stop it. Also, it does not really make sense to support that because: no honest nodes would start `ic-gateway` manually. Thus, only malicious ones could. But if they're malicious, they could modify the orchestrator's code to their will, for example by removing the proposed change.

Still updated the docs to not mention "keep it stopped".

rs/orchestrator/src/upgrade.rs
```diff
@@ -632,48 +594,30 @@ impl Upgrade {
             );
             // Restarting the replica is enough to pass the unsigned CUP forward.
             // If we fail, restart the current process instead.
-            if let Err(e) = self.stop_replica() {
+            if let Err(e) = self.processes_manager.write().unwrap().stop_replica() {
```
pierugo-dfinity, line 595: Stopping only the replica should be enough, but I am tempted to just restart all children instead for simplicity. WDYT?

rs/orchestrator/src/upgrade.rs
```diff
@@ -1729,20 +1734,13 @@ mod tests {
...
+        // should mean that the child processes were started by a previous iteration of the upgrade
         // loop.
-        fn was_replica_process_started_previously(&self) -> bool {
+        fn were_child_processes_started_previously(&self) -> bool {
             matches!(
                 self.initial_subnet_assignment,
                 SubnetAssignment::Assigned(_)
-            )
-            && self.has_local_cup.is_some()
-            // TODO(CON-1630): After mocking the process management, we can remove the condition below.
```
pierugo-dfinity, line 1738: > After mocking the process management

This was done in #9528, so now we can remove the TODO (orthogonal to the changes of this PR).

rs/orchestrator/src/upgrade.rs
```diff
@@ -2734,10 +2777,8 @@ mod tests {
         #[values(
             None,
             Some(RegistryVersion::from(5)),
-            Some(RegistryVersion::from(10)),
             Some(RegistryVersion::from(50)),
-            Some(RegistryVersion::from(100)),
-            Some(RegistryVersion::from(150))
```
pierugo-dfinity, line 2738: Here and below, I removed some (imo redundant) test cases to decrease the cardinality of the Cartesian product which was starting to get a bit big.

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,729 @@
...
+                | Some(SubnetType::Unspecified)
+                | Some(SubnetType::Application)
+                | Some(SubnetType::System)
+                | Some(SubnetType::VerifiedApplication) => {
+                    self.ic_gateway_manager.stop()?;
+                }
+                Some(SubnetType::CloudEngine) => {
+                    self.ic_gateway_manager.ensure_running(replica_version)?;
+                }
+            }
```
eichhorl, line 482: I guess it's fine to return an error here (even if the replica above worked) because this is called near the very end of the upgrade loop? I think it's a good idea to not let a failing gateway interfere with the replica too much

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,729 @@
...
+    pub(crate) fn stop_replica(&mut self) -> OrchestratorResult<()> {
+        self.replica_manager.stop()
+    }
+
+    /// Stop every managed process.
+    pub(crate) fn stop_all(&mut self) -> OrchestratorResult<()> {
+        self.ic_gateway_manager.stop()?;
+        self.replica_manager.stop()?;
+
+        Ok(())
```
eichhorl: Also here: if stopping the gateway fails, we wouldn't even try to stop the replica. Maybe something like this?

```suggestion
        let ic_gateway_result = self.ic_gateway_manager.stop();
        let replica_result = self.replica_manager.stop();

        ic_gateway_result.and(replica_result)
```
(Assuming returning an error at all is a good idea if only the gateway fails)

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,729 @@
...
+        }
+    }
+
+    pub(crate) fn stop(&mut self) -> OrchestratorResult<()> {
+        self.inner.stop()
+    }
+}
+
+// ---------------------------------------------------------------------------
+// MultipleProcessManager
```
eichhorl:
```suggestion
// MultipleProcessesManager
```

rs/orchestrator/src/process_manager.rs
```diff
@@ -27,22 +30,34 @@ pub(crate) trait Process {
...
     /// Return the version of the [`Process`]
     fn get_version(&self) -> &Self::Version;
 
     /// Return the path to the binary of the [`Process`]
-    fn get_binary(&self) -> &Path;
+    fn get_binary(&self) -> PathBuf;
 
     /// Return the arguments passed to the [`Process`]
-    fn get_args(&self) -> &[OsString];
+    fn get_args(&self) -> Vec<OsString>;
```
blind-oracle, line 53: nit: do we really need to use `OsString`? I guess we don't really have any nulls or whatever in our arguments, to maybe we can just use normal string here, and pass them to `process` to be converted to `OsString` when we spawn it?

Same below in `get_env`.

rs/orchestrator/src/upgrade.rs
blind-oracle: Should be ok

rs/orchestrator/src/processes.rs
pierugo-dfinity: > because this is called near the very end of the upgrade loop?

Yes, I was careful to start the replica before anything else. 

> I think it's a good idea to not let a failing gateway interfere with the replica too much

I agree. Though failing to start `ic-gateway` should imo still be considered an error as it would be a sign of broken/unexpected internal state.

rs/orchestrator/src/processes.rs
pierugo-dfinity: Yes, I also thought about that. I first though of stopping the replica first, to be as least intrusive as possible (similar to `start_all`), but then it felt wrong because I thought it would be better to be consistent in that started processes should be stopped in reverse order (here, `ic-gateway` _depends_ on the replica, so we start it second and stop it first).
Your suggestion works too, but I think in that case, we should be consistent with `start_all` and have the same logic there (of trying to start `ic-gateway` even if the replica failed to start).

> Assuming returning an error at all is a good idea if only the gateway fails

I think it is because returning an error prevents from removing the state and CUP, meaning that on the next iteration of the loop, we will try to stop the processes again. If we did not return an error, we would remove the CUP on disk and not retry, potentially leaving an `ic-gateway` running indefinitely.

rs/orchestrator/src/processes.rs
pierugo-dfinity: Thanks

rs/orchestrator/src/processes.rs
pierugo-dfinity: Whatever we do, we should probably still gate it: https://github.com/dfinity/ic/pull/10499/commits/45aeab234bbb1840cc9abc842416c44a8edf87b4

rs/orchestrator/src/upgrade.rs
```diff
@@ -118,13 +89,12 @@ impl RegistryReplicatorForUpgrade for RegistryReplicator {
 pub(crate) struct Upgrade {
     pub registry: Arc<RegistryHelper>,
     pub metrics: Arc<OrchestratorMetrics>,
-    replica_process: Arc<Mutex<dyn ProcessManager<ReplicaProcess>>>,
+    processes_manager: Arc<RwLock<MultipleProcessesManager>>,
```
blind-oracle, line 92: Why `RwLock` here? I see only `write()` locks I think

rs/orchestrator/src/process_manager.rs
pierugo-dfinity: The motivation for using `OsString`s is that some arguments (like the replica's `--config-file` and `--catch-up-package`) are `PathBuf`s, meaning that we would have to call either `.to_string_lossy()` or `.into_os_string().into_string().unwrap()`, none of which are really satisfactory.

`get_env` does not have this problem, but we kept `OsString` for consistency.

rs/orchestrator/src/upgrade.rs
pierugo-dfinity: The lock is shared with the `OrchestratorDashboard` which `read`s it

rs/orchestrator/src/args.rs
```diff
@@ -27,11 +27,19 @@ pub struct OrchestratorArgs {
...
+    #[clap(long)]
+    pub(crate) ic_boundary_env_file: PathBuf,
+
+    /// The path to the IC gateway environment file
+    #[clap(long)]
+    pub(crate) ic_gateway_env_file: PathBuf,
+
     /// The path to the Replica binary location containing the following in case
     /// of guest OS deployment: version.txt, manageboot.sh, replica,
     /// install-upgrade.sh
```
eichhorl: I guess it also contains the ic-boundary/gateway env files?

rs/orchestrator/src/processes.rs
```diff
@@ -0,0 +1,731 @@
...
+                {
+                    warn!(
+                        self.logger,
+                        "Failed to start {}: {}",
+                        IcBoundaryProcess::NAME,
+                        err
+                    );
+                }
+
+                self.current_domain_name = Some(domain_name);
```
eichhorl: Should we only update the current domain name, if we actually succeeded in restarting the process?

rs/tests/driver/src/driver/test_env_api.rs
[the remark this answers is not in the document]
eichhorl: Or we could add an assert making sure that the metrics are indeed prefix free, so that it would fail fast.

rs/orchestrator/src/process_manager.rs
```diff
@@ -166,6 +184,16 @@ impl<P: Process + Send> ProcessManager<P> for ProcessManagerImpl<P> {
...
+                // Put the child into a new process group of which it is the
+                // leader (PGID == PID). Any sub-processes it spawns inherit this
+                // group, which lets `kill()` reliably signal the whole group by
+                // negating the PID. We establish the group here, in the
+                // orchestrator, rather than relying on each managed binary to
+                // call `setpgid` itself. This is equivalent to `setpgid(0, 0)`
+                // run in the forked child before `exec`, while it is still in
+                // the orchestrator's SELinux domain -- which is permitted to set
+                // its own process group.
+                .process_group(0)
```
eichhorl, line 196: Then do we still need to call `setpgid` in the replica and boundary node?

rs/orchestrator/src/upgrade.rs
```diff
@@ -456,7 +428,7 @@ impl Upgrade {
             )
             .await
             .map_err(OrchestratorError::FileDownloadError)?;
-        if let Err(e) = self.stop_replica() {
+        if let Err(e) = self.stop_children() {
             // Even though we fail to stop the replica, we should still
             // replace the registry local store, so we simply issue a warning.
             warn!(self.logger, "Failed to stop replica with error {:?}", e);
```
eichhorl: This seems outdated

eichhorl (approved):

rs/orchestrator/src/args.rs
pierugo-dfinity: The env files are given in separate arguments above, but it indeed contains the binaries. Updated the docs thanks. https://github.com/dfinity/ic/pull/10499/commits/5b993db15bcc69fbee99332bbaacf235c871a5ad

rs/orchestrator/src/processes.rs
pierugo-dfinity: This is different from the current implementation, but I agree this is more robust, thanks! https://github.com/dfinity/ic/pull/10499/commits/55e2587a3d3ba797cd551afe9ee3d8ef7a709f1d
PTAL, I also modified to clear it when succeeding to _stop_

rs/tests/driver/src/driver/test_env_api.rs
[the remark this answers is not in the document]
pierugo-dfinity: Did that 👍 https://github.com/dfinity/ic/pull/10499/commits/a22c0ded19fde18394fa3bff16a2554a35a33b93

rs/orchestrator/src/process_manager.rs
pierugo-dfinity: Indeed, I don't think we should. That said, given how long ago that `setpgid` call was introduced, I wouldn't be confident claiming that no other parts of the replica or `ic-boundary` have started to depend on it, especially considering the sandboxing-related stuff.

rs/orchestrator/src/upgrade.rs
pierugo-dfinity: Thank you https://github.com/dfinity/ic/pull/10499/commits/8f3a7403ddd0239216816a7b025e9fa41029a00b

rs/orchestrator/src/processes.rs
eichhorl: But now wouldn't we still update it even if `stop` above fails, and then `ensure_running` trivially succeeds (meaning the process wasn't actually restarted)?

pierugo-dfinity: True... Maybe something like https://github.com/dfinity/ic/pull/10499/commits/90016fd2a79c9d855f7ff47eefc9f2cbe53c4437? A cleaner alternative code-wise would be to return early if we fail to stop (or if we fail to start), but that would change the functionality.

eichhorl: Makes sense! 👍

pierugo-dfinity: > return early if we fail to stop (or if we fail to start)

So I gave it a go here: https://github.com/dfinity/ic/pull/10499/commits/fe8ffaac9fca11776c708fd782510dbc2df8a095. I also resolved the TODO that I added in this PR to return the errors and let the caller log them. I think it reads better, PTAL 😅

eichhorl: LGTM

rs/orchestrator/src/processes.rs
pierugo-dfinity: Went with your suggestion here https://github.com/dfinity/ic/pull/10499/commits/748ee21b9cd04a9c24580eacee9cc95ab31a4706

rs/orchestrator/src/process_manager.rs
pierugo-dfinity: @blind-oracle Maybe you can confirm that launching `ic-gateway` as a progress group leader does not raise any issues?

blind-oracle (approved):

basvandijk (approved):

PR #10500, opened 2026-06-17
feat(firewall): prepare cloud engines' firewalls for running `ic-gateway` as a side-car to the replica

pierugo-dfinity:
To make cloud engines self-contained, this PR prepares each of their nodes' firewalls to support running `ic-gateway` next to the replica, with the former routing external traffic to the latter.

Cloud engines now have their own firewall config in `ic.json5`. The only differences with a normal replica is that it additionally opens port 80 (for `ic-gateway`, can be later changed to 443 when supporting TLS). Port 9314 (for `ic-gateway`'s metrics) can be open through a proposal, which also justifies the newly-added `CloudEngines` firewall scope.
You can visualize the introduced changes by looking at `nftables_assigned_cloud_engine.conf.golden`. These apply only to **assigned** cloud engine nodes, i.e. based on their subnet type and not their node reward type (see how `nftables_unassigned_cloud_engine.conf.golden` is unchanged). Indeed, `ic-gateway` will run only when nodes are assigned so it makes sense to open the firewall only then.

Naturally, we also include port 9314 to the HTTP adapter blacklist. While we're at it, we also include ports 9324 (`ic-boundary`'s metrics) and port 19522 (disk encryption's key exchange server during upgrades) to the list, which were likely forgotten by mistake.

The motivation for using a separate template than the replica's is that it will diverge even more in the future when adding `ic-boundary` and will include parts of the `boundary_node_firewall`. The cloud engine's firewall will be mid-way between the replica's and the API BN's so they might as well have their own template.


rs/ic_os/config/tool/templates/ic.json5.template
```diff
@@ -327,7 +327,157 @@ table ip6 filter {\n\
...
+    ct state new add @rate_limit { ip6 saddr limit rate over 1000/minute burst 500 packets } counter name rate_limit_v6_counter drop\n\
+    #  Notes about the rule below:\n\
+    #   - The rule allows a maximum of <<MAX_SIMULTANEOUS_CONNECTIONS_PER_IP_ADDRESS>> persistent connections to any ip6 address.\n\
+    #   - The rule drops all new connections that goes over the configured limit.\n\
+    ct state new add @connection_limit { ip6 saddr ct count over <<MAX_SIMULTANEOUS_CONNECTIONS_PER_IP_ADDRESS>> } counter name connection_limit_v6_counter drop\n\
+    icmpv6 type $icmp_v6_types_accept accept\n\
+    # DHCPv6\n\
+    udp dport { 546 } accept\n\
+    # TCP ports required for GuestOS functionality\n\
+    ip6 saddr { {{ ipv6_prefix }} } ct state { new } tcp dport { 7070, 9090, 9091, 9100, 9314, 19531, 19100, 19522 } accept\n\
```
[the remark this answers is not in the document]
pierugo-dfinity: Similarly to the other metrics ports (9090, 9091, 9324), we open these to the local IPv6 prefix.

rs/ic_os/config/tool/templates/ic.json5.template
```diff
@@ -327,7 +327,157 @@ table ip6 filter {\n\
...
+            "2602:fb2b:110::/48",
+            "2600:c00:2:100::/64",
+            "2001:4c08:2003:b09::/64",
+            "2600:3007:4401::/48",
+            "2a00:fb01:400::/56",
+            "2a00:fb01:400:200::/64",
+            "2a05:d01c:e2c:a700::/56",
+            "2a05:d01c:d9:2b00::/56",
+          ],
+          ports: [22, 2497, 4100, 7070, 8080, 9090, 9091, 9100, 9314, 19100, 19523, 19531],
```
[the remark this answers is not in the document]
pierugo-dfinity: Similarly to the other metrics ports (9090, 9091, 9324), we open these when no registry rules are present (but only to the given prefixes above).

rs/orchestrator/src/firewall.rs
```diff
@@ -549,14 +571,19 @@ impl Firewall {
             // Whitelisting for node IPs
             // In addition to any explicit firewall rules we might apply, we also ALWAYS whitelist
             // all nodes in the registry on the ports used by the protocol
-            Role::AssignedReplica(_) | Role::UnassignedReplica => {
+            Role::AssignedReplica(_) | Role::AssignedCloudEngine(_) | Role::Unassigned => {
                 let (more_tcp_rules, more_udp_rules) =
                     self.get_node_whitelisting_rules(registry_version);
```
eichhorl: Should this function take the role as an argument? Otherwise we will always use the `replica_config`. Maybe we should also add a test where the `cloud_engine_config` is different than the `replica_config`?

daniel-wong-dfinity-org-twin (approved):
For Governance-owned code, this seems more or less harmless (and I assume, adds a new feature). I see some test code, but am not sure about coverage, but not being familiar with firewall (since it is not an "architectural" level aspect of Registry, but rather a "subdomain" issue), I assume it's sufficient -> approve.

rs/orchestrator/src/firewall.rs
pierugo-dfinity: Wow, indeed, thanks for the careful review 🙏 
I added a test for that as suggested. Introduced a named struct containaing all the different whitelists/blacklists that we can pass directly instead of checking the role again inside the body of the function (which is now agnostic of whether we are an assigned replica or cloud engine).

eichhorl (approved):

rs/ic_os/config/tool/templates/ic.json5.template
r-birkner: Actually, here we should also add the prefix of `tp1`: `2600:3007:4401::/48`

rs/ic_os/config/tool/templates/ic.json5.template
r-birkner, line 430: Did you omit `9324` on purpose? This rule allows the node provider to scrape metrics/logs from the same prefix as the one within that node runs.

rs/ic_os/config/tool/templates/ic.json5.template
```diff
@@ -327,7 +327,157 @@ table ip6 filter {\n\
...
+  }\n\
+\n\
+  chain FORWARD {\n\
+    type filter hook forward priority 0; policy drop;\n\
+  }\n\
+\n\
+  chain OUTPUT {\n\
+    type filter hook output priority 0; policy accept;\n\
+    meta skuid ic-http-adapter fib daddr type local ct state { new } tcp dport { 1-19999 } reject # Block restricted local addresses ic-http-adapter HTTPS access\n\
+    meta skuid ic-http-adapter ip6 daddr { 2a00:fb01:400:42::/64, 2602:fb2b:110::/48, 2602:fb2b:100::/48, 2602:fb2b:120::/48 } ct state { new } tcp dport { 1-19999 } reject # Block restricted outbound ic-http-adapter HTTPS access\n\
```
r-birkner: Here as well, let's add the `tp1` prefix: `2600:3007:4401::/48`

r-birkner (approved):

rs/ic_os/config/tool/templates/ic.json5.template
pierugo-dfinity: https://github.com/dfinity/ic/pull/10500/commits/4570ae679d2d458fa3605535fc8a50e8601cf67b

rs/ic_os/config/tool/templates/ic.json5.template
pierugo-dfinity: I preferred to only touch on `ic-gateway` for now (whose metrics port is 9314, which is included in the list). When adding `ic-boundary`, I can update the firewall again, in particular by opening port 9324 here. What do you think?

rs/ic_os/config/tool/templates/ic.json5.template
pierugo-dfinity: https://github.com/dfinity/ic/pull/10500/commits/4570ae679d2d458fa3605535fc8a50e8601cf67b

rs/ic_os/config/tool/templates/ic.json5.template
r-birkner: Yes makes sense!

r-birkner (approved):

PR #10501, opened 2026-06-18
feat(https-outcalls): Introduce pay-as-you-go base fee

eichhorl:
## Background
Currently, HTTPS outcalls are charged upfront based on a `max_response_bytes` parameter. This is done by subtracting the full cost from the caller's payment. The remaining cycles are stored in the request context and refunded once a response is delivered.

Instead, we want to introduce pay-as-you-go pricing, which charges cycles whenever resources are consumed. This happens in three stages:
1. Base cost. This fee is charged for every request upfront
2. Per-replica cost. The remaining cycles after charging the base cost are split evenly between the participating replicas. Each replica consumes some of their allowance as the HTTP request is processed. In the end, the remaining cycles are gossiped as part of refund shares.
3. Consensus cost. This fee is charged for including the aggregated HTTP response as part of a block. The cost is covered by the sum of refund shares that were included in the aggregated response. Cycles remaining after charging the consensus cost are refunded to the user asynchronously.

## Proposed Changes
This PR introduces a function to calculate the base cost (1.) based on the request context and the subnet's cycles cost config. The pricing formula is defined [here](https://colab.research.google.com/drive/1MyZBivzU_5kQtqQ8Z71gYbrMubGrTXmv#scrollTo=vxa3d-NYyEOi) (internal). Unlike legacy pricing, for pay-as-you-go pricing, the user's _entire_ payment is taken out of the request. This means nothing will be refunded when a response is returned. The asynchronous refunds will instead take a different path. After charging the base cost, the payment is used to populate the context's `RefundStatus` which determines the per-replica allowance as described above.

The exception is if the subnet's cycle cost schedule is `Free`, in which case the caller's payment in the request remains untouched, and no refundable cycles are generated. All of the attached cycles are returned back to the caller when a response is delivered.

Additionally, we introduce a (disabled) feature flag for flexible outcalls, in order to test the changes end-to-end.

The legacy charging flow should be unchanged by this PR.


rs/cycles_account_manager/src/cycles_account_manager.rs
```diff
@@ -1236,6 +1236,55 @@ impl CyclesAccountManager {
         CompoundCycles::new(amount, subnet_cycles_config.cost_schedule)
     }
 
+    pub fn http_request_base_fee(
+        &self,
+        request_size: NumBytes,
+        replication: &Replication,
+        subnet_cycles_config: CyclesAccountManagerSubnetConfig,
+    ) -> CompoundCycles<HTTPOutcalls> {
+        const BASE_FEE: u128 = 1_000_000;
```
schneiderstefan: It would be better to have these constants at a central place. Like `rs/config/src/execution_environment`, which has lots of contants. Furthermore, you even already touch that file in this PR.

schneiderstefan (approved):

rs/cycles_account_manager/src/cycles_account_manager.rs
eichhorl: I moved them to `rs/config/src/subnet_config.rs` instead, which seems to have some fee constants. Now they will be visible to all config users

PR #10505, opened 2026-06-18
feat: CON-1467 CON-1705 Set registry version of `CanisterHttpContext`s

eichhorl:
https://github.com/dfinity/ic/pull/10484 introduced a `RegistryVersion` field to `CanisterHttpContext`s. For compatibility reasons, the field was introduced to the proto definition but always set to 0 in all production cases.

With this PR, we start setting the field to the registry version at which new `CanisterHttpContext`s are created. In the future this field will be used as the single source of truth for the registry version throughout the lifetime of an HTTP outcall.

For compatibility reasons, a subnet running the version of this PR should not be directly downgraded to a version earlier than https://github.com/dfinity/ic/pull/10484 (corresponding replica version proposed [here](https://dashboard.internetcomputer.org/proposal/142444)).


rs/types/types/src/canister_http.rs
```diff
@@ -1937,6 +1861,38 @@ mod tests {
             replication,
         }
     }
+
+    /// Generates a context from `args` and `node_ids`, filling in dummy values
+    /// for the time, request, registry version, and rng.
+    fn generate_args_context(
```
mraszyk:
```suggestion
    fn generate_context(
```

mraszyk (approved):

PR #10510, opened 2026-06-18
perf: [DSM-144] Misc certification optimizations

alin-at-dfinity:
* Use precomputed empty leaf hash instead of computing it from scratch every time.
* Pass the key to `MapTransformFork::mk_tree` by reference, don't clone it (especially since it's only used in a couple of cases).
* Simplify `CanisterFork`.


alin-at-dfinity:
This speeds up sequential `HashTree` computation by about 10%. Unfortunately, the effect on concurrent hashing is a lot smaller (1%-2%), because the bottleneck is memory accesses, not computation.

rs/canonical_state/src/lazy_tree_conversion.rs
```diff
@@ -771,7 +771,6 @@ fn status_to_tree(status: &IngressStatus) -> LazyTree<'_> {
 }
 
 const CERTIFIED_DATA_LABEL: &[u8] = b"certified_data";
-const CONTROLLER_LABEL: &[u8] = b"controller";
```
schneiderstefan, line 774: Is this gone in all certification versions?

alin-at-dfinity: Yes, only the labels in `CANISTER_LABELS` and `CANISTER_NO_MODULE_LABELS` are (and were, before this change) populated.

This was left behind from some old certification version, likely because the `match` statements below still referenced it, so the compiler did not report it as unused.

schneiderstefan (approved):

daniel-wong-dfinity-org-twin:
afaict, the files here are not owned by Governance. Please, let me know in Slack if review and/or approval from Governance team is required. If not, please, feel free to close the automated reminder with a reply like, "No Governance-owned code in this PR.", and maybe explain why Governance team was tagged (by accident). Thanks.

PR #10511, opened 2026-06-18
fix: deflake //rs/nervous_system/integration_tests:integration_tests_test_tests/sns_lifecycle

basvandijk:
## Summary

Deflakes `//rs/nervous_system/integration_tests:integration_tests_test_tests/sns_lifecycle` by moving it into a dedicated `rust_test` target that limits how many SNS lifecycle scenarios run concurrently against the shared PocketIC server.

## Root cause

The `sns_lifecycle` binary contains 10 `#[tokio::test]` scenarios, each of which spins up a full NNS + SNS. `libtest` runs them concurrently (one thread per available CPU), and they all share a single PocketIC server process (every instance talks to the same `127.0.0.1:<port>`).

In the flaky runs (2026-06-16) **every one of the 10 PocketIC instances failed simultaneously** with:

```
HTTP failure: reqwest::Error { ... /update/await_ingress_message,
  source: ... Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" } }
```

(and `hyper::Error(IncompleteMessage)`), panicking at `packages/pocket-ic/src/nonblocking.rs:1733` — `builder.send().await.expect("HTTP failure")`, which does not retry on transient connection errors.

All instances dying at once points to the shared PocketIC **server process** itself going down under memory pressure (running ~10 full NNS+SNS replicas at the same time). It is transient — the test passes on the next attempt — which is consistent with memory pressure rather than a logic bug.

## Fix

Move `sns_lifecycle` out of the shared `integration_tests_test` suite into its own `rust_test` target (the pattern already documented in this BUILD file for the other flaky tests) and cap concurrency:

- `RUST_TEST_THREADS = "4"` — at most 4 live PocketIC instances at a time, which bounds the shared server's memory usage (the established pattern for PocketIC tests, e.g. `packages/pocket-ic`).
- `cpu:8` — matches the sibling `rs/sns/integration_tests` SNS tests and gives more scheduling headroom.
- `RUST_TEST_NOCAPTURE = "1"` — better logging, consistent with the other dedicated targets in this file.

The test label changes to `//rs/nervous_system/integration_tests:sns_lifecycle` (same convention as the other dedicated targets in this package). No CI/Bazel config referenced the old label.

## Verification

```
bazel test --test_output=errors --runs_per_test=3 --jobs=3 //rs/nervous_system/integration_tests:sns_lifecycle
```

→ `Stats over 3 runs: max = 301.7s, min = 295.8s, avg = 298.8s` — 3/3 passed. Runtime settles at ~300s, well under the `long` (900s) timeout (it was ~125–231s at full 10-way parallelism).

---

### DIFF
--- a/rs/nervous_system/integration_tests/BUILD.bazel
+++ b/rs/nervous_system/integration_tests/BUILD.bazel
@@ -196,6 +196,7 @@ rust_test_suite_with_extra_srcs(
             "tests/sns_topics.rs",
             "tests/sns_extension_test.rs",
             "tests/custom_proposal_criticality.rs",
+            "tests/sns_lifecycle.rs",
         ],
     ),
     data = DEV_DATA,
@@ -207,6 +208,30 @@ rust_test_suite_with_extra_srcs(
     ],
     tags = [
         "cpu:6",
+    ],
+    deps = [":nervous_system_integration_tests"] + DEPENDENCIES_WITH_TEST_FEATURES,
+)
+
+rust_test(
+    name = "sns_lifecycle",
+    timeout = "long",
+    srcs = ["tests/sns_lifecycle.rs"],
+    data = DEV_DATA,
+    env = DEV_ENV | {
+        "RUST_TEST_NOCAPTURE": "1",
+        # Each SNS lifecycle scenario spins up a full NNS + SNS on the shared
+        # PocketIC server. Running all scenarios concurrently exhausts the
+        # server's memory, causing it to drop connections ("Connection reset by
+        # peer") and flake the test. Cap the number of concurrently running
+        # scenarios to keep the server's memory usage bounded.
+        "RUST_TEST_THREADS": "4",
+    },
+    proc_macro_deps = [
+        # Keep sorted.
+        "@crate_index//:rust_decimal_macros",
+    ],
+    tags = [
+        "cpu:8",
     ],
     deps = [":nervous_system_integration_tests"] + DEPENDENCIES_WITH_TEST_FEATURES,
 )


daniel-wong-dfinity-org-twin (approved):
This change seems to cause more or less no harm, and since Bas says this reduces flakiness, it strongly seems to be a net win -> approve

OTOH, I do not super understand how this change works. The comment within the new rule is nice. The story there sounds pretty plausible. But as I said, I am not very familiar with the test. IIUC, the critical change here is the introduction of RUST_TEST_THREADS=4, which limits the amount of memory that the test needs, and because exhaustion is prevented, the test becomes more reliable (so the theory goes).

One thing I do not understand in the PR description is where it says that there is a single PocketIc server. That surprises me. How can many tests share the same server? Wouldn't that make it very difficult to maintain isolation between tests?? Well, I guess that is kind of what's happening, except on a resource level (specifically, memory), not on in terms of functionality/behavior. In any case, this mystery is tangential to the main question (i.e. "How/why does this deflake?"), and therefore, does not block approval.

# Minor Feedback

These BUILD files are becoming pretty repetitive. OTOH, trying to factor out the repetitive bits would most likely result in the creation of some esoteric function (which is something we should be particularly sensitive to in BUILD files). Not sure if the power to weight ratio of that is sufficiently high at this point to justify it yet, but we are slowly inching towards that point...

3 runs seems small. OTOH, this test takes a "long" time to run (~5 minutes). Still, with the power of --runs_per_test, it really shouldn't be that troublesome to run more times (e.g. 10) so that we have a better idea of how reliable this test has become as a result of this change. Maybe 100 runs is excessive, but 3 seems kind of stingy.

PR #10519, opened 2026-06-19
feat: CON-1639 HTTPS outcalls pay-as-you-go and dark launch budget trackers

eichhorl:
## Background
Currently, HTTPS outcalls are charged upfront based on a `max_response_bytes` parameter. This is done by subtracting the full cost from the caller's payment. The remaining cycles are stored in the request context and refunded once a response is delivered.

Instead, we want to introduce pay-as-you-go pricing, which charges cycles whenever resources are consumed. This happens in three stages:
1. Base cost. This fee is charged for every request upfront
2. Per-replica cost. The remaining cycles after charging the base cost are split evenly between the participating replicas. Each replica consumes some of their allowance as the HTTP request is processed. In the end, the remaining cycles are gossiped as part of refund shares.
3. Consensus cost. This fee is charged for including the aggregated HTTP response as part of a block. The cost is covered by the sum of refund shares that were included in the aggregated response. Cycles remaining after charging the consensus cost are refunded to the user asynchronously.

The per-replica cost (2.) is calculated by the "budget tracker": This struct is instantiated with the per-replica cycles allowance, whenever a new request is starting to be processed by the HTTP adapter. As the response is downloaded, transformed and gossiped, the tracker charges the consumed cycles from the initial allowance. If at any point the remaining allowance _does not_ cover an outstanding charge, an error is returned and a reject response is gossiped. If the initial allowance _does_ cover all charges, the remaining cycles are refunded. To this end, the budget tracker creates a payment receipt which is gossiped alongside the response.

Before this PR only one budget tracker exited (`LegacyTracker`) which doesn't compute any per-replica cost, and also doesn't refund anything.

## Proposed Changes
This PR introduces the "pay-as-you-go" budget tracker, whose purpose is to calculate the per-replica cost for outcalls using the "pay-as-you-go" pricing (note that such outcalls do not exist yet). This is done by implementing the per-replica part of the pricing formula defined [here](https://colab.research.google.com/drive/1MyZBivzU_5kQtqQ8Z71gYbrMubGrTXmv#scrollTo=vxa3d-NYyEOi) (internal). To do this, we additionally pass the subnet's size and cycle cost schedule to the HTTP adapter. This is needed to calculate the correct pricing. 

"Pay-as-you-go" pricing charges for the amount of bytes that are gossiped explicitly in the case of flexible and non-replicated outcalls (where the whole response is gossiped). To do this, we move the truncation of oversized reject messages into the adapter, such that the correct payload length is charged.

Additionally, we implement and start to use a `DarkLaunchTracker`. This tracker calculates both, the real (legacy, i.e. 0) and the new (pay-as-you-go) per-replica cost. In the end, only the "real" refund is gossiped. However, this allows us to compare both trackers, and observe whenever the pay-as-you-go tracker returns an out-of-cycles error, while the legacy tracker succeeds. Such an event indicates that the outcall would not be covered by enough cycles under the new pricing. In this case, the canister ID is logged and a metric increased.

The legacy charging flow should be unchanged by this PR.


rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -292,6 +315,20 @@ impl CanisterHttpPoolManagerImpl {
...
+                    warn!(
+                        every_n_seconds => 10,
+                        self.log,
+                        "Skipping canister http request {} because the subnet size or cost \
+                         schedule could not be determined at registry version {}",
+                        id,
+                        context.registry_version
+                    );
+                    continue;
+                };
```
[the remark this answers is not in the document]
eichhorl: That's true, but the registry lookup failing is unexpected, and would only lead to a timeout of the request in the worst case

rs/https_outcalls/client/src/client.rs
```diff
@@ -262,6 +271,38 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
...
+                    .message
+                    .ellipsize(MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES, 90);
+            }
+
+            // Account for the cost of gossiping the final (post-transform)
+            // response to peers before creating the receipt.
+            let response_size = match &payload {
+                Ok(response) => response.len(),
+                Err(reject) => reject.message.len(),
+            };
```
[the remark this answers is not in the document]
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/commits/d8ec86a875782ede62b54ab0eb06ce42ce9a4c93

rs/https_outcalls/client/src/client.rs
```diff
@@ -262,6 +271,38 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
...
+            // Account for the cost of gossiping the final (post-transform)
+            // response to peers before creating the receipt.
+            let response_size = match &payload {
+                Ok(response) => response.len(),
+                Err(reject) => reject.count_bytes(),
+            };
+            let payload = budget
+                .subtract_gossip_usage(NumBytes::from(response_size as u64))
+                .map_err(|PricingError::InsufficientCycles| CanisterHttpReject {
+                    reject_code: RejectCode::SysFatal,
```
mraszyk: is this consistent?

rs/https_outcalls/client/src/client.rs
```diff
@@ -1128,6 +1172,71 @@ mod tests {
...
+            status: 200,
+            headers: Vec::new(),
+            content: Vec::new(),
+        };
+        let mock_grpc_channel = setup_adapter_mock(Ok(create_result_from_response(response))).await;
+        let (svc, mut handle) = setup_system_query_mock();
+
+        // A multi-byte message whose 1200 bytes exceed the 1024-byte limit, with
+        // emoji straddling the truncation boundary to exercise char safety.
+        let oversized_message = "😀".repeat(300);
```
mraszyk: I don't think this crosses the boundary of 1024 chars.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -247,6 +247,29 @@ impl CanisterHttpPoolManagerImpl {
...
+    ) -> Option<(NumberOfNodes, CanisterCyclesCostSchedule)> {
+        let record = self
+            .registry_client
+            .get_subnet_record(self.replica_config.subnet_id, registry_version)
+            .ok()
+            .flatten()?;
+        let subnet_size = u32::try_from(record.membership.len()).ok()?;
+        if subnet_size == 0 {
+            return None;
+        }
```
mraszyk: why this special case? could you please add a comment?

rs/types/types/src/canister_http.rs
```diff
@@ -90,6 +90,9 @@ pub const MAX_CANISTER_HTTP_REQUEST_BYTES: u64 = 2_000_000;
 /// Maximum number of response bytes for a canister http request.
 pub const MAX_CANISTER_HTTP_RESPONSE_BYTES: u64 = 2_000_000;
 
+/// Maximum size of a canister http reject message.
+pub const MAXIMUM_ALLOWED_ERROR_MESSAGE_BYTES: usize = 1024; // 1KB
```
mraszyk: nit: MAXIMUM_CANISTER_HTTP_ERROR_MESSAGE_BYTES sounds more consistent with the neighbouring constants

rs/https_outcalls/pricing/src/metrics.rs
```diff
@@ -0,0 +1,42 @@
+use ic_metrics::MetricsRegistry;
+use prometheus::IntCounterVec;
+
+/// Label identifying the accounting step at which the shadow tracker diverged
```
mraszyk: "diverged" (or "disagreed" below) sounds like they don't agree on the actual charged amount which is expected; I'd rephrase this so that it is clear that we refer to the step at which the shadow tracker reported an error while the legacy tracker succeeded

rs/https_outcalls/pricing/src/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// share of requests that would NOT be backwards compatible.
+    pub shadow_incompatible_total: IntCounterVec,
+}
+
+impl PricingMetrics {
+    pub fn new(metrics_registry: &MetricsRegistry) -> Self {
+        Self {
+            shadow_requests_total: metrics_registry.int_counter_vec(
+                "canister_http_pricing_shadow_requests_total",
+                "Total canister http requests evaluated by the dark-launch budget tracker, by \
```
mraszyk: could you please point out why this count is different from the total number of all canister http requests (likely tracked separately already)?

rs/https_outcalls/pricing/src/lib.rs
```diff
@@ -21,11 +33,20 @@ pub trait BudgetTracker: Send {
...
     /// Returns the maximum instructions allowed for the transformation function.
     fn get_transform_limit(&self) -> NumInstructions;
-    /// Deducts the actual instructions consumed by the transformation.
+    /// Deducts the cost of the instructions consumed by the transformation.
     ///
     /// # Invariants
     ///  - This method returns `Ok(())` if and only if `usage <= get_transform_limit()`.
     fn subtract_transform_usage(&mut self, usage: NumInstructions) -> Result<(), PricingError>;
+    /// Deducts the cost of the final (post-transform) response that this replica
+    /// produced and that will be gossiped to peers.
```
mraszyk:
```suggestion
    /// produced and that will be gossiped to peers. This cost does not apply to fully-replicated requests.
```

rs/https_outcalls/pricing/src/dark_launch.rs
```diff
@@ -0,0 +1,332 @@
...
+pub struct DarkLaunchTracker {
+    real: Box<dyn BudgetTracker>,
+    shadow: Box<dyn BudgetTracker>,
+    canister_id: CanisterId,
+    replication: &'static str,
+    metrics: PricingMetrics,
+    log: ReplicaLogger,
+    /// Whether an incompatibility has already been recorded for this request.
+    /// Ensures we count and log at most once per request.
+    reported: bool,
```
mraszyk:
```suggestion
    error_reported: bool,
```

rs/https_outcalls/pricing/src/dark_launch.rs
```diff
@@ -0,0 +1,332 @@
...
+            self.replication,
+            real,
+            shadow,
+        );
+    }
+}
+
+impl BudgetTracker for DarkLaunchTracker {
+    fn get_adapter_limits(&self) -> AdapterLimits {
+        self.real.get_adapter_limits()
```
mraszyk, line 84: could add a debug_assert that the shadow limits are equal, also for other getters below

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+
+impl PayAsYouGoTracker {
+    pub fn new(
+        context: &CanisterHttpRequestContext,
+        subnet_size: NumberOfNodes,
+        cost_schedule: CanisterCyclesCostSchedule,
+    ) -> Self {
+        Self {
+            subnet_size,
+            is_flexible_pricing: match context.replication {
```
mraszyk: "flexible" is confusing since non-replicated also makes this true, how about `is_ghossipping`?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+                .max_response_bytes
+                .unwrap_or(NumBytes::from(MAX_CANISTER_HTTP_RESPONSE_BYTES)),
+            spent: 0,
+        }
+    }
+
+    /// Charges `amount` against the budget. Returns an error if the total spent
+    /// now exceeds the available allowance.
+    fn charge(&mut self, amount: u128) -> Result<(), PricingError> {
+        // A free cost schedule means the subnet charges nothing for resources.
```
mraszyk, line 83: I think we should still track the cost so that canister metrics could be updated based on the actual work done. Canister metrics are used on "free" subnets for cost accounting in user space.

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+
+impl BudgetTracker for PayAsYouGoTracker {
+    fn get_adapter_limits(&self) -> AdapterLimits {
+        AdapterLimits {
+            max_response_size: self.max_response_size,
+            // Mirror the legacy limit: the server enforces a 30s timeout, so 60s
+            // here is just a safety margin.
+            max_response_time: Duration::from_secs(60),
```
mraszyk: Could this use a constant instead of hard-coded value?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+        let cost = (usage.get() as u128) / TRANSFORM_INSTRUCTION_DIVISOR;
+        self.charge(cost)
+    }
+
+    fn subtract_gossip_usage(
+        &mut self,
+        transformed_response_size: NumBytes,
+    ) -> Result<(), PricingError> {
+        // For fully replicated outcalls the gossip term is a
+        // consensus cost (ignored here for now). For flexible outcalls each
```
mraszyk: Why "for now"? Is there a plan to change this later?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        let transformed_size = 500_u64;
+        assert_eq!(
+            tracker.subtract_gossip_usage(NumBytes::from(transformed_size)),
+            Ok(())
+        );
+        let expected =
+            FLEXIBLE_PER_TRANSFORMED_BYTE_NODE_FEE * transformed_size as u128 * subnet_size as u128;
+        assert_eq!(tracker.spent, expected);
+    }
```
mraszyk, line 292: This and the previous test are almost identical - could you please factor out a common helper?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,362 @@
...
+        );
+        assert_eq!(tracker.create_payment_receipt().refund, Cycles::zero());
+    }
+
+    #[test]
+    fn charges_nothing_on_free_cost_schedule() {
+        // On a free subnet the tracker charges nothing, even for usage that
+        // would otherwise exceed the allowance, so the full allowance is
+        // refunded. A flexible request is used so the gossip term (which would
+        // normally be charged) is also exercised.
```
mraszyk:
```suggestion
        // not be charged for fully-replicated requests) is also exercised.
```

rs/https_outcalls/client/src/client.rs
mraszyk: the reject code is inconsistent with
```
rs/execution_environment/src/execution_environment.rs, lines 2174-2184, inside try_add_http_context_to_replicated_state:

  if request.payment < charged_fee.real() {
      return Err(UserError::new(
          ErrorCode::CanisterRejectedMessage,
          format!(
              "{} request sent with {} cycles, but {} cycles are required.",
              request.method_name,
              request.payment,
              charged_fee.real()
          ),
      ));
  }
```

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Thanks: https://github.com/dfinity/ic/pull/10519/commits/608b767688ee8aa9cee78d736e6d8fc3fcfe184e

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/commits/ebd2a9be2c362cece767caa33960f390d8cfcc82

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Removed, thanks: https://github.com/dfinity/ic/pull/10519/changes/22c5465fbfd07a6e7583ff72a822d8a7835e81b0

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Switched to a constant: https://github.com/dfinity/ic/pull/10519/changes/6d32da54b9f9cb6f3a2e570f96b8dc6c0872520c

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Good idea: https://github.com/dfinity/ic/pull/10519/commits/47738078bbc0571278bffc08f718e76d6c8585b1

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: I haven't gotten to that part yet, but following the original design they might not be equal anymore at some point, because the pay-as-you-go limits could be made more dynamic. The idea is that, as the cycles budget is consumed, we reduce the limits for the next step.

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/commits/591ececda428cedf7fd5b8873f44e9f435471bba

rs/https_outcalls/pricing/src/lib.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/commits/a91a45b3ef23cb841c282cc1be204b2b4ccb6071

rs/https_outcalls/pricing/src/metrics.rs
eichhorl: Good point: https://github.com/dfinity/ic/pull/10519/changes/b8df347aedcdd5f1b97c78eecd187fb98dd39b31

rs/types/types/src/canister_http.rs
eichhorl: Done, thanks: https://github.com/dfinity/ic/pull/10519/commits/35c202a29401c50303a36a5c05848e05abc8e840

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Removed it instead, since it is covered by registry invariants: https://github.com/dfinity/ic/pull/10519/commits/5132054cabf4fc47708e8e35d656dbf3071af8f3

rs/https_outcalls/client/src/client.rs
eichhorl: Why not? There is an assert just below saying that it is

rs/https_outcalls/client/src/client.rs
eichhorl: Thanks changed to `CanisterReject`: https://github.com/dfinity/ic/pull/10519/commits/5ed23a83569fbe53d11c135bbb348be9de438ea6

rs/https_outcalls/pricing/src/metrics.rs
eichhorl: Good point, we can use `canister_http_requests_total` instead, just had to add the replication label: https://github.com/dfinity/ic/pull/10519/commits/81c0f8194fcb5c89b57a32f6db7a4d9097b54640

rs/https_outcalls/client/src/client.rs
mraszyk: What I mean is that the emoji is encoded using 4 bytes and the byte limit of 1024 bytes is an integer multiple of that so we don't exercise the case of the emoji crossing the byte limit (in this case truncating to exactly 1024 bytes would cut the emoji encoding and the blob won't be a valid string anymore).

rs/https_outcalls/pricing/src/metrics.rs
mraszyk: There are still occurrences of "divergence" and "disagreed" that I'd suggest to clean up, too.

rs/https_outcalls/pricing/src/dark_launch.rs
mraszyk: Could you please add a comment so that future readers don't raise the same question?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,341 @@
...
+//     + 50 * transformed_response_bytes_i * N + transform_instructions_i / 13
+const PER_DOWNLOADED_BYTE_FEE: u128 = 50;
+const PER_RESPONSE_MS_FEE: u128 = 300;
+const TRANSFORM_INSTRUCTION_DIVISOR: u128 = 13;
+const FLEXIBLE_PER_TRANSFORMED_BYTE_NODE_FEE: u128 = 50;
+
+pub struct PayAsYouGoTracker {
+    /// Number of nodes (`N`) on the subnet.
+    subnet_size: NumberOfNodes,
+    /// Whether this responses to this outcalls are gossiped (only flexible and non-replicated).
```
mraszyk:
```suggestion
    /// Whether responses to this outcalls are gossiped (only flexible and non-replicated).
```

rs/https_outcalls/pricing/src/dark_launch.rs
```diff
@@ -0,0 +1,304 @@
...
+        if real.is_ok() == shadow.is_ok() {
+            return;
+        }
+        if self.error_reported {
+            return;
+        }
+        self.error_reported = true;
+        self.metrics
+            .shadow_incompatible_total
+            .with_label_values(&[step, self.replication.as_str()])
```
pierugo-dfinity, line 67: Would it make sense to add a label to distinguish which one (real vs shadow) resulted in insufficient cycles? Such that you could easily query how many requests succeeded with the real but not with the shadow, which is what we are mostly interested in IIUC.

pierugo-dfinity: Concerning reporting once per request, wouldn't we be interested to know if the trackers diverge at different steps?

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -0,0 +1,341 @@
...
+//
+// Fully-replicated per replica:
+//   50 * downloaded_bytes_i + 300 * request_ms_i + transform_instructions_i / 13
+//
+// Non-replicated/Flexible per replica:
+//   50 * downloaded_bytes_i + 300 * request_ms_i
+//     + 50 * transformed_response_bytes_i * N + transform_instructions_i / 13
+const PER_DOWNLOADED_BYTE_FEE: u128 = 50;
+const PER_RESPONSE_MS_FEE: u128 = 300;
+const TRANSFORM_INSTRUCTION_DIVISOR: u128 = 13;
```
pierugo-dfinity, line 34: Just making sure, this doesn't need to be dynamic based on the actual size of the subnet, right?

rs/https_outcalls/pricing/src/metrics.rs
eichhorl: https://github.com/dfinity/ic/pull/10519/commits/8e46e6ba5dd21f56a96dfb81d9b24ee20b7422a4

rs/https_outcalls/pricing/src/payg.rs
eichhorl: https://github.com/dfinity/ic/pull/10519/commits/a89e86b1632be4f2a73346929bc178a34b1af47f

rs/https_outcalls/client/src/client.rs
eichhorl: Ah, note that the 1024 byte limit includes the 3 bytes for `"..."`. Also `.ellipsize(max, percentage)` actually makes two cuts. The `"..."` are placed at `percentage%` of the truncated message.

I changed the test such that now both cuts would happen inside an emoji (not just the first one): https://github.com/dfinity/ic/pull/10519/commits/eaeaa56964f0629b0757dacce654ac6f64b6e2ff

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/changes/f8d302f1a8329a4f38cb4cf8965650a0a86767ed

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: The (real) legacy tracker always returns `Ok`, so the only way an incompatibility can be noticed is if the (shadow) payg tracker fails because it ran out of cycles.

rs/https_outcalls/pricing/src/dark_launch.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Compares the results of the real and shadow trackers for a given
+    /// accounting `step` and increment the shadow_incompatible_total metric
+    /// if they differ.
+    fn compare(
+        &mut self,
+        step: &str,
+        real: &Result<(), PricingError>,
+        shadow: &Result<(), PricingError>,
+    ) {
+        if real.is_ok() == shadow.is_ok() {
```
mraszyk: I guess we are actually interested in cases where real is ok and shadow is error, the opposite case of real being err and shadow ok is not interesting since it does not affect backward compatibility. So we could just adjust this condition instead of distinguishing the kind of incompatibility (as suggested in the comment below).

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Correct, the instruction cost depends on the reference subnet size, which is taken out here (by dividing), so that we can determine the correct cost per node.

However, IIUC, since very recently this reference subnet size may actually be dynamic (in case of SEV subnets), so should probably also be determined per request at runtime. I added a TODO here: https://github.com/dfinity/ic/pull/10519/commits/63ee45a609074c75770f578e1ccdb035663e1169

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/10519/changes/a7b8261c7b3a8ce3a3b2a850d69d30b4e9ac7fb1

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Supporting this would require some bigger changes. The original design did not consider free subnets at all. The problem is that the per-replica allowance is derived from the cycles attached to the call (which on free subnets may be zero). For that reason, currently free subnets would only report the base cost under the new pricing.

Fabricating an allowance out of thin air instead risks minting cycles through the refund shares, but making up a fake (+ appropriate) allowance would probably be required to support this if necessary.

mraszyk: I see, let's then not do it in this PR, but could you please create a ticket or so to track this?

eichhorl: https://dfinity.atlassian.net/browse/CON-1724

mraszyk: Thank you!

rs/https_outcalls/pricing/src/dark_launch.rs
pierugo-dfinity: I see you changed the condition at the top of the function. Works for me!

rs/https_outcalls/pricing/src/payg.rs
pierugo-dfinity: Thanks 👍

pierugo-dfinity (approved):

PR #10521, opened 2026-06-19
fix: CONTAINER_RUNTIME=docker ./ci/container/container-run.sh

basvandijk:
This is to fix:
```
$ CONTAINER_RUNTIME=docker ./ci/container/container-run.sh
...
jq: error (at <stdin>:1): Cannot index string with string "Type"
jq: error (at <stdin>:2): Cannot index string with string "Type"
jq: error (at <stdin>:3): Cannot index string with string "Type"
jq: error (at <stdin>:4): Cannot index string with string "Type"
```


nmattia (approved):
Thanks!

PR #10532, opened 2026-06-22
feat: [MR-144] Reuse unchanged canister digests across `HashTrees`

alin-at-dfinity:
### Summary

`hash_lazy_tree` re-hashes every canister's subtree each round, even though almost all canisters are untouched. This PR lets a new `HashTree` reuse subtree digests from a previous one, keyed by the canister's backing `Arc<CanisterState>` (shared copy-on-write, replaced only on mutation).

### What changed

- **Stub nodes.** A new `NodeKind::Stub` collapses a reusable subtree to a digest-only node. A `LazyFork` opts in via `subtree_source()`; `CanisterFork` returns a `SubtreeSource` over its `Arc<CanisterState>`. Other forks are still materialized inline.
- **`SubtreeSource`.** A type-erased, owned handle (`Arc<dyn Any>` + expander fn pointer) that keeps the source alive (no ABA on the address) and re-expands the stub for witnesses. `Eq` is a conservative, false-negative-only reuse gate: same source allocation **and** same expander.
- **Version in the expander.** `expand_canister::<const V: u32>` bakes the certification version into the function pointer, so it isn't stored per stub and reuse can't mix versions.
- **`hash_lazy_tree_with_baseline`.** Lockstep-joins the new tree against a baseline; matching `SubtreeSource`s reuse the baseline digest, otherwise rebuild. Witnesses expand stubs via `SubtreeSource::expand()`, so the tree stays self-contained.
- **Adaptive parallelization.** Forks build sequentially, then hand the *remaining* children to the thread pool once the observed rebuild rate (after a warmup) projects more than `PARALLEL_MIN_CHILDREN` rebuilds. The common reuse-heavy path stays sequential; large/from-scratch forks parallelize.

### Tests

New `tests/subtree.rs`: stubbing, witness/absence-proof expansion from source, baseline builds (full and partial change) matching from-scratch, and `reuse_is_by_identity_not_by_value` (shared `Arc`s reuse, replaced `Arc`s rebuild).

### Follow-up

Actually invoke `hash_lazy_tree_with_baseline` from production code, with the latest available `HashTree`. For now, it is only invoked from the benchmark and tests.

Stubbing is implemented for `/canisters` only; ingress `/messages` are left for later.


rs/canonical_state/tree_hash/src/lazy_tree.rs
```diff
@@ -116,3 +134,94 @@ pub fn follow_path<'a>(t: &LazyTree<'a>, path: &[&[u8]]) -> Option<LazyTree<'a>>
...
+}
+
+/// Rebuilds a stubbed subtree's [`HashTree`](crate::hash_tree::HashTree) from
+/// its type-erased [`SubtreeSource`], by [downcasting](SubtreeSource::downcast)
+/// the held `Arc` back to its concrete source and re-materializing it. Used to
+/// expand a [`NodeKind::Stub`](crate::hash_tree::HashTree) on demand during
+/// witness generation.
+///
+/// It is a plain (non-capturing) function pointer, so the producer of the stub
+/// must bake the certification version into it. The pointer alone fully
```
mraszyk: I didn't get the sentence "It is a plain (non-capturing) function pointer, so the producer of the stub must bake the certification version into it."

rs/canonical_state/tree_hash/src/lazy_tree.rs
```diff
@@ -116,3 +134,94 @@ pub fn follow_path<'a>(t: &LazyTree<'a>, path: &[&[u8]]) -> Option<LazyTree<'a>>
...
+
+impl SubtreeSource {
+    /// Creates a handle that shares ownership of the subtree's `source` and can
+    /// rebuild the subtree from it, via the `expander`.
+    pub fn new<T: Any + Send + Sync>(source: &Arc<T>, expander: SubtreeExpander) -> Self {
+        let this = Self {
+            source: Arc::clone(source) as Arc<dyn Any + Send + Sync>,
+            expander,
+        };
+        debug_assert!((expander)(&this).is_ok());
```
mraszyk:
```suggestion
        debug_assert!(this.expand().is_ok());
```

rs/canonical_state/tree_hash/src/lazy_tree.rs
```diff
@@ -116,3 +134,94 @@ pub fn follow_path<'a>(t: &LazyTree<'a>, path: &[&[u8]]) -> Option<LazyTree<'a>>
...
+    /// source, to expand a stub on demand during witness generation.
+    pub fn expand(&self) -> Result<crate::hash_tree::HashTree, crate::hash_tree::HashTreeError> {
+        (self.expander)(self)
+    }
+}
+
+impl PartialEq for SubtreeSource {
+    /// A conservative, false-negative-only reuse-gate; see the type-level note.
+    fn eq(&self, other: &Self) -> bool {
+        self.addr() == other.addr() && std::ptr::fn_addr_eq(self.expander, other.expander)
```
mraszyk: Why do we need `&& std::ptr::fn_addr_eq(self.expander, other.expander)`? Isn't `self.addr() == other.addr()` sufficient?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -16,6 +17,17 @@ const NUMBER_OF_CERTIFICATION_THREADS: u32 = 16;
...
 
+/// A fork with fewer than this many (expensive to build) children is always
+/// built sequentially: it is too small for the thread pool to pay for itself.
+pub const PARALLEL_MIN_CHILDREN: usize = 1000;
+
+/// When building against a baseline, a large fork starts building sequentially
+/// and samples this many children before extrapolating, over the whole fork, the
+/// rate at which they have to be *built* — i.e. hashed, rather than cheaply
+/// reused from the baseline. With no baseline nothing can be reused, so the
+/// switch to the thread pool happens immediately, without a warmup.
```
mraszyk: This explanation sounds a bit confusing to me.

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -284,6 +321,32 @@ pub struct HashTree {
...
+    /// subtree's contents are not materialized; when needed for building a witness
+    /// they are rebuilt on demand from the `SubtreeSource` (see
+    /// [`HashTree::witness`]).
+    stubs: Vec<Vec<StubNode>>,
+}
+
+/// A reusable subtree collapsed to a single digest ("stub"), stored in a
+/// [`NodeKind::Stub`] node.
+///
+/// Holds one `Arc` (inside the [`SubtreeSource`]) plus a cheap [`Digest`], so it
```
mraszyk: plus a function pointer

rs/canonical_state/tree_hash/src/lazy_tree.rs
```diff
@@ -116,3 +134,94 @@ pub fn follow_path<'a>(t: &LazyTree<'a>, path: &[&[u8]]) -> Option<LazyTree<'a>>
...
+    /// Creates a handle that shares ownership of the subtree's `source` and can
+    /// rebuild the subtree from it, via the `expander`.
+    pub fn new<T: Any + Send + Sync>(source: &Arc<T>, expander: SubtreeExpander) -> Self {
+        let this = Self {
+            source: Arc::clone(source) as Arc<dyn Any + Send + Sync>,
+            expander,
+        };
+        debug_assert!((expander)(&this).is_ok());
+        this
+    }
```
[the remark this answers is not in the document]
alin-at-dfinity: Ensuring correct pairing of source and expander is more important than whatever performance gains are to be had in debug builds. We're talking subtrees (e.g. canister states) with 4 nodes here, so expanding them only makes a difference if there are hundreds of thousands of them and we're building a `HashTree` a couple of times per second.

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -666,6 +778,24 @@ impl HashTree {
...
+                return match t {
+                    // Requested partial tree descends into the subtree: rebuild it from source and
+                    // continue witness generation there.
+                    LabeledTree::SubTree(children) if !children.is_empty() => {
+                        let expanded = ht.stubs[pos.bucket()][pos.index()]
+                            .source
+                            .expand()
+                            .expect("expanding a stub should not fail");
+                        go::<B>(&expanded, NodeId::empty(), expanded.root, t)
+                    }
```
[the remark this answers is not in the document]
alin-at-dfinity: This is why we have a `debug_assert` in the constructor. This would only panic due to a bug in the code (a bad pairing of source `Arc` and expander bundled into a `SubtreeSource`.

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -814,12 +1047,76 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
+
+    /// Builds the hash tree for `t`, returning the [`NodeId`] of its root.
+    ///
+    /// The hash tree is always materialized; collapsing a subtree fork into a
+    /// digest-only [`NodeKind::Stub`] happens one level up, in [`build_child`].
+    fn build_tree(
         t: &LazyTree<'_>,
         ht: &mut HashTree,
         parent: NodeId,
         par_strategy: &mut ParStrategy,
```
mraszyk: ParStrategy => ParallelismStrategy or something like that so that "Par" does not read like related to "parent" (the line just above)

mbjorkqvist (approved):
The DeFi-related changes seem to be coming from a sub-optimal merge of `master`, but approving nonetheless!

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -210,17 +246,18 @@ impl NodeIndexRange {
...
 ///
-///  * (0,00) is an empty tree.
-///  * (0,01) is a leaf.
-///  * (0,10) is a labeled node.
-///  * (0,11) is a fork.
+///  * (0,000) is an empty tree.
+///  * (0,001) is a leaf.
+///  * (0,010) is a labeled node.
+///  * (0,011) is a fork.
+///  * (0,100) is a reusable stub.
```
mraszyk: looks inconsistent with
```
const LEAF_KIND: u32 = 0x2000_0000;
const NODE_KIND: u32 = 0x4000_0000;
const FORK_KIND: u32 = 0x6000_0000;
const STUB_KIND: u32 = 0x8000_0000;
```

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -210,17 +246,18 @@ impl NodeIndexRange {
...
 ///
-///  This means that the tree can store at most 2^30 nodes of the same type.  As
+///  This means that the tree can store at most 2^29 nodes of the same type. As
 ///  each tree node has a 32-byte hash associated with it, the tree needs to
-///  occupy at least 32 GiB of data before the index overflows.
+///  occupy at least 16 GiB of data before the index overflows.
```
mraszyk, line 257: > occupy at least 16 GiB of data before the index overflows

what's the point of this statement?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -666,6 +778,24 @@ impl HashTree {
             pos: NodeId,
             t: &LabeledTree<Vec<u8>>,
         ) -> Result<B, WitnessGenerationError<B>> {
+            if pos.kind() == NodeKind::Stub {
+                // A stub, only storing its root digest.
+                return match t {
+                    // Requested partial tree descends into the subtree: rebuild it from source and
+                    // continue witness generation there.
+                    LabeledTree::SubTree(children) if !children.is_empty() => {
+                        let expanded = ht.stubs[pos.bucket()][pos.index()]
```
mraszyk, line 788:
```suggestion
                        let expanded = ht.stubs[pos.bucket() - ht.bucket_offset][pos.index()]
```
or is it guaranteed that `ht.bucket_offset` is always zero because this never runs during tree construction?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -674,12 +804,16 @@ impl HashTree {
                         B::make_node(label.clone(), B::make_pruned(ht.digest(child).clone()))
                     }
                     HashTreeView::Fork(digest, _left, _right) => B::make_pruned(digest.clone()),
+                    // Intercepted above; a stub behaves like an opaque subtree.
+                    HashTreeView::Stub(digest) => B::make_pruned(digest.clone()),
```
mraszyk: this should be unreachable, right?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -674,12 +804,16 @@ impl HashTree {
...
                 }),
                 LabeledTree::SubTree(children) if children.is_empty() => Ok(match ht.view(pos) {
                     HashTreeView::Empty => B::make_empty(),
                     HashTreeView::Leaf(digest) => B::make_pruned(digest.clone()),
                     HashTreeView::Fork(digest, _left, _right) => B::make_pruned(digest.clone()),
                     HashTreeView::Node(digest, _label, _child) => B::make_pruned(digest.clone()),
+                    // Intercepted above; a stub behaves like an opaque subtree.
+                    HashTreeView::Stub(digest) => B::make_pruned(digest.clone()),
```
mraszyk: this should be unreachable, right?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -719,6 +856,10 @@ impl PartialEq<crypto::HashTree> for HashTree {
         fn eq_recursive(ht: &HashTree, ht_root: NodeId, other: &crypto::HashTree) -> bool {
             ht.digest(ht_root) == other.digest()
                 && match (ht_root.kind(), other) {
+                    // A stub collapses a whole subtree to its root digest, which the top-level
+                    // digest comparison above already checked; there is no materialized structure
+                    // left to compare.
+                    (NodeKind::Stub, _) => true,
```
mraszyk: shouldn't we materialize the stub and compare recursively? if hash equality suffices, then why do we perform any descent from the root at all (after checking the root hashes for equality)?

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -1017,9 +1359,18 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
+        tree,
+        node: NodeId::empty(),
+    });
 
+    let mut ht = HashTree::new();
+    let strategy = &mut ParStrategy::Concurrent;
+    // The root is always materialized; only *descendants* that carry a
+    // `subtree_source` are collapsed into stubs (see `build_child`). Building a
+    // stand-alone subtree is just `hash_lazy_tree` on that subtree's root, which
+    // is in turn materialized for the same reason.
```
mraszyk: `hash_lazy_tree` is not called recursively though so not sure what this means to say

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -863,40 +1160,67 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
+
+                    num_built += was_built as usize;
+                    num_processed += 1;
+                    do_parallelize |= may_parallelize
+                        // Beyond the warmup, switch to parallel once the sampled build rate
+                        // (`num_built / num_processed`) projects more than the number of children
+                        // required for parallel processing over all `num_children` (rearranged to avoid
+                        // division).
+                        && num_processed >= ADAPTIVE_WARMUP_CHILDREN
+                            && num_built * num_children >= PARALLEL_MIN_CHILDREN * num_processed;
```
mraszyk, line 1240: should there be one more condition that at least `PARALLEL_MIN_CHILDREN` are still left - we shouldn't spawn a threadpool if only a few children are left at this point (e.g., if there are only `PARALLEL_MIN_CHILDREN` + 42 many children)

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -863,40 +1160,67 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
+                // the *remaining* children to the thread pool. This covers both stubbed forks
+                // (where reuse keeps the rate low) and regular forks (where every child is
+                // materialized; so a large fork always parallelizes).
                 //
-                // We do not pass the thread pool down after use, so we are not spawning new threads
-                // in a nested way.
-                if num_children > 100 && par_strategy.is_concurrent() {
-                    fork_parallel(
+                // We only materialize the unprocessed tail into a `Vec` if and when we
+                // switch; the common, all-sequential path stays fully lazy.
```
mraszyk: What does this mean? in particular, the distinction between "materialized" and "lazy"

alin-at-dfinity:
FWIW, in `bench_traversal` with 500k canisters with `ExecutionState` this change has virtually no effect on building a cold `HashTree` (480 ms -> 490 ms).

But it improves the hot path (reusing subtree digests from an earlier `HashTree`) by 35%+ (to just under 300 ms).

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -970,21 +1300,32 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
-                        match root {
-                            Ok(root) => {
+                            depth,
+                            *base,
+                        ) {
+                            Ok((root, _was_built)) => {
+                                let children_range = if root.kind() == NodeKind::Stub {
+                                    NodeIndexRange::default()
+                                } else {
+                                    ht.root_labels_range.clone()
```
mraszyk, line 1375: this is a "hack", right? we set the parent as "root" (via `NodeId::empty()`) in the above call to build_child so that its labels range is set as `ht.root_labels_range` (over-writen for every new child which is fine since splice_subtree ignores root_labels_range anyway)

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -970,21 +1300,32 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
+                        // `subtree_source` is collapsed to a stub here.
+                        //
+                        // A stub has no materialized labeled children of its own, so its
+                        // `children_range` is empty (and is never consulted: stubs are descended into
+                        // via their source during witness generation).
+                        match build_child(
                             child,
                             &mut ht,
                             NodeId::empty(),
+                            // Run with `ParStrategy::Sequential`, so thread pools are never nested.
```
mraszyk: this is likely also a correctness invariant for splice_subtree, right?

daniel-wong-dfinity-org-twin:
AFAICT, Governance approval is not needed here and I'm being emailed in error. If that's not right, please, contact me in Slack.

rs/canonical_state/tree_hash/src/lazy_tree.rs
alin-at-dfinity: Changed to

> It is a plain function pointer (not a closure)

In order to pass a closure, you would need something like `Box<dyn Fn>` instead of `fn`. Which is heavier, requires an explicit memory allocation (and we clone these things a lot) and would break the `fn_addr_eq()` we use below to check whether it's the same certification version (implied by equal function pointers).

rs/canonical_state/tree_hash/src/lazy_tree.rs
alin-at-dfinity: Would look nicer, but unfortunately it doesn't work. I did manage to get rid of the parentheses around `expander` though:
```rust
debug_assert!(expander(&this).is_ok());
```
That's only required for something like `(self.expander)(self)`, further below.

rs/canonical_state/tree_hash/src/lazy_tree.rs
alin-at-dfinity: It's not. We need the same data AND the same certification version. Equal function pointers implies the same certification version. (The reverse is not true, you could have two functions that produce output with the same certification version, but we don't have that and we wouldn't care anyway.)

It's described in the `SubtreeSource` doc comment. If that's not clear or it's too far from the spot, I can move it / duplicate it here.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Rewritten, PTAL.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: No, it's exactly right. These are the first 3 bits of the ID, so `001` actually stands for `0b0010_0000_0000_0000_0000_0000_0000_0000` or `0x2000_0000`.

Looking at it more closely though, it should probably have been something like `(0, 001...)`, although that's not exactly clear either.

Happy to take suggestions.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: I don't know for sure. Probably to reassure the reader that things would definitely grind to a halt before the data structure fills up. But I guess pointing out that `2^29` is 500M nodes (of any given type) would have been sufficient.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Precisely. `bucket_offset` is indeed only non-zero while building chunks of a `LazyFork` in parallel.

This is similar to code further below in this same function. In code that I added (`BaselineCursor`) there are explicit `debug_assert_eq!(self.tree.bucket_offset, 0)` checks. But in that case too, we're dealing with an earlier, fully built `HashTree`, so it is safe to assume all-zero `bucket_offsets`.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Indeed. AI-generated code.

Switched to `unreachable!()`.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Right.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Good point, done.

This code was first written for a version of the code in which stub nodes were actually only digests (and probably the `Arc` too, but only for equality comparison reasons); and you had to pass the original `LazyTree` to `HashTree::witness()` in order to expand it.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Changed to:

```rust
                // We only collect the unprocessed tail into a `Vec` if and when we switch; the
                // common, all-sequential path uses the `joined` iterator directly.
```

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: I suppose we could, but it doesn't matter much. It's unlikely that we are rebuilding just over `PARALLEL_MIN_CHILDREN` and that we notice it only when we've done the vast majority of the work sequentially. The two likely cases are (i) we reuse most of the children (so this condition doesn't trigger) or (ii) we can't reuse any (e.g. we've just charged everyone for storage) or they're not reusable (in which case we'd trigger as soon as we've passed `ADAPTIVE_WARMUP_CHILDREN`.

I've reduced `ADAPTIVE_WARMUP_CHILDREN` to 500 (half of `PARALLEL_MIN_CHILDREN`), so we also avoid the "1001 children, all need rebuilding, do 1000 sequentially and 1 in parallel" case.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Good point, updated.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: I suppose that on the one hand it's just how you build an independent `HashTree` (you start with the root); and on the other hand, you then need to record the range of its children. This code was already here, see a couple of lines below.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: It refers to (temporarily) building stand-alone subtrees in order to compute the root hash of stub nodes. Let me know if you have a better wording suggestion.

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: > explicit debug_assert_eq!(self.tree.bucket_offset, 0) checks

how about adding them here, too?

rs/canonical_state/tree_hash/src/lazy_tree.rs
mraszyk: What if
```
    // The bare address, used for identity comparison.
    fn addr(&self) -> *const () {
        Arc::as_ptr(&self.) as *const ()
    }
```
(I thought this is what `self.addr()` returns when writing this comment.)

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: `(0, 0x0010...)` would be easier to match against the const definitions of the individual kinds above.

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: So you want to point out that we use `build_tree` as opposed to `build_child` here? If so, I'm not actually sure why we couldn't use `build_child` here, too.

Or you want to point out that we need to assign the `root` field? If so, I'd justify it by the fact that the _recursive_ `build_child`/`build_tree` functions cannot set the root (because they are recursive).

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: > This code was already here, see a couple of lines below.

Indeed, and I believe your change makes sense, I just wanted to confirm my understanding of using `ht.root_labels_range.clone()` repeatedly on a shared tree.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: I guess the point of this comment is to clarify that (i) the root of the tree is never a stub; and (ii) when building a stand-alone subtree (i.e. for a stub) we do it by calling `build_tree` (the comment was out of date) rather than `hash_lazy_tree` or `build_child` (which would end up in an infinite loop, creating a stub for a stub, etc.).

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Oh, I see what you mean now. As said, the code was already here, so I clearly didn't read it as carefully as you did. )o:

But yeah, it appears that we repeatedly add parentless subtrees to the tree, then take the resulting `root_labels_range` and use it as the `children_range` of the subtree that was just created. The implication being that `root_labels_range` is rewritten rather than extended on every iteration.

rs/canonical_state/tree_hash/src/lazy_tree.rs
alin-at-dfinity: We don't just want to compare the address of `self` with the address of `other`. They would virtually always be different (one is from a previous `HashTree` build, one is just created). What we want to ensure is that `self.source` is the same `Arc` as `other.source` (i.e. the data hasn't changed) and `self.expander` is the same function as `other.expander (indirectly, we are encoding the same certification version; or more precisely, we are not encoding a different certification version).

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: Thanks for confirming! Maybe we could capture this in a comment, but since it's unrelated to this PR's goal, I'll leave the decision up to you (feel free to close this thread anytime).

rs/canonical_state/tree_hash/src/lazy_tree.rs
mraszyk: > They would virtually always be different (one is from a previous HashTree build, one is just created).

That's what I missed - thanks for explaining!

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -814,12 +1060,76 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
         }
     }
 
-    fn go(
+    /// Builds one labeled `child` of a fork (linked under `parent`), returning its
```
mraszyk: What is a bit confusing here is that `child` must be a `LazyFork(f)`, but in principle a SubtreeSource could be associated to any other variant of LazyTree, e.g., we could (in a follow-up PR) unify blob hash caching using SubtreeSource.

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: > (i) the root of the tree is never a stub

that's true, but rather coincidental, I think we could call build_child here as well (which would make no difference in practice at this point, because the entire tree has no `SubtreeSource` associated with it, but for the sake of explaining the design, I don't see a reason why the root cannot be a stub);

> build_child (which would end up in an infinite loop, creating a stub for a stub, etc.)

I don't think so: build_tree always calls build_child on _proper_ subtrees.

My suggestion would be to drop the comment entirely.

alin-at-dfinity: > I don't think so: build_tree always calls build_child on _proper_ subtrees.

Precisely. If you were to call `build_tree` on a node that is supposed to be a stub, it would create a stub; which would then call `build_tree`; which could then create a stub; etc.

So the point of the comment is to say that `build_tree` will always materialize the tree, regardless of whether it's supposed to be stubbed or not. Because stubbing relies on it to do so.

But if you think this is the wrong place to state that; or that it's obvious; then I can definitely drop the comment.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Actually, on my dev branch I've switched to parents designating their children as stubbable (by optionally providing a different iterator, parallel to `children()`; a method that returns an `Option<Iterator<Item=SubtreeSource>>`).

This is precisely because it makes it easier to stub ingress messages (which have 3 different "shapes", so you'd have to implement `subtree_source()` for each); and possible to implement stubbing at all for canister messages (which are leaves).

Would you like me (or Claude) to try backporting that refactor to this PR?

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: > So the point of the comment is to say that build_tree will always materialize the tree, regardless of whether it's supposed to be stubbed or not. Because stubbing relies on it to do so.

I see, that comment might better read as a doc comment of `fn build_tree` itself.

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: Wouldn't it be easier if `fn subtree_source(&self) -> Option<SubtreeSource>` is a method on `LazyTree`?

If not, then

> stubbable (by optionally providing a different iterator, parallel to children(); a method that returns an Option<Iterator<Item=SubtreeSource>>).

How about:
```
fn children(&self) -> Box<dyn Iterator<Item = (Label, LazyTree<'a>, Option<SubtreeSource>)> + '_>;
```
then?

alin-at-dfinity: Creating the `LazyTree` is somewhat expensive (it requires an extra allocation and drop). It is more efficient to return either `LazyTree` or `SubtreeSource` than both.

Similarly, making `subtree_source()` a method of `LazyTree`, would mean first creating the `LazyTree` only to immediately drop it (if it returns a source that is unchanged).

mraszyk: > Creating the LazyTree is somewhat expensive (it requires an extra allocation and drop).

But this is what happens in the current implementation, the child must be built explicitly _and_ have the form `if let LazyTree::LazyFork(f) = child`. So I suppose you're talking about your branch in which you want to avoid this by
- allowing any child to be stubbed (including blobs);
- making the parent provide the subtree source.

I don't know much performance we'd additionally gain, but it sounds quite a bit more complex.

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -1310,14 +1329,14 @@ fn hash_lazy_tree_impl(
...
-                        // `subtree_source` is collapsed to a stub here.
+                        // and fix the link from `root` to the parent later. A `Child::Stub` is
+                        // collapsed to a stub here.
                         //
                         // A stub has no materialized labeled children of its own, so its
                         // `children_range` is empty (and is never consulted: stubs are descended into
                         // via their source during witness generation).
                         match build_child(
-                            child,
+                            child.clone(),
```
mraszyk: Why do we need to clone here?

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: It's not more complex, it mostly just shifts code around. Here's the commit, I believe you've already seen it: 119c532ef719c3a85a40f9d541fe9871d86e50a3.

> But this is what happens in the current implementation, the child must be built explicitly

Which is one of the reasons why I switched to parents declaring children as stubbable.

> I don't know much performance we'd additionally gain

I just benchmarked it again, after pulling in yet another commit (44d89f707470a1f9313a8f726992dfcd4266c98f) from my work branch having to do with exactly the `clone()` you singled out. Nice job, I needed a CPU profile to point it out to me! (o:

This is about 15% faster on the slow path (w/ no baseline); and around 5% faster on the fast path (w/ unchanged baseline).

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Because we have a `Vec<(..., Child, ...)>` (`tail`) as input to `build_fork_parallel()`. And then we use `tail.chunks(per_thread)` on it to break it into per-thread chunks. This returns immutable slices, hence the need to clone. There's also a `chunks_mut()`, that would return `&mut []` slices, but that still won't allow us to move values out of the underlying `Vec`.

So I've added a follow-up commit (44d89f707470a1f9313a8f726992dfcd4266c98f) that takes the iterator as an argument (instead of the `Vec`) and builds 16 separate `Vecs` out of it. Then we don't need the clone and things are meaningfully faster (and much easier to work with for other node types).

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done, only left behind a one-liner.

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Done.

rs/canonical_state/tree_hash/src/hash_tree.rs
mraszyk: Looks great - I like the new version better.

rs/canonical_state/tree_hash/src/hash_tree.rs
```diff
@@ -1004,22 +1405,33 @@ pub fn hash_lazy_tree(t: &LazyTree<'_>) -> Result<HashTree, HashTreeError> {
...
-    let mut ht = HashTree::new();
-    ht.root = go(t, &mut ht, NodeId::empty(), &mut ParStrategy::Concurrent, 0)?;
+    let baseline = baseline.map(|tree| BaselineCursor {
+        tree,
+        node: NodeId::empty(),
+    });
 
+    let mut ht = HashTree::new();
+    let strategy = &mut ParallelismStrategy::Concurrent;
+    // Always materialize the root, avoiding infinite recursion for stub subtrees.
```
mraszyk:
```suggestion
    // Always materialize the root, stubbing the root doesn't help in practice (the value of time stored in the tree always changes) and thus it is only supported for descendants of a fork.
```

mraszyk (approved):
Great work!

rs/canonical_state/tree_hash/src/hash_tree.rs
alin-at-dfinity: Reworded. Mentioning stubbed subtrees here does not make sense, indeed.

PR #10536, opened 2026-06-22
feat(Core): Switch to deterministic memory tracker

michael-weigelt:
This PR unifies page instruction charging. The deterministic memory tracker (DMT) now charges for all combinations of {heap, stable} x {read, write} page access types. The cost is uniform and increased to 5000 instructions per OS page (4KiB). 

Unchanged is the additional cost of copying dirty pages after the message execution (3000 instructions per page). 

So overall, the costs change like this: 

```
               previously     dmt   
heap read:              0    5000  
heap write:         *1000   *5000 
stable read:            0    5000  
stable write:       *1000   *5000   

* additionally charged for copy overhead after the message (+3000)
```

The dirty stable pages tracking via the injected system API replacements is changed: The instruction counter is no longer decreased, because the DMT does it automatically. But the bitmap tracking of stable pages is kept, because it's still needed to limit overall stable memory access. 

As a consequence of this change, heap accesses are charged deterministically and immediately (during the message/slice, rather than afterwards), which means that messages with lots of heap accesses are DTS'd properly. While this incurs additional cycle costs, it will allow us to increase the Wasm heap size to much more than the current 6GiB limit.


rs/config/src/subnet_config.rs
```diff
@@ -149,7 +149,7 @@ pub const DEFAULT_REFERENCE_SUBNET_SIZE: usize = 13;
 pub const SEV_REFERENCE_SUBNET_SIZE: usize = 7;
 
 /// Costs for each newly created dirty page in stable memory.
```
schneiderstefan, line 151: Is this comment still accurate?

michael-weigelt: No. I plan to do a pass over the naming next and I will include this there.

michael-weigelt:
Squashed into https://github.com/dfinity/ic/pull/10789

PR #10541, opened 2026-06-22
fix: Fix nightly baremetal tests

frankdavid:
After upgrading Bazel to the latest version, the path `IdracRedfishSupport-0.0.8.data` was changed to `idracredfishsupport-0.0.8.data` and the scripts were no longer found. To be more flexible, the implementation now searches for the dependency under `*.data/scripts`.

In addition, reuse the same bazel output base in all bazel tests in `Run Bazel Launch Bare Metal`.


Bownairo (approved):

.github/workflows/schedule-daily.yml
basvandijk, line 86: @nmattia do you think this step should use:
```
uses: ./.github/actions/bazel
```
like our other steps that invoke bazel?

That way:
* The output_base is automatically set for all bazel invocations.
* We get BuildBuddy links.
* We use our Bazel remote cache.

nmattia: SGTM! do you know why it wasn't originally used?

frankdavid: I assume the original author did not consider it. I added it, PTAL.

basvandijk: Look good. Thanks!

basvandijk (approved):

PR #10545, opened 2026-06-23
feat: New features for bare-metal deployment

frankdavid:
- Support deploying a GuestOS *upgrade image* to the selected target boot alternative. Optionally enable wiping the corresponding var partition. Previously, only full GuestOS disk images could be deployed thereby overwriting all (including the data) partition.
- Support deploying a released HostOS/GuestOS version by specifying the 40-char git hash.
- Some fixes, e.g. don't mount over the entire /opt/ic/bin library (this led to some issues previously), only the necessary binaries/scripts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1146,6 +1146,7 @@ dependencies = [
  "anyhow",
  "clap 4.6.0",
  "deterministic_ips",
+ "grub",
  "http 1.4.0",
  "macaddr",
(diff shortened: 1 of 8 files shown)


basvandijk (approved):
`rs/tests/driver/src/driver/bootstrap.rs` LGTM

r-birkner (approved):

PR #10547, opened 2026-06-23
fix(Rosetta): DEFI-2902: Enforce a 24h ingress window in Rosetta /construction/payloads

mbjorkqvist:
## Why

The `/construction/payloads` endpoint of both ICP Rosetta and ICRC1 Rosetta accepts caller-controlled `ingress_start`/`ingress_end` values and expands them into a list of ingress expiries without enforcing the 24-hour bound documented on `ConstructionPayloadsRequestMetadata`. This lets an unauthenticated request drive the expiry loop to exhaust memory and crash an operator's Rosetta instance.

Two distinct payloads trigger it: an oversized window (a huge `ingress_end - ingress_start` span), and a near-`u64::MAX` window whose span is tiny but whose increment wraps so the loop never terminates.

## What

`validate_ingress_window` now runs before the loop in both services and rejects a request when either:
- the span `ingress_end - ingress_start` exceeds 24h, or
- `ingress_end` is more than 24h in the future.

The second bound is what closes the wraparound payload — its span is sub-minute, so a span-only check would not catch it.

The change is built up across three commits: a regression test characterizing the current (unbounded) behavior, the enforcement itself, and the flip of those tests to assert rejection plus a negative control and a boundary case. The existing roundtrip proptests are adapted to the new behavior.

Scope is limited to operator-deployed Rosetta instances; IC mainnet, NNS, consensus and ledger canisters are unaffected.

### DIFF
--- a/rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
+++ b/rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
@@ -43,7 +43,9 @@ const MAX_INGRESS_WINDOW: Duration = Duration::from_secs(24 * 60 * 60);
 ///
 /// Rejects windows that would make the expiry loop iterate an excessive (or, on
 /// arithmetic wraparound, unbounded) number of times (see ICPBB-134 /
-/// DEFI-2902). Two independent bounds are enforced:
+/// DEFI-2902). It enforces that:
+/// - `ingress_start` is strictly before `ingress_end` (an empty or reversed
+///   window would otherwise produce an empty/degenerate set of payloads),
 /// - the span `ingress_end - ingress_start` must not exceed 24h, and
 /// - `ingress_end` must not be more than 24h in the future, which also rejects
 ///   the near-`u64::MAX` payloads that would otherwise wrap the loop counter.
@@ -57,6 +59,11 @@ fn validate_ingress_window(
     let start = ingress_start.as_nanos_since_unix_epoch();
     let end = ingress_end.as_nanos_since_unix_epoch();
 
+    if start >= end {
+        return Err(ApiError::invalid_request(
+            "ingress_start must be strictly before ingress_end.",
+        ));
+    }
     if end.saturating_sub(start) > max_window {
         return Err(ApiError::invalid_request(format!(
             "The ingress window (ingress_end - ingress_start) must not exceed {} hours.",
@@ -1187,6 +1194,14 @@ mod tests {
     fn near_u64_max_ingress_window_is_rejected() {
         let start = t(u64::MAX - 50 * 1_000_000_000);
         assert!(validate_ingress_window(t(NOW_NANOS), start, t(u64::MAX)).is_err());
+    }
+
+    // A reversed or empty window (ingress_end <= ingress_start) is rejected
+    // rather than silently producing an empty set of payloads.
+    #[test]
+    fn reversed_ingress_window_is_rejected() {
+        assert!(validate_ingress_window(t(NOW_NANOS), t(NOW_NANOS + 1), t(NOW_NANOS)).is_err());
+        assert!(validate_ingress_window(t(NOW_NANOS), t(NOW_NANOS), t(NOW_NANOS)).is_err());
     }
 
     // Negative control: a realistic window (5 minutes ahead of now) is accepted.
--- a/rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
+++ b/rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
@@ -34,6 +34,20 @@ use rosetta_core::convert::principal_id_from_public_key;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 use tracing::log::debug;
 
+/// Validate the caller-provided ingress window before it is expanded into a
+/// list of ingress expiries.
+///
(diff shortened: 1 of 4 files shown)


rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
```diff
@@ -34,6 +34,44 @@ use rosetta_core::convert::principal_id_from_public_key;
...
+fn validate_ingress_window(
+    now: ic_types::time::Time,
+    ingress_start: ic_types::time::Time,
+    ingress_end: ic_types::time::Time,
+) -> Result<(), ApiError> {
+    let max_window = MAX_INGRESS_WINDOW.as_nanos() as u64;
+    let now = now.as_nanos_since_unix_epoch();
+    let start = ingress_start.as_nanos_since_unix_epoch();
+    let end = ingress_end.as_nanos_since_unix_epoch();
+
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — fixed in bd5d72891a. `validate_ingress_window` now rejects `ingress_start >= ingress_end` before the span/future checks, so the `saturating_sub`-based span check no longer treats a reversed/empty window as a zero-length (valid) span. ICP had no prior start/end ordering guard (unlike ICRC1's handler), so this closes that gap and avoids the empty `ingress_expiries` outcome. Added a `reversed_ingress_window_is_rejected` regression test.

rs/rosetta-api/icrc1/src/construction_api/services.rs
```diff
@@ -461,6 +536,20 @@ mod tests {
...
+                        let eff_end = payloads_metadata
+                            .ingress_end
+                            .unwrap_or(eff_start + ingress_interval);
+                        let max_window = MAX_INGRESS_WINDOW.as_nanos() as u64;
+                        if eff_end.saturating_sub(eff_start) > max_window
+                            || eff_end.saturating_sub(now) > max_window
+                        {
+                            assert!(construction_payloads_response.is_err());
+                            continue;
+                        }
```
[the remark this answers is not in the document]
mbjorkqvist: Done in bd5d72891a — `test_construction_parse` now calls `validate_ingress_window(now, eff_start, eff_end)` directly instead of re-implementing the bound inline, so the test can no longer drift from the production validation.

rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
```diff
@@ -1110,3 +1157,68 @@ fn neuron_subaccount(
...
+    const HOUR_NANOS: u64 = 60 * 60 * 1_000_000_000;
+    // A realistic "now" (~2023) that is well away from the u64 boundary.
+    const NOW_NANOS: u64 = 1_700_000_000 * 1_000_000_000;
+
+    fn t(nanos: u64) -> Time {
+        Time::from_nanos_since_unix_epoch(nanos)
+    }
+
+    // Vector A (ICPBB-134): a window far larger than the documented 24h bound is
```
gregorydemay: nit: do we want the bugbounty number to appear in the code? Maybe better to keep it only in the Jira ticket?

rs/rosetta-api/icrc1/src/construction_api/services.rs
```diff
@@ -108,6 +108,32 @@ pub fn construction_combine(
...
+/// list of ingress expiries.
+///
+/// Rejects windows that would make the expiry loop iterate an excessive (or, on
+/// arithmetic wraparound, unbounded) number of times (see ICPBB-134 /
+/// DEFI-2902). Two independent bounds are enforced:
+/// - the span `ingress_end - ingress_start` must not exceed 24h, and
+/// - `ingress_end` must not be more than 24h in the future, which also rejects
+///   the near-`u64::MAX` payloads that would otherwise wrap the loop counter.
+fn validate_ingress_window(now: u64, ingress_start: u64, ingress_end: u64) -> Result<(), Error> {
+    let max_window = MAX_INGRESS_WINDOW.as_nanos() as u64;
```
gregorydemay, line 122: Contrary to the validating function in ICP Rosetta the check
```
    if start >= end {
        return Err(ApiError::invalid_request(
            "ingress_start must be strictly before ingress_end.",
        ));
    }
```

seems to be missing, is that on purpose?

gregorydemay (approved):
Thanks @mbjorkqvist ! Just a comment asking why the validating functions are slightly different.

rs/rosetta-api/icp/src/request_handler/construction_payloads.rs
mbjorkqvist: Good point — dropped the `ICPBB-134` (and `DEFI-2902`) identifiers from the code comments in 2cb6d4cafd; the ticket reference lives in the PR title now.

rs/rosetta-api/icrc1/src/construction_api/services.rs
mbjorkqvist: Not intentional as a behavioral difference — the check did exist, but as a separate guard in `construction_payloads` itself rather than in `validate_ingress_window`, which is why the two validators looked different. I've moved it into the ICRC1 `validate_ingress_window` (and removed the now-redundant caller guard) in 2cb6d4cafd, so both validators now enforce the same set of checks (`start < end`, span <= 24h, `ingress_end <= now + 24h`).

PR #10551, opened 2026-06-23
feat: Add fuse2fs tool built from e2fsprogs

frankdavid:
fuse2fs will be used to extract the contents of released ext4 images without root for automated creation of SEV recovery images, it comes from the same e2fsprogs package that we already have but a few compile options had to be tweaked. In addition, libfuse must be provided as a dependency.

### DIFF
--- a/third_party/BUILD.libfuse.bazel
+++ b/third_party/BUILD.libfuse.bazel
@@ -29,6 +29,5 @@ meson(
         " && test -n \"$$libfuse_archive\"" +
         " && cp \"$$libfuse_archive\" $$INSTALLDIR/lib/libfuse3.a"
     ),
-    #visibility = ["//visibility:private"],
     visibility = ["//visibility:public"],
 )
--- a/third_party/BUILD.libfuse.bazel
+++ b/third_party/BUILD.libfuse.bazel
@@ -12,7 +12,6 @@ filegroup(
 meson(
     name = "libfuse",
     env = {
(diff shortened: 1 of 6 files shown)


third_party/BUILD.libfuse.bazel
```diff
@@ -0,0 +1,36 @@
...
+    },
+    lib_source = ":all_srcs",
+    postfix_script = (
+        "libfuse_archive=$$(find $$INSTALLDIR/lib -name libfuse3.a -print -quit)" +
+        " && test -n \"$$libfuse_archive\"" +
+        " && cp \"$$libfuse_archive\" $$INSTALLDIR/lib/libfuse3.a"
+    ),
+    out_shared_libs = [],
+    out_static_libs = ["libfuse3.a"],
+    #visibility = ["//visibility:private"],
```
basvandijk:
```suggestion
```

third_party/BUILD.libfuse.bazel
```diff
@@ -0,0 +1,36 @@
+load("@//bazel:defs.bzl", "disable_hermetic_cc_target")
```
basvandijk: I don't see `disable_hermetic_cc_target` being called in this file. Am I missing something?

frankdavid: No, it's not needed, it got cleaned up by the automation.

basvandijk (approved):

Bownairo (approved):

PR #10559, opened 2026-06-24
fix: Use registry version of `CanisterHttpRequestContext` to determine committee membership

eichhorl:
## Background

Before this PR, each node determined the registry version (thus committee) of an HTTP outcall by itself, depending on the current finalized height at the time of processing the request:
https://github.com/dfinity/ic/blob/fe9e1b7a9a98ebafbf38230368226f4e928cc14f/rs/https_outcalls/consensus/src/pool_manager.rs#L577-L599

Additionally, HTTP shares with the wrong registry version were filtered out by the payload builder:
https://github.com/dfinity/ic/blob/fe9e1b7a9a98ebafbf38230368226f4e928cc14f/rs/https_outcalls/consensus/src/payload_builder.rs#L212-L213

However, since the exact timing of when an HTTP request is handled on each replica may be different, the current finalized height (and therefore the registry version) may be different, as well.

This has some adverse effects:

1. If the registry version changes after shares were produced, but before they are combined, the HTTP request will time out because shares with the old registry version are filtered out by the payload builder (see [this metric](https://victoria.ch1-obs1.dfinity.network/select/0/vmui/#/?g0.range_input=4d20h20m29s519ms&g0.end_input=2026-06-16T15%3A16%3A04&g0.relative_time=none&g0.tab=0&g0.tenantID=0&g0.expr=increase%28max+%28canister_http_timeouts_delivered%7Bic_subnet%3D%22pzp6e-ekpqk-3c5x7-2h6so-njoeq-mt45d-h3h6c-q3mxf-vpeq5-fk5o7-yae%22%7D%29+%3E+0%29&g1.expr=increase%28max%28consensus_membership_registry_version%7Bic_subnet%3D%22pzp6e-ekpqk-3c5x7-2h6so-njoeq-mt45d-h3h6c-q3mxf-vpeq5-fk5o7-yae%22%7D%29%29&g1.range_input=4d20h20m29s519ms&g1.end_input=2026-06-16T15%3A16%3A04&g1.relative_time=none&g1.tab=0&g1.tenantID=0) for reference).
2. If a node produces some shares just before leaving the subnet, once these shares are received by its peers they might be invalidated, because to the rest of the subnet the node already left.
3. During membership changes, the HTTP outcalls committee will leave the subnet immediately without finishing outstanding requests first.

## Proposed Changes
Instead, this PR proposes to use the registry version of the `CanisterHttpRequestContext`. This registry version is used as the single source of truth for all nodes on the subnet, throughout the entire duration of the request. The registry version included in the share metadata is removed instead.

Additionally, during membership changes, this allows the committee of the request to remain on the subnet until the request context disappears from the state. This is achieved by considering HTTP contexts in `get_oldest_registry_version_in_use_by_repliated_state`.

One implication that this change has is that now, registry versions of signatures passed to `verify_basic_sig_batch_multi_msg` may be different. To solve this, we pass a separate registry version for each signature instead.

### DIFF
--- a/rs/consensus/utils/src/lib.rs
+++ b/rs/consensus/utils/src/lib.rs
@@ -740,19 +740,17 @@ mod tests {
         }
     }
 
-    fn fake_state_with_canister_http_contexts(
-        contexts: Vec<CanisterHttpRequestContext>,
-    ) -> ReplicatedState {
-        fake_state_with_contexts(vec![], vec![], contexts)
-    }
-
     #[test]
     fn test_get_oldest_state_registry_version_canister_http_only() {
-        let state = fake_state_with_canister_http_contexts(vec![
-            fake_canister_http_context(RegistryVersion::from(8)),
-            fake_canister_http_context(RegistryVersion::from(4)),
-            fake_canister_http_context(RegistryVersion::from(6)),
-        ]);
+        let state = fake_state_with_contexts(
+            vec![],
+            vec![],
+            vec![
+                fake_canister_http_context(RegistryVersion::from(8)),
+                fake_canister_http_context(RegistryVersion::from(4)),
+                fake_canister_http_context(RegistryVersion::from(6)),
+            ],
+        );
         assert_eq!(
             Some(RegistryVersion::from(4)),
             get_oldest_state_registry_version(&state)
--- a/rs/consensus/utils/src/lib.rs
+++ b/rs/consensus/utils/src/lib.rs
@@ -583,6 +583,7 @@ mod tests {
     fn fake_state_with_contexts(
         sign_with_threshold: Vec<SignWithThresholdContext>,
         setup_initial_dkg: Vec<SetupInitialDkgContext>,
+        canister_http: Vec<CanisterHttpRequestContext>,
     ) -> ReplicatedState {
         let mut state = ReplicatedStateBuilder::default().build();
         state
@@ -604,12 +605,21 @@ mod tests {
                 .map(|(i, c)| (CallbackId::from(i as u64), c)),
         );
         state
+            .metadata
+            .subnet_call_context_manager
+            .canister_http_request_contexts = BTreeMap::from_iter(
+            canister_http
+                .into_iter()
+                .enumerate()
+                .map(|(i, c)| (CallbackId::from(i as u64), c)),
+        );
+        state
     }
 
     fn fake_state_with_signature_contexts(
         contexts: Vec<SignWithThresholdContext>,
     ) -> ReplicatedState {
-        fake_state_with_contexts(contexts, vec![])
+        fake_state_with_contexts(contexts, vec![], vec![])
     }
 
     fn fake_setup_initial_dkg_context(registry_version: RegistryVersion) -> SetupInitialDkgContext {
@@ -625,7 +635,7 @@ mod tests {
     fn fake_state_with_setup_initial_dkg_contexts(
         contexts: Vec<SetupInitialDkgContext>,
     ) -> ReplicatedState {
-        fake_state_with_contexts(vec![], contexts)
+        fake_state_with_contexts(vec![], contexts, vec![])
     }
 
     fn fake_key_ids() -> Vec<MasterPublicKeyId> {
@@ -730,26 +740,18 @@ mod tests {
         }
     }
 
-    fn fake_state_with_canister_http_contexts(versions: Vec<RegistryVersion>) -> ReplicatedState {
-        let mut state = ReplicatedStateBuilder::default().build();
-        state
-            .metadata
-            .subnet_call_context_manager
-            .canister_http_request_contexts = BTreeMap::from_iter(
-            versions
-                .into_iter()
-                .enumerate()
-                .map(|(i, v)| (CallbackId::from(i as u64), fake_canister_http_context(v))),
-        );
-        state
+    fn fake_state_with_canister_http_contexts(
+        contexts: Vec<CanisterHttpRequestContext>,
+    ) -> ReplicatedState {
+        fake_state_with_contexts(vec![], vec![], contexts)
     }
 
     #[test]
     fn test_get_oldest_state_registry_version_canister_http_only() {
         let state = fake_state_with_canister_http_contexts(vec![
-            RegistryVersion::from(8),
-            RegistryVersion::from(4),
-            RegistryVersion::from(6),
+            fake_canister_http_context(RegistryVersion::from(8)),
+            fake_canister_http_context(RegistryVersion::from(4)),
+            fake_canister_http_context(RegistryVersion::from(6)),
         ]);
         assert_eq!(
             Some(RegistryVersion::from(4)),
@@ -762,21 +764,15 @@ mod tests {
         // Sign and setup-dkg contexts at v5, canister http at v2: the canister
         // http version must be reflected as the oldest.
         let key_id = fake_key_ids().into_iter().next().unwrap();
-        let mut state = fake_state_with_contexts(
+        let state = fake_state_with_contexts(
             vec![fake_signature_request_context_with_registry_version(
                 Some(PreSigId(0)),
                 &key_id,
                 RegistryVersion::from(5),
             )],
             vec![fake_setup_initial_dkg_context(RegistryVersion::from(5))],
+            vec![fake_canister_http_context(RegistryVersion::from(2))],
         );
-        state
-            .metadata
-            .subnet_call_context_manager
-            .canister_http_request_contexts = BTreeMap::from([(
-            CallbackId::from(0),
-            fake_canister_http_context(RegistryVersion::from(2)),
-        )]);
         assert_eq!(
             Some(RegistryVersion::from(2)),
             get_oldest_state_registry_version(&state)
@@ -799,6 +795,7 @@ mod tests {
         let state = fake_state_with_contexts(
             signature_contexts,
             vec![fake_setup_initial_dkg_context(RegistryVersion::from(2))],
+            vec![],
         );
         assert_eq!(
             Some(RegistryVersion::from(2)),
@@ -822,6 +819,7 @@ mod tests {
         let state = fake_state_with_contexts(
             signature_contexts,
             vec![fake_setup_initial_dkg_context(RegistryVersion::from(11))],
+            vec![],
         );
         assert_eq!(
             Some(RegistryVersion::from(2)),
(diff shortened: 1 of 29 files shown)


rs/consensus/utils/src/lib.rs
```diff
@@ -699,6 +713,76 @@ mod tests {
...
+            get_oldest_state_registry_version(&state)
+        );
+    }
+
+    #[test]
+    fn test_get_oldest_state_registry_version_canister_http_younger_than_others() {
+        // Sign and setup-dkg contexts at v5, canister http at v2: the canister
+        // http version must be reflected as the oldest.
+        let key_id = fake_key_ids().into_iter().next().unwrap();
+        let mut state = fake_state_with_contexts(
```
pierugo-dfinity: Alternatively, `fake_state_with_contexts` could now take a third argument, a vector of canister HTTP contexts and `fake_state_with_canister_http_contexts`'s implementation could look like the existing `fake_state_with_[setup_initial_dkg/signature]_contexts`

rs/crypto/src/sign/basic_sig/tests.rs
```diff
@@ -607,13 +613,90 @@ mod verify_sigs_batch {
...
+                signature: &sig_1,
+                message: &msg_1,
+                registry_version: REG_V2,
+            },
+            BasicSigBatchEntry {
+                signer: NODE_2,
+                signature: &sig_2,
+                message: &msg_2,
+                registry_version: REG_V1,
+            },
```
mraszyk: shouldn't we have two separate cases
- (node_1, reg_v1), (node2, reg_v1) -> fails
- (node_1, reg_v2), (node2, reg_v2) -> succeeds

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -265,6 +234,22 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                            .get_canister_http_committee(request.registry_version)
+                        {
+                            Ok(committee) => {
+                                let faults_tolerated = get_faults_tolerated(committee.len());
+                                (committee.len() - faults_tolerated, faults_tolerated)
+                            }
+                            Err(err) => {
+                                warn!(self.log, "Failed to get canister http committee: {:?}", err);
+                                continue;
+                            }
```
mraszyk: maybe this can be extracted to a separate function to avoid code duplication (the threshold is also computed further below)

mraszyk:
> Additionally, during membership changes, this allows the committee of the request to remain on the subnet until the request context disappears from the state. This is achieved by considering HTTP contexts in get_oldest_registry_version_in_use_by_repliated_state.

Doesn't the unassigned node stop the replica process meaning that the node can no longer share its artifacts anyway?

mraszyk (commented):
The PR looks good to me, but maybe somebody with more expertise around this code could double check.

rs/crypto/src/sign/basic_sig/tests.rs
eichhorl: After https://github.com/dfinity/ic/pull/10559/commits/92b1bbe051a77a44078941ca6f392484a0f9bdcb we check all four cases

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -284,9 +309,10 @@ impl CanisterHttpPoolManagerImpl {
...
-            if !context
-                .replication
-                .is_authorized_signer(&self.replica_config.node_id)
+            // Only make a request if this node belongs to the request's committee
+            if !self
+                .node_belongs_to_request_committee(context, self.replica_config.node_id)
+                .unwrap_or(false)
             {
                 continue;
             }
```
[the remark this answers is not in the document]
eichhorl: done: https://github.com/dfinity/ic/pull/10559/commits/92b1bbe051a77a44078941ca6f392484a0f9bdcb

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Done in https://github.com/dfinity/ic/pull/10559/changes/176f724120b4061c076bf9d36e767bcc91d1e360 and moved to `membership.rs` in https://github.com/dfinity/ic/pull/10559/changes/e3405b7a776393cee8f22e6ca13559fb4b032204

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -265,6 +234,22 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                            .membership
+                            .get_canister_http_committee(request.registry_version)
+                        {
+                            Ok(committee) => {
+                                let faults_tolerated = get_faults_tolerated(committee.len());
+                                (committee.len() - faults_tolerated, faults_tolerated)
+                            }
+                            Err(err) => {
+                                warn!(self.log, "Failed to get canister http committee: {:?}", err);
+                                continue;
```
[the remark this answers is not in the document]
eichhorl: This seems like premature optimization

eichhorl:
> > Additionally, during membership changes, this allows the committee of the request to remain on the subnet until the request context disappears from the state. This is achieved by considering HTTP contexts in get_oldest_registry_version_in_use_by_repliated_state.
> 
> Doesn't the unassigned node stop the replica process meaning that the node can no longer share its artifacts anyway?

Yes, "remaining on the subnet" here precisely means that a node will not turn into an unassigned node while there are still open HTTP requests that require contribution from that node.

This is done in this PR by letting the registry version of the context influence the "oldest registry version in use", which is stored in the CUP.

The orchestrator will not stop the replica process, as long as it is still part of the subnet in any registry version between the oldest one in use and the latest version, see here: https://github.com/dfinity/ic/blob/e3405b7a776393cee8f22e6ca13559fb4b032204/rs/orchestrator/src/upgrade.rs#L811-L849

mraszyk:
> The orchestrator will not stop the replica process, as long as it is still part of the subnet in any registry version between the oldest one in use and the latest version, see here:

Thanks a lot for explaining - this is what I was missing.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -247,6 +246,41 @@ impl CanisterHttpPoolManagerImpl {
...
+                    warn!(
+                        every_n_seconds => 10,
+                        self.log,
+                        "Unable to check HTTP committee membership at registry version {}, {:?}",
+                        context.registry_version,
+                        e
+                    );
+                }),
+            replication @ (Replication::NonReplicated(_) | Replication::Flexible { .. }) => {
+                Ok(replication.is_authorized_signer(&node_id))
```
pierugo-dfinity: `is_authorized_signer` always returning `true` for `FullyReplicated` might be misused. Since here is now its only usage (excluding tests), I think an alternative would be to inline it.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -631,26 +616,13 @@ impl CanisterHttpPoolManagerImpl {
...
-            // Make any requests that need to be made
-            self.make_new_requests(canister_http_pool);
-
-            // Create shares from any responses that are now available
-            change_set.extend(self.create_shares_from_responses(finalized_height));
-        }
+        // Make any requests that need to be made and create shares from responses
+        // that are now available.
+        self.make_new_requests(canister_http_pool);
+        change_set.extend(self.create_shares_from_responses());
```
pierugo-dfinity, line 621: We used to check the committee before generating those two change sets. Understood for `make_new_requests`, as it now checks inside its body. For `create_shares_from_response`, it looks like we do not check anymore. Do I understand properly that this is perfectly fine because we receive responses form the HTTP adapter for requests whose committee was precisely checked in `make_new_requests`? I.e., if `make_new_requests`'s committee check is correct, then so will `create_shares_from_response`'s because they are in a 1-1 relationship.

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity, line 602: Unrelated: Do you know why we generate change sets for all subcomponents instead of round-robin like most other protocols in Consensus?

rs/protobuf/src/gen/types/types.v1.rs
```diff
@@ -553,8 +553,6 @@ pub struct CanisterHttpResponseMetadata {
     pub id: u64,
     #[prost(bytes = "vec", tag = "3")]
     pub content_hash: ::prost::alloc::vec::Vec<u8>,
-    #[prost(uint64, tag = "4")]
-    pub registry_version: u64,
```
pierugo-dfinity, line 557: Is it fine to remove the protobuf fields without any rollout phase because they are not part of the replicated state? Slow replicas could still receive artifacts from replicas that already upgraded but I assume we would receive an alert in that case and it shouldn't do much more harm than that?

pierugo-dfinity (approved):
This is such a huge improvement that both simplifies the code and fixes the functionality. Very satisfying!

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Nice, this is much better: https://github.com/dfinity/ic/pull/10559/changes/eba3fb1b79eee87bf5caf4775b8fb15eaf3843a5

rs/consensus/utils/src/lib.rs
eichhorl: Good point, done: https://github.com/dfinity/ic/pull/10559/changes/8bda09f94c53633bc52c97a54c67d7e395f619ba

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Correct, we will only create shares for responses to requests that we previously made in `make_new_requests`

rs/protobuf/src/gen/types/types.v1.rs
eichhorl: Yes, HTTP outcalls are part of the batch payload which doesn't exist in the CUP. And we also don't store these types in the replicated state.

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Not sure, maybe because there is only one artifact type it seemed fine to do it all in one go

PR #10560, opened 2026-06-24
feat: publish to crates.io from a pinned commit

mraszyk:
This PR allows to publish crates to crates.io from a pinned commit. For security reasons, the commit must be an ancestor of HEAD to not bypass the branch whitelist for crates publishing.

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -43,27 +43,28 @@ jobs:
 
       - name: Verify commit and check out
         if: ${{ github.event.inputs.commit != '' }}
+        env:
+          COMMIT: ${{ github.event.inputs.commit }}
         run: |
           set -euo pipefail
-          commit="${{ github.event.inputs.commit }}"
-          if ! git merge-base --is-ancestor -- "${commit}" HEAD; then
-            echo "Commit '${commit}' is not an ancestor of HEAD — aborting for security"
+          if ! git merge-base --is-ancestor -- "${COMMIT}" HEAD; then
+            echo "Commit '${COMMIT}' is not an ancestor of HEAD — aborting for security"
             exit 1
           fi
-          git checkout --detach -- "${commit}"
+          git checkout --detach -- "${COMMIT}"
 
       - uses: rust-lang/crates-io-auth-action@bbd81622f20ce9e2dd9622e3218b975523e45bbe
         id: auth
 
       - name: Check crate version
+        env:
+          CRATE_NAME: ${{ github.event.inputs.crate-name }}
+          CRATE_VERSION: ${{ github.event.inputs.crate-version }}
         run: |
           set -euo pipefail
-          crate_name="${{ github.event.inputs.crate-name }}"
-          crate_version="${{ github.event.inputs.crate-version }}"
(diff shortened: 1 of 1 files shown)


basvandijk (approved):
LGTM. 

You could consider having `ref` instead of `commit` such that you can also run it on branches and tags (as long as they're ancestors of `master`).

Then you could also consider dropping the regex match and the `git rev-parse --verify` and just rely on `git merge-base --is-ancestor` to fail on non-sensible inputs.

mraszyk:
> You could consider having `ref` instead of `commit` such that you can also run it on branches and tags (as long as they're ancestors of `master`).
> 
> Then you could also consider dropping the regex match and the `git rev-parse --verify` and just rely on `git merge-base --is-ancestor` to fail on non-sensible inputs.

Done: could you please take one more look to ensure there's no security hole now, @basvandijk ?

.github/workflows/publish-crates.yml
```diff
@@ -34,6 +38,19 @@ jobs:
...
+      - name: Verify ref and check out
+        if: ${{ github.event.inputs.ref != '' }}
+        run: |
+          set -euo pipefail
+          ref="${{ github.event.inputs.ref }}"
+          if ! git merge-base --is-ancestor -- "${ref}" HEAD; then
+            echo "Ref '${ref}' is not an ancestor of HEAD — aborting for security"
+            exit 1
+          fi
+          git checkout --detach -- "${ref}"
```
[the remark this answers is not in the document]
mraszyk: https://github.com/dfinity/ic/pull/10560/commits/dca2abdc04c9e769b7243acf9f4706f15cd1e67a CC @basvandijk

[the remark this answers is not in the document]
basvandijk: Ok. Probably also safer to use a commit.

.github/workflows/publish-crates.yml
```diff
@@ -34,6 +38,19 @@ jobs:
...
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: ${{ github.event.inputs.commit != '' && '0' || '1' }}
+
+      - name: Verify commit and check out
+        if: ${{ github.event.inputs.commit != '' }}
+        run: |
+          set -euo pipefail
+          commit="${{ github.event.inputs.commit }}"
```
jwndlng: Use environment variables for workflow inputs to prevent script injection E.g.

```
- name: Verify commit and check out
  if: ${{ github.event.inputs.commit != '' }}
  env:
    COMMIT: ${{ github.event.inputs.commit }}
```

While the workflow requires write access to run, it's still good practice to do.

mraszyk: Could you please take a look: https://github.com/dfinity/ic/pull/10560/commits/7ba1deab07891d1a4ff8601ccdea35be54db485c

jwndlng: LGTM!

mraszyk: Thanks a lot!

PR #10561, opened 2026-06-24
chore: remove the ict system-test wrapper

basvandijk:
After https://github.com/dfinity/ic/pull/10494 the `ict` wrapper around `bazel` to launch system-tests and deploy testnets provides little value above `bazel` itself. IDX therefor doesn't want to maintain it anymore and suggests all users to switch to using `bazel` directly instead. The two primary use-cases are changed like:

* Running a system-test (note that the full bazel target has to be specified):
  ```diff
  -ict test tecdsa_performance_test_colocate 
  +bazel test //rs/tests/consensus/tecdsa:tecdsa_performance_test_colocate
  ```

* Deploying a testnet (note to use `bazel run`):
  ```diff
  -ict testnet create small
  +bazel run //rs/tests/testnets:small
  ```


rs/tests/networking/canister_http_soak_test.rs
```diff
@@ -3,7 +3,7 @@ Title:: Soak test for the http_requests feature
 
 Goal:: Measure the evolving qps of http_requests originating from one canister. The test should be run with the following command:
 ```
-ict testnet create //rs/tests/networking:canister_http_soak_test --output-dir=./canister_http_soak_test -- --test_tmpdir=./canister_http_soak_test
+bazel run //rs/tests/networking:canister_http_soak_test --test_tmpdir=./canister_http_soak_test
```
mraszyk: how about the output-dir that is now lost - is it not needed anymore or not supported?

basvandijk: Good question, it's indeed not supported anymore. Do you have a use case for it?

mraszyk: I don't know what it does so I don't know how useful it could be...

basvandijk: I frankly also never use it and forgot what it was for ;)

mraszyk (approved):

rs/tests/networking/canister_http_soak_test.rs
mraszyk: I think that's enough of an answer, approving now.

schneiderstefan: The [subnet load tester](https://github.com/dfinity/subnet-load-tester) currently makes use of the fact that the logs are there

michael-weigelt: I think it needs test_tmpdir. Does it also need this other one?

basvandijk: I didn't know about the subnet load tester. I'll take a look at how it works ...

basvandijk: @schneiderstefan , @michael-weigelt see https://github.com/dfinity/subnet-load-tester/pull/75.

nmattia (approved):
🥳

NikolaMilosa (approved):

r-birkner (approved):

PR #10565, opened 2026-06-25
chore(Core): Add metric for Wasm locals

michael-weigelt:


rs/embedders/src/wasm_utils/validation.rs
```diff
@@ -1325,9 +1327,19 @@ fn validate_function_section(
                 name: truncated_name,
             });
         }
+        // Check number of locals
+        let num_locals = module
+            .functions
+            .get(id)
+            .kind()
+            .unwrap_local() /* we are looping over locals only */
+            .unwrap() /* we retrieved the id from the same module */
```
mraszyk: Wouldn't it be safer to default to 0 here (and potentially log an error/raise an alert)?

michael-weigelt: I think this cannot fail barring an error in Wasm-tools. The double unwrap is a bit misleading, this is actually just part of the unwrap_local, which can only err if it's not local. I'll change this comment.

mraszyk (approved):

rs/embedders/src/wasm_utils/validation.rs
mraszyk: sure

PR #10571, opened 2026-06-25
feat: Implement new bazel rules for building SEV recovery GuestOS images

frankdavid:
Introduces the build infrastructure for SEV recovery GuestOS images. These images are recovery images whose measured boot components (kernel, initrd, OVMF, boot args) are taken from a previously released GuestOS version and whose launch measurements are verified against an NNS-signed `BlessAlternativeGuestOsVersion` proposal at build time. The signed proposal is embedded in the boot partition (and later verified at boot time).

## Build usage

```bash
bazel run //ic-os/guestos/envs/sev-recovery:build-sev-recovery -- \
    --base=<guestos git commit> --proposal=<proposal id>
```

The `bazel build` itself is hermetic (no network access). The `build-sev-recovery` runner performs the two steps that need network out-of-band and hands the results to the build as source files:

- **base image**: downloads the released GuestOS update image for `--base` via `curl`, caching it as `update-img-<version>.tar.zst` (skipping the download if already present) and pointing the fixed-name `base-update-img.tar.zst` symlink at it.
- **proposal**: downloads the certified `get_proposal_info` response for `--proposal` via the `alternative_guestos_tool` and writes it to `alternative_guestos_proposal.cbor`.

### New crates
- **`rs/ic_os/alternative_guestos`** — Extracts the existing proposal verification logic (`read_and_verify_signed_bless_alternative_guest_os_version_proposal`) out of `open_rootfs` into a reusable library. The `nns_public_key_override` parameter is now always present (previously conditionally compiled), with `None` passed in production builds so the hardcoded NNS public key is used.
- **`rs/ic_os/build_tools/alternative_guestos`** (`alternative_guestos_tool`) — A build tool with two subcommands:
  - `download-signed-proposal`: fetches a certified `get_proposal_info` response from the NNS governance canister via ic-agent and stores the CBOR certificate to disk, verifying it immediately.
  - `validate-measurements`: checks that locally generated launch measurements overlap with the measurements blessed in the proposal.

### SEV recovery build flow (`ic-os/guestos/envs/sev-recovery`)
- **`build-sev-recovery.sh`** (`bazel run` target `build-sev-recovery`) — orchestrates the downloads and then invokes the hermetic `bazel build`.

### Build wiring (`ic-os/defs.bzl`)
- Adds a `build_alternative_guestos_image` flag to `icos_build`. When enabled:
  - the bootfs tree is extracted from the supplied base GuestOS update image (`base-update-img.tar.zst`) via fuse2fs and the proposal is embedded into the boot partition;
  - the boot args are reused from the base release;
  - the generated launch measurements are validated against the proposal.
- Refactors boot partition file extraction (`extract_boot_partition_files`) to extract initrd, vmlinuz, OVMF, and boot args directly from the built boot partition image via `debugfs`, so launch measurements reflect the actual partition contents.


ic-os/defs.bzl
```diff
@@ -230,15 +273,19 @@ def icos_build(
 
         if image_deps.get("requires_root_signing", False):
             # Sign the root partition and substitute ROOT_HASH in boot args
+            signed_partition_root_hash = partition_root_hash
+            if build_alternative_guestos_image:
+                signed_partition_root_hash = partition_root_hash + ".local"
```
Bownairo: Why do we need this branch? It looks like when we take the args directly, we don't ever reference `partition_root_hash`.

ic-os/defs.bzl
```diff
@@ -251,20 +298,28 @@ def icos_build(
...
+                    name = boot_args,
+                    actual = ":" + extracted_boot_args,
+                    tags = ["manual"],
+                )
+            else:
+                native.genrule(
+                    name = "generate-" + boot_args,
+                    outs = [boot_args],
+                    srcs = [partition_root_hash, ":boot_args_template"],
+                    cmd = "sed -e s/ROOT_HASH/$$(cat $(location " + partition_root_hash + "))/ " +
```
Bownairo: nit: Should these be `signed_partition_root_hash`, even though they are set to be the same?

frankdavid: Removed signed_partition_root_hash

ic-os/defs.bzl
frankdavid: I think it remained from a previous approach, we can stick with `partition_root_hash`.

ic-os/defs.bzl
```diff
@@ -280,22 +331,26 @@ def icos_build(
...
-                    source $(execpath """ + boot_args + """)
+            # Build the genrule inputs and command conditionally.  When building an
+            # alternative GuestOS image (e.g. for recovery), the generated measurements
+            # are validated against the proposal *in the same genrule* so that the
+            # validation can never become a leaf target that is silently skipped.
+            measurement_srcs = [
+                ":extracted_OVMF_SEV.fd",
+                ":extracted_boot_args",
+                ":extracted_initrd.img",
+                ":extracted_vmlinuz",
```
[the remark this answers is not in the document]
frankdavid: Fixed.

ic-os/alternative_guestos.bzl
```diff
@@ -0,0 +1,96 @@
...
+cleanup() {{
+  set +e
+  if [[ "$$mounted" -eq 1 ]]; then
+    fusermount3 -u "$$tmpdir/bootfs" || fusermount -u "$$tmpdir/bootfs" || umount "$$tmpdir/bootfs"
+  fi
+  rm -rf "$$tmpdir"
+}}
+trap cleanup EXIT
+
+curl --fail --silent --show-error --location \
```
nmattia: we should definitely not be downloading via `curl` inside a build. Can you turn this into an `http_file` or similar?

ic-os/alternative_guestos.bzl
```diff
@@ -0,0 +1,96 @@
+def _download_alternative_guestos_proposal_impl(ctx):
+    output = ctx.actions.declare_file(ctx.label.name)
+
+    command = """
+set -euo pipefail
+
+proposal_id="$(cat {proposal_id_file})"
+
+if [[ -z "$proposal_id" ]]; then
+  echo "//{package}:{name} requires ALTERNATIVE_GUESTOS_PROPOSAL_ID to be set in the environment before invoking Bazel." >&2
```
nmattia: My instincts say this is probably not the right way to do it. In practice, when will this be built? And can the targets still be built on a clean checkout with e.g. `bazel build //...`? Otherwise it's probably best to run this as a `bazel run`

rs/ic_os/alternative_guestos/BUILD.bazel
nmattia, line 1: nit: lots of new code in this directory, would be good to have a README or a few top-level comments explaining what this is for. Also, what's the difference between `rs/ic_os/alternative_guestos` and `‎rs/ic_os/build_tools/alternative_guestos`?

ic-os/alternative_guestos.bzl
frankdavid: Restructured the flow into a script and a hermetic bazel build.

ic-os/alternative_guestos.bzl
frankdavid: PTAL

rs/ic_os/alternative_guestos/BUILD.bazel
frankdavid: WDYM? There is only one function in the directory. It's a shared library for verifying alternative guestos proposals. It's shared between the build tools and the production GuestOS. The reason: we want to run the same verification logic at build time and in production so that we can see early during the build if the proposal won't be accepted.

ic-os/defs.bzl
```diff
@@ -199,29 +220,79 @@ def icos_build(
             tags = ["manual", "no-cache"],
         )
 
-        ext4_image(
-            name = partition_boot_tzst,
-            src = ":rootfs-tree.tar",
-            file_contexts = ":file_contexts",
-            partition_size = image_deps["bootfs_size"],
-            subdir = "boot",
+        if build_alternative_guestos_image:
```
nmattia: in general it's best to avoid conditionals and keep the logic ultra simple and declarative. But there's already a fair bit of logic in this file, so might require some more structural changes if we want to get rid of the if/elses altogether...

ic-os/defs.bzl
```diff
@@ -199,29 +220,79 @@ def icos_build(
...
+                tmpdir="$$(mktemp -d)"
+                trap 'rm -rf "$$tmpdir"' EXIT
+
+                tar --extract -a --file "$<" --directory "$$tmpdir"
+
+                # initrd and vmlinuz are symlinks, first extract the symlink targets
+                initrd_target="$$(/usr/sbin/debugfs -R "stat /initrd.img" "$$tmpdir/partition.img" | sed -n 's/^Fast link dest: "\\(.*\\)"$$/\\1/p')"
+                vmlinuz_target="$$(/usr/sbin/debugfs -R "stat /vmlinuz" "$$tmpdir/partition.img" | sed -n 's/^Fast link dest: "\\(.*\\)"$$/\\1/p')"
+
+                /usr/sbin/debugfs -R "dump /boot_args $(@D)/""" + extracted_boot_args + """" "$$tmpdir/partition.img"
```
nmattia: is `debugfs` always available on Linux? or does it need to be somehow installed?

ic-os/defs.bzl
```diff
@@ -251,20 +322,28 @@ def icos_build(
...
-            native.genrule(
-                name = "generate-" + boot_args,
-                outs = [boot_args],
-                srcs = [partition_root_hash, ":boot_args_template"],
-                cmd = "sed -e s/ROOT_HASH/$$(cat $(location " + partition_root_hash + "))/ " +
-                      "< $(location :boot_args_template) > $@",
-                tags = ["manual"],
-            )
+
+            if build_alternative_guestos_image:
```
nmattia: same thoughts as above re. the conditionals; maybe it's best to declare both targets and later on select the one you need

rs/ic_os/alternative_guestos/BUILD.bazel
nmattia: I understand; a README.md in `rs/ic_os/alternative_guestos` could say exactly that

ic-os/defs.bzl
```diff
@@ -133,21 +135,40 @@ def icos_build(
...
-        wildcards = True,
-        tags = ["manual"],
-    )
+    # Extract the bootfs tree from the base GuestOS update image. The image is
+    # supplied by build-sev-recovery as a source file `base-update-img.tar.zst`,
+    # a symlink to a version-named `update-img-<version>.tar.zst`
+    # (see //ic-os/guestos/envs/sev-recovery:build-sev-recovery).
+    native.genrule(
+        name = "alternative_guestos_base_bootfs_tree_tar",
+        srcs = [":base-update-img.tar.zst"],
```
nmattia, line 144: in most builds (unless the build-sev-recovery.sh was run) this will be a broken symlink, correct? but since `build_alternative_guestos_image` is `False`, and because this target is tagged with `"manual"`, it will never be built, correct? Just making sure I understand.

nmattia (approved):

ic-os/defs.bzl
frankdavid: We already have quite a bit of logical targets where depending on the IC-OS version, we build a target differently. I'm happy to discuss if you have ideas for how to make this easier to understand.

ic-os/defs.bzl
frankdavid: It's actually available via e2fsprogs which we already build from scratch, we can reuse that.

ic-os/defs.bzl
frankdavid: I made a few smaller improvements - moved the rules out of the condition and added a ternary selection below.

ic-os/defs.bzl
frankdavid: Yep, that's right. The symlink is only created by build-sev-recovery.sh

Bownairo (approved):

PR #10582, opened 2026-06-26
chore: Extract Bazel Test Bare Metal job from schedule‑daily

frankdavid:
We can limit the secrets the workflow needs access to, so it's less risky if we ever want to run the workflow on non-submitted branches.


blind-oracle (approved):

.github/workflows/schedule-daily-bare-metal.yml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  bazel-test-bare-metal:
+    name: Bazel Test Bare Metal
+    runs-on: &dind-large-setup
+      labels: dind-large
+    container: &container-setup
+      image: ghcr.io/dfinity/ic-build@sha256:0ed93dc568f53cce56042a86b31f98cd1bbb9cd3ea8e649532c366494b9c160a
+      options: >-
+        -e NODE_NAME --privileged --cgroupns host --mount type=tmpfs,target="/tmp/containers"
+    environment: Nightly Tests
```
basvandijk:
```suggestion
    environment: Bare Metal Tests
```

basvandijk (approved):
Note I added a dedicated `Bare Metal Tests` environment with the secrets:

* `ZH2_DLL01_CSV_SECRETS `
* `ZH2_DLL01_INI_SECRET`
* `ZH2_FILE_SHARE_KEY`

That environment is available on the branches:

* `baremetal-*`
* `master`
* `rc--*`

Let's use that instead of the `Nightly Tests` environment.

Once this PR is merged I'll remove those secrets from the  `Nightly Tests` environment.

PR #10590, opened 2026-06-26
perf(system-tests): cache image extraction in the local backend

basvandijk:
## What

The local (libvirt + QEMU) system-test backend (`SYSTEM_TEST_INFRA=local`) re-extracted the **same** primary GuestOS image once per node when starting a testnet's VMs. Decompressing a multi-gibibyte `*.tar.zst` / `*.img.zst` N times (once per node) is wasteful.

## How

`start_vm` now routes the primary-image extraction through a new `extract_image_cached`:

- Each `src` is extracted **once** into a shared, content-addressed base under the backend's working dir: `image_cache/<key>.img` (keyed by a hash of the source path, which is stable within a bazel invocation).
- Concurrent callers — other setup threads and forked task subprocesses running `vm.start()` — serialize on a per-key **`flock`** (blocking, exclusive). The first holder extracts (to a temp file, then atomically `rename`s it into place); the rest wake to find the base present and skip straight to the copy.
- Each VM disk is then materialized with `cp --reflink=auto` (copy-on-write where the filesystem supports it; a full copy otherwise).

This mirrors Farm's base-image trick (`backend/Farm/VMManager.hs`).

Per-VM-unique config disks (`attach_disk_images`) are deliberately left on the direct extraction path — caching them would dedupe nothing.

## Validation

`bazel test //rs/tests/idx:basic_health_test_local` — from the test-driver log, the shared GuestOS image is extracted once and copied per node:

| metric | value |
| --- | --- |
| \`tar.zst\` extractions | **1** |
| \`Materializing\` (per-VM copies) | **9** (8 nodes + 1 UVM) |
| GuestOS base materialized | **8×** from one \`image_cache/<key>.img\` |

\`rustfmt\` + \`clippy\` clean.

### DIFF
--- a/rs/tests/driver/src/driver/local_backend.rs
+++ b/rs/tests/driver/src/driver/local_backend.rs
@@ -885,6 +885,103 @@ impl LocalBackend {
         Ok(())
     }
 
+    /// Extract `src` into a shared base location exactly once, then materialize
+    /// `dst` as a copy-on-write (where supported) copy of that base.
+    ///
+    /// [`start_vm`](Self::start_vm) extracts the *same* primary GuestOS image
+    /// for every IC node in a testnet. Decompressing a multi-gibibyte
+    /// `*.tar.zst` / `*.img.zst` once per node is wasteful, so the first caller
+    /// for a given `src` extracts it to `image_cache/<key>.img` under
+    /// `working_dir`, while concurrent callers (other setup threads, or a forked
+    /// task subprocess running `vm.start()`) block on a per-key file lock,
+    /// observe the finished base, and skip straight to the copy. This mirrors
+    /// Farm's `cp --reflink=auto` base-image trick.
+    ///
+    /// The cache lives under `working_dir` so the base shares a filesystem with
+    /// the per-VM disks (making the publish `rename` atomic and letting
+    /// `cp --reflink` share blocks) and is torn down with the group.
+    ///
+    /// The base is keyed by a hash of the `src` *path*, not its contents: within
+    /// a single bazel invocation a given image always resolves to the same
+    /// immutable runfiles path, so the path identifies the content without
+    /// reading the (large) file to hash it. Distinct paths to identical content
+    /// merely cache twice (correct, marginally less optimal).
+    fn extract_image_cached(&self, src: &Path, dst: &Path) -> Result<()> {
+        use ic_crypto_sha2::Sha256;
+        use std::os::unix::io::AsRawFd;
+
+        let cache_dir = self.active_local_backend.working_dir.join("image_cache");
+        std::fs::create_dir_all(&cache_dir)
+            .with_context(|| format!("creating image cache dir {}", cache_dir.display()))?;
+
+        let key = hex::encode(&Sha256::hash(src.to_string_lossy().as_bytes())[0..16]);
+        let base = cache_dir.join(format!("{key}.img"));
+
+        // Serialize extraction of a given `src` across threads *and* processes
+        // with a blocking, exclusive advisory lock on a per-key lock file. Only
+        // the first holder extracts; others wake to find `base` already present.
+        // The lock is released (dropped) before the copy below: a finished base
+        // is immutable, so any number of callers may copy from it concurrently.
+        {
+            let lock_path = cache_dir.join(format!("{key}.lock"));
+            // `append` (rather than `write`) gives the open a defined,
+            // non-truncating behavior; the file is only a lock holder, never
+            // written to.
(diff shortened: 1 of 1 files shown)


basvandijk:
## Follow-up: rework to qcow2 backing-file overlays (filesystem-independent CoW)

**Why.** This PR materializes each VM's primary disk with `cp --reflink=auto base dst`. `--reflink` only avoids a full copy on CoW filesystems (btrfs/XFS); on **ext4** (the dev container, and likely CI) it silently falls back to a full copy, so the cache doesn't actually speed things up. Measured on the GuestOS image (556M zst → 1.6G sparse), 8 nodes, 32 cores:

| filesystem | BEFORE (8× decompress) | this PR (1× decompress + 8× `cp`) |
|---|---|---|
| local disk (ext4, `/dev/vdd`) | 11.5s | 13–17.6s |
| CEPH-backed (ext4, `/dev/vdb` = `/ic`) | 125–135s | 116–149s |

i.e. neutral-to-slightly-slower where there's no reflink. (Also: `/ic` is CEPH-backed and ~10× slower than the local disk, so benchmark/test with `--test_tmpdir` on a local-disk path such as `~/.cache/...`, not under `/ic`.)

**Idea.** Give each VM a thin **qcow2 overlay** backed by the shared raw base instead of copying it. Overlay creation is near-instant and filesystem-independent (CoW lives in the qcow2 layer):

```
qemu-img create -f qcow2 -F raw -b <image_cache/<key>.img> <vm>/primary.qcow2 [<size>]
```

Expected disk-prep for 8 nodes ≈ one base decompress + instant overlays ≈ **~2–3s local / ~15–30s CEPH** (vs 11.5s / 125s today), with lower disk usage (base + thin overlays vs N full copies).

### Plan
1. **Base-only extraction.** Refactor `extract_image_cached` → `ensure_base_image(src) -> PathBuf`: under the existing per-key `flock`, extract once to `image_cache/<key>.img` (raw, `0o444`), atomic rename, return the base path. Drop the per-VM `cp`.
2. **Overlay in `start_vm`.** `primary_disk` → `vm_dir/primary.qcow2`. After `ensure_base_image`, run `qemu-img create -f qcow2 -F raw -b <abs base> primary.qcow2`. Honour `min_boot_image_size_gib` as the overlay's *virtual* size (create / `qemu-img resize` to `max(base_virtual_size, min_gib)` — never shrink). chmod overlay `0o600`; base stays `0o444`. Keep the `if !primary_disk.exists()` guard so `vm().kill()` + `vm().start()` reuses the overlay (persisted writes).
3. **Per-disk format in the domain XML.** Add a `driver_type` field to `DiskEntry` (`"qcow2"` for the primary, `"raw"` for extras). In `rs/tests/driver/templates/guestos_vm_template.xml`, change `<driver name='qemu' type='raw' .../>` → `type='{{ d.driver_type }}'`. The overlay header records the backing file, so no explicit `<backingStore>` is needed.
4. **Alignment.** `pad_to_request_alignment` no longer applies to the primary: the overlay is qcow2 (qemu manages alignment) and the base is opened **read-only** as a backing file (no `resize` permission → avoids the "image size is not a multiple of request alignment" error). Extra/config disks keep the raw + pad path.
5. **`qemu-img` is a system binary** (`/usr/bin/qemu-img`, present), used like the existing `cp`/`truncate`/`tar`/`qemu-system-x86_64`; confirm it's reachable inside the bazel linux-sandbox.
6. **Teardown unchanged** — overlays in vm dirs, base in `image_cache`, both under `working_dir`.

### Validate
- `bazel test //rs/tests/idx:basic_health_test_local` with KVM and `--test_tmpdir` on a local-disk path — expect pass + much shorter disk-prep.
- A test doing `vm().kill()` + `vm().start()` to confirm overlay persistence.
- Re-run the micro-benchmark with `qemu-img create` to quantify local + CEPH.

### Risks / notes
- Backing-file path must be absolute & stable across the setup process and forked task subprocesses (it's under the canonicalized `working_dir` → OK).
- Validate qcow2 + `cache='none'` (O_DIRECT) + `discard='unmap'` together.
- Base must stay immutable for the overlays' lifetime (read-only `0o444` in `image_cache`).

basvandijk:
Alternative implementation using **qcow2 backing-file overlays** instead of `cp --reflink`: #10591. It's faster than master on both local-disk and CEPH-backed ext4 (whereas this PR's `cp --reflink` regresses on ext4 since it has no CoW). See the benchmark table in #10591.

basvandijk:
Closing in favour of https://github.com/dfinity/ic/pull/10591.

PR #10594, opened 2026-06-29
perf: avoid deep cloning full network topology on each message execution

hpeebles:
Every wasm execution builds a SandboxSafeSystemState, which until now deep cloned the entire NetworkTopology ([here](https://github.com/dfinity/ic/blob/909803a5c8b33acdfda9063df1ca4123acac39f6/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs#L857)). Since this happens once per message, a round executing M messages performed M full topology clones.

This PR makes SandboxSafeSystemState hold an Arc<NetworkTopology> instead of an owned value, so the per message construction becomes a reference count bump rather than a deep clone.

The full topology is still deep cloned for each install and each query (this can potentially be improved as a follow on but it is a significantly larger change).

I benchmarked NetworkTopology::clone() (the old cost) against Arc::clone (the new cost) on a reconstructed mainnet topology (49 subnets / 673 nodes):

Topology | deep clone (old) | Arc::clone (new) | speedup
-- | -- | -- | --
mainnet (49 subnets / 673 nodes) | ~5.9 µs | ~3.6 ns | ~1,650×
2× mainnet (98 subnets / 1,346 nodes) | ~12.1 µs | ~3.6 ns | ~3,350×

The deep clone cost scales linearly with topology size while Arc::clone stays flat at ~3.6 ns, so the saving grows as the network adds subnets/nodes.


rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -657,7 +658,17 @@ pub struct SandboxSafeSystemState {
...
     pub is_wasm64_execution: bool,
-    network_topology: NetworkTopology,
+    // The `Arc` lets the (round-invariant) topology be shared cheaply across the
+    // per-message `SandboxSafeSystemState`s instead of being deep-cloned each
+    // time. It is safe to serialize via `serde_arc`: `NetworkTopology` is
+    // read-only here (no interior mutability to drift across copies) and there is
+    // a single reference to it per `SandboxSafeSystemState`, so the identity/
+    // duplication caveats of serializing `Arc`s don't apply. The inner topology
+    // crosses the sandbox IPC boundary by value regardless, exactly as it would
+    // if this were a plain owned field.
```
alin-at-dfinity: Nit: Please reduce this to the (gist of the) first sentence. And turn it into a doc comment. The extensive description only makes sense in the context of this change.

hpeebles: Shall I simply remove the comment? It isn't really adding much.
I didn't plan on adding it to begin with, but within `serde_arc.rs` it says to provide justification each time you use the `(de)serialize_arc` funcions. But it seems like its just adding noise.

alin-at-dfinity: That would work too.

hpeebles: Done

alin-at-dfinity (approved):

PR #10601, opened 2026-06-29
feat: [DSM-144] Reuse stream message subtrees in `HashTree`

alin-at-dfinity:
Extend subtree reuse to stream messages: each message's certified subtree (a single leaf) is stored as a reusable stub keyed by its backing `Arc<Request|Response|Refund>`. `StreamQueueFork` gains an `mk_source` function; per-variant `expand_*` expanders (and `message_source`) rebuild a stub. Witness generation expands a stubbed *leaf* (carrying its value, not a pruned digest), and `SubtreeSource::downcast_arc` recovers the owned `Arc` that the message expanders move into the rebuilt subtree.

### DIFF
--- a/rs/canonical_state/src/lazy_tree_conversion.rs
+++ b/rs/canonical_state/src/lazy_tree_conversion.rs
@@ -32,7 +32,7 @@ use ic_replicated_state::{
 use ic_types::{
     CanisterId, Height, NodeId, PrincipalId, SubnetId,
     ingress::{IngressState, IngressStatus, WasmResult},
-    messages::{EXPECTED_MESSAGE_ID_LENGTH, MessageId},
+    messages::{EXPECTED_MESSAGE_ID_LENGTH, MessageId, Refund, Request, Response, StreamMessage},
     xnet::{StreamHeader, StreamIndex, StreamIndexedQueue},
 };
 use std::convert::{AsRef, TryFrom, TryInto};
@@ -293,6 +293,10 @@ where
     queue: &'a StreamIndexedQueue<T>,
     certification_version: CertificationVersion,
     mk_tree: fn(StreamIndex, &'a T, CertificationVersion) -> LazyTree<'a>,
+    /// Produces a [`SubtreeSource`] for a child, to be used to create a reusable
+    /// [stub](`NodeKind::Stub`). Stream messages' contents are immutable once
+    /// enqueued, so the digest of any index present in a baseline can be reused.
+    mk_source: fn(&'a T, CertificationVersion) -> SubtreeSource,
 }
 
 impl<'a, T: Send + Sync> LazyFork<'a> for StreamQueueFork<'a, T> {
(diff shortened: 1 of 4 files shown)


rs/canonical_state/src/lazy_tree_conversion.rs
```diff
@@ -560,6 +573,73 @@ fn streams_as_tree<'a>(
     }
 }
 
+/// Expands `$ty`/`$variant` to a 1-leaf [`HashTree`] (re-encoding the message
+/// under the const certification version `V`), and a `select_*` helper that
+/// picks the `V`-specific monomorphization — so the stored function pointer
+/// alone fully determines the expansion. See [`message_source`].
+macro_rules! message_expander {
```
mraszyk, line 580: Wouldn't it make sense to store `StreamMessage` under an Arc so that we don't need to pattern-match on `StreamMessage` and deal with all the complexity?

rs/canonical_state/tree_hash/tests/subtree.rs
```diff
@@ -408,6 +408,177 @@ fn disjoint_stub_sources(a: &HashTree, b: &HashTree) -> bool {
     a.stub_count() == b.stub_count() && a.stub_sources().zip(b.stub_sources()).all(|(x, y)| x != y)
 }
 
+// --- Leaf stubs (mirroring stream messages) -------------------------------
```
mraszyk, line 411: Why don't we use the real code here?

rs/canonical_state/tree_hash/tests/subtree.rs
```diff
@@ -408,6 +408,177 @@ fn disjoint_stub_sources(a: &HashTree, b: &HashTree) -> bool {
...
+fn messages(n: usize) -> Messages {
+    (0..n)
+        .map(|i| (message_id_label(i), Arc::new(message_bytes(i))))
+        .collect()
+}
+
+/// Rebuilds a stubbed message leaf from its source `Arc` (mirrors the per-variant
+/// `expand_*` message expanders in production).
+fn expand_test_message(source: &SubtreeSource) -> Result<HashTree, HashTreeError> {
+    let bytes = source.downcast::<Vec<u8>>();
```
mraszyk: that would be downcast_arc in production

rs/canonical_state/src/lazy_tree_conversion.rs
alin-at-dfinity: You mean have `Stream` hold `Arc<StreamMessage>` instead of directly `StreamMessage`?

`StreamMessage` is already an enum wrapping one of 3 kinds of `Arcs`, so it would be kind of redundant and inefficient to store a pointer to a pointer.

I.e. neither the current approach, nor wrapping `StreamMessage` within an `Arc` seems ideal to me.

rs/canonical_state/tree_hash/tests/subtree.rs
alin-at-dfinity: [It's AI-generated code, but] `ic_canonical_state` (which defines the various `LazyTree` implementations) depends on `ic_canonical_state_tree_hash`. So it would be something of a circular dependency (not really, it would work in practice, because only the `ic_canonical_state_tree_hash` system tests would depend on `ic_canonical_state`).

mraszyk: Sure, let's then fix the downcast to downcast_arc (see below) and merge this.

rs/canonical_state/tree_hash/tests/subtree.rs
alin-at-dfinity: Done, although I'm not sure it's relevant here (beyond ensuring coverage for `downcast_arc()`).

mraszyk: > beyond ensuring coverage for downcast_arc()

that was my motivation for pointing this out

mraszyk (approved):

PR #10603, opened 2026-06-29
feat(registry): grant elevated rate limits to allowlisted node providers

bjoernek:
## Summary

Adds a hardcoded, temporary allowlist of trusted node providers that receive **elevated (10x) rate limits** for node add/remove operations in the registry canister. Initially the allowlist contains only the DFINITY Foundation (`bvcsg-3od6r-jnydw-eysln-aql7w-td5zn-ay5m6-sibd2-jzojt-anwag-mqe`). This is required in the context of the cloud engine initiative which requires the onboarding of many new nodes (note that these do not receive node provider rewards, but are rather paid by the cloud engine creator). 

## Motivation

Node providers are rate limited when adding/removing nodes (standard: avg 20/day, burst 140, per node operator and per node provider). This blocks legitimate bulk onboarding, e.g. on-demand provisioning of nodes in the cloud. Creating an explicit allowlist of node IDs is impractical because a node ID is derived from the node signing key, which only exists once the node has started. Granting a higher limit to specific, trusted node providers is a simpler, controlled alternative.

## What changes

- `rate_limits.rs`: a hardcoded `ELEVATED_NODE_PROVIDERS` allowlist plus a second pair of node-operator / node-provider rate limiters configured at 10x the standard limits (avg 200/day, burst 1400). Operations by node operators belonging to an allowlisted provider are routed to the elevated limiters at both reserve and commit time.
- `storage.rs`: two new stable-memory regions (IDs 4 and 5) for the elevated limiters, so their capacity persists across upgrades.
- Unit test verifying an allowlisted provider can reserve beyond the standard spike while a standard provider cannot.

## Scope / safety

- All non-allowlisted node providers remain subject to the unchanged standard limits.
- The per-IP `add_node` rate limit (1/day per IP) continues to apply to everyone, including allowlisted providers, as a residual safeguard.
- This is intended as a temporary measure; the allowlist and elevated limiters are marked with `TODO` for removal once no longer needed.


rs/registry/canister/src/rate_limits.rs
```diff
@@ -30,6 +31,42 @@ const AVG_ADD_NODE_BY_IP_PER_DAY: u64 = 1;
...
+const ELEVATED_NODE_PROVIDERS: &[&str] = &[
+    // DFINITY Foundation
+    "bvcsg-3od6r-jnydw-eysln-aql7w-td5zn-ay5m6-sibd2-jzojt-anwag-mqe",
+];
+
+fn is_elevated_node_provider(node_provider_id: PrincipalId) -> bool {
+    let node_provider_id = node_provider_id.to_string();
+    ELEVATED_NODE_PROVIDERS
+        .iter()
+        .any(|allowlisted| *allowlisted == node_provider_id)
```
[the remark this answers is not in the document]
bjoernek: Addressed in eaacb0e5ed: the allowlist is now parsed once into `Vec<PrincipalId>` in a thread_local and compared directly (no per-call `String` allocation), and a malformed entry now fails fast (panics at first use) instead of silently never matching. A new test `test_elevated_node_providers_allowlist_parses` forces the parse so CI catches a bad entry.

rs/registry/canister/src/rate_limits.rs
```diff
@@ -320,4 +443,55 @@ mod tests {
...
         assert_eq!(initial_provider_capacity, final_provider_capacity);
     }
+
+    #[test]
+    fn test_allowlisted_node_provider_gets_elevated_rate_limit() {
+        use std::str::FromStr;
+
+        let now = SystemTime::now();
+        // The first entry of the elevated allowlist (DFINITY).
+        let elevated_provider = PrincipalId::from_str(ELEVATED_NODE_PROVIDERS[0]).unwrap();
```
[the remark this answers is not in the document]
bjoernek: Addressed in eaacb0e5ed: the test now references the DFINITY principal explicitly (not by allowlist index) and asserts `is_elevated_node_provider(elevated_provider)` membership, so it is independent of allowlist ordering.

rs/registry/canister/src/rate_limits.rs
```diff
@@ -1,6 +1,7 @@
 use crate::mutations::node_management::common::get_node_provider_id_for_operator_id;
 use crate::registry::Registry;
 use crate::storage::{
+    get_elevated_node_operator_rate_limiter_memory, get_elevated_node_provider_rate_limiter_memory,
```
daniel-wong-dfinity-org-twin, line 4: How about instead of "elevated" (which does not explain why), let's call it "trusted"?

rs/registry/canister/src/rate_limits.rs
```diff
@@ -30,6 +32,54 @@ const AVG_ADD_NODE_BY_IP_PER_DAY: u64 = 1;
...
+/// Node providers granted elevated (higher) rate limits. This is a hardcoded,
+/// temporary allowlist of trusted node providers; everyone else is subject to
+/// the standard limits. All node operator operations (add, remove, and the
+/// direct node config updates) performed by node operators belonging to one of
+/// these node providers are routed to the elevated limiters, mirroring the
+/// scope of the standard node operator limiter.
+/// TODO: Remove entries (and the elevated limiters) once no longer needed.
+const ELEVATED_NODE_PROVIDERS: &[&str] = &[
+    // DFINITY Foundation
+    "bvcsg-3od6r-jnydw-eysln-aql7w-td5zn-ay5m6-sibd2-jzojt-anwag-mqe",
```
daniel-wong-dfinity-org-twin, line 69: FWIW, I concur that this matches what's on Dashboard.

rs/registry/canister/src/rate_limits.rs
```diff
@@ -320,4 +456,81 @@ mod tests {
...
+
+        // The committed capacity is actually consumed from the elevated
+        // limiters: a follow-up reservation for the full elevated spike must now
+        // fail because `over_standard_spike` was already used.
+        let result = registry.try_reserve_capacity_for_node_operator_operation(
+            now,
+            elevated_operator,
+            ELEVATED_NODE_OPERATOR_MAX_SPIKE,
+        );
+        assert_eq!(result.err(), Some(RateLimiterError::NotEnoughCapacity));
```
daniel-wong-dfinity-org-twin:
```suggestion
        assert_eq!(result, Err(Some(RateLimiterError::NotEnoughCapacity)));
```

rs/registry/canister/src/rate_limits.rs
```diff
@@ -320,4 +456,81 @@ mod tests {
...
+
+        // A standard provider's operator is still bound by the standard spike.
+        let result = registry.try_reserve_capacity_for_node_operator_operation(
+            now,
+            standard_operator,
+            over_standard_spike,
+        );
+        assert_eq!(result.err(), Some(RateLimiterError::NotEnoughCapacity));
```
daniel-wong-dfinity-org-twin: Ditto


```suggestion
        assert_eq!(result, Err(Some(RateLimiterError::NotEnoughCapacity)));
```

rs/registry/canister/src/rate_limits.rs
bjoernek: I prefer elevated because also other parties are trusted, but we just need elevated levels for DFINITY in the context of cloud engines.

daniel-wong-dfinity-org-twin: Do as you see fit.

Since this is temporary, I guess it does not super matter, but we can maybe briefly comment more WHY elevated. I think what we can say is something approximately like

> To prepare for the launch of engines, we need a large-ish initial pool of nodes from which to form engines. Whereas, the standard rate limits would get in the way of that, because they are only designed to allow steady state flow of mutations, not these special one-time mega bursts. This loop hole is safe, because we trust these node providers will not abuse this special privilege.

This could be added at around line 35, or so.

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/src/rate_limits.rs
bjoernek: Kept the name `elevated` and added the rationale you suggested in d3eec7834b (near the elevated-limits constants): explains that the one-time bulk onboarding of nodes to form cloud engines is a burst the steady-state standard limits would otherwise block, and that this is safe because the allowlisted providers are trusted not to abuse it.

bjoernek:
Heads up on the test assertions: the two applied suggestions (`assert_eq!(result, Err(Some(...)))`) broke the build and turned CI red. `result` is `Result<RateLimitReservation, RateLimiterError>`, and asserting on the whole result needs `RateLimitReservation: PartialEq + Debug` — but it can't be `PartialEq` (it holds a lib `Reservation` that wraps a `Weak` pointer), so it fails to compile (E0369/E0277).

I fixed this in two steps:
- `d3eec78`: reverted to the `assert_eq!(result.err(), Some(RateLimiterError::NotEnoughCapacity))` form to get CI green again.
- `946509aef6`: to properly address the reviewer's intent (assert on the error directly, not via `.err()`), derived `Debug` on `RateLimitReservation` and switched both assertions to `assert_eq!(result.unwrap_err(), RateLimiterError::NotEnoughCapacity)`.

So the final form is `unwrap_err()`, not `.err()`.

rs/registry/canister/src/rate_limits.rs
bjoernek: Followed up in 946509aef6 to honor the intent here. The literal suggestion could not compile: asserting on the whole `result` needs `RateLimitReservation: PartialEq + Debug`, but it can't be `PartialEq` (it holds the lib `Reservation<String>`, which wraps a `Weak` pointer and isn't `PartialEq`). Instead I derived `Debug` on `RateLimitReservation` and switched both assertions to `assert_eq!(result.unwrap_err(), RateLimiterError::NotEnoughCapacity)` — asserts on the error directly (no `.err()`/`Some` wrapping) and keeps `assert_eq!` with a clean failure message.

daniel-wong-dfinity-org-twin (approved):

PR #10604, opened 2026-06-29
chore: Remove attached LUKS header from Store partition

frankdavid:
The Store partition is now formatted with a detached LUKS header only. A legacy attached header still present on a device (which was the previous approach) is wiped on the next open, so only the detached header remains going forward.

In addition, `guest_disk` now uses a shared metrics `Registry` and writes all metrics once. The PR adds the `guest_disk_store_attached_luks2_header_status` gauge to track the header migration state.

### DIFF
--- a/rs/ic_os/guest_upgrade/client/src/lib.rs
+++ b/rs/ic_os/guest_upgrade/client/src/lib.rs
@@ -197,13 +197,12 @@ impl DiskEncryptionKeyExchangeClientAgent {
             .key
             .context("GetKeyResponse does not contain a key")?;
 
-        self.persist_disk_encryption_artifacts(
-            disk_encryption_key,
-            // This can be None when upgrading from older GuestOS-s that do not populate this field.
-            // TODO: Error on None once all GuestOS-s support detached headers
-            disk_encryption_data.luks_header,
-        )
-        .await?;
+        let luks_header = disk_encryption_data
+            .luks_header
+            .context("Server did not send a Store LUKS header")?;
+
+        self.persist_disk_encryption_artifacts(disk_encryption_key, luks_header)
+            .await?;
 
         Ok(())
     }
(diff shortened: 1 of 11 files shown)


rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -197,46 +197,32 @@ impl DiskEncryptionKeyExchangeClientAgent {
...
-                self.crypto_ops
-                    .backup_luks_header(&self.store_device_path, &self.store_luks_header_path)
-                    .context("Local LUKS header backup failed")?;
-            }
-        }
+        tokio::fs::write(&self.store_luks_header_path, &luks_header)
+            .await
+            .with_context(|| {
+                format!(
+                    "Failed to write store LUKS header to {}",
```
blind-oracle: Maybe uppercase Store or quote it, otherwise looks like a typo

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -18,39 +21,64 @@ pub struct SevDiskEncryption {
...
-                device_path.display(),
-                self.store_luks_header_path.display()
+                "Failed to check for attached LUKS2 header on {}: {e:#}. \
+                Continuing without wiping.",
+                device_path.display()
+            );
+            false
+        });
+
+        if self.store_luks_header_path.exists() && attached_header_present {
```
blind-oracle, line 61: Should we actively check that the existing header is valid? Or this is already done before somewhere?

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -196,35 +226,34 @@ impl DiskEncryption for SevDiskEncryption {
...
-        // backup. Once all GuestOS-s support detached headers, we can switch to using only detached
-        // headers.
-        // TODO: Remove attached header usage for store partition
+
+        let header_location = match partition {
+            Partition::Store => {
+                if self.store_luks_header_path.exists() {
+                    bail!(
+                        "Refusing to format Store because detached LUKS header {} already exists. \
+                        Remove the stale header first if you really want to reformat the device.",
```
blind-oracle, line 254: Hmm is this error even actionable? How can a user remove it?

blind-oracle (approved):

rs/ic_os/os_tools/guest_disk/src/sev.rs
frankdavid: No, but 1) we'd find it in testing 2) if it fails the boot, we could run a recovery GuestOS. Being asked to format the Store partition when a detached header already exists just doesn't make sense, so I rather don't touch the disk.

rs/ic_os/os_tools/guest_disk/src/sev.rs
frankdavid: Added a check + test case

PR #10609, opened 2026-06-29
docs: clarify SNS ledger fee parameter updates

bjoernek:
## Summary

Clarifies the SNS Governance protobuf/API comments around ledger fee updates:

- `ManageLedgerParameters.transfer_fee` is the intended way to change the SNS ledger's actual transfer fee.
- Successful `ManageLedgerParameters` execution also syncs `NervousSystemParameters.transaction_fee_e8s` in Governance.
- Directly changing `NervousSystemParameters.transaction_fee_e8s` does not update the ledger.

## Why

DAO operators can otherwise infer that changing `transaction_fee_e8s` through `ManageNervousSystemParameters` changes the ledger fee. That can leave Governance and the SNS ledger fee out of sync, causing ordinary ledger transfers initiated by Governance to use a stale fee.

## Validation

- `git diff --check`
- `rustfmt --check rs/sns/governance/src/gen/ic_sns_governance.pb.v1.rs`

No runtime tests were run because this is a comment-only documentation change.


rs/sns/governance/proto/ic_sns_governance/pb/v1/governance.proto
```diff
@@ -390,9 +390,13 @@ message TransferSnsTreasuryFunds {
   optional Subaccount to_subaccount = 5;
 }
 
-// A proposal function that changes the ledger's parameters.
+// A proposal function that changes the SNS ledger canister's parameters.
 // Fields with None values will remain unchanged.
 message ManageLedgerParameters {
+  // Changes the SNS ledger's actual transfer fee. If this proposal executes
+  // successfully, Governance also syncs NervousSystemParameters.transaction_fee_e8s
```
daniel-wong-dfinity-org-twin, line 397: Do as you see fit.

> If this proposal executes successfully

IMHO, this goes without saying, because the whole point of proposals is that nothing unless they get voted in (and subsequently, get executed successfully). I think it is understood that these comments explain what happens upon successful execution.

This isn't "actually" harmful though; just some "unnecessary" verbosity.

daniel-wong-dfinity-org-twin (approved):
Ideally, we'd get rid of this footgun, but ofc, that would require more effort, which we do not want to spend now.

PR #10613, opened 2026-06-30
chore: build dnsmasq from source

nmattia:
This builds `dnsmasq` — used by local system tests — from source, instead of pulling it in as an Ubuntu dependency in the build container.


basvandijk (approved):
Nice and simple.

tmu0 (approved):

PR #10619, opened 2026-07-01
fix: Try to lock the crypt device around FS creation

Bownairo:
Locking and triggering udev within the script that creates var is hopefully a fix to the udev flakiness we've seen.

### DIFF
--- a/ic-os/components/guestos/init/setup-encryption/setup-var-encryption.sh
+++ b/ic-os/components/guestos/init/setup-encryption/setup-var-encryption.sh
@@ -80,4 +80,6 @@ else
         # and continue operating.
         cryptsetup luksClose old_var_crypt || echo "Failed to close old /var crypto partition"
     fi
+
+    udevadm trigger --settle --action=change /dev/mapper/var_crypt
 fi
--- a/ic-os/components/guestos/init/setup-encryption/setup-var-encryption.sh
+++ b/ic-os/components/guestos/init/setup-encryption/setup-var-encryption.sh
@@ -39,6 +39,7 @@ else
     /opt/ic/bin/guest_disk crypt-format var "$VAR_PARTITION"
     /opt/ic/bin/guest_disk crypt-open var "$VAR_PARTITION"
     echo "Populating /var filesystem in ${VAR_PARTITION} on first boot."
+    # Prevent udev from processing the device while we format it.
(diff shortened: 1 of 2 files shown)


Bownairo:
If this works, we should be able to cleanup the changes from #8368 and #9984.

blind-oracle (approved):

ic-os/components/upgrade/systemd-generators/mount-generator
```diff
@@ -134,6 +134,7 @@ function make_var_cryptsetup() {
     echo "Before=blockdev@dev-mapper-var_crypt.target"
     echo "Wants=blockdev@dev-mapper-var_crypt.target"
     echo "Before=var.mount"
+    echo "Before=systemd-fsck@dev-mapper-var_crypt.service"
```
frankdavid: Is this necessary? fsck already has `After=dev-mapper-var_crypt.device` so if the device unit works well, we should not have to manually order these.

ic-os/components/guestos/init/setup-encryption/setup-var-encryption.sh
```diff
@@ -39,7 +39,7 @@ else
     /opt/ic/bin/guest_disk crypt-format var "$VAR_PARTITION"
     /opt/ic/bin/guest_disk crypt-open var "$VAR_PARTITION"
     echo "Populating /var filesystem in ${VAR_PARTITION} on first boot."
-    mkfs.ext4 -F /dev/mapper/var_crypt -d /var
+    udevadm lock --device=/dev/mapper/var_crypt mkfs.ext4 -F /dev/mapper/var_crypt -d /var
```
frankdavid, line 43: Could you add a comment, e.g. 

```
Prevent udev from processing the device while we format it.
```

frankdavid:
Consider adding a lock to `setup-shared-*.sh` scripts, too.

frankdavid (approved):

ic-os/components/upgrade/systemd-generators/mount-generator
Bownairo: With only the `device` dependency, fsck could start as soon as `mkfs.ext4` exits. I think that's mostly fine, but I think it would be good to wait for the debugfs stuff and log copying too. I guess the logs should probably come after fsck... but maybe not for this change.

Bownairo: I've moved the remaining changes into https://github.com/dfinity/ic/pull/10656.

PR #10622, opened 2026-07-01
feat(ICP_Rosetta): Release ICP Rosetta 2.1.10

mbjorkqvist:
Release ICP Rosetta v2.1.10 with the following changes:

- Use a deterministic memo of 0 instead of a random one for transfers without a specified memo in `construction/payloads`.
- Optimize database query performance.
- Enforce a 24h ingress window in `construction/payloads`.


gregorydemay (approved):
Thanks @mbjorkqvist

PR #10623, opened 2026-07-01
chore: drop critical error for unexpected infinite loops

mraszyk:
In every loop iteration, we do `output_iter.exclude_queue()` or `output_iter.next().unwrap()` so the loop indeed always terminates and not asserting that it does so requires additional error handling downstream (in a follow-up PR) whose complexity is not worth the benefit of this safe-guard.


alin-at-dfinity (approved):
LGTM, thank you.

PR #10624, opened 2026-07-01
perf: cache the result of `read_registry`

hpeebles:
At the start of each round we call `read_registry`.

Prior to this PR, this would involve computing the full registry state each time.

However, the registry at a given version is immutable, so we can safely cache the result of the latest call to `read_registry`, and simply return a clone of that value whenever subsequent requests pass in the matching registry version.

Given that the registry version changes very infrequently relative to the round rate, the cache hit ratio will be extremely high.

This change also wraps the `NetworkTopology` returned by `read_registry` in an `Arc`.


rs/messaging/src/message_routing.rs
```diff
@@ -594,6 +594,9 @@ where
     log: ReplicaLogger,
     #[allow(dead_code)]
     malicious_flags: MaliciousFlags,
+    /// Caches the most recent `read_registry` result so it can be reused within
+    /// subsequent rounds whose registry version is unchanged (the common case).
+    cached_registry_read: Mutex<Option<(RegistryVersion, ReadRegistryResult)>>,
```
alin-at-dfinity: While this will definitely work as written, this goes against the grain of how information flows through Message Routing / Replicated State.

All the Message Routing components, including `BatchProcessorImpl` are essentially stateless. `state_manager` above is internally mutable (and uses `Mutex` to achieve that) but it's the one accepted exception: you take the later state out of it and you commit the new state; that's all there is to it.

The more consistent/cleaner way would be to consolidate `ReadRegistryResult` into a couple of structs. I would have put everything it holds inside `NetworkTopology`, as that was its goal, but right now it seems to make more sense to move the "network" parts of it (which I guess is only `ApiBoundaryNodes`) inside `NetworkTopology` and the rest, having to do with the subnet's features, limits, node keys, inside a new `SubnetTopology` (?) struct. Then you wrap both within `Arcs` inside `SystemMetadata`, store the registry version next to them and reuse them or not from one `ReplicatedState` to the next, depending on whether the registry version is the same or not.

If that seems too involved, I'm happy to pick this up and do it myself. Whichever way you prefer.

hpeebles: Right that all makes sense.
I'm happy to have a quick go at that when I get some time (probably tomorrow). But if it looks as though it's becoming too large or requires decisions to be made I think I'll throw it over to you.

hpeebles: I've had a go at this, and it is "working", but it isn't releasable since the changes to the system metadata will result in a different certified hash, so these changes would need to be orchestrated carefully in order to be forward/backward compatible, which pushes this out of my depth 😅, so I'll throw this over to you.

In my version I added this struct which results in a few nice simplifications:

```
pub struct OwnSubnetTopology {
    /// The subnet ID this data was populated for.
    pub own_subnet_id: SubnetId,
    pub features: SubnetFeatures,
    pub resource_limits: ResourceLimits,
    pub registry_execution_settings: RegistryExecutionSettings,
    /// DER-encoded public keys of the subnet's nodes.
    pub node_public_keys: BTreeMap<NodeId, Vec<u8>>,
}
```

The `read_registry` response simplifies to `Result<(NetworkTopology, OwnSubnetTopology), ReadRegistryError>`.

And `execute_round` becomes `fn execute_round(&self, state: ReplicatedState, batch: Batch) -> ReplicatedState;`.

In fact I'll push my latest changes to this branch in case they'll be of use to you.

PR #10625, opened 2026-07-01
feat: allowing engine controller canister to perform directly membership changes

NikolaMilosa:
In order to allow people to replace nodes in their engines we need to allow the engine controller to proxy the calls on behalf of engine owners.

### DIFF
--- a/rs/engine_controller/engine_controller.did
+++ b/rs/engine_controller/engine_controller.did
@@ -110,7 +110,7 @@ type DeployGuestosToAllSubnetNodesPayload = record {
 // Mirror of `registry_canister::mutations::do_change_subnet_membership::
 // ChangeSubnetMembershipPayload`. Forwarded unchanged to the registry; the
 // registry enforces that the target subnet is of type `CloudEngine` when the
-// caller is the engine controller.
+// caller is the engine controller canister.
 type ChangeSubnetMembershipPayload = record {
   subnet_id : principal;
   node_ids_add : vec principal;
--- a/rs/engine_controller/engine_controller.did
+++ b/rs/engine_controller/engine_controller.did
@@ -107,9 +107,20 @@ type DeployGuestosToAllSubnetNodesPayload = record {
   replica_version_id : text;
 };
 
+// Mirror of `registry_canister::mutations::do_change_subnet_membership::
+// ChangeSubnetMembershipPayload`. Forwarded unchanged to the registry; the
+// registry enforces that the target subnet is of type `CloudEngine` when the
+// caller is the engine controller.
+type ChangeSubnetMembershipPayload = record {
+  subnet_id : principal;
+  node_ids_add : vec principal;
+  node_ids_remove : vec principal;
+};
+
 service : (opt EngineControllerInitArgs) -> {
   create_engine : (CreateEngineArgs) -> (CreateEngineResult);
   delete_engine : (DeleteEngineArgs) -> (Result);
   update_subnet : (UpdateSubnetPayload) -> (Result);
+  change_subnet_membership : (ChangeSubnetMembershipPayload) -> (Result);
   deploy_guestos_to_all_subnet_nodes : (DeployGuestosToAllSubnetNodesPayload) -> (Result);
 }
--- a/rs/engine_controller/unreleased_changelog.md
+++ b/rs/engine_controller/unreleased_changelog.md
@@ -11,7 +11,7 @@ on the process that this file is part of, see
 
 * New `change_subnet_membership` update method that proxies to the registry's
   `change_subnet_membership` endpoint. The registry restricts this to
-  `CloudEngine` subnets when the caller is the engine controller.
+  `CloudEngine` subnets when the caller is the engine controller canister.
 
 ## Changed
 
--- a/rs/engine_controller/unreleased_changelog.md
+++ b/rs/engine_controller/unreleased_changelog.md
@@ -9,12 +9,11 @@ on the process that this file is part of, see
 
 ## Added
 
-## Changed
+* New `change_subnet_membership` update method that proxies to the registry's
+  `change_subnet_membership` endpoint. The registry restricts this to
+  `CloudEngine` subnets when the caller is the engine controller.
 
-* `change_subnet_membership` may now be called by the engine controller canister
-  in addition to the governance canister. When invoked by the engine controller,
-  the target subnet must be of type `CloudEngine`; governance retains
-  unrestricted access to any subnet.
+## Changed
 
 ## Deprecated
 
--- a/rs/registry/canister/src/mutations/do_change_subnet_membership.rs
+++ b/rs/registry/canister/src/mutations/do_change_subnet_membership.rs
@@ -22,7 +22,7 @@ impl Registry {
         let subnet_id = SubnetId::from(payload.subnet_id);
         let mut subnet_record = self.get_subnet_or_panic(subnet_id);
 
-        // When invoked by the engine controller, restrict the operation to
+        // When invoked by the engine controller canister, restrict the operation to
         // `CloudEngine` subnets. Governance retains full ability to change
         // membership of any subnet.
         if caller == ENGINE_CONTROLLER_CANISTER_ID.get() {
--- a/rs/registry/canister/src/mutations/do_change_subnet_membership.rs
+++ b/rs/registry/canister/src/mutations/do_change_subnet_membership.rs
@@ -6,6 +6,8 @@ use candid::{CandidType, Deserialize};
 #[cfg(target_arch = "wasm32")]
 use dfn_core::println;
 use ic_base_types::{NodeId, PrincipalId, SubnetId};
+use ic_nns_constants::ENGINE_CONTROLLER_CANISTER_ID;
+use ic_protobuf::registry::subnet::v1::SubnetType;
 use ic_registry_keys::make_subnet_record_key;
 use ic_registry_transport::upsert;
 use prost::Message;
@@ -13,15 +15,28 @@ use serde::Serialize;
 
 impl Registry {
     /// Changes membership of nodes in a subnet record in the registry.
-    ///
-    /// This method is called by the governance canister, after a proposal
-    /// for modifying a subnet by changing the membership (adding/removing) has been accepted.
     pub fn do_change_subnet_membership(&mut self, payload: ChangeSubnetMembershipPayload) {
-        println!("{LOG_PREFIX}do_change_subnet_membership started: {payload:?}");
+        let caller = dfn_core::api::caller();
+        println!("{LOG_PREFIX}do_change_subnet_membership started (caller: {caller}): {payload:?}");
 
-        let nodes_to_add = payload.node_ids_add.clone();
         let subnet_id = SubnetId::from(payload.subnet_id);
         let mut subnet_record = self.get_subnet_or_panic(subnet_id);
+
+        // When invoked by the engine controller, restrict the operation to
+        // `CloudEngine` subnets. Governance retains full ability to change
+        // membership of any subnet.
+        if caller == ENGINE_CONTROLLER_CANISTER_ID.get() {
+            assert_eq!(
+                subnet_record.subnet_type,
+                i32::from(SubnetType::CloudEngine),
+                "{LOG_PREFIX}do_change_subnet_membership: the engine controller may \
+                 only change membership of CloudEngine subnets; subnet {subnet_id} \
+                 has subnet_type {}",
+                subnet_record.subnet_type
+            );
+        }
+
+        let nodes_to_add = payload.node_ids_add.clone();
 
         let current_subnet_nodes: Vec<NodeId> = subnet_record
             .membership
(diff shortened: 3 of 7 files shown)


rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+    // to remove; take the first one from the subnet record.
+    let mut builder = RegistryCanisterInitPayloadBuilder::new();
+    builder.push_init_mutate_request(invariant_compliant_mutation_as_atomic_req(0));
+    builder.push_init_mutate_request(cloud_engine_mutate);
+    builder.push_init_mutate_request(extra_node_mutate);
+    install_registry_canister_with_payload_builder(&pocket_ic, builder.build(), true).await;
+
+    // Read a current member of the CloudEngine subnet to use as the removal
+    // target.
+    use ic_protobuf::registry::subnet::v1::SubnetRecord as SubnetRecordPb;
```
daniel-wong-dfinity-org-twin: why?? I think you already brought in SubnetRecord and can use that? In any case, this should not be down here; `use` at the top of module.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+            "change_subnet_membership",
+            Encode!(&payload).unwrap(),
+        )
+        .await
+        .unwrap_or_else(|err| {
+            panic!(
+                "change_subnet_membership call by the engine controller on a CloudEngine \
+                 subnet was unexpectedly rejected: {err:?}"
+            )
+        });
```
daniel-wong-dfinity-org-twin: Do as you see fit.

Probably, `.unwrap()` is sufficient.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+
+    // change_subnet_membership returns nothing on success; if decoding as `()`
+    // works, the call completed successfully.
+    Decode!(&response).unwrap();
```
daniel-wong-dfinity-org-twin: I didn't know you are not required to explicitly tell Decode what type you were expecting.

Regardless, let's be explicit: tell it what type you want. One benefit of explicitness here is that it reduces the need for commenting.

Another argument in favor of explicitness here: consistency. I've never seen Decode called with***OUT*** explicitly specifying the expected type.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+
+    // Verify the membership actually changed.
+    use ic_protobuf::registry::subnet::v1::SubnetRecord as SubnetRecordPb;
```
daniel-wong-dfinity-org-twin: Ditto.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+        make_subnet_record_key(ic_base_types::SubnetId::from(PrincipalId::from(
+            cloud_engine_subnet_id,
+        ))),
+    )
+    .await;
+    let members: Vec<NodeId> = subnet_record
+        .membership
+        .iter()
+        .map(|bytes| NodeId::from(PrincipalId::try_from(bytes.as_slice()).unwrap()))
+        .collect();
```
daniel-wong-dfinity-org-twin: Do as you see fit (including nothing).

Despite what seems to be encouraged in Rust, I am pretty darn sure that

```
let widgets: Vec<...> = foo
    .iter()
    .map(...)
    .etc(...)
    .collect();
```

is an anti-pattern. Instead, we should do

```
let widgets = foo
    .iter()
    .map(...)
    .etc(...)
    .collect::<Vec<...>>();
```

***Why***: the first way involves an arbitrarily large separation between two closely related things: collect what its actual behavior (i.e. return a Vec).

If you want to also indicate the type of `widgets` next to `widgets` itself, that's fine; you can ALSO do `let widgets: Vec<...>`, if you think that helps.

But when I reach `collect` in my reading of this code, I shouldn't have to scan back up to `let widgets` to confirm that `collect` will return a `Vec`. Instead, that information should be right where I need it, where my eyes already are.

AFAICT, the only reason the first way is considered "idiomatic" is that it uses Rusts powerful inference capabilities. But to what end?? Seems that the only thing it accomplishes is cuteness. But clever code is NOT good! Boring code is good! The thing that Rust infers is "What type of thing does this `collect` call produce/return?". Clearly, this fact should be right where it is most relevant: right next to the `collect` call itself! Locality good!

When you go to the grocery store, you do not leave the shopping list at home. You keep it on you, because that maximizes locality. The "idiomatic" way here is equivalent to leaving the shopping list at home!

> Explicit is better than implicit.
> --Zen of Python

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+    let non_cloud_engine_subnet =
+        Principal::try_from(subnet_list.subnets.first().unwrap().as_slice()).unwrap();
+
+    // The exact contents of the payload do not matter: the CloudEngine check
+    // in `do_change_subnet_membership` fires before any mutation would be
+    // applied. Use an empty add/remove so the payload is otherwise valid.
+    let payload = ChangeSubnetMembershipPayload {
+        subnet_id: PrincipalId::from(non_cloud_engine_subnet),
+        node_ids_add: vec![],
+        node_ids_remove: vec![],
```
daniel-wong-dfinity-org-twin, line 555: I am surprised that this is valid. Can we maybe have an assertion that it is valid? Otherwise, one can reasonably play devil's advocate: ok, the Registry call failed, but maybe that's just because Registry does not like these empty Vecs, not because of the reason we want (i.e. the Engine Controller Canister is blocked from touching the NNS subnet)?

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+
+    // Identify the (only) non-CloudEngine subnet.
+    let subnet_list =
+        decode_registry_value::<SubnetListRecord>(&pocket_ic, make_subnet_list_record_key()).await;
+    let non_cloud_engine_subnet =
+        Principal::try_from(subnet_list.subnets.first().unwrap().as_slice()).unwrap();
+
+    // The exact contents of the payload do not matter: the CloudEngine check
+    // in `do_change_subnet_membership` fires before any mutation would be
+    // applied. Use an empty add/remove so the payload is otherwise valid.
```
daniel-wong-dfinity-org-twin: Tests should not refer to implementation details. This comment mentions lots of implementation details.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+        .await;
+
+    assert!(
+        response.as_ref().is_err_and(|err| err
+            .reject_message
+            .contains("may only change membership of CloudEngine subnets")),
+        "expected the engine controller's change_subnet_membership call on a \
+         non-CloudEngine subnet to be rejected with a CloudEngine-specific message, \
+         got {response:?}"
+    );
```
daniel-wong-dfinity-org-twin: Do as you see fit.

This can be done using match:

```
let err = match response {
    Err(err) => err,
    _ => panic!("{response:?}"),
};
assert!(err.reject_message.contains(...));
```

This has less nesting, and does not require esoteric things like `is_err_and`. Primitive is boring. Boring is good.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+            "change_subnet_membership",
+            Encode!(&payload).unwrap(),
+        )
+        .await
+        .unwrap_or_else(|err| {
+            panic!(
+                "change_subnet_membership call by governance on a non-CloudEngine \
+                 subnet was unexpectedly rejected: {err:?}"
+            )
+        });
```
daniel-wong-dfinity-org-twin: Ditto.

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+            .contains("may only change membership of CloudEngine subnets")),
+        "expected the engine controller's change_subnet_membership call on a \
+         non-CloudEngine subnet to be rejected with a CloudEngine-specific message, \
+         got {response:?}"
+    );
+
+    // For a comparison baseline, governance (which is exempt from the
+    // CloudEngine restriction) should have its call accepted by the caller
+    // check; the empty payload is a no-op mutation, which succeeds.
+    let response = pocket_ic
```
daniel-wong-dfinity-org-twin: Do as you see fit.

```suggestion
    let response: Vec<u8> = pocket_ic
```

IMHO, this is an example of "dropping useful hints for the reader, hints that the compiler does not need, but are nevertheless valuable". Always consider the reader, because someday, that poor soul might be YOU!

rs/registry/canister/tests/change_subnet_membership.rs
```diff
@@ -410,3 +429,182 @@ fn test_change_subnet_membership_duplicate_nodes() {
...
+            Encode!(&payload).unwrap(),
+        )
+        .await
+        .unwrap_or_else(|err| {
+            panic!(
+                "change_subnet_membership call by governance on a non-CloudEngine \
+                 subnet was unexpectedly rejected: {err:?}"
+            )
+        });
+    Decode!(&response).unwrap();
```
daniel-wong-dfinity-org-twin: Ditto.

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/tests/change_subnet_membership.rs
NikolaMilosa: Also we could do: `response.unwrap_err();` and continue with assertion. I've picked that if that is okay with you.

rs/engine_controller/engine_controller.did
```diff
@@ -107,9 +107,20 @@ type DeployGuestosToAllSubnetNodesPayload = record {
   replica_version_id : text;
 };
 
+// Mirror of `registry_canister::mutations::do_change_subnet_membership::
+// ChangeSubnetMembershipPayload`. Forwarded unchanged to the registry; the
+// registry enforces that the target subnet is of type `CloudEngine` when the
+// caller is the engine controller.
```
r-birkner:
```suggestion
// registry enforces that the target subnet is of type `CloudEngine` when the
// caller is the engine controller canister.
```
It took me a while to understand the comment as I was somehow thinking `engine controller == subnet admin`, but here we actually talk about the `canister`.

rs/engine_controller/unreleased_changelog.md
```diff
@@ -0,0 +1,24 @@
...
+`rs/nervous_system/changelog_process.md`.
+
+
+# Next Upgrade Proposal
+
+## Added
+
+* New `change_subnet_membership` update method that proxies to the registry's
+  `change_subnet_membership` endpoint. The registry restricts this to
+  `CloudEngine` subnets when the caller is the engine controller.
```
r-birkner:
```suggestion
  `CloudEngine` subnets when the caller is the engine controller canister.
```

rs/registry/canister/src/mutations/do_change_subnet_membership.rs
```diff
@@ -6,23 +6,38 @@ use candid::{CandidType, Deserialize};
...
     pub fn do_change_subnet_membership(&mut self, payload: ChangeSubnetMembershipPayload) {
-        println!("{LOG_PREFIX}do_change_subnet_membership started: {payload:?}");
+        let caller = dfn_core::api::caller();
+        println!("{LOG_PREFIX}do_change_subnet_membership started (caller: {caller}): {payload:?}");
 
-        let nodes_to_add = payload.node_ids_add.clone();
         let subnet_id = SubnetId::from(payload.subnet_id);
         let mut subnet_record = self.get_subnet_or_panic(subnet_id);
 
+        // When invoked by the engine controller, restrict the operation to
```
r-birkner:
```suggestion
        // When invoked by the engine controller canister, restrict the operation to
```

rs/engine_controller/engine_controller.did
NikolaMilosa: Yup, you are right, nice catch. Fixing...

r-birkner (approved):

PR #10579, opened 2026-06-26
chore: evaluate Remote Build Execution (RBE) @ Namespace

basvandijk:
## Why

To reduce reliance on DFINITY's own Data Centers and to make developing, building and testing the IC more portable we're planning to switch the majority of CI to run on Namespace, a CI provider who DFINITY already uses for ARM64 builds on Linux and MacOS.

In particular we're planning to move the `bazel-test-all` job (AMD64 bazel build and test) to Namespace. This includes system-tests. Since Namespace does not have access to Farm the system-tests on Namespace will use the `Local` system-test backend, i.e. each test will spawn its QEMU VMs locally on the worker running the bazel action.

Considering that local system-tests are resource heavy, i.e. they can easily consume 10s of CPUs, they shouldn't be executed on a single worker because that would effectively sequentialise all system-tests which would consume almost 4 hours.

We therefor configure bazel to run its actions on a [Remote Build Execution (RBE) cluster @ Namespace](https://namespace.so/docs/bazel/execution). This way builds and tests distribute over many workers and run in parallel. 15 workers for now of shape 32x64 that each can accept 16 action slots.

Before switching to RBE we first need to evaluate its performance and stability. That's this PR. 

## What

To compare our existing `bazel-test-all` with RBE we need to have a copy of that job that is as similar as possible with the only difference that it runs bazel on RBE @ Namespace. To this end:

* `ci-kickoff.yml` is extended with a `ci-rbe-evaluation` job (like `ci-main`)  which calls the new `ci-rbe-evaluation.yml` workflow.

* `ci-rbe-evaluation.yml` is a slimmed down copy of `ci-main.yml` with similar jobs like `config`, `infer-bazel-targets` and `bazel-test-all-rbe`. The `infer-bazel-targets` job will behave the same as in `ci-main` except that it excludes Farm-based system-tests; the reverse is true for  `infer-bazel-targets` in `ci-main` which excludes local system-tests.

* `bazel-test-all-rbe` is similar to `bazel-test-all` except that it runs on a `namespace-profile-rbe-driver-amd64-linux-16x32` runner and configures bazel to use Namespace's RBE cluster.
  Note that this job doesn't use the same container image as `bazel-test-all` but instead uses a special optimised variant stored at Namespace's Container Registry (`nscr.io`).

[...]

PR #10633, opened 2026-07-02
test: run permission-denied tests as an unprivileged user when root

basvandijk:
# What

Supports running rust_tests, that assume being run as an unprivileged user, as root.

Spin off from the [RBE @ Namespace experiment](https://github.com/dfinity/ic/pull/10579).

# Why

In certain environments, like on a Remote Build Execution cluster @ Namespace, bazel runs as root. This breaks several tests that assert `PermissionDenied` on files whose permission bits have been restricted. Root holds `CAP_DAC_OVERRIDE` and bypasses filesystem permission bits, so when run as root the operations succeed and the tests fail.

# How

This adds the `ic-test-utilities-privileges` crate exposing `run_as_nobody_if_root`, which runs a test closure in a forked child that drops to the `nobody` user when the process is root (redirecting `TMPDIR`/`TEST_TMPDIR` to a `nobody`-owned base and mirroring the child's outcome — including panic messages — so `#[should_panic]` keeps working), and runs in-process otherwise. The helper is gated to Linux since the semantics (uid 65534, `CAP_DAC_OVERRIDE`) are Linux-specific and `nix` 0.24.3 lacks `setgroups` on Apple targets.

Since wrapping a whole test body in a closure pollutes it with extra code and indentation, the companion proc-macro crate `ic-test-utilities-privileges-macros` provides the `#[as_nobody_when_root]` attribute, which performs that wrapping at compile time. Proc-macros must live in their own crate, so it is re-exported through `ic-test-utilities-privileges` (the `serde`/`serde_derive` facade pattern, like `canlog` re-exports `canlog_derive`) and consumers keep a single dependency. The attribute composes with `#[test]` and `#[should_panic]`, and rejects attribute arguments, `async fn`s, and non-unit return types at compile time.

The attribute is applied to the affected permission tests in `rs/sys`, the crypto service provider, `rs/state_manager`, and `rs/sns/cli`.

Tested on a RBE cluster running as root @ Namespace in https://github.com/dfinity/ic/pull/10579 and manually verified by running `sudo bazel test`.


rs/test_utilities/privileges/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        let (mut reader, writer) = std::io::pipe().expect("failed to create pipe");
+
+        match unsafe { fork() }.expect("fork failed") {
+            ForkResult::Child => {
+                drop(reader);
+                let mut writer = writer;
+                let code = run_child(action, &mut writer);
+                // Skip destructors / at-exit handlers inherited from the parent.
+                std::process::exit(code);
+            }
```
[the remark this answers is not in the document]
basvandijk: Fixed in 75903f43b7: the child now terminates with `_exit(2)` (via `nix::libc::_exit`), skipping the inherited `atexit(3)` handlers and stdio flushing.

rs/test_utilities/privileges/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        std::env::temp_dir().join(format!("ic-nobody-{pid}"))
+    }
+
+    fn redirect_tmp_and_drop_privileges() -> nix::Result<()> {
+        // Point the temp-dir helpers at a base the `nobody` child can own, so
+        // that files it creates (and then restricts) deny `nobody` as an owner.
+        let base = nobody_tmp_base(std::process::id() as i32);
+        std::fs::create_dir_all(&base).map_err(|_| nix::errno::Errno::EACCES)?;
+        std::fs::set_permissions(&base, std::fs::Permissions::from_mode(0o777))
+            .map_err(|_| nix::errno::Errno::EACCES)?;
```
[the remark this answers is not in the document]
basvandijk: Fixed in 75903f43b7: the base is now `chown`ed to `nobody` (65534:65534) with mode `0700`, and created with `create_dir` rather than `create_dir_all` so a pre-existing directory of unknown ownership is rejected instead of silently adopted.

rs/test_utilities/privileges/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// `expected = "..."`, keeps working); if it returns, the caller returns.
+    ///
+    /// # Notes
+    /// * The *whole* closure runs in the child, so any temporary files it
+    ///   creates are owned by `nobody`; restricting their permissions therefore
+    ///   denies `nobody` exactly as it would a normal non-root owner. Threads
+    ///   spawned by `action` also run as `nobody`.
+    /// * Temporary directories must be derived from `TMPDIR`/`TEST_TMPDIR` (as
+    ///   `tempfile` and `ic_test_utilities_tmpdir::tmpdir` are); the child points
+    ///   both at a world-writable, `nobody`-owned base.
```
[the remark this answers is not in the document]
basvandijk: Fixed in 75903f43b7: the docs now describe a `nobody`-owned base with mode `0700`, prepared by the parent.

rs/test_utilities/privileges/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+            Ok(()) => 0,
+            Err(payload) => {
+                let _ = write!(writer, "{}", panic_message(payload.as_ref()));
+                PANIC_EXIT_CODE
+            }
+        }
+    }
+
+    /// A world-writable base directory, created (as root) before privileges are
+    /// dropped, in which the `nobody` child owns everything it creates.
```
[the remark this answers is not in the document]
basvandijk: Fixed in 75903f43b7: this now lives in the `TmpDirRedirect` docs, which describe the `nobody`-owned `0700` base.

rs/sys/BUILD.bazel
```diff
@@ -37,6 +37,7 @@ rust_test(
     crate = ":sys",
     deps = [
         # Keep sorted.
+        "//rs/test_utilities/privileges",
```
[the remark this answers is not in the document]
basvandijk: Addressed differently in 75903f43b7. Forcing `--test-threads=1` here would serialize a large suite for a hazard that only exists when running as root, so instead the helper itself was made fork-safe: the temp base and `putenv(3)` strings are prepared in the parent *before* forking, and the child no longer calls `std::env::set_var` (whose global environment lock another parent thread could have held at fork time), using `putenv` with the pre-allocated strings plus direct privilege-drop syscalls and `_exit(2)`. The remaining child-side work is allocation, which glibc makes fork-safe via its own atfork handlers. A concurrency regression test (8 threads x 8 forks) was added in rs/test_utilities/privileges/tests/smoke.rs and verified as root.

rs/state_manager/BUILD.bazel
```diff
@@ -69,6 +69,7 @@ rust_test(
         "//packages/ic-error-types",
         "//rs/test_utilities/consensus",
         "//rs/test_utilities/logger",
+        "//rs/test_utilities/privileges",
         "//rs/test_utilities/state",
```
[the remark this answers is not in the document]
basvandijk: Addressed differently in 75903f43b7. Forcing `--test-threads=1` here would serialize a large suite for a hazard that only exists when running as root, so instead the helper itself was made fork-safe: the temp base and `putenv(3)` strings are prepared in the parent *before* forking, and the child no longer calls `std::env::set_var` (whose global environment lock another parent thread could have held at fork time), using `putenv` with the pre-allocated strings plus direct privilege-drop syscalls and `_exit(2)`. The remaining child-side work is allocation, which glibc makes fork-safe via its own atfork handlers. A concurrency regression test (8 threads x 8 forks) was added in rs/test_utilities/privileges/tests/smoke.rs and verified as root.

rs/crypto/internal/crypto_service_provider/BUILD.bazel
```diff
@@ -109,6 +109,7 @@ rust_test(
         "//rs/crypto/test_utils/reproducible_rng",
         "//rs/crypto/utils/basic_sig",
         "//rs/test_utilities/in_memory_logger",
+        "//rs/test_utilities/privileges",
         "//rs/test_utilities/time",
```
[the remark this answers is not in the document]
basvandijk: Addressed differently in 75903f43b7. Forcing `--test-threads=1` here would serialize a large suite for a hazard that only exists when running as root, so instead the helper itself was made fork-safe: the temp base and `putenv(3)` strings are prepared in the parent *before* forking, and the child no longer calls `std::env::set_var` (whose global environment lock another parent thread could have held at fork time), using `putenv` with the pre-allocated strings plus direct privilege-drop syscalls and `_exit(2)`. The remaining child-side work is allocation, which glibc makes fork-safe via its own atfork handlers. A concurrency regression test (8 threads x 8 forks) was added in rs/test_utilities/privileges/tests/smoke.rs and verified as root.

rs/sns/cli/src/propose/propose_tests.rs
```diff
@@ -34,25 +34,27 @@ fn test_ensure_file_exists_and_is_writeable_succeeds_when_creating_file() {
...
-        ensure_file_exists_and_is_writeable(temp_file_path),
-        Err(SaveToErrors::FileOpenFailed(ref path, _)) if path == &temp_file_path.to_path_buf(),
-
-    ));
-
-    // Teardown
-    // Reset permissions so the file can be deleted
-    let permissions = fs::Permissions::from_mode(READ_WRITE_PERMISSION);
-    fs::set_permissions(temp_file_path, permissions).expect("Failed to reset permissions");
+    ic_test_utilities_privileges::run_as_nobody_if_root(|| {
```
daniel-wong-dfinity-org-twin: Can you comment why this is needed? Is there a way we can get this effect without this distracting code, e.g. could this be an attribute instead? Not only would that avoid polluting the body with extra code, it also avoids "extra unnecessary" indentation.

rs/sns/cli/src/propose/propose_tests.rs
```diff
@@ -75,22 +77,24 @@ fn test_save_proposal_id_to_file_succeeds() {
...
-    assert!(matches!(
-        save_proposal_id_to_file(temp_file_path, &ProposalId { id: 1 }),
-        Err(SaveToErrors::FileWriteFailed(ref path, _)) if path == &temp_file_path.to_path_buf(),
-    ));
-
-    // Teardown
-    // Reset permissions so the file can be deleted
-    let permissions = fs::Permissions::from_mode(READ_WRITE_PERMISSION);
-    fs::set_permissions(temp_file_path, permissions).expect("Failed to reset permissions");
+    ic_test_utilities_privileges::run_as_nobody_if_root(|| {
```
daniel-wong-dfinity-org-twin: Ditto.

daniel-wong-dfinity-org-twin (approved):

rs/sns/cli/src/propose/propose_tests.rs
basvandijk: Added a short comment at this (and every other) call site in 071d0ee6d2 explaining why it is needed: root bypasses file permission bits (`CAP_DAC_OVERRIDE`), so the permission denial the test expects only happens when dropping to `nobody` on root-run workers (e.g. Bazel remote execution).

Regarding an attribute: yes, that is possible with a small proc-macro crate (e.g. `ic-test-utilities-privileges-macros`) exposing something like `#[as_nobody_when_root]` that rewrites the test fn to wrap its body in `run_as_nobody_if_root(|| { ... })` — the same fn-rewriting pattern as `rs/nervous_system/instruction_stats_update_attribute` (and `serial_test`'s `#[serial]` in the ecosystem). `#[should_panic]` keeps working since the helper re-raises the child's panic. Trade-offs: a new proc-macro crate + Bazel `rust_proc_macro` target, a proc-macro dependency for every consuming test crate, and it would have to reject `async fn` (forking around a Tokio runtime does not compose). No existing crate covers this (rusty-fork is macro-block based, forks unconditionally, and does not drop privileges). Happy to add the attribute as sugar on top of the function in a follow-up PR to keep this one small — let me know.

rs/sns/cli/src/propose/propose_tests.rs
basvandijk: Done in 071d0ee6d2 (see the sibling thread for the attribute research).

rs/sns/cli/src/propose/propose_tests.rs
basvandijk: Actually I'll go ahead and implement your attribute idea. I like it.

daniel-wong-dfinity-org-twin: Feel free to do in another PR. As you wish.

basvandijk: Implemented in e224f9f7cf: there is now a `#[ic_test_utilities_privileges::as_nobody_when_root]` attribute, so the test bodies are free of the closure and extra indentation. All 12 call sites are converted. The proc-macro lives in a companion crate re-exported through `ic-test-utilities-privileges` (canlog/canlog_derive-style facade), so consumers keep a single dependency. It composes with `#[test]`/`#[should_panic]` and rejects `async fn` and non-unit returns at compile time. Verified as root: the permission-denied tests only pass because the attribute drops privileges.

rs/sns/cli/src/propose/propose_tests.rs
basvandijk: Converted to the attribute form in e224f9f7cf.

rs/crypto/internal/crypto_service_provider/Cargo.toml
```diff
@@ -81,6 +81,7 @@ ic-crypto-utils-basic-sig = { path = "../../utils/basic_sig" }
 ic-management-canister-types-private = { path = "../../../types/management_canister_types" }
 ic-test-utilities-compare-dirs = { path = "../../../test_utilities/compare_dirs" }
 ic-test-utilities-in-memory-logger = { path = "../../../test_utilities/in_memory_logger" }
+ic-test-utilities-privileges = { path = "../../../test_utilities/privileges" }
```
alin-at-dfinity, line 84: Hmm, interesting...

Are `Cargo.toml` files not owned by anyone?

rs/test_utilities/privileges/tests/smoke.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[test]
+fn should_spawn_many_concurrently() {
+    // Regression test for fork-safety: forking from a multi-threaded process
+    // must not deadlock the children (e.g. on locks that another thread of the
+    // parent held at fork time).
+    let handles: Vec<_> = (0..8)
+        .map(|i| {
+            std::thread::spawn(move || {
```
alin-at-dfinity, line 44: Just curious: does the parent process hold any implicit locks here? If not, should it hold an explicit one?

alin-at-dfinity (approved):
LGTM, thank you.

rs/crypto/internal/crypto_service_provider/Cargo.toml
basvandijk: [Indeed](https://github.com/dfinity/ic/blob/cc1287f38c519256d7189d0730fb5f2c145fb27a/.github/CODEOWNERS#L233).

rs/test_utilities/privileges/tests/smoke.rs
basvandijk: Claude answer: 

Good question! Yes — implicitly, and they're exactly the locks the child needs after the fork:

1. **glibc's malloc arena locks** — the sibling threads allocate constantly, and the child allocates right away (`catch_unwind`, `tempdir()`). Safe only because glibc's `fork(3)` re-initializes malloc state in the child via its internal atfork handling; the test would catch e.g. a raw `clone(2)` or a non-fork-safe allocator.
2. **Rust std's env `RwLock` in read mode** — each iteration takes it in the parent (`std::env::temp_dir()` in `TmpDirRedirect::prepare`) and in the child (`tempfile::tempdir()`). Read/read is compatible, so the `putenv` child is fine. But if `apply_in_child` regressed to `std::env::set_var`, the child would need the *write* lock, which blocks until the reader count reaches 0 — and readers stranded by the fork never unlock. With 64 forks racing 8 read-locking threads, that hangs with high probability; it's the regression this test is really aimed at.

An explicit lock only adds value if the child also acquires it — but a lock the child never touches exercises nothing, and one it does touch deadlocks under *any* fork-based implementation (the child inherits the locked state), i.e. it would test a violation of the closure's contract rather than the implementation. The locks that matter are the internal ones above, and there's no public API to hold them for a controlled duration, so allocation/env churn across threads is the practical way to get them held at fork time.

(Related caveat this implies: a thread calling `std::env::set_var` concurrently in the same test binary could stall even the correct implementation, since the closure reads the env via `tempfile` — an inherent fork limitation rather than something this crate can fix.)

PR #10634, opened 2026-07-02
chore: wip qemu no libvirt

nmattia:


rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -1076,67 +931,208 @@ impl LocalBackend {
...
+    /// Reboot the VM `vm_name` via a QMP `system_reset` (a hard reset; the guest
+    /// firmware re-runs and boots again). The transient VM is not destroyed, so
+    /// its disks and console log persist across the reset.
+    pub fn reboot_vm(&self, _group_name: &str, vm_name: &str) -> Result<()> {
+        let vm_dir = self.vm_dir(vm_name);
+        let qmp_path = Self::qmp_socket_path(&vm_dir);
+        qmp_command(&qmp_path, "system_reset", &self.logger)
+            .with_context(|| format!("rebooting VM {vm_name} via QMP {}", qmp_path.display()))
+    }
+
```
basvandijk: The previous semantics was to do a *soft* reboot. 
See: https://www.libvirt.org/manpages/virsh.html#reboot.
But this does a hard reset. This might break certain assumptions in tests so I suggest keeping the soft reboot semantics. 

For a hard reset the user can just to a destroy (kill) followed by a start again which is what tests are already doing.

ci/container/files/packages.common
```diff
@@ -45,11 +45,12 @@ rsync
...
-# Provides per-test libvirtd subprocess + KVM-accelerated QEMU for the libvirt
-# network's DHCP.
-libvirt-daemon
-libvirt-daemon-driver-qemu
+# Local system-test backend (see rs/tests/driver/src/driver/local_backend.rs),
+# which launches KVM-accelerated QEMU VMs directly (no libvirt) and runs a
+# `dnsmasq` per group as an IPv6 RA daemon + DHCPv4 server. `dnsmasq-base` is
+# listed explicitly because nothing else pulls it in (it used to arrive as a
+# libvirt recommends); `ovmf` provides the guests' UEFI firmware.
+dnsmasq-base
```
nmattia: I'm surprised by this. Can't we use the one we build from source?

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -1235,24 +1231,55 @@ fn sanitize_libvirt_name(s: &str) -> String {
...
 }
 
+/// Send a single no-argument QMP command to a VM's monitor socket and wait for
+/// its reply.
+///
+/// QMP requires a capabilities handshake before commands, and may interleave
+/// asynchronous events with replies, so we complete the handshake and then skip
+/// events until the command's `return`/`error` reply arrives. Read/write
+/// timeouts keep a wedged VM from blocking teardown.
+fn qmp_command(socket_path: &Path, execute: &str, logger: &Logger) -> Result<()> {
```
nmattia, line 1300: for this and `qmp_reply`, are there no qemu cli tools that can make our lives easier?

nmattia (commented):
Couple of notes:
* In general, don't mention 'libvirt' in the comments unless there's a real reason: we don't care about previous iterations, we have the git history for that
* When the local backend and libvirt were introduced, I believe some workarounds were implemented to make session mode work, etc. Looking at the commit that introduced those changes, is there anything that could be reverted?

PR #10639, opened 2026-07-02
feat: Cache and reuse registry data across rounds

alin-at-dfinity:
Cache previously read registry values (`NetworkTopology`, `OwnSubnetInfo` and `RegistryExecutionSettings`) along with the registry version they were read at. On the next registry read (during the next execution round), if the same registry version is requested, return the cached values.

Reading the registry is now about 50% of the (admittedly already low) DSM overhead (~20 ms of a total ~40 ms; which is still 10% of the average 400 ms round).


rs/messaging/src/message_routing.rs
```diff
@@ -724,10 +771,25 @@ impl<RegistryClient_: RegistryClient> BatchProcessorImpl<RegistryClient_> {
...
         own_subnet_id: SubnetId,
-    ) -> (NetworkTopology, OwnSubnetInfo, RegistryExecutionSettings) {
-        loop {
+    ) -> (
+        Arc<NetworkTopology>,
+        Arc<OwnSubnetInfo>,
+        Arc<RegistryExecutionSettings>,
+    ) {
+        // Return the cached values if they were read at the requested version.
+        if let Some(cached) = self.cache.borrow().as_ref()
```
michael-weigelt, line 780: Can you guarantee that this borrow does not panic? 
~I'd have expected an `RwLock` since we already have `Arc`s here (rather than `Rc`s), but perhaps the `Arc`s are handed out to threads whereas this `cache` can only ever be accessed by one single thread?~

michael-weigelt: Never mind, you are only ever reading from this which means it cannot panic.

michael-weigelt (approved):

rs/messaging/src/message_routing.rs
alin-at-dfinity: Also writing it, every now and then. But all reads and writes happen from the main DSM thread only (virtually all of Message Routing runs on this one thread).

PR #10640, opened 2026-07-02
chore: vendor dfx-core identity/network resolution, drop external dfx-core dependency

lwshang:
## Why

`ic-sns-cli` and `ic-sns-testing` depended on the external
[`dfx-core`](https://crates.io/crates/dfx-core) crate. `dfx-core` re-exports
`ic-agent` in its public API, so its **major version is bumped in lockstep with
`ic-agent`** — every `ic-agent` upgrade in this repo requires a matching
`dfx-core` release.

Once the [`dfinity/sdk`](https://github.com/dfinity/sdk) repo is archived, those
`dfx-core` releases can no longer be cut, which would block this monorepo from
upgrading `ic-agent` at all. It also drags in a pile of outdated transitive deps
(`bip32 0.4`, `tiny-bip39`, `sec1 0.3`, `handlebars`, `byte-unit`, …) that other
teams have hit while trying to upgrade their own dependencies.

SNS only uses a small, self-contained slice of `dfx-core`: resolving a dfx
identity + network into an `ic_agent::Agent`, and resolving an identity name to
a principal. This PR vendors just that slice as a new internal crate and removes
the external `dfx-core` dependency.

This direction was approved by the NNS team, who own `ic-sns-cli`: .

## What

- New crate **`rs/sns/dfx-core-vendored`** — a trimmed, in-tree subset of
  `dfx-core 0.4.0` exposing two functions:
  - `get_agent(network, identity) -> Agent`
  - `get_identity_principal(name) -> Principal`
- `ic-sns-cli` and `ic-sns-testing` rewired onto it; the `dfx-core` dependency
  removed from both crates and from the workspace (`Cargo.toml`,
  `bazel/rust.MODULE.bazel`, lockfiles).
- The keyring/encrypted-PEM deps `dfx-core` used to pull in
  (`aes-gcm`, `argon2`, `dialoguer`, `keyring`, `security-framework`) become
  direct deps of the new crate; the unused ones (`bip32 0.4`, `tiny-bip39`,
  `sec1 0.3`, `handlebars`, `byte-unit`, `dunce`, `tar`, `flate2`, …) drop out of
  the graph.

## Behaviour

No user-facing change. dfx identities are untouched (`~/.config/dfx/identity/…`)
and the SNS CLI behaves as before for the supported cases:

- **Identities:** plaintext PEM, password-encrypted PEM, OS keyring, and HSM —
  all still load exactly as dfx loads them.
- **Networks:** `ic` (mainnet), `local` (shared or project, honouring a running
  replica's `webserver-port`), and explicit IC HTTP endpoint URLs (e.g.
  `sns-testing --network http://127.0.0.1:8080`).

Not vendored (unused by SNS): identity creation/rename/removal/export, wallets,
the extension manager, playground networks, and general `dfx.json`/`networks.json`
parsing beyond the `local` bind. `IdentityManager::new` no longer auto-creates a
default identity for `InitializeIdentity::Allow` (that is dfx's job), but SNS only
ever calls it with `Disallow`, so there is no behavioural change in practice.

See `rs/sns/dfx-core-vendored/README.md` for how to verify the subset against
upstream and the maintenance expectations (the crate is treated as frozen
vendored code — the only expected changes are the mechanical adjustments to track
future `ic-agent` majors, which is the whole reason it exists).

## Reviewing the vendored crate

The crate is derived from **`dfx-core 0.4.0`** (which itself depends on
`ic-agent 0.47`, this workspace's version). Most modules are copied at the **same
relative path** as upstream and trimmed; two files (`network.rs`, `lib.rs`) are
reimplementations. To confirm the copied modules really are a subset:

```sh
diff -ur sdk/src/dfx-core/src ic/rs/sns/dfx-core-vendored/src
```

That diff is ~2,530 lines, but the review surface is tiny:

| | Lines | Notes |
| --- | --- | --- |
| Deletions (trimmed-out code) | ~1,865 | Skim-only |
| `Only in …` notices for dropped files | 34 | ~60 dropped files/dirs, collapsed to one line each |
| **Additions** | **~190** | The only lines to actually read |

Of the ~190 additions:

- **~150 are the two reimplementations** — `lib.rs` (+112) and `network.rs`
  (~300 lines, which shows only as `Only in …/vendored: network.rs`, so review it
  directly). These need normal code review; there is no upstream counterpart to
  diff against.
- **~39 across all 19 copied modules**, and none of it is rustfmt noise:
  ~25 shortened `use` statements (each verifiably a subset of upstream's imports,
  since imports are dropped alongside the code that used them), 9 doc-comment
  lines (two `Unlike upstream …` notes), and 5 code lines for the edits those
  notes describe (`_initialize_identity` ignored, the `NoIdentityConfigurationFound`
  early-return, and `password_prompt` dropping its `PromptMode` argument).

(Don't run `diff -urN`: it expands every dropped file into a full deletion,
ballooning to ~11,000 lines for no benefit.)

## Testing

- `cargo build` + `cargo clippy` (`-D warnings`, per `ci/scripts/rust-lint.sh`)
  on `dfx-core-vendored`, `ic-sns-cli`, `ic-sns-testing`.
- `bazel build` of the new crate and all `//rs/sns/cli/...` + `//rs/sns/testing/...`
  targets.
- `bazel test`:
  - `//rs/sns/cli:sns_test`, `//rs/sns/cli:sns_doctest`,
    `//rs/sns/testing:sns_testing_test`
  - `//rs/sns/testing:sns_testing_ci` (full PocketIC NNS + SNS lifecycle)
  - `//rs/nervous_system/integration_tests:sns_extension_test`
- After merging the latest `master`, the crate compiles cleanly against the
  upgraded `ic-agent` / `ic-identity-hsm 0.47`, with no `dfx-core` release
  required — a live confirmation of the decoupling.


rs/sns/dfx-core-vendored/src/network.rs
```diff
@@ -0,0 +1,187 @@
...
+    Err(NetworkResolutionError::NetworkNotFound(
+        network_name.to_string(),
+    ))
+}
+
+fn parse_provider_url(url: &str) -> Result<(), NetworkResolutionError> {
+    Url::parse(url)
+        .map(|_| ())
+        .map_err(|e| NetworkResolutionError::ParseProviderUrlFailed(url.to_string(), e))
+}
```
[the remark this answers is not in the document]
lwshang: fixed in https://github.com/dfinity/ic/pull/10640/commits/dcfd201d78050a70fe5764d5f11b9b36cfe1f39d

rs/sns/dfx-core-vendored/src/identity/pem_safekeeping.rs
```diff
@@ -0,0 +1,124 @@
...
+        content,
+        if was_encrypted {
+            IdentityType::EncryptedLocal
+        } else {
+            IdentityType::Plaintext
+        },
+    ))
+}
+
+/// If the IndentityConfiguration suggests that the content of the pem file is encrypted,
```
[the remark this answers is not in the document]
lwshang: fixed in https://github.com/dfinity/ic/pull/10640/commits/dcfd201d78050a70fe5764d5f11b9b36cfe1f39d

nmattia (approved):

rs/sns/cli/src/utils.rs
```diff
@@ -9,35 +8,19 @@ use serde::{Deserialize, Serialize};
 
 /// Gets an agent for a given network and identity. This is similar to the code DFX uses internally to get an agent.
 /// If no identity is provided, it will use the identity currently selected in the DFX CLI.
+///
+/// Network and identity resolution is handled by the in-tree `dfx-core-vendored`
+/// crate (a trimmed vendoring of `dfx-core`), so this behaves like dfx for the
+/// supported networks (`ic`, `local`, and IC HTTP endpoint URLs) and identity
+/// storage modes (plaintext/encrypted PEM, keyring, and HSM).
```
daniel-wong-dfinity-org-twin: Do as you see fit.

This seems like an implementation detail, which means, it should not be mentioned in triple slash comments.

rs/sns/dfx-core-vendored/src/config/directories.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::foundation::get_user_home;
+use crate::fs::composite::ensure_dir_exists;
+use directories_next::ProjectDirs;
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::sync::{LazyLock, Mutex};
+
+pub fn project_dirs() -> Result<&'static ProjectDirs, GetUserHomeError> {
+    static DIRS: LazyLock<Option<ProjectDirs>> =
+        LazyLock::new(|| ProjectDirs::from("org", "dfinity", "dfx"));
```
daniel-wong-dfinity-org-twin: This seems like an unnecessary diff?

Also, I think people are more familiar with `lazy_static! { ... }`. Personally, I have never heard of `LazyLock`.

Unless there is some big advantage to doing it this way (e.g. the old way suddenly just doesn't work for some reason), let's minimize diffs.

rs/sns/dfx-core-vendored/src/error/fs.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[derive(Error, Debug)]
+#[error("failed to read from {path}")]
+pub struct ReadFileError {
+    pub path: PathBuf,
+    pub source: std::io::Error,
+}
+
+#[derive(Error, Debug)]
+#[error("failed to read from {path}")]
```
daniel-wong-dfinity-org-twin:
```suggestion
#[error("failed to read {path}")]
```

I think this ☝️ minimizes diffs.

rs/sns/dfx-core-vendored/src/error/get_user_home.rs
```diff
@@ -0,0 +1,8 @@
+//! Verbatim from `dfx_core::error::get_user_home`.
```
daniel-wong-dfinity-org-twin: Do as you see fit.

This seems unnecessary. The README already says that this whole dir is a copy & paste from the dfinity/sdk. IMHO, it does not seem esp helpful to restate this fact in every file. Plus, this fact is also stated (in strange code word) in the name of the dir: dfx-core-***vendored***.

rs/sns/dfx-core-vendored/src/error/identity.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    #[error("Failed to load identity manager configuration")]
+    LoadIdentityManagerConfigurationFailed(#[source] StructuredFileError),
+
+    #[error("The specified identity must exist")]
+    OverrideIdentityMustExist(#[source] RequireIdentityExistsError),
+
+    #[error(
+        r#"No identity configuration found.  Please run "dfx identity get-principal" or "dfx identity new <identity name>" to create a new identity."#
+    )]
```
daniel-wong-dfinity-org-twin: In the sdk repo, this is one line. Did rustfmt force you to do this? If not, can you make this one line to minimize diffs?

rs/sns/dfx-core-vendored/src/identity/identity_manager.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+}
+
+pub(super) fn get_dfx_hsm_pin() -> Result<String, String> {
+    std::env::var("DFX_HSM_PIN")
+        .map_err(|_| "There is no DFX_HSM_PIN environment variable.".to_string())
+}
+
+fn load_configuration(path: &std::path::Path) -> Result<Configuration, StructuredFileError> {
```
daniel-wong-dfinity-org-twin:
```suggestion
fn load_configuration(path: &Path) -> Result<Configuration, StructuredFileError> {
```

Was there a reason you changed this? Otherwise, it seems like we can avoid textual differences vs. the original, and make it easier to tell that all this is copied straight from the original.

rs/sns/dfx-core-vendored/src/identity/pem_safekeeping.rs
```diff
@@ -0,0 +1,124 @@
...
+        content,
+        if was_encrypted {
+            IdentityType::EncryptedLocal
+        } else {
+            IdentityType::Plaintext
+        },
+    ))
+}
+
+/// If the IdentityConfiguration suggests that the content of the pem file is encrypted,
```
daniel-wong-dfinity-org-twin: Do as you see fit.

Keeping the typo would be acceptable.

```suggestion
/// If the IndentityConfiguration suggests that the content of the pem file is encrypted,
```

rs/sns/dfx-core-vendored/src/identity/pem_safekeeping.rs
```diff
@@ -0,0 +1,124 @@
...
+        let pem = decrypt(pem_content, decryption_config, &password)?;
+        // print to stderr so that output redirection works for the identity export command
+        eprintln!("Decryption complete.");
+        Ok((pem, true))
+    } else {
+        Ok((Vec::from(pem_content), false))
+    }
+}
+
+fn password_prompt() -> Result<String, EncryptionError> {
```
daniel-wong-dfinity-org-twin, line 93: Can you document how this is different from the original? Perhaps, something like,

> Unlike the original, this does NOT take PromptMode. Instead, this behaves like the original when PromptMode::DecryptingToUse is passed.

rs/sns/dfx-core-vendored/src/foundation/mod.rs
daniel-wong-dfinity-org-twin, line 1: Can you do like the original, and put this in foundation/mod.rs?

rs/sns/dfx-core-vendored/src/json/mod.rs
daniel-wong-dfinity-org-twin, line 1: Ditto.

rs/sns/dfx-core-vendored/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+//!   endpoint URL.
+//! * Building an `ic_agent::Agent` (round-robin route provider + rustls client).
+//!
+//! Dropped (unused by SNS): identity creation/rename/removal/export, wallets,
+//! the extension manager, `dfx.json`/`networks.json` project config parsing
+//! beyond the `local` bind, playground networks, and arbitrary named networks.
+//!
+//! The identity modules are copied close to verbatim from `dfx-core` so they can
+//! be diffed against upstream; the network resolution is a compact
+//! reimplementation of the `ic`/`local`/URL cases.
```
daniel-wong-dfinity-org-twin: Do as you see fit.

Seems redundant vs. the README file.

rs/sns/dfx-core-vendored/src/network.rs
daniel-wong-dfinity-org-twin, line 1: This code seems to come from various files, not network/mod.rs. Can you make it so that if code comes from path p (relative to sd/src/dfx-core) that it gets copied to the same path p (relative to ic/.../dfx-core-vendored) ?

rs/sns/dfx-core-vendored/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "dfx-core-vendored"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+description.workspace = true
+documentation.workspace = true
+
+[lib]
+path = "src/lib.rs"
```
daniel-wong-dfinity-org-twin: This seems unnecessary, and is not done in the original.

rs/sns/dfx-core-vendored/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "dfx-core-vendored"
+version.workspace = true
```
daniel-wong-dfinity-org-twin: Do as you see fit.

```suggestion
version.workspace = "0.4.0"
```

I mean, since nobody else is allowed to use this, I guess it doesn't really matter, but again, let's minimize diffs.

rs/sns/dfx-core-vendored/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load("@rules_rust//rust:defs.bzl", "rust_library")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_library(
+    name = "dfx-core-vendored",
+    srcs = glob(["src/**/*.rs"]),
+    crate_name = "dfx_core_vendored",
+    version = "0.9.0",
+    deps = [
```
daniel-wong-dfinity-org-twin, line 13: Seems like the versions here could get out of sync with ./Cargo.toml. OTOH, I am not sure what to do about this fact... Maybe, Cargo.toml should be doing workspace = true?

rs/sns/dfx-core-vendored/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load("@rules_rust//rust:defs.bzl", "rust_library")
+
+package(default_visibility = ["//visibility:public"])
+
+rust_library(
+    name = "dfx-core-vendored",
+    srcs = glob(["src/**/*.rs"]),
+    crate_name = "dfx_core_vendored",
+    version = "0.9.0",
```
daniel-wong-dfinity-org-twin:
```suggestion
    version = "0.4.0",
```

rs/sns/dfx-core-vendored/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load("@rules_rust//rust:defs.bzl", "rust_library")
+
+package(default_visibility = ["//visibility:public"])
```
daniel-wong-dfinity-org-twin:
```suggestion
package(default_visibility = [
    "//rs/sns/cli:__pkg__",
    "//rs/sns/testing:__pkg__",
])
```

rs/sns/dfx-core-vendored/README.md
daniel-wong-dfinity-org-twin, line 1: Do as you see fit (probably nothing).

It seems weird to me that we would have vendored code under rs/sns. I mean, I understand your motivation: this is just a subset of the upstream code, a subset that specifically targets code in rs/sns.

One can easily imagine a future where there is more vendored code in the ic repo. Instead of it being spread out all over the place, it might make more sense for all of that to live in ic/vendored or something. idk. FWIW, at Google, there is a third_party dir. This avoids people depending on different versions of the same lib, which I guess could be a nightmare if those two pieces of code end up in the same binary. ofc, in this case, that won't happen, since dfx is on death roe.

rs/sns/dfx-core-vendored/README.md
```diff
@@ -0,0 +1,68 @@
+# `dfx-core-vendored`
+
+A minimal, in-tree vendoring of the parts of the external
```
daniel-wong-dfinity-org-twin:
```suggestion
A subset of the [`dfx-core`][...] crate.
```

"vendoring" is not a very accessible word; it is jargon, and I think many intelligent programmers are not familiar with it. Moreover, it isn't necessary. "subset" explains perfectly well.

rs/sns/dfx-core-vendored/README.md
```diff
@@ -0,0 +1,68 @@
...
+[`dfx-core`](https://crates.io/crates/dfx-core) crate that SNS tooling
+(`ic-sns-cli`, `ic-sns-testing`) depends on: resolving a dfx identity + network
+into an [`ic_agent::Agent`], and resolving a dfx identity name to its principal.
+
+## Motivation
+
+`dfx-core` re-exports `ic-agent` in its public API, so its **major version is
```
daniel-wong-dfinity-org-twin: I think the lede here is that dfx is deprecated.

Lead with the lede.

rs/sns/dfx-core-vendored/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `identity/identity_file_locations.rs` | `identity/identity_file_locations.rs` | Verbatim, minus the unused `root()` accessor. |
+| `config/directories.rs` | `config/directories.rs` | Trimmed. Keeps `project_dirs`, `get_shared_network_data_directory`, `get_user_dfx_config_dir`, and `DFX_CONFIG_ROOT`. Drops `get_shared_wallet_config_path` (wallets). |
+| `foundation.rs` | `foundation/mod.rs` | Trimmed to `get_user_home` (drops `get_current_exe`). |
+| `fs/mod.rs`, `fs/composite.rs` | `fs/mod.rs`, `fs/composite.rs` | Trimmed to the read-side helpers used by the load paths (`read`, `read_to_string`, `create_dir_all`, `ensure_dir_exists`). Drops archive extraction, all write/permission helpers, and `canonicalize` (→ also drops the `tar`/`flate2`/`dunce` deps). |
+| `json.rs` | `json/mod.rs`, `json/structure.rs` | Trimmed to `load_json_file` (drops `save_json_file` and `structure.rs`). |
+| `error/*.rs` | `error/*.rs` | Consolidated from 27 error modules to 7 (`config`, `encryption`, `fs`, `get_user_home`, `identity`, `keyring`, `structured_file`), keeping only the enums and variants reachable on the retained load/resolve paths. Messages are preserved for the retained variants. |
```
daniel-wong-dfinity-org-twin: I don't think we need this table.

Ideally, the way this should be copied over is as follows:

```
cp  -R dfx/src/dfx-core ic/rs/sns/dfx-core-vendored
```

Then, delete stuff that rs/sns/... doesn't need.

By doing it this way, one can pretty easily verify that dfx-core-vendored is a subset by simply doing

```
diff -uR dfx/src/dfx-core ic/rs/sns/dfx-core-vendored
```

In theory, this also makes it pretty clear when there are upstream changes that we want to pull in. Ofc, in practice, we do not expect to do such operations, but that's not the point. Those operations being easy is telling us that what we are doing here is easy to understand, and does not require a translation table to explain what's going on.

rs/sns/cli/src/utils.rs
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/src/error/get_user_home.rs
lwshang: Fixed in `38cba29b3c` — dropped the per-file provenance headers from all copied modules.

rs/sns/dfx-core-vendored/src/identity/pem_safekeeping.rs
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/src/identity/pem_safekeeping.rs
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/src/foundation/mod.rs
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/src/json/mod.rs
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/Cargo.toml
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/BUILD.bazel
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/BUILD.bazel
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/README.md
lwshang: Fixed in `38cba29b3c`.

rs/sns/dfx-core-vendored/src/error/fs.rs
lwshang: Fixed in `38cba29b3c`, though I used the exact upstream string rather than the suggestion: this is `ReadToStringError`, whose upstream message is `"failed to read {path} as string"` (not `"failed to read {path}"`). Matching upstream verbatim keeps it diffable.

rs/sns/dfx-core-vendored/src/error/identity.rs
lwshang: Fixed in `38cba29b3c`. And no, rustfmt didn't force the wrapping — I put it back on one line and confirmed rustfmt leaves it that way, matching upstream.

rs/sns/dfx-core-vendored/src/identity/identity_manager.rs
lwshang: Fixed in `38cba29b3c`. No reason for the change — reverted to `&Path` (re-added `Path` to the existing `std::path` import) to match upstream.

rs/sns/dfx-core-vendored/src/config/directories.rs
lwshang: Fixed in `38cba29b3c` — reverted `project_dirs()` to `lazy_static!` to match upstream. `lazy_static` is already a workspace dependency, so this adds nothing new. (Upstream keeps `LazyLock` for the `DFX_CONFIG_ROOT` static, so I left that as-is — it already matched.)

rs/sns/dfx-core-vendored/Cargo.toml
lwshang: Fixed in `38cba29b3c` — set to `version = "0.4.0"` to match the `dfx-core` release this is copied from.

rs/sns/dfx-core-vendored/BUILD.bazel
lwshang: Fixed in `38cba29b3c` — both pinned to `0.4.0` to match the `dfx-core` release this is copied from.

rs/sns/dfx-core-vendored/src/network.rs
lwshang: Done for the genuinely-copied modules (they now live at their upstream paths, e.g. `foundation/mod.rs`, `json/mod.rs`), which is what makes `diff -ur` meaningful. `network.rs` is the exception: it's not a copy of any single upstream file — it's a compact reimplementation replacing `dfx-core`'s whole `network/` + `config/` machinery (`network/provider.rs`, `network/root_key.rs`, several `config/model/*.rs`). There's no single origin path to mirror, so I kept it as one file with a header explaining it's a reimplementation. The README now calls out `network.rs` and `lib.rs` as the only two files that need to be read directly rather than diffed.

rs/sns/dfx-core-vendored/src/lib.rs
lwshang: Trimmed in `38cba29b3c`. I kept a short crate-level doc (it's the rustdoc landing page) that says what the crate is and defers to the README for the why / kept-vs-dropped / maintenance details, so it's no longer a duplicate. Like `network.rs`, `lib.rs` is a reimplementation (it replaces `interface/builder.rs` + `dfx.rs`), so a brief provenance note stays on it.

rs/sns/dfx-core-vendored/README.md
lwshang: Done in `38cba29b3c`. The README no longer has the per-module table; it explains the same-relative-path + trim approach and gives the `diff -ur sdk/src/dfx-core/src …/dfx-core-vendored/src` command to verify the subset directly (I also rebased the whole thing onto `dfx-core 0.4.0`, which matches this workspace's `ic-agent 0.47`). I put the concrete diff breakdown — how many lines are deletions vs. the ~39 real additions across copied files — in the PR description rather than the README, since it's review guidance rather than crate documentation.

rs/sns/dfx-core-vendored/README.md
lwshang: Done — all prose in the README and crate docs now uses "subset"; "vendoring/vendored" no longer appears anywhere except the crate name. I kept the name `dfx-core-vendored`: as a *name* it's the standard term for an in-tree copy of an external crate and it conveys *why* the code is here, which "subset" (a statement about content) doesn't. Happy to rename to `dfx-core-subset` if you feel strongly, but it's churn across both consumers for a name-only change.

rs/sns/dfx-core-vendored/README.md
lwshang: Left under `rs/sns/` for now. This isn't a shared third-party library — it's a private, frozen slice used only by the two SNS crates, and the README + the restricted Bazel visibility both spell out that nothing else should depend on it. If we ever accumulate genuinely shared vendored code, a top-level `third_party`/`vendor` dir would make sense, but I'd rather not set that precedent from a single-consumer special case.

rs/sns/dfx-core-vendored/README.md
daniel-wong-dfinity-org-twin: I don't see it. Perhaps, I wasn't clear. What I meant is, let's make the first sentence more like, "`dfx` is deprecated."

rs/sns/dfx-core-vendored/src/network.rs
```diff
@@ -0,0 +1,181 @@
...
+
+// Kept identical to the corresponding constants in dfx-core.
+const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
+const DEFAULT_IC_GATEWAY_TRAILING_SLASH: &str = "https://icp0.io/";
+const DEFAULT_SHARED_LOCAL_BIND: &str = "127.0.0.1:4943"; // hex for "IC"
+const DEFAULT_PROJECT_LOCAL_BIND: &str = "127.0.0.1:8000";
+
+/// The subset of a dfx `NetworkDescriptor` that SNS actually consumes.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ResolvedNetwork {
```
daniel-wong-dfinity-org-twin: Why not just re-use the original name? Boring code is very often best.

rs/sns/dfx-core-vendored/src/network.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::error::get_user_home::GetUserHomeError;
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+use url::Url;
+
+// Kept identical to the corresponding constants in dfx-core.
+const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
+const DEFAULT_IC_GATEWAY_TRAILING_SLASH: &str = "https://icp0.io/";
+const DEFAULT_SHARED_LOCAL_BIND: &str = "127.0.0.1:4943"; // hex for "IC"
+const DEFAULT_PROJECT_LOCAL_BIND: &str = "127.0.0.1:8000";
```
daniel-wong-dfinity-org-twin: Consider "address" instead of "bind" when this process is not the one listening/binding. Ditto elsewhere in this file.

daniel-wong-dfinity-org-twin (approved):

rs/sns/dfx-core-vendored/src/network.rs
lwshang: fixed in https://github.com/dfinity/ic/pull/10640/commits/2419205403f52a257d866b17944ca5ea3941a1c5

rs/sns/dfx-core-vendored/src/network.rs
lwshang: fixed in https://github.com/dfinity/ic/pull/10640/commits/2419205403f52a257d866b17944ca5ea3941a1c5

rs/sns/dfx-core-vendored/README.md
lwshang: fixed in https://github.com/dfinity/ic/pull/10640/commits/2419205403f52a257d866b17944ca5ea3941a1c5

PR #10642, opened 2026-07-03
feat(local-backend): drop libvirt and use qemu

nmattia:
This drops `libvirt` in the local system test backend, and instead calls out to qemu directly. `libvirt` was also calling out to qemu, but adding a layer of indirection as well as a layer of complexity in the build. Instead we issue the qemu commands ourselves and monitor the process directly (we don't use any of libvirt's other backends like lxc).


nmattia:
local systests passed: https://github.com/dfinity/ic/actions/runs/28601395384/job/84810254998

.github/workflows/local-system-tests.yml
```diff
@@ -1,6 +1,10 @@
 name: local-system-tests
 
 on:
+  # TEMPORARY: smoke-test this workflow from the feature branch before it's on
+  # master. REMOVE this `push` trigger before merging.
+  push:
+    branches: [nm-systest-qemu]
```
basvandijk: Just adding a comment to not forget to remove this before merge

nmattia: Thanks, would have completely forgotten!

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -1,52 +1,57 @@
...
-use virt::domain::Domain;
 
-/// Persistent record (in the root TestEnv) of the libvirtd socket and working
-/// dir, so forked task subprocesses can connect to the daemon spawned by the
-/// setup task instead of spawning their own.
+/// UEFI firmware (OVMF) split-image paths, provided by the container's `ovmf`
+/// package. The code image is read-only and shared; each VM gets a writable copy
+/// of the variable store (its per-VM UEFI NVRAM).
+const OVMF_CODE: &str = "/usr/share/OVMF/OVMF_CODE_4M.fd";
+const OVMF_VARS_TEMPLATE: &str = "/usr/share/OVMF/OVMF_VARS_4M.fd";
```
basvandijk, line 44: Future work: provide OVMF firmware via Bazel as well.

tmu0 (approved):

basvandijk (approved):
LGTM!

mbjorkqvist (approved):

blind-oracle (approved):

.github/workflows/local-system-tests.yml
nmattia:
```suggestion
```

PR #10643, opened 2026-07-03
feat: management canister's endpoint list_canisters as inter-canister call

mraszyk:
This PR allows the management canister's endpoint `list_canisters` to be called as an inter-canister call by subnet admins. The endpoint is evaluated on the same subnet as the calling canister.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api/routing.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/routing.rs
@@ -65,7 +65,8 @@ pub(super) fn resolve_destination(
     // Figure out the destination subnet based on the method and the payload.
     let method = Ic00Method::from_str(method_name);
     match method {
-        Ok(Ic00Method::CreateCanister)
+        Ok(Ic00Method::ListCanisters)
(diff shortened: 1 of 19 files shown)


michael-weigelt:
Regarding the special casing around the instructions: Have you considered just giving this endpoint a constant instruction cost and always charging that? It seems like the worst case, iterating over 1M canister IDs, cannot be that expensive, so being more precise is maybe not worth the additional complexity.

mraszyk:
> Regarding the special casing around the instructions: Have you considered just giving this endpoint a constant instruction cost and always charging that? It seems like the worst case, iterating over 1M canister IDs, cannot be that expensive, so being more precise is maybe not worth the additional complexity.

The benchmarks yielded these values:
```
    const BASE_INSTRUCTIONS: u64 = 20_000_000;
    const INSTRUCTIONS_PER_CANISTER: u64 = 16_000;
```
so 1M canisters would correspond to 16B instructions and we would actually drop a FR drop if we were enumerating so many canisters.

michael-weigelt (approved):

PR #10650, opened 2026-07-03
feat: allow local system-tests to be run as root

basvandijk:
## Overview

Make the Local (QEMU) system-test backend (`SYSTEM_TEST_INFRA=local`) work when the driver runs **as root** inside a nested user namespace whose network namespace is owned by an *ancestor* user namespace — e.g. on a Remote Build Execution cluster, where the `_local` action runs as root but is left in a netns it does not own, so RTNETLINK operations are denied.

These changes were extracted from #10579 (the RBE @ Namespace experiment) so they can land independently. The unprivileged dev-container / internal-CI path is unchanged.

## Changes

Both changes are confined to the local backend:

- **`net_admin`** (`local_backend.rs`): when running as root, run the `net-admin` script directly via `/bin/sh -c` instead of the file-capability `capsh` launcher (`NET_ADMIN_LAUNCHER`). A root process already holds the needed capabilities and its children inherit them; meanwhile `capsh`'s *file* capabilities are not honored across the user-namespace boundary, so its inheritable-set step fails with `EPERM`.

- **`ensure_administrable_netns`** (`local_backend.rs`): at driver startup, when running as root without `CAP_NET_ADMIN` over the current network namespace, `unshare(CLONE_NEWNET)` into a fresh, self-owned netns so RTNETLINK operations succeed — recreating what `linux-sandbox` gives the unprivileged case. It must run before the tokio runtime and any task subprocess is spawned so the whole process tree (task subprocesses, QEMU, `dnsmasq`) shares the netns. A `probe_netlink` helper (create + delete a short-lived bridge) decides whether the unshare is needed and verifies the fresh netns is administrable.

- **`group.rs`**: invokes `ensure_administrable_netns` early, only under the `Local` backend, before the tokio runtime and reaper are set up.

Also adds a small `running_as_root()` helper (effective uid 0).

### DIFF
--- a/rs/tests/driver/src/driver/local_backend.rs
+++ b/rs/tests/driver/src/driver/local_backend.rs
@@ -1380,13 +1380,12 @@ fn sanitize_name(s: &str) -> String {
 
 /// Whether the driver process is running as root (effective uid 0).
 ///
-/// The `bazel-test-rbe` RBE cluster runs `_local` test actions as root. A
-/// monolithic `libvirtd` selects *system* vs. *session* mode purely by
-/// `geteuid() == 0`, so as root it would ignore `$XDG_RUNTIME_DIR` and place a
-/// privileged socket under `/run`, breaking this module's session-mode design.
-/// When this returns true the backend drops libvirtd/QEMU/`dnsmasq` to `nobody`
-/// to stay in session mode; the normal (unprivileged `ubuntu`) dev-container
-/// case returns false and behaves exactly as before.
+/// The `bazel-test-rbe` RBE cluster may run `_local` test actions as root.
+///
+/// When this returns true, the local backend adjusts its networking strategy:
+/// it bypasses the file-capability `capsh` launcher in [`net_admin`] and may
+/// `unshare(CLONE_NEWNET)` in [`ensure_administrable_netns`] so RTNETLINK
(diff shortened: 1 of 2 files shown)


rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -225,6 +244,85 @@ impl LocalBackend {
...
+    /// the unprivileged case. The backend needs no external connectivity (the
+    /// bridge, TAPs and the driver's `lo` addresses are namespace-internal;
+    /// that is also why the `_local` targets do not carry `requires-network`,
+    /// see `rs/tests/system_tests.bzl`), so an isolated netns loses nothing.
+    ///
+    /// Must run in the root (parent) driver process before the tokio runtime
+    /// and the task subprocesses are spawned: `unshare` moves only the calling
+    /// thread, while threads and processes created *afterwards* inherit its
+    /// namespace. Calling this early therefore puts the entire process tree —
+    /// task subprocesses, libvirtd, QEMU, `dnsmasq` — into the same netns.
```
basvandijk:
```suggestion
    /// task subprocesses, QEMU, `dnsmasq` — into the same netns.
```

nmattia (approved):

PR #10651, opened 2026-07-03
chore: Improve logging in sev_host

frankdavid:


blind-oracle (approved):
Are we sure that tracing is set up to log correctly to stderr/out here?
Or since it's a library - we don't care and leave this to the user of it?

PR #10652, opened 2026-07-03
docs: design for supporting deposits from CEX (DEFI-2096)

gregorydemay:
Design doc for [DEFI-2096](https://dfinity.atlassian.net/browse/DEFI-2096): allow onramping ckUSDC/ckUSDT (and generally any ckERC20, later ckETH) directly from a centralized exchange withdrawal.

Deposits currently require calling the helper smart contract, which a CEX cannot do: a CEX withdrawal is a plain transfer from a shared hot wallet, carrying no IC principal. The proposed design gives each IC account a unique tECDSA-derived deposit address and sweeps funds to the minter using EIP-7702 (Pectra), so deposit addresses never need ETH for gas and remain key-recoverable independently of any contract code.

The doc covers the requirements, claim-based deposit detection, the sweeper delegate, fee model, a phased delivery (ckERC20 first, then ckETH with its balance-based detection and fixed-21k-gas constraints), and the discussed alternatives (CREATE2 forwarders, ERC-4337, permit-based sponsoring).

The runnable proof-of-concept the doc references lives in a stacked follow-up, #10670.

[DEFI-2096]: https://dfinity.atlassian.net/browse/DEFI-2096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+        event Transfer(address indexed from, address indexed to, uint256 value);
+    }
+}
+
+const CKSWEEPER_BYTECODE: &str = include_str!("../artifacts/CkSweeper.bin.hex");
+const MOCKUSDT_BYTECODE: &str = include_str!("../artifacts/MockUSDT.bin.hex");
+
+// Anvil's first two well-known dev accounts.
+const MINTER_PK: &str = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+const CEX_PK: &str = "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d";
```
gregorydemay: nit: PK can be misunderstood for PUBLICK_KEY, which is not the case here. remove the abbrevation -> MINTER_PRIVATE_KEY

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+async fn main() -> Result<()> {
+    let rpc_url =
+        std::env::var("ETH_RPC_URL").unwrap_or_else(|_| "http://127.0.0.1:8545".to_string());
+
+    let minter_signer: PrivateKeySigner = MINTER_PK.parse()?;
+    let cex_signer: PrivateKeySigner = CEX_PK.parse()?;
+    let minter = minter_signer.address();
+    let cex = cex_signer.address();
+    let mut wallet = EthereumWallet::from(minter_signer);
+    wallet.register_signer(cex_signer);
```
gregorydemay: what has the minter_signer anything to do with the cex_signer? Shouldn't they be 2 separate wallets?

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+    sweeper: Address,
+    chain_id: u64,
+) -> Result<SignedAuthorization> {
+    let authorization = Authorization {
+        chain_id: U256::from(chain_id),
+        address: sweeper,
+        nonce: 0,
+    };
+    let signature = deposit_signer.sign_hash_sync(&authorization.signature_hash())?;
+    Ok(authorization.into_signed(signature))
```
gregorydemay: nit: print the full hex of the send transaction

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    ensure!(single_receipt.status(), "single sweep reverted");
+    ensure!(usdt.balanceOf(deposit_addresses[0]).call().await?.is_zero());
+    ensure!(usdt.balanceOf(minter).call().await? == U256::from(amounts[0]));
+    ok("250 USDT swept to the minter; deposit address needed 0 ETH throughout");
+    let single_gas = single_receipt.gas_used;
+    ensure!(
+        (60_000..=90_000).contains(&single_gas),
+        "unexpected single-sweep gas: {single_gas}"
+    );
```
gregorydemay: It should be deterministic, let's hard-code the gas used.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+    let expected_total: u64 = amounts.iter().sum();
+    ensure!(usdt.balanceOf(minter).call().await? == U256::from(expected_total));
+    ok("all deposit addresses swept; minter now holds all 475 USDT");
+
+    let batch_gas = batch_receipt.gas_used;
+    ensure!(
+        batch_gas < 3 * single_gas,
+        "batching brought no amortization: {batch_gas} vs 3 x {single_gas}"
+    );
```
gregorydemay: also here hard-code the gas used.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+            format!("{} USDT", amount / 1_000_000),
+        );
+    }
+    for address in &deposit_addresses {
+        ensure!(provider.get_balance(*address).await?.is_zero());
+    }
+    ok("deposit addresses still have 0 ETH (cannot pay gas themselves)");
+
+    step("4) Minter sweeps ONE deposit address in ONE EIP-7702 transaction");
+    let single_authorization = sign_delegation(&deposit_signers[0], sweeper_address, chain_id)?;
```
gregorydemay: Let's print the minter's nonce every time the minter signs something (number of transactions sent by the minter is critical)

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,414 @@
...
+  `Transfer` event, no principal) or a native ETH send (no log at all).
+* The sender is the exchange's omnibus hot wallet, shared by all its customers, so
+  sender-based attribution is impossible.
+* Ethereum has no memo/data side-channel on plain transfers.
+
+Consequently a user holding e.g. USDT or USDC on Coinbase/Binance cannot onramp into
+ckUSDT/ckUSDC without first withdrawing to a self-custody wallet, funding it with ETH
+for gas, and interacting with the helper contract — a prohibitive UX. Funds sent
+directly to the minter address today are simply unaccounted, with no recovery path
+(see the comment on `EthBalance::eth_balance` in `src/state.rs`).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in cc578797c6: the doc now references the documentation of the `eth_balance` field of the `EthBalance` struct in `src/state.rs`.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Renamed to `MINTER_PRIVATE_KEY` / `CEX_PRIVATE_KEY` in cc578797c6.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Agreed — they are unrelated parties. Split into two separate wallets/providers (`minter_wallet`/`minter_provider` and `cex_wallet`/`cex_provider`) in cc578797c6.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Done in cc578797c6: each sweep now prints the full raw signed transaction hex (type-0x04 payload including the authorization list) before sending it.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Hard-coded in cc578797c6 as `SINGLE_SWEEP_GAS_USED = 66_889`, asserted with equality.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Hard-coded in cc578797c6 as `BATCH_SWEEP_GAS_USED = 118_982`, asserted with equality.

rs/ethereum/cketh/docs/deposit_from_cex_demo/src/main.rs
gregorydemay: 🤖 Done in cc578797c6: the minter's nonce is printed for every transaction it signs (CkSweeper deployment and both sweeps; the CEX deployment nonce is printed too for completeness).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,684 @@
...
+* `R7`: The per-token `deposit_fee` and minimum deposit amount are configurable
+  (upgrade argument / NNS proposal) such that fees cover the amortized sweep gas cost.
+* `R8`: All new state transitions (address registration, accepted/invalid deposit,
+  delegation, sweep sent/confirmed) are recorded as audit events, replayable on
+  upgrade, consistent with the minter's event-sourcing architecture.
+* `R9`: The minter dashboard and metrics expose: registered deposit addresses,
+  credited-but-unswept balances per token, delegation status, and sweep activity.
+* `R10`: Withdrawals (ckERC20 → ERC-20 and ckETH → ETH) are unaffected: they continue
+  to be served from the minter's main address and its existing nonce sequence.
+* `R13`: Registering a deposit address (`get_deposit_address`) triggers no
```
gregorydemay: Note that it must be an update call so that the minter can register it needs to scan that address for a given time period. The naming `get_` may hint to a query call, which it cannot be. Maybe something like `retrieve_deposit_address`.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,684 @@
...
+  consolidation, batched to amortize gas, and never blocks or reverts a mint (`R5`).
+  This decouples UX latency from gas-optimization policy. Variant B instead credits on
+  the sweep's own finalized event, see the open decision below.
+* **Claim-based detection instead of continuous scraping.** Deposits are detected when
+  a `notify_deposit` endpoint is called for a specific account (by the user, or
+  transparently by a frontend/timer polling recently active addresses). A targeted
+  `eth_getLogs` on one token contract filtered by one recipient address is small and
+  cheap; continuously scraping `Transfer` logs for an unbounded, growing set of
+  addresses is not. This mirrors ckBTC's `update_balance`. (The existing helper-contract
+  scraping is unchanged.)
```
gregorydemay: That goes against the target UX

> Target UX: *"I have USDT on a CEX and I want ckUSDT: I paste my deposit address into
the exchange withdrawal form and the tokens automagically appear as ckUSDT."*


It's very hard for front-end to do correctly a 2 step process (what if browser is closed after 1st step). ckETH/ckERC20 deposits are until now a single step process (call the helper smart contract)

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,684 @@
...
+* Accepted residual limitations:
+  * A CEX that batches ETH withdrawals through a contract (internal transactions)
+    provides no sender information without trace APIs; Phase 2 compliance screening is
+    therefore weaker for ETH than for ERC-20 (see Phase 2 section).
+  * An account's ETH deposit address (Phase 2) differs from its ERC-20 one, matching
+    the per-asset deposit-address UX of exchanges; sending the wrong asset class to
+    an address is not credited automatically, but funds always remain recoverable
+    (key-controlled addresses, `R12` guarantees no loss on the sender side).
+
+## Design Decisions
```
gregorydemay: How are transactions of deposits to user-derived addresses scanned against the blocklist (https://github.com/dfinity/ic/blob/97e9e67ea4232d1f40255eb63de463d7ff3e9415/rs/ethereum/cketh/minter/src/blocklist.rs)?

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Renamed to `retrieve_deposit_address` in d83eb4faf2, with the rationale spelled out in step 1: it must be an update call because it has side effects (registers the address and arms its scanning window), which a `get_`-style query could not do.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Reworked in d83eb4faf2 to keep the single-step UX: `retrieve_deposit_address` now also arms a per-address *scanning window*, and a minter background task scans all active addresses in bounded batches — no second canister call is needed (new `R15`). `notify_deposit` is demoted to an optional accelerator / re-arming mechanism for dormant addresses. The window bound is what keeps the anti-DoS property of `R13` (the registered set is inflatable for free, so an unbounded standing scan would be a cycles drain); see the variants table in step 3.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Detailed in d83eb4faf2 (step 3): the `Transfer` log's `from` address (the CEX hot wallet) is screened against the same compiled-in blocklist (`src/blocklist.rs`) as helper deposits today. If blocked: no mint, the deposit is recorded as invalid, and — an improvement over today, where blocked helper-deposit funds land commingled at the minter address — the deposit address is excluded from sweeping so the funds stay segregated at the key-controlled deposit address until explicit manual/governance release (`R3` strengthened accordingly). Under variant B the screening must run before scheduling the sweep, since the helper event's `owner` is the deposit EOA, not the real sender. For native ETH (Phase 2) balance deltas carry no sender — screening is limited to address-level checks plus optional caller-supplied tx hashes, flagged as an accepted weakening to review with compliance.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+  sweeper address' gas balance (`R17`).
+* `R10`: Withdrawals (ckERC20 → ERC-20 and ckETH → ETH) are unaffected: they continue
+  to be served from the minter's main address and its existing nonce sequence.
+* `R13`: Registering a deposit address (an unsponsored `deposit_erc20` call)
+  triggers no threshold-ECDSA signature and no Ethereum transaction (a *sponsored*
+  call may trigger both — compensated by the caller's ckETH fee). The minter only signs a
+  delegation and sweeps an address after having observed there a balance of a
+  supported token of at least the per-token minimum deposit amount. (Registrations
+  are free for callers; anything the minter spends per registration is a DoS vector
+  on its cycles and ETH.)
```
mbjorkqvist, line 126: nit: The numbering jumps from `R10` to `R13` (`R11` and `R12` are in "Phase 2" below).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+* `R6`: A sweep transaction moves funds only to the minter's main address, regardless
+  of who triggers it. No other destination is reachable through the sweeper delegate.
+* `R7`: The per-token `deposit_fee` and minimum deposit amount are configurable
+  (upgrade argument / NNS proposal) such that fees cover the amortized sweep gas cost.
+* `R8`: All new state transitions (address registration, accepted/invalid deposit,
+  delegation, sweep sent/confirmed) are recorded as audit events, replayable on
+  upgrade, consistent with the minter's event-sourcing architecture.
+* `R9`: The minter dashboard and metrics expose: registered deposit addresses,
+  credited-but-unswept balances per token, delegation status, sweep activity, and the
+  sweeper address' gas balance (`R17`).
```
mbjorkqvist, line 117: Is "registered deposit addresses" listing them all, or a count? Does the minter store something per `(principal, subaccount)`, in which case there could be a potential DoS attack if someone e.g., asks for lots of deposit addresses.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+  helper contract lands its funds, unminted, commingled at the minter address), and
+  under permissionless sweeping a third party can sweep a tainted deposit address.
+* `R4`: Transfers below the per-token minimum deposit amount, and transfers of
+  unsupported ERC-20 tokens, are not credited. No funds are ever burned or destroyed:
+  they remain at a tECDSA-controlled address and remain recoverable by the minter.
+* `R5`: Every credited deposit is eventually swept to the minter's main address. A
+  sweep failure or delay never affects already-minted balances; sweeps are retried
+  until confirmed.
+* `R6`: A sweep transaction moves funds only to the minter's main address, regardless
+  of who triggers it. No other destination is reachable through the sweeper delegate.
```
mbjorkqvist: At what cadence are sweeps performed? Can the minter detect deposits e.g., from log events, or is the only way to find out about deposits to try to sweep them? Is it free for anyone to trigger a sweep, in which case this could be a cycle draining attack vector?

rs/ethereum/cketh/docs/deposit_from_cex.md
mbjorkqvist, line 126: This part about registrations (partially) answers my earlier question(s). Can we do something to limit our exposure? E.g., rate limit registrations per principal (in `inspect_message`?), or something else?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+  tracked and offset against subsequent burns; they are never re-minted. If the fee
+  account cannot cover a sweep, no sweep is submitted. (The burn happens ahead of
+  time: funding the sweeper address is an ordinary ckETH withdrawal from the fee
+  account, covering many sweeps — see step 0.)
+* `R15`: A single user-visible step suffices: after one `deposit_erc20`
+  call, a deposit arriving at that address within its *scanning window* is credited
+  with no further canister call by the user or frontend. Re-calling
+  `deposit_erc20` (idempotent, free of per-address spending per `R13`)
+  re-arms the window; a deposit arriving on a dormant address is credited once the
+  address is re-armed and is never lost in the meantime.
```
mbjorkqvist, line 140: I could imagine that one use case would involve users depositing to the same address (same `(principal, subaccount)`) multiple times, but perhaps days apart. For the first deposit, I assume a user would click a "get deposit address" button somewhere (in OISY, or in the MULTI/DEX), which would trigger a call to `deposit_erc20` to derive the address (and open the scanning window). However, if the user then simply reuses that address (maybe stored in their address book in a CEX) for a subsequent deposit a week later, they may be confused as to why their deposit is not arriving. Is there something that could be done to improve this flow (other than a reminder/note next to the button the user initially pressed)?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+  under permissionless sweeping a third party can sweep a tainted deposit address.
+* `R4`: Transfers below the per-token minimum deposit amount, and transfers of
+  unsupported ERC-20 tokens, are not credited. No funds are ever burned or destroyed:
+  they remain at a tECDSA-controlled address and remain recoverable by the minter.
+* `R5`: Every credited deposit is eventually swept to the minter's main address. A
+  sweep failure or delay never affects already-minted balances; sweeps are retried
+  until confirmed.
+* `R6`: A sweep transaction moves funds only to the minter's main address, regardless
+  of who triggers it. No other destination is reachable through the sweeper delegate.
+* `R7`: The per-token `deposit_fee` and minimum deposit amount are configurable
```
mbjorkqvist, line 108: R7 says the fee covers the "amortized sweep gas cost", but two costs aren't captured:

1. **Detection cycles.** The background scan (step 3) spends cycles on every armed address whether or not a deposit arrives. At current mainnet gas (~0.1 gwei) a sweep is ~$0.017, while scanning one address for a full window is ~$0.02–0.03 in cycles — i.e. detection currently costs *more* than the sweep. The fee floor should be solved against detection + sweep, not sweep alone.
2. **Currency mismatch.** `deposit_fee` accrues in the deposited ckToken, but sweep gas is spent in ckETH (from the fee account) and detection in cycles. "Fees cover gas" only holds after the out-of-band treasury conversion noted in Non-goals — worth making explicit here.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+   the optional fee arguments (step 1) transfers the caller-specified ckETH amount
+   into the fee account (`icrc2_transfer_from`, see the variants below) — plus
+   treasury top-ups and converted `deposit_fee` revenue (see Non-goals).
+1. **Daily funding task**: read the fee account's balance on the ckETH ledger and
+   the sweeper address' ETH balance (`eth_getBalance`); if the sweeper balance is
+   below its low-water mark, **withdraw ckETH from the fee account to the sweeper
+   address** — an ordinary ckETH withdrawal through the existing pipeline (burn
+   from the fee account, then send the ETH on the main address' nonce sequence).
+   `R14` holds by construction, with no new burn path to audit; this pipeline is
+   the *only* way ETH is spent on sweeps.
```
mbjorkqvist: The sweeper *funding* tx ("send the ETH on the main address' nonce sequence") is not isolated by `R17`. `R17` keeps *sweep* txs off the main nonce sequence so a stuck sweep can't block withdrawals — but the funding tx rides that same sequence, so a stuck funding tx can head-of-line-block withdrawals just like any main-address tx. Probably acceptable (it's an ordinary withdrawal-shaped tx with the same resubmission machinery), but worth stating that funding shares the withdrawal lane.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+Detection runs as a **minter background task over "active" addresses**:
+
+* When `deposit_erc20` is called, the minter records the address together with the
+  then-current **last observed block number** (the latest finalized block) — the
+  scan floor: history before registration is never scanned.
+* The number of addresses tracked in parallel is **capped** (configurable). When
+  the active set is full, an unsponsored call still registers and returns the
+  address but signals that scanning is saturated; a *sponsored* call bypasses the
+  cap (the caller pays).
+* Each active address carries a **cycles budget**, decremented by its share of
```
mbjorkqvist, line 455: The window is defined as a cycles budget per tick, but the *tick cadence* is left unspecified ("bounded by cost rather than wall time" bounds total cost but not detection latency — how fast the budget is spent). Two concrete additions worth pinning down (also relevant to the cadence/DoS questions on the sweep side above):

- **Schedule.** A CEX withdrawal usually lands within minutes, so a backoff cadence (e.g. every 30s for 0–10 min, every 2 min to 30 min, every 5 min to 60 min ≈ 36 ticks) covers the likely-arrival window with ~40% fewer calls than a flat 1/min (60 ticks).
- **Cost.** Detection goes through the EVM-RPC canister on the 34-node fiduciary subnet × ~3 providers: base ≈ 0.8B cycles/call (~$0.0011) plus byte terms (~$0.0003/req-byte, ~$0.0001/resp-byte; 1T cycles = 1 XDR ≈ $1.37). A `balanceOf` tick over the active set is ~$0.0003–0.0005 per address per tick, so a full 60-min window is ~$0.017–0.03 per address (T≈3–5 tokens).

Key economic point: detection cycles are spent on **every armed address whether or not a deposit ever arrives**, but the fee is only collected on successful deposits — so empty/abandoned windows are pure minter loss. `cap × per-address-budget` is the real bound (e.g. 256 addrs × ~15B cycles ≈ 3.8T cycles ≈ $5.25 to re-arm the whole set once). Stating concrete cap / budget / cadence values would let the fee floor and the DoS ceiling both be solved explicitly.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+| **B — mint via the existing pipeline, on the sweep's own finalized helper event** | The battle-tested scrape→parse→dedup→mint pipeline is reused **unchanged** — detection (step 3) is demoted from correctness-critical to a mere scheduling hint; much smaller minter change. No liquidity window: the mint is triggered by the consolidation itself, so minted supply is always covered by the main address — today's helper-flow invariant, and `R16` is satisfied trivially | Mint follows the sweep: crediting halts if sweeping halts (empty fee account); latency tied to sweep scheduling — mitigated by sweeping on `latest`-block observations without waiting for deposit finality (a reorged deposit only wastes the sweep's gas: the delegate sweeps a zero balance, and a reorged sweep tx is absorbed by the existing nonce-tracking/resubmission machinery), making end-to-end latency comparable to today's helper flow |
```
mbjorkqvist, line 589: Worth surfacing at requirement level, not just in this cell's cons: because variant B mints *from the sweep's* helper event, crediting is coupled to sweeping. A gas spike or an empty ckETH fee account halts not just sweeps but **all new ckERC20 CEX crediting** (funds stay safe at key-controlled addresses, but new deposits stop being credited until the fee account is topped up / gas subsides). `R5`'s "a sweep delay never affects already-minted balances" holds for *already-minted* balances, but new deposits stall — this is the main operational cost of choosing B.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+sweeper delegate. Sweep transactions are type-`0x04` transactions sent from a
+**dedicated sweeper address** (tECDSA-derived with derivation path `[3u8]` — its
+own schema tag, no account components), whose nonce sequence is independent of the main address' so that a stuck
+sweep can never delay a withdrawal (`R17`); the sweeper address holds only gas
+money (funded by ckETH withdrawals from the fee account, step 0) while swept funds always land at
+the main address (`R6`). Many deposit addresses are swept in one transaction, the
+deployed delegate instance doubling as the batcher. No deposit address ever needs
+an ETH balance for gas. A periodic task selects addresses with
+observed-but-unswept balances where `unswept_value ≥ sweep_gas_cost × margin` or
+`age > max_age` — or, under variant A, **as soon as a queued withdrawal waits for
```
mbjorkqvist, line 605: `age > max_age` is effectively `R5`'s liveness net: it fires when `unswept_value` never crosses the economic threshold on its own — chiefly during gas spikes (which inflate `sweep_gas_cost × margin` above the deposit value) or for a lone below-margin balance waiting to be batched. Two notes:

- Under decided variant B this also bounds **crediting latency**, not just consolidation: since the mint follows the sweep, a below-margin deposit isn't credited until `age > max_age` forces the sweep. So `max_age` is the worst-case "my ckToken hasn't arrived yet" time for small deposits during a spike.
- `max_age` and `margin` are unspecified (as are the step-3 window cap / budget / cadence).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+| **A — direct sweep** (`CkSweeper`): delegate transfers straight to the minter's main address | Permissionless-safe: destination hardcoded (`R6`), any caller only donates gas → no access control in the delegate (with the `R3` caveat that a third party may sweep a tainted address — no worse than today); cheapest — measured 66'854 gas for a first single-address sweep incl. authorization, ≈ 26k marginal per additional address in a batch | Requires the new crediting path of step 4 variant A |
+| **B — sweep through the helper** (`CkSweeperViaHelper`): delegate approves + calls `depositErc20(token, balance, principal, subaccount)` on the existing helper | Sweep emits the canonical `ReceivedEthOrErc20` event → step 4 variant B's pipeline reuse; native ETH works symmetrically via `depositEth` | The principal is a sweep argument → it must be protected against spoofing (an arbitrary caller crediting a deposit to their own principal) by caller-gating or a one-time attestation — see the sub-variants below; more gas — measured 82'207 single (+15'353), 164'746 for a batch re-delegating three EOAs and sweeping two |
+
+**Sub-variants (variant B only) — how to prevent principal spoofing:**
```
mbjorkqvist, line 624: The anti-spoofing sub-variant reads as undecided while variant B itself is decided: the delegate in [Sweeper delegate contract](#sweeper-delegate-contract) and the demo both implement **caller-gating**, but this table declares no winner (unlike the other variant tables). The choice is cross-cutting — it determines whether sweeps are permissionless (self-attestation) or minter-only (caller-gating), and whether `R3`'s best-effort sweep-exclusion survives. Recommend either marking a decision here or explicitly flagging it as the one open sub-decision.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -0,0 +1,1029 @@
...
+(EIP-7702 delegates share the EOA's storage; using none avoids collision hazards
+entirely and leaves nothing behind on re-delegation) and its configuration held in
+`immutable`s (immutables live in the code, so they travel with the delegation).
+Per decided variant B, sweeping one token is three steps, all executing *as* the
+deposit EOA — under delegation, `address(this)` **is** the deposit address:
+
+1. **Get the balance**: `balanceOf(address(this))`. The whole current balance is
+   swept; there is no partial sweep.
+2. **Approve the balance**: approve the helper for exactly that balance, through a
+   USDT-tolerant `_safeApprove` (USDT's `approve` returns no value and reverts
```
mbjorkqvist, line 860: The sweep assumes `balanceOf(address(this))` equals the amount that will actually transfer, and that `depositErc20` moves exactly `balance`. That holds for standard tokens (USDC, USDT today) but not for fee-on-transfer / rebasing ERC-20s. Probably out of scope (same assumption as today's helper flow), but a one-line "supported tokens are assumed standard, non-fee-on-transfer" would close it.

mbjorkqvist (commented):
Thanks @gregorydemay! I have mostly clarifying questions.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: That's correct. I think it still makes sense to separate the 2, because the funding tx is expected to be infrequent, contrary to sweeps that is more user-driven.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: Yes the costs are currently not clear enough. Basically each address would get a cycle budget to account for

1. signatures
2. HTTPs outcalls

Once we have a better understanding of those cost, these will translate into a fee that will be taken by the minter. The currency mismatch is currently outscoped, but should be do able with OISY TRADE.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: We could alleviate the problem by making Multicall3 /eth_call to fetch the ERC-20 balances of multiple addresses with a single HTTPs outcall (in the order of 1000s addresses). The limit there will probably the max response size of 2MB and this is per token but this would mean keep making HTTPs outcalls forever as soon as someone called the deposit endpoint. I would say let's me conservative now, and if this is a problem we can always throw more cycles at it.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: yes, I plan to do something similar as what was planed to do for ckSOL: an address as a limited time window (say 24h) and a limited amount of cycles to reach a deposit.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: This we will refine as we go, the dashboard is not that critical. For the DoS see the comments below.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: I tend to permissionless with self-attestation but want to have a better understanding of the cost involved first (tECDSA, gas, sweeper contract complexity)

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: Ethereum blocks are produced every 12s, so something like 30s, 1min, 2 min, 4min,... seems to make sense the exact schedule is dependent of the cost calculation. See the new cost section.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Good catch. The requirements are grouped by phase rather than numbered sequentially (`R11`/`R12` are Phase 2); renumbering would ripple through dozens of `R`-references across the doc, so instead I added a one-line note under the Phase 1 heading clarifying this. Done in 8c33a9d755.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Agreed, worth making explicit. Added to step 5: under decided variant B the mint follows the sweep, so `max_age` doubles as the worst-case crediting latency for a below-margin deposit (credited only once age forces its sweep, chiefly during gas spikes that keep `unswept_value` under `sweep_gas_cost × margin`). Done in 8c33a9d755.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Correct, and it is the same standard-token assumption as today’s helper flow. Added a note to the sweeper delegate section: supported tokens are assumed non-fee-on-transfer and non-rebasing, so `balanceOf(address(this))` equals the amount `depositErc20` transfers. Done in 8c33a9d755.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 A `Cost estimation` section quantifying variant B’s operational cost was added in 3e7dc1913b (tECDSA + outcalls + gas across four scenarios). The crediting-coupled-to-sweeping point you raise is called out today in step 0 (“under variant B, crediting pauses with sweeping”) and in the variant-B cons cell; elevating it to an explicit requirement/non-goal is still open and a good follow-up.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 The cost understanding this decision was waiting on is now in the doc: the `Cost estimation` section (3e7dc1913b) sizes tECDSA signatures, HTTPS outcalls and gas, and notes the self-attestation sub-variant adds one tECDSA signature (≈ $0.036, one-time per address) to each address’ first sweep. So the sub-decision can now be made on cost grounds. It is tracked as the one open sub-decision (Jira DEFI-2918).

mbjorkqvist (approved):

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: See the new cost section.

PR #10657, opened 2026-07-03
fix: remove the system-test child-subreaper

basvandijk:
# What

Remove the child-subreaper / descendant-reaper machinery from system-tests:

- `rs/tests/driver/src/driver/group.rs`: drop the `enable_child_subreaper()` / `spawn_descendant_reaper()` calls and the `kill_all_descendants()` final safety net.
- `rs/tests/driver/src/driver/process.rs`: delete `enable_child_subreaper`, `spawn_descendant_reaper`, `kill_all_descendants` and their now-unused helpers and imports.
- `rs/tests/driver/src/driver/local_backend.rs`: update the comments that referenced the removed machinery (no behavior change).

Without the subreaper, orphaned local-backend daemons (the daemonized `qemu-system` processes and `dnsmasq`) are reparented to the bazel sandbox's init process (PID 1 of the test's PID namespace), which reaps their zombies — so `stop_qemu`'s `/proc/<pid>` polling and `await_qemu_exit` keep working. They are stopped explicitly in `LocalBackend::delete_group`, and anything that outlives the driver is killed when the sandbox's PID namespace is torn down.

# Why

The child-subreaper was not strictly needed anymore but was kept as a safety net. However it's causing flakiness:

`//rs/tests/message_routing:global_reboot_test` flakes on `master` (e.g. on 2026-07-02 and 2026-07-03), always failing attempt 1 during group setup with:

```
thread 'main' (12) panicked at rs/tests/driver/src/driver/test_env.rs:373:18:
Expected ssh-keygen to finish successfully: Os { code: 10, kind: Uncategorized, message: "No child processes" }
```

`code: 10` is `ECHILD`: `wait()` was called on a child that had already been reaped by someone else.

# Root cause

A race between the descendant-reaper thread (introduced in #10253) and the synchronous `ssh-keygen` child that group setup spawns — both living in the same parent driver process:

1. `SystemTestGroup::execute()` calls `enable_child_subreaper()` and `spawn_descendant_reaper()`, then `create_group_setup()`.
2. `create_group_setup()` calls `ssh_keygen()`, which spawns `ssh-keygen` as a synchronous `std::process::Command` child and then calls `.wait()` on it.
3. The descendant-reaper thread reaps, every 200 ms, **any** direct zombie child whose `comm` differs from the driver's own `comm` (the filter was designed to spare only Tokio's task subprocesses, which re-`exec` the driver binary and thus share its `comm`).
4. `ssh-keygen` is a foreign-`comm` direct child, so if it exits and a reaper scan runs before the main thread reaches `.wait()`, the reaper steals its exit status and `.wait()` fails with `ECHILD` → panic.

The failing logs show exactly this: the reaper starts, `ssh-keygen` completes successfully (keys saved, fingerprint printed), yet `.wait()` still reports "No child processes". The race window is small (`ssh-keygen -t ed25519` is fast, the reaper polls every 200 ms), hence the flakiness. Any other synchronous foreign-`comm` child the parent driver spawns and waits on (e.g. the `kill` invocations in `LocalBackend::delete_group`) was exposed to the same race.

---


basvandijk:
Force merging since this fixes flakiness and reduces complexity.

PR #10664, opened 2026-07-06
perf: Compare proper `CanisterIds` as `u64` values

alin-at-dfinity:
If a `CanisterId` represents an actual canister (the majority case; as opposed to a `SubnetId` masquerading as a canister), it is actually a big-endian `u64` value plus a couple of marker bytes underneath. Comparing such `CanisterIds` (both for equality but particularly for ordering) as `u64` values gives speed-ups between 1.25x and 2.8x (with the less common path only slowing down by 6-20%). And because the `u64` is encoded as big-endian, relative ordering remains the same.

While the `CanisterId` struct will grow by 1 byte (from 30 to 31 bytes) (i) that's all the increase (e.g. no extra 7 bytes due to memory alignment) and (ii) my guess is that in the vast majority of cases it actually takes up 32 bytes anyway (due to external alignment constraints).

| Benchmark | `master` | this branch | speedup |
|---|---:|---:|---|
| `cmp` u64 | 3.49 ns | 1.71 ns | **2.04× faster** |
| `binary_search` 1024 | 80.5 ns | 28.4 ns | **2.83× faster** |
| `sort` 1024 | 5.84 µs | 3.79 µs | **1.54× faster** |
| `eq` u64, differ | 2.22 ns | 1.77 ns | **1.25× faster** |
| `eq` u64, equal | 2.22 ns | 1.76 ns | **1.26× faster** |
| `linear_find` (eq scan) | 810 ns | 638 ns | **1.27× faster** |
| `cmp` long/opaque | 3.46 ns | 3.68 ns | ~6% slower |
| `eq` long, differ | 2.46 ns | 2.87 ns | ~17% slower |
| `eq` long, equal | 2.26 ns | 2.70 ns | ~20% slower |
| `hash` one u64 ID | 32.4 ns | 14.1 ns | **2.30× faster** |
|`HashMap` lookup, hit | 35.9 ns | 17.2 ns | **2.10× faster** |
|`HashMap` lookup, miss | 31.0 ns | 16.3 ns | **1.90× faster** |
|`hash` one long/opaque ID | 31.2 ns | 27.5 ns | ~wash (both hash the full principal) |

## Motivation

While after the recent scheduler improvements we perform significantly fewer lookups in the large `BTreeMap` of cold canisters, we still mutate and look up in the (usually) smaller `BTreeMap` of hot canisters and in `SubnetSchedule`. Testing shows 2.1-2.2x improvements in the `RoundSchedule/iteration` and `RoundSchedule/finish_round` scheduler benchmarks. While the absolute numbers are small (single-digit milliseconds) this is because the benchmark assumes few active canisters. With lots of active canisters (e.g. another "Bob" incident), time increase is faster than linear, so a 2x improvement becomes visible.

`scheduler_bench` (`RoundSchedule`, 100k canisters in `BTreeMap`/`BTreeSet<CanisterId>`)

| Benchmark | `master` | branch | speedup |
|---|---|---|---|
| `RoundSchedule/iteration` | 2.23 / 2.14 ms | 0.995 / 0.992 ms | **~2.2× faster** |
| `RoundSchedule/finish_round` | 4.55 / 4.38 ms | 2.12 / 2.12 ms | **~2.1× faster** |

Charging for storage (every 50 rounds) also removes and replaces every single `CanisterState` only to do a simple computation and update one numeric field. No benchmark explicitly covers this, but the time spent, instrumented in production is 50+ ms for 100k canisters. And the increase is similarly faster than linear with more canisters.

Routing (and various canister call validations) rely on routing table lookups (and `CanisterId` comparisons). These are normally a small fraction of the round duration, but that is due to low canister message traffic. With higher (canister controlled) traffic volumes, they can become noticeable. And they are on the critical path (DSM main thread), so they can have material effects under high load. (E.g. any canister can have one message each to arbitrarily many other canisters; the only limit is on the number of callbacks per subnet (millions); if the stream to the respective canisters' subnet is full, Message Routing will try and fail to route each of those messages every single round; that's millions of routing table lookups per round.)

To summarize, the absolute gains are limited. But the relative gains are measurable. And measurable relative gains mean visible absolute gains under high subnet load (large subnets / many active canisters / many canister calls). I.e. (more) graceful degradation.


basvandijk (approved):

michael-weigelt (approved):

rs/types/base_types/src/canister_id.rs
```diff
@@ -4,23 +4,88 @@ use candid::{CandidType, Principal};
...
+}
+
+impl Ord for CanisterId {
+    #[inline(always)]
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.is_u64 && other.is_u64 {
+            as_u64(self.id).cmp(&as_u64(other.id))
+        } else {
+            self.id.cmp(&other.id)
+        }
```
pierugo-dfinity, line 42: One of the conditions that must hold when implementing both `PartialOrd` and `PartialEq` is `a.eq(&b) <==> a.partial_cmp(&b) == Some(Equal)`, though it looks like if `self.is_u64 != other.is_u64`, couldn't we still have `a.partial_cmp(&b) == Some(Equal)` but `a.eq(&b) == False`?

alin-at-dfinity: I don't think we could. `a.is_u64` is equivalent to `a.as_slice().len() == 10 && a.as_slice()[8] == 1 && a.as_slice()[9] == 1`.

So `self.is_u64 != other.is_u64` means that one of them has length 10 and ends in `..., 1, 1]` and one either has a different length or it ends in different bytes. Regardless, `cmp()` would directly compare the slices and they could not possibly be equal.

pierugo-dfinity: I see, thanks!

rs/types/base_types/src/canister_id.rs
```diff
@@ -141,25 +224,62 @@ impl CanisterId {
             )));
         }
 
-        Ok(CanisterId(principal_id))
+        Ok(CanisterId {
+            is_u64: is_canister_id(principal_id),
```
pierugo-dfinity: Isn't this call to `is_canister_id` redundant with the body of the current function? Said differently, is it necessary to have two places in code where we check `length == 10 && slice[8] == 1 && slice[9] == 1`?

alin-at-dfinity: You're right, this is entirely redundant.

Replaced with `true`.

rs/types/base_types/src/canister_id.rs
```diff
@@ -4,23 +4,88 @@ use candid::{CandidType, Principal};
...
+
+impl PartialOrd for CanisterId {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Hash for CanisterId {
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
```
daniel-wong-dfinity-org-twin:
```suggestion
    fn hash<H: Hasher>(&self, hasher: &mut H) {
```

Everything is "state".

rs/types/base_types/src/canister_id.rs
```diff
@@ -4,23 +4,88 @@ use candid::{CandidType, Principal};
...
+    {
+        self.id.serialize(serializer)
+    }
+}
+
+impl CandidType for CanisterId {
+    fn id() -> candid::types::TypeId {
+        candid::types::TypeId::of::<CanisterId>()
+    }
+    fn _ty() -> candid::types::Type {
```
daniel-wong-dfinity-org-twin, line 80: AFAIK, there is ALWAYS a blank line between `fn`s (and this is a good thing beyond consistency).

rs/registry/routing_table/src/lib.rs
```diff
@@ -12,19 +12,9 @@ use std::{collections::BTreeMap, convert::TryFrom};
...
-        bytes.len(),
-        LENGTH + 2,
-        "canister_id: {canister_id}; raw {canister_id:?}"
-    );
-    let mut array = [0; LENGTH];
-    array[..LENGTH].copy_from_slice(&bytes[..LENGTH]);
-    u64::from_be_bytes(array)
+    canister_id
+        .as_u64()
+        .unwrap_or_else(|| panic!("canister_id: {canister_id}; raw {canister_id:?}"))
```
daniel-wong-dfinity-org-twin, line 17: Just an FYI, not an objection.

I think this might add new panic cases (e.g. if as_u64 more aggressively checks the content of bytes).

daniel-wong-dfinity-org-twin (commented):
Ok, but what _actual_ problem does this solve?

This question ☝️ really deserves more explanation in the PR description. Otherwise, we are at risk of violating

> Premature optimization is the root of all evil.
> --Donald Knuth

Ok, so you increased the speed of some operations. Are those ops super common such that there is a meaningful difference _overall_? If so, great. I'm just asking for that to be explained.

I don't think this is a slam dunk, because this replaces thoroughly battle-tested code (derives) with hand-crafted code. And AFAICT, the new code has no tests.

rs/types/base_types/src/canister_id.rs
alin-at-dfinity: Done.

rs/types/base_types/src/canister_id.rs
alin-at-dfinity: Done.

rs/registry/routing_table/src/lib.rs
alin-at-dfinity: I would argue the opposite: `CanisterId::from_u64()` and `CanisterId::from(u64_value)` are implemented by `CanisterId` in `ic_base_types`. Having the inverse conversion implemented somewhere completely unrelated is a bigger potential source of issues.

As is, `CanisterId::as_u64()` does expect bytes 8 and 9 to be `0x01` in addition to `length == 10`, but that's what `CanisterId::from_u64()` always sets them to.

mbjorkqvist (approved):

alin-at-dfinity:
> Ok, but what _actual_ problem does this solve?
> 
> This question ☝️ really deserves more explanation in the PR description.

Done. This is a change I had lying around for a long time. It produced measurable improvements (on the order of 100 ms) in DSM main thread (i.e. critical path) operations. But after various other optimizations, the difference is down in the single digit milliseconds (because the whole DSM overhead is 40 ms and will soon be 20 ms). But this only holds with well-behaved canisters. As soon as you have tens of thousands of canisters with heartbeats/timers or millions of canister-to-canister messages, you go back to the DSM performance from before these optimizations. And a 2x improvement becomes significant. I.e. this is about (somewhat more) graceful degradation.

> And AFAICT, the new code has no tests.

Fair point. Added a bunch of tests covering all the custom implemented comparison and hash code. And both "proper" `CanisterIds` and other kinds of principals wrapped in a `CanisterId`.

rs/types/base_types/src/canister_id.rs
pierugo-dfinity: Couldn't we replace the entire body of `try_from_principal_id` with something like the following?
```rust
if is_canister_id(principal_id) {
    Ok(Self {
        is_u64: true,
        id: principal_id,
    })
} else {
    Err(CanisterIdError::InvalidPrincipalId(...))
}
```
Just to avoid duplicating the logic.

rs/nns/sns-wasm/src/sns_wasm.rs
```diff
@@ -1620,7 +1620,7 @@ where
 
         let canister_status_result = call_response.map_err(|(code, description)| {
             format!(
-                "Could not get the controllers of {target_canister_id:?} \
+                "Could not get the controllers of canister {target_canister_id} \
                     due to an error from the replica. {code:?}:{description:?}"
```
daniel-wong-dfinity-org-twin: align

rs/registry/canister/canbench/canbench_results.yml
daniel-wong-dfinity-org-twin, line 1: Can you share the diff report table? That might help support "this provides a meaningful overall boost".

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/canbench/canbench_results.yml
alin-at-dfinity: I had Claude run the benchmarks, so I don't have the raw output. This is the summary it gave me:
| Workload | Bottleneck | Branch vs master |
|---|---|---|
| **`scheduler` RoundSchedule/iteration** | `CanisterId::cmp` over 100k-entry `BTreeMap`/`BTreeSet` | **~2.2× faster** ✅ |
| **`scheduler` RoundSchedule/finish_round** | same | **~2.1× faster** ✅ |
| `100k_canisters` round | Wasm execution | tie (within noise) |
| `100k_canisters` checkpoint | checkpoint I/O | tie (within noise) |
| `state_manager` hash_tree_new | SHA-256 hashing | tie (within noise) |
| `state_manager` hash_tree_cached | SHA-256 hashing | ~2–3% slower (near noise) |

And the numbers (2 runs on each branch) for `scheduler_bench`:

## `scheduler_bench` (`RoundSchedule`, 100k canisters in `BTreeMap`/`BTreeSet<CanisterId>`)

| Benchmark | `master` | branch | speedup |
|---|---|---|---|
| `RoundSchedule/iteration` | 2.23 / 2.14 ms | 0.995 / 0.992 ms | **~2.2× faster** |
| `RoundSchedule/finish_round` | 4.55 / 4.38 ms | 2.12 / 2.12 ms | **~2.1× faster** |

rs/nns/sns-wasm/src/sns_wasm.rs
alin-at-dfinity: Done.

rs/types/base_types/src/canister_id.rs
alin-at-dfinity: Done.

I had chosen to stick with the detailed error messages (wrong length, wrong class), but you're right: the code duplication is potentially worse than a verbose error message (that is never really expected to be produced anyway).

rs/types/base_types/src/canister_id/tests.rs
```diff
@@ -1,6 +1,159 @@
...
+                "cmp disagrees with principal cmp for {a} vs {b}"
+            );
+            assert_eq!(a.partial_cmp(b), Some(a.cmp(b)));
+
+            if a == b {
+                assert_eq!(
+                    hash_of(a),
+                    hash_of(b),
+                    "equal IDs {a} and {b} have different hashes"
+                );
```
daniel-wong-dfinity-org-twin: I mean, if a != b, it would be super surprising/highly suspicious if your small hand-crafted set of CanisterIds somehow had a collision, right? Let's maybe also assert that the hashes of != CanisterIds are distinct?

rs/types/base_types/src/canister_id.rs
pierugo-dfinity: Cool thanks!

rs/registry/canister/canbench/canbench_results.yml
alin-at-dfinity: And if you meant the canbench reports, here they are, since I've had to redo them. There isn't any meaningful change, as e.g. `get_subnet_for_canister()` will load and parse all routing table shards before even picking the right shard, so any improvement in `CanisterId` comparison is swamped by everything else going on (hence the tens to hundreds of millions of instructions required for 100 lookups). The routing table used in Message Routing is a plain `BTreeMap` and is only constructed once per registry version. So routing performance is much more closely tied to `CanisterId` comparison performance.

```
$ bazel run //rs/registry/canister:registry-canbench_update
[...]
---------------------------------------------------

Summary:
  instructions:
    status:   Regressions and improvements 🔴🟢
    counts:   [total 14 | regressed 1 | improved 1 | new 0 | unchanged 12]
    change:   [max +7.21M | p75 +484.94K | median -2 | p25 -926.61K | min -237.64M]
    change %: [max +2.19% | p75 +0.09% | median -0.00% | p25 -0.51% | min -2.65%]

  heap_increase:
    status:   Improvements detected 🟢
    counts:   [total 14 | regressed 0 | improved 2 | new 0 | unchanged 12]
    change:   [max +6 | p75 0 | median 0 | p25 0 | min -3]
    change %: [max +1.75% | p75 0.00% | median 0.00% | p25 0.00% | min -100.00%]

  stable_memory_increase:
    status:   No significant changes 👍
    counts:   [total 14 | regressed 0 | improved 0 | new 0 | unchanged 14]
    change:   [max 0 | p75 0 | median 0 | p25 0 | min 0]
    change %: [max 0.00% | p75 0.00% | median 0.00% | p25 0.00% | min 0.00%]

---------------------------------------------------

Only significant changes:
| status | name                                         | calls |     ins |  ins Δ% |  HI |    HI Δ% | SMI |  SMI Δ% |
|--------|----------------------------------------------|-------|---------|---------|-----|----------|-----|---------|
|   +    | get_subnet_for_canister_100_times_100        |       |  29.07M |  +2.19% |   0 |    0.00% |   0 |   0.00% |
|   +    | upgrade_with_routing_table_10k::post_upgrade |     1 |   5.44B |  +0.13% | 223 |   +2.76% |   0 |   0.00% |
|   -    | upgrade_with_routing_table_1k::post_upgrade  |     1 | 667.09M |  +0.11% |  22 |  -12.00% |   0 |   0.00% |
|   -    | upgrade_with_routing_table_1k                |       | 669.96M |  +0.11% |  37 |   -7.50% |   0 |   0.00% |
|   -    | migrate_canisters_10_times_1k                |       | 887.18M |  -1.66% |   0 | -100.00% |   0 |   0.00% |
|   -    | migrate_canisters_10_times_10k               |       |   8.74B |  -2.65% |   0 |    0.00% |   0 |   0.00% |

ins = instructions, HI = heap_increase, SMI = stable_memory_increase, Δ% = percent change

---------------------------------------------------

$ bazel run //rs/bitcoin/ckbtc/minter:ckbtc_minter_canbench_update
[...]
---------------------------------------------------

Summary:
  instructions:
    status:   No significant changes 👍
    counts:   [total 7 | regressed 0 | improved 0 | new 0 | unchanged 7]
    change:   [max +2 | p75 +2 | median +2 | p25 +2 | min +2]
    change %: [max 0.00% | p75 0.00% | median 0.00% | p25 0.00% | min 0.00%]

  heap_increase:
    status:   No significant changes 👍
    counts:   [total 7 | regressed 0 | improved 0 | new 0 | unchanged 7]
    change:   [max 0 | p75 0 | median 0 | p25 0 | min 0]
    change %: [max 0.00% | p75 0.00% | median 0.00% | p25 0.00% | min 0.00%]

  stable_memory_increase:
    status:   No significant changes 👍
    counts:   [total 7 | regressed 0 | improved 0 | new 0 | unchanged 7]
    change:   [max 0 | p75 0 | median 0 | p25 0 | min 0]
    change %: [max 0.00% | p75 0.00% | median 0.00% | p25 0.00% | min 0.00%]

---------------------------------------------------

```

rs/types/base_types/src/canister_id/tests.rs
alin-at-dfinity: Done.

pierugo-dfinity (approved):

PR #10670, opened 2026-07-07
test: end-to-end deposit-from-CEX EIP-7702 sweep demo against a local Ethereum node

gregorydemay:
A hermetic, monorepo-native integration test that validates the "deposit from CEX" sweep (design in `rs/ethereum/cketh/docs/deposit_from_cex.md`) end-to-end against a local `anvil` node, through the minter's productive EIP-7702 transaction layer (#10671).

It sweeps unfunded deposit addresses to the minter in minter-paid EIP-7702 transactions, checks each emits the canonical deposit event with the right IC principal, measures gas across batch sizes (1/10/20), and covers both sweeper designs — caller-gated and permissionless-with-minter-attestation — including rejection of unauthorized sweeps.

foundry and solc are vendored via Bazel (no docker, IC replica, or alloy).


mbjorkqvist (approved):
LGTM, thanks @gregorydemay!

MODULE.bazel
```diff
@@ -417,6 +417,71 @@ http_file(
...
+    url = "https://binaries.soliditylang.org/linux-arm64/solc-linux-arm64-v0.8.35+commit.47b9dedd",
+)
+
+http_file(
+    name = "solc_0_8_35_macosx_amd64",
+    downloaded_file_path = "solc",
+    executable = True,
+    sha256 = "6e432580d973cae74c869ef4f737529e73fec5b200dce3c08d5e76c4a6b19159",
+    url = "https://binaries.soliditylang.org/macosx-amd64/solc-macosx-amd64-v0.8.35+commit.47b9dedd",
+)
```
basvandijk, line 447: It would be nice to consolidate on a single `solc` version. Not a blocker, just a suggestion for a follow up.

basvandijk (approved):

MODULE.bazel
gregorydemay: Good idea, I will do a separate PR for that.

PR #10671, opened 2026-07-07
feat: EIP-7702 (SetCode) transaction layer for the ckETH/ckERC20 minter

gregorydemay:
Adds EIP-7702 (type `0x04`, SetCode) support as pure library code: no behavior change and no runtime wiring yet. This unblocks the later sweeping task that pays gas on behalf of unfunded per-account deposit addresses. Authorizations are signed over `keccak256(0x05 || rlp([chain_id, delegate, nonce]))` through the existing threshold-ECDSA path, rejecting `chain_id == 0` to prevent cross-chain replay.

The transaction layer is also refactored so EIP-1559 and EIP-7702 share one abstraction rather than duplicating it: a generic signed-transaction wrapper over a common `SignableTransaction` trait, with the module split into per-EIP submodules. CBOR and RLP encodings are unchanged, so persisted state and on-chain transactions stay byte-compatible — verified against published EIP-7702 test vectors and an authority sign-and-recover round-trip.

Fee-bump/resubmission for EIP-7702 is deferred to DEFI-2926, where it is wired into the send path and its resubmission semantics are decided.


rs/ethereum/cketh/minter/src/tx/tests.rs
```diff
@@ -218,6 +218,174 @@ fn should_cbor_encoding_be_stable() {
...
+    use crate::tx::{
+        AccessList, Authorization, Eip1559Signature, Eip7702TransactionRequest,
+        SignedAuthorization, SignedEip7702TransactionRequest,
+    };
+    use ethnum::u256;
+    use ic_ethereum_types::Address;
+    use std::str::FromStr;
+
+    // Published test vector from the trust-wallet/wallet-core EIP-7702 test suite:
+    // https://github.com/trustwallet/wallet-core/blob/master/rust/tw_evm/src/transaction/transaction_eip7702.rs
```
mbjorkqvist: nit: Would it make sense to link to a specific commit for this file?

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -0,0 +1,188 @@
...
+            return Err(
+                "BUG: EIP-7702 authorization chain_id must be set explicitly and never 0"
+                    .to_string(),
+            );
+        }
+        let hash = self.hash();
+        let key_name = read_state(|s| s.ecdsa_key_name.clone());
+        let signature = crate::management::sign_with_ecdsa(
+            key_name,
+            DerivationPath::new(crate::MAIN_DERIVATION_PATH),
```
mbjorkqvist: This signs the authorization with `MAIN_DERIVATION_PATH` (the empty path → the minter's main-address key), so the tuple recovers to the **main address**. But the whole point of the authorization is to delegate a per-account **deposit EOA**: EIP-7702 installs the delegate code at whichever address `ecrecover` of the tuple yields, so it must be signed with the *deposit address'* derivation path.

The design (`rs/ethereum/cketh/docs/deposit_from_cex.md`, PR #10652) is explicit — this section it implements says at L826–827: "signed over `keccak256(0x05 || rlp([chain_id, delegate, nonce]))` with `sign_with_ecdsa` using **the deposit address' derivation path**", and the primer at L737: "The minter signs (`sign_with_ecdsa`, **`D`'s derivation path**) the authorization tuple". Deposit-address paths are `[SCHEMA, principal, subaccount]` (non-empty, L375), explicitly distinct from `MAIN_DERIVATION_PATH` (L379/L684).

As written, if wired up, `ecrecover(auth)` yields the main address and the sweep of the deposit EOA would do nothing / the delegate code would land on the wrong (main) address. Same hardcoding in `signed.rs::sign` (the tx *sender* should be the dedicated sweeper address, path `[3u8]`, not main — correct for today's EIP-1559 withdrawals, but not for EIP-7702 sweeps).

How about parameterizing the path — have `Authorization::sign` (and `sign<T>`) take a `DerivationPath` argument so the caller passes the deposit-EOA path for authorizations and the sweeper-address path for the sender? Not a live bug (no callers yet, and this method is untested — the round-trip test signs with a local key rather than through `Authorization::sign`), but the current signature can only ever produce an unusable authorization, and the only `TODO` here (DEFI-2926) is about resubmission, not the path.

rs/ethereum/cketh/minter/src/tx/signed.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// the constructor to create the correct value.
+    memoized_hash: Hash,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Decode, Encode)]
+struct InnerSigned<T> {
+    #[n(0)]
+    transaction: T,
+    #[n(1)]
+    signature: Eip1559Signature,
```
mbjorkqvist: nit: now that the signature is embedded generically in `Signed<T>`/`InnerSigned<T>` and reused for EIP-7702 (tests build `SignedEip7702TransactionRequest::from((tx, Eip1559Signature { .. }))`), the name `Eip1559Signature` is misleading. The `(y_parity, r, s)` layout is EIP-2718-generic — consider renaming to `TransactionSignature` (or a type alias) to reflect that it's shared across transaction types.

rs/ethereum/cketh/minter/src/tx/signed.rs
```diff
@@ -0,0 +1,192 @@
...
+        TransactionPrice {
+            gas_limit: self.gas_limit(),
+            max_fee_per_gas: self.max_fee_per_gas(),
+            max_priority_fee_per_gas: self.max_priority_fee_per_gas(),
+        }
+    }
+}
+
+/// Immutable signed transaction.
+/// Use `<request>.sign()` to create a newly signed transaction or
```
mbjorkqvist: nit: stale doc comment — `<request>.sign()` is now the free function `super::sign(request)` (and `withdraw.rs` was updated to call `crate::tx::sign(tx)`); the `.sign()` method no longer exists.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -0,0 +1,188 @@
...
+    #[n(5)]
+    pub destination: Address,
+    #[n(6)]
+    pub amount: Wei,
+    #[cbor(n(7), with = "minicbor::bytes")]
+    pub data: Vec<u8>,
+    #[n(8)]
+    pub access_list: AccessList,
+    #[n(9)]
+    pub authorization_list: Vec<SignedAuthorization>,
```
mbjorkqvist, line 47: nit: a type-`0x04` transaction with an empty `authorization_list` is invalid per EIP-7702. Nothing constructs one yet, but a `debug_assert!`/constructor check that `authorization_list` is non-empty would prevent building an invalid transaction once this is wired.

rs/ethereum/cketh/minter/src/tx/signed.rs
```diff
@@ -0,0 +1,192 @@
...
+            signature,
+        };
+        let hash = Hash(ic_sha3::Keccak256::hash(inner.raw_bytes()));
+        Self {
+            inner,
+            memoized_hash: hash,
+        }
+    }
+
+    pub fn raw_transaction_hex(&self) -> Vec<u8> {
```
mbjorkqvist: nit (pre-existing, carried over): `raw_transaction_hex` returns `Vec<u8>` (raw bytes, not hex); `raw_transaction_hex_string` is the actual hex. Could rename to `raw_transaction_bytes` while this code is being reshuffled.

rs/ethereum/cketh/minter/src/tx/tests.rs
```diff
@@ -218,6 +218,174 @@ fn should_cbor_encoding_be_stable() {
     assert_eq!(decoded_signed_tx, signed_tx);
 }
 
+mod eip7702 {
```
mbjorkqvist, line 222: This module asserts RLP (`raw_transaction_hex_string`, `rlp_bytes`) and the signing hashes, but never CBOR: nothing here exercises `minicbor::encode`/`decode` for the EIP-7702 types. The only CBOR golden-bytes + round-trip test (`should_cbor_encoding_be_stable`, above) covers `SignedEip1559TransactionRequest` only, whereas `Eip7702TransactionRequest` and `SignedAuthorization` derive `Encode`/`Decode` and `Signed<Eip7702…>` has the hand-written CBOR impl in `signed.rs`.

Since these transactions will be persisted in the event-sourced state (design #10652, R8), a CBOR-format regression would break upgrades with nothing to catch it. Suggest adding a golden-bytes + decode round-trip test for `SignedEip7702TransactionRequest` (and ideally `SignedAuthorization`), mirroring the EIP-1559 one.

Related: `Authorization::sign` is currently dead (no callers) and untested — `should_recover_authority_from_signed_authorization` signs with a local `PrivateKey` rather than through it, so its body (incl. the `chain_id == 0` guard) never runs. If it's kept in this PR it's worth a test; see the separate comment on it re: the hardcoded derivation path.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/tx/signed.rs
```diff
@@ -0,0 +1,192 @@
+use super::eip_1559::Eip1559Signature;
+use super::{TransactionPrice, compute_recovery_id, split_in_two};
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in af272f89ec. Renamed `Eip1559Signature` → `TransactionSignature` and moved it out of the `eip_1559` module into the generic signing layer (`tx/signed.rs`), so `Signed<T>` no longer reaches into an EIP-1559-specific module for its signature type. No compatibility alias — this matches alloy, whose `Signed<T>` embeds a transaction-agnostic `alloy_primitives::Signature`.

rs/ethereum/cketh/minter/src/tx/tests.rs
gregorydemay: 🤖 Done in af272f89ec — pinned the reference URL to commit `83823464a621115cf7c06c86079a2597fe46e55b`.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
gregorydemay: 🤖 Done in af272f89ec. `sign<T>` and `Authorization::sign` now take a `DerivationPath` argument; the only current caller (the EIP-1559 withdrawal path in `withdraw.rs`) passes `MAIN_DERIVATION_PATH` explicitly. F3 (DEFI-2921) will supply the per-account deposit-EOA path for authorizations and the sweeper path for the sender when the sweep is wired. End-to-end correctness of the EIP-7702 signing/delegate path is validated by the foundry test in the stacked PoC #10670 (real EVM); the unit tests here pin format/regression (RLP, CBOR) and known-answer recovery.

rs/ethereum/cketh/minter/src/tx/signed.rs
gregorydemay: 🤖 Done in af272f89ec — renamed to `TransactionSignature` and hoisted into the generic layer. See the reply on the related thread.

rs/ethereum/cketh/minter/src/tx/signed.rs
gregorydemay: 🤖 Fixed in af272f89ec — the doc comment now references the free function `sign`.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
gregorydemay: 🤖 Added a `debug_assert!(!authorization_list.is_empty())` in the EIP-7702 RLP encoding chokepoint (`rlp_inner`) in af272f89ec. Note it is debug-only (release canister builds are `--release`), so it catches the mistake in tests but not in production; there is no constructor today (fields are `pub`), so a runtime/constructor-level check is deferred to F3, when a real construction site is introduced.

rs/ethereum/cketh/minter/src/tx/signed.rs
gregorydemay: 🤖 Done in af272f89ec — renamed `raw_transaction_hex` → `raw_transaction_bytes` (kept `raw_transaction_hex_string`).

rs/ethereum/cketh/minter/src/tx/tests.rs
gregorydemay: 🤖 Done in af272f89ec. Added golden-bytes + decode round-trip tests for `SignedEip7702TransactionRequest` and `SignedAuthorization` (mirroring the EIP-1559 CBOR test), guarding the event-sourced state format. `Authorization::sign` is now parameterized (see the derivation-path thread); it still cannot be unit-tested end-to-end because it calls `sign_with_ecdsa` on the management canister, but the crypto is covered by a known-answer authority-recovery vector from viem plus the existing hash/RLP known-answer tests. End-to-end correctness is validated by the foundry test in the stacked PoC #10670.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -158,6 +157,10 @@ impl SignableTransaction for Eip7702TransactionRequest {
     }
 
     fn rlp_inner(&self, rlp: &mut RlpStream) {
+        debug_assert!(
+            !self.authorization_list.is_empty(),
+            "BUG: EIP-7702 transaction must have a non-empty authorization_list"
+        );
```
gregorydemay: Make it a true assert (not a debug)

rs/ethereum/cketh/minter/src/tx/tests.rs
```diff
@@ -384,6 +345,145 @@ mod eip7702 {
 
         assert_eq!(recovered.as_bytes(), authority.as_ref());
     }
+
+    // Known-answer recovery vector reproducing the viem signAuthorization/recoverAuthorizationAddress
+    // fixture: the anvil account 0 key
+    // 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
+    // -> 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 signs the authorization
+    // { chain_id: 1, delegate: wagmi test contract, nonce: 0 } and recovers to that address.
+    // https://github.com/wevm/viem/blob/main/src/utils/authorization/recoverAuthorizationAddress.test.ts
```
gregorydemay: also here pinpoint an exact commit and also where those constants are defined. There are not visible in https://github.com/wevm/viem/blob/349eb2eae6a84d0a7ea9e73b81b51f3fe2f17df8/src/utils/authorization/recoverAuthorizationAddress.test.ts

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -0,0 +1,191 @@
...
+impl SignableTransaction for Eip7702TransactionRequest {
+    fn transaction_type(&self) -> u8 {
+        SET_CODE_TX_ID
+    }
+
+    fn rlp_inner(&self, rlp: &mut RlpStream) {
+        debug_assert!(
+            !self.authorization_list.is_empty(),
+            "BUG: EIP-7702 transaction must have a non-empty authorization_list"
+        );
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in 5bec618153 — changed to a runtime `assert!` in `rlp_inner`, so an invalid type-0x04 transaction (empty `authorization_list`) traps at encode/sign time in release canister builds too, not just in debug.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
gregorydemay: 🤖 Done in 5bec618153 — now a runtime `assert!` instead of `debug_assert!`.

rs/ethereum/cketh/minter/src/tx/tests.rs
gregorydemay: 🤖 Done in 5bec618153. Pinned to viem commit `349eb2eae6a84d0a7ea9e73b81b51f3fe2f17df8` and cited the definitions the fixture imports (they are not inlined in the test): `accounts[0]` privateKey + address in `test/src/constants.ts` (L1-6), and `wagmiContractConfig.address` in `test/src/abis.ts` (L1252-1253).

PR #10676, opened 2026-07-07
fix: add `requires-network` tag to delegation manager unit tests

pierugo-dfinity:
One delegation manager unit test (`load_root_delegation_on_cloud_engine_should_contact_api_bn_test`) tests that a cloud engine nodes correctly contacts an API BN to fetch its delegation. The test ensures that by [asserting](https://github.com/dfinity/ic/blob/82ff6e57713a88cc27aaee485dad6f6041dd566e/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs#L1279) that the node correctly attempted to resolve the API BN's domain. Before `hickory-dns`'s bump in #10645, we just asserted that the domain resolution failed, which would get triggered by a timeout (which justified the [manually set](https://github.com/dfinity/ic/blob/82ff6e57713a88cc27aaee485dad6f6041dd566e/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs#L467-L473) timeout/attempts options). #10645 changed that to `NoRecordsFound`, meaning the resolution should not timeout, but return `NXDOMAIN`. This works on CI because its execution environment resolves the domain, but not locally because Bazel's sandbox doesn't have access to the network. So the request times out and thus does not match `NoRecordsFound`.

This PR adds the `requires-network` tag to the Bazel target, which makes the test pass locally. We can now also remove the manual exception on the resolver options because we do not need to force a timeout.


eichhorl (approved):

basvandijk (approved):
Draft PR https://github.com/dfinity/ic/pull/10693 explains why there's a difference between local and CI runs. `requires-network` would make the test non-hermetic which is unfortunate. The other PR makes the test hermetic by using a stub DNS resolver. However that increases the complexity so I think this is the lesser of two evils.

pierugo-dfinity:
> I think this is the lesser of two evils.

Thanks for the compromise.

PR #10678, opened 2026-07-07
fix: Validate in-memory size of ingress payload

eichhorl:
Due to the hashes-in-blocks feature, ingress payloads have both an in-memory and an on-wire size limit (if hashes-in-blocks is disabled, both are the same). 

During payload building, both limits are enforced while selecting new ingress messages to be included:
https://github.com/dfinity/ic/blob/0f0a8d48219d90b10c2aad869217f96926d4b149/rs/ingress_manager/src/ingress_selector.rs#L210-L220

With this PR, we also enforce the in-memory limit during payload validation.

### DIFF
--- a/rs/ingress_manager/src/ingress_selector.rs
+++ b/rs/ingress_manager/src/ingress_selector.rs
@@ -104,9 +104,18 @@ impl IngressSelector for IngressManager {
         let max_expiry = context.time + MAX_INGRESS_TTL;
         let expiry_range = min_expiry..=max_expiry;
 
-        let settings = self
-            .get_ingress_message_settings(context.registry_version)
-            .expect("Couldn't fetch ingress message parameters from the registry.");
+        let settings = match self.get_ingress_message_settings(context.registry_version) {
+            Ok(settings) => settings,
+            Err(err) => {
+                warn!(
+                    every_n_seconds => 5,
+                    self.log,
+                    "Failed to get the ingress message settings from the registry: {err}"
+                );
+
+                return PayloadWithSizeEstimate::default();
+            }
+        };
 
         // Select valid ingress messages and stop once the total size
         // becomes greater than byte_limit.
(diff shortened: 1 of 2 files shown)


rs/ingress_manager/src/ingress_selector.rs
```diff
@@ -379,6 +379,16 @@ impl IngressSelector for IngressManager {
...
 
+        let size_estimates = self.payload_size_estimates(payload);
+        if size_estimates.memory.get() > settings.max_ingress_bytes_per_block as u64 {
+            return Err(ValidationError::InvalidArtifact(
+                InvalidIngressPayloadReason::IngressPayloadTooBig(
+                    size_estimates.memory.get() as usize,
+                    settings.max_ingress_bytes_per_block,
+                ),
+            ));
+        }
```
pierugo-dfinity, line 416: Should we instead call `self.memory_byte_limit` to mirror payload building?

rs/ingress_manager/src/ingress_selector.rs
pierugo-dfinity: Sorry, not your change. Maybe we could return an error here instead of panicking

pierugo-dfinity (commented):
Nice catch!

rs/ingress_manager/src/ingress_selector.rs
eichhorl: Good point, there was a second occurrence in `get_ingress_payload`, where we can return an empty payload instead: https://github.com/dfinity/ic/pull/10678/commits/f41788c4845ad60f2a660d27fe46164be9dc0f3b

rs/ingress_manager/src/ingress_selector.rs
eichhorl: We would have to come up with a fake `wire_limit`, so I think we shouldn't.

pierugo-dfinity: Right. I was thinking of `size_estimates.wire`, but dove deeper and realized it did not help.

pierugo-dfinity (approved):

PR #10680, opened 2026-07-07
fix: deflake //rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test

basvandijk:
## Summary

`//rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test` was flaky. Over the past week it flaked in 7 CI runs, with three subtests failing intermittently:

- `load_root_delegation_times_out_on_send_request_test`
- `load_root_delegation_times_out_on_receive_body_test`
- `load_root_delegation_on_cloud_engine_should_contact_api_bn_test`

## Root cause

In test mode `CONNECTION_TIMEOUT`, `NNS_DELEGATION_REQUEST_SEND_TIMEOUT` and `NNS_DELEGATION_BODY_RECEIVE_TIMEOUT` were all set to `1s`. `try_fetch_delegation_from_nns` runs three sequential, wall-clock-timed stages, and the **connect** stage runs first with the same `1s` budget as the stage each test actually targets:

1. `timeout(CONNECTION_TIMEOUT, connect(...))` — a real TCP + TLS handshake to an in-process mock server (or a DNS lookup for the API BN case).
2. `timeout(NNS_DELEGATION_REQUEST_SEND_TIMEOUT, send_request(...))`
3. `timeout(NNS_DELEGATION_BODY_RECEIVE_TIMEOUT, collect body)`

The Rust test harness runs the tests in parallel (one thread per test), and each test drives the client and the spawned mock server on the same `current_thread` Tokio runtime. Under CPU contention on CI, the connect stage exceeds `1s`, so the connection timeout fires and returns `"Timed out while connecting to the node after 1s"` — pre-empting and masking the send/receive/DNS error each test asserts on. In every observed CI failure the actual error was this connection-timeout string.

## Fix

- Use the same `10s` `CONNECTION_TIMEOUT` in production and tests (removed the `#[cfg(test)]` `1s` override). A generous connect budget means a slow TLS handshake no longer pre-empts the send/receive/DNS errors the tests actually assert on, while the `1s` send/receive timeouts those tests target are unchanged.
- Limit the test binary's concurrency (`RUST_TEST_THREADS=4` and `tags = ["cpu:4"]`) so connection setup isn't starved of CPU and completes within the timeout. This directly addresses the CPU-contention root cause.

## Verification

- `rustfmt` and `cargo clippy` (incl. `--tests`) are clean.
- Stress-ran `bazel test --runs_per_test=10 --jobs=10 --nocache_test_results //rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test`; with the connect stage no longer sharing a `1s` budget it stops timing out spuriously.
- Downstream `//rs/http_endpoints/public:*` and `//rs/replica:*` tests pass.

Created following the steps in `.claude/skills/fix-flaky-tests/SKILL.md`.

### DIFF
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -1270,19 +1270,8 @@ mod tests {
         )
         .await;
 
-        // Since the API BN is configured with a domain that does not resolve, we expect the
-        // connection to fail with a name resolution error, which indicates that we indeed tried to
-        // connect to the API BN instead of an NNS node. Depending on the environment the resolver
-        // either reports `NoRecordsFound` (a DNS server answered that the domain does not exist) or
-        // `Timeout` (no DNS server answered within the resolver timeout); both prove that we went
-        // through the API BN DNS-resolution path rather than connecting to an NNS node by IP.
-        assert_matches!(
-            response,
-            Err(err) if {
-                let err = format!("{err:?}");
-                err.contains("NoRecordsFound") || err.contains("Timeout")
-            }
-        );
+        // connect to the API BN instead of an NNS node.
+        assert_matches!(response, Err(err) if format!("{err:?}").contains("NoRecordsFound"));
     }
 
     #[tokio::test]
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -59,11 +59,6 @@ const DELEGATION_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
 
 const DELEGATION_RETRY_MAX_BACKOFF_SECONDS: u64 = 15;
 
-// Unlike the send/receive timeouts below (which we deliberately lower in test mode), the
-// connection timeout uses the same value in production and in tests. Establishing the TLS
-// connection to the in-process mock server can take a while under load, and a low connection
-// timeout would spuriously fire before the connection is established and mask the
-// send/receive/DNS-resolution errors that the tests assert on.
 const CONNECTION_TIMEOUT: Duration = Duration::from_secs(10);
 
 #[cfg(not(test))]
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -1270,6 +1270,8 @@ mod tests {
         )
         .await;
 
+        // Since the API BN is configured with a domain that does not resolve, we expect the
+        // connection to fail with a name resolution error, which indicates that we indeed tried to
         // connect to the API BN instead of an NNS node.
         assert_matches!(response, Err(err) if format!("{err:?}").contains("NoRecordsFound"));
     }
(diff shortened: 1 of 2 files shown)


rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -59,10 +59,15 @@ const DELEGATION_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
...
 
-#[cfg(not(test))]
+// Unlike the send/receive timeouts below (which we deliberately lower in test mode), the
+// connection timeout uses the same value in production and in tests. Establishing the TLS
+// connection to the in-process mock server can take a while under load, and a low connection
+// timeout would spuriously fire before the connection is established and mask the
+// send/receive/DNS-resolution errors that the tests assert on. To keep this timeout meaningful in
+// tests, the concurrency of the test binary is limited (see `RUST_TEST_THREADS` and the `cpu:4`
+// tag in `BUILD.bazel`), and the test that specifically exercises the connection timeout passes
+// its own short timeout instead.
```
pierugo-dfinity:
```suggestion
// send/receive/DNS-resolution errors that the tests assert on.
```
The end of the comment doesn't look super relevant

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -245,6 +251,7 @@ async fn load_root_delegation(
 /// Returns a BoxError if any step of the process fails.
 async fn try_fetch_delegation_from_nns(
     config: &Config,
+    connection_timeout: Duration,
```
pierugo-dfinity: I think I'd prefer not adding this new argument just for the sake of making one test happy. The whole unit test suite already takes > 10s to complete, so I'd argue that making that single test take 10s instead of 1s is not a big deal if it avoids us passing `CONNECTION_TIMEOUT` at every call-site.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -1275,8 +1283,17 @@ mod tests {
...
+        // either reports `NoRecordsFound` (a DNS server answered that the domain does not exist) or
+        // `Timeout` (no DNS server answered within the resolver timeout); both prove that we went
+        // through the API BN DNS-resolution path rather than connecting to an NNS node by IP.
+        assert_matches!(
+            response,
+            Err(err) if {
+                let err = format!("{err:?}");
+                err.contains("NoRecordsFound") || err.contains("Timeout")
+            }
+        );
```
pierugo-dfinity: I think this got affected by the recent https://github.com/dfinity/ic/pull/10645. Please let me know what you think of the following alternative fix: https://github.com/dfinity/ic/pull/10676.

pierugo-dfinity (commented):
I think limiting the number of threads to 4 and keeping the timeout to 10s (without the exception for one single test) should be helpful. For the API BN case, see my comment.

Thanks for looking into it!

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
basvandijk: That seems like a better fix. Although it still feels rather hacky to me. Let me comment on that PR. But I'll remove this change here.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
basvandijk: Done.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: > Let me comment on that PR.

Yes please do, that's why I had asked for Nicolas' review, but he looks out for the moment.

pierugo-dfinity (approved):

PR #10683, opened 2026-07-07
feat(registry): allow deleting non-System subnets

mraszyk:
Lift the CloudEngine-only restriction in `do_delete_subnet` now that the deterministic state machine handles deletion of any remote subnet.

Authorization by subnet type (the caller is already restricted to governance or the engine controller at the endpoint):
- System subnets (e.g. the NNS) may never be deleted.
- The engine controller may only delete CloudEngine subnets.
- Governance may delete any non-System subnet.

Adds/updates registry tests covering governance deleting an Application subnet, the engine controller being refused, and System subnets remaining undeletable.

### DIFF
--- a/rs/registry/canister/src/mutations/do_delete_subnet.rs
+++ b/rs/registry/canister/src/mutations/do_delete_subnet.rs
@@ -1,6 +1,7 @@
 use candid::{CandidType, Principal};
 #[cfg(target_arch = "wasm32")]
 use dfn_core::println;
+use ic_nns_constants::ENGINE_CONTROLLER_CANISTER_ID;
 use ic_protobuf::registry::subnet::v1::{SubnetListRecord, SubnetType};
 use ic_registry_keys::{
     make_catch_up_package_contents_key, make_crypto_threshold_signing_pubkey_key,
@@ -14,10 +15,6 @@ use serde::{Deserialize, Serialize};
 use crate::{common::LOG_PREFIX, registry::Registry};
 
 impl Registry {
-    /// Note: The method name implies general subnet deletion. Currently, however, only CloudEngine subnets
-    ///       can be deleted. The reason is that general subnet deletion requires changes in the deterministic
-    ///       state machine, which are planned in the near future.  
-    ///
     /// Deleting a subnet means to:
     /// - Remove its subnet ID from the key `subnet_list`.
     /// - Remove its subnet record.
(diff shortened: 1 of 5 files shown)


rs/registry/canister/src/mutations/do_delete_subnet.rs
```diff
@@ -29,16 +26,34 @@ impl Registry {
     /// consumers that must take deleted subnets into account do so via old registry versions (the
     /// registry client method `get_versioned_value` allows the caller to distinguish between deleted
     /// and non-existing values via `version ?= 0`).
-    pub fn do_delete_subnet(&mut self, payload: DeleteSubnetPayload) -> Result<(), String> {
-        println!("{LOG_PREFIX}do_delete_subnet: {payload:?}");
+    ///
+    /// Authorization by subnet type (the calling principal is already restricted to governance or the
+    /// engine controller at the endpoint):
+    /// - System subnets (e.g. the NNS) may never be deleted.
```
daniel-wong-dfinity-org-twin: I guess this follows from the next two, which is fine, but maybe, it would make sense to change the order of these to 

1. Governance can delete any non-system subnets, including engines.
2. The controller of an engine can delete the engine.
3. A corollary of the above is that nobody can delete a system subnet.

rs/registry/canister/src/mutations/do_delete_subnet.rs
```diff
@@ -29,16 +26,34 @@ impl Registry {
...
 
         let DeleteSubnetPayload { subnet_id } = payload;
         let subnet_id_ = SubnetId::from(PrincipalId::from(subnet_id));
         let subnet_record = self.get_subnet(subnet_id_, self.latest_version())?;
 
-        // Currently, only CloudEngines can be deleted.
-        if subnet_record.subnet_type != i32::from(SubnetType::CloudEngine) {
-            return Err("Only CloudEngines may be deleted".to_string());
+        // System subnets (e.g. the NNS) may never be deleted.
+        if subnet_record.subnet_type == i32::from(SubnetType::System) {
```
daniel-wong-dfinity-org-twin: I know the code was already like this, but I think we do `E as i32` in the (vast) majority of cases.

rs/registry/canister/tests/delete_subnet.rs
```diff
@@ -48,6 +50,50 @@ async fn setup_with_existing_subnet() -> (PocketIc, Principal) {
...
+        .update_call(
+            REGISTRY_CANISTER_ID.get().0,
+            caller.0,
+            "delete_subnet",
+            Encode!(&payload).unwrap(),
+        )
+        .await
+        .unwrap_or_else(|err| {
+            panic!("delete_subnet call by {caller} was unexpectedly rejected: {err:?}")
+        });
```
daniel-wong-dfinity-org-twin: Do as you see fit.

unwrap would be acceptable.

daniel-wong-dfinity-org-twin (approved):

PR #10685, opened 2026-07-07
feat: per-account deposit-address derivation for the ckETH/ckERC20 minter

gregorydemay:
Each IC account gets a unique, deterministic Ethereum deposit address derived from the minter's threshold-ECDSA key, so a plain exchange withdrawal to that address can later be attributed and swept. The addresses are per-account and per-asset (separate ckERC20 and ckETH schemas), never collide with the minter's main address, and are signable by the minter — each address matches the derivation path the minter later signs with, so it is exactly what the minter can sweep from. Pure library code; registration, detection, and sweeping arrive in later tickets (F3 / DEFI-2921 onward).

Also in this PR:
- a design-doc subsection consolidating the addressing model — the derivation tree (main, per-asset deposit, and sweeper addresses as independent siblings), which key signs which part of a sweep, and why Ethereum nonces stay per-address;
- a behaviour-preserving consolidation of a duplicated principal-encoding helper, which now also rejects the empty principal (an input the minter can never receive).

Satisfies R1: determinism, uniqueness across principals / subaccounts / asset schemas, non-collision with the minter's main address, and EIP-55 encoding.


rs/ethereum/cketh/minter/src/deposit_address.rs
```diff
@@ -0,0 +1,96 @@
...
+    schema: DepositAddressSchema,
+    owner: &Principal,
+    subaccount: Option<&LedgerSubaccount>,
+) -> Vec<ByteBuf> {
+    vec![
+        ByteBuf::from(vec![schema.tag()]),
+        ByteBuf::from(principal_to_bytes32(owner).to_vec()),
+        ByteBuf::from(subaccount_to_bytes32(subaccount).to_vec()),
+    ]
+}
```
gregorydemay: Shouldn't the sweeper be used as prefixed in the derivation path so that the sweeper's address is used for the transactions instead of the minter's? (see spec)

gregorydemay: 🤖 The sweeper *is* used as the sender for sweep transactions (path `[3]`, per R17) — but the deposit EOAs are independent siblings of it in the derivation tree, not nested beneath it, and they do not need to be: the minter holds the master key and signs for every sibling independently. A sweep uses two independent signatures on two unrelated paths — the deposit-EOA path `[1|2, p, s]` signs the authorization tuple, the sweeper path `[3]` signs the type-0x04 transaction. Documented this explicitly in 29bdbd4d47 (new "Address derivation tree, signing, and nonces" subsection in the design doc, merged into this PR): the tree, who signs what, and the per-address nonce lanes.

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
```diff
@@ -244,6 +244,22 @@ pub enum EventSourceError {
...
+/// - the remaining bytes are zero
+///
+/// This is the inverse of [`parse_principal_from_slice`].
+pub fn principal_to_bytes32(principal: &Principal) -> [u8; 32] {
+    let bytes = principal.as_slice();
+    let n = bytes.len();
+    assert!(n <= 29, "BUG: principal is longer than 29 bytes: {n}");
+    let mut fixed = [0_u8; 32];
+    fixed[0] = n as u8;
+    fixed[1..=n].copy_from_slice(bytes);
```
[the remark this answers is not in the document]
gregorydemay: 🤖 This is a false positive: `fixed[1..=n]` does not panic when `n == 0`. The inclusive range `1..=0` normalizes to the empty range `1..1`, a valid empty subslice, and `copy_from_slice(&[])` is then a no-op (verified with a standalone program). In addition `n == 0` only occurs for the management-canister principal, which is never a deposit owner — `parse_principal_from_slice` already rejects it on decode. This is also pre-existing code, moved unchanged in this PR, so leaving it as-is.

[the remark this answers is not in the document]
gregorydemay: 🤖 Follow-up: rather than tolerate the empty principal, we now assert it cannot occur (389731d021). `principal_to_bytes32` asserts the principal is non-empty, since the only zero-length principal is the management canister, which can never call the minter and is therefore never a deposit owner; added a `#[should_panic]` test locking the invariant.

rs/ethereum/cketh/minter/bin/principal_to_hex.rs
```diff
@@ -22,10 +22,6 @@ fn main() {
             std::process::exit(1)
         }
     };
-    let n = principal.as_slice().len();
-    assert!(n <= 29);
-    let mut fixed_bytes = [0_u8; 32];
-    fixed_bytes[0] = n as u8;
-    fixed_bytes[1..=n].copy_from_slice(principal.as_slice());
+    let fixed_bytes = ic_cketh_minter::eth_logs::principal_to_bytes32(&principal);
     println!("0x{}", hex::encode(fixed_bytes));
```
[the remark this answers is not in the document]
gregorydemay: no the management canister will never call this.

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
```diff
@@ -244,6 +244,26 @@ pub enum EventSourceError {
...
+/// - the next N bytes are the principal
+/// - the remaining bytes are zero
+///
+/// This is the inverse of [`parse_principal_from_slice`].
+pub fn principal_to_bytes32(principal: &Principal) -> [u8; 32] {
+    let bytes = principal.as_slice();
+    let n = bytes.len();
+    assert!(
+        !bytes.is_empty(),
+        "BUG: principal is empty (the management canister principal cannot call the minter)"
```
mbjorkqvist: It's not clear to me how this will be called - in [the design](https://github.com/dfinity/ic/blob/389731d021c2d567754807603d6de94233d0899b/rs/ethereum/cketh/docs/deposit_from_cex.md#step-1-retrieve-the-deposit-address) we have

> Endpoint `deposit_erc20(account, fee?) -> String`

but it's not clear to me if the `account` is the principal of the caller, or if it's possible to pass in any `account`, in which case, someone could call `deposit_erc20` and pass in the management canister principal? In that case I don't consider this as a bug (or maybe a call at this layer *is* a bug, and it should have been rejected at a higher layer already).

rs/ethereum/cketh/minter/src/deposit_address/tests.rs
```diff
@@ -0,0 +1,224 @@
...
+    for checksummed in [
+        "0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed",
+        "0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359",
+        "0xdbF03B407c01E7cD3CBea99509d93f8DDDC8C6FB",
+        "0xD1220A0cf47c7B9Be7A2E6BA89F429762e7b9aDb",
+    ] {
+        let parsed = Address::from_str(&checksummed.to_lowercase()).unwrap();
+        assert_eq!(parsed.to_string(), checksummed);
+    }
+}
```
mbjorkqvist: nit: This doesn't seem to test anything related to this PR?

mbjorkqvist (approved):
Thanks @gregorydemay, just a couple of minor questions/comments!

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -188,6 +188,7 @@ rust_binary(
     srcs = ["bin/principal_to_hex.rs"],
     deps = [
         # Keep sorted.
+        ":minter",
```
gregorydemay: undo the changes to principal_to_hex, this is meant as a small utility that can be run independently of the minter.

rs/ethereum/cketh/minter/src/deposit_address/tests.rs
gregorydemay: you're right, sorry about that. It seems I forgot to review those tests, I thought I did.

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
gregorydemay: you're right that we can address derivation for the caller (minter sponsors gas) or for a provided account (e.g. caller sponsors gas). But here using principal_to_bytes32 doesn't make sense to me. Key derivation path does not need to be align on 32 bytes.

rs/ethereum/cketh/minter/src/deposit_address.rs
```diff
@@ -0,0 +1,96 @@
...
+/// [`crate::MAIN_DERIVATION_PATH`] used for the minter's main address.
+pub fn deposit_derivation_path(
+    schema: DepositAddressSchema,
+    owner: &Principal,
+    subaccount: Option<&LedgerSubaccount>,
+) -> Vec<ByteBuf> {
+    vec![
+        ByteBuf::from(vec![schema.tag()]),
+        ByteBuf::from(principal_to_bytes32(owner).to_vec()),
+        ByteBuf::from(subaccount_to_bytes32(subaccount).to_vec()),
```
gregorydemay: The derivation path elements don't need to be align on 32 bytes, no? there is no need for principal_to_bytes32/subaccount_to_bytes32 we could just use something like ckBTC: https://github.com/dfinity/ic/blob/b608c374f23d4ae68237d9fa14dc60240b16565c/rs/bitcoin/ckbtc/minter/src/address.rs#L92

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
gregorydemay: 🤖 Addressed. The derivation path no longer uses `principal_to_bytes32`: `deposit_address`/`deposit_derivation_path` now take a single `&Account` (like the ckBTC minter) and encode the owner as its variable-length principal bytes plus `effective_subaccount()` — no 32-byte alignment, and no empty-principal edge case in the path (78338c12bb, 177cf9a618).

On your question about `deposit_erc20`: whether `account` is the caller or a caller-provided account (minter- vs caller-sponsored gas) is a design choice for a later ticket; either way this library layer just derives from the given account, and validation of who may be an owner belongs at the endpoint layer. Since `principal_to_bytes32` had no remaining productive caller after this, I removed it entirely and reverted `eth_logs` to its prior state.

rs/ethereum/cketh/minter/src/deposit_address/tests.rs
gregorydemay: 🤖 Removed this unrelated EIP-55 reference-vector test (177cf9a618).

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: 🤖 Done — reverted the `principal_to_hex` BUILD change (dropped the `:minter` dep) and restored its inline encoder, so the utility stays independent of the minter (177cf9a618).

rs/ethereum/cketh/minter/src/deposit_address.rs
gregorydemay: 🤖 Done — switched to the ckBTC-style encoding: the path now takes an `&Account` and uses variable-length owner bytes + `effective_subaccount()`, so `principal_to_bytes32`/`subaccount_to_bytes32` are gone (78338c12bb, 177cf9a618).

PR #10687, opened 2026-07-07
fix: account for fetch_canister_logs instructions

mraszyk:
This PR fixes accounting for instructions executed by the `fetch_canister_logs` endpoint of the management canister invoked via an inter-canister call:
- benchmarks are fixed to fully fill the buffer with data, new benchmarks with a filter (producing no record, a single record, and keeping all records) and with a maximum response size of 2MB are added, and the benchmark is now focused on `fetch_canister_logs_response` (other parts of the stack are metered separately);
- instructions are computed as a linear approximation over the record count and total content length with factors derived from the benchmarks;
- instructions are deducted from the round instruction budget (`round_limits`);
- the call fee is dropped since it is negligible compared to the update call base fee (for the caller's response callback) and xnet per-byte fee;
- new tests:
  - checking that all attached cycles are refunded;
  - checking that the result is trimmed to be within the maximum result size and its Candid-encoded size does not exceed the maximum result size allowing for up to 4KiB overhead due to Candid.


rs/execution_environment/src/canister_logs.rs
```diff
@@ -66,3 +150,35 @@ pub(crate) fn check_log_visibility_permission(
...
+        // 150_000 + 900 × record_count + 1 × content_size instructions.
+        assert_eq!(
+            fetch_canister_logs_instructions(50_000, NumBytes::new(0)),
+            NumInstructions::new(150_000 + 900 * 50_000)
+        );
+        assert_eq!(
+            fetch_canister_logs_instructions(10, NumBytes::new(4_096)),
+            NumInstructions::new(150_000 + 900 * 10 + 4_096)
+        );
+        // Monotonically non-decreasing in both the record count and content size.
```
mraszyk:
```suggestion
        // Monotonically increasing in both the record count and content size.
```

rs/replicated_state/src/canister_state/system_state/log_memory_store/tests/log_memory_store.rs
```diff
@@ -469,6 +470,76 @@ fn max_response_size_respected_with_filtering_by_timestamp() {
...
+            "content_len={content_len}: encoded response of {encoded_len} bytes exceeds \
+             the result cap plus Candid framing ({max_response})"
+        );
+
+        // The last (largest) content case is a single record whose `data_size()` equals
+        // `RESULT_MAX_SIZE` exactly (content = `RESULT_MAX_SIZE` minus the fixed per-record
+        // overhead), so the trim keeps precisely that one record and the returned size hits
+        // the cap exactly — not merely below it. This pins the trim boundary: an off-by-one
+        // that dropped the boundary record (`>` vs `>=`) would leave `returned_size` short
+        // of the cap and fail here. And because the stored per-record overhead (40 B)
```
mraszyk:
```suggestion
        // of the cap and fail here. And although the stored per-record overhead (40 B)
```

rs/execution_environment/benches/management_canister/canister_logging.rs
```diff
@@ -133,41 +162,202 @@ fn run_bench_resize_canister_log<M: criterion::measurement::Measurement>(
...
-                );
-                assert!(result.is_ok());
+            || {
+                let mut request = FetchCanisterLogsRequest::new(target);
+                request.filter = filter;
+                request
+            },
+            |request| fetch_canister_logs_response_for_bench(sender, canister, request),
+            // The response (up to ~2 MiB) is dropped outside the timed region,
+            // with bounded memory across the batch.
```
mraszyk:
```suggestion
```
looks misplaced here

rs/execution_environment/src/canister_logs.rs
```diff
@@ -42,17 +39,104 @@ pub(crate) fn fetch_canister_logs(
     })
 }
 
+/// Derives the number of round instructions to deduct for executing a
```
michael-weigelt: IMO this wordy explanation adds little. Comments have to be maintained too, after all. A 2 line explanation to each constant is certainly good, and a hint on how to re-run the benchmark would be great (which is already hidden inside this wall of text). But every mention of microsecond duration and all that reasoning..
In case you see value in it, perhaps it could go in the PR description instead?

michael-weigelt (approved):

PR #10692, opened 2026-07-08
feat: Free subnets can spend more than their allowance on HTTPS outcalls

eichhorl:
Free subnets may not attach cycles to their HTTP outcalls. Therefore, the per-replica allowance of the outcall is zero.

In order for HTTP outcalls to still be functional on free subnets under the new pricing, this PR allows nodes on free subnets to spend more cycles than their per-replica allowance. The amount of cycles that were spent are gossiped as part of the metadata shares, just like they are for normal subnets. This to enable accurate cycle Accounting metric.

However, the check ensuring that the spent amount is smaller than the allowance is disabled. Still, the spent cycles must be smaller than a newly defined constant (1 trillion), even on free subnets.

To do this, we need to consider they subnet's cycle cost schedule during share and block payload validation. For now, we use the cycle cost schedule found in the latest replicated state. In the future, we will switch to the cycle cost schedule that is stored in the request's context (see https://github.com/dfinity/ic/pull/10712).


rs/types/types/src/canister_http.rs
```diff
@@ -1091,6 +1093,53 @@ impl CountBytes for CanisterHttpPaymentReceipt {
     }
 }
 
+/// The maximum cycles a single replica may report having spent on an HTTP
+/// outcall when the subnet uses a [`CanisterCyclesCostSchedule::Free`] schedule.
+///
+/// Free subnets charge nothing, so the reported spend (used only for cost
+/// accounting) is not bounded by the per-replica allowance. To keep it from
+/// being arbitrarily large, it is instead bounded by this constant, which is a
+/// above the largest per-replica cost the pay-as-you-go tracker can ever compute
```
schneiderstefan:
```suggestion
/// being arbitrarily large, it is instead bounded by this constant, which is
/// above the largest per-replica cost the pay-as-you-go tracker can ever compute
```

schneiderstefan (approved):

PR #10693, opened 2026-07-08
fix: make DNS resolution hermetic in nns_delegation_manager unit tests

basvandijk:
## Summary

`load_root_delegation_on_cloud_engine_should_contact_api_bn_test` asserts that resolving the non-existing API BN domain (`domain.invalid`) fails with `NoRecordsFound` (NXDOMAIN). The resolver uses the system DNS configuration (`/etc/resolv.conf`), so the observed error depends on the environment: the test failed locally while passing on CI.

## Root cause

Bazel runs this test in a network-isolated sandbox **both** locally and on CI (`--nosandbox_default_allow_network` in `bazel/conf/.bazelrc.build`); only loopback is available inside the sandbox's network namespace. What differs is the container's `/etc/resolv.conf`, which is generated by the container engine based on the network mode:

- The local dev container runs with `--network=host` (`ci/container/container-run.sh`) and inherits the host's `resolv.conf` — e.g. a systemd-resolved stub at `127.0.0.53`, which is bound only in the host netns and hence unreachable from the sandbox netns → the lookup **times out**, failing the `NoRecordsFound` assert.
- CI's job container (`ci-main.yml` container options) is attached to a Docker bridge network and resolves via Docker's embedded DNS → the lookup returns **NXDOMAIN**, passing the assert.

So the test outcome depended on the environment's resolver setup rather than on the code under test.

#10676 proposes opting the test target out of network isolation with the `requires-network` tag. That makes the test pass locally today, but keeps it environment-dependent: the strict `NoRecordsFound` assertion then relies on real-world DNS honestly answering NXDOMAIN (e.g. ISP/VPN resolvers that hijack NXDOMAIN would break it), and it removes sandbox network isolation from the whole test target.

## Fix

Make the resolution hermetic instead:

- The test starts a minimal stub DNS server on `127.0.0.1` which answers `NXDOMAIN` to every query, mimicking how a real nameserver answers for a non-existing domain.
- In unit tests (`#[cfg(test)]`), `connect()` points the hickory resolver at that stub instead of the system configuration. Loopback is available inside Bazel's sandbox (the other tests in this file already bind mock TLS servers on `127.0.0.1`), so resolution now behaves identically in every environment.
- The test-only fast-fail resolver options (100ms timeout, 1 attempt) are no longer needed and are removed: the stub answers immediately.

## Verification

- On clean master the target fails locally in the default (network-isolated) sandbox; with this change it passes.
- Stress run: `bazel test --runs_per_test=10 --jobs=10 --nocache_test_results //rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test` → 10/10 pass.
- `rustfmt` and `cargo clippy` (incl. `--tests`) are clean.
- Directly affected downstream tests pass: `//rs/http_endpoints/public:*` and `//rs/replica:*`.

### DIFF
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -423,6 +423,11 @@ async fn try_fetch_delegation_from_nns(
     Ok(nns_delegation_builder)
 }
 
+/// Port of the stub DNS server which makes DNS resolution hermetic in unit tests, see
+/// `tests::start_nxdomain_dns_server`.
+#[cfg(test)]
+static TEST_DNS_SERVER_PORT: std::sync::OnceLock<u16> = std::sync::OnceLock::new();
+
 async fn connect(
     log: ReplicaLogger,
     rt_handle: &tokio::runtime::Handle,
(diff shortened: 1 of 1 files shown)


basvandijk:
Closed in favour of https://github.com/dfinity/ic/pull/10676.

PR #10694, opened 2026-07-08
feat: introduce `BackwardsCompatible`

pierugo-dfinity:
This PR introduces a new `BackwardsCompatible` type, intended to be used when introducing new fields to structs that cross upgrade boundaries and which need to be backwards compatible, like in the replicated state or in CUPs. The current rollout usually looks like:
- Introducing the field as optional, setting it to `None` everywhere.
- After the previous step is rolled out to all subnets, start filling the value with `Some` everywhere.
- After the previous step is rolled out to all subnets, turn the field non-optional.

This is well understood by now but could still be error-prone. The new `BackwardsCompatible` acts as a regular `Option`, though is impossible to instantiate and is ignored during hashing if still in the first step. Though it still "understands" the field and can be loaded from a future version that populated it.

### DIFF
--- a/rs/types/types/src/backwards_compatibility.rs
+++ b/rs/types/types/src/backwards_compatibility.rs
@@ -8,6 +8,7 @@ use std::hash::{Hash, Hasher};
 /// Compared to [`Option`], its [`Hash`] implementation:
 ///  - ignores `None` (so adding an unset field preserves the surrounding struct hash), and
 ///  - hashes `Some(v)` like `v` (so later replacing it with `T` preserves hashes).
+///
 /// The purpose of this is to have a compatible `Hash` implementation across versions of a struct,
 /// so that replicas running different versions of the code can still agree on the hash of a struct
 /// instance.
--- a/rs/types/types/src/backwards_compatibility.rs
+++ b/rs/types/types/src/backwards_compatibility.rs
@@ -9,6 +9,17 @@ use std::hash::{Hash, Hasher};
 ///  - ignores `None` (so adding an unset field preserves the surrounding struct hash), and
 ///  - hashes `Some(v)` like `v` (so later replacing it with `T` preserves hashes).
 ///
+/// IMPORTANT: there should be only one field of type `BackwardsCompatible` in a struct. Otherwise,
+/// two different instances of the struct could have the same hash. For example:
+/// ```
+/// #[derive(Hash)]
+/// struct S { a: BackwardsCompatible<u8, false>, b: BackwardsCompatible<u8, false> }
+///
+/// S { a: Some(1), b: None }
+/// S { a: None, b: Some(1) }
+/// ```
+/// would have the same hash.
+///
 /// Lifecycle of adding a new field to a struct in a backwards compatible way:
 /// 1. Add a new field to the struct with type `BackwardsCompatible<T, false>`. At this point the
 ///    field is *NOT* allowed to be instantiated with any value other than `None`. Though, it still
--- a/rs/types/types/src/backwards_compatibility.rs
+++ b/rs/types/types/src/backwards_compatibility.rs
@@ -60,3 +60,170 @@ impl<T, const SETTABLE: bool> BackwardsCompatibleOption<T, SETTABLE> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::hash::{DefaultHasher, Hash, Hasher};
+
+    fn hash_of<T: Hash>(value: &T) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        value.hash(&mut hasher);
+        hasher.finish()
+    }
+
+    #[test]
+    fn none_variant_is_ignored_in_hash() {
+        let none = BackwardsCompatibleOption::<u64, false>::new_for_test_only(None);
+
+        // Hashing the `None` variant must not write anything to the hasher, so
+        // the result must be identical to that of a fresh, untouched hasher.
+        assert_eq!(hash_of(&none), DefaultHasher::new().finish());
+    }
+
+    #[test]
+    fn some_variant_hashes_like_inner_value() {
+        let value = 42_u64;
+        let some = BackwardsCompatibleOption::<u64, true>::new(value);
+
+        assert_eq!(hash_of(&some), hash_of(&value));
+    }
+
+    #[derive(Hash)]
+    struct WithoutField {
+        a: u64,
+        b: String,
+    }
+    #[derive(Hash)]
+    struct WithUnsettableField {
+        a: u64,
+        b: String,
+        unsettable_field: BackwardsCompatibleOption<u32, false>,
+    }
+    #[derive(Hash)]
+    struct WithSettableField {
+        a: u64,
+        b: String,
+        settable_field: BackwardsCompatibleOption<u32, true>,
+    }
+
+    #[test]
+    fn adding_none_field_preserves_struct_hash() {
+        let without = WithoutField {
+            a: 7,
+            b: "ic".to_string(),
+        };
+        let with = WithUnsettableField {
+            a: 7,
+            b: "ic".to_string(),
+            unsettable_field: BackwardsCompatibleOption::default(),
+        };
+
+        // The whole point of the type: introducing a `None` field leaves the
+        // hash of the surrounding struct unchanged.
+        assert_eq!(hash_of(&without), hash_of(&with));
+    }
+
+    #[test]
+    fn adding_some_field_changes_struct_hash() {
+        let without = WithoutField {
+            a: 7,
+            b: "ic".to_string(),
+        };
+        let with = WithSettableField {
+            a: 7,
+            b: "ic".to_string(),
+            settable_field: BackwardsCompatibleOption::default(),
+        };
+
+        // Introducing a `Some` field changes the hash of the surrounding struct.
+        assert_ne!(hash_of(&without), hash_of(&with));
+    }
+
+    #[test]
+    fn default_is_none_when_not_settable() {
+        let opt = BackwardsCompatibleOption::<u64, false>::default();
+
+        assert_eq!(opt.as_ref(), None);
+    }
+
+    #[test]
+    fn default_is_some_default_when_settable() {
+        let opt = BackwardsCompatibleOption::<u64, true>::default();
+
+        assert_eq!(opt.as_ref(), Some(&u64::default()));
+    }
+
+    #[test]
+    fn new_wraps_value_in_some() {
+        let opt = BackwardsCompatibleOption::<u64, true>::new(123);
+
+        assert_eq!(opt.as_ref(), Some(&123));
+    }
+
+    #[derive(Debug, PartialEq)]
+    struct Value(u64);
+
+    struct ValidProto(u64);
+    impl TryFrom<ValidProto> for Value {
+        type Error = ProxyDecodeError;
+        fn try_from(proto: ValidProto) -> Result<Self, Self::Error> {
+            Ok(Value(proto.0))
+        }
+    }
+
+    struct InvalidProto;
+    impl TryFrom<InvalidProto> for Value {
+        type Error = ProxyDecodeError;
+        fn try_from(_: InvalidProto) -> Result<Self, Self::Error> {
+            Err(ProxyDecodeError::MissingField("Value"))
+        }
+    }
+
+    #[test]
+    fn try_from_proto_maps_none_to_none() {
+        let opt_settable =
+            BackwardsCompatibleOption::<Value, true>::try_from_proto(None::<ValidProto>)
+                .expect("conversion of `None` should never fail");
+        let opt_not_settable =
+            BackwardsCompatibleOption::<Value, false>::try_from_proto(None::<ValidProto>)
+                .expect("conversion of `None` should never fail");
+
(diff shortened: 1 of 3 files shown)


rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    pub fn try_from_proto<Proto: TryInto<T, Error = ProxyDecodeError>>(
+        proto: Option<Proto>,
+    ) -> Result<Self, ProxyDecodeError> {
+        match proto {
+            Some(value) => Ok(Self(Some(value.try_into()?))),
+            None => Ok(Self(None)),
+        }
+    }
```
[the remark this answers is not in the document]
alin-at-dfinity: I'm pretty sure this is how it is supposed to be, but it may be worth explicitly spelling it out in the doc comment.

rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,229 @@
...
+impl<T> BackwardsCompatibleOption<T, true> {
+    pub const fn new(value: T) -> Self {
+        Self(Some(value))
+    }
+}
+
+impl<T, const SETTABLE: bool> BackwardsCompatibleOption<T, SETTABLE> {
+    pub const fn new_for_test_only(value: Option<T>) -> Self {
+        Self(value)
+    }
```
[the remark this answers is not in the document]
alin-at-dfinity: I suppose you can add a `#[doc(hidden)]` (or whatever it is) attribute here.

rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,229 @@
...
+/// variant, which means that the hash of the struct without the field would be the same as the hash
+/// of the struct with the field present but set to `None`.
+///
+/// Lifecycle of adding a new field to a struct in a backwards compatible way:
+/// 1. Add a new field to the struct with type `BackwardsCompatibleOption<T, false>`. At this point
+///    the field is *NOT* allowed to have any value other than `None`.
+/// 2. When the change is deployed to all replicas, we can switch the type to
+///    `BackwardsCompatibleOption<T, true>` and the field can begin to be populated.
+/// 3. When the change is deployed to all replicas, we can replace the type with `T`.
+pub struct BackwardsCompatibleOption<T, const SETTABLE: bool>(Option<T>);
```
alin-at-dfinity: `BackwardsCompatibleOption` sounds like a field that is eventually intended to be an `Option<T>`, whereas, if I'm reading this correctly, the intended end state here is to have a field of the plain type `T`.

If so, I would rather name it something like `BackwardsCompatibleField` or something.

But this is just a slight preference, so please leave it as is if you disagree.

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
pierugo-dfinity: Here's how I understand it: If we have `SETTABLE == false`, we are running a replica version that "understands" the field (i.e. can read it with `as_ref()`) yet doesn't populate it. The only way we can observe a protobuf with `Some` is if a later replica version populated that field and we then rolled back. In my opinion, we should still deserialize the field and interpret it as `Some` as otherwise the hash would be different and we would ignore it when reading it through `as_ref()`. This is the first time I'm involved with backwards compatible changes though, so I'm happy to hear about your thoughts. @eichhorl you can also jump in.

In any case, I think more documentation can definitely help here, so I'll look into that.

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
pierugo-dfinity: Thanks. Indeed, we'd want to use it in other crates so `#[cfg(test)]` wouldn't work.

rs/types/types/src/backwards_compatibility.rs
pierugo-dfinity: That's very fair. I simplified it to `BackwardsCompatible`.

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
pierugo-dfinity: Then in that case actually maybe `ExhaustiveSet` should be the current implementation of `BackwardsCompatible<T, true>`, but also for `BackwardsCompatible<T, false>`

rs/types/types/src/exhaustive.rs
```diff
@@ -235,6 +236,21 @@ impl ExhaustiveSet for String {
...
+        vec![Default::default()]
+    }
+}
+
+impl<T: Clone + ExhaustiveSet> ExhaustiveSet for BackwardsCompatibleOption<T, true> {
+    fn exhaustive_set<R: RngCore + CryptoRng>(rng: &mut R) -> Vec<Self> {
+        Option::<T>::exhaustive_set(rng)
+            .into_iter()
+            .map(Self::new_for_test_only)
+            .collect()
```
eichhorl: Doesn't this also generate the `None` variant (during step 2. of the rollout)? If so, wouldn't it be incompatible with step 3, where we switch to `T`?

I wonder if we could add a test that showcases the whole rollout (and rollback) process

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
eichhorl: Yes, the value may be set due to a rollback

rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,239 @@
+use ic_protobuf::proxy::ProxyDecodeError;
+use serde::{Deserialize, Serialize};
+use std::hash::{Hash, Hasher};
+
+#[derive(Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
```
eichhorl: Usually I think the convention is to have this below the docs

eichhorl: Do we need to derive `Deserialize, Serialize`? Contrary to what the name suggests, I think by default, adding a `BackwardsCompatible` type to a struct actually isn't `serde` compatible (unlike adding a plain `Option` would be for some formats).

rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,239 @@
...
+///
+/// Lifecycle of adding a new field to a struct in a backwards compatible way:
+/// 1. Add a new field to the struct with type `BackwardsCompatible<T, false>`. At this point the
+///    field is *NOT* allowed to be instantiated with any value other than `None`. Though, it still
+///    "understands" it: if it receives a protobuf containing a set value, it will still convert it
+///    and store it in the `BackwardsCompatible` value.
+/// 2. When the change is deployed to all replicas, we can switch the type to
+///    `BackwardsCompatible<T, true>` and the field can begin to be populated.
+/// 3. When the change is deployed to all replicas, we can replace the type with `T`.
+pub struct BackwardsCompatible<T, const SETTABLE: bool>(Option<T>);
```
eichhorl, line 36: This is cool, but also not foolproof. Maybe we should mention that the type should not be used in collections or in multiple adjacent fields of a struct. Otherwise the following would have the same hash:
```rust
#[derive(Hash)]
struct S { a: BackwardsCompatible<u8, false>, b: BackwardsCompatible<u8, false> }

S { a: Some(1), b: None } 
S { a: None, b: Some(1) }
```

alin-at-dfinity: Excellent point, I hadn't even considered it.

I suppose that one way to work around it (still not foolproof) is to require the rollout to progress as fast as possible:
1. replica version N - add field;
2. replica version N+1 - flip `SETTABLE` to true;
3. replica version N+2 - replace type with `T`.

If you did this, there would be no point in time when field `a` (presumably added first) would be `None` but field `b` would be `Some(_)` (or `T`).

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
alin-at-dfinity: It doesn't need a whole lot more documentation. Just clarify that the type cannot be constructed with a value in phase 1; but it could be loaded from a checkpoint with a value set by a phase 2 implementation.

rs/types/types/src/backwards_compatibility.rs
schneiderstefan: Don't forget to update the PR title and description with the new name.

rs/types/types/src/backwards_compatibility.rs
pierugo-dfinity: A-ha nice catch. I tried to see if having a derive/attribute macro on the outer struct could help, but it would require quite a big effort. I hope this case should be rare enough not to run against it too much. I added this example to the docs. https://github.com/dfinity/ic/pull/10694/commits/e75e7bf19cd8362b321364f54276dda6cd1a2b8d

rs/types/types/src/backwards_compatibility.rs
pierugo-dfinity: `DkgSummary` derives both traits, so I just added them here without thinking more than that to make the compiler happy. Would https://github.com/dfinity/ic/pull/10694/commits/7528663a76800051c6e718272eafbd3b9f7ae55e work instead?

pierugo-dfinity: Thanks https://github.com/dfinity/ic/pull/10694/commits/48ebe69143c8e49e226668ac7d3ae7418fdecd18

rs/types/types/src/exhaustive.rs
pierugo-dfinity: > I wonder if we could add a test that showcases the whole rollout (and rollback) process

Good point, just added. https://github.com/dfinity/ic/pull/10694/commits/8b1ae152f20af5cf74e1d1cce4c76b5396e83597

> Doesn't this also generate the None variant (during step 2. of the rollout)? If so, wouldn't it be incompatible with step 3, where we switch to T?

Indeed, but the field in step 2. could also be empty right? When deserializing from step 1.

rs/types/types/src/backwards_compatibility.rs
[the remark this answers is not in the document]
pierugo-dfinity: Ah sorry Alin, I thought by
> this is how it is supposed to be

you were referring to Copilot's suggestion.

rs/types/types/src/exhaustive.rs
eichhorl: > Indeed, but the field in step 2. could also be empty right? When deserializing from step 1.

Yes, but the `ExhaustiveSet` implementation should reflect whatever the current replica version produces. Otherwise, the compatibility test will fail. See this part in the new test:
https://github.com/dfinity/ic/blob/e75e7bf19cd8362b321364f54276dda6cd1a2b8d/rs/types/types/src/backwards_compatibility.rs#L432-L437

During step 2, only variants that set the fields may be produced, otherwise the step 3 check would fail:
https://github.com/dfinity/ic/blob/e75e7bf19cd8362b321364f54276dda6cd1a2b8d/rs/types/types/src/backwards_compatibility.rs#L448

Maybe we should also generate the structs in this test via the `ExhaustiveSet` (like we do in the cup compatibility test), which would reveal the problem.

rs/types/types/src/backwards_compatibility.rs
eichhorl: I see, I'm also fine with deriving it if we need to, but maybe clarify in the docs that the purpose of this struct is to make `Hash` compatible

rs/types/types/src/backwards_compatibility.rs
eichhorl: Works for me 👍

rs/types/types/src/exhaustive.rs
pierugo-dfinity: Ah, I understand how `ExhaustiveSet` is used now, you're right. Maybe something like https://github.com/dfinity/ic/pull/10694/commits/c0319f17d90d5d63c170fa0ac9159f8442a1f89e? Or did you think of something else?

rs/types/types/src/backwards_compatibility.rs
pierugo-dfinity: Reverted + added some docs 👍 https://github.com/dfinity/ic/pull/10694/commits/8c10145e80cad6096b4b0fb5454e599178aa863c

rs/types/types/src/backwards_compatibility.rs
```diff
@@ -0,0 +1,341 @@
...
+
+impl<T> BackwardsCompatible<T, true> {
+    pub const fn new(value: T) -> Self {
+        Self(Some(value))
+    }
+}
+
+impl<T, const SETTABLE: bool> BackwardsCompatible<T, SETTABLE> {
+    #[doc(hidden)]
+    pub const fn new_for_test_only(value: Option<T>) -> Self {
```
eichhorl, line 75: this seems to be only used in tests now

eichhorl (approved):

rs/types/types/src/backwards_compatibility.rs
pierugo-dfinity: The plan is to be able to call this function from unit tests of other crates, such that one can test a feature even if the field would not be settable yet. Added a comment 👍

PR #10695, opened 2026-07-08
fix: Unmap initial state data

schneiderstefan:
The mmapped initial state data is never munmapped, causing unnecessary increases in resident memory after canister installs and upgrades. This wasn't a large problem, as the memory could be reclaimed if necessary, but it's still cleaner to stop mapping it as soon as we know that we won't need it anymore.

### DIFF
--- a/rs/embedders/src/serialized_module.rs
+++ b/rs/embedders/src/serialized_module.rs
@@ -13,7 +13,7 @@ use ic_interfaces::execution_environment::{HypervisorError, HypervisorResult};
 use ic_replicated_state::canister_state::execution_state::WasmMetadata;
 use ic_types::{DiskBytes, NumInstructions, methods::WasmMethod};
 use ic_wasm_types::WasmEngineError;
-use nix::sys::mman::{MapFlags, ProtFlags, mmap};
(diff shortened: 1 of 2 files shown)


rs/embedders/src/serialized_module.rs
```diff
@@ -259,8 +259,15 @@ impl OnDiskSerializedModule {
...
-            .expect("Error parsing initial state data file")
+        let initial_state_data = bincode::deserialize::<InitialStateData>(data)
+            .expect("Error parsing initial state data file");
+        // Safety: `mmap_ptr`/`mmap_size` are the pointer and length returned by
+        // the `mmap` above; `data` is not used past this point.
+        unsafe {
+            munmap(mmap_ptr as *mut std::ffi::c_void, mmap_size)
+                .expect("Unable to unmap initial state data file");
+        }
+        initial_state_data
```
alin-at-dfinity, line 270: While I was debugging this, Claude claimed that there was a similar `mmap` leak in `SandboxManager::create_execution_state()`.

alin-at-dfinity (approved):
LGTM, thank you.

PR #10699, opened 2026-07-08
feat(delegations): reactively fetch delegations on new public key and/or canister ranges

pierugo-dfinity:
During canister migrations and soon subnet splitting, the public key of a subnet and its canister ranges are modified. Because the delegation manager fetches new delegations every 5 minutes independently from the state, there can be up to 5 minutes of additional downtime during those operations just to let each replica fetch a new delegation that matches their state.

This PR proposes to add a new asynchronous path to the delegation manager to fetch new delegations. To enable this, the PR also refactors the delegation *reader*.

## Delegation manager changes

The current "fetch every 5 minutes" logic stays with one important difference: if the fetched delegation does not match the latest certified state, hold it back, i.e. ignore it and try again 5 minutes later. This is to avoid the case where we would have fetched a delegation too early.

Asynchronously, the delegation manager locally checks every 10 seconds whether its latest delegation matches its latest certified state. In almost all cases, they will, and it won't do anything. Though if they do mismatch, then it reactively fetches a new delegation. This allows to reduce the maximum downtime from 5 minutes to 10 seconds.

## Delegation reader refactor

The `NNSDelegationBuilder/Reader` are moved to their own crate `ic_nns_delegation_reader`, which contains the relevant code for comparing the delegations with the replicated state. By putting the comparison/validation logic directly in `NNSDelegationBuilder`, this will allow the HTTP handler to check whether its delegation matches the latest certified state and return a 503 error if not instead of serving an inconsistent delegation.
For now, "delegation matching the state" means the same subnet public key and the same canister ranges for one's own subnet, but we can add more variants to `CanisterRangesCheck` to check whether a specific canister ID is present in both the delegation and state (instead of checking for all ranges).

The motivation for moving it to a different crate is that the query handler will need to depend on it to perform the comparison and it would be better not to also depend on all the dependencies of `ic_nns_delegation_manager`. In particular, `ic_nns_delegation_reader` cannot depend on `ReplicatedState` as it would otherwise create a circular dependency.


rs/canonical_state/src/delegation.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// The embedded certificate could not be CBOR-decoded.
+    MalformedCertificate(String),
+    /// The certificate's mixed hash tree could not be converted into a labeled tree.
+    MalformedHashTree(String),
+    /// An expected path was missing from the certificate or had an unexpected shape.
+    UnexpectedTreeShape(String),
+    /// A certified canister-ranges leaf could not be CBOR-decoded.
+    MalformedCanisterRanges(String),
+    /// The state has no topology for the delegated subnet.
+    UnknownSubnet(SubnetId),
```
eichhorl: Alternatively we could keep the real error types where possible, and only do the string conversion in the `Display` implementation. That might simplify some of the `map_err` calls.

rs/canonical_state/src/delegation.rs
```diff
@@ -0,0 +1,615 @@
...
+                Some(LabeledTree::SubTree(children)) => {
+                    // In the tree layout, we only check that the ranges present in the delegation
+                    // match the ranges assigned to the subnet in the state.
+                    if children.is_empty() {
+                        // This could genuinely happen if the routing table has changed but we
+                        // haven't refreshed the NNS delegation just yet. In that case, we would
+                        // have built a delegation with no ranges, which is invalid. So we return
+                        // false here.
+                        return Ok(false);
+                    }
```
eichhorl: Isn't an empty range encoded as a single leaf for canister id 0 with an empty list?
https://github.com/dfinity/ic/blob/67dee00ef30d27f2197a67a563d508869279d301/rs/canonical_state/src/lazy_tree_conversion.rs#L346-L347

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -143,23 +182,61 @@ impl DelegationManager {
...
+            return Some(self.fetch().await);
+        }
+
+        None
+    }
+
     async fn run(self, sender: watch::Sender<Option<NNSDelegationBuilder>>) {
-        let mut interval = tokio::time::interval(DELEGATION_UPDATE_INTERVAL);
+        let mut proactive_interval = tokio::time::interval(DELEGATION_PROACTIVE_UPDATE_INTERVAL);
+        let mut reactive_interval = tokio::time::interval(DELEGATION_REACTIVE_UPDATE_INTERVAL);
```
eichhorl, line 229: This uses `MissedTickBehavior::Burst` by default, meaning if one `fetch` takes very long, and we miss some ticks, the missed ticks will be done back to back until the interval has "caught" up. Maybe it would be better to set `MissedTickBehavior::Delay`?

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -143,23 +182,61 @@ impl DelegationManager {
...
-                let modified = if &delegation != old_delegation {
-                    std::mem::swap(old_delegation, &mut delegation);
+            // Fetch the delegation if enough time has passed
+            let Some(new_delegation) = select!(
+                _ = proactive_interval.tick() => self.proactive_fetch().await,
+                _ = reactive_interval.tick() => self.reactive_fetch(last_delegation.as_ref()).await,
+            ) else {
+                // No new delegation was fetched. Retry on the next tick.
+                continue;
+            };
```
eichhorl, line 253: If the first delegation we fetch (with `proactive_fetch`) right after a restart doesn't match the one in the state, then we would hold it back. But in that case, `reactive_fetch` will not work because there is no `last_delegation`. This means we will wait for 5 minutes until we try again with `proactive_fetch`, at which point the state is hopefully caught up. Is that intended?

rs/canonical_state/src/delegation.rs
pierugo-dfinity: True, thanks https://github.com/dfinity/ic/pull/10699/commits/5be6b6296088c6aa322561357e04a38ee843f7f3

rs/canonical_state/src/delegation.rs
pierugo-dfinity: I had added this if-check because I thought that we could enter that branch in practice when building a delegation for a canister that we are not responsible for ([here](https://github.com/dfinity/ic/blob/fa09a04fd83bdf03759c14a3b19ea6c633d9bb39/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_reader.rs#L287-L292)). Though when diving deeper, I realize that it would generate a delegation without even the `canister_ranges` node, meaning we would hit the `UnexpectedTreeShape` error below. The if-check is effectively dead code, but I would still keep it for robustness (updating the comment and turning `Ok(false)` to `Err(_)`, as it is clearly unexpected https://github.com/dfinity/ic/pull/10699/commits/cd8ca34c0a165f1f29b1205036ad061ba655d1f0): if the delegation is malformed and does not contain any ranges for our subnet, we would not want to return `Ok(true)`.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: I hadn't looked at the internals of `interval` and this definitely makes sense, thank you! Also applied `Delay` to the proactive one. https://github.com/dfinity/ic/pull/10699/commits/805f44bb2865a5a864a5f5e06ecf280b66768f4b

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: That's a very good point. I am thinking that this scenario should be relatively rare and not so harmful (this would add 5 more potential minutes of downtime but to a node which is already catching up/unhealthy) to justify some added complexity to cover that edge-case, as we would probably have to fetch the delegation, hold it back but still keep it in memory, to compare it with the state every 10 seconds, and once the latter catches up, "deliver" it to the HTTP handler.

eichhorl: Wouldn't it be enough to only hold back the delegation if we already have one in proactive fetch?

pierugo-dfinity: You're right, we could do that actually. Though what I like with the current implementation is that there is an implicit invariant that ensures that the delegation is never ahead of the state. This would break by handling that edge case. In particular, if we do what you suggest, the manager would start querying the NNS every 10 seconds in an effort to update its delegation (that it thinks is outdated, but in reality, is ahead), which introduces unnecessary load.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -143,23 +182,69 @@ impl DelegationManager {
...
+            return None;
+        }
+
+        Some(new_delegation)
+    }
+
+    /// Fetches a delegation from the NNS subnet reactively, i.e. only if the current delegation is
+    /// incompatible with the certified state. If the delegation is still valid, it will not be
+    /// fetched again (i.e. returns `None`).
+    async fn reactive_fetch(
```
eichhorl, line 214: Should we add some new metrics to `DelegationManagerMetrics`? I.e. whenever a delegation is held back, or fetched reactively

rs/canonical_state/src/delegation.rs
```diff
@@ -560,10 +560,13 @@ mod tests {
         // The delegation certifies no ranges at all.
         let delegation = delegation(subnet_id, &build_tree(format, subnet_id, &public_key, &[]));
 
-        assert_matches!(
-            is_delegation_valid_with_respect_to_state(&delegation, format, &state),
-            Ok(is_valid) if is_valid == matches!(format, CertificateDelegationFormat::Pruned)
-        );
+        assert_eq!(
+            matches!(
```
daniel-wong-dfinity-org-twin: Why use a macro when you can simply use `==` ??

Also, we need explicit breadcrumbs here, because if this fails, all it will say is (approximately), "expected true, got false", which is not so helpful.

# Sermon

In general, I really think we should avoid these match macros entirely. I know they are used regularly in the ic repo (esp in tests), but I think that's a mistake (albeit, a small one). I just do not see how they pull their own weight.

Imagine if they were banned. What you would do in that case is simply use the `match` statement. That would not be significantly more verbose, but it would be a fair bit less magical.

One of the things that makes macros magical is that their arguments need not be Rust expressions. This is why, for example, rustfmt does not touch code within a macro call. In particular, the second argument to match macros is NOT an expression.

daniel-wong-dfinity-org-twin (commented):
Woops. I was not looking at the whole PR somehow. Clicked on a wrong link.

daniel-wong-dfinity-org-twin (approved):

rs/canonical_state/src/delegation.rs
pierugo-dfinity: I could not use `==` because `Result<bool, DelegationValidationError>` does not implement `PartialEq`, and it cannot because `serde_cbor::Error` does not implement it either.

Breadcrumbs added [here](https://github.com/dfinity/ic/pull/10699/commits/b1534a22cc7a03338c41ad820271853655d17376), thanks.

daniel-wong-dfinity-org-twin: > serde_cbor::Error does not implement it

🤦 This is why my Claude instructions includes [this][eq].

[eq]: https://sourcegraph.com/r/github.com/dfinity/ic@61d44387f3b41eaa24c6542d287105269656ea48/-/blob/rs/nervous_system/standards_and_conventions/claude_author_instructions.md?L197&view=raw

rs/registry/routing_table/src/lib.rs
```diff
@@ -555,6 +555,12 @@ impl RoutingTable {
         lookup_in_ranges(&self.0, canister_id)
     }
 
+    /// Returns the corresponding `SubnetId` that the given `range` is assigned to or `None` if an
+    /// assignment cannot be found.
```
eichhorl: Do I understand correctly that the given range could be assigned to a subnet, but we still return `None`, unless it is an exact match?

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -115,13 +125,42 @@ struct DelegationManager {
...
+    ) -> Option<bool> {
+        let Some(delegation) = delegation else {
+            // No delegation: Initialization or on the NNS subnet: return true to proactively fetch
+            // a new one (which is a no-op on the NNS).
+            return Some(true);
+        };
+
+        ic_canonical_state::delegation::is_delegation_valid_with_respect_to_state(
+            // Build the delegation in the flat format to capture all ranges
+            &delegation.build_or_original(CanisterRangesFilter::Flat, &self.log),
```
eichhorl: Doesn't the `NNSDelegationBuilder` already contain the tree in deserialized form? Why do we need to pass the blob here, and then deserialize it again in `is_delegation_valid_with_respect_to_state`?

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: Good idea!

rs/registry/routing_table/src/lib.rs
pierugo-dfinity: True. I did not think it would be a problem in our use-case but it is probably better to be more correct. Let me see if I even need that function anyways after I refactor things out (see next comment).

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: It's indeed a bit awkward. The reason was that I also wanted to use that function from the query handler to return a 503 on an outdated delegation, which only has the serialized delegation. Let me try to see if I can refactor things out more nicely.

rs/http_endpoints/nns_delegation_reader/src/reader.rs
```diff
@@ -123,29 +139,120 @@ impl NNSDelegationBuilder {
...
+    /// Verifies that the delegation is consistent with the given view of the subnet
+    /// information recorded in a replicated state and, only if it is, builds it
+    /// according to the ranges check to be applied and returns it. The builder is
+    /// an immutable snapshot of the delegation, so the returned delegation is
+    /// guaranteed to be exactly the one which was verified.
+    ///
+    /// `ranges_check` specifies what to check the certified canister ranges against
+    /// (see [`CanisterRangesCheck`]). For the meaning of `state_view_for_subnet`,
+    /// see [`Self::is_consistent_with`].
+    pub fn build_verified(
```
pierugo-dfinity: Unused for now (except in unit tests), but will be used by the HTTP handler when checking if the latest delegation matches the latest state.

rs/http_endpoints/nns_delegation_reader/src/validation.rs
```diff
@@ -0,0 +1,485 @@
...
+
+/// What to check the canister ranges certified in a delegation against.
+///
+/// The certified threshold public key is always checked, independently of the variant.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum CanisterRangesCheck {
+    /// Check that the certified canister ranges at `/subnet/<subnet_id>/canister_ranges`
+    /// exactly match all the ranges which the state assigns to the delegated subnet. Ranges
+    /// certified at `/canister_ranges/<subnet_id>` are ignored.
+    AllSubnetRanges,
```
pierugo-dfinity, line 36: If the delegation is built correctly, the canister ranges in the flat and tree paths should be consistent and it should be sufficient to look only at the flat canister ranges

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: Hey @eichhorl, PTAL at the latest changes. Sorry for the big diff, I had to refactor the `NNSDelegationBuilder/Reader` to get what I wanted. I think it looks better than the original implementation. I updated the PR description to describe the motivation of this refactor.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
```diff
@@ -1372,4 +1595,356 @@ mod tests {
...
+            DELEGATION_REACTIVE_UPDATE_INTERVAL * 2,
+            reader.wait_until_initialized(),
+        )
+        .await
+        .expect("Should try to reactively refresh the delegation until the latter matches the current state")
+        .unwrap();
+    }
+
+    #[tokio::test]
+    #[rstest]
```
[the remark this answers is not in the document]
pierugo-dfinity: Looks like the current code works as intended but the suggested alternative just doesn't compile. Still moved tokio's attribute last as `rstest` is expected to run first. https://github.com/dfinity/ic/pull/10699/commits/fb3c994059e7fb653c91f8c9376a65cf7c4f8293

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
eichhorl: > The reason was that I also wanted to use that function from the query handler to return a 503 on an outdated delegation, which only has the serialized delegation.

Then couldn't we just have two functions, one that takes the serialized delegation, does the deserialization, and then calls the second one which takes the deserialized delegation?

rs/http_endpoints/nns_delegation_manager/src/metrics.rs
eichhorl, line 21: I think this currently counts the number of fetches (not necessarily updates). Although this metric seems redundant anyway since `update_duration` below also contains the total number of observations.

eichhorl: Similarly this measures the fetch duration

rs/http_endpoints/nns_delegation_reader/src/reader.rs
```diff
@@ -123,29 +139,120 @@ impl NNSDelegationBuilder {
...
+    /// ```
+    /// Resolving to `None` maps to [`DelegationValidationError::UnknownSubnet`].
+    /// `ranges_check` specifies what to check the certified canister ranges against
+    /// (see [`CanisterRangesCheck`]).
+    ///
+    /// See [`is_tree_consistent_with`] for the exact semantics.
+    pub fn is_consistent_with(
+        &self,
+        ranges_check: CanisterRangesCheck,
+        state_view_for_subnet: impl FnOnce(SubnetId) -> Option<(Vec<u8>, CanisterIdRanges)>,
```
eichhorl: Why does this take a closure for the subnet ID? Shouldn't all callers already know the subnet ID?

rs/http_endpoints/nns_delegation_manager/src/metrics.rs
pierugo-dfinity: Good point, I moved it to count the actual updates. https://github.com/dfinity/ic/pull/10699/commits/e029f1e26a3d89873f197a4df9ecb8348932386a

pierugo-dfinity: Renamed https://github.com/dfinity/ic/pull/10699/commits/e029f1e26a3d89873f197a4df9ecb8348932386a

rs/http_endpoints/nns_delegation_reader/src/reader.rs
pierugo-dfinity: The `Builder` already has a `subnet_id` and I thought it would be better to use it to avoid inconsistencies.

> Shouldn't all callers already know the subnet ID?

The HTTP handler and the query handler will be such callers, and neither of them know the subnet ID (we could pass it around, but that would be additional plumbing).

Optimally, it should be sufficient to pass a `&ReplicatedState`, but this crate cannot depend on it as this would otherwise create a cyclic dependency.

rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
pierugo-dfinity: We would then still serialize + deserialize the delegation for every query :/
The plan is to pass an `Arc<NNSDelegationBuilder>` to the query handler instead of the serialized delegation, which would be reference-shared with the delegation manager and the HTTP handler

rs/http_endpoints/nns_delegation_manager/src/metrics.rs
```diff
@@ -16,7 +19,9 @@ impl DelegationManagerMetrics {
                 "nns_delegation_manager_updates_total",
                 "How many times has the nns delegation been updated",
             ),
-            update_duration: metrics_registry.histogram(
+            fetch_duration: metrics_registry.histogram(
+                // Metric name kept as `update_duration` for backwards compatibility, even though it
+                // measures the fetch duration.
                 "nns_delegation_manager_update_duration_seconds",
```
eichhorl: Is this metric even used anywhere? https://github.com/search?q=repo%3Adfinity-ops%2Fk8s+nns_delegation_manager_update_duration_seconds&type=code

rs/http_endpoints/nns_delegation_reader/src/reader.rs
eichhorl: What is the cyclic dependency exactly? Claude seems pretty convinced that there is none, but maybe I prompted the wrong crates

eichhorl: I assume once we start using this for queries, we would also want a different `CanisterRangesCheck` for a single canister ID? But the closure as it is now always needs to return some ranges

rs/http_endpoints/nns_delegation_manager/src/metrics.rs
pierugo-dfinity: Lol ok, let me rename it then https://github.com/dfinity/ic/pull/10699/commits/525398804625424f10bccee16a9ee36796c39f53

rs/http_endpoints/nns_delegation_reader/src/reader.rs
pierugo-dfinity: > What is the cyclic dependency exactly?

`ic-interfaces` defines `QueryExecutionInput` which will have an `Option<Arc<NNSDelegationBuilder>>`, so `ic-interfaces -> ic-nns-delegation-reader -> ic-replicated-state -> ic-interfaces`

> I assume once we start using this for queries, we would also want a different CanisterRangesCheck for a single canister ID?

Correct

> But the closure as it is now always needs to return some ranges

We would pass all canister ranges of the subnet, and check that the given canister ID indeed falls in one of them. Now actually, we could simply pass the entire routing table and check with `lookup_entry` when checking for a specific canister, even better: https://github.com/dfinity/ic/pull/10699/commits/e4d5921ba1520824bfe8d294d4d303aeb78456dc (no functional differences for the moment)

eichhorl: > we could pass it around, but that would be additional plumbing

I guess technically we are already passing it around as part of the `NNSDelegationBuilder`? Alternatively, claude suggests moving the functions that would take `ReplicatedState` somewhere else, i.e. `ic_canonical_state`.

I'm also fine with the current version, but maybe we could change the closure to return a slice so that we won't need to clone the public key for every query?

rs/http_endpoints/nns_delegation_reader/src/reader.rs
```diff
@@ -123,29 +139,121 @@ impl NNSDelegationBuilder {
...
-            builder.build_or_original(CanisterRangesFilter::Flat, logger);
-
-        Self {
-            builder,
-            precomputed_delegation_with_flat_canister_ranges,
-            precomputed_delegation_without_canister_ranges,
+            // Placeholders, replaced right below (`build_uncached_or_original` needs the
+            // other fields to be initialized).
+            precomputed_delegation_with_flat_canister_ranges: original_delegation.clone(),
+            precomputed_delegation_without_canister_ranges: original_delegation.clone(),
```
eichhorl: Is there an easy way to avoid this?

rs/http_endpoints/nns_delegation_reader/src/reader.rs
pierugo-dfinity: > Alternatively, claude suggests moving the functions that would take ReplicatedState somewhere else, i.e. ic_canonical_state.

And have the validation logic split into two crates..? :/

> return a slice

Yes, I can definitely do that 👍 https://github.com/dfinity/ic/pull/10699/commits/b65e63dc9152b04f2017d22ed2916acb82a5c86d

rs/http_endpoints/nns_delegation_reader/src/validation.rs
eichhorl: Maybe we can add this reasoning also to the doc comment

rs/http_endpoints/nns_delegation_reader/src/validation.rs
```diff
@@ -0,0 +1,475 @@
...
+            b"canister_ranges",
+        ],
+    ) {
+        Some(LabeledTree::Leaf(bytes)) => {
+            Ok(Some(decode_ranges(bytes)?.as_slice() == state_ranges))
+        }
+        Some(LabeledTree::SubTree(_)) => Err(DelegationValidationError::UnexpectedTreeShape(
+            format!("unexpected subtree at /subnet/{subnet_id}/canister_ranges"),
+        )),
+        None => Ok(None),
```
eichhorl: Do we have a test for these cases?

rs/http_endpoints/nns_delegation_reader/src/reader.rs
eichhorl: > And have the validation logic split into two crates..?

Two functions that take the delegation, state and `CanisterRangesCheck`, and then call `delegation.is_consistent_with` or `delegation.build_verified`

eichhorl (approved):

rs/http_endpoints/nns_delegation_reader/src/reader.rs
pierugo-dfinity: Sure, let me revert the inline of `BuilderInner` https://github.com/dfinity/ic/pull/10699/commits/4b570b3f11007b84fb3405fa46970f2e575785e9

rs/http_endpoints/nns_delegation_reader/src/validation.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10699/commits/a54a8e13e7bde137cfc7b7e091066460a7ef1661

rs/http_endpoints/nns_delegation_reader/src/validation.rs
pierugo-dfinity: Indeed not: https://github.com/dfinity/ic/pull/10699/commits/566488e295797b61885fc89b1e55bba1d8a320d9 I took the opportunity to directly return an error instead of `Ok(None)` https://github.com/dfinity/ic/pull/10699/commits/1f056f053443814689c039ecde068ced132eb31f

rs/http_endpoints/nns_delegation_reader/src/reader.rs
pierugo-dfinity: Hmmm, yes that could work. Though not sure if that simple shortcut truly justifies an additional dependency... 

And holders of a `NNSDelegationBuilder` wouldn't be so sure whether to pull the `ic_canonical_state` dependency and call its function, or directly call the method.

PR #10705, opened 2026-07-09
feat(cketh): TimedSizedMap bounded, time-expiring map

gregorydemay:
Adds a generic in-heap map bounded by **both** a hard entry cap and a per-entry time-to-live, which will be used to store addresses that needs to be watch over for deposits.

On insertion, re-inserting a key that still has a live entry is **rejected up front** with `AlreadyPresent`, leaving the map unchanged (a caller cannot refresh/extend an entry by re-inserting it; re-admission is only possible once it has expired). Otherwise, admitting a new key first evicts entries that have outlived their TTL, then **rejects rather than evicts a live entry** — returning `AtCapacity` if the map is still full of live entries. Time is supplied by the caller (a `Timestamp` newtype), so the structure keeps no internal clock and is deterministic to test.
Modeled on the bounded cache in `dfinity/canhttp` (not a workspace dependency, hence a small local structure).


rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn remove_from_time_index(&mut self, key: &K, inserted_at: Timestamp) {
+        if let Some(bucket) = self.by_time.get_mut(&inserted_at) {
+            if let Some(position) = bucket.iter().position(|indexed| indexed == key) {
+                bucket.remove(position);
+            }
+            if bucket.is_empty() {
+                self.by_time.remove(&inserted_at);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in b880d83251 — `remove_from_time_index` now `expect("BUG: ...")`s on a missing bucket or key instead of silently skipping, so an inconsistent index fails loudly rather than drifting and panicking later in eviction. Also bound `is_empty` to end the bucket borrow before the outer `by_time.remove`.

rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// counting against capacity. Returns the entries evicted by this call.
+    pub fn insert(&mut self, now: Timestamp, key: K, value: V) -> Vec<(K, V)> {
+        let mut evicted = self.evict_expired(now);
+        self.remove(&key);
+        while self.entries.len() >= self.capacity.get() {
+            match self.evict_oldest() {
+                Some(entry) => evicted.push(entry),
+                None => break,
+            }
+        }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in aaf656e620 — the capacity loop now `expect("BUG: ...")`s instead of breaking, so an empty time index while `entries` is non-empty (index corruption) fails loudly rather than inserting past the hard capacity.

rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,168 @@
...
+            by_time: BTreeMap::new(),
+        }
+    }
+
+    /// Insert `value` under `key`, evicting expired then oldest entries as needed to respect the
+    /// `ttl` and `capacity`. Re-inserting an existing key refreshes its value and lifetime without
+    /// counting against capacity. Returns the entries evicted by this call.
+    pub fn insert(&mut self, now: Timestamp, key: K, value: V) -> Vec<(K, V)> {
+        let mut evicted = self.evict_expired(now);
+        self.remove(&key);
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in aaf656e620 — `insert` now removes the existing key *before* evicting expired entries, so re-inserting a key never reports its own previous value as an eviction regardless of whether it had expired. Added a test (`should_not_report_refresh_of_expired_key_as_eviction`).

rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,169 @@
...
+            by_time: BTreeMap::new(),
+        }
+    }
+
+    /// Insert `value` under `key`, evicting expired then oldest entries as needed to respect the
+    /// `ttl` and `capacity`. Re-inserting an existing key refreshes its value and lifetime; the
+    /// previous value is dropped and never reported as an eviction. Returns the entries evicted by
+    /// this call.
+    pub fn insert(&mut self, now: Timestamp, key: K, value: V) -> Vec<(K, V)> {
+        self.remove(&key);
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive: none of these are `#[must_use]` — not `Option<T>` as a type, nor `BTreeMap::insert`/`BTreeMap::remove`/`VecDeque::remove`, nor this crate's `remove` — so no `unused_must_use` warning fires. CI proves it: clippy runs with `-D warnings` (which would promote `unused_must_use` to a hard error) and passes clean, as does the bazel build. No change needed.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn insert(&mut self, now: Timestamp, key: K, value: V) -> Vec<(K, V)> {
+        self.remove(&key);
+        let mut evicted = self.evict_expired(now);
+        while self.entries.len() >= self.capacity.get() {
+            evicted.push(
+                self.evict_oldest()
+                    .expect("BUG: entries is non-empty but the time index is empty"),
+            );
+        }
+        self.entries.insert(
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive: none of these are `#[must_use]` — not `Option<T>` as a type, nor `BTreeMap::insert`/`BTreeMap::remove`/`VecDeque::remove`, nor this crate's `remove` — so no `unused_must_use` warning fires. CI proves it: clippy runs with `-D warnings` (which would promote `unused_must_use` to a hard error) and passes clean, as does the bazel build. No change needed.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+            .expect("BUG: entry timestamp missing from the time index");
+        let position = bucket
+            .iter()
+            .position(|indexed| indexed == key)
+            .expect("BUG: entry missing from its timestamp bucket");
+        bucket.remove(position);
+        let is_empty = bucket.is_empty();
+        if is_empty {
+            self.by_time.remove(&inserted_at);
+        }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive: none of these are `#[must_use]` — not `Option<T>` as a type, nor `BTreeMap::insert`/`BTreeMap::remove`/`VecDeque::remove`, nor this crate's `remove` — so no `unused_must_use` warning fires. CI proves it: clippy runs with `-D warnings` (which would promote `unused_must_use` to a hard error) and passes clean, as does the bazel build. No change needed.

rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn evict_oldest(&mut self) -> Option<(K, V)> {
+        let inserted_at = *self.by_time.keys().next()?;
+        let bucket = self
+            .by_time
+            .get_mut(&inserted_at)
+            .expect("BUG: bucket must exist");
+        let key = bucket.pop_front().expect("BUG: bucket must be non-empty");
+        if bucket.is_empty() {
+            self.by_time.remove(&inserted_at);
+        }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive: none of these are `#[must_use]` — not `Option<T>` as a type, nor `BTreeMap::insert`/`BTreeMap::remove`/`VecDeque::remove`, nor this crate's `remove` — so no `unused_must_use` warning fires. CI proves it: clippy runs with `-D warnings` (which would promote `unused_must_use` to a hard error) and passes clean, as does the bazel build. No change needed.

rs/ethereum/cketh/minter/src/timed_sized_map.rs
```diff
@@ -0,0 +1,169 @@
...
+            entries: BTreeMap::new(),
+            by_time: BTreeMap::new(),
+        }
+    }
+
+    /// Insert `value` under `key`, evicting expired then oldest entries as needed to respect the
+    /// `ttl` and `capacity`. Re-inserting an existing key refreshes its value and lifetime; the
+    /// previous value is dropped and never reported as an eviction. Returns the entries evicted by
+    /// this call.
+    pub fn insert(&mut self, now: Timestamp, key: K, value: V) -> Vec<(K, V)> {
```
gregorydemay: In this context, we cannot remove an entry whose ttl is not expired: it would mean dropping a user's address from the watchlist which may in between have received funds. We need to be able to return an error if we reached capacity, even after having removed expired entries

gregorydemay: 🤖 Fixed in the pushed commit. `insert` now evicts only expired entries and, if the map is still at capacity with only live entries, returns an `AtCapacity` error instead of evicting a live one — a still-armed address is never dropped. Re-inserting an existing key still refreshes it and is never rejected. Removed the now-unused `evict_oldest`; the deposit-address wiring PR will map `AtCapacity` to a `TooManyActiveAddresses` endpoint error.

rs/ethereum/cketh/minter/src/timed_sized_map/tests.rs
```diff
@@ -0,0 +1,148 @@
...
+    map.insert(ts(3), "c", 3).unwrap();
+    assert_eq!(map.len(), 3);
+
+    let rejected = map.insert(ts(4), "d", 4);
+
+    assert_eq!(rejected, Err(AtCapacity { key: "d", value: 4 }));
+    assert_eq!(map.len(), 3);
+    assert_eq!(map.get(ts(4), &"a"), Some(&1));
+    assert_eq!(map.get(ts(4), &"d"), None);
+    assert_consistent(&map);
```
gregorydemay: clone the map before insertion and assert_eq on after (failed) insertion on the whole map

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty()
+    }
+
+    pub fn capacity(&self) -> NonZeroUsize {
+        self.capacity
+    }
+
+    fn is_expired(&self, inserted_at: Timestamp, now: Timestamp) -> bool {
+        inserted_at.saturating_add(self.ttl) <= now
```
gregorydemay: shouldn't that be strict? Better to now.checked_sub(inserted_at) > self.ttl

gregorydemay: 🤖 Done — `is_expired` now uses `now.checked_duration_since(inserted_at) > self.ttl` (strict), so an entry lives for the full `ttl` and expires only once its age exceeds it. Dropped the old inclusive `saturating_add` comparison and updated the tests to the new boundary.

rs/ethereum/cketh/minter/src/timed_sized_map/tests.rs
gregorydemay: 🤖 Done — derived `Clone`/`Eq`/`PartialEq` on `TimedSizedMap` and the rejection test now clones the map before the failed insert and asserts the whole map is unchanged afterwards (`assert_eq!(map, before)`).

rs/ethereum/cketh/minter/src/timed_sized_map/tests.rs
```diff
@@ -0,0 +1,146 @@
...
+    map.insert(ts(2), "b", 2).unwrap();
+
+    let evicted = map.insert(ts(3), "a", 11).unwrap();
+
+    assert!(evicted.is_empty());
+    assert_eq!(map.len(), 2);
+    assert_eq!(map.get(ts(3), &"a"), Some(&11));
+    assert_eq!(map.get(ts(3), &"b"), Some(&2));
+    assert_consistent(&map);
+}
```
gregorydemay, line 104: A caller should not be able to refresh (otherwise the caller will keep calling deposit_erc20). We should return an error in that case.

gregorydemay: 🤖 Done. `insert` now returns `InsertError::AlreadyPresent` when a live entry already exists under the key, so a caller cannot refresh/extend it by re-calling — re-admission is only possible after the entry has expired (evicted first, then re-inserted fresh). The error type is now an enum (`AlreadyPresent` | `AtCapacity`); dropped the now-unused by-key remove helpers. Test renamed to `should_reject_refresh_of_live_key`.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+        self.entries.insert(
+            key.clone(),
+            Entry {
+                value,
+                inserted_at: now,
+            },
+        );
+        self.by_time.entry(now).or_default().push_back(key);
+        Ok(evicted)
```
mbjorkqvist: A failed insert returning an `AlreadyPresent` error could silently evict entries (if `self.evict_expired(now)` returns a non-empty vector, but `self.entries.contains_key(&key)` is `true`), which doesn't seem intended.

gregorydemay: 🤖 Good catch — fixed in 5d422152f3. `insert` now checks for a live entry under the key *before* calling `evict_expired`, so a rejected refresh (`AlreadyPresent`) leaves the map untouched instead of silently evicting other expired entries and discarding them. Behaviour is otherwise unchanged (a key that survives `evict_expired` is exactly a live key). Added a test: `should_not_evict_expired_entries_when_rejecting_a_live_refresh`.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+            by_time: BTreeMap::new(),
+        }
+    }
+
+    /// Insert `value` under `key`. Returns an [`InsertError`] if the key already has a live entry
+    /// ([`InsertError::AlreadyPresent`] — refreshing a live entry is not allowed, so a caller cannot
+    /// extend an entry's lifetime by re-inserting it; the map is left unchanged) or, after evicting
+    /// any entries that have outlived their `ttl`, the map is still full of live entries
+    /// ([`InsertError::AtCapacity`] — a live entry is never evicted, since a still-armed address may
+    /// already have received funds). On success, returns the expired entries evicted by this call.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good point. You are right that the rejected-live-refresh path returns before `evict_expired`. Clarified the ordering in the struct-level doc comment (`af88c1028b`) and updated the PR description to reflect the actual order: a live-key re-insert is rejected up front leaving the map unchanged; expired-entry eviction happens only when admitting a genuinely new key. The `insert` method doc already documented this precisely.

PR #10709, opened 2026-07-09
feat: use rules_qemu for local system tests

nmattia:
This removes the `qemu-` dependencies from the build container and instead pulls the qemu binaries in via the hermetic `rules_qemu`.


tmu0 (approved):

basvandijk (approved):
Nice!

PR #10712, opened 2026-07-09
feat: Add subnet size and cycle cost schedule to HTTP request context

eichhorl:
This is needed to correctly calculate part of the new pricing.

Technically we could retrieve both of the new fields via the registry version (which is already in the context), but there is no registry client available in the execution environment.

For compatibility reasons, the registry version is set to `0`, and the cycle cost schedule is introduced as optional (`None`) for now.

### DIFF
--- a/rs/types/types/src/canister_http.rs
+++ b/rs/types/types/src/canister_http.rs
@@ -60,6 +60,7 @@ use ic_management_canister_types_private::{
 };
 use ic_protobuf::{
     proxy::{ProxyDecodeError, try_from_option_field},
+    registry::subnet::v1::CanisterCyclesCostSchedule as CanisterCyclesCostScheduleProto,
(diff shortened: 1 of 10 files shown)


rs/types/types/src/canister_http.rs
```diff
@@ -610,6 +633,10 @@ impl CanisterHttpRequestContext {
             // based on the request's payment and the base fee.
             refund_status: RefundStatus::default(),
             registry_version,
+            // TODO: populate with the actual subnet size this request is processed at.
```
michael-weigelt, line 636: Are these (4) todos here because this change needs a 2 phase rollout?

eichhorl: Yes, exactly

michael-weigelt (approved):

PR #10713, opened 2026-07-09
fix(registry): adapt delete_subnet_test to non-System subnet deletion

basvandijk:
## Problem

`//rs/tests/nns:delete_subnet_test` has been failing since 65ad179d4ae53ff4244d57829f5cfe368e680f0f (#10683) with:

```
thread 'main' panicked at rs/tests/nns/delete_subnet_test.rs:207:14:
called `Result::unwrap_err()` on an `Ok` value: ()
```

That commit lifted the CloudEngine-only restriction in `do_delete_subnet`: governance may now delete any non-System subnet, and System subnet deletion is refused with a new error message. The registry unit tests were updated, but the system test still asserted the old behavior. Since the test's universal canister is installed at the governance canister ID, deleting the Application/VerifiedApplication subnets now succeeds, making the test's `unwrap_err()` panic.

## Fix

Update the system test to the new semantics:

- Governance successfully deletes the CloudEngine, Application, and VerifiedApplication subnets.
- Deleting the System (NNS) subnet is refused with `System subnets may not be deleted`.
- Assert that only the NNS subnet remains, that the routing table entries of all deleted subnets are gone, and that the nodes of all deleted subnets are unassigned and wiped.
- Replace the hardcoded registry version `2` with the actual latest registry version, robust against the additional deletions.

The engine-controller-only path is covered by the PocketIC unit tests in `rs/registry/canister/tests/delete_subnet.rs` and is not exercisable from this system test (the UC occupies the governance canister ID).

## Validation

- rustfmt + clippy clean
- `bazel test //rs/tests/nns:delete_subnet_test` passes on farm:
  - setup PASSED (69.8s), test PASSED (21.6s), assert_no_metrics_errors PASSED, assert_no_unallowed_log_patterns PASSED

### DIFF
--- a/rs/tests/nns/delete_subnet_test.rs
+++ b/rs/tests/nns/delete_subnet_test.rs
@@ -1,7 +1,8 @@
 /* tag::catalog[]
 Title:: Delete Subnet
 
-Goal:: Ensure that CloudEngines can be deleted, and that regular App and System subnets cannot be deleted.
+Goal:: Ensure that governance can delete non-System subnets (CloudEngine, Application, and
+VerifiedApplication subnets), and that System subnets cannot be deleted.
 
 end::catalog[] */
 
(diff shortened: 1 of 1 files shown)


basvandijk:
Force merging since this will fix the broken master.

rs/tests/nns/delete_subnet_test.rs
```diff
@@ -128,53 +136,69 @@ pub fn test(env: TestEnv) {
...
         assert!(!final_subnets.contains(&engine_subnet.subnet_id));
-        assert_eq!(final_subnets.len(), 3);
-
-        // The subnet record and routing table entries of the engine should be gone.
-        let routing_table = new_topology_snapshot.subnet_canister_ranges(engine_subnet.subnet_id);
-        assert!(routing_table.is_empty());
+        assert!(!final_subnets.contains(&app_subnet.subnet_id));
+        assert!(!final_subnets.contains(&vapp_subnet.subnet_id));
+        assert!(final_subnets.contains(&nns_subnet.subnet_id));
+        assert_eq!(final_subnets.len(), 1);
```
daniel-wong-dfinity-org-twin, line 175: FYI

No need to follow up, but a possibly better way to do this is

```
assert_eq!(final_subnets, vec![nns_subnet_id]);
```

On advantage of this is that if the assert fails, the diagnostics might be more useful. E.g. instead of "got 0; expected 1", we get "got vec![a, nns, c]; expected vec![nns]".

daniel-wong-dfinity-org-twin (commented):
Thank you!

rs/tests/nns/delete_subnet_test.rs
basvandijk: That makes sense: https://github.com/dfinity/ic/pull/10722

PR #10714, opened 2026-07-10
refactor: remove legacy canister_log and drop log memory store feature flag

mraszyk:
The one-time migration from the legacy `CanisterLog` store to the `LogMemoryStore` has completed on all subnets, so the migration scaffolding and the legacy store can be removed.

- Drop the `log_memory_store_feature` flag (and the `LOG_MEMORY_STORE_FEATURE`/`_ENABLED` consts); the log memory store is now always enabled.
- Remove the `SystemState.canister_log` field and the `CanisterState` log wrappers now operate solely on `log_memory_store`.
- Remove the migration scaffolding: the per-canister `migrated` flag (`is_migrated`/`set_migrated`/`clear_migrated`), the subnet-level `logs_migrated` metadata, the scheduler migration code, and the migration-duration metric. Reads, writes and metrics now use `log_memory_store` unconditionally.
- Stop serializing `canister_log_records`, `next_canister_log_record_idx` and `log_memory_store_migrated`; the corresponding proto fields (43, 44, 66) are marked reserved.
 - Replace the `StateMachine::canister_log test` helper (which rebuilt a `CanisterLog` aggregate) with `canister_log_records` and `canister_log_bytes_used`, reading directly from `LogMemoryStore`. Add a test-only `LogMemoryStore::all_records_for_testing` that returns all records without the response-size trimming that `records()` applies, and migrate all callers.

The `CanisterLog` type and the transient `SystemStateModifications` delta log buffer are retained.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
@@ -393,16 +393,10 @@ impl SystemStateModifications {
     ) -> HypervisorResult<RequestMetadataStats> {
         // Append delta logs.
         if !self.canister_log.is_empty() {
-            // TODO(DSM-11): Move this into append_delta_log() once there is only one of it.
             metrics.observe_delta_log_size(self.canister_log.bytes_used());
         }
-        if system_state.log_memory_store.is_migrated() {
-            system_state
-                .log_memory_store
-                .append_delta_log(&mut self.canister_log.clone());
-        }
         system_state
-            .canister_log
+            .log_memory_store
             .append_delta_log(&mut self.canister_log);
 
         // Verify total cycle change is not positive and update cycles balance.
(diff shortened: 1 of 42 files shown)


rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -393,16 +393,10 @@ impl SystemStateModifications {
     ) -> HypervisorResult<RequestMetadataStats> {
         // Append delta logs.
         if !self.canister_log.is_empty() {
-            // TODO(DSM-11): Move this into append_delta_log() once there is only one of it.
```
mraszyk, line 396: It is easier to keep it here since `append_delta_log` is also called in tests and dummy metrics would have to be passed at all those places.

mbjorkqvist (approved):

alin-at-dfinity (approved):

rs/execution_environment/src/execution/install_code.rs
```diff
@@ -618,12 +609,8 @@ impl InstallCodeHelper {
...
     /// Takes the canister log.
-    pub(crate) fn take_canister_log(&mut self) -> (CanisterLog, LogMemoryStore) {
-        // TODO(DSM-105): Remove duplication when the migration is fully done.
-        (
-            self.canister.system_state.canister_log.take(),
-            self.canister.system_state.log_memory_store.clone(),
-        )
+    pub(crate) fn take_canister_log(&mut self) -> LogMemoryStore {
+        self.canister.system_state.log_memory_store.clone()
     }
```
eichhorl: I guess technically this no longer "takes" the canister log, but clones it. Maybe we could update the name and documentation. Also I think we can change the argument to just `&self`

mraszyk: https://github.com/dfinity/ic/pull/10714/commits/9803bf1e579a350bf093c74fbc5ab34a951f7ab5

eichhorl (approved):

PR #10723, opened 2026-07-10
fix(nns): Drive node provider rewards mint to completion in golden state upgrade test

pietrodimarco-dfinity:
## Problem

`test_upgrade_canisters_with_golden_nns_state` was failing in CI with:

```
After upgrading and advancing time, node provider rewards timestamp did not increase.
Before: 1781451541, After: 1781451541
```

(panic at `rs/nns/integration_tests/src/upgrade_canisters_with_golden_nns_state.rs:556`)

## Root cause

`advance_time_to_allow_for_voting_and_node_rewards` jumps the state machine clock ~1–2 months forward to trigger the monthly node provider rewards distribution, then ticks a **fixed 100 times** expecting the mint to have completed. That is not enough ticks:

- The mint (in the governance heartbeat) can only validate once the **node-rewards canister has synced** its node metrics up to the requested period — its sync runs on an hourly timer, so it needs several ticks to fire and complete after the abrupt time jump. Until then the node-rewards canister rejects the request with `"Metrics and registry are not synced up to to_date"`.
- Once it validates, the mint makes **one inter-canister call per day** in the reward period (which spans ~55 days here, because the golden state's last reward is well behind the test's execution date) plus **one ledger transfer per node provider**.

When the mint does not finish within the 100-tick budget, the reward timestamp is never updated (the mint error is only logged, not fatal), and the `assert_increased` check on the node provider rewards timestamp fails. This gets worse the staler the pinned golden state is relative to the test's run date.

## Fix

Replace the fixed `for _ in 0..100` tick loop with `tick_until_node_provider_rewards_distributed`, which ticks (1s/tick) and polls the reward timestamp until a new reward is actually distributed (timestamp increases past the previous one), up to a generous cap (`MAX_TICKS = 2000`). This makes the wait adaptive to how stale the golden state is, while still failing with the same informative assertion in `check_all` if no reward is ever distributed.

Behavior for the passing case is preserved: the distribution is still triggered at the same moment; only the wait is now adaptive instead of a fixed count.

## Testing

- `rustfmt` clean.
- `bazel build //rs/nns/integration_tests:upgrade_canisters_with_golden_nns_state` passes.
- The test itself is a `manual`/nightly golden-state test (requires downloading the golden NNS state) and was not run locally.


rs/nns/integration_tests/src/upgrade_canisters_with_golden_nns_state.rs
```diff
@@ -471,6 +478,36 @@ mod sanity_check {
...
+    /// Ticks the state machine (advancing time by one second per tick) until a new
+    /// monthly node provider reward has been distributed, i.e. its timestamp has
+    /// increased past `before_timestamp`, up to a generous cap on the number of ticks.
+    fn tick_until_node_provider_rewards_distributed(
+        state_machine: &StateMachine,
+        before_timestamp: u64,
+    ) {
+        // Generous upper bound. Reaching the reward distribution requires the
+        // node-rewards canister to sync (hourly timer) plus a minting call whose length
+        // scales with the reward period and the number of node providers, so this needs
```
daniel-wong-dfinity-org-twin, line 490: Please, give some idea about how many node providers there are, because this is NOT common knowledge. E.g.

> As of July, 2026, there are 104 node providers. The current number can be found by going to https://dashboard.internetcomputer.org/network/providers

That way, we can have a better understanding of why "few hundred rounds" should be sufficient, and why this 2_000 is "very generous".

rs/nns/integration_tests/src/upgrade_canisters_with_golden_nns_state.rs
```diff
@@ -471,6 +478,36 @@ mod sanity_check {
...
+        // to comfortably exceed a few hundred rounds.
+        const MAX_TICKS: usize = 2000;
```
daniel-wong-dfinity-org-twin: you might want to change this to a "nice round" number like 3_600. This is "round" in the sense that since in the main loop, time is advanced by 1 s, this keeps trying for 1 hr of simulated time, which intuitively seems like an "improbably long" time, and therefore "very generous".

daniel-wong-dfinity-org-twin (approved):

rs/nns/integration_tests/src/upgrade_canisters_with_golden_nns_state.rs
pietrodimarco-dfinity: Good call, done in 0343852 — changed the cap to 3600 (one hour of simulated time) and updated the comment to explain the reasoning.

PR #10727, opened 2026-07-10
feat: Credit delivered HTTP refunds to canisters and report cycles metric

eichhorl:
https://github.com/dfinity/ic/pull/10897 added the infrastructure for delivering initial and asynchronous HTTP outcalls refunds to the DSM. For now, only initial refunds are delivered.

This PR implements the DSM side, which is responsible for:
1. Updating the context's `RefundStatus` based on the delivered refunds
2. Crediting the refunded cycles to the calling canister of the HTTP outcall
3. Reporting the amount of cycles spent (both normal and free subnets) as a metric.

The refund lifecycle of an HTTP context happens after a response to the canister was delivered: When a response is delivered, the context is removed from `canister_http_request_contexts`. If the context uses pay-as-you-go pricing, then the context is instead re-inserted into the `delivered_canister_http_request_contexts` collection.

Refunds may only be applied for contexts residing in the latter collection, which is why it happens at the end of the execution round.

While the context is in `delivered_canister_http_request_contexts`, delivered refunds may be applied, respecting the following invariants:
1. Initial refunds must be applied first, while `refunding_nodes` is still empty. 
2. A node may not issue refunds multiple times.
3. Collectively, not more than `refundable_cycles = committee_size * per_replica_allowance` may be refunded.

After a timeout of 2 minutes (after the original context was created), contexts in `delivered_canister_http_request_contexts` time out, and refund the full allowance for all replicas that haven't responded so far.


rs/messaging/src/message_routing.rs
```diff
@@ -359,6 +360,9 @@ pub(crate) struct MessageRoutingMetrics {
     /// Metrics for query stats aggregator
     pub query_stats_metrics: QueryStatsAggregatorMetrics,
 
+    /// Metrics for the accounting of HTTP outcall spend reports and refunds.
```
mraszyk:
```suggestion
    /// Metrics for the accounting of HTTP outcall (cycles) spent reports and refunds.
```

rs/replicated_state/src/metadata_state/subnet_call_context_manager.rs
```diff
@@ -318,6 +326,14 @@ impl SubnetCallContextManager {
                             context.request.sender_reply_callback,
                             context.request.sender
                         );
+                        if context.pricing_version == PricingVersion::PayAsYouGo {
+                            // If the pricing version is pay-as-you-go, move the context
+                            // to the delivered contexts. This lets us keep accounting
+                            // late per-replica spend reports (and refund the replicas
```
mraszyk:
```suggestion
                            // late per-replica spend reports (and refund cycles from the replicas
```

rs/replicated_state/src/metadata_state/tests.rs
```diff
@@ -977,6 +978,121 @@ fn subnet_call_contexts_deserialization() {
...
+            .is_empty()
+    );
+    assert_eq!(2, manager.delivered_canister_http_request_contexts.len());
+
+    // Exactly at its timeout, only the first context is timed out.
+    let at_first_timeout = UNIX_EPOCH + DELIVERED_CANISTER_HTTP_REQUEST_CONTEXT_TIMEOUT;
+    assert_eq!(
+        vec![contexts[0].clone()],
+        manager.time_out_delivered_canister_http_request_contexts(at_first_timeout)
+    );
```
mraszyk, line 1085:
```suggestion
    );
    assert_eq!(1, manager.delivered_canister_http_request_contexts.len());
```

rs/messaging/src/state_machine.rs
```diff
@@ -238,6 +247,26 @@ impl StateMachine for StateMachineImpl {
...
         }
+
+        // Apply HTTP outcall spend reports and time out delivered request
+        // contexts. This runs after execution, so that contexts that were just
+        // responded to during the round have already been moved into the
+        // delivered collection and can therefore receive their refunds and have
+        // their consumed cycles reported.
+        deliver_canister_http_spent(
+            &canister_http_spent,
+            &mut state_after_execution,
```
mraszyk: we could consider flipping these two args to be consistent with the arg order in the subsequent refund_timed_out_canister_http_contexts

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+    consumed: NominalCycles,
+}
+
+impl CanisterAccounting {
+    /// Accumulates `amount` as consumed by the caller, according to the cost
+    /// schedule of the context the amount was reported for.
+    fn observe_consumed(&mut self, amount: Cycles, cost_schedule: CanisterCyclesCostSchedule) {
+        self.consumed += CompoundCycles::<HTTPOutcalls>::new(amount, cost_schedule).nominal();
+    }
+}
```
mraszyk: Why not use `CompoundCycles::<HTTPOutcalls>` directly?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+    {
+        let contexts = &mut state
+            .metadata
+            .subnet_call_context_manager
+            .delivered_canister_http_request_contexts;
+
+        // Initial spending
+        for report in &spent.initial {
+            let Some(context) = contexts.get_mut(&report.callback) else {
+                continue;
```
mraszyk, line 109: isn't this unexpected and thus a critical error should be raised here?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+    }
+
+    apply_accounting(state, accounting, log);
+}
+
+/// Records `amount` as refunded against `refund_status`, capped so that
+/// `refunded_cycles` never exceeds `refundable_cycles`. Returns the amount that
+/// was actually applied (and therefore should be credited to the canister).
+///
+/// Capping is not expected to happen (the per-replica allowances are sized so
```
mraszyk, line 262: So a refund report by a malicious node that reports an overly high refund amount would be caught earlier?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+                .extend(report.nodes.iter());
+            let entry = accounting.entry(context.request.sender).or_default();
+            entry.refund += applied;
+            entry.observe_consumed(report.amount, context.cost_schedule);
+        }
+
+        // Asynchronous spending
+        for report in &spent.asynchronous {
+            let Some(context) = contexts.get_mut(&report.callback) else {
+                continue;
```
mraszyk, line 149: isn't this unexpected and thus a critical error should be raised here?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+        (1..=SUBNET_SIZE).map(node_test_id).collect()
+    }
+
+    /// Builds a state holding a single caller canister (`canister_test_id(1)`) on
+    /// a [`SUBNET_SIZE`]-node subnet and, if `context` is `Some`, a delivered
+    /// `CanisterHttpRequestContext` registered under [`CALLBACK`], with a
+    /// [`CanisterCyclesCostSchedule::Normal`] cost schedule and the refund
+    /// status of a request that paid the given refundable cycles (see
+    /// [`refund_status()`]).
+    fn setup(context: Option<(Replication, Cycles)>) -> (ReplicatedState, CanisterId) {
```
mraszyk, line 404: the replication is not mentioned in the doc comment

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+    }
+
+    /// The cycles reported as consumed for HTTPS outcalls at the subnet level,
+    /// both in the dedicated field and in the by-use-case map. The two must
+    /// always agree, as they are observed together.
+    fn subnet_consumed(state: &ReplicatedState) -> u128 {
+        let subnet_metrics = &state.metadata.subnet_metrics;
+        let http_outcalls = subnet_metrics.get_consumed_cycles_http_outcalls();
+        let by_use_case = subnet_metrics
+            .get_consumed_cycles_by_use_case()
```
mraszyk: could further extend this by checking that get_consumed_cycles_by_use_case_as_counters also agree

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+                nodes: node_set(&[1, 2, 3]),
+            }],
+            asynchronous: vec![],
+        };
+        deliver_canister_http_spent(&report, &mut state, &no_op_logger(), &metrics);
+
+        assert_eq!(balance(&state, caller), INITIAL_BALANCE);
+        assert_eq!(consumed(&state, caller), spent.get());
+        assert_eq!(subnet_consumed(&state), spent.get());
+        assert_eq!(get_refund_status(&state).refunded_cycles, Cycles::zero());
```
mraszyk:
```suggestion
        assert_eq!(get_refund_status(&state).refunded_cycles, Cycles::zero());
        assert_eq!(status.refunding_nodes.len(), 3);
```

mraszyk: same below

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+        assert_errors(&[], &metrics_registry);
+    }
+
+    /// Same as above, for asynchronous reports: the per-node spend is reported
+    /// as consumed, while the (zero) per-replica allowance refunds nothing.
+    #[test]
+    fn free_subnet_asynchronous_report_reports_consumed_without_refund() {
```
mraszyk, line 810: normal subnet and asynchronous reports should also be tested on the happy path

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+            asynchronous: vec![CanisterHttpAsyncSpent {
+                callback: CALLBACK,
+                shares: BTreeMap::from([
+                    (node_test_id(1), Cycles::new(400)),
+                    (node_test_id(2), Cycles::new(600)),
+                ]),
+            }],
+        };
+        deliver_canister_http_spent(&first, &mut state, &log, &metrics);
+        // refund = (1_000 − 400) + (1_000 − 600) = 1_000; consumed = 1_000.
```
mraszyk: the test would be stronger if refund is not equal to consumed

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+        assert_eq!(
+            balance(&state, caller),
+            INITIAL_BALANCE + Cycles::new(1_000)
+        );
+        assert_eq!(consumed(&state, caller), 1_000);
```
mraszyk:
```suggestion
        assert_eq!(consumed(&state, caller), 1_000);
        assert_eq!(get_refund_status(&state).refunding_nodes.len(), 2);
```

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1094 @@
...
+                nodes: all_nodes(),
+            }],
+            asynchronous: vec![],
+        };
+        deliver_canister_http_spent(&report, &mut state, &no_op_logger(), &metrics);
+
+        assert!(state.canister_state(&caller).is_none());
+        // Nothing is reported at the subnet level either, as the cycles were never
+        // credited back to (nor consumed by) an existing canister.
+        assert_eq!(subnet_consumed(&state), 0);
```
mraszyk: The subnet-level metrics also carry cycles consumed by deleted canisters so we should account for them (otherwise, some cycles minted by the CMC would be "lost" from the metrics).

rs/replicated_state/src/metadata_state/tests.rs
eichhorl: Added

rs/messaging/src/canister_http_spent.rs
eichhorl: Good point, added

rs/messaging/src/canister_http_spent.rs
eichhorl: Thanks!

rs/messaging/src/canister_http_spent.rs
eichhorl: Nodes no longer report the refund amount. They now report how much they spent. So even if a node reports the maximum amount x (which is 1T on free subnets, and the `per_replica_allowance` on normal subnets), the refund is calculated by `per_replica_allowance - x`. On free subnets, the `per_replica_allowance` is zero.

This means, the refund here can be at most the `per_replica_allowance`, which is calculated as `refundable_cycles / committee_size`. So assuming we correctly only accept one share per committee member, it should be impossible to exceed `refundable_cycles`.

rs/messaging/src/canister_http_spent.rs
eichhorl: see above

rs/messaging/src/canister_http_spent.rs
eichhorl: For now, it is still expected that we end up in this branch for all outcalls (as they still use legacy pricing). We only insert contexts with pay-as-you-go pricing into `delivered_canister_http_request_contexts`, but we already started delivering refunds even for legacy pricing (although they are all zero).

rs/messaging/src/canister_http_spent.rs
eichhorl: Do you mean instead of `CanisterAccounting`? That wouldn't work, because `refund` and `consumed` have different semantics. We could use it instead of `NominalCycles` for consumed, but we don't really need the `real` part anyway.

rs/messaging/src/message_routing.rs
eichhorl: Added

rs/replicated_state/src/metadata_state/subnet_call_context_manager.rs
eichhorl: Added

rs/messaging/src/state_machine.rs
eichhorl: Done 👍

rs/messaging/src/canister_http_spent.rs
eichhorl: Done 👍

rs/messaging/src/canister_http_spent.rs
eichhorl: Good point 👍

rs/messaging/src/canister_http_spent.rs
mraszyk: > because refund and consumed have different semantics.

of course; I just saw the pattern and made the proposal...

rs/messaging/src/canister_http_spent.rs
eichhorl: Done

rs/messaging/src/canister_http_spent.rs
mraszyk: I see and without the context, we can't really check if the canister http outcall was legacy...

rs/messaging/src/canister_http_spent.rs
eichhorl: Now we're reporting the "spent" cycles as consumed by the subnet, and the "refunded" cycles as lost by deleted canisters. PTAL

rs/messaging/src/canister_http_spent.rs
mraszyk: I see, so this could only happen due to a bug in the code, not due to a malicious node. Not sure if it's worthwhile to point out in the comment.

rs/messaging/src/canister_http_spent.rs
eichhorl: How do we differentiate deleted canisters from canisters that were migrated or split to a different subnet?

mraszyk: > How do we differentiate deleted canisters from canisters that were migrated or split to a different subnet?

"Deleted" canisters are those for which the `delete_canister` endpoint of the management canister was invoked, i.e., migrated canisters or canisters split to a different subnet do not qualify as "deleted". But I think it's better to "misuse" the "deleted" canister case than lose these cycles from metrics and have the cycles alert drift away by not having the cycles recorded [here](https://github.com/dfinity-ops/k8s/blob/d9a91826cec161dff8945da4aea5f079cdbb9333/bases/apps/ic-metrics-configs/rules/rules.yaml#L1806).

mraszyk: The cycles should be recorded in consumed_cycles_http_outcalls, consumed_cycles_by_use_case[HttpOutcalls], and consumed_cycles_by_use_case_as_counters[HttpOutcalls] in SubnetMetrics.

mraszyk:
```
subnet_metrics.observe_consumed_cycles_with_use_case(
            CyclesUseCase::DeletedCanisters,
            lost_by_deleted_canisters,
        );
        subnet_metrics.observe_consumed_cycles_by_deleted_canisters(lost_by_deleted_canisters);
```
is not ideal because this is not what would happen if the canister existed, got the cycles refunded, and only then got deleted.

  Why: at the canister level, observe_consumed_cycles_for_https_outcall (rs/replicated_state/src/canister_state/system_state.rs:2138) writes only into the canister's
  consumed_cycles_by_use_cases_as_counters map — it touches neither the canister's consumed_cycles scalar nor its consumed_cycles_by_use_cases map. delete_canister
  (rs/execution_environment/src/canister_manager.rs:1330-1367) reads only those two:

  - consumed_cycles_by_deleted_canisters += leftover_cycles + canister_metrics().consumed_cycles() — HTTP outcalls absent from the second term,
  - the loop re-observing consumed_cycles_by_use_cases() into the subnet map — no HTTPOutcalls entry to re-observe.

So canister http outcall cycles do not get folded into CyclesUseCase::DeletedCanisters when a canister is deleted.

eichhorl: Do you mean like this: https://github.com/dfinity/ic/pull/10727/commits/1fbe7614d36bc745fea9433dbc46dd5554af7d93?

I guess ultimately for this feature to work cleanly with subnet splitting, we would have to start splitting also the `(delivered_)canister_http_request_contexts`. Otherwise all cycles attached to the call would be lost, because the excess is no longer refunded upfront.

mraszyk: > Do you mean like this: https://github.com/dfinity/ic/commit/1fbe7614d36bc745fea9433dbc46dd5554af7d93?

Yes, but now I realized that if the canister got only deleted after processing the refund, the flow would have been:
- report accounting.consumed as "consumed" in the canister counter
- credit accounting.refund to the canister balance
- [delete canister]: merge balance into CyclesUseCase::DeletedCanisters and the scalar consumed_cycles_by_deleted_canisters

So what you had before was actually the right approach... I apologize!

mraszyk: Essentially I confused accounting.consumed (which has always been correctly added to subnet_consumed) and accounting.refund (which should end up in the consumed cycles by deleted canisters after all).

mraszyk: > I guess ultimately for this feature to work cleanly with subnet splitting, we would have to start splitting also the (delivered_)canister_http_request_contexts. Otherwise all cycles attached to the call would be lost, because the excess is no longer refunded upfront.

I'd expect pending canister http outcalls to be rejected upon subnet split (at least on the "new" subnet that was forked off) and at that point we can refund everything back (it is not the canister's fault that the subnet was split).

eichhorl: > before was actually the right approach

Reverted!

> I'd expect pending canister http outcalls to be rejected upon subnet split

Yes, either that, or properly move the contexts to the new subnet. Neither of these are implemented today (the contexts simply stay untouched on the original subnet). I will take a look in a follow up.

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+};
+use prometheus::IntCounterVec;
+use std::collections::BTreeMap;
+
+const METRIC_ACCOUNTING_ERRORS_TOTAL: &str = "mr_canister_http_accounting_errors_total";
+
+const LABEL_ERROR: &str = "error";
+
+/// An initial spend report was received for a callback for which some node(s)
+/// had already been accounted (through an asynchronous report).
```
mraszyk: > through an asynchronous report

or a duplicate initial report, no?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+///    once (tracked via `refunding_nodes`), which makes these idempotent.
+pub(crate) fn deliver_canister_http_spent(
+    state: &mut ReplicatedState,
+    spent: &CanisterHttpSpent,
+    log: &ReplicaLogger,
+    metrics: &CanisterHttpSpentMetrics,
+) {
+    // First update the contexts' refund status and accumulate the amounts to
+    // credit/report per canister. The crediting happens in a second pass to
+    // avoid borrowing both the subnet call context manager and the canister
```
mraszyk: why do we avoid borrowing both at the same time?

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+    let timed_out = state
+        .metadata
+        .subnet_call_context_manager
+        .time_out_delivered_canister_http_request_contexts(current_time);
+
+    let mut accounting: BTreeMap<CanisterId, CanisterAccounting> = BTreeMap::new();
+    for (callback, mut context) in timed_out {
+        // The number of replicas assigned to the request; the responders already
+        // refunded via `deliver_canister_http_spent`, the rest refund in full
+        // here.
```
mraszyk: this comment reads a bit strange

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+                    log,
+                    "Canister {} for an HTTP outcall no longer exists; its refund of {} cycles is \
+                     lost and reported as consumed by deleted canisters (consumed {} cycles).",
+                    sender,
+                    accounting.refund,
+                    accounting.consumed
+                );
+                lost_by_deleted_canisters += CompoundCycles::<HTTPOutcalls>::new(
+                    accounting.refund,
+                    CanisterCyclesCostSchedule::Normal,
```
mraszyk, line 333: let's please add a note why we pass Normal here

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+            map.get(&use_case)
+                .copied()
+                .unwrap_or_else(NominalCycles::zero)
+        };
+        let gauge = get(subnet_metrics.get_consumed_cycles_by_use_case());
+        assert_eq!(
+            gauge,
+            get(subnet_metrics.get_consumed_cycles_by_use_case_as_counters())
+        );
+        if use_case == CyclesUseCase::HTTPOutcalls {
```
mraszyk: CyclesUseCase::DeletedCanisters could be handled here, too, for consistency

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+                    (node_test_id(1), Cycles::new(400)),
+                    (node_test_id(2), Cycles::new(600)),
+                ]),
+            }],
+        };
+        deliver_canister_http_spent(&mut state, &report, &no_op_logger(), &metrics);
+
+        assert_eq!(balance(&state, caller), INITIAL_BALANCE);
+        assert_eq!(consumed(&state, caller), 1_000);
+        assert_eq!(subnet_consumed(&state), 1_000);
```
mraszyk: these two values should be different to make the test stronger

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1209 @@
...
+    fn free_subnet_timeout_refunds_nothing() {
+        let (mut state, caller) =
+            setup_free_subnet(Some((Replication::FullyReplicated, Cycles::zero())));
+        let (metrics_registry, metrics) = metrics();
+
+        let timeout = UNIX_EPOCH + Duration::from_secs(3 * 60); // > 2min timeout.
+        refund_timed_out_canister_http_contexts(&mut state, timeout, &no_op_logger(), &metrics);
+
+        assert_eq!(balance(&state, caller), INITIAL_BALANCE);
+        assert_eq!(consumed(&state, caller), 0);
```
mraszyk, line 848:
```suggestion
        assert_eq!(consumed(&state, caller), 0);
        assert_eq!(subnet_consumed(&state, caller), 0);
```

mraszyk (commented):
Is the initial report always delivered together with the response? Otherwise, a malicious node could produce an "asynchronous" report earlier and make the initial report ignored.

rs/messaging/src/canister_http_spent.rs
eichhorl: Good point, fixed

rs/messaging/src/canister_http_spent.rs
eichhorl: The borrow checker forces us to: We can't hold a mutable reference to `delivered_canister_http_request_contexts`, while also calling `apply_accounting` which takes `&mut state`. But the comment is more confusing than helpful, so I removed it.

rs/messaging/src/canister_http_spent.rs
eichhorl: Reworded :+1:

rs/messaging/src/canister_http_spent.rs
eichhorl: I tried to answer this here: https://github.com/dfinity/ic/commit/1fbe7614d36bc745fea9433dbc46dd5554af7d93#r194197739:

> I guess technically it should be the cost schedule of this specific outcall which may be different than the one of the subnet. Specifically, system subnets have normal cost schedule, but HTTP outcalls are still free, so the context is created with free cost schedule.
>
> Either way, the cost schedule doesn't affect the nominal part of the `CompoundCycles` so shouldn't really matter. Does anything speak against adding a proper constructor for `NominalCycles` and using that instead?

What do you think of a `NominalCycles` constructor?

rs/messaging/src/canister_http_spent.rs
eichhorl: Good idea

rs/messaging/src/canister_http_spent.rs
eichhorl: Done

rs/messaging/src/canister_http_spent.rs
eichhorl: Added

eichhorl:
> Is the initial report always delivered together with the response? Otherwise, a malicious node could produce an "asynchronous" report earlier and make the initial report ignored.

Yes, the initial report is derived from the same artifact as the response

rs/messaging/src/canister_http_spent.rs
mraszyk: > What do you think of a NominalCycles constructor?

I agree it is cleaner to introduce it.

eichhorl: Done: https://github.com/dfinity/ic/pull/10727/commits/002521b07fdfab69b80fe645b645d3cdb935020c

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+                callback: CALLBACK,
+                amount: spent,
+                nodes: node_set(&[1, 2, 3]),
+            }],
+            asynchronous: vec![],
+        };
+        deliver_canister_http_spent(&mut state, &report, &no_op_logger(), &metrics);
+
+        assert_eq!(balance(&state, caller), INITIAL_BALANCE);
+        assert_eq!(consumed(&state, caller), spent.get());
```
mraszyk, line 736: This means we'll "manufacture" cycles in metrics because minted is no longer equal to balance + consumed on "normal" cost schedule subnets. But since it is unexpected and the logs contain the cycles amounts (so we could fix the metrics afterwards), I think we can accept this fact.

eichhorl: Sounds good 👍

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+    // `apply_accounting()`. This aggregates all reports of a canister into a single
+    // balance and metric update.
+    let mut accounting: BTreeMap<CanisterId, CanisterAccounting> = BTreeMap::new();
+    {
+        let contexts = &mut state
+            .metadata
+            .subnet_call_context_manager
+            .delivered_canister_http_request_contexts;
+
+        // Initial spending
```
mraszyk: Add a comment that this must be processed before asynchronous reports in case a node produced an asynchronous report in the same block as the initial report for a callback.

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+                    );
+                    let refund = allowance - spent.real();
+                    entry.refund += apply_capped(
+                        &mut context.refund_status,
+                        refund,
+                        report.callback,
+                        log,
+                        metrics,
+                    );
+                    entry.consumed += spent.nominal();
```
mraszyk: This part can now be moved into a helper shared by initial and asynchronous reports.

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+                     lost and reported as consumed by deleted canisters (consumed {} cycles).",
+                    sender,
+                    accounting.refund,
+                    accounting.consumed
+                );
+                lost_by_deleted_canisters += CompoundCycles::<HTTPOutcalls>::new(
+                    accounting.refund,
+                    CanisterCyclesCostSchedule::Normal,
+                )
+                .nominal();
```
mraszyk, line 335: here the constructor for NominalCycles would still be useful.

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+        let mut state = ReplicatedStateBuilder::new()
+            .with_canister(canister)
+            .with_node_ids(all_nodes().into_iter().collect())
+            .build();
+
+        let consumed_before = CompoundCycles::<HTTPOutcalls>::new(
+            Cycles::new(SUBNET_CONSUMED_BEFORE),
+            CanisterCyclesCostSchedule::Normal,
+        )
+        .nominal();
```
mraszyk, line 469: another use case for NominalCycles::new

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+            INITIAL_BALANCE + Cycles::new(3_500)
+        );
+        assert_eq!(consumed(&state, caller), spent.get());
+        // The spend is also reported at the subnet level, like the base fee.
+        assert_eq!(
+            subnet_consumed(&state),
+            SUBNET_CONSUMED_BEFORE + spent.get()
+        );
+        let status = get_refund_status(&state);
+        assert_eq!(status.refunded_cycles, Cycles::new(3_500));
```
mraszyk, line 669: We could also test that refundable_cycles do not decrease.

rs/messaging/src/canister_http_spent.rs
```diff
@@ -0,0 +1,1330 @@
...
+            Replication::FullyReplicated,
+            raw_refund_status(refundable, allowance),
+        );
+        let (metrics_registry, metrics) = metrics();
+
+        // amount = 0 → uncapped refund would be 13 * 1_000 = 13_000.
+        let report = CanisterHttpSpent {
+            initial: vec![CanisterHttpInitialSpent {
+                callback: CALLBACK,
+                amount: Cycles::zero(),
```
mraszyk: we could use a non-zero amount and assert that it is still reported in the metrics

mraszyk (approved):
Leaving some final comments, but let's maybe apply them in a follow-up PR to not drag this one along for too long.

rs/messaging/src/canister_http_spent.rs
eichhorl: Done

rs/messaging/src/canister_http_spent.rs
eichhorl: Good idea, done

rs/messaging/src/canister_http_spent.rs
eichhorl: Will take a look in a follow up

rs/messaging/src/canister_http_spent.rs
eichhorl: Will take a look in a follow up

rs/messaging/src/canister_http_spent.rs
eichhorl: Added

rs/messaging/src/canister_http_spent.rs
eichhorl: Done

PR #10728, opened 2026-07-10
feat(release-runscript): set engine-controller authorized_caller on upgrade

pietrodimarco-dfinity:
## What

The `release-runscript` release tool hardcoded `()` as the engine-controller upgrade argument in the *Create Proposal Texts* step. `()` cannot set the canister's `authorized_caller`, so it was impossible to change the authorized caller through the normal release flow.

This changes the engine-controller branch to prompt for the authorized caller principal and build the upgrade arg:

```candid
(opt record {
    authorized_caller = opt principal "<CALLER>";
    initial_dkg_subnet_id = null
})
```

## Why

The engine-controller's `post_upgrade` re-applies its init args on **every** upgrade (`apply_init_args` in `rs/engine_controller/canister/canister.rs`). If `authorized_caller` is `null`, the canister falls back to its built-in `DEFAULT_AUTHORIZED_CALLER`. So to set (or retain) a specific authorized caller, the value must be supplied on each release — which the tool previously couldn't do.

The candid string is passed as `CANDID_ARGS` to `prepare-nns-upgrade-proposal-text.sh`, which encodes it via `didc encode | xxd -r -p` to compute the proposal's `arg_hash` and the verification section.

### DIFF
--- a/rs/nervous_system/tools/release-runscript/src/main.rs
+++ b/rs/nervous_system/tools/release-runscript/src/main.rs
@@ -318,16 +318,36 @@ fn run_create_proposal_texts(cmd: CreateProposalTexts) -> Result<()> {
             // Some canisters require an explicit upgrade arg. cycles-minting
             // needs a hand-provided one (usually '()'). engine-controller's
             // post_upgrade traps when decoding an empty arg, so it must be
-            // given at least '()'; otherwise the upgrade silently rolls back
-            // while the proposal still reports as executed.
+            // given a non-empty record; otherwise the upgrade silently rolls
+            // back while the proposal still reports as executed. We prompt for
+            // the authorized caller principal and build the record around it.
             let output = match canister.as_str() {
                 "cycles-minting" => {
                     let upgrade_arg = input_with_default("Upgrade arg for CMC?", "()")?;
                     run_script(script, &[canister, &commit, &upgrade_arg], &ic)
                         .expect("Failed to run NNS proposal text script")
                 }
-                "engine-controller" => run_script(script, &[canister, &commit, "()"], &ic)
-                    .expect("Failed to run NNS proposal text script"),
+                "engine-controller" => {
+                    // An empty answer emits `null`, which makes the canister
+                    // fall back to its built-in default authorized caller; a
+                    // non-empty value is wrapped in `opt principal "..."`.
+                    // The canister re-applies this on every post_upgrade, so
+                    // the caller must be supplied each release to retain it
+                    // (otherwise it reverts to the default).
+                    let new_caller = input(
+                        "Authorized caller principal for engine-controller? (leave empty for canister default)",
+                    )?;
+                    let authorized_caller = if new_caller.is_empty() {
+                        "null".to_string()
+                    } else {
+                        format!("opt principal \"{new_caller}\"")
+                    };
+                    let upgrade_arg = format!(
+                        "(opt record {{ authorized_caller = {authorized_caller}; initial_dkg_subnet_id = null }})"
+                    );
+                    run_script(script, &[canister, &commit, &upgrade_arg], &ic)
+                        .expect("Failed to run NNS proposal text script")
+                }
                 _ => run_script(script, &[canister, &commit], &ic)
                     .expect("Failed to run NNS proposal text script"),
             };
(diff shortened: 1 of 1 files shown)


daniel-wong-dfinity-org-twin:
I removed ephemera content from PR description (note to reviewer, and test status sections). PTAL (in case you want to restore any of that content).

Why such content does not belong: The PR (title and) description are used as the main content of the eventual commit message. Therefore, it should not contain information that is only relevant at review time, at least IMHO.

OTOH, GitHub really does not provide a good way to convey such information. What I like to do is reply to my own PR, but this can easily be missed by reviewers.

rs/nervous_system/tools/release-runscript/src/main.rs
```diff
@@ -318,16 +318,39 @@ fn run_create_proposal_texts(cmd: CreateProposalTexts) -> Result<()> {
             // Some canisters require an explicit upgrade arg. cycles-minting
             // needs a hand-provided one (usually '()'). engine-controller's
             // post_upgrade traps when decoding an empty arg, so it must be
-            // given at least '()'; otherwise the upgrade silently rolls back
-            // while the proposal still reports as executed.
+            // given a non-empty record; otherwise the upgrade silently rolls
+            // back while the proposal still reports as executed. We prompt for
```
daniel-wong-dfinity-org-twin, line 322: > while the proposal still reports as executed

This is tangential. Therefore, I would either not mention it, or put it in partheneses. I would also maybe mention that this behavior is a bug.

(We have no plan to fix it, because while it is pretty misleading, it's not "fatal".)

rs/nervous_system/tools/release-runscript/src/main.rs
```diff
@@ -318,16 +318,39 @@ fn run_create_proposal_texts(cmd: CreateProposalTexts) -> Result<()> {
...
+                        "null".to_string()
+                    } else {
+                        let principal = candid::Principal::from_text(&new_caller).map_err(|e| {
+                            anyhow::anyhow!("Invalid principal '{new_caller}': {e}")
+                        })?;
+                        format!("opt principal \"{}\"", principal.to_text())
+                    };
+                    let upgrade_arg = format!(
+                        "(opt record {{ authorized_caller = {authorized_caller}; initial_dkg_subnet_id = null }})"
+                    );
```
daniel-wong-dfinity-org-twin, line 350: Spinning this out. In general, match arms should be short.

rs/nervous_system/tools/release-runscript/src/main.rs
```diff
@@ -318,16 +318,39 @@ fn run_create_proposal_texts(cmd: CreateProposalTexts) -> Result<()> {
...
                 }
-                "engine-controller" => run_script(script, &[canister, &commit, "()"], &ic)
-                    .expect("Failed to run NNS proposal text script"),
+                "engine-controller" => {
+                    // An empty answer emits `null`, which makes the canister
+                    // fall back to its built-in default authorized caller; a
+                    // non-empty value is wrapped in `opt principal "..."`.
+                    // The canister re-applies this on every post_upgrade, so
+                    // the caller must be supplied each release to retain it
+                    // (otherwise it reverts to the default).
```
daniel-wong-dfinity-org-twin, line 336: > An empty answer emits `null`

This is just a direct transcription of code to prose. And the code is not particularly opaque. Do not write such comments. Comments are supposed to supply information that is not readily gleaned from the code, such as intent.

> a non-empty value is wrapped in `opt principal "..."`

Ditto: this is just code to prose transcription.

> makes the canister fall back to its built-in default authorized caller

This seems tangential to the code being commented.

The real beneficiary of this information is the user of this program, not maintainer. But the user cannot read this comment. What the user can see is the prompt, which already has some explanation (albeit very brief): "(leave empty for canister default)". Therefore, this doesn't seem to add information beyond what can already be readily gleaned from the code.

> The canister re-applies this on every post_upgrade, so the caller must be supplied each release to retain it (otherwise it reverts to the default).

Ditto: this is more of a concern for the user of this program, not maintainer.

**TANGENT**: Wouldn't it make more sense for the Engine Controller canister to not require passing the same information on every upgrade? Seems pretty surprising.

daniel-wong-dfinity-org-twin (approved):

PR #10729, opened 2026-07-10
feat(cketh): deposit_erc20 endpoint (derive + store deposit addresses)

gregorydemay:
Second slice of DEFI-2921 ([F3] registration state and `deposit_erc20`), stacked on #10705 (the `TimedSizedMap` structure).

Adds the `deposit_erc20` update endpoint. Given a subaccount, it derives the caller's per-account ckERC20 deposit address (threshold-ECDSA, via the F2 derivation) and registers the account → address mapping in a bounded, time-expiring in-heap watchlist. It returns the derived deposit address together with a `valid_until` timestamp indicating until when a deposit sent to that address is guaranteed to be noticed by the minter.

Behaviour:
- re-registering a still-armed account returns the same address and its original validity window without re-arming it (idempotent);
- a full watchlist returns `TooManyActiveAddresses`;
- sponsored (caller-paid fee) registration is a later ticket (DEFI-2927).

The registration state is encapsulated in a dedicated type and survives upgrades via a snapshot event emitted at `pre_upgrade` and replayed on `post_upgrade` (the existing `log_scrapings` pattern), covered by `is_equivalent_to`.

**Store-only:** nothing is done with a registered address yet — no scanning, detection, sweeping, or fee charging. Using the stored addresses (detection/crediting, sweeping) comes in separate follow-up PRs.


rs/ethereum/cketh/minter/src/deposit_erc20/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+        DepositAddressSchema::CkErc20,
+        &account,
+    );
+    let address_string = address.to_string();
+
+    match state.deposit_addresses.insert(now, account, address) {
+        Ok(_) => Ok(address_string),
+        Err(InsertError::AlreadyPresent { .. }) => Ok(address_string),
+        Err(InsertError::AtCapacity { .. }) => Err(DepositErc20Error::TooManyActiveAddresses),
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done. On `AlreadyPresent`, the endpoint now returns the address stored in the registry (`deposit_addresses.get(now, &account)`) rather than the freshly derived one, so the "same address" guarantee holds by construction and we never return an address that is not actually registered.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -129,6 +129,30 @@ impl<K: Ord + Clone, V> TimedSizedMap<K, V> {
...
+        entries: impl IntoIterator<Item = (Timestamp, K, V)>,
+    ) -> Self {
+        let mut map = Self::new(ttl, capacity);
+        for (inserted_at, key, value) in entries {
+            map.entries
+                .insert(key.clone(), Entry { value, inserted_at });
+            map.by_time.entry(inserted_at).or_default().push_back(key);
+        }
+        map
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done — `from_entries` now asserts each key is distinct (`BUG: from_entries received a duplicate key`), so a corrupt snapshot fails loudly at restore instead of silently diverging `entries`/`by_time` and panicking later during eviction. Added a `#[should_panic]` test. In practice the snapshot comes from a `BTreeMap` iterator so keys are always unique, but this guards the invariant on the `pub` method.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -128,6 +132,23 @@ fn emit_preupgrade_events() {
...
+            .map(
+                |(registered_at, account, address)| DepositAddressRegistration {
+                    owner: account.owner,
+                    subaccount: account.subaccount,
+                    address: *address,
+                    registered_at_nanos: registered_at.as_nanos(),
+                },
+            )
+            .collect::<Vec<_>>()
+    });
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done — replaced `iter_with_time` with `iter_live(now)`, which filters out expired entries, and the pre_upgrade snapshot now uses it, so dormant (expired) registrations are no longer encoded/persisted.

rs/ethereum/cketh/minter/src/deposit_erc20/mod.rs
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed — on `AlreadyPresent` the endpoint now returns the *stored* address via `state.deposit_addresses.get(now, &account)`, not a freshly derived one, preserving the idempotency guarantee even if the key/chain-code changes.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed — `from_entries` now asserts each key is unique (`assert!(previous.is_none(), "BUG: from_entries received a duplicate key")`), so an inconsistent `by_time`/`entries` map can no longer be constructed on replay.

rs/ethereum/cketh/minter/src/main.rs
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed — the pre-upgrade snapshot is now built from `iter_live(now)`, which filters expired entries, so only live registrations are encoded into the snapshot event.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -865,13 +896,13 @@ fn get_events(arg: GetEventsArg) -> GetEventsResult {
...
+        })
     }
 
     let events = storage::with_event_iter(|it| {
         it.skip(arg.start as usize)
             .take(arg.length.min(MAX_EVENTS_PER_RESPONSE) as usize)
-            .map(map_event)
+            .filter_map(map_event)
             .collect()
     });
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in 9511ea5a7c — rather than dropping `RegisteredDepositAddresses`, it is now mapped 1:1 to a Candid `EventPayload::RegisteredDepositAddresses` variant, exactly like `SyncedToBlock`/`SyncedErc20ToBlock`. `get_events` is back to `.map(map_event)` (no `filter_map`), so paging is dense again: every event in `[start, start+length)` is returned and `total_event_count` matches the enumerable set. Added the matching `.did` variant and the reverse-mapping arms in the audit/dump test helpers.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -722,6 +745,10 @@ service : (MinterArg) -> {
     // Withdraw the specified amount of ERC-20 tokens to the given Ethereum address.
     withdraw_erc20 : (WithdrawErc20Arg) -> (variant { Ok : RetrieveErc20Request; Err : WithdrawErc20Error });
 
+    // Derive and register the ckERC20 deposit address for the given account.
+    // Returns the EIP-55 checksummed Ethereum deposit address.
+    deposit_erc20 : (Account, DepositMode) -> (variant { Ok : text; Err : DepositErc20Error });
```
gregorydemay: Let's change the parameters into a struct, named `DepositErc20Arg`:

1. it has a field mode, which is an enum for sponsored/unsponsored.
2. Unsponsored is a variant which contains an optional subaccount. Address is derived from caller's principal an subaccount (default to all zeroes).
3. Sponsored is currently a commented out variant with a todo to DEFI-2927

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -153,6 +175,14 @@ async fn minter_address() -> String {
     state::minter_address().await.to_string()
 }
 
+#[update]
+async fn deposit_erc20(account: Account, mode: DepositMode) -> Result<String, DepositErc20Error> {
+    validate_ckerc20_active();
+    let (public_key, chain_code) = state::lazy_call_ecdsa_public_key_with_chain_code().await;
+    let now = Timestamp::from_nanos(ic_cdk::api::time());
+    mutate_state(|s| register_deposit_address(s, &public_key, &chain_code, now, account, mode))
+}
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in a749f44fd9 — `deposit_erc20` no longer accepts an arbitrary `Account`. It now calls `validate_caller_not_anonymous()` and derives the account from the caller: `Account { owner: caller, subaccount }`. A caller can therefore only register addresses for their own principal, so the bounded registry can no longer be filled with other principals' accounts.

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Done in a749f44fd9. `deposit_erc20` now takes `DepositErc20Arg { mode }`; `DepositMode::Unsponsored { subaccount }` is the only active variant (address derived from the caller's principal + subaccount, default all-zeroes), and `Sponsored` is commented out with a `// TODO(DEFI-2927)`.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -153,6 +175,22 @@ async fn minter_address() -> String {
     state::minter_address().await.to_string()
 }
 
+#[update]
+async fn deposit_erc20(arg: DepositErc20Arg) -> Result<String, DepositErc20Error> {
+    validate_ckerc20_active();
+    let caller = validate_caller_not_anonymous();
+    let subaccount = match arg.mode {
+        DepositMode::Unsponsored { subaccount } => subaccount,
+    };
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Addressed: the public API intentionally exposes only `DepositMode::Unsponsored`; `Sponsored` (caller-paid fee) is deferred to DEFI-2927 and the PR description now reflects the actually exposed API. `DepositErc20Error::TemporarilyUnavailable` is not dead — `State::register_deposit_address` returns it when the minter threshold-ECDSA public key has not been fetched yet.

rs/ethereum/cketh/minter/src/deposit_erc20/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    match state.deposit_addresses.insert(now, account, address) {
+        Ok(_) => Ok(derived_address),
+        Err(InsertError::AlreadyPresent { .. }) => Ok(state
+            .deposit_addresses
+            .get(now, &account)
+            .expect("BUG: AlreadyPresent implies a live stored entry")
+            .to_string()),
+        Err(InsertError::AtCapacity { .. }) => Err(DepositErc20Error::TooManyActiveAddresses),
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The free `register_deposit_address` has since been inlined into `State::register_deposit_address`. The idempotent-path micro-optimization (skipping address derivation when the account is already armed) is not being pursued: derivation is a cheap local computation and the endpoint already `await`s the tECDSA public key regardless. Resolving as outdated.

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -130,6 +130,40 @@ impl<K: Ord + Clone, V> TimedSizedMap<K, V> {
...
+    /// inserted.
+    pub fn iter_live(&self, now: Timestamp) -> impl Iterator<Item = (Timestamp, &K, &V)> {
+        self.entries.iter().filter_map(move |(key, entry)| {
+            if self.is_expired(entry.inserted_at, now) {
+                None
+            } else {
+                Some((entry.inserted_at, key, &entry.value))
+            }
+        })
+    }
```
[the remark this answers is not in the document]
gregorydemay: iter_live is no longer there.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    pub fn watchlist_iter(&self) -> impl Iterator<Item = (&Account, &Entry<DepositRequest>)> {
+        self.watchlist.iter()
+    }
+
+    pub fn watchlist_snapshot(&self) -> Vec<DepositAddressRegistration> {
+        let mut snapshot: Vec<_> = self
+            .watchlist
+            .iter()
```
mbjorkqvist: **Event-sourcing equivalence can break: `check_audit_log` traps once a registered address expires.**

`watchlist_snapshot()` iterates *all* entries, including expired-but-not-yet-evicted ones (eviction only runs inside `insert`/`insert_entry` — there is no timer sweep), so the pre-upgrade snapshot persists expired entries. On replay, `rebuild_watchlist` → `insert_entry` *drops* already-expired entries. The two directions are asymmetric, so the replayed state no longer matches the live state and `is_equivalent_to` (via `check_audit_log`) traps.

Verified locally with a PocketIC test: register one address, `advance_time(DEPOSIT_ADDRESS_SCAN_WINDOW + 1s)`, then `check_audit_logs_and_upgrade_as_ref(..)` traps with `replaying the audit log should produce an equivalent state` (live map still holds the expired entry; replayed map is empty). It scales too: when the map is full of live entries, new inserts are *rejected*, so nothing triggers eviction and expired entries linger indefinitely.

Note that filtering the snapshot to live entries alone would not fix it (the live map still physically holds the expired entry) — the fix likely needs eager eviction, or `is_equivalent_to`/`PartialEq` comparing only live entries.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Rebuild the watchlist from a snapshot previously produced by
+    /// [`Self::watchlist_snapshot`], replacing any existing content.
+    ///
+    /// Entries are re-admitted under *this* version's limits rather than the
+    /// ones in effect when the snapshot was taken: entries already expired as of
+    /// `now` are dropped, each entry's validity is clamped to the current scan
+    /// window, and the current capacity is enforced. Since the snapshot lists
+    /// entries longest-lived first, the soonest-to-expire are the ones dropped
+    /// when the snapshot exceeds the current capacity.
+    pub fn rebuild_watchlist(&mut self, now: Timestamp, snapshot: &[DepositAddressRegistration]) {
```
mbjorkqvist: **Cross-version constant changes silently drop still-armed addresses / shorten promised validity (no trap, but data loss).**

Because rebuild re-admits under *this* version's limits:
- Lowering `MAX_ACTIVE_DEPOSIT_ADDRESSES` in an upgrade: if the snapshot has more live entries than the new capacity, the soonest-to-expire are silently dropped — potentially addresses that have already received funds. This reintroduces at upgrade time exactly the loss the runtime `AtCapacity`-reject logic was designed to prevent.
- Lowering `DEPOSIT_ADDRESS_SCAN_WINDOW`: existing entries are clamped to the shorter `now + ttl`, so a `valid_until` already returned to a user is no longer honored after the upgrade.

Neither is caught by tests (the harness always upgrades to the same wasm). Worth a deliberate decision / comment on whether shrinking these constants is ever allowed.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+            .watchlist
+            .iter()
+            .map(|(account, deposit)| DepositAddressRegistration {
+                owner: account.owner,
+                subaccount: account.subaccount,
+                address: deposit.value.address,
+                expires_at_nanos: deposit.expires_at,
+            })
+            .collect();
+        snapshot.sort_unstable_by_key(|deposit| Reverse(deposit.expires_at_nanos));
```
mbjorkqvist: **Minor: identical-expiry entries can reorder across replay, undermining equivalence.**

`sort_unstable_by_key(Reverse(expires_at))` does not preserve input order for equal `expires_at`. Multiple `deposit_erc20` calls executing in the same round share `ic_cdk::api::time()`, so they get the same `expires_at` and land in the same `by_time` bucket. After replay the bucket's `VecDeque` order can differ from the live chronological insertion order, and `TimedSizedMap`'s derived `PartialEq` compares `by_time` order-sensitively → another latent `is_equivalent_to` divergence, independent of the expiry issue above.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::cmp::Reverse;
+use std::num::NonZeroUsize;
+use std::time::Duration;
+
+/// Time window during which a registered ckERC20 deposit address is kept armed.
+pub const DEPOSIT_ADDRESS_SCAN_WINDOW: Duration = Duration::from_secs(24 * 60 * 60);
+
+// Ethereum blocktime is 12s (on average), so that there are 7_200 blocks per day.
+// Use 1 transaction per block to a minter-controlled address as a crude upper-bound.
+const MAX_ACTIVE_DEPOSIT_ADDRESSES: NonZeroUsize = NonZeroUsize::new(7_000).unwrap();
```
mbjorkqvist, line 17: **Resource exhaustion / DoS: a single principal can fill the global watchlist for 24h.**

`deposit_erc20` is currently free (fee handling deferred to DEFI-2927), gated only on non-anonymous + ckERC20 active, and derives the account from caller + subaccount with no per-principal cap. A single caller can register up to 7,000 distinct subaccounts, each armed 24h; since `insert` rejects at capacity rather than evicting live entries, every other user then gets `TooManyActiveAddresses` for up to 24h. Worth confirming this is acceptable while the endpoint is store-only, or gating it until the fee mechanism lands.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// State machine to handle deposits made to addresses controlled by the minter individually-derived for each user.
+/// This is in particular allows deposits from central exchanges (CEX).
```
mbjorkqvist: Nit: "This is in particular allows deposits from central exchanges" reads awkwardly (grammar), and the "deposit lifecycle" list only has step 1 — either complete the enumeration or trim it.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -153,6 +160,27 @@ async fn minter_address() -> String {
...
 }
 
+#[update]
+async fn deposit_erc20(arg: DepositErc20Arg) -> Result<DepositErc20Response, DepositErc20Error> {
+    validate_ckerc20_active();
+    let caller = validate_caller_not_anonymous();
+    let subaccount = match arg.mode {
+        DepositMode::Unsponsored { subaccount } => subaccount,
+    };
+    let account = Account {
```
mbjorkqvist, line 170: **`None` and `Some([0u8;32])` subaccounts derive the same address but are distinct watchlist keys.**

`deposit_derivation_path` uses `account.effective_subaccount()`, which maps `None` → `[0;32]`. So a caller passing `subaccount: None` and later `Some([0;32])` derives the *same* Ethereum address but occupies two distinct `Account` keys/slots, and the idempotency guarantee (keyed on `Account`) does not hold across the two. Consider normalizing the subaccount (map `None`→`[0;32]`, or reject an explicit all-zero subaccount) before building the key.

rs/ethereum/cketh/minter/src/state/audit.rs
```diff
@@ -165,6 +166,9 @@ pub fn apply_state_transition(state: &mut State, payload: &EventType) {
                 *block_number,
             );
         }
+        EventType::RegisteredDepositAddresses(registrations) => state
+            .automatic_deposits
+            .rebuild_watchlist(Timestamp::from_nanos(ic_cdk::api::time()), registrations),
```
mbjorkqvist: **`apply_state_transition` now reads the wall clock, making replay non-deterministic w.r.t. the log.**

Every other arm is a pure function of the event; this one pulls `ic_cdk::api::time()`. The `Event` carries a `timestamp` (discarded via `..` in `replay_events_internal`) that would be the faithful, deterministic "now" for the snapshot — consider threading the event timestamp through instead. (Also: the commit message states the snapshot is rebuilt "with the snapshot event's timestamp," which doesn't match the code.)

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -587,6 +597,50 @@ impl State {
...
+    /// [`DepositErc20Error::TemporarilyUnavailable`] if the minter's public key
+    /// has not been fetched yet.
+    pub fn register_deposit_address(
+        &mut self,
+        now: Timestamp,
+        account: Account,
+    ) -> Result<Entry<DepositRequest>, DepositErc20Error> {
+        let (master_public_key, chain_code) =
+            self.public_key_and_chain_code()
+                .ok_or(DepositErc20Error::TemporarilyUnavailable(
```
mbjorkqvist, line 632: Minor: this `TemporarilyUnavailable` arm is unreachable from `deposit_erc20`, which always awaits `lazy_call_ecdsa_public_key_with_chain_code()` (caching the key) before calling this. Fine as defensive code — just flagging that it's effectively dead on the endpoint path.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -722,6 +759,11 @@ service : (MinterArg) -> {
     // Withdraw the specified amount of ERC-20 tokens to the given Ethereum address.
     withdraw_erc20 : (WithdrawErc20Arg) -> (variant { Ok : RetrieveErc20Request; Err : WithdrawErc20Error });
 
+    // Derive and register the ckERC20 deposit address for the caller.
+    // The account is derived from the caller's principal and the subaccount in the argument.
+    // Returns the EIP-55 checksummed Ethereum deposit address.
```
mbjorkqvist: Nit: the comment only mentions returning the address; the response is now `DepositErc20Response { address; valid_until }`. Worth documenting `valid_until` here too.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: Note that check_audit_log is only used for testing. The reasoning for the asymmetry is as follows:

1. pre_upgrade: be as simple as possible, since any trap there is critical/potentially catastrophic. We could consider doing eviction there though.
2. post_upgrade: only consider live entries, ok to have more logic there.

I'm not sure whether a different `ic_cdk::api::time()` could be returned between pre_upgrade and post_upgrade. WDYT?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: yes this is done on purpose. I do not expect those constants to change too often. To test this properly I would need to make those constants a config option and I currently don't have the impression that this is worth it. WDYT?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: yes, there is nothing we can really do here. Having a bound per principal would be also be trivial to bypass (same physical user just creates multiple principals). For the sponsored case, we will be able to burn the ckETH upfront so that abuse will not be economically reasonable.

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: This is not true because Account has a custom PartialEq implementation that considers effective_subaccount, but it's definitely worth a unit test.

rs/ethereum/cketh/minter/src/state/audit.rs
gregorydemay: 🤖 Fixed. `apply_state_transition` no longer reads `ic_cdk::api::time()` for the `RegisteredDepositAddresses` arm — the watchlist is now rebuilt *verbatim* from the snapshot event itself, so replay is a pure function of the log again. The snapshot records each entry with its stored `expires_at` (plus the ttl/capacity in force), and `rebuild_watchlist` restores them as-is, so there is no wall-clock read and the code matches its commit message.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed by making snapshot and restore symmetric instead of evicting. `watchlist_snapshot()` now dumps *all* entries — live and expired-but-not-yet-evicted — in time-index order, and `rebuild_watchlist` restores them verbatim via `TimedSizedMap::from_ordered_entries`: no expiry-dropping, no clamping, no capacity re-check. The replayed map therefore equals the live map by construction, including any lingering expired entry, so `is_equivalent_to` no longer diverges.

Re your question: yes, `ic_cdk::api::time()` can differ between `pre_upgrade` and `post_upgrade` since they run in separate messages — which is exactly why replay now uses each entry's stored `expires_at` from the snapshot rather than a fresh clock.

Trade-off: expired entries are now encoded into the snapshot event. This is bounded (at most `capacity`), they are re-snapshotted on each upgrade so they don't accumulate in the map, and they are evicted lazily on the next insert. Filtering them out would require eagerly evicting from the live map in `pre_upgrade`, which we're deliberately keeping as simple as possible. (Noted on the earlier snapshot-size thread too.)

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed. The snapshot is now emitted in `by_time` order — ascending expiry, and insertion order within a shared-expiry bucket — via a new `iter_by_expiry`, and `from_ordered_entries` rebuilds the `by_time` index bucket-for-bucket in that same order. So same-round registrations that share an `expires_at` keep their chronological order across replay, and the derived `PartialEq` matches. Covered by a `TimedSizedMap` round-trip unit test that deliberately places two same-expiry entries in non-key order.

rs/ethereum/cketh/minter/src/main.rs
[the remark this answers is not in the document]
gregorydemay: 🤖 Heads-up: a later change on this branch partially reverses this. To keep event-log replay equivalent to the live state, the pre-upgrade snapshot and the replayed map must match *exactly*, so the snapshot now intentionally includes expired-but-not-yet-evicted entries rather than filtering them. The cost is bounded — at most `capacity` entries, re-snapshotted on each upgrade (so they don't accumulate in the map) and evicted lazily on the next insert. Filtering them out would instead require eagerly evicting from the live map in `pre_upgrade`, which we're keeping as simple as possible.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 This is now handled differently after the verbatim-restore change. The snapshot records the ttl (`scan_window_nanos`) and `capacity` in force when it was taken, and `rebuild_watchlist` restores entries as-is under those limits, so:

- Existing entries are never dropped or re-clamped on upgrade — an already-returned `valid_until` stays honored, and lowering `MAX_ACTIVE_DEPOSIT_ADDRESSES` no longer silently discards funded addresses (the restored map may sit temporarily above capacity and simply rejects *new* inserts until it drains below it).
- Because the limits live in the snapshot, changing the code constants doesn't affect existing state at all; propagating a change would require an explicit migration (e.g. an upgrade arg), which is the deliberate future-work path.

So shrinking the constants is no longer a silent-data-loss footgun — it's a no-op on existing state until that migration is added.

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Agreed it's worth locking in — added a unit test. `should_watch_address_for_account` now includes a case *"treats missing and all-zero subaccount as the same account"*: it arms `subaccount: None`, re-registers the same owner with `Some([0; 32])`, and asserts the second call returns the already-stored entry (same address, same validity) with the watchlist still holding a single entry — i.e. they collide on one `Account` key via `effective_subaccount`, so the idempotency guarantee holds across the two forms.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Kept as defensive code — agreed it is effectively unreachable from `deposit_erc20`, which awaits `lazy_call_ecdsa_public_key_with_chain_code()` before calling this. `register_deposit_address` is a `pub` method that doesn't itself guarantee the key has been fetched, so the arm guards that precondition rather than relying on every future caller ordering the await correctly.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed — reworded the `AutomaticDeposits` doc comment (grammar) and replaced the single-item "lifecycle" list with a short description of what exists today plus a pointer to the scanning/minting future work (DEFI-2927).

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Fixed — the `deposit_erc20` comment in the .did now documents `valid_until` (timestamp, ns since the Unix epoch, until which a deposit is guaranteed to be noticed) alongside the returned address.

mbjorkqvist (approved):
Thanks for the fixes and clarifications, LGTM!

PR #10733, opened 2026-07-10
fix: return all transactions when start is None in index get_account_…

Rachit2323:
## Bug

Calling `get_account_transactions` without a `start` value always returns empty results, even when the account has transactions.

## Why

The code was using `u64::MAX` as a default for `start`. Then it filtered for blocks less than `u64::MAX` — but since the search started at `u64::MAX`, nothing was ever returned.

## Fix

When `start` is not given, return all transactions instead of filtering.

## Changes

- Fixed `rs/ledger_suite/icrc1/index-ng/src/main.rs`
- Fixed `rs/ledger_suite/icp/index/src/main.rs`
- Added tests to both test files


rs/ledger_suite/icrc1/index-ng/tests/tests.rs
```diff
@@ -904,6 +904,35 @@ fn test_get_account_transactions_start_length() {
     }
 }
 
+#[test]
+fn test_get_account_transactions_without_start_matches_unbounded_start() {
+    let initial_balances = vec![(account(1, 0), 1_000_000_000_000u64)];
```
basvandijk: To fix [the clippy error](https://github.com/dfinity/ic/actions/runs/29121596911/job/86457710880#step:5:2909) use:
```suggestion
    let initial_balances = vec![(account(1, 0), 1_000_000_000_000_u64)];
```

mbjorkqvist (commented):
This is inaccurate - all transactions are already returned when `start` is `None`, as verified by [`test_get_account_transactions`](https://github.com/dfinity/ic/blob/5210e6b112cd674a64ae4d89712ad1b9001b2b2c/rs/ledger_suite/icrc1/index-ng/tests/tests.rs#L727-L730) for the ICRC1 index, and [`test_get_account_identifier_transactions`](https://github.com/dfinity/ic/blob/5210e6b112cd674a64ae4d89712ad1b9001b2b2c/rs/ledger_suite/icp/index/tests/tests.rs#L1121-L1125) for the ICP index. The added tests also already pass on `master`, and don't add anything new.

PR #10734, opened 2026-07-11
refactor(ic-admin): Re-use fns to build clients and keys in ic-admin

Bownairo:
Where possible, use helpers to avoid building `RegistryClientImpl`s and registry keys, directly.

### DIFF
--- a/rs/registry/admin/bin/main.rs
+++ b/rs/registry/admin/bin/main.rs
@@ -88,16 +88,17 @@ use ic_protobuf::registry::{
 };
 use ic_registry_client::client::RegistryClientImpl;
 use ic_registry_client_helpers::{
-    chain_keys::ChainKeysRegistry, crypto::CryptoRegistry, deserialize_registry_value,
-    ecdsa_keys::EcdsaKeysRegistry, hostos_version::HostosRegistry,
+    api_boundary_node::ApiBoundaryNodeRegistry, chain_keys::ChainKeysRegistry,
+    crypto::CryptoRegistry, deserialize_registry_value, ecdsa_keys::EcdsaKeysRegistry,
+    hostos_version::HostosRegistry, node_operator::NodeOperatorRegistry,
     replica_version::ReplicaVersionRegistry, subnet::SubnetRegistry,
 };
 use ic_registry_keys::{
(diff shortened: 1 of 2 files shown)


daniel-wong-dfinity-org-twin:
I changed the title. PTAL.

rs/registry/admin/bin/main.rs
```diff
@@ -4661,19 +4662,19 @@ async fn main() {
...
-                &opts.hsm_slot.expect(
+                opts.hsm_slot.as_ref().expect(
                     "HSM slot must also be provided for --use-hsm; use --hsm-slot or see --help.",
                 ),
-                &opts.hsm_key_id.expect(
+                opts.hsm_key_id.as_ref().expect(
                     "HSM key ID must also be provided for --use-hsm; use --key-id or see --help.",
                 ),
-                &opts.hsm_pin.expect(
+                opts.hsm_pin.as_ref().expect(
```
daniel-wong-dfinity-org-twin: This changes seems to be orthogonal? Do you simply need to merge from master (or rebase)?

rs/registry/admin/bin/main.rs
```diff
@@ -6750,32 +6741,22 @@ async fn get_subnet_pk(registry: &RegistryCanister, subnet_id: SubnetId) -> Publ
...
 }
 
-fn get_api_boundary_node_ids(nns_url: Vec<Url>) -> Vec<String> {
-    let registry_client = RegistryClientImpl::new(
-        Arc::new(NnsDataProvider::new(
-            tokio::runtime::Handle::current(),
-            nns_url,
-        )),
-        None,
+fn get_api_boundary_node_ids(opts: Opts, nns_url: Vec<Url>) -> Vec<String> {
```
daniel-wong-dfinity-org-twin: Could you pass Registry client itself? (As opposed to passing parameters needed to construct it.)

One thing I don't like about passing Opts is that you are passing "the whole hardware store" instead of just the one or two "nails" (verify_nns_resposnes, nns_public_key_pem_file) that you actually actually need/use. Passing Registry client avoids that.

daniel-wong-dfinity-org-twin (approved):

rs/registry/admin/bin/main.rs
Bownairo: They're required to pass opts into the helper fns, below. We borrow at the field instead of the parent struct.

PR #10711, opened 2026-07-09
feat: bump prost-related crates (tonic, etc) from 0.13 to 0.14

nmattia:
This bumps one of our tonic duplicates and removes it from the dependency graph. Same goes for related crates like prost, tonic_build, etc.

Prost now derives Eq and Hash automatically where possible (explained in [their release notes][prost]). Therefore, there is no need to configure Prost to derive these traits anymore (and so, we stop doing that).

[prost]: https://github.com/tokio-rs/prost/releases/tag/v0.14.0

PR #10735, opened 2026-07-11
chore(registry): Update registry helpers to return keyed types

Bownairo:
Followup from https://github.com/dfinity/ic/pull/9772#discussion_r3062917623.

The values returned are derived from `RegistrySnapshot`, which is a `BTreeMap`, and so are already unique. The new definitions express this on the type level.


michael-weigelt (approved):

rs/protobuf/generator/src/lib.rs
```diff
@@ -243,7 +243,7 @@ fn build_registry_proto(def: &Path, out: &Path) {
 
     config.type_attribute(
         ".registry.hostos_version",
-        "#[derive(serde::Serialize, serde::Deserialize)]",
+        "#[derive(serde::Serialize, serde::Deserialize, Ord, PartialOrd, Eq)]",
```
daniel-wong-dfinity-org-twin: I think Prost has started generating derive Eq recently. See https://github.com/dfinity/ic/pull/10711/ Ask Nicolas M.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -62,15 +62,15 @@ pub(crate) fn get_all_node_records(snapshot: &RegistrySnapshot) -> BTreeMap<Node
 /// Returns all replica version records in the snapshot.
 pub(crate) fn get_all_replica_version_records(
     snapshot: &RegistrySnapshot,
-) -> Vec<(String, ReplicaVersionRecord)> {
-    let mut replica_versions = Vec::new();
+) -> BTreeMap<String, ReplicaVersionRecord> {
```
daniel-wong-dfinity-org-twin, line 65: Technically, this isn't being changed to a set, but rather a map. I mean, I'm not sure what (if anything) to do about this, but think about how the PR title and/or description could be improved to maximize truthiness.

daniel-wong-dfinity-org-twin (approved):
I updated the title of this PR. PTAL.

The important things that I added:

1. Motivation (avoid linear scan). If you can add more (e.g. some important method is "too" slow), that would be great. Doesn't have to be in the title.

2. Scope: otherwise, "a few" could be anywhere in this large repo.

PR #10736, opened 2026-07-11
chore: Normalize use of version in `TopologySnapshot`

Bownairo:
Use the internally held registry version when checking `TopologySnapshot`, leaving updates only to calls that explicitly ask for a new version. This makes subsequent calls idempotent, so they won't change as the underlying registry does, unless explicitly updated.


basvandijk (approved):

daniel-wong-dfinity-org-twin (approved):
Just reviewing Governance code changes.

Not sure what this does/accomplishes, but I also see no harm -> Approve.

Is this just a little optimization? Can the PR description (or title) explain (more)?

PR #10738, opened 2026-07-11
fix: correct block IDs returned by get_blocks when start is non-zero

Rachit2323:
## Bug

`decode_block_range` was passing `start + i` to the decoder instead of `i`. Logically wrong, but no observable bug since `decode_icrc1_block` ignores that parameter anyway.

## Fix

Removed the unused `_txid: u64` parameter from `decode_icrc1_block` and simplified `decoder` from `Fn(u64, Vec<u8>)` to `Fn(Vec<u8>)`.

## Changes

- `rs/ledger_suite/icrc1/index-ng/src/main.rs` — removed unused txid parameter


basvandijk:
@Rachit2323 thanks for the contribution. Note there's a [clippy error](https://github.com/dfinity/ic/actions/runs/29182430898/job/86622500756#step:5:2909). Also note both of your PRs are drafts. If you want us to review them I suggest to undraft them.

rs/ledger_suite/icrc1/index-ng/src/main.rs
```diff
@@ -1249,7 +1249,7 @@ fn decode_block_range<R>(start: u64, length: u64, decoder: impl Fn(u64, Vec<u8>)
     with_blocks(|blocks| {
         let limit = blocks.len().min(start.saturating_add(length));
         (start..limit)
-            .map(|i| decoder(start + i, blocks.get(i).unwrap()))
+            .map(|i| decoder(i, blocks.get(i).unwrap()))
```
mbjorkqvist: This change is accurate, but since the `decoder` function (in this case [`decode_icrc1_block`](https://github.com/dfinity/ic/blob/5210e6b112cd674a64ae4d89712ad1b9001b2b2c/rs/ledger_suite/icrc1/index-ng/src/main.rs#L1228-L1231), since currently the only caller to `decode_block_range` is the `get_blocks` function above) ignores the first parameter, this changes no observable behavior. I'd still be happy to merge a fix, but I would recommend the following:

- Remove the added test (it already passes on `master`, and it doesn't provide any regression protection).
- Remove the unused `_txid: u64` parameter in the `decode_icrc1_block` function that `decoder` points to, and adjust the `decoder` parameter in this `decode_block_range` function to also remove `u64` there (leaving `decoder: impl Fn(Vec<u8>`).
- Adjust the PR description to reflect the fact that there is actually no observable bug.

mbjorkqvist (commented):
Thanks for the PR, @Rachit2323!

rs/ledger_suite/icrc1/index-ng/src/main.rs
Rachit2323: Done — removed the test, removed the unused _txid: u64 parameter, and updated the PR description.

mbjorkqvist (approved):
Thanks @Rachit2323!

PR #10739, opened 2026-07-11
feat(governance): Standards and Conventions

daniel-wong-dfinity-org-twin:
The "root" file here is STANDARDS_AND_CONVENTIONS.md. It is human readable.

From this, a couple sets of Claude instructions are generated:

1. How to WRITE code.
2. How to REVIEW code. (not yet implemented; future work)

The idea is to keep the 1 small, to not overwhelm Claude, while the second set of instructions is comprehensive. To avoid overloading the reviewer Claudes, the instructions are "sharded out". That is, the rules are broken out into several subsets, and each subset is handled by one agent.

The derivation of these instructions is purely mechanical. Furthermore, in a later PR, using GitHub Actions, it will be enforced that the derived files stay in sync.

# Preview

Currently, this is only applied to Governance code. This is done via a CLAUDE.md file in the main Governance directories that points to the derived author instructions.

# Derivation of Claude Instructions

A subset of S&C.md becomes the instructions for how to write code. Basically, content in parentheses is excluded. There are two styles of parentheses:

1. single, i.e. "(" + ")" - This can be used to exclude an entire rule, or pieces of a rule. Empty lines are not allowed between single parentheses.

2. double, i.e. "((" + "))" - This is used to exclude whole swaths of S&C.md.


bjoernek:
Really like the overall direction here.

One suggestion on structure. Right now the ruleset mixes three kinds of rules that an agent handles very differently, and I think it would be worth separating them explicitly (a dimension orthogonal to the existing write/review split):

**Tier 1: mechanically checkable.** Naming prefixes (`is_`, `try_`, `_count`), `list_` for collection RPCs, `.collect::<Vec<T>>()` over bare `.collect()`, no bare `.into()`, no `pub` fields, separate test files, banned words. Agents apply these reliably and consistently.

**Tier 2: judgment-based but tractable.** Return early, short branch arms, spin out, meaningful doc comments, breadcrumbs in errors, "abstractions must pull their own weight". This is where the LLM's judgment genuinely adds value over a linter, and where an AI reviewer earns its keep.

**Tier 3: subjective aphorisms.** "Lead with the lede", "guide the eye", "be MEANINGFUL", the blank-line-when-a-statement-spans-3-lines rule. Great as author-facing inspiration, but agents apply these inconsistently and tend to over-apply them, producing nitpicky feedback that varies run to run. In always-on context they also dilute the signal of Tiers 1 and 2.

Why this matters for the agent specifically: unlike a human who skims and internalizes the spirit, an agent treats all the lines as roughly equal-weight instructions, so signal-to-noise drives behavior more than for a human reader. Concretely, for the future reviewer phase, I'd suggest agents only *flag* Tier 1 and Tier 2, and keep Tier 3 as guidance that informs writing but is never surfaced as a review comment. Otherwise reviews risk drowning in taste-nits.

The sharding architecture already gets us most of the way there; this is mostly about tagging each rule with its tier so the derivation can route them.

bjoernek:
On the "Derive as much as possible" rule under **Defining Types**: I'd suggest pulling `Copy` and `Default` out of that list. `Debug`, `Eq`, `Ord`, and `Hash` are safe to derive freely, but `Copy` and `Default` change how the type behaves and can create hidden problems, so I think they should be added deliberately rather than by default.

**`Copy`** means every use of the value silently makes a full copy. That's fine for a tiny type today, but it plants a trap for later: if someone adds a bigger field, every use starts silently copying that field (a quiet performance cost), and the day the new field can't be copied, the derive stops compiling and every call site that relied on copying breaks at once. Better to add `Copy` only when small-and-copyable is genuinely intended.

**`Default`** invents a default value even for types that don't have a sensible one. What is a default `PrincipalId` or `Account`? Deriving `Default` produces an all-zeros value anyway, and that "valid-looking but meaningless" value tends to leak past the validation that was supposed to reject it.

Suggestion: keep "derive as much as possible" for `Debug`, `Eq`, `Ord`, `Hash`, and move `Copy` / `Default` to a separate note like "derive only when the semantics are genuinely intended."

bjoernek:
The **Consistency is king** rule and the **Grandfathering** section point in opposite directions for the most common case an agent will hit: writing new code right next to non-compliant legacy code.

"When nearby code and far away code follow different patterns, follow the nearby code" says match the legacy. Grandfathering says the legacy isn't compliant yet. So when I add a function beside older code that, say, uses banned words or bare `.into()`, should I match its style or follow the standards? As written, both readings are valid, and an agent will resolve it differently from run to run.

Suggestion: add an explicit tie-breaker, something like:

> New code follows these standards even when adjacent legacy code does not. The "follow nearby code" rule applies only among code that is already compliant (e.g. matching naming or structure between two conforming modules).

That keeps the good intent of the consistency rule (don't invent a third style) without letting non-compliant legacy pull new code out of compliance.

rs/nervous_system/standards_and_conventions/derive_claude_instructions
```diff
@@ -0,0 +1,146 @@
...
+    OUTPUT.write_text(FRONT_MATTER + "\n\n" + text)
+
+    print(f"Wrote {OUTPUT.relative_to(LOCAL_DIRECTORY.parent)}")
+    print(f"Original non-whitespace chars: {original_chars}")
+    print(f"Kept (author AI):              {kept_chars}")
+    print(f"Cut:                           {cut_chars} ({100 * cut_chars / original_chars:.1f}%)")
+    print(f"Whole sections dropped:        {len(dropped_sections)}")
+    print(f"Paren spans dropped:           {len(dropped_spans)}")
+    if violations:
+        print(f"\n{len(violations)} markup violation(s) found (see warnings above)", file=sys.stderr)
```
bjoernek, line 144: `find_violations` warns but the run still exits `0`. Once the CI sync-check lands, malformed markup that still produces stable output would pass, so these rules end up documented but unenforced. Suggestion: keep generation lenient, add a `--check` mode for CI that writes nothing and exits non-zero on any violation (and on drift).

rs/nervous_system/STANDARDS_AND_CONVENTIONS.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Banned Words
+
+Do not use the following words (because they convey little to no information,
+just like the word "marklar" in _South Park_):
+
+* Nouns:
+    * data
+    * info(rmation)
+    * state
```
bjoernek, line 126: Question on the **Banned Words** list: how do you see `state` playing out in this codebase? I'm on board with `data`, `info`, `do`, `run`, `execute` as filler, but `state` is a core domain term across the IC (canister state, replicated state, state machine), so a hard ban seems likely to push authors toward vaguer names than `state` itself. Would it make sense to drop `state`, or reframe it as "avoid as a bare, unqualified name" so things like `replicated_state` are still fine?

rs/nervous_system/standards_and_conventions/derive_claude_instructions
```diff
@@ -0,0 +1,146 @@
...
+
+    text, dropped_spans = strip_parens_to_fixed_point(text)
+    text = cleanup_whitespace(text).strip() + "\n"
+    text = restore_code(text, saved_code)
+
+    def non_white_space_len(s):
+        return sum(1 for c in s if not c.isspace())
+
+    original_chars = non_white_space_len(original)
+    kept_chars = non_white_space_len(text)
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: Done.

According to Merriam-Webster, "white space" is a word, not "whitespace". OTOH, Wikipedia has an entry for "whitespace character"...

I do not buy your point about clarity, but I agree that in this business, people usually spell it without space.

Another argument in your favor: there is another function: cleanup_whitespace. I hadn't noticed this inconsistency.

rs/nervous_system/STANDARDS_AND_CONVENTIONS.md
daniel-wong-dfinity-org-twin: I have added an exception: existing compound terms (like the ones you mentioned) are grandfathered in.

I consider this exception to be "just clarification" though. The theory here is that if you name your function `fetch_canister_state`, you are relying on the pre-existing widely-used compound term "canister state", not the independent word "state".

rs/nervous_system/standards_and_conventions/derive_claude_instructions
daniel-wong-dfinity-org-twin: I like this idea. Let's do it in a future PR to avoid scope creep in this PR. I've added a TODO. I think it would make the most sense to include this in the PR where we enforce syncing via GitHub Actions.

daniel-wong-dfinity-org-twin (commented):
re consistency vs. grandfathering: To me, they seem complementary:

"Consistency is king" is saying that you might have to break some (other) rules, because of legacy code.

vs.

Grandfathering is saying that we'll fix compliance issues in legacy code in a separate dedicated campaign, so do not try to opportunistically fix legacy code when you add or modify code.

So, if you are looking at "Consistency is king" and worrying that it will just increase the amount of non-compliant code, then Grandfathering should be of some relief, because it is saying, "yes, but we'll sweep through and fix that later".

daniel-wong-dfinity-org-twin:
re derive: Yes, of course, the compiler might allow a trait impl to be derived even when it makes no sense (either currently or in some likely future evolution). The point here is not "do the max that the COMPILER would allow"; but rather, "do the max that GOOD DESIGN would allow". I have changed the wording to avoid this confusion. PTAL.

daniel-wong-dfinity-org-twin:
re tiers: Yeah, some of the rules are pretty abstract, and Claude might not do a good job implementing them, or they might be distracting from things that it can apply more easily. I would rather observe how Claude handles what we have here so far (currently doing that), and course correct as we gain experience.

bjoernek (approved):

PR #10741, opened 2026-07-12
fix: `CanisterHttpPoolImpl::get()` serves full response for flexible & non-replicated requests

eichhorl:
## Background
For fully replicated HTTP outcalls requests, the gossiped shares include only the response hash, because each replica determines the HTTP response content by making the outcall themselves.

For flexible & non-replicated requests, not all replicas make the HTTP outcall. Yet, all replicas should still know the results of all outcalls, such that all replicas can potentially include them in a block. Therefore, the gossiped responses also include the full HTTP response content. Otherwise, a replica making a non-replicated request (where only one replica makes the outcall), would need to wait 1-13 blocks (13 node subnet) until it is the blockmaker to include the response in a block.

Additionally, HTTP outcalls shares are not relayed, or advertised. They are pushed to all peers immediately (`is_latency_sensitive: true`).

## Problem
Because HTTP outcalls shares are pushed without adverts, the code before this PR assumed that `CanisterHttpPoolImpl::get()` would never actually be called (`get()` is usually how a node returns an artifact that is requested / "pulled" by other nodes, i.e. when resolving an advert). Because of this assumption, `get()` never returned the full response content, and instead set it to `None` for simplicity: https://github.com/dfinity/ic/blob/5ee93a413156c9fdc2787194731d6bec9488de15/rs/artifact_pool/src/canister_http_pool.rs#L194-L207

However, this assumption is incorrect due to the following reason: When a node receives an HTTP outcalls share, it calls its "bouncer" to see if the artifact is `Wanted`, `Unwanted` or `MaybeWantsLater`: https://github.com/dfinity/ic/blob/478eb8343cb4be99db7e715b1fc96c8bc0c14f41/rs/https_outcalls/consensus/src/gossip.rs#L58-L67

If the bouncer returns `MaybeWantsLater`, then P2P will _only_ stash the advert, and take out the (larger) artifact content in order to conserve memory, even if the advert was pushed together with its content:
https://github.com/dfinity/ic/blob/478eb8343cb4be99db7e715b1fc96c8bc0c14f41/rs/p2p/artifact_downloader/src/fetch_artifact/download.rs#L180-L184

If at a later point, the bouncer value switches to `Wants`, then the advert needs to be resolved from scratch, meaning the artifact content is downloaded from a peer which calls `CanisterHttpPoolImpl::get()`.

In case of non-replicated or flexible outcalls, due to the assumption above, the contacted peer will then serve the artifact without the HTTP response content. The HTTP share without response content is unexpected for non-replicated and flexible requests, meaning the receiving peer will invalidate it.

## Proposed Changes
This PR fixes the problem by recording in the pool (at insertion time), if the response content should be served via `get()` or not. This will allow nodes to correctly fetch shares for flexible or non-replicated requests (including the response content) from other nodes.

### DIFF
--- a/rs/artifact_pool/src/canister_http_pool.rs
+++ b/rs/artifact_pool/src/canister_http_pool.rs
@@ -25,7 +25,7 @@ use prometheus::IntCounter;
 const POOL_CANISTER_HTTP: &str = "canister_http";
 const POOL_CANISTER_HTTP_CONTENT: &str = "canister_http_content";
 
-type ValidatedCanisterHttpPoolSection = PoolSection<CanisterHttpResponseShare, ResponseGossip>;
+type ValidatedCanisterHttpPoolSection = PoolSection<CanisterHttpResponseShare, ServedResponse>;
 
 type UnvalidatedCanisterHttpPoolSection =
     PoolSection<CanisterHttpResponseShare, CanisterHttpResponseArtifact>;
@@ -37,7 +37,7 @@ type ContentCanisterHttpPoolSection =
 /// when the artifact is served to peers that *pull* it via
 /// [`ValidatedPoolReader::get`].
 #[derive(Clone, Copy)]
-enum ResponseGossip {
+enum ServedResponse {
     /// The request is not fully replicated (`NonReplicated` / `Flexible`): the
     /// `CanisterHttpResponse` is produced by only a subset of nodes, so it must
     /// be included with the artifact.
@@ -47,7 +47,7 @@ enum ResponseGossip {
     Exclude,
 }
 
-impl HasLabel for ResponseGossip {
+impl HasLabel for ServedResponse {
     fn label(&self) -> &str {
         ""
     }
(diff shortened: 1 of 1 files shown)


rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -25,14 +25,34 @@ use prometheus::IntCounter;
...
+    /// be included with the artifact.
+    Include,
+    /// The request is fully replicated: every node recomputes the response
+    /// locally, so it must be excluded from the artifact.
+    Exclude,
+}
+
+impl HasLabel for ServedResponse {
+    fn label(&self) -> &str {
+        ""
```
schneiderstefan: Why not include an actual label?

schneiderstefan (approved):

rs/artifact_pool/src/canister_http_pool.rs
eichhorl:
```suggestion
        "served_response"
```

eichhorl: This is what it was before, but before the type was `()`. I changed it to an actual label now.

PR #10742, opened 2026-07-13
fix(governance): Correct DeleteSubnet proposal description

mraszyk:
Update the note "Currently limited to CloudEngine subnets" clause (no longer accurate after https://github.com/dfinity/ic/commit/65ad179d4ae53ff4244d57829f5cfe368e680f0f) to saying that system subnets cannot be deleted and add the removal of the subnet from the subnet list, which the registry canister performs but the description omitted.

### DIFF
--- a/rs/nns/governance/api/src/types.rs
+++ b/rs/nns/governance/api/src/types.rs
@@ -4285,6 +4285,7 @@ pub enum NnsFunction {
     /// Delete a subnet. The subnet record, catch-up package, threshold signing key
     /// and routing table entries are removed from the registry, the subnet is
     /// removed from the subnet list, and the subnet's nodes become unassigned.
+    /// System subnets (e.g. the NNS or II subnet) cannot be deleted.
     DeleteSubnet = 57,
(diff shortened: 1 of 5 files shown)


rs/nns/governance/api/src/types.rs
```diff
@@ -4283,9 +4283,8 @@ pub enum NnsFunction {
     /// The proposal requests to split a subnet.
     SplitSubnet = 56,
     /// Delete a subnet. The subnet record, catch-up package, threshold signing key
-    /// and routing table entries are removed from the registry, and the subnet's
-    /// nodes become unassigned.
-    /// Currently limited to CloudEngine subnets.
+    /// and routing table entries are removed from the registry, the subnet is
+    /// removed from the subnet list, and the subnet's nodes become unassigned.
```
daniel-wong-dfinity-org-twin, line 4287: Perhaps, we can mention that deleting system subnets is not allowed (for now)?

daniel-wong-dfinity-org-twin (approved):

PR #10745, opened 2026-07-13
feat(node-rewards): targeted per-provider rewards reduction

pietrodimarco-dfinity:
## What

Adds a hardcoded, per-day rewards-reduction hook to the performance-based rewards algorithm (`rs/node_rewards/rewards_calculation/src/performance_based_algorithm/mod.rs`).

For node providers listed in `REDUCTION_TARGET_PROVIDERS`, on calendar days within `reduction_window()` (inclusive UTC), their **adjusted** rewards are scaled by `reduction_multiplier()` — `dec!(0.5)`, i.e. a 50% reduction.

The reduction is applied inside `calculate_rewards_for_date`, the single point both public endpoints flow through, so it affects:
- `get_node_providers_rewards(from_day, to_day)` — the summed XDR payout, and
- `get_node_providers_rewards_calculation(day)` — the per-day breakdown.

Base rewards are left untouched; each node's `adjusted_rewards_xdr_permyriad` is scaled and the truncated `total_adjusted_rewards_xdr_permyriad` is recomputed the same way (`.trunc().to_u64()`) as the original code.

## Status: ships inert ⚠️

`REDUCTION_TARGET_PROVIDERS` is **empty** and `reduction_window()` uses a **placeholder** date (`1970-01-01`), so `is_reduction_target` always returns `false` and behavior is unchanged. This is deliberate so it can't cut anyone's rewards by accident.

## Before this can go to mainnet

- [ ] Populate `REDUCTION_TARGET_PROVIDERS` with the actual provider principals.
- [ ] Set the real inclusive `reduction_window()` dates.
- [ ] Replace the `TODO` with a reference to the decision/proposal authorizing the reduction (this ships as a public NNS upgrade proposal that node providers can diff).
- [ ] Confirm window semantics (inclusive both ends, matched against reward day, UTC).

## Notes

This is a hardcoded intervention that bypasses the performance-based algorithm and should be removed once no longer needed.

### DIFF
--- a/rs/node_rewards/rewards_calculation/src/performance_based_algorithm/mod.rs
+++ b/rs/node_rewards/rewards_calculation/src/performance_based_algorithm/mod.rs
@@ -17,11 +17,77 @@ use rust_decimal::prelude::{FromPrimitive, ToPrimitive, Zero};
 use rust_decimal_macros::dec;
 use std::cmp::max;
 use std::collections::{BTreeMap, HashMap};
+use std::str::FromStr;
 
 pub mod results;
 pub mod test_utils;
 pub mod v1;
 pub mod v2;
+
+// ================================================================================================
+// TARGETED REWARDS REDUCTION (hardcoded, temporary intervention)
+// ================================================================================================
+//
+// Applies a fixed multiplier to the *adjusted* rewards of a specific set of node providers, for
+// every calendar day within an inclusive UTC window. This changes the XDR amounts returned by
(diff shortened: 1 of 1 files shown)


pietrodimarco-dfinity:
Closing in favor of #10746, which implements the same 50% node provider reward reduction at the canister boundary (keeping the versioned reward-calculation algorithm pure), tied to NNS motion proposal 142724, with the real provider list and test coverage.

PR #10751, opened 2026-07-13
fix(SDK-1936): Kill and reap the PocketIC server when all tests finish

basvandijk:
## What

`start_server` used to spawn the PocketIC server, detach it into its own process group (`process_group(0)`) and discard the `Child` handle. As a result the server — and the canister-sandbox processes it re-execs, which inherit the server's stdio — outlived the test process and shut down only on its TTL (this was the TODO: [SDK-1936](https://dfinity.atlassian.net/browse/SDK-1936)).

A lingering server keeps the test action's inherited stdout/stderr pipe write-ends open past process exit. On some CI runners that don't do proper PID namespacing (like on RBE runners @ Namespace) that only consider an action complete once those pipes reach EOF, this caused spurious failures.

In the past we also observed many pocket-ic-server zombie processes existing on our self-hosted darwin runners while their associated bazel invocations were long gone.

## Change

- Every spawned server is transferred to a process-global registry (`server_process.rs`) that owns the `Child` exclusively. This covers servers spawned implicitly through `PocketIc`/`PocketIcBuilder`, not just direct `start_server` callers.
- On unix, a `libc::atexit` callback runs when the process exits **normally** (for a test binary: once all subtests have finished and the harness is exiting) and, for each registered server:
  - sends `SIGTERM` to the whole **process group** — so the launcher/sandbox children die too and release the inherited pipes — escalating to `SIGKILL` after a short grace period. All groups are signalled up front and reaped against one shared deadline, so process exit is delayed by at most ~2×2s in total, no matter how many servers are still running;
  - `waitpid`-reaps the direct child;
  - removes the killed server's port file (the server only cleans it up itself on graceful shutdown), so a stale `reuse: true` port file can't hand a later process with a recycled pid a dead URL.
  - The cleanup does **not** run if the process terminates abnormally (killed by a signal, abort); the server then falls back to shutting down on its TTL, as before.
- `start_server` now returns a `ServerHandle` instead of a `std::process::Child`:
  - `ServerHandle::kill_and_wait` — terminate the server now (see `tests/unix.rs`);
  - `ServerHandle::pid` — the OS pid of the spawned process (replaces `Child::id`);
  - `ServerHandle::detach` — deliberately opt back into the old detached behaviour when the server must outlive the spawning process (it is then bounded by its TTL again).
- Registration opportunistically cleans up candidates that already exited (e.g. `reuse: true` candidates that lost the port-file race), reaping their zombies instead of accumulating them for the lifetime of the process — this also closes [SDK-1936](https://dfinity.atlassian.net/browse/SDK-1936)'s zombie leak during long runs, not just at exit — and the `#[allow(clippy::zombie_processes)]` workaround is removed.
- `fork()` is handled: a forked child never kills the parent's servers (the reaper checks the registry's owner pid), and a forked child that spawns servers of its own adopts the registry so they are still cleaned up at its exit.

This is gated behind `#[cfg(unix)]`; Windows behaviour is unchanged (the server is still bounded by its TTL there).

## API impact

Breaking change to the public `start_server` (return type) plus the new `ServerHandle` type — documented under `## Unreleased` in `CHANGELOG.md` with migration guidance; warrants a major version bump at release time.

## Testing

Two new regression tests in `tests/unix.rs` pin the new behaviour by re-invoking the test binary as a helper that spawns a server and exits:

- `server_is_killed_when_process_exits` asserts the helper's stdout reaches EOF (i.e. the server and its sandbox children released the inherited pipes — exactly the CI hang this PR fixes) and that the server pid is gone afterwards;
- `detached_server_survives_process_exit` asserts a `detach()`ed server outlives the helper.

`rustfmt`, `clippy --all-features --all-targets -p pocket-ic` (strict flags), and `bazel build` all clean. Ran the pocket-ic package suites, the `pocket_ic_server` integration tests (incl. the gateway / live-mode / bitcoind / dogecoind cases that spawn long-lived servers and sidecars), and the full set of ~148 bazel test targets depending on the `pocket-ic` crate (NNS/SNS integration tests, ledger, rosetta, …) — all pass, and every test binary exits cleanly, confirming the atexit reaper runs without hanging.

Builds and tests involving pocket-ic also [succeed](https://github.com/dfinity/ic/actions/runs/29412516732/job/87342706386?pr=10760#step:9:7754) on the RBE cluster @ Namespace.

[SDK-1936]: https://dfinity.atlassian.net/browse/SDK-1936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/pocket-ic/Cargo.toml
+++ b/packages/pocket-ic/Cargo.toml
@@ -48,6 +48,9 @@ tracing = { workspace = true }
 tracing-appender = { workspace = true }
 tracing-subscriber = { workspace = true }
 
+[target.'cfg(unix)'.dependencies]
+libc = { workspace = true }
+
 [target.'cfg(windows)'.dependencies]
 wslpath = "0.0.2"
 
--- a/packages/pocket-ic/Cargo.toml
+++ b/packages/pocket-ic/Cargo.toml
@@ -88,5 +88,6 @@ ic-registry-proto-data-provider = { path = "../../rs/registry/proto_data_provide
 ic-registry-transport = { path = "../../rs/registry/transport" }
 ic-types = { path = "../../rs/types/types" }
 ic-universal-canister = { path = "../../rs/universal_canister/lib" }
+libc = { workspace = true }
 maplit = "1.0.2"
 prost = { workspace = true }
--- a/packages/pocket-ic/Cargo.toml
+++ b/packages/pocket-ic/Cargo.toml
@@ -88,6 +88,5 @@ ic-registry-proto-data-provider = { path = "../../rs/registry/proto_data_provide
 ic-registry-transport = { path = "../../rs/registry/transport" }
 ic-types = { path = "../../rs/types/types" }
 ic-universal-canister = { path = "../../rs/universal_canister/lib" }
-libc = { workspace = true }
 maplit = "1.0.2"
 prost = { workspace = true }
--- a/packages/pocket-ic/src/server_process.rs
+++ b/packages/pocket-ic/src/server_process.rs
@@ -68,8 +68,10 @@ fn lock_registry() -> MutexGuard<'static, Vec<Option<ServerProc>>> {
 /// process-global registry and returns a [`ServerHandle`] referring to it.
 ///
 /// The registry keeps the child alive for the remainder of the process (so the server
-/// stays up for the whole test run) and guarantees it is killed and reaped at process
-/// exit, replacing the previous "spawn and leak" behaviour.
+/// stays up for the whole test run). On unix it is then killed (its process group) and
+/// reaped at process exit on a bounded, best-effort basis, replacing the previous
+/// "spawn and leak" behaviour; on other platforms the child is not reaped automatically
+/// and shuts down on its TTL.
 pub(crate) fn register(child: Child) -> ServerHandle {
     #[cfg(unix)]
     let pgid = child.id() as libc::pid_t;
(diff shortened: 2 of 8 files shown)


packages/pocket-ic/tests/unix.rs
```diff
@@ -935,3 +935,122 @@ fn test_cost_threshold_keys() {
...
+    let (server, server_url) = start_server(StartServerParams::default()).await;
+    // Make sure the server is fully up before exiting.
+    let resp = reqwest::Client::new()
+        .get(server_url.join("instances").unwrap())
+        .send()
+        .await
+        .unwrap();
+    assert_eq!(resp.status(), StatusCode::OK);
+    if mode == "detach" {
+        server.detach();
```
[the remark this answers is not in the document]
basvandijk: Switched to `std::env::var` in 9d0caa2314 for clarity. Small correction though: the original code did compile and the test passed — `std` provides `impl PartialEq<&str> for OsString`, so the comparison was valid; `var` is just the simpler choice here since the value is always UTF-8.

packages/pocket-ic/src/server_process.rs
```diff
@@ -0,0 +1,439 @@
...
+/// behaviour — hence the `catch_unwind`.
+#[cfg(unix)]
+extern "C" fn reap_all() {
+    let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        // atexit registrations and the registry's memory are inherited across `fork()`:
+        // a forked child exiting normally must not kill the servers its parent is still
+        // using.
+        // SAFETY: getpid cannot fail and has no preconditions.
+        if unsafe { libc::getpid() } != registry().owner_pid {
+            return;
```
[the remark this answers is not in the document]
basvandijk: Implemented in 9d0caa2314: `owner_pid` is now an `AtomicI32`, and the first registration after a fork disowns the inherited entries (slots are cleared in place, so inherited `ServerHandle`s go inert instead of aliasing new entries — dropping a `Child` neither signals nor waits, so the parent's servers are untouched) and adopts the registry by updating `owner_pid`. The inherited atexit hook then reaps exactly the child's own servers. Children that never spawn servers keep the old fast path: `reap_all` returns before touching the (potentially fork-poisoned) registry lock.

packages/pocket-ic/Cargo.toml
```diff
@@ -85,5 +88,6 @@ ic-registry-proto-data-provider = { path = "../../rs/registry/proto_data_provide
 ic-registry-transport = { path = "../../rs/registry/transport" }
 ic-types = { path = "../../rs/types/types" }
 ic-universal-canister = { path = "../../rs/universal_canister/lib" }
+libc = { workspace = true }
 maplit = "1.0.2"
```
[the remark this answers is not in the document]
basvandijk: Removed in 9d0caa2314 — verified with `cargo check --tests -p pocket-ic` that the integration tests compile against the crate's `cfg(unix)` normal dependency alone (the bazel `unix` test target keeps its explicit `libc` dep, since bazel deps don't propagate that way).

packages/pocket-ic/src/server_process.rs
```diff
@@ -0,0 +1,462 @@
+//! Process-lifetime management for the spawned `pocket-ic-server`.
+//!
+//! Historically [`crate::start_server`] spawned the server, detached it into its own
+//! process group (`process_group(0)`) and then discarded the [`Child`] handle, so the
+//! server — and the canister-sandbox processes it re-execs — outlived the spawning
+//! process. Under a remote-execution worker that only completes once the test action's
+//! stdout/stderr pipes reach EOF, the lingering server (and its sandbox children, which
+//! inherit those pipes) kept the pipe write-ends open, and the action failed with
+//! `WaitDelay expired before I/O complete`.
```
mraszyk, line 9: If inheriting the pipes is the issue, then an easier solution would be to do
```
  cmd.stdin(std::process::Stdio::null());
  cmd.stdout(std::process::Stdio::null());
  cmd.stderr(std::process::Stdio::null());
```
when spawning the PocketIC server in `start_server` and let the pipes be not inherited. CC @basvandijk

mraszyk: Although that would likely hide all server output so it has its trade-off.

mraszyk: I checked the PocketIC server output and it basically consists of canister logs produced during the test run so it might be a reasonable trade-off to turn them off on CI.

basvandijk: Note that Namespace just implemented proper sandboxing so we no longer strictly need this. 

I do remember we used to have many stray pocket-ic-server zombies on our Darwin runners. So in general I think it's good that a test kills all processes it started. However this does add quite a bit of complexity which might be overkill.

@nmattia what's your opinion?

PR #10755, opened 2026-07-13
refactor: stop using log memory store feature flag and migration fields in production

mraszyk:
The one-time migration from the legacy `CanisterLog` store to the `LogMemoryStore` has completed on all subnets, so production code no longer needs to consult the feature flag or the per-canister migration status. This decouples the runtime logic from those fields while keeping them serialized for downgrade compatibility (unlike #10714, which was reverted).

- Drop the `log_memory_store_feature` flag from `Config` and all plumbing (canister manager config, `SystemState::new*`, `LogMemoryStore::new()`, `fetch_canister_logs`, scheduler). The `LOG_MEMORY_STORE_FEATURE` / `_ENABLED` consts are retained but now unused by production logic.
- Remove the migration scaffolding: the subnet-level `logs_migrated` metadata, the scheduler migration code (and `filter_canister_log_records`), the migration-duration metric, and the orphaned `set_migrated` / `clear_migrated` methods.
- Remove the per-canister migration flag from the runtime entirely: drop the `migrated` field from `LogMemoryStore`, the `is_migrated()` getter, and the `migrated` parameter from `from_checkpoint` / `SystemState::new_from_checkpoint`. `next_canister_log_record_idx` and `log_memory_store_migrated` (proto 44/66) are still serialized — `log_memory_store_migrated` now written as a constant `true` — and the legacy `canister_log` store is retained and kept up to date via dual-write, so checkpoints remain readable by replicas that predate the log memory store. Dropping the serialized fields entirely is left for a follow-up once this PR is rolled out to all subnets.


rs/replicated_state/src/canister_state/system_state/log_memory_store/mod.rs
```diff
@@ -137,23 +142,14 @@ impl LogMemoryStore {
 
     /// Returns `true` if the one-time migration from `CanisterLog` has already
     /// been performed for this canister.
+    ///
+    /// Always `true` for stores created after the migration was completed on
+    /// all subnets; retained only so that the value keeps round-tripping
+    /// through checkpoints for replicas that predate the log memory store.
     pub fn is_migrated(&self) -> bool {
```
schneiderstefan: I don't quite understand. All the other changes make sense if you assume is_migrated is always true. But we keep the function anyway for roundtrips? Can't we remove the getter here and hardcode true in the checkpoint writing logic?

mraszyk: Good point: https://github.com/dfinity/ic/pull/10755/commits/3eb3a78c1089bd013ee1a68f6d99f39f16a50e46

schneiderstefan (approved):

PR #10756, opened 2026-07-13
chore: bump ic-management-canister-types to 0.8.0 and agent-rs to 0.49.0

mraszyk:
Bump `ic-management-canister-types` 0.7.1 -> 0.8.0 and the agent-rs crates (`ic-agent`, `ic-utils`, `ic-transport-types`, `ic-identity-hsm`) 0.47.3 -> 0.49.0, and fix the resulting build breakages:

- ic-pub-key: `ic-vetkeys` 0.7.0 is still built against `ic-management-canister-types` 0.7.1, so `for_mainnet_key` expects `VetKDKeyId` from `ic-management-canister-types` 0.7.1. Bridge the two by rebuilding the key id from its fields via `ic-cdk-management-canister` (added as an optional dependency gated behind the `vetkeys` feature).
- several canisters: add the new `snapshot_visibility` field to `CanisterSettings` and/or `DefiniteCanisterSettings`.
- system-test driver: add the new `permissions` field to `ic_agent::identity::Delegation`.
- consensus upgrade test: `ic_utils` still depends on the older `ic-management-canister-types`, so call the management canister directly via the agent instead of its typed `take_canister_snapshot` helper.


rs/tests/execution/general_execution_tests/canister_lifecycle.rs
```diff
@@ -354,8 +357,8 @@ pub fn update_settings_multiple_controllers(env: TestEnv) {
                 logger,
                 "Assert that the user cannot access canister C's status."
             );
-            assert_http_submit_fails(
-                mgr.canister_status(&canister_c).call().await,
+            assert_reject(
```
mraszyk, line 360: this change does not force rejection by the node receiving the HTTP request, but this is fine since what matters that the call is rejected, not whether the reject is uncertified (by a single node -- this is what we asserted before) or certified (this is what we also allow now).

rs/tests/execution/canister_migration_test.rs
```diff
@@ -125,7 +125,8 @@ async fn add_controller(
 ) {
     let (status_result,) = management_canister
         .canister_status(&canister.canister_id())
-        .call_and_wait()
+        .as_update()
```
schneiderstefan, line 128: What triggered this change (and the very similar ones)? Is this the new agent version?

mraszyk: Yes — this is driven by the agent-rs bump.
  
  In ic-utils ≤ 0.48.1, `canister_status` returned an `impl AsyncCall`, so `.call_and_wait()` always issued a replicated update call:
 
```
  pub fn canister_status(&self, canister_id: &Principal)
      -> impl 'agent + AsyncCall<Value = (CanisterStatusResult,)>
 ```

  In 0.49.0 it now returns a new `QueryOrUpdateCall` builder that defaults to a (fast, non-replicated) query call:

```
  /// Issued as a (fast, non-replicated) query call by default; call
  /// `QueryOrUpdateCall::as_update` on the returned builder for a replicated
  /// update call instead.
  pub fn canister_status(&self, canister_id: &Principal)
      -> QueryOrUpdateCall<'agent, '_, (CanisterStatusResult,)>
 ```

  So to preserve the previous behavior (replicated update call), each call site now has to add `.as_update()` before `.call()`. `.call_and_wait()` no longer exists on this builder — `.call()` does a
  single query round-trip by default, or an update call + polling once `.as_update()` is set.

schneiderstefan (approved):

basvandijk (approved):

mbjorkqvist (approved):

daniel-wong-dfinity-org-twin (approved):
Governance-owned files LG.

However, I think there is one more missing thing in cmc.did before we can upgrade the canister. Please, compare to cmc.did to what the canister currently offers. E.g.

```
dfx canister metadata --ic rkp4c-7iaaa-aaaaa-aaaca-cai 'candid:service'
```

Alternatively, I'm pretty sure it's running code at git commit ID `e114e3d7f4efef3dc9876b619e71deb79868640e`.

mraszyk:
> Governance-owned files LG.
> 
> However, I think there is one more missing thing in cmc.did before we can upgrade the canister. Please, compare to cmc.did to what the canister currently offers. E.g.
> 
> ```
> dfx canister metadata --ic rkp4c-7iaaa-aaaaa-aaaca-cai 'candid:service'
> ```
> 
> Alternatively, I'm pretty sure it's running code at git commit ID `e114e3d7f4efef3dc9876b619e71deb79868640e`.

I'm getting the following on this branch:
```
$ dfx canister metadata --ic rkp4c-7iaaa-aaaaa-aaaca-cai 'candid:service' | diff - rs/nns/cmc/cmc.did
24a25
>   log_memory_limit : opt nat;
28d28
<   log_memory_limit : opt nat;
```
while the `log_memory_limit` field is merely moved around so the candid interface does match the deployed canister on this branch.

PR #10757, opened 2026-07-13
feat: Further limit message memory on small nodes

schneiderstefan:
This PR addresses an issue when running a subnet with nodes that have limited RAM. The message memory, i.e. the storage required by messages in queues, used to be bound only by a constant of 20 GB. This is however not safe as they are currently kept fully in RAM, and some newer nano nodes will have as little as 8 GB of RAM.

Those subnets however already limit the heap delta using a parameter in the registry, as the heap delta is also stored in RAM. This commit simply reuses the heap delta limit to also give an upper bound on the message memory. Both bounded and unbounded wait messages are now limited to min(heap_delta/3, old_constant). On large nodes, this commit makes no difference at all.

Most code changes are due to the fact that message memory is no longer a fixed value that can be set at replica startup, and accompanying necessary changes to the test framework.

### DIFF
--- a/rs/config/src/execution_environment.rs
+++ b/rs/config/src/execution_environment.rs
@@ -46,21 +46,6 @@ const SUBNET_MEMORY_THRESHOLD: NumBytes = NumBytes::new(750 * GIB);
 /// IC protocol requires storing copies of the canister state.
 const SUBNET_MEMORY_CAPACITY: NumBytes = NumBytes::new(2 * TIB);
 
-/// This is the upper limit on how much memory can be used by all guaranteed
-/// response canister messages on a given subnet.
-///
-/// Guaranteed response message memory usage is calculated as the total size of
-/// enqueued guaranteed responses; plus the maximum allowed response size per
-/// reserved guaranteed response slot.
-const SUBNET_GUARANTEED_RESPONSE_MESSAGE_MEMORY_CAPACITY: NumBytes = NumBytes::new(15 * GIB);
-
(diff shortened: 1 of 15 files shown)


daniel-wong-dfinity-org-twin (approved):
Approving tiny change in Governance-owned code.

rs/config/src/execution_environment.rs
```diff
@@ -110,7 +110,53 @@ pub const STOP_CANISTER_TIMEOUT_DURATION: Duration = Duration::from_secs(5 * 60)
...
+/// message memory relative to a subnet's heap delta capacity (which is itself
+/// sized relative to available RAM).
+fn message_memory_capacity(
+    configured_default_capacity: NumBytes,
+    heap_delta_capacity: NumBytes,
+) -> NumBytes {
+    configured_default_capacity.min(NumBytes::new(
+        heap_delta_capacity.get() / MESSAGE_MEMORY_HEAP_DELTA_DIVISOR,
+    ))
+}
```
alin-at-dfinity: At this point, I don't think it makes much sense to have this logic in `ic_config::execution_environment` anymore. I'd just consolidate it all under `ic_replicated_state::metadata_state` / `ic_replicated_state::metadata_state::testing`.

alin-at-dfinity (approved):
LGTM.

rs/config/src/execution_environment.rs
schneiderstefan: Done.

PR #10763, opened 2026-07-14
fix: only run //rs/tests/message_routing/xnet:subnet_delete_test on Farm

basvandijk:
`//rs/tests/message_routing/xnet:subnet_delete_test_local` [fails](https://dash.dm1-idx1.dfinity.network/invocation/e78f0983-049f-4460-8c15-0735eb21ec37?target=%2F%2Frs%2Ftests%2Fmessage_routing%2Fxnet%3Asubnet_delete_test_local&targetStatus=6) so we configure the test to run on Farm only.

### DIFF
--- a/rs/tests/message_routing/xnet/BUILD.bazel
+++ b/rs/tests/message_routing/xnet/BUILD.bazel
@@ -140,6 +140,8 @@ system_test(
 
 system_test(
     name = "subnet_delete_test",
+    backend = "farm",  # Too big for the "local" backend.
+    cpus = MIN_LOCAL_CPUS + (1 + 5 * 4) * DEFAULT_VCPUS_PER_VM,  # 1 API BN + 5 subnets * 4 nodes * 6 vCPUs each
     tags = [
         "long_test",
     ],


rs/tests/message_routing/xnet/BUILD.bazel
```diff
@@ -140,6 +140,8 @@ system_test(
 
 system_test(
     name = "subnet_delete_test",
+    backend = "farm",  # Too big for the "local" backend.
+    cpus = MIN_LOCAL_CPUS + (1 + 5 * 4) * DEFAULT_VCPUS_PER_VM,  # 1 API BN + 5 subnets * 4 nodes * 6 vCPUs each
```
mraszyk, line 144: Ideally, this would be derived from the declaration inside the test, but I suppose it's too late at that point.

mraszyk (approved):

rs/tests/message_routing/xnet/BUILD.bazel
basvandijk: Yeah unfortunately `setup()` functions are dynamic and not declarative / static.

basvandijk: Of course you could imagine having a `declarative_system_test` bazel macro that would allow configuring the testnet configuration via bazel. Then the macro would know the amount of vCPUs statically and can generate a `setup()` function.

PR #10771, opened 2026-07-14
chore: cache mainnet ICOS image downloads via sha256

basvandijk:
## Why

During the `Bazel Build with RBE` step of `rbe-namespace-test.yml`, the execution profile shows many long "Fetching repository" traces — e.g. `@@+mainnet_icos_images+mainnet_app_images` takes **over 5 minutes** because the `mainnet_icos_images` repo rule re-downloads large ICOS image artifacts from `download.dfinity.systems` on every run.

Root cause: `bazel/mainnet-icos-images.bzl` called `repository_ctx.download(...)` **without a `sha256`**. The Namespace RBE job enables Bazel's Remote Asset API (`nsc bazel execution setup --enable_remote_asset_api`), which can serve a blob straight from CAS — and Bazel's local `--repository_cache` can be used — **only when the expected content digest is supplied**. So supplying a correct `sha256` per download turns these multi-minute fetches into cache hits.

This is the low-risk alternative to the experimental `--experimental_remote_repo_contents_cache` (which would require dropping `repository_ctx.watch()` and returning `repo_metadata(reproducible = True)`).

## What

- **`bazel/mainnet-icos-images.bzl`**: pass `sha256` to both `download()` calls (dev/prod aware).
  - guest-os update image → existing `update_img_hash` / `update_img_hash_dev`.
  - setup-os disk image → new `setupos_disk_img_hash` / `setupos_disk_img_hash_dev`.
- **`ci/src/mainnet_revisions/mainnet_revisions.py`**: record the setup-os disk-image hash in the JSON, sourced from the CDN `setup-os/disk-img[-dev]/SHA256SUMS` files (setup-os is not NNS-elected, so no proposal hash exists). The "already up to date" guards now also backfill entries that predate the new field.
- **`mainnet-icos-revisions.json`**: backfilled `setupos_disk_img_hash[_dev]` for the current pinned versions.

Naming distinguishes the SetupOS disk image (`setupos_disk_img_hash`) from the GuestOS update image (`update_img_hash`).

## Verification

- `bazel build` of the app-image repos (prod + dev) and the hostos repos succeeds → Bazel sha256-verifies every download; a wrong hash aborts the fetch with a checksum mismatch. Re-fetching a warm artifact drops from minutes to a cache hit.
- Python compiles; a mocked unit check confirms the fetchers populate the new fields from the correct SHA256SUMS URLs; `buildifier` clean.

### DIFF
--- a/ci/src/mainnet_revisions/mainnet_revisions.py
+++ b/ci/src/mainnet_revisions/mainnet_revisions.py
@@ -31,6 +31,11 @@ class VersionInfo:
     dev_hash: str
     launch_measurements: dict
     dev_measurements: dict
+    # SHA256 of the setup-os disk-img.tar.zst (prod and dev channels). Unlike
+    # `hash` (the NNS-elected update image), the setup-os image is not elected
+    # via the NNS, so this is read from the CDN SHA256SUMS file.
+    setupos_hash: str
+    setupos_dev_hash: str
 
 
 def sync_main_branch_and_checkout_branch(
@@ -162,7 +167,13 @@ def get_replica_version_info(replica_version: str) -> VersionInfo:
         f"https://download.dfinity.systems/ic/{version}/guest-os/update-img-dev/launch-measurements.json"
     )
 
-    return VersionInfo(version, hash, dev_hash, launch_measurements, dev_measurements)
+    setup_base = f"https://download.dfinity.systems/ic/{version}/setup-os"
+    setupos_hash = download_and_read_sha256sums(f"{setup_base}/disk-img/SHA256SUMS", "disk-img.tar.zst")
+    setupos_dev_hash = download_and_read_sha256sums(f"{setup_base}/disk-img-dev/SHA256SUMS", "disk-img.tar.zst")
+
+    return VersionInfo(
+        version, hash, dev_hash, launch_measurements, dev_measurements, setupos_hash, setupos_dev_hash
+    )
 
 
 def get_latest_replica_version_info() -> VersionInfo:
@@ -192,7 +203,13 @@ def get_latest_replica_version_info() -> VersionInfo:
         f"https://download.dfinity.systems/ic/{version}/guest-os/update-img-dev/launch-measurements.json"
     )
 
-    return VersionInfo(version, hash, dev_hash, launch_measurements, dev_measurements)
+    setup_base = f"https://download.dfinity.systems/ic/{version}/setup-os"
+    setupos_hash = download_and_read_sha256sums(f"{setup_base}/disk-img/SHA256SUMS", "disk-img.tar.zst")
+    setupos_dev_hash = download_and_read_sha256sums(f"{setup_base}/disk-img-dev/SHA256SUMS", "disk-img.tar.zst")
+
+    return VersionInfo(
+        version, hash, dev_hash, launch_measurements, dev_measurements, setupos_hash, setupos_dev_hash
+    )
 
 
 def get_latest_hostos_version_info(logger: logging.Logger) -> VersionInfo:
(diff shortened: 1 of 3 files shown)


ci/src/mainnet_revisions/mainnet_revisions.py
```diff
@@ -354,6 +385,18 @@ def download_and_read_file(url: str):
...
 
+def download_and_read_sha256sums(url: str, filename: str) -> str:
+    """Download a SHA256SUMS file and return the hex sha256 recorded for `filename`."""
+    with tempfile.NamedTemporaryFile() as tmp_file:
+        urllib.request.urlretrieve(url, tmp_file.name)
+        with open(tmp_file.name, "r", encoding="utf-8") as f:
+            for line in f:
+                parts = line.split()
+                if len(parts) == 2 and parts[1] == filename:
+                    return parts[0]
```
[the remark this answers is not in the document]
basvandijk: In our case the `SHA256SUMS` file looks like:
```
curl 'https://download.dfinity.systems/ic/8a655ede1f59ea0cb3079a9aa0601b7905b7ca37/setup-os/disk-img/SHA256SUMS'
9f8bbe68b592592344cfbe70485892f4061229072b7b9dd9a8f6f4e2c5e06864 disk-img.tar.zst
```

Bownairo (approved):

ci/src/mainnet_revisions/mainnet_revisions.py
```diff
@@ -354,6 +381,18 @@ def download_and_read_file(url: str):
             return json.loads(f.read().decode())
 
 
+def download_and_read_sha256sums(url: str, filename: str) -> str:
```
nmattia, line 384: how about computing the sha256 explicitly (locally) instead of trusting the upstream? that would also mean we don't rely on any specific format or upstream infra

mainnet-icos-revisions.json
```diff
@@ -272,6 +276,8 @@
       "version": "992628c4f26cc7320396b6b8be37133a666a4386",
       "update_img_hash": "82d7e19ee6b708984db97bc1dbf03fd6188be01607ce1d13ef84fe6fc62d274a",
       "update_img_hash_dev": "6f50d00ae6c74d4cf042259528ab7568d9f81215f005d38ca2fbd5bd659c1aa8",
+      "setupos_disk_img_hash": "50bcb55955d60b45d4d09c734fb4e8c49567413b11e1565e66b8fe63a0769c4d",
+      "setupos_disk_img_hash_dev": "56b31731089f233f0672a5bbe9b9093c9d3c9c9a5d61b9e7b4c1a665b361f394",
```
nmattia, line 280: out of curiosity, where did the original hashes come from and why was there none for setupos?

nmattia (commented):
LGTM, though would prefer calculating the hashes explicitly in  mainnet_revisions.py!

Also, am a bit surprised that the repo rule itself is not updated to return the hashes; did I miss this?

ci/src/mainnet_revisions/mainnet_revisions.py
basvandijk: My reasoning for using the SHA256SUMS file was that if there is a mismatch we'll catch it. If we calculate it ourselves there can be a mismatch which will go unnoticed.

mainnet-icos-revisions.json
basvandijk: IDK. @Bownairo do you know?

basvandijk: Or do you mean with "original hashes" the existing hashes like`update_img_hash`? If so, they come from the upgrade proposals in the NNS. SetupOS is not upgraded via the NNS which is why we can't get its hashes from it.

Bownairo:
I remember doing something like this when I first wrote it, does it make sense to bring something similar back?
https://github.com/dfinity/ic/blob/35321e206851e2519c029c7063ef2701210ecf61/bazel/mainnet-icos-images.bzl#L92-L105

(Is there any meaningful difference between integrity and hash? 🤔 It doesn't seem like it.)

This would depend on the download directly, so we would not have:
> If we calculate it ourselves there can be a mismatch which will go unnoticed.[^1]


[^1]: https://github.com/dfinity/ic/pull/10771#discussion_r3581032564

basvandijk:
As discussed [here](https://dfinity.slack.com/archives/C026JQ1B2AF/p1784037020801069), I'll merge this now to get the caching we want and then we evaluate how this holds up over time.

PR #10773, opened 2026-07-14
chore: WIP drop libcap

nmattia:


rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -248,68 +224,168 @@ impl LocalBackend {
         Ok(())
     }
```
basvandijk, line 249: We should just get rid of these whole `net_admin` and `run_net_admin` functions then right?

PR #10782, opened 2026-07-15
test(execution): reject mgmt canister /call under queries-only delegation

aterga:
## Summary

- Adds a new `queries_only_delegation_test` system-test target in `rs/tests/execution/` that verifies the [queries-only delegation permissions](https://github.com/dfinity/ic/pull/10449) feature end-to-end.
- The test authenticates every request through a delegation chain whose top delegation carries `permissions = queries`, iterates over every variant of `ic_management_canister_types_private::Method`, and POSTs an ingress `/call` to `aaaaa-aa` for each method.
- Each request must be rejected during ingress validation with `400 Bad Request` and a body containing `"Update calls are not permitted"`.

## Why raw HTTP

`ic-agent 0.47.3` (the version this workspace is currently on) does not expose the `permissions` field on `Delegation`; that landed in `ic-agent 0.48.0` (crates.io release [`agent-rs#734`](https://github.com/dfinity/agent-rs/pull/734)). Once the workspace bumps to 0.48+, this test can migrate to `ic-agent::identity::{Delegation, DelegatedIdentity}` directly and drop the raw-HTTP scaffolding.

## Test plan

- [ ] `bazel test //rs/tests/execution:queries_only_delegation_test` passes on CI
- [ ] Extending `Method` (e.g. a new IC00 endpoint) automatically exercises the new variant — the loop is driven by `strum::IntoEnumIterator`


rs/tests/execution/queries_only_delegation_test.rs
```diff
@@ -0,0 +1,157 @@
...
+            .signature
+            .expect("delegation signature bytes");
+        let signed_delegation = SignedDelegation::new(delegation, delegation_sig_bytes);
+
+        let url = format!("{}api/v2/canister/{}/call", node_url, CanisterId::ic_00());
+
+        for method in Ic00Method::iter() {
+            let content = HttpCallContent::Call {
+                update: HttpCanisterUpdate {
+                    canister_id: Blob(CanisterId::ic_00().get().to_vec()),
```
mraszyk: Why only this special canister and not (also) a regular canister?

aterga: Added a `regular_canister_call_rejected_with_queries_only_delegation` test that installs a UC via the default agent and asserts an ordinary `update` call is rejected with the same 400 / queries-only reason (shared helpers `build_queries_only_agent` and `assert_update_rejected`).

mraszyk (approved):

PR #10785, opened 2026-07-15
feat: Introduce system_metadata in the state tree

schneiderstefan:
This commit introduces a leaf system_metadata in the state tree for each canister. It's a CBOR encoded map that currently only contains last_install_timestamp, that corresponds to the block time when the wasm binary was last installed, reinstalled, upgraded, or restored from a snapshot. The map will be empty for uninstalled canisters and canisters that were last installed before the leaf was introduced.

Note that the information can also be found in the canister's history. However, the size of the history is limited, and hence it might not be available there anymore.

This commit introduces the tracking of the timestamp on each new install, and also bumps the maximum supported certification version. The full rollout will include a follow-up where the current certification version is bumped, and the new paths are whitelisted for read_state requests.


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -568,6 +568,14 @@ pub struct SystemState {
     /// Canister global timer.
     pub global_timer: CanisterTimer,
 
+    /// The round time at which the canister's code was most recently deployed
+    /// (install, reinstall, or upgrade) or restored from a snapshot, in
+    /// nanoseconds since the Unix epoch. It is `None` when the canister has no
+    /// installed code (never installed, or uninstalled). It is also `None` for
+    /// canisters installed before this field was introduced, until their next
+    /// (re)install, upgrade, or snapshot restore.
+    pub last_install_timestamp: Option<Time>,
```
mraszyk: How about putting the field into `ExecutionState`? Then it's guaranteed to be `None` whenever the canister is uninstalled and we're forced to set it when creating a new `ExecutionState`.

rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -5114,6 +5117,48 @@ fn uninstall_code_clears_canister_state() {
...
+
+    // Upgrade updates the install time.
+    test.state_mut().metadata.batch_time += std::time::Duration::from_secs(1);
+    let upgrade_time = test.state().time();
+    assert!(upgrade_time > reinstall_time);
+    test.upgrade_canister(canister_id, UNIVERSAL_CANISTER_WASM.to_vec())
+        .unwrap();
+    assert_eq!(last_install_timestamp(&test), Some(upgrade_time));
+
+    // Uninstall clears the install time. This exercises the same
```
mraszyk: We could also consider testing the uninstall path due to the canister being out of cycles explicitly.

rs/replicated_state/src/canister_state/system_state.rs
schneiderstefan: Good point. I moved it there.

rs/execution_environment/src/canister_manager/tests.rs
schneiderstefan: Now that it's in ExecutionState, I think this wouldn't add much.

rs/canonical_state/certification_version/src/lib.rs
```diff
@@ -22,6 +22,8 @@ pub enum CertificationVersion {
     V25 = 25,
     /// Introdue `EngineNotAllowed` signals
     V26 = 26,
+    /// Add `/canister/<canister_id>/last_install_timestamp` leaf.
```
mraszyk, line 25:
```suggestion
    /// Add `/canister/<canister_id>/system_metadata` leaf.
```

rs/canonical_state/src/traversal.rs
```diff
@@ -434,6 +451,70 @@ mod tests {
...
+        let tree = replicated_state_as_lazy_tree(&state, Height::new(0));
+
+        for canister_id in [codeless_id, no_timestamp_id] {
+            let principal = canister_id.get();
+            let path: [&[u8]; 3] = [b"canister", principal.as_slice(), b"system_metadata"];
+            let leaf =
+                follow_path(&tree, &path).expect("every canister exposes `system_metadata` at V27");
+            let bytes = match leaf {
+                LazyTree::LazyBlob(f) => f(),
+                LazyTree::Blob(b, _) => b.to_vec(),
```
mraszyk: only one of this cases can occur, right?

mraszyk (approved):

rs/canonical_state/src/traversal.rs
schneiderstefan: [fef4cc0](https://github.com/dfinity/ic/pull/10785/commits/fef4cc0979426e5fd1efad48c39f479e7d2a9e30)

PR #10786, opened 2026-07-15
chore: set bazel cpu reservation via exec_properties instead of via the cpu:n tag

basvandijk:
Bazel doesn't forward the `cpu:n` tag, for reserving cpu cores, to the Remote Execution API. This means that on Remote Build Execution clusters (like @ Namespace) actions don't get the minimal CPUs they specified. To fix this we replace `cpu:n` tags with `exec_properties = {"cpu": "n"}` settings which are forwarded to the REAPI and work locally as well.

Note that this is supported since bazel-9.2.0. See: https://github.com/bazelbuild/bazel/issues/29419.


mbjorkqvist (approved):
Thanks @basvandijk!

nmattia (approved):
LGTM. For reference: https://github.com/bazelbuild/bazel/pull/29426

schneiderstefan (approved):

bjoernek (approved):

PR #10787, opened 2026-07-15
feat(registry): Added update_standard_engine_replica_version to Registry.

daniel-wong-dfinity-org-twin:
As usual, only callable by Governance. Will be used in the future to implement a new proposal type.

[prev]: https://github.com/dfinity/ic/pull/10774
[next]: https://github.com/dfinity/ic/pull/10840

### DIFF
--- a/rs/registry/canister/src/invariants/replica_version.rs
+++ b/rs/registry/canister/src/invariants/replica_version.rs
@@ -8,11 +8,13 @@ use crate::invariants::common::{
 
 use ic_base_types::SubnetId;
 use ic_protobuf::registry::{
-    replica_version::v1::ReplicaVersionRecord, subnet::v1::SubnetRecord,
-    unassigned_nodes_config::v1::UnassignedNodesConfigRecord,
+    replica_version::v1::ReplicaVersionRecord,
+    standard_engine_replica_version::v1::StandardEngineReplicaVersionRecord,
+    subnet::v1::SubnetRecord, unassigned_nodes_config::v1::UnassignedNodesConfigRecord,
 };
 use ic_registry_keys::{
-    make_replica_version_key, make_subnet_record_key, make_unassigned_nodes_config_record_key,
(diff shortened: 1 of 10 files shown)


bjoernek (approved):

rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
```diff
@@ -0,0 +1,471 @@
...
+/// deployment must start where the previous one left off, i.e. the new
+/// old_replica_version_id must match the current/old new_replica_version_id.
+impl Registry {
+    pub fn do_update_standard_engine_replica_version(
+        &mut self,
+        payload: UpdateStandardEngineReplicaVersionPayload,
+    ) {
+        println!("{LOG_PREFIX}do_update_standard_engine_replica_version: {payload:?}");
+
+        let new_record = StandardEngineReplicaVersionRecord::try_from(payload).unwrap();
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: ok, but then, the CODE is noisier. What you are suggesting is (IMHO) a "large" amount of code to achieve very little value. I mean, how much noisier is it??

rs/registry/canister/src/invariants/replica_version.rs
pierugo-dfinity, line 58: In practice, cloud engines following the upgrade train will have an empty replica version in their subnet record, which would not be possible because of these checks IIUC: `versions_in_use` would contain `""` (from `get_all_replica_versions_of_subnets`). From there, either `elected_set` also contains `""`, passing the first check but not the second, or it does not, passing the second check but not the first.

@daniel-wong-dfinity-org-twin PTAL

daniel-wong-dfinity-org-twin: Thanks. Yes, I know. Working on this now.

pierugo-dfinity: Great to hear thanks!

PR #10794, opened 2026-07-16
fix: account for Schnorr/VetKd/dropped-message cycles in consumed_cycles gauge

mraszyk:
`replicated_state_consumed_cycles_since_replica_started` was computed by manually adding only the deleted-canisters, ECDSA-outcall and HTTP-outcall subnet-level use cases to the per-canister totals. The `SchnorrOutcalls`, `VetKd` and `DroppedMessages` subnet-level use cases (which have no dedicated scalar field and live only in the by-use-case map) were omitted, so the gauge under-counted consumed cycles.

Add dedicated getters `get_consumed_cycles_schnorr_outcalls`, `get_consumed_cycles_vetkd` and `get_consumed_cycles_dropped_messages` on `SubnetMetrics` (reading from the by-use-case map) and include them in the gauge, for parity with the existing ECDSA/HTTP getters.


rs/replicated_state/src/metrics.rs
```diff
@@ -535,6 +535,21 @@ impl ReplicatedStateMetrics {
...
+        // Add the remaining subnet-level use cases. Unlike ECDSA/HTTP outcalls
+        // and deleted canisters, these have no dedicated scalar field, but their
+        // getters read the by-use-case map. The canister-level use cases in that
+        // map originate from deleted canisters and are already covered by
+        // `get_consumed_cycles_by_deleted_canisters()`.
+        consumed_cycles_total += state
+            .metadata
+            .subnet_metrics
+            .get_consumed_cycles_schnorr_outcalls();
+        consumed_cycles_total += state.metadata.subnet_metrics.get_consumed_cycles_vetkd();
```
eichhorl, line 547: Maybe we could implement a function `CyclesUseCase::is_charged_to_canister` with an exhaustive match, and then use `EnumIter` to iterate over all use cases here, and sum up only those where the function returns true.

Maybe something like that could also be used [here](https://github.com/dfinity/ic/pull/10842#discussion_r3622544673), or in other places that were previously forgotten because there was no exhaustive match.

eichhorl: Should we also add some tests?

mraszyk: > Maybe we could implement a function CyclesUseCase::is_charged_to_canister

The tricky part is that `HttpOutcalls` are collected at the subnet level, but since recently also in consumed_cycles_by_use_cases_as_counters at the canister level [here](https://github.com/dfinity/ic/blob/c63afc36cbe3fae931450cfda91ec3498c528bb9/rs/replicated_state/src/canister_state/system_state.rs#L2124) so there's no clean conceptual split. I'd suggest to use exhaustive matches as much as possible, but not to define a global split into canister/subnet cycles use cases. I added this task into the document tracking issues in cycles metrics on the IC and would suggest to do this as a follow-up.

> Should we also add some tests?

https://github.com/dfinity/ic/pull/10794/commits/6b9eab8bd7c110e2466f822e80c316ea0b4dc763

eichhorl (approved):

PR #10797, opened 2026-07-16
chore: do not run `@@+mainnet_icos_images+...//:guest-img` remotely

basvandijk:
The `@@+mainnet_icos_images+...//:guest-img` targets have 2.4G setupOS images as input. Running this remotely would mean these multi gigabyte inputs would have to be copied to the remote worker. Additionally Namespace currently has a 1G `/tmp` which causes the `:guest-img` action to run into a `No space left on device` error since it attempts to extract the 2.4G setupOS image under `/tmp`. So to work around both issues we forbid the action to run remotely (tag: `no-remote-exec`).

### DIFF
--- a/bazel/mainnet-icos-images.bzl
+++ b/bazel/mainnet-icos-images.bzl
@@ -78,7 +78,10 @@ genrule(
     name = "guest-img",
     srcs = ["disk-img.tar.zst"],
     outs = ["guest-img.tar.zst"],
-    tags = [ "manual" ],
+    # no-remote-exec because the setupOS image input is a multi gigabyte file
+    # which would then have to be copied to the remote worker.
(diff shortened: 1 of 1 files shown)


nmattia (approved):

bazel/mainnet-icos-images.bzl
```diff
@@ -78,7 +78,10 @@ genrule(
     name = "guest-img",
     srcs = ["disk-img.tar.zst"],
     outs = ["guest-img.tar.zst"],
-    tags = [ "manual" ],
+    # no-remote-exec because the setupOS image input is a multi gigabyte file
+    # which would then have to be copied to the remote worker.
+    # Additionally Namespace had a /tmp of only 1G which was too small to extract the setupOS image to.
```
basvandijk:
```suggestion
```

PR #10798, opened 2026-07-16
perf(state_manager): [DSM-145] Compute the manifest concurrently with validation

alin-at-dfinity:
Move manifest computation onto a dedicated thread so it overlaps `ValidateReplicatedStateAndFinalize` instead of running after it on the tip thread.

The `ComputeManifest` request is enqueued on the tip channel ahead of validation and forwarded by the tip thread to the manifest thread only once`TipToCheckpointAndSwitch` has finished serializing the checkpoint's protos. The computed `BundledManifest` is handed back and published from a paired `WaitForManifest` request enqueued on the tip channel after validation, so the root hash is only visible to Consensus once the checkpoint is verified.

At 200k idle canisters, the manifest (~28s to root hash) now runs entirely within the validation window (~31s), cutting batch->root-hash from ~44s to ~34s while also moving the post-publish rehash off the critical path.

Supporting changes:

- `files_with_sizes` skips checkpoint marker files during the directory walk. They are never part of a manifest, and skipping them by name avoids `stat`ing a marker that concurrent checkpoint finalization is removing, which would otherwise fail manifest computation. This makes the previous post-hoc marker exclusion in `compute_manifest` unnecessary.
- `flush_tip_channel` forwards its `Wait` barrier to the manifest thread as well, so it still waits for all outstanding checkpoint work -- including the post-publish rehash -- to drain.


rs/state_manager/src/tip.rs
```diff
@@ -1580,6 +1655,26 @@ fn handle_compute_manifest_request(
...
+fn publish_bundled_manifest(
+    log: &ReplicaLogger,
+    metrics: &StateManagerMetrics,
+    state_layout: &StateLayout,
+    states: &parking_lot::RwLock<SharedState>,
+    persist_metadata_guard: &Arc<Mutex<()>>,
+    height: Height,
+    bundled_manifest: crate::BundledManifest,
+) {
+    let mut states = states.write();
```
schneiderstefan, line 1749: How about checking here that the unverified marker is indeed gone?

rs/state_manager/src/tip.rs
```diff
@@ -178,11 +190,63 @@ pub(crate) fn spawn_tip_thread(
...
+                                &log,
+                            );
+                        }
+
+                        // Flush barrier (see `flush_tip_channel`): forwarded by the tip thread so a
+                        // flush also waits for in-flight manifest work — including rehash — to drain.
+                        TipRequest::Wait { sender } => {
+                            let _ = sender.send(());
+                        }
+                        _ => unreachable!("manifest thread only handles ComputeManifest and Wait"),
```
schneiderstefan: I think this should just be a different type with only those two options.

schneiderstefan (approved):

rs/state_manager/src/tip.rs
alin-at-dfinity: Done. It's a bit more verbose now, as I have to unpack e.g. the `TipRequest::ComputeManifest` and repack it as a `ManifestRequest::ComputeManifest`, but the `unreachable!()` is gone.

rs/state_manager/src/tip.rs
alin-at-dfinity: Done.

PR #10814, opened 2026-07-17
test: extract loopback accept-queue-saturation helper into ic-test-utilities-net

basvandijk:
## What

Extracts the loopback **accept-queue-saturation** trick — used to test connect-timeout behavior without network egress — into a new small, `tokio`-only, `testonly` crate `ic-test-utilities-net`, and migrates both existing copies to it.

New API (`rs/test_utilities/net/src/lib.rs`):

```rust
pub async fn saturated_loopback_listener() -> SaturatedTcpListener;
impl SaturatedTcpListener { pub fn addr(&self) -> SocketAddr; }
```

The returned guard owns both the listener **and** the blocker connections, so the address stays unconnectable for as long as the guard is alive (dropping it frees the port and empties the accept queue).

Both call sites collapse to two lines:

```rust
let _saturated = ic_test_utilities_net::saturated_loopback_listener().await;
let addr = _saturated.addr();
```

## Why

The trick was introduced in #10803 (https_outcalls `test_connect_timeout`) and then copied verbatim into the bitcoin adapter in #10813 (`initialization_times_out_after_five_seconds`) to drop its `requires-network` tag — leaving two ~55-line near-identical inline copies. This deduplicates them into one documented helper.

## Files

- **New:** `rs/test_utilities/net/{Cargo.toml,BUILD.bazel,README.md,src/lib.rs}`; added to root `Cargo.toml` workspace members.
- `rs/bitcoin/adapter/src/stream.rs` + `Cargo.toml` + `BUILD.bazel` (dev-dep + `//rs/test_utilities/net` in `adapter_test`).
- `rs/https_outcalls/adapter/tests/server_test.rs` + `Cargo.toml` + `BUILD.bazel` (dev-dep + `//rs/test_utilities/net` in both `adapter_integration` and `adapter_integration_with_http`).

## Verification

```
bazel test //rs/bitcoin/adapter:adapter_test \
  //rs/https_outcalls/adapter:adapter_integration \
  //rs/https_outcalls/adapter:adapter_integration_with_http
```
All three PASS in the network-less sandbox. `cargo check`/`clippy` clean on all three crates.

## Note

No new third-party crate was added (the helper depends only on `tokio`, already in the crate index), so no repin was needed.


mbjorkqvist (approved):
Nice, thanks @basvandijk!

michael-weigelt (approved):

PR #10816, opened 2026-07-17
chore(Core): Limit wasm locals #2

michael-weigelt:
Essentially reverts this revert: https://github.com/dfinity/ic/pull/10475
Metrics show that this is almost twice the max of existing canisters over all subnets.


schneiderstefan (approved):
Please mention in the description that this is more than any existing canisters are using, according to the metrics.

PR #10818, opened 2026-07-17
feat: avoid rustc LTO warnings

nmattia:
This patches the hermetic_cc_toolchain to drop the linker optimization flag specified by rustc: zig ignores & warns about them, which (since rust 1.97) is bubbled back up by rustc.


basvandijk (approved):
Nice, thanks!

PR #10821, opened 2026-07-19
refactor: drop the legacy canister_log store from SystemState

mraszyk:
The one-time migration from the legacy `CanisterLog` store to the `LogMemoryStore` has completed on all subnets and the dual-write kept the legacy store around only so checkpoints stayed readable by replicas that predate the log memory store (see #10755). That downgrade window has now elapsed and the migration code as well as additional sanity checks have already been removed, so the legacy store and its serialized fields can be fully removed.

- Remove the `SystemState.canister_log` field;
- Stop serializing `canister_log_records`, `next_canister_log_record_idx` and `log_memory_store_migrated`; the corresponding proto fields (43, 44, 66) are marked reserved.
- Replace the `StateMachine::canister_log` test helper with `canister_log_records` and `canister_log_bytes_used`, reading directly from `LogMemoryStore` (via a new test-only `LogMemoryStore::all_records_for_testing`), and migrate all callers.

The `CanisterLog` type and the transient `SystemStateModifications` delta log buffer are retained. This re-applies the parts of the reverted #10714 that were left as a follow-up in #10755.


rs/execution_environment/tests/canister_logging.rs
```diff
@@ -1746,46 +1744,12 @@ fn test_logging_of_long_running_dts_over_checkpoint() {
 const METRIC_BYTES_OVERHEAD_FACTOR: f64 = 1.05;
 const METRIC_PAYLOAD_SIZE: usize = 1_000;
 
-#[test]
-fn test_metric_canister_log_memory_usage_bytes_from_canister_log() {
```
mraszyk, line 1686: covered by test_metric_canister_log_memory_usage_bytes_from_log_memory_store

schneiderstefan (approved):

mbjorkqvist (approved):

PR #10825, opened 2026-07-20
fix: account for canister history memory usage

mraszyk:
Recording a canister history entry previously bypassed the subnet's memory and cycles accounting: `add_canister_change` applied the resulting execution-memory delta to the subnet available memory (letting it go negative), and it neither reserved storage cycles nor counted towards the freezing threshold. This PR makes canister history memory be accounted for like any other canister memory.

Additionally,
- the subnet memory usage invariant is tightened since canister history usage is now properly accounted towards subnet available memory;
- creating a canister with a low initial cycles balance in tests requires setting freezing threshold to 0 so that the canister history entry for canister creation does not make the canister frozen (which would fail the canister creation itself).


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -7249,6 +7336,169 @@ fn only_controllers_can_rename() {
...
+            .count(),
+        changes_before
+    );
+    assert_eq!(
+        test.subnet_available_memory().get_execution_memory(),
+        rename_bytes - 1
+    );
+
+    // The migration canister received a reject caused by the subnet being
+    // oversubscribed on memory.
```
mraszyk, line 7495: The migration canister will be hardened to gracefully handle this in a follow-up PR.

rs/execution_environment/src/scheduler/tests/charging.rs
```diff
@@ -279,6 +279,66 @@ fn canisters_with_insufficient_cycles_are_uninstalled() {
     );
 }
 
+// Uninstalling a canister because it ran out of cycles succeeds even when the
+// subnet has no available execution memory: the out-of-cycles path drops the
+// canister history (it does not record a `CanisterCodeUninstall` change), so it
+// never needs to decrement the subnet available execution memory. This is in
+// contrast to `uninstall_code`, which records a canister history change and fails
+// if the subnet cannot account for it.
```
schneiderstefan, line 287: Just to double check, if the uninstall clears more memory than the history entry takes (which should be the normal case), then the uninstall will succeed, even if old_memory + history entry would be over the limit?

schneiderstefan (approved):

rs/execution_environment/src/scheduler/tests/charging.rs
mraszyk: Indeed, `uninstall_code` snapshots `old_memory_usage` before dropping the execution state, records the `CanisterCodeUninstall` history entry, and only then reads `new_memory_usage`. So the history entry is folded into a single net figure — there is never an intermediate `old_memory + history_entry` quantity that gets checked.

PR #10837, opened 2026-07-20
chore(deps): remove unused ansible dependency

tmu0:
## What

Removes the `ansible` dependency from `requirements.in` and regenerates `requirements.txt`.

## Why

`ansible` has no first-party consumer in this repo — no Python imports, no playbooks/roles, no scripts invoking `ansible-playbook`, and no Bazel target depends on it. The only remaining references are stale comments/docs. It was pulled into the dependency-scanning venv purely because it was declared here.

## Effect

Dropping it also removes its exclusive transitive deps: `ansible-core`, `jinja2`, `markupsafe`, `resolvelib` (none of which are used first-party either). Net diff is a clean deletion (88 lines out of `requirements.txt`, nothing added).

This supersedes the earlier `ansible-core` version bump on this branch.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,4 +1,5 @@
 ansible>=12.2.0
+ansible-core>=2.20.7
 cbor>=1.0.0
 cbor2>=5.9.0
 configargparse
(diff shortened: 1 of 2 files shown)


basvandijk (approved):

basvandijk:
BTW @tmu0 do we still use Ansible for something?

tmu0:
@basvandijk good point, claude doesn't find any usage. let me remove it and see if something breaks...

PR #10840, opened 2026-07-20
feat: Consult the standard engine replica version.

daniel-wong-dfinity-org-twin:
When determining what replica version an engine is supposed to run (and it doesn't explicitly specify one).

# Future Work

Orchestrator does not use this. That will be fixed in an imminent PR.

# References

[prev]: https://github.com/dfinity/ic/pull/10787
[next]: https://github.com/dfinity/ic/pull/10851

### DIFF
--- a/rs/registry/helpers/src/subnet.rs
+++ b/rs/registry/helpers/src/subnet.rs
@@ -622,7 +622,7 @@ fn engine_upgrade_priority(subnet_id: SubnetId, new_replica_version_id: &str) ->
     hasher.write(subnet_id.get().to_string().as_bytes());
     let digest = hasher.finish();
 
-    let first_8_bytes: [u8; 8] = digest[0..8].try_into().unwrap();
+    let first_8_bytes = <[u8; 8]>::try_from(&digest[0..8]).unwrap();
     let priority_int = u64::from_le_bytes(first_8_bytes);
 
     priority_int as f64 / u64::MAX as f64
--- a/rs/registry/helpers/src/subnet.rs
+++ b/rs/registry/helpers/src/subnet.rs
@@ -620,7 +620,7 @@ fn get_standard_engine_replica_version_record<T: RegistryClient + ?Sized>(
 fn engine_upgrade_priority(subnet_id: SubnetId, new_replica_version_id: &str) -> f64 {
     let mut hasher = Sha256::new_with_context(&DomainSeparationContext::new("upgrade priority"));
     hasher.write(new_replica_version_id.as_bytes());
-    hasher.write(subnet_id.get().to_string().as_bytes());
+    hasher.write(subnet_id.to_string().as_bytes());
     let digest = hasher.finish();
 
     let first_8_bytes = <[u8; 8]>::try_from(&digest[0..8]).unwrap();
--- a/rs/registry/helpers/src/subnet.rs
+++ b/rs/registry/helpers/src/subnet.rs
@@ -1,4 +1,5 @@
 use crate::deserialize_registry_value;
+use ic_crypto_sha2::{DomainSeparationContext, Sha256};
 use ic_interfaces_registry::{
     RegistryClient, RegistryClientResult, RegistryClientVersionedResult, RegistryVersionedRecord,
 };
@@ -7,13 +8,15 @@ use ic_protobuf::{
     registry::{
         node::v1::NodeRecord,
         replica_version::v1::ReplicaVersionRecord,
+        standard_engine_replica_version::v1::StandardEngineReplicaVersionRecord,
         subnet::v1::{CatchUpPackageContents, SubnetListRecord, SubnetRecord, SubnetType},
     },
     types::v1::SubnetId as SubnetIdProto,
 };
 use ic_registry_keys::{
     DEFAULT_INITIAL_DKG_SUBNET_ID_KEY, ROOT_SUBNET_ID_KEY, make_catch_up_package_contents_key,
-    make_node_record_key, make_replica_version_key, make_subnet_list_record_key,
+    make_node_record_key, make_replica_version_key,
+    make_standard_engine_replica_version_record_key, make_subnet_list_record_key,
     make_subnet_record_key,
 };
 use ic_registry_subnet_features::{ChainKeyConfig, SubnetFeatures};
@@ -155,8 +158,26 @@ pub trait SubnetRegistry {
         version: RegistryVersion,
     ) -> RegistryClientResult<bool>;
 
-    /// Return the [ReplicaVersion] as recorded in the subnet record
-    /// at the given height.
+    /// Except for CloudEngine subnets, this just returns the value in the
+    /// [SubnetRecord]'s `replica_version_id` field, which is generally the git
+    /// commit ID from which Replica was built.
+    ///
+    /// But when the subnet is of type CloudEngine, `replica_version_id` can be
+    /// blank, which means that the engine is following
+    /// [StandardEngineReplicaVersionRecord]. Non-blank `replica_version_id`
+    /// means the same thing as in the non-CloudEngine case.
+    ///
+    /// Err is returned in various cases, but call out a couple in particular,
+    /// because the error type is unintuitive: DecodeError is returned in the
+    /// following cases:
+    ///
+    /// 1. CloudEngine with blank replica_version_id, but no
+    ///    StandardEngineReplicaVersionRecord.
+    ///
+    /// 2. Non-CloudEngine with blank replica_version_id.
+    ///
+    /// In practice, such data problems are prevented from happening elsewhere,
+    /// but we mention them here for completeness.
     fn get_replica_version(
         &self,
         subnet_id: SubnetId,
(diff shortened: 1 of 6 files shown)


rs/registry/helpers/src/subnet.rs
```diff
@@ -531,6 +598,36 @@ impl<T: RegistryClient + ?Sized> SubnetRegistry for T {
...
+/// the standard new replica version (otherwise, it takes the old one).
+///
+/// Based on 2 things (their text/display representations, for consistency):
+///
+/// 1. the engine's ID
+/// 2. the new replica version
+fn engine_upgrade_priority(subnet_id: SubnetId, new_replica_version_id: &str) -> f64 {
+    let mut hasher = Sha256::new_with_context(&DomainSeparationContext::new("upgrade priority"));
+    hasher.write(new_replica_version_id.as_bytes());
+    hasher.write(subnet_id.get().to_string().as_bytes());
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: Premature optimization, so no.

rs/registry/helpers/src/subnet.rs
```diff
@@ -531,6 +598,36 @@ impl<T: RegistryClient + ?Sized> SubnetRegistry for T {
...
+    let digest = hasher.finish();
+
+    let first_8_bytes: [u8; 8] = digest[0..8].try_into().unwrap();
+    let priority_int = u64::from_le_bytes(first_8_bytes);
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: Claude says copy_from_slice also panics, so this just sweeps that problem under the rug.

rs/registry/helpers/src/subnet.rs
```diff
@@ -155,8 +158,26 @@ pub trait SubnetRegistry {
...
+    /// Err is returned in various cases, but call out a couple in particular,
+    /// because the error type is unintuitive: DecodeError is returned in the
+    /// following cases:
+    ///
+    /// 1. CloudEngine with blank replica_version_id, but no
+    ///    StandardEngineReplicaVersionRecord.
+    ///
+    /// 2. Non-CloudEngine with blank replica_version_id.
+    ///
+    /// In practice, such data problems are prevented from happening elsewhere,
```
eichhorl: Does "elsewhere" refer to a registry invariant?

rs/registry/helpers/src/subnet.rs
```diff
@@ -531,6 +598,36 @@ impl<T: RegistryClient + ?Sized> SubnetRegistry for T {
...
+    let mut hasher = Sha256::new_with_context(&DomainSeparationContext::new("upgrade priority"));
+    hasher.write(new_replica_version_id.as_bytes());
+    hasher.write(subnet_id.get().to_string().as_bytes());
+    let digest = hasher.finish();
+
+    let first_8_bytes = <[u8; 8]>::try_from(&digest[0..8]).unwrap();
+    let priority_int = u64::from_le_bytes(first_8_bytes);
+
+    priority_int as f64 / u64::MAX as f64
+}
```
eichhorl, line 630: This seems to be slightly different than the "recipe" introduced by one of the previous PRs:

https://github.com/dfinity/ic/blob/cc1a281bdafdd25147c5c19b37fbf851301db0bb/rs/protobuf/def/registry/standard_engine_replica_version/v1/standard_engine_replica_version.proto#L19-L35

According to the proto, 
1. we should hash the hex decoded replica version whereas here, we hash the ASCII bytes
2. we should hash the raw subnet ID bytes, but here we hash the bytes of the string representation

eichhorl (approved):

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: Yes. Added clarification.

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: Ah. Yes. Sorry. I changed my mind, but forgot to update the .proto file. Nice catch! I have fixed the .proto file.

eichhorl:
The branch name contains `update-replica_version_id-readers`. Another reader (that doesn't use this "helper" function) is the orchestrator. Did you plan to update it as part of this PR or do so in a follow up?

rs/registry/helpers/src/subnet.rs
```diff
@@ -531,6 +599,36 @@ impl<T: RegistryClient + ?Sized> SubnetRegistry for T {
...
+/// the standard new replica version (otherwise, it takes the old one).
+///
+/// Based on 2 things (their text/display representations, for consistency):
+///
+/// 1. the engine's ID
+/// 2. the new replica version
+fn engine_upgrade_priority(subnet_id: SubnetId, new_replica_version_id: &str) -> f64 {
+    let mut hasher = Sha256::new_with_context(&DomainSeparationContext::new("upgrade priority"));
+    hasher.write(new_replica_version_id.as_bytes());
+    hasher.write(subnet_id.get().to_string().as_bytes());
```
eichhorl: I think this should be the same
```suggestion
    hasher.write(subnet_id.to_string().as_bytes());
```

eichhorl (approved):

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: Done.

daniel-wong-dfinity-org-twin:
> The branch name contains `update-replica_version_id-readers`. Another reader (that doesn't use this "helper" function) is the orchestrator. Did you plan to update it as part of this PR or do so in a follow up?

Yes, I'm working on a follow up now. Hope to send it out today.

bjoernek:
Could you add a line to the description noting the remaining steps in this stack (e.g. the orchestrator reader follow-up)? This PR is clear on its own, but a quick "what's still coming" would save future readers from walking the whole prev/next chain to see where it sits. Non-blocking.

bjoernek (approved):

rs/registry/helpers/src/subnet.rs
```diff
@@ -726,6 +825,175 @@ mod tests {
...
+            high_priority_result,
+            Some(ReplicaVersion::try_from("old").unwrap())
+        );
+        assert_eq!(
+            override_result,
+            Some(ReplicaVersion::try_from("override").unwrap())
+        );
+    }
+
+    // This wouldn't occur in practice, so this test is "just" for completeness.
```
pierugo-dfinity, line 940: > This wouldn't occur in practice

In the current state, I do not think there is anything preventing this from happening, except due diligence when reviewing proposals. I think it would be a good idea to add an invariant that ensures that a `StandardEngineReplicaVersionRecord` MUST exist if a `SubnetRecord` has an empty replica version. Otherwise, replicas would not know what version to run.

@daniel-wong-dfinity-org-twin PTAL

rs/registry/helpers/src/subnet.rs
```diff
@@ -726,6 +825,175 @@ mod tests {
...
+        registry.update_to_latest_version();
+
+        // Step 2: Run the code under test.
+        let result = registry.get_replica_version(subnet_id(0xBABE), RegistryVersion::from(2));
+
+        // Step 3: Verify result(s).
+        assert_matches!(result, Err(RegistryClientError::DecodeError { .. }));
+    }
+
+    // This wouldn't occur in practice, so this test is "just" for completeness.
```
pierugo-dfinity, line 967: > This wouldn't occur in practice

Same here, I think it would be worth adding an invariant for this.

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: iiuc, currently, blank is not allowed. Ofc, we will relax that. But there is a middle path: engine can have blank replica_version_id if there is a (valid) StandardEngineReplicaVersionRecord. That's being implemented now.

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: I think this already exists. Or at least, it is already required that there are no "dangling references" (i.e. if someone says they are using an elected replica version, then the replica version must actually be elected). I would consider that to be good enough already. Ofc, locking it down more is good. Belt and suspenders. To really lock it down, we might want to require (if it isn't already) that an elected replica version must have an ID of length > 0. We can lock it down even more by requiring that it must have the "shape" of a (full) git commit ID (i.e. 40 char, hexidecimal (and maybe even lower case)).

rs/registry/helpers/src/subnet.rs
pierugo-dfinity: > That's being implemented now.

Ok great to hear!

rs/registry/helpers/src/subnet.rs
pierugo-dfinity: Sorry my comment was referring to the test `blank_replica_version_id_on_a_non_engine_subnet_is_an_error`. What I meant is that it would be useful to have an invariant that ensures that a blank replica version is possible only on cloud engines, and not on regular subnets. But I assume this will be part of the changes you are currently implementing.

pierugo-dfinity: > I assume this will be part of the changes you are currently implementing

It is, indeed. All good for me.

PR #10842, opened 2026-07-21
fix: avoid double counting deleted canisters in SubnetMetrics::consumed_cycles_total

mraszyk:
When a canister is deleted, its per-use-case consumption is added both to the `consumed_cycles_by_deleted_canisters` scalar and to the `consumed_cycles_by_use_case` map. `SubnetMetrics::consumed_cycles_total()` summed both, double counting the cycles consumed by deleted canisters.

Fix `consumed_cycles_total()` to skip the canister-level use cases (which only ever enter the map via deleted canisters and are already covered by the deleted canisters scalar), summing only the subnet-level use cases (SchnorrOutcalls, VetKd, DroppedMessages) that are not covered by any scalar.

Since this changes the certified state, the legacy behavior is preserved in `consumed_cycles_total_v27()`, which the canonical state consumer keeps using for all certification versions up to and including V27. A future certification version can switch the consumer over to the fixed `consumed_cycles_total()`.

### DIFF
--- a/rs/replicated_state/src/metadata_state/tests.rs
+++ b/rs/replicated_state/src/metadata_state/tests.rs
@@ -2386,44 +2386,72 @@ fn stream_responses_tracking() {
 
 #[test]
 fn consumed_cycles_total_calculates_the_right_amount() {
+    // Each entry gets a distinct power of two so that the resulting total is a
+    // bitmask: any use case that is miscategorized (added when it should be
+    // skipped, or vice versa) changes the total by a unique amount that cannot
+    // be masked by other entries cancelling out.
     let mut consumed_cycles_by_use_case = BTreeMap::new();
-    // Use cases already covered by the scalar metrics below; these must not be
+    // Use cases covered by a dedicated scalar metric below; these must not be
     // added to the total again (otherwise the cycles consumed by deleted
-    // canisters would be double counted).
-    consumed_cycles_by_use_case.insert(CyclesUseCase::DeletedCanisters, NominalCycles::new(5));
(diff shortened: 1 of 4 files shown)


rs/replicated_state/src/metadata_state/tests.rs
```diff
@@ -2387,12 +2387,20 @@ fn stream_responses_tracking() {
...
     consumed_cycles_by_use_case.insert(CyclesUseCase::HTTPOutcalls, NominalCycles::new(12));
     consumed_cycles_by_use_case.insert(CyclesUseCase::ECDSAOutcalls, NominalCycles::new(30));
     consumed_cycles_by_use_case.insert(CyclesUseCase::Instructions, NominalCycles::new(100));
     consumed_cycles_by_use_case.insert(CyclesUseCase::Memory, NominalCycles::new(50));
     consumed_cycles_by_use_case.insert(CyclesUseCase::CanisterCreation, NominalCycles::new(40));
+    // Subnet-level use cases not covered by any scalar metric; these are added
+    // to the total.
+    consumed_cycles_by_use_case.insert(CyclesUseCase::SchnorrOutcalls, NominalCycles::new(7));
+    consumed_cycles_by_use_case.insert(CyclesUseCase::VetKd, NominalCycles::new(8));
+    consumed_cycles_by_use_case.insert(CyclesUseCase::DroppedMessages, NominalCycles::new(9));
```
eichhorl: There are a couple more use cases that aren't part of this test, should we populate all of them?

eichhorl (approved):
Should we add debug_asserts here:
https://github.com/dfinity/ic/blob/c63afc36cbe3fae931450cfda91ec3498c528bb9/rs/replicated_state/src/canister_state/system_state.rs#L2139-L2144

PR #10844, opened 2026-07-21
feat: Migrate HTTP/ECDSA outcall cycles in SubnetMetrics from scaler to per-use-case map

mraszyk:
The cycles consumed by HTTP and ECDSA outcalls are tracked both in the legacy scalar fields consumed_cycles_http_outcalls / consumed_cycles_ecdsa_outcalls of SubnetMetrics and in the
  consumed_cycles_by_use_case map. The scalar fields predate use-case tracking, so they are a superset of the corresponding use-case entries.

  This adds a migration that brings the HTTPOutcalls / ECDSAOutcalls entries of consumed_cycles_by_use_case up to the scalar value (via max).

  The migration runs whenever any use case is observed (from observe_consumed_cycles_with_use_case), i.e. independently of whether the scalar fields themselves are observed, so the use-case
  entries also catch up on subnets that stop performing outcalls. It runs after the increments; since the scalar fields are bumped before the matching use-case observation at the call sites,
  the max reconciliation keeps the entries in lockstep with the scalar fields without double counting.
  
  Only the consumed_cycles_by_use_case map is migrated. The monotonic consumed_cycles_by_use_case_as_counters map is intentionally left untouched: backfilling it would introduce a spurious
  counter jump (the counters were introduced very recently, unlike the gauges).

  The scalar fields are intentionally kept (and kept up to date) rather than zeroed, so they remain the source of truth for the readers that still use them (e.g. consumed_cycles_total) and
  so that downgrading to an earlier replica version keeps observing the correct totals. Read paths are left unchanged for now.

### DIFF
--- a/rs/replicated_state/src/metadata_state.rs
+++ b/rs/replicated_state/src/metadata_state.rs
@@ -480,6 +480,7 @@ impl SubnetMetrics {
     }
 
     pub fn observe_consumed_cycles_http_outcalls(&mut self, cycles: NominalCycles) {
+        self.migrate_outcalls_cycles_to_use_cases();
         self.consumed_cycles_http_outcalls += cycles;
     }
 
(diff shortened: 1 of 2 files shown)


eichhorl:
The description claims that `consumed_cycles_by_use_case_as_counters` is left untouched because backfilling it would introduce a spurious counter jump. 

However, it seems that `consumed_cycles_by_use_case_as_counters` (`get_consumed_cycles_by_use_case_as_counters()`) is never actually read outside of tests, and the metric `consumed_cycles_by_use_case_as_counters` is also filled using `consumed_cycles_by_use_case`:
https://github.com/dfinity/ic/blob/31862028e4fcaf83dc97af94633fd77e2d799423/rs/replicated_state/src/metrics.rs#L518-L524

So this migration would already lead to a counter jump. 

Maybe the problem is that we should call `get_consumed_cycles_by_use_case_as_counters()` in the code above? But then the metric might actually decrease, since the counters were introduced after the gauges.

mraszyk:
> it seems that `consumed_cycles_by_use_case_as_counters` (`get_consumed_cycles_by_use_case_as_counters()`) is never actually read outside of tests

Good point! I'll consider cleaning that up in a follow-up.

> the metric `consumed_cycles_by_use_case_as_counters` is also filled using `consumed_cycles_by_use_case`:
> 
> https://github.com/dfinity/ic/blob/31862028e4fcaf83dc97af94633fd77e2d799423/rs/replicated_state/src/metrics.rs#L518-L524
> 
> So this migration would already lead to a counter jump.

correct, but the jump would only affect the exported metric, not data (the map `consumed_cycles_by_use_case_as_counters`) stored in the checkpoint (that I wanted to keep consistent as tracking since the "counters" were introduced, not since the total or gauges were introduced)

> Maybe the problem is that we should call `get_consumed_cycles_by_use_case_as_counters()` in the code above?

The gauges actually already behave as counters at the moment which is likely why they are used to feed the "counter" metric.

eichhorl:
> I'll consider cleaning that up in a follow-up.

Fine with me 👍 

> The gauges actually already behave as counters at the moment which is likely why they are used to feed the "counter" metric.

Then I'm not sure why the `_as_counters` map was introduced

eichhorl (approved):

mraszyk:
> Then I'm not sure why the `_as_counters` map was introduced

The only reason I can think of is to have them be tracked since the same point in time when the canister-level "counters" were introduced (canister-level "gauges" are really gauges which is why canister-level "counters" were introduced, i.e., they are also decremented unlike subnet-level "gauges" that are never decremented).

PR #10846, opened 2026-07-21
feat: Add canister creation timestamp to state tree

schneiderstefan:
Like the recently added last install timestamp, this adds the canister creation timestamp to the state tree.

This PR only introduces the nex certification version. In a follow-up, we will bump the current version.


rs/execution_environment/src/canister_manager/tests.rs
```diff
@@ -5158,6 +5158,38 @@ fn last_install_timestamp_tracks_code_deployment_and_uninstall() {
     assert_eq!(last_install_timestamp(&test), None);
 }
 
+#[test]
+fn canister_creation_timestamp_is_set_at_creation_and_stable() {
+    let mut test = ExecutionTestBuilder::new().build();
+
+    let creation_time = test.state().time();
```
mraszyk, line 5168: We could bump time beforehand so that the creation timestamp is not the default timestamp.

mraszyk (approved):

rs/execution_environment/src/canister_manager/tests.rs
schneiderstefan: Done.

PR #10848, opened 2026-07-21
chore: replace deprecated backoff with backon

nmattia:
The `backoff` crate is deprecated and unmaintained:
* https://rustsec.org/advisories/RUSTSEC-2025-0012.html
* https://github.com/ihrwein/backoff/issues/66

This replaces all local uses of `backoff` with `backon`. The API is a bit different, though this tries to keep the same configuration/behavior as we had with `backoff`.

The `backoff` crate is still pulled transitively via agent-rs. The latest version (ic-agent v0.49.1) will allow us to drop the dependency completely but the upgrade will be done separately.


rs/orchestrator/src/orchestrator.rs
```diff
@@ -534,17 +534,22 @@ impl Orchestrator {
             // e.g. (roughly) 1, 1.75, 3, 5.25, 9.5, 16.5, 28.75, 50.25, 88, 120, 120
             //
             // Additionally, there's a random +=50% range added to each delay, for jitter.
```
mraszyk: I think this comment about +=50% is outdated

mraszyk (approved):

basvandijk (approved):

daniel-wong-dfinity-org-twin (approved):

lwshang (approved):

blind-oracle (approved):

rs/orchestrator/src/orchestrator.rs
nmattia: good catch, here's what `backon` actually uses:

>     /// When jitter is enabled, [`ExponentialBackoff`] will add a random jitter within `(0, current_delay)`
>    /// to the current delay.

PR #10859, opened 2026-07-22
fix: cycles accounting for mgmt canister operations

mraszyk:
The management canister does not refund cycles for its charged instructions (the only exceptions being install_code and load_canister_snapshot), but counters are only set upon refund - fix the management canister to always refund zero cycles immediately after its charges by using a new function `CyclesAccountManager::consume_cycles_for_final_instructions` whenever there's no explicit refund. (The function `CyclesAccountManager::consume_cycles_for_instructions` was unused so it is dropped.)

Moreover, do not decompose CompoundCycles into real cycles and back to CompoundCycles which leads to not accounting for the nominal cycles properly on free subnets.


rs/cycles_account_manager/src/cycles_account_manager.rs
```diff
@@ -381,28 +381,35 @@ impl CyclesAccountManager {
...
+        cycles: CompoundCycles<Instructions>,
         subnet_cycles_config: CyclesAccountManagerSubnetConfig,
-        execution_mode: WasmExecutionMode,
+        reveal_top_up: bool,
     ) -> Result<(), CanisterOutOfCyclesError> {
-        let memory_usage = canister.memory_usage();
-        let message_memory = canister.message_memory_usage();
-        let cycles = self.execution_cost(amount, subnet_cycles_config, execution_mode);
-        let reveal_top_up = canister.controllers().contains(sender);
         self.consume_cycles(
```
schneiderstefan, line 401: Since `consume_cycles` is still public, it's possible that we reintroduce this bug in the future. Not sure what we can do about it exactly, but maybe a comment on it could help.

rs/execution_environment/src/execution_environment/tests/canister_snapshots.rs
```diff
@@ -1858,6 +1859,73 @@ fn take_canister_snapshot_charges_canister_cycles() {
     );
 }
 
+// On a free subnet the real instruction charge for `take_canister_snapshot` is
+// zero, but its nominal amount must still be recorded on both the
+// consumed-cycles gauge and the monotonic counter. Regression test that the
+// nominal amount is not dropped along the way.
+#[test]
+fn take_canister_snapshot_records_nominal_cycles_on_free_subnet() {
```
schneiderstefan, line 1867: Would it make sense to have a similar test for `load_canister_snapshot` (or reload the snapshot in this test)?

schneiderstefan (approved):

rs/cycles_account_manager/src/cycles_account_manager.rs
mraszyk: I'd propose to do the following in a follow-up: introduce `CyclesUseCaseNonRefundableKind` and restrict the public methods `consume_with_threshold<T: CyclesUseCaseNonRefundableKind>` and `consume_cycles<T: CyclesUseCaseNonRefundableKind>`. I took a note to do that.

rs/execution_environment/src/execution_environment/tests/canister_snapshots.rs
mraszyk: I'm not sure it's worthwhile since there are other snapshot operations and also `update_settings` resizing log memory buffer and `upload_chunk` (to WASM chunk store) that charge mgmt canister instructions (and having a test for each of them would be an overkill). Moreover, `load_canister_snapshot` also charges instructions for execution (via the prepay + refund paradigm) so such a test could not easily assert the precise consumed cycles which limits its coverage.

PR #10860, opened 2026-07-22
fix: Validate summary blocks while halted

eichhorl:
While halting, data blocks are empty so their content does not need to be validated explicitly. However, summary blocks may still contain some data that should be validated for consistency, even if the subnet is already halted.

### DIFF
--- a/rs/consensus/src/consensus/validator.rs
+++ b/rs/consensus/src/consensus/validator.rs
@@ -1274,9 +1274,10 @@ impl Validator {
             .into());
         }
 
-        // If the replica is halted, the block payload is empty so we can skip the rest of the
+        // If the replica is halted, data block payloads are empty so we can skip the rest of the
         // validation.
-        if status == Status::Halting || status == Status::Halted {
+        if !proposal.payload.is_summary() && (status == Status::Halting || status == Status::Halted)
+        {
             return Ok(());
         }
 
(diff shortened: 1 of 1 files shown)


rs/consensus/src/consensus/validator.rs
```diff
@@ -1274,9 +1274,10 @@ impl Validator {
...
         }
 
-        // If the replica is halted, the block payload is empty so we can skip the rest of the
+        // If the replica is halted, data block payloads are empty so we can skip the rest of the
         // validation.
-        if status == Status::Halting || status == Status::Halted {
+        if !proposal.payload.is_summary() && (status == Status::Halting || status == Status::Halted)
+        {
             return Ok(());
         }
```
pierugo-dfinity: Could we combine this and the check on lines 1214-1219 with something like the following?

```suggestion
        if status == Status::Halting || status == Status::Halted {
            let payload = proposal.payload.as_ref();
            if !payload.is_summary() {
                // If the replica is halted, data block payloads should be empty, and we can skip the rest of the validation if they are.
                return if !payload.is_empty() {
                    Err(InvalidArtifactReason::NonEmptyPayloadPastUpgradePoint.into());
                } else {
                    Ok(())
                };
            }
        }
```

rs/consensus/src/consensus/validator.rs
```diff
@@ -2805,6 +2806,84 @@ pub mod test {
...
+                .expect_latest_certified_height()
+                .return_const(context.certified_height);
+            time_source.set_time(context.time).unwrap();
+
+            // The block must be validated, so the failing payload validation is hit.
+            let result = validator.check_block_validity(&PoolReader::new(&pool), &summary_proposal);
+            assert_matches!(
+                result,
+                Err(ValidationError::InvalidArtifact(
+                    InvalidArtifactReason::InvalidPayload(_)
```
pierugo-dfinity: Maybe
```suggestion
                    InvalidArtifactReason::InvalidPayload(InvalidPayloadReason::PayloadTooBig { .. })
```
to be extra explicit?

pierugo-dfinity (approved):
Wow, very good catch

rs/consensus/src/consensus/validator.rs
eichhorl: Good idea, done: https://github.com/dfinity/ic/pull/10860/changes/3675114757cf6fc95efe7467075370abc53bc6de

rs/consensus/src/consensus/validator.rs
eichhorl: Thanks, done: https://github.com/dfinity/ic/pull/10860/changes/3675114757cf6fc95efe7467075370abc53bc6de

PR #10862, opened 2026-07-22
fix(nns_delegation_manager): use custom DNS resolver in tests

basvandijk:
## Problem

`load_root_delegation_on_cloud_engine_should_contact_api_bn_test` fails under sandboxed remote execution (e.g. namespaced Bazel RBE) with:

```
assertion failed: `Err(Io(Custom { kind: Other, error: "no nameservers found in config" }))`
    does not match `Err(err) if format!("{err:?}").contains("Could not connect to node [::1]:443")`
```

The `CloudEngine` branch of `try_fetch_delegation_from_nns` builds a DNS resolver via `Resolver::builder_tokio()`, which reads `/etc/resolv.conf` **at construction time**. In an environment without nameservers (a sandbox), that call fails with `"no nameservers found in config"` before the code ever reaches the connect step. The test's API BN "domain" is the literal loopback address `::1`, so no DNS is actually needed — but the resolver is built regardless, and its construction reads the (empty) system config.

In non-sandboxed environments (with a populated `resolv.conf`) the resolver builds fine, the literal `::1` short-circuits, and the connection is refused, yielding the expected `"Could not connect to node [::1]:443"`.

## Fix

Avoid reading the system `resolv.conf` in test builds:

- In **production** (`!cfg!(test)`), the resolver is built from the system's default nameservers exactly as before, via `Resolver::builder(TokioRuntimeProvider::default())` (equivalent to the previous `Resolver::builder_tokio()`).
- In **test** builds (`cfg!(test)`), the resolver is built from an explicit `ResolverConfig` containing a single placeholder nameserver (`::1`) instead of reading `/etc/resolv.conf`. This placeholder is never actually contacted, because the domain used by the test is an IP literal (`::1`) which resolves without issuing a query.

This keeps the test hermetic under sandboxed remote execution — it still exercises the full DNS-resolution code path and reaches the connect step, failing with the expected `"Could not connect to node [::1]:443"` — while leaving production DNS behavior unchanged.

## Testing

`bazel test //rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test` — PASSED (15/15 tests).

### DIFF
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -1,9 +1,4 @@
-use std::{
-    convert::TryFrom,
-    net::{IpAddr, SocketAddr},
-    sync::Arc,
-    time::Duration,
-};
+use std::{convert::TryFrom, net::SocketAddr, sync::Arc, time::Duration};
 
 use axum::body::Body;
 use futures::FutureExt;
--- a/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
+++ b/rs/http_endpoints/nns_delegation_manager/src/nns_delegation_manager.rs
@@ -1,4 +1,9 @@
-use std::{convert::TryFrom, net::SocketAddr, sync::Arc, time::Duration};
+use std::{
+    convert::TryFrom,
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
 
 use axum::body::Body;
 use futures::FutureExt;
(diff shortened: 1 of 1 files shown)


pierugo-dfinity:
I'm not a big fan of changing the functionality also for production code. Because the problem is that sandboxed environments do not have nameservers, this [alternative](https://github.com/dfinity/ic/pull/10862/commits/7b90228a995e25d1a418d50b307b4dccdc2c2081) passes a dummy nameserver in test-only code. WDYT? The test passes locally, but also on `master`. Do you know how could I test my change?

basvandijk:
@pierugo-dfinity I'll cherry-pick this into https://github.com/dfinity/ic/pull/10579 to see if it works on RBE. It probably will. Thanks. Could you (or your Claude) adapt the PR description accordingly?

basvandijk:
@pierugo-dfinity testing this now in https://github.com/dfinity/ic/actions/runs/29945328053/job/89009140401?pr=10579 🤞

basvandijk:
@pierugo-dfinity it [passed](https://github.com/dfinity/ic/actions/runs/29945328053/job/89009140401?pr=10579#step:9:1198):
```
//rs/http_endpoints/nns_delegation_manager:nns_delegation_manager_test   PASSED in 12.5s
```

basvandijk:
I also adapted the PR description.

pierugo-dfinity (approved):

PR #10863, opened 2026-07-22
fix(system-tests): keep local backend dnsmasq non-root under RBE sandbox

basvandijk:
## Problem

Under Namespace's RBE sandbox (`--remote_default_exec_properties=namespace_action_isolation=sandboxed`) the `LocalBackend::create_group` step fails while starting the group's `dnsmasq` RA daemon:

```
dnsmasq: failed to change group-id to dip: Operation not permitted
```

The local backend relies on `dnsmasq` staying **unprivileged** so it skips its privilege-drop path. That path is gated purely on `getuid() == 0`, and when taken it fails inside the driver's nested user namespace: `setgroups` is denied (mandatory before an unprivileged `gid_map` write) and the default `dip` gid is unmapped.

This invariant held because `ensure_administrable_netns` identity-maps the caller's uid and the action historically ran as an ordinary (non-root) user. The Namespace sandbox runs actions as **uid 0** in its own user namespace, so the identity map makes the driver root inside its namespace too — flipping `dnsmasq` into the drop path.

## Fix

Map a (fake-)root caller to a **non-zero** inner uid/gid instead of identity-mapping `0 → 0`, so `getuid()` is never `0` inside the driver's namespace and `dnsmasq` (and any other root-sensitive child) skips its privilege-drop path — restoring the invariant the code already documents. The map's *outer* id is still the caller's real uid, so on-disk ownership, `/dev/kvm` and `/dev/net/tun` access are unchanged. A single-line self-map is always permitted, so it is robust to whatever uid the RBE backend uses.

No fix is required on the Namespace side; running sandboxed actions as fake-root is a legitimate sandbox design and this change makes the backend robust to it.

## Testing

This has been successfully tested on https://github.com/dfinity/ic/pull/10579.

### DIFF
--- a/rs/tests/driver/src/driver/local_backend.rs
+++ b/rs/tests/driver/src/driver/local_backend.rs
@@ -238,18 +238,36 @@ impl LocalBackend {
                  networking without host capabilities",
             );
         }
-        // Identity-map the caller's uid/gid into the new user namespace, so it
-        // keeps its usual identity (files, `/dev/kvm` and `/dev/net/tun` are
-        // opened exactly as before) while being the namespace owner. `setgroups`
-        // must be denied before an unprivileged process may write `gid_map`; that
-        // is fine here because nothing in the process tree calls `setgroups`
-        // (`dnsmasq`, the only program that would, stays unprivileged and so
-        // skips its privilege-drop entirely — see `start_ra_daemon`).
+        // Map the caller's uid/gid into the new user namespace so it keeps its
+        // usual identity (files, `/dev/kvm` and `/dev/net/tun` are opened exactly
+        // as before) while being the namespace owner. The mapping is an identity
+        // one, with a single exception: if the caller is (fake-)root we map it to
+        // a *non-zero* inner uid/gid instead of `0`.
+        //
+        // The reason: the backend relies on `dnsmasq` staying unprivileged so it
+        // skips its privilege-drop path (see `start_ra_daemon`). That path is
+        // gated purely on `getuid() == 0`, and when taken it fails in this
+        // namespace — `setgroups` is denied (below) and the default `dip` gid is
+        // unmapped. This normally holds because the action runs as an ordinary
+        // user, but under an RBE sandbox that runs actions as uid 0 in its own
+        // user namespace (e.g. Namespace's `namespace_action_isolation=sandboxed`)
+        // an identity map would make the driver root here too and trip that path.
+        // Presenting a non-zero uid keeps `dnsmasq` (and any other root-sensitive
+        // child) out of it regardless of the outer uid. On-disk ownership,
+        // `/dev/kvm` and `/dev/net/tun` are still accessed as the mapping's
+        // *outer* uid, so nothing else changes.
+        //
+        // `setgroups` must be denied before an unprivileged process may write
+        // `gid_map`; that is fine here because nothing in the process tree needs
+        // `setgroups` to succeed. A single-line self-map is always permitted, with
+        // or without `CAP_SETUID`/`CAP_SETGID` in the parent user namespace.
+        let inner_uid = if uid == 0 { 1 } else { uid };
+        let inner_gid = if gid == 0 { 1 } else { gid };
         std::fs::write("/proc/self/setgroups", "deny")
             .context("denying setgroups for the private user namespace")?;
-        std::fs::write("/proc/self/uid_map", format!("{uid} {uid} 1"))
+        std::fs::write("/proc/self/uid_map", format!("{inner_uid} {uid} 1"))
             .context("writing uid_map for the private user namespace")?;
-        std::fs::write("/proc/self/gid_map", format!("{gid} {gid} 1"))
+        std::fs::write("/proc/self/gid_map", format!("{inner_gid} {gid} 1"))
             .context("writing gid_map for the private user namespace")?;
         // Raise the networking capabilities into the ambient set so the
         // unprivileged `ip`/`dnsmasq`/QEMU children inherit them across `exec`.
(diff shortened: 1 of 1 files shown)


rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -238,18 +238,36 @@ impl LocalBackend {
...
+        // child) out of it regardless of the outer uid. On-disk ownership,
+        // `/dev/kvm` and `/dev/net/tun` are still accessed as the mapping's
+        // *outer* uid, so nothing else changes.
+        //
+        // `setgroups` must be denied before an unprivileged process may write
+        // `gid_map`; that is fine here because nothing in the process tree needs
+        // `setgroups` to succeed. A single-line self-map is always permitted, with
+        // or without `CAP_SETUID`/`CAP_SETGID` in the parent user namespace.
+        let inner_uid = if uid == 0 { 1 } else { uid };
+        let inner_gid = if gid == 0 { 1 } else { gid };
```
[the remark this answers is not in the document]
basvandijk: This isn't correct — `uid_map` *is* the mechanism that determines the uid inside the namespace. `getuid()` is defined as the process's kernel uid rendered through the current user namespace's `uid_map`, so writing `1 0 1` (parent uid 0 ↔ inner uid 1) makes a process that is uid 0 in the parent report `getuid() == 1` inside. It's the same mechanism `unshare -r` uses in reverse (an ordinary user writes `0 <uid> 1` to *become* root-in-namespace); we do the inverse.

Demonstrated by replicating exactly what the driver does (here the outer uid is 1000; in the sandboxed fake-root case it's 0 and the entry is `1 0 1` — identical mechanism):

```
BEFORE: getuid=1000 getgid=1002  CapEff:	0000000000000000
AFTER : getuid=1 getgid=1  CapEff:	000001ffffffffff
  id -> uid=1(daemon) gid=1(daemon) ...
  write to outer-owned file: OK (owner check passed)
  overflowuid would be 65534 (we are 1, so NOT overflow)
```

So `getuid()` returns 1 inside → dnsmasq's `getuid() == 0` guard is false → it skips the privilege-drop path. That's the fix working.

On the capabilities aside: writing `uid_map` is not a `setuid()` transition — the kernel uid is unchanged, only its rendering differs, so nothing clears caps. As the run above shows, `CapEff` is the full set *after* the map write (the namespace creator is granted full caps regardless of its in-namespace uid), and `raise_ambient_net_caps()` runs after the map write exactly as in the pre-existing identity-map path. dnsmasq/ip/qemu inherit the ambient `CAP_NET_*` across exec unchanged.

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -238,18 +238,36 @@ impl LocalBackend {
...
         std::fs::write("/proc/self/setgroups", "deny")
             .context("denying setgroups for the private user namespace")?;
-        std::fs::write("/proc/self/uid_map", format!("{uid} {uid} 1"))
+        std::fs::write("/proc/self/uid_map", format!("{inner_uid} {uid} 1"))
             .context("writing uid_map for the private user namespace")?;
-        std::fs::write("/proc/self/gid_map", format!("{gid} {gid} 1"))
+        std::fs::write("/proc/self/gid_map", format!("{inner_gid} {gid} 1"))
```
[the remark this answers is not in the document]
basvandijk: This prediction (credentials stay `0` inside → uid 0 becomes unmapped → overflowuid) doesn't hold: the process is not uid 0 inside after the map. The entry `1 0 1` maps the process's *own* outer uid (0) to inner uid 1, so the process becomes inner uid **1**, which is mapped — never overflowuid. It's inner uid 0 that is unmapped, but the process is no longer inner uid 0.

Empirically (replicating the driver's calls; outer uid 1000 here, `1 0 1` in the sandboxed case — same mechanism):

```
AFTER : getuid=1 getgid=1
  id -> uid=1(daemon) gid=1(daemon) ...
  write to outer-owned file: OK (owner check passed)
  overflowuid would be 65534 (we are 1, so NOT overflow)
```

Access checks are unaffected because inner uid 1 resolves to the mapping's *outer* uid (the real caller), so files, `/dev/kvm` and `/dev/net/tun` are accessed as before — the outer identity is stable, which is what the comment intends. The only thing that changes is that `getuid()` inside is non-zero, which is exactly what keeps dnsmasq out of its root-only privilege-drop path.

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -238,18 +238,36 @@ impl LocalBackend {
...
             );
         }
-        // Identity-map the caller's uid/gid into the new user namespace, so it
-        // keeps its usual identity (files, `/dev/kvm` and `/dev/net/tun` are
-        // opened exactly as before) while being the namespace owner. `setgroups`
-        // must be denied before an unprivileged process may write `gid_map`; that
-        // is fine here because nothing in the process tree calls `setgroups`
-        // (`dnsmasq`, the only program that would, stays unprivileged and so
-        // skips its privilege-drop entirely — see `start_ra_daemon`).
+        // Map the caller's uid/gid into the new user namespace so it keeps its
```
nmattia, line 241: This is very roundabout. I had to read this many times but I think the issue is simple: when dnsmasq is run as root (uid 0) it tries to drop its privileges:

> Dnsmasq must normally be started as root, but it will drop root privileges after startup by changing id to another user.

But the sandbox won't allow this, so we simply run it as non-root via id mapping. Is this correct?

nmattia: follow-up question: is it `setgroup` that's problematic, or is it the group we set it to? if it's the latter, can we supply a different group with `--group`?

basvandijk: Both the [`setgroups` and `setgid` in dnsmasq](https://github.com/imp/dnsmasq/blob/master/src/dnsmasq.c#L746-L747) are problematic. `setgroups` will be denied by the `std::fs::write("/proc/self/setgroups", "deny")` in `ensure_administrable_netns` (which is required by the kernel) and `setgid` will fail since the `dip` group is not mapped. We could use `--group=<some_group>` but then we still run into the `setgroups` deny issue. The other issue is what to use for `<some_group>`; we need to make sure to map that group into the user namespace as well. That's also extra complexity.

I researched this a bit further with Fable 5 and there are some alternatives:

1. --user=root --group=root + stop denying setgroups when the caller is root. 
2. no dnsmasq flag changes at all: richer gid/uid maps.
3. -d/--no-daemon

Fable recommends the 1st but I think it's too complex. I like 3 but it would require backgrounding `dnsmasq` and tracking its PID ourselves. Definitely doable but I feel we should go with the current solution for now since it's proven to work in all situations. Then we can revisit this later when the dust settles around RBE. 

WDYT?

nmattia (approved):

PR #10867, opened 2026-07-22
fix(system-tests): pin rent_subnet_test to the Farm backend

basvandijk:
## Why

`rent_subnet_test` boots a **36-VM** testnet on a single host — 32 App IC Node VMs (the canister-ID-range hack) + 1 System + 1 fast System + 1 unassigned — demanding **~51 vCPUs and ~328 GiB of RAM**.

That fits DFINITY's local-backend runners (**64 CPUs**, plenty of RAM) but massively overcommits a Namespace Bazel Execution Worker (**16 CPUs, 32 GiB**). Under Namespace remote execution (see #10579) the host thrashes and the awaited System-subnet node never becomes reachable, so `setup` times out after ~503s with:

```
await_status_is_healthy ... timed out after 503.2s on attempt 62.
Last error: error sending request ... No route to host (os error 113)
```

Unlike the sibling per-VM memory caps (e.g. `unstuck_subnet_test`), **no memory cap rescues this test**: 32 of the VMs are full replica VMs, and the declared 51 vCPUs alone exceed the worker's 16.

## What

Pin the test to the Farm backend, where each VM runs on its own host. This marks the `_local` variant `manual`, dropping it from the `//rs/tests/...` wildcard on Namespace RBE while the Farm variant keeps running as before.

### DIFF
--- a/rs/tests/nns/BUILD.bazel
+++ b/rs/tests/nns/BUILD.bazel
@@ -264,8 +264,8 @@ system_test(
     # + 1 unassigned), demanding ~51 vCPUs and ~328 GiB of RAM. That fits on
     # DFINITY's local-backend runners (64 CPUs, plenty of RAM) but not on a
     # Namespace Bazel Execution Worker (16 CPUs, 32 GiB), where the host thrashes
-    # and nodes never become reachable. Pin this test to the Farm backend, where
-    # each VM runs on its own host.
+    # and nodes never become reachable. Pin this test to the Farm backend, where VMs
+    # are provisioned by the Farm cluster instead of being booted on the Bazel worker.
     backend = "farm",
     cpus = MIN_LOCAL_CPUS + 32 * 1 + 3 * DEFAULT_VCPUS_PER_VM,  # 32 App IC Node VMs with 1 vCPU override + 1 System + 1 fast System + 1 unassigned IC Node VMs * 6 vCPUs.
     tags = [
--- a/rs/tests/nns/BUILD.bazel
+++ b/rs/tests/nns/BUILD.bazel
@@ -259,6 +259,14 @@ system_test(
 
 system_test(
     name = "rent_subnet_test",
+    # Too big for the "local" backend under Namespace remote execution: this test
+    # boots 36 VMs on a single host (32 App IC Node VMs + 1 System + 1 fast System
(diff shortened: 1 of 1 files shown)


daniel-wong-dfinity-org-twin:
Yikes! I did not realize this test required so much resources! The code that this test is trying to exercise does not INTRINSICALLY need that many subnets. Rather, an exorbitant amount of subnets are created as a hack to work around an issue that blocks the test. See [this comment]. If there is another way to resolve the issue without so many subnets, that would be better. I think I got this hack from Martin R. Perhaps, he knows of a better way now?

[this comment]: https://sourcegraph.com/r/github.com/dfinity/ic@54f381972d6eeb4ba944c6f3d2413df9790991cc/-/blob/rs/tests/nns/rent_subnet_test.rs?L137-141

rs/tests/nns/BUILD.bazel
```diff
@@ -259,6 +259,14 @@ system_test(
 
 system_test(
     name = "rent_subnet_test",
+    # Too big for the "local" backend under Namespace remote execution: this test
+    # boots 36 VMs on a single host (32 App IC Node VMs + 1 System + 1 fast System
+    # + 1 unassigned), demanding ~51 vCPUs and ~328 GiB of RAM. That fits on
+    # DFINITY's local-backend runners (64 CPUs, plenty of RAM) but not on a
+    # Namespace Bazel Execution Worker (16 CPUs, 32 GiB), where the host thrashes
+    # and nodes never become reachable. Pin this test to the Farm backend, where VMs
+    # are provisioned by the Farm cluster instead of being booted on the Bazel worker.
```
daniel-wong-dfinity-org-twin, line 268: I think you can simply say something like

> Requires a very large amount of resources (328 GiB of RAM), which only farm can provide.

Lead with the lede!

daniel-wong-dfinity-org-twin (approved):
Please, consider [my earlier] comment before merging.

[my earlier]: https://github.com/dfinity/ic/pull/10867#issuecomment-5047057110

basvandijk:
@daniel-wong-dfinity-org-twin thanks for pushing back on this. I forwarded it to Fable 5 which produced: https://github.com/dfinity/ic/pull/10869 which is excellent IMHO.

basvandijk:
Closing in favour of https://github.com/dfinity/ic/pull/10869.

PR #10869, opened 2026-07-22
test(nns): Drastically reduce the resource requirements of rent_subnet_test

basvandijk:
# Motivation

Previously, a very resource-intensive hack was used in order to create a mock Exchange Rate canister at its usual production ID in `rent_subnet_test`.

The hack was to create 32 filler subnets, which requires 328 GiB of RAM (!).

# Changes

To achieve that requirement, we now simply use `use_specified_ids_allocation_range` and `at_ids`. The test's topology thereby shrinks from 35 node VMs (328 GiB of RAM) to 3 (72 GiB).

### DIFF
--- a/rs/nns/test_utils/src/itest_helpers.rs
+++ b/rs/nns/test_utils/src/itest_helpers.rs
@@ -510,41 +510,19 @@ pub async fn install_rust_canister_from_path<P: AsRef<Path>>(
     .await
 }
 
-/// Runtime must be built from a node belonging to a subnet that can host
-/// EXCHANGE_RATE_CANISTER_ID.
-///
-/// Warning: This assumes that canisters with ID smaller than that of the
-/// Exchange Rate canister have all already been created.
+/// Runtime must be built from a node belonging to a subnet that hosts
+/// EXCHANGE_RATE_CANISTER_ID and supports creating canisters at specified IDs
+/// in the canister ID range containing it. In system tests, such a subnet can
+/// be obtained with `InternetComputer::use_specified_ids_allocation_range`.
 pub async fn create_and_install_mock_exchange_rate_canister(
     runtime: &'_ Runtime,
     price_of_icp_in_xdr_cents: u64,
 ) {
-    // Step 1: Create the canister.
-
-    // Create canisters in a loop until we hit EXCHANGE_RATE_CANISTER_ID. Yes,
-    // this is a hack. You might think that
-    // runtime.create_canister_with_specified_id(...) would get the desired
-    // effect (more straightforwardly), but trying to create an Exchange Rate
-    // canister that way results in
-    //
-    //     The `specified_id` uf6dk-hyaaa-aaaaq-qaaaq-cai is invalid because it belongs to the canister allocation ranges of the test environment.
-    //
-    // This is because of the way that the routing table is set up in system
-    // tests. If we wanted to get rid of this hack, we would have to change how
-    // the routing table is set up in system tests:
-    // https://github.com/dfinity/ic/pull/6053#discussion_r2329340517
-    //
-    // The "Warning" in the triple slash comments of this function is because of
-    // this hack.
-    let mut found = false;
-    for _ in 0..100 {
-        let canister = runtime.create_canister(Some(0)).await.unwrap();
-        if canister.canister_id() == EXCHANGE_RATE_CANISTER_ID {
-            found = true;
-            break;
-        }
-    }
-    assert!(found);
+    // Step 1: Create the canister at its usual (mainnet) canister ID.
+    runtime
+        .create_canister_at_id_max_cycles_with_retries(EXCHANGE_RATE_CANISTER_ID.get())
+        .await
+        .unwrap();
 
     // Step 2: Install code into the canister.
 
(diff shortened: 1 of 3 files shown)


rs/nns/test_utils/src/itest_helpers.rs
```diff
@@ -510,41 +510,19 @@ pub async fn install_rust_canister_from_path<P: AsRef<Path>>(
...
 
-/// Runtime must be built from a node belonging to a subnet that can host
-/// EXCHANGE_RATE_CANISTER_ID.
-///
-/// Warning: This assumes that canisters with ID smaller than that of the
-/// Exchange Rate canister have all already been created.
+/// Runtime must be built from a node belonging to a subnet that hosts
+/// EXCHANGE_RATE_CANISTER_ID and supports creating canisters at specified IDs
+/// in the canister ID range containing it. In system tests, such a subnet can
+/// be obtained with `InternetComputer::use_specified_ids_allocation_range`.
```
daniel-wong-dfinity-org-twin, line 516: > use_specified_ids_allocation_range

Ah. This seems to be our savior. Excellent! I would state this more prominently in the PR title and/or description.

rs/tests/nns/rent_subnet_test.rs
```diff
@@ -132,38 +132,36 @@ fn main() -> Result<()> {
...
+            1, // Node count.
+        ))
+        // Assign the first subnet a pair of canister ID ranges: one covering
+        // all mainnet canister IDs (where canisters can be created at
+        // specified IDs), and a disjoint one for the allocation of new
+        // canister IDs. Without this, no canister ID range in the test
+        // environment would contain the usual Exchange Rate canister ID, and
+        // we would have to pad the routing table by creating 32 bogus
+        // single-node Application subnets to make a system subnet end up with
+        // the canister ID range containing that ID.
```
daniel-wong-dfinity-org-twin:
```suggestion
        // environment would contain the usual Exchange Rate canister ID.
```

I don't think we need to talk about how we used to get around the problem. We just need to state the problem that this addresses.

rs/tests/nns/rent_subnet_test.rs
```diff
@@ -573,7 +571,12 @@ fn install_nns_canisters(env: &TestEnv) {
         nns_node.node_id
     );
 
+    // at_ids is needed because, with use_specified_ids_allocation_range (see
+    // setup), new canister IDs on the root subnet are no longer allocated
+    // sequentially starting from 0; instead, the NNS canisters must be
+    // explicitly created at their usual canister IDs.
```
daniel-wong-dfinity-org-twin: Thanks. I would move this to be directly above

```
.at_ids()
```

because this is explaining exactly that. It is unusual, but certainly not unheard of for a comment to appear WITHIN a statement. IMHO, the most important thing is that comments are next to the code they are talking about. Imagine going to an art museum, and the placards are on the other side of the room from the painting they describe. That would be silly.

daniel-wong-dfinity-org-twin:
```suggestion
    // This causes NNS canisters to be created with their well-known
    // production IDs. In order for this to work,
    // use_specified_ids_allocation_range must also be used.
```

daniel-wong-dfinity-org-twin (approved):
Woo! You might want to look for more tests that use this hack, and fix those as well.

daniel-wong-dfinity-org-twin:
Suggestion for PR description

> # Motivation
>
> Previously, a very resource-intensive hack was used in order to create a mock Exchange Rate canister at its usual production ID in a rent_subnet_test.
>
> # Changes
>
> To achieve that requirement, we now simply use use_specified_ids_allocation_range and at_ids.
>
> The hack was to create 32 filler subnets, which requires 328 GiB of RAM (!).

I kind of don't see the point of the "Testing" section, because it just goes without saying that if this is in master, it passed tests.

My suggested description elides secondary changes. I don't think anyone expects the description to be comprehensive, rather, they want to quickly catch the drift.

I think it is important to live by "Lead with the lede", which means selectively eliding secondary information.

daniel-wong-dfinity-org-twin:
Title suggestion:

> test(nns): Drastically reduce the resource requirements of rent_subnet_test.

The number 32 is certainly NOT the lede.

rs/nns/test_utils/src/itest_helpers.rs
basvandijk: Done — the PR title and description now lead with `use_specified_ids_allocation_range` (+ `at_ids`).

rs/tests/nns/rent_subnet_test.rs
basvandijk: Done in 63f2c1238b.

rs/tests/nns/rent_subnet_test.rs
basvandijk: Done in 63f2c1238b — the comment now sits directly above `.at_ids()` within the statement.

basvandijk: Done in 63f2c1238b, with exactly this wording.

basvandijk:
Applied both suggestions: the PR title is now "test(nns): Drastically reduce the resource requirements of rent_subnet_test" and the description leads with the motivation and `use_specified_ids_allocation_range` + `at_ids`, with the Testing section dropped.

Re: looking for more tests that use this hack — the only remaining copy is in `rs/tests/testnets/src_testing.rs` (from which rent_subnet_test originally copied it). Its 32 Application subnets are also used as CMC-authorized subnets for the cycles wallet there, so removing them changes that testnet's topology for its users; I'd rather tackle that in a separate PR.

rs/tests/nns/rent_subnet_test.rs
```diff
@@ -132,38 +132,33 @@ fn main() -> Result<()> {
...
-    // a system subnet that is assigned a canister ID range containing the usual
-    // Exchange Rate canister ID (uf6dk-hyaaa-aaaaq-qaaaq-cai).
-    for _ in 0..32 {
-        ic = ic.add_subnet(
-            Subnet::new(SubnetType::Application)
-                .with_resource_overrides(VmResourceOverrides {
-                    vcpus: Some(NrOfVCPUs::new(1)),
-                    memory_kibibytes: Some(AmountOfMemoryKiB::new(8_389_000)),
-                    ..VmResourceOverrides::default()
+    InternetComputer::new()
```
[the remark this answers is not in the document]
basvandijk: The code compiles so I'll resolve this.

rs/nns/test_utils/src/itest_helpers.rs
```diff
@@ -587,9 +565,13 @@ pub async fn create_and_install_mock_exchange_rate_canister(
...
 /// Subnet Rental canister have all already been created.
 pub async fn create_and_install_mock_subnet_rental_canister(runtime: &'_ Runtime) -> Canister<'_> {
-    // Create canisters in a loop until we hit SUBNET_RENTAL_CANISTER_ID.
-    // This is a hack similar to the one in `create_and_install_mock_exchange_rate_canister`.
-    // See the comment there for more details.
+    // Create canisters in a loop until we hit SUBNET_RENTAL_CANISTER_ID. Yes,
+    // this is a hack. You might think that
+    // runtime.create_canister_with_specified_id(...) would get the desired
+    // effect (more straightforwardly), but the Subnet Rental canister ID
+    // belongs to the canister allocation range of the (root) subnet in test
```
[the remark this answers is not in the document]
basvandijk: Let's handle this for-loop in `create_and_install_mock_subnet_rental_canister` in a separate PR where we also remove this whole comment.

PR #10873, opened 2026-07-23
feat(cketh): detect deposits above minimum

gregorydemay:
A background task periodically reads the on-chain ERC-20 balance of the armed deposit addresses that are due for a scan and flags the ones at or above a per-token minimum as deposit candidates (reported via metrics). A balance is only a cheap scheduling hint — it triggers work but never credits; filter 2 (DEFI-2924) re-verifies against finalized logs before any mint. Nothing runs downstream yet.

Highlights:
- **Deployless batcher instead of Multicall3.** Balances are read with a single create-style `eth_call` (`to` omitted) that runs a small program returning them as a compact flat array — ~5× less response overhead and a trivial decode, with no deployed contract. The shipped bytecode is pinned by a byte-for-byte assembly golden and its behaviour verified on a local anvil node (balances across tokens/holders; fail-loud on a bad token), in addition to end-to-end validation against Ethereum mainnet across all four providers the minter uses and through the deployed EVM-RPC canister.
- **Driven by the schedule.** Scans only the addresses due per #10881's block-based backoff, pins the call to the refreshed latest block height, and advances each address' schedule after a successful scan.
- **Per-token minimums** worth ~0.005 ETH each, covering every token the mainnet and Sepolia minters support (a hard-coded rate snapshot; DEFI-2961 tracks refreshing them daily from the exchange-rate canister).
- **Discoverable progress.** `deposit_erc20` now returns each address' `last_scanned_block` and `scan_count`, and an end-to-end test exercises register → scan → observe.

Deferred: filter 2, blocklist/dedup and crediting (DEFI-2924); exposing the minimum to users (DEFI-2962).

## PR stack
- **#10881** — scan-scheduling layer — base `master`
- **#10873 (this PR)** — balance scan filter 1 (deployless-batcher balances) — stacked on #10881

#10878 (the earlier wall-clock scan-schedule) was closed as superseded by #10881.


rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    let (addresses_scanned, pairs, calls) = read_state(|s| build_calls(s, now));
+    if calls.is_empty() {
+        mutate_state(|s| {
+            s.last_balance_scan = Some(BalanceScanStats {
+                scanned_at_ns: now.as_nanos(),
+                addresses_scanned: 0,
+                candidates_found: 0,
+                chunks_failed: 0,
+            })
+        });
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in a0519137dc — the empty-`calls` early return now records the computed `addresses_scanned` instead of a hard-coded 0, so an armed watchlist with no supported ERC-20 tokens configured is reported accurately (candidates/chunks remain 0 since no call is issued).

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -81,6 +84,12 @@ impl AutomaticDeposits {
         self.watchlist.iter()
     }
 
+    pub fn live_addresses(&self, now: Timestamp) -> impl Iterator<Item = (Account, Address)> + '_ {
+        self.watchlist_iter()
+            .filter(move |(_, entry)| entry.expires_at >= now)
+            .map(|(account, entry)| (*account, entry.value.address))
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive: `icrc_ledger_types::icrc1::account::Account` is `Copy`, so `*account` is a copy, not a move — this compiles. Confirmed by the green `Cargo Build Linux` and `bazel build` checks on this exact code; `clippy::clone_on_copy` earlier in this stack in fact required *removing* an `account.clone()`. No change needed.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    for (account, deposit_address) in state.automatic_deposits.live_addresses(now) {
+        addresses_scanned += 1;
+        for token in &tokens {
+            pairs.push((account, *token));
+            calls.push(BalanceOfCall {
+                token: *token,
+                holder: deposit_address,
+            });
+        }
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 False positive for the same reason: `Account` is `Copy`, so pushing `(account, *token)` inside the per-token loop copies the account rather than moving it — the loop compiles (per the green `Cargo Build Linux` / `bazel build` checks). No change needed.

rs/ethereum/cketh/minter/src/balance_scan/multicall3/mod.rs
```diff
@@ -0,0 +1,209 @@
+use crate::numeric::Erc20Value;
```
gregorydemay: rename file to rs/ethereum/cketh/minter/src/balance_scan/multicall3/mod.rs

gregorydemay: 🤖 Done in 3c9d57d5b8 — `git mv`d `balance_scan/multicall3.rs` → `balance_scan/multicall3/mod.rs` (tests stay at `multicall3/tests.rs`). Build + tests green.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+            record_stats(now, 0, 0, 0);
+            return;
+        }
+    };
+
+    let client = read_state(rpc_client);
+    // Chunk by address so an address' per-token calls never straddle a chunk boundary; this keeps
+    // the per-address scan-state advance below all-or-nothing per chunk.
+    let addresses_per_chunk = (MAX_CALLS_PER_BATCH / tokens.len()).max(1);
+    let mut candidates = 0_usize;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Intentional, and documented in a comment now. Chunks are aligned to whole addresses so the per-address scan-state advance stays all-or-nothing — a partially-scanned address must never be advanced. `tokens.len()` is the number of supported ckERC20 tokens (11 today), far below the 200 cap, so a chunk holds many addresses. 2D chunking (splitting one address across chunks) would reintroduce the partial-advance hazard, so even if the token set ever grew past the cap a single address' calls are still sent together; the batcher response is only 32 bytes per call, so an over-cap single-address chunk stays cheap.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    calls
+        .iter()
+        .zip(balances)
+        .filter(|(call, balance)| **balance >= min_deposit(&call.token))
+        .count()
+}
+
+/// Minimum balance for `token` to count as a scan candidate; a token absent from
+/// [`MIN_DEPOSITS`] never counts.
+fn min_deposit(token: &Address) -> Erc20Value {
```
mbjorkqvist, line 194: IIUC, `supported_ck_erc20_tokens` and `MIN_DEPOSITS` can silently diverge, and returning `Erc20Value::MAX` for an absent token means a supported token with no `MIN_DEPOSITS` entry is scanned (cost incurred) but can never be flagged, so deposits are never detected. The unknown-token case is tested (`should_not_count_candidates_for_an_unsupported_token`), but nothing asserts that a `MIN_DEPOSITS` entry exists for all supported tokens.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    let mut scanned: Vec<Account> = Vec::new();
+
+    for batch in plan_batches(&addresses_to_scan, &erc20_tokens) {
+        let input = batcher::encode_balance_batch(&batch.calls);
+        match client
+            .call(call_args(input, latest_block))
+            .with_cycles(MIN_ATTACHED_CYCLES)
+            .try_send()
+            .await
+            .reduce_with_strategy(AnyOf)
```
mbjorkqvist, line 103: IIUC, using `AnyOf` here could result in a situation where if all other providers return a decode failure, a single bad/malicious provider could result in a deposit not being detected, if it always returned a well-formed all-zeros response.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -149,6 +149,17 @@ impl AutomaticDeposits {
         self.watchlist.get_entry(now, account)
     }
 
+    /// Record that `account`'s deposit address was scanned at `block`, advancing it along the
+    /// backoff schedule (`last_scanned_block = block`, `scan_count += 1`). No-op if the account is
+    /// no longer live as of `now` (expired or evicted).
+    // TODO DEFI-2923: move the watched address with balance to a separate queue for sweeping.
+    pub fn record_scan(&mut self, now: Timestamp, account: &Account, block: BlockNumber) {
```
mbjorkqvist, line 156: nit: Since this PR edits `deposit_from_cex.md` to swap Multicall3 to deployless batcher, it could also clarify that a block-based implementation is used for scheduling, rather than wall-clock time.

rs/ethereum/cketh/minter/src/lib.rs
```diff
@@ -35,6 +36,8 @@ use std::time::Duration;
 pub const MAIN_DERIVATION_PATH: Vec<ByteBuf> = vec![];
 pub const SCRAPING_ETH_LOGS_INTERVAL: Duration = Duration::from_secs(3 * 60);
 pub const REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL: Duration = Duration::from_secs(30);
+// provisional
+pub const BALANCE_SCAN_INTERVAL: Duration = Duration::from_secs(60);
```
mbjorkqvist: Is it correct that the `BALANCE_SCAN_INTERVAL` value of `60s` here is used together with the value from `SCAN_GAP_SECS` to determine the interval (the maximum being chosen)? If so, does it make sense to have any values smaller than `BALANCE_SCAN_INTERVAL` in `SCAN_GAP_SECS`? Currently, the first two values are `30`, but in practice, the scans will now be performed at 60s intervals?

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -1006,6 +1018,29 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+                    )?;
+                    w.encode_gauge(
+                        "cketh_minter_balance_scan_candidates",
+                        stats.candidates_found as f64,
+                        "Number of deposit candidates found in the last balance scan.",
+                    )?;
+                    w.encode_gauge(
+                        "cketh_minter_balance_scan_failed_chunks",
+                        stats.chunks_failed as f64,
+                        "Number of multicall chunks that failed in the last balance scan.",
```
mbjorkqvist: `chunks_failed` is a last-run gauge that lumps decode and call failures together. This may be fine, but it makes it more difficult to figure out the exact source of the problem if there are issues (logs would probably need to be inspected). Also, having this be a *gauge* could make it more troublesome to alert on - an increasing counter could potentially be better.

Also, the "multicall" wording in the description is stale - this metric counts deployless-batcher chunks now, not Multicall3.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+                }
+                Err(e) => {
+                    chunks_failed += 1;
+                    log!(INFO, "balance scan decode error: {e:?}");
+                }
+            },
+            Err(e) => {
+                chunks_failed += 1;
+                log!(INFO, "balance scan eth_call error: {e:?}");
+            }
```
mbjorkqvist, line 118: Every chunk queries every supported token for its addresses, and the batcher `REVERT`s the whole `eth_call` on any sub-call anomaly (revert or non-32-byte return). So a single supported token that breaks will result in  every chunk every tick, so no deposits will be detected. IIUC, the only detection mechanism is the gauge for `chunks_failed`?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -149,6 +149,17 @@ impl AutomaticDeposits {
...
+        if let Some(request) = self.watchlist.get_value_mut(now, account) {
+            request.last_scanned_block = Some(block);
+            request.scan_count = request.scan_count.saturating_add(1);
```
mbjorkqvist, line 159: The first scan in `addresses_to_scan_iter` above takes the `None` path. For the `Some` path, we check `SCAN_GAP_SECS[index]`, where `let index = request.scan_count as usize`, but since we here set `request.scan_count = request.scan_count.saturating_add(1)`, `index` is never `0`, so the value in `SCAN_GAP_SECS[0]` is never used.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/lib.rs
gregorydemay: 🤖 Fixed in 8398925d42 — BALANCE_SCAN_INTERVAL is now 30s (was 60s), matching the smallest SCAN_GAP_SECS gap so the 30s gaps are actually reachable. You're right that the tick interval and the per-address gaps combine: a scan task fires every 30s and an address is due once its block-based elapsed time meets its current gap, so gaps below the tick interval were previously unachievable. (Paired with the off-by-one fix on the indexing so SCAN_GAP_SECS[0]=30s is the first backoff gap.)

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed in 8398925d42 — the index is now `(request.scan_count as usize).saturating_sub(1)`, so SCAN_GAP_SECS[0] is used for the first backoff scan (scan_count 1) and all 33 entries are consulted; the schedule now exhausts at scan_count 34. Retargeted the schedule tests to the N -> SCAN_GAP_SECS[N-1] mapping (incl. the exhaustion boundary).

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Fixed in c7db28ab27 — replaced the single last-run `cketh_minter_balance_scan_failed_chunks` gauge with two cumulative counters, `cketh_minter_balance_scan_decode_errors_total` and `cketh_minter_balance_scan_call_errors_total`, so decode vs eth_call failures are distinguishable and alertable (monotonic, reset on upgrade — standard Prometheus counter-reset semantics, hence excluded from is_equivalent_to like the transient last_balance_scan). Also fixed the stale "multicall" wording to "deployless-batcher". Per-kind INFO logs remain for pinpointing the failing chunk.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Fixed in 6c97d8fc01 — added a unit test (should_have_a_min_deposit_for_every_deployed_supported_token) asserting every currently-deployed supported ckERC20 contract (mainnet + Sepolia minters) has a MIN_DEPOSITS entry, using an independently transcribed hex list so a dropped/typo'd entry is caught (it caught a transcription typo while writing it). Long term, the static MIN_DEPOSITS list is superseded by the live-rate work (DEFI-2961, exchange-rate-canister lookups), which removes the divergence entirely; no new tokens are planned in the near term.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed in 3f38346726 — the Detection schedule section now states scheduling is block-based (elapsed measured as elapsed_blocks × SECS_PER_BLOCK, ~12s/block; see automatic_deposits/mod.rs::addresses_to_scan_iter), not wall-clock, and that the scan task fires on a fixed 30s timer. Reconciled the cadence table with the actual 33-entry SCAN_GAP_SECS (all now used): an immediate initial scan plus 33 gap-gated scans (tail is 23 hourly, not 24), 34 total.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Good catch — this will be hardened in a follow-up (tracked as DEFI-2964, relates-to the DEFI-2917 epic). The plan: the automatic-deposit `latest`-block scans will use a 2-of-3 RPC client with `NoReduction` (instead of today's 3-of-4 + `AnyOf`), so at least two providers must agree on the batch result, defeating a single malicious/faulty provider that always returns a well-formed all-zeros response. This matches the existing `TODO DEFI-2923` at the call site — a lower threshold is acceptable here because the scan reads the latest (non-finalized) height only to notify the sweeper; no minting is gated on it. Keeping it out of this PR to keep the change scoped.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Correct that a single broken supported token would REVERT every chunk every tick, so no deposits would be detected until it's resolved. This is an intentional fail-loud design for a trusted/whitelisted token set: a broken token is an operational incident, not a per-address 'no balance', so we surface it as a loud failure rather than silently recording empty balances. Detection is no longer only a lumped gauge — this PR splits it into the cumulative `cketh_minter_balance_scan_call_errors_total` / `cketh_minter_balance_scan_decode_errors_total` counters (alertable) plus the per-kind INFO logs. No code change planned.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -90,6 +91,17 @@ pub struct State {
...
+    /// Statistics from the most recent balance scan, if any.
+    pub last_balance_scan: Option<BalanceScanStats>,
+
+    /// Cumulative count of balance-scan chunks whose response failed to decode.
+    /// Monotonic in-memory counter, reset on upgrade.
+    pub balance_scan_decode_errors: u64,
+
+    /// Cumulative count of balance-scan `eth_call`s that failed.
+    /// Monotonic in-memory counter, reset on upgrade.
+    pub balance_scan_call_errors: u64,
```
gregorydemay: why are those a separate fields instead of being added to `BalanceScanStats`?

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Follow-up: on reflection we are removing the balance-scan metrics from this PR entirely (fixed in 7db8031194) and deferring observability — including the split decode/eth_call error counters proposed here — to DEFI-2965, so it lands as a coherent, alerting-aware set rather than piecemeal gauges. The per-kind INFO error logs remain in place in the meantime.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Good question — rather than reconcile the placement, we are dropping `BalanceScanStats` / `last_balance_scan` and both error-counter fields altogether (fixed in 7db8031194) and deferring balance-scan observability to DEFI-2965, where the metrics (and their state) will be designed as one coherent set. So there's no longer a placement to reconcile here.

PR #10878, opened 2026-07-23
feat(cketh): dynamic per-address balance-scan schedule

gregorydemay:
Implements the design docs per-address detection cadence on top of the D1 balance scan, replacing D1s flat "scan every live address every tick" with a backoff schedule.

Each armed deposit address is now scanned on the `deposit_from_cex.md` "Detection schedule": burst (30s, 60s, 120s, 240s, 360s, 600s), ramp (every 5 min to 30 min), then hourly to 24h — 33 ticks total. A pure `is_scan_due` check compares an address's age (derived from its watchlist expiry) against a cumulative schedule and its per-account scan count; the global tick timer drops to 30s (the finest step) and each tick scans only the addresses that are due. Per-account progress is transient state (excluded from `is_equivalent_to`), pruned when an address is no longer live and reset on upgrade — a restored address simply re-bursts. No `.did` change; still no downstream action on candidates (filter 2 / crediting remains a later PR).

## PR stack
- #10729 — [F3] deposit_erc20 registration state (base of the stack)
- #10873 — [D1] balance scan, filter 1 (this PR's base)
- **this PR** — dynamic per-address scan schedule

Merge bottom-up: #10729 → #10873 → this. Review only the top commit here; the diff below the base belongs to the PRs above.

### DIFF
--- a/rs/ethereum/cketh/minter/src/balance_scan/mod.rs
+++ b/rs/ethereum/cketh/minter/src/balance_scan/mod.rs
@@ -7,14 +7,25 @@ use crate::eth_rpc_client::{AnyOf, MIN_ATTACHED_CYCLES, ToReducedWithStrategy, r
 use crate::guard::TimerGuard;
 use crate::logs::INFO;
 use crate::numeric::Erc20Value;
+use crate::state::automatic_deposits::DEPOSIT_ADDRESS_SCAN_WINDOW;
 use crate::state::{State, TaskType, mutate_state, read_state};
 use crate::timed_sized_map::Timestamp;
 use ic_canister_log::log;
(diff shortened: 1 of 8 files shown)


gregorydemay:
🤖 Closing as superseded by #10881, which implements the same per-address scan scheduling/selection with a block-based cadence (`last_scanned_block` + `scan_count`, `addresses_to_scan_iter`, latest-block-height timer) instead of the wall-clock `SCAN_SCHEDULE_SECS` approach here. The remaining piece unique to this PR — wiring the schedule into the balance-scan execution task — will land in the follow-up Multicall-execution PR on top of #10881, consuming `addresses_to_scan_iter`. #10873 stays open.

PR #10882, opened 2026-07-23
feat: Make query limits configurable in the SubnetRecord

schneiderstefan:
This commit adds new fields to ResourceLimits (and therefore SubnetRecord) in the registry canister:
- maximum_query_instructions
- maximum_query_walltime_seconds

This way, individual subnets and/or engines can be independently configured to allow more expensive queries. The drawback of setting the value too large is that it will be easy for expensive queries to use up all the resources that other queries/canisters might want to use.

This PR only makes it possible to set the values in the registry. They are read and stored by the replica due to the shared types, but otherwise they are still ignored by the replica until the follow-up PR.

### DIFF
--- a/rs/messaging/src/message_routing/tests.rs
+++ b/rs/messaging/src/message_routing/tests.rs
@@ -749,6 +749,7 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
             resource_limits: ResourceLimits {
                 maximum_state_size: Some(own_maximum_state_size),
                 maximum_state_delta: Some(own_maximum_state_delta),
+                ..Default::default()
             },
 
             ..Default::default()
--- a/rs/messaging/src/message_routing/tests.rs
+++ b/rs/messaging/src/message_routing/tests.rs
@@ -38,7 +38,8 @@ use ic_types::crypto::threshold_sig::ni_dkg::{NiDkgTag, NiDkgTranscript};
 use ic_types::time::Time;
 use ic_types::xnet::{StreamIndexedQueue, StreamSlice};
 use ic_types::{
-    CanisterId, ExecutionRound, NodeId, NumBytes, PrincipalId, Randomness, ReplicaVersion,
+    CanisterId, ExecutionRound, NodeId, NumBytes, NumInstructions, PrincipalId, Randomness,
+    ReplicaVersion,
 };
 use maplit::{btreemap, btreeset};
 use std::{fmt::Debug, str::FromStr, sync::Arc, time::Duration};
@@ -714,6 +715,8 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
         let own_subnet_id = subnet_test_id(13);
         let own_maximum_state_size = NumBytes::new(1 << 30);
         let own_maximum_state_delta = NumBytes::new(1 << 20);
+        let own_maximum_query_instructions = NumInstructions::new(7_000_000_000);
+        let own_maximum_query_call_walltime_seconds = 15;
         let own_subnet_record = SubnetRecord {
             membership: &[node_test_id(1), node_test_id(2)],
             subnet_type: SubnetType::Application,
@@ -749,7 +752,8 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
             resource_limits: ResourceLimits {
                 maximum_state_size: Some(own_maximum_state_size),
                 maximum_state_delta: Some(own_maximum_state_delta),
-                ..Default::default()
+                maximum_query_instructions: Some(own_maximum_query_instructions),
+                maximum_query_call_walltime_seconds: Some(own_maximum_query_call_walltime_seconds),
             },
 
             ..Default::default()
@@ -1057,6 +1061,16 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
         assert_eq!(
             latest_state.resource_limits().maximum_state_delta,
             Some(own_maximum_state_delta)
+        );
+        assert_eq!(
+            latest_state.resource_limits().maximum_query_instructions,
+            Some(own_maximum_query_instructions)
+        );
+        assert_eq!(
+            latest_state
+                .resource_limits()
+                .maximum_query_call_walltime_seconds,
+            Some(own_maximum_query_call_walltime_seconds)
         );
         assert_eq!(
             *registry_settings.lock().unwrap(),
--- a/rs/messaging/src/message_routing/tests.rs
+++ b/rs/messaging/src/message_routing/tests.rs
@@ -716,7 +716,7 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
         let own_maximum_state_size = NumBytes::new(1 << 30);
         let own_maximum_state_delta = NumBytes::new(1 << 20);
         let own_maximum_query_instructions = NumInstructions::new(7_000_000_000);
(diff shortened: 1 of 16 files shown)


rs/protobuf/def/registry/subnet/v1/subnet.proto
```diff
@@ -406,4 +406,12 @@ enum CanisterCyclesCostSchedule {
 message ResourceLimits {
   optional uint64 maximum_state_size = 1;
   optional uint64 maximum_state_delta = 2;
+  // The maximum number of instructions a single query (or composite query)
+  // method execution is allowed to consume.
+  // The protocol uses a default value if the limit of `0` is specified.
+  optional uint64 maximum_query_instructions = 3;
```
daniel-wong-dfinity-org-twin: How can a canister operator find out that they are in danger of (or actually) exceeding this? And how are they supposed to figure out what changes they need to make when they become aware of the problem? I know this is out of scope for this PR, but I think this needs to be considered at some point.

rs/protobuf/def/registry/subnet/v1/subnet.proto
```diff
@@ -406,4 +406,12 @@ enum CanisterCyclesCostSchedule {
...
+  // The maximum number of instructions allowed across an entire composite query
+  // call graph.
+  // The protocol uses a default value if the limit of `0` is specified.
+  optional uint64 maximum_composite_query_instructions = 4;
```
daniel-wong-dfinity-org-twin: Seems like the previous field also covers composite queries. I'm a bit confused how these two new fields interact. I mean, obviously, the stricter one will end up blocking. But then, why have the other one? E.g. if you have

```
maximum_query_instructions: 42
maximum_composite_query_instructions: 43
```

then, it seems to me that isn't that maximum_composite_query_instructions never actually has an effect. Similarly, if it's the other way around (i.e.

```
maximum_query_instructions: 43
maximum_composite_query_instructions: 42
```
) then, the first setting has an effect on composite queries. But if you are going to limit composite queries to 42 cycles, why would you have a higher limit for non-composite queries??

Obviously, I'm missing something here. The comments should fill that gap.

rs/protobuf/def/registry/subnet/v1/subnet.proto
daniel-wong-dfinity-org-twin:
```suggestion

  // The maximum number of instructions a single query (or composite query)
  // method execution is allowed to consume.
  // The protocol uses a default value if the limit of `0` is specified.
  optional uint64 maximum_query_instructions = 3;

```

This file is not consistent, but I think this pattern is better.

why: "empty" space is meaningful. Here, we things that are conceptually separate, so there should be commensurate spatial separation.

rs/registry/admin/bin/update_subnet.rs
```diff
@@ -346,14 +346,24 @@ impl ProposeToUpdateSubnetCmd {
...
+                maximum_composite_query_instructions.or(subnet_record
+                    .resource_limits
+                    .maximum_composite_query_instructions);
             ResourceLimits {
                 maximum_state_size,
                 maximum_state_delta,
+                maximum_query_instructions,
+                maximum_composite_query_instructions,
             }
         });
```
daniel-wong-dfinity-org-twin: This should be spun out, rather than inlined. The name `inherit_from` has [some precedent][prec].

[prec]: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/dfinity/ic%24+fn%20inherit_from&patternType=regexp&case=yes&sm=0

For one thing, this function is already kind of unreasonably long (this is just a heuristic though). The "real" justification here is that what this is doing is a "coherent" unit of work: when a field in resource_limits is blank, populate the field with the value currently used in production (if any)

rs/protobuf/def/registry/subnet/v1/subnet.proto
schneiderstefan: Yeah good point. I guess the main thing they can currently do is observing that their queries trap with OutOfInstructions and adjust their code accordingly.

schneiderstefan: Done.

rs/protobuf/def/registry/subnet/v1/subnet.proto
schneiderstefan: They are not quite the same. Composite queries involves multiple canisters calling each other. The first one is a limit per message, while the second is a limit for the whole call tree.

But I agree it's unnecessarily confusing, and in prod we set both to 5 Billion anyway. I simplified the PR to just have a single parameter, that applies to both.

rs/registry/admin/bin/update_subnet.rs
schneiderstefan: Thanks a lot for the suggestion. PTAL if I understood it correctly.

rs/protobuf/def/registry/subnet/v1/subnet.proto
daniel-wong-dfinity-org-twin: That's a bit like saying, "a motorist can judge if they are going too fast by if they crash into the car in front of them, or the police fines them". What they really need is a speedometer, which they might see in time BEFORE damage is inflicted. More concretely, I think what people would want is (at a minimum) a couple of Prometheus metrics that tells them the total number of query calls and their total cycles consumption (from this, they can derive the average number of cycles per query during the recent past). Even better would be a distribution of cycles consumed per call (they could derive average cycles / query from this as well).

But anyway, this is "just a suggestion for future work". No need to resolve this here.

daniel-wong-dfinity-org-twin (approved):

rs/messaging/src/message_routing/tests.rs
```diff
@@ -749,6 +749,7 @@ fn try_read_registry_succeeds_with_fully_specified_registry_records() {
             resource_limits: ResourceLimits {
                 maximum_state_size: Some(own_maximum_state_size),
                 maximum_state_delta: Some(own_maximum_state_delta),
+                ..Default::default()
```
mraszyk: We should provide a value here as the test exercises "fully specified" registry records.

rs/registry/canister/src/mutations/do_update_subnet.rs
```diff
@@ -788,6 +788,7 @@ mod tests {
                 ResourceLimits {
                     maximum_state_size: Some(NumBytes::new(42)),
                     maximum_state_delta: Some(NumBytes::new(64)),
+                    ..Default::default()
```
mraszyk: this test is supposed to override all fields

mraszyk (approved):
We could also consider refactoring the tests in rs/registry/admin/bin/update_subnet.rs to try all combinations of None, Some(field=None), Some(field=Some) to get systematic test coverage.

mraszyk:
There's also a wall clock time setting: max_query_call_walltime — wall-clock cap for a non-replicated query call; default 10 s (corresponding to the 5B instructions at the rate of 2B instructions per second). Maybe we also want to make this configurable?

rs/registry/canister/src/mutations/do_update_subnet.rs
schneiderstefan: Done.

rs/messaging/src/message_routing/tests.rs
schneiderstefan: Done.

NikolaMilosa (approved):

PR #10884, opened 2026-07-23
feat(governance): Added a new NNS proposal type: UpdateStandardEngineReplicaVersion.

daniel-wong-dfinity-org-twin:

[prev]: https://github.com/dfinity/ic/pull/10865
[next]: https://github.com/dfinity/ic/pull/10885

### DIFF
--- a/rs/nns/governance/api/src/types.rs
+++ b/rs/nns/governance/api/src/types.rs
@@ -693,6 +693,8 @@ pub mod proposal {
         LoadCanisterSnapshot(super::LoadCanisterSnapshot),
         /// Create a canister in a (possibly non-NNS) subnet and install code into it.
         CreateCanisterAndInstallCode(super::CreateCanisterAndInstallCode),
+        /// Change what replica version(s) are run by Cloud Engines.
+        UpdateStandardEngineReplicaVersion(super::UpdateStandardEngineReplicaVersion),
     }
 }
 /// Empty message to use in oneof fields that represent empty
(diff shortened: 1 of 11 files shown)


bjoernek:
**`validate()` on `update_standard_engine_replica_version.rs`** (or the api→pb conversion)

Because the api type makes all three fields `opt`, the two version IDs and `deployment_progress` behave differently when omitted:

- Omitting `new_/old_replica_version_id` → `unwrap_or_default()` yields `""` → `is_potential_full_git_commit_id` rejects it. Good, effectively required.
- Omitting `deployment_progress` → `unwrap_or_default()` yields `0.0`, which is inside `[0.0, 1.0]` and passes validation silently.

So a proposer who forgets (or typos) `deployment_progress` gets an accepted proposal that deploys to 0% rather than an error. Is 0.0-on-omission intended, or should a missing `deployment_progress` be rejected the way a missing version ID effectively is? If it's meant to be required, validating on the `Option` before defaulting (e.g. `None => invalid_proposal_error("deployment_progress is required")`) would make the two behaviors consistent. If 0.0 is a deliberate, safe default, a one-line comment saying so would save the next reader this same question.

daniel-wong-dfinity-org-twin:
Good point. Claude thought of an interesting way to address that: instead of unwrap_or_default, he changed it to unwrap_or(-1.0). This leads to proposal with no explicit deployment_progress being not being created.

bjoernek (approved):

daniel-wong-dfinity-org-twin:
Now that the previous PR has gone in, I rebased+force pushed (and stripped the DO NOT MERGE from the title). Getting ready to merge...

PR #10886, opened 2026-07-23
chore: use new ic-cdk naming

nmattia:
This starts the migration to ic-cdk 0.20, which removes some APIs marked as deprecated in 0.18 (version currently in use: 0.19).

The first step is to use the new module naming. The changes here only change the naming used in imports.

https://github.com/dfinity/cdk-rs/blob/main/ic-cdk/CHANGELOG.md


Bownairo (approved):

rs/nervous_system/long_message/src/lib.rs
```diff
@@ -51,7 +51,7 @@ async fn make_noop_call() {}
 /// Makes a call to a no-op function defined in this library.
 #[cfg(target_arch = "wasm32")]
 async fn make_noop_call() {
-    () = ic_cdk::call(ic_cdk::id(), "__long_message_noop", ())
+    () = ic_cdk::call(ic_cdk::api::canister_self(), "__long_message_noop", ())
```
daniel-wong-dfinity-org-twin: Can you do

```
use ic_cdk::api::canister_self;
```

at the top (and consolidate with existing `use`s if possible)? I think this is usually how things are done.

Ditto elsewhere.

## Rationale

I think originally, we spelled it `ic_cdk::id()`, because if we do the usual thing (i.e. `use` at the top), it would just be spelled `id()`, which is super ambiguous (to a human; obviously, the compiler can hold in its head all the `use`s in effect and resolve it). But `canister_self` doesn't need to be qualified in order to be not confusing in that way.

rs/sns/governance/canister/canister.rs
```diff
@@ -5,7 +5,8 @@ use async_trait::async_trait;
 use ic_base_types::{CanisterId, PrincipalId};
 use ic_canister_log::log;
 use ic_canister_profiler::{measure_span, measure_span_async};
-use ic_cdk::{caller as cdk_caller, init, post_upgrade, pre_upgrade, println, query, update};
+use ic_cdk::api::msg_caller;
+use ic_cdk::{init, post_upgrade, pre_upgrade, println, query, update};
```
daniel-wong-dfinity-org-twin: Consolidate.

daniel-wong-dfinity-org-twin (approved):

eichhorl (approved):

PR #10887, opened 2026-07-23
feat: add crate-upgrade skill

nmattia:
This adds a skill that can be used by both humans and agents to break down a large migration into multiple, readable and reviewable steps.


basvandijk (approved):
This is great.

PR #10888, opened 2026-07-23
feat(registry): Enable blank replica_version_id for engines.

daniel-wong-dfinity-org-twin:
This is "just" a flag flip, but it suddenly makes a whole bunch of code active.

# Future Work

Existing engines cannot yet transition to blank replica_version_id. That will be fixed in an upcoming PR.

# References

[👈 Previous PR][prev]

[prev]: https://github.com/dfinity/ic/pull/10885

### DIFF
--- a/rs/registry/canister/src/flags.rs
+++ b/rs/registry/canister/src/flags.rs
@@ -9,7 +9,7 @@ thread_local! {
     static IS_SUBNET_SPLITTING_ENABLED: Cell<bool> = const { Cell::new(false) };
     static IS_CHUNKIFYING_LARGE_VALUES_ENABLED: Cell<bool> = const { Cell::new(true) };
     static IS_NODE_SWAPPING_ENABLED: Cell<bool> = const { Cell::new(true) };
-    static IS_BLANK_REPLICA_VERSION_ID_FOR_CLOUD_ENGINES_ENABLED: Cell<bool> = const { Cell::new(false) };
+    static IS_BLANK_REPLICA_VERSION_ID_FOR_CLOUD_ENGINES_ENABLED: Cell<bool> = const { Cell::new(true) };
 
     // Temporary flags related to the node swapping feature.
     //
--- a/rs/registry/canister/src/invariants/replica_version.rs
+++ b/rs/registry/canister/src/invariants/replica_version.rs
@@ -33,10 +33,9 @@ use prost::Message;
 /// * Each URL is well-formed.
 /// * Release package hash is a well-formed hex-encoded SHA256 value.
 ///
-/// Exception: a CloudEngine can have a blank replica_version_id in its
-/// SubnetRecord if there is a StandardEngineReplicaVersionRecord. As of July
-/// 22, 2026, this feature is disabled via a flag (but the plan is to enable it
-/// in the not too distant future).
+/// Exception: a CloudEngine is allowed to have a blank replica_version_id in
+/// its SubnetRecord, provided a StandardEngineReplicaVersionRecord exists. In
+/// that case, that record determines the Cloud Engine's replica version.
 pub(crate) fn check_replica_version_invariants(
     snapshot: &RegistrySnapshot,
 ) -> Result<(), InvariantCheckError> {
--- a/rs/registry/canister/unreleased_changelog.md
+++ b/rs/registry/canister/unreleased_changelog.md
@@ -11,6 +11,10 @@ on the process that this file is part of, see
 
 ## Changed
 
+* Cloud Engines are now allowed to have blank `replica_version_id` (in their
+  `SubnetRecord`). In this case, `StandardEngineReplicaVersionRecord` is used to
+  determine the Cloud Engine's replica version.
+
 ## Deprecated
 
 ## Removed


rs/registry/canister/src/flags.rs
```diff
@@ -9,7 +9,7 @@ thread_local! {
     static IS_SUBNET_SPLITTING_ENABLED: Cell<bool> = const { Cell::new(false) };
     static IS_CHUNKIFYING_LARGE_VALUES_ENABLED: Cell<bool> = const { Cell::new(true) };
     static IS_NODE_SWAPPING_ENABLED: Cell<bool> = const { Cell::new(true) };
-    static IS_BLANK_REPLICA_VERSION_ID_FOR_CLOUD_ENGINES_ENABLED: Cell<bool> = const { Cell::new(false) };
+    static IS_BLANK_REPLICA_VERSION_ID_FOR_CLOUD_ENGINES_ENABLED: Cell<bool> = const { Cell::new(true) };
```
bjoernek, line 12: Flipping this makes the doc comment on `check_replica_version_invariants` false. `rs/registry/canister/src/invariants/replica_version.rs:34-38` still reads:

> Exception: a CloudEngine can have a blank replica_version_id in its SubnetRecord if there is a StandardEngineReplicaVersionRecord. As of July 22, 2026, this feature is disabled via a flag (but the plan is to enable it in the not too distant future).

The first sentence is still accurate; the second is falsified by this PR. Could you update it here, so we don't leave a "this is disabled" comment sitting next to code that is now enabled? Suggested replacement:

```rust
/// Exception: a CloudEngine is allowed to have a blank replica_version_id in
/// its SubnetRecord, provided a StandardEngineReplicaVersionRecord exists. In
/// that case, that record determines the Cloud Engine's replica version.
```

(Commenting here rather than inline, since `replica_version.rs` isn't part of this diff.)

bjoernek:
Question rather than a finding: what's the migration path for **existing** Cloud Engines?

I went through every writer of `SubnetRecord.replica_version_id` in the registry canister:

- `do_deploy_guestos_to_all_subnet_nodes.rs:40` calls `check_replica_version_is_elected`, which rejects `""`.
- `do_split_subnet.rs:95` copies the source subnet's value verbatim.
- `do_create_subnet.rs:565` passes the payload straight through, with no election check.

So as far as I can tell this flip only reaches **newly created** engines, via `create_engine` (`rs/engine_controller/canister/canister.rs:137`), which forwards `replica_version_id` unvalidated. Existing Cloud Engines can't be moved to a blank `replica_version_id` at all.

Is that intentional for now, with a follow-up planned, or am I missing a path? Either way it'd be worth a line in the PR description: it shrinks the blast radius of this flip considerably, and it's exactly the context whoever approves the eventual NNS proposal will want.

bjoernek:
Minor, non-blocking, and possibly already considered: 10840 taught `SubnetRegistry::get_replica_version` to resolve a blank `replica_version_id` via the `StandardEngineReplicaVersionRecord`, but its sibling `get_replica_version_record` (`rs/registry/helpers/src/subnet.rs:512-528`) still does `make_replica_version_key(record.replica_version_id)` on the raw field, so it returns `None` for a blank Cloud Engine.

Only a test calls it today, so nothing is broken. But the two methods now disagree, and the name doesn't hint that one is blank-aware and the other isn't. Was leaving it as-is deliberate? If so, a one-line comment there would save the next person the grep.

bjoernek (approved):

rs/registry/canister/src/flags.rs
daniel-wong-dfinity-org-twin: Done.

daniel-wong-dfinity-org-twin:
I think you are right about transitioning engines from non-blank to blank replica_version_id. I've updated the PR description accordingly.

daniel-wong-dfinity-org-twin:
On your point about get_replica_version_record. I think we can do that in another follow up PR.

daniel-wong-dfinity-org-twin:
> Minor, non-blocking, and possibly already considered: 10840 taught `SubnetRegistry::get_replica_version` to resolve a blank `replica_version_id` via the `StandardEngineReplicaVersionRecord`, but its sibling `get_replica_version_record` (`rs/registry/helpers/src/subnet.rs:512-528`) still does `make_replica_version_key(record.replica_version_id)` on the raw field, so it returns `None` for a blank Cloud Engine.
> 
> Only a test calls it today, so nothing is broken. But the two methods now disagree, and the name doesn't hint that one is blank-aware and the other isn't. Was leaving it as-is deliberate? If so, a one-line comment there would save the next person the grep.

Done: https://github.com/dfinity/ic/pull/11244

FYI: Claude tells me that get_replica_version_record was only used in test. This means that the bug wouldn't have shown up in production, but it would have started showing up at some unfortunate time in the future if/when get_replica_version_record started being used.

daniel-wong-dfinity-org-twin:
> Question rather than a finding: what's the migration path for **existing** Cloud Engines?
> 
> I went through every writer of `SubnetRecord.replica_version_id` in the registry canister:
> 
>     * `do_deploy_guestos_to_all_subnet_nodes.rs:40` calls `check_replica_version_is_elected`, which rejects `""`.
> 
>     * `do_split_subnet.rs:95` copies the source subnet's value verbatim.
> 
>     * `do_create_subnet.rs:565` passes the payload straight through, with no election check.
> 
> 
> So as far as I can tell this flip only reaches **newly created** engines, via `create_engine` (`rs/engine_controller/canister/canister.rs:137`), which forwards `replica_version_id` unvalidated. Existing Cloud Engines can't be moved to a blank `replica_version_id` at all.
> 
> Is that intentional for now, with a follow-up planned, or am I missing a path? Either way it'd be worth a line in the PR description: it shrinks the blast radius of this flip considerably, and it's exactly the context whoever approves the eventual NNS proposal will want.

Done: https://github.com/dfinity/ic/pull/11172

Yes, ofc, we want to be able to migrate existing engines, but at the time of this PR, it was something planned for the future. The future has now arrived. Sorry for the delayed reply.

PR #10889, opened 2026-07-23
docs(registry): Explain blank replica_version_id for Cloud Engines.

daniel-wong-dfinity-org-twin:
In short, use StandardEngineReplicaVersionRecord.

### DIFF
--- a/rs/protobuf/def/registry/subnet/v1/subnet.proto
+++ b/rs/protobuf/def/registry/subnet/v1/subnet.proto
@@ -20,7 +20,16 @@ message SubnetRecord {
   // propagation.
   uint64 initial_notary_delay_millis = 8;
 
-  // ID of the Replica version to run
+  // Replica version IDs are generally git commit IDs in the ic git repository,
+  // pointing to the sources from which the Replica was built.
+  //
+  // Must refer to an "elected" Replica version. Concretely, this means that
+  // there is a Registry record (of type ReplicaVersionRecord) with
+  // replica_version_${replica_version_id} as its key.
+  //
+  // For Cloud Engine subnets, this may be blank. In that case, the Replica
+  // version is determined by StandardEngineReplicaVersionRecord (see
+  // registry/standard_engine_replica_version/v1/standard_engine_replica_version.proto).
   string replica_version_id = 9;
 
   // The length of all DKG intervals. The DKG interval length is the number of rounds following the DKG summary.
--- a/rs/protobuf/def/registry/subnet/v1/subnet.proto
+++ b/rs/protobuf/def/registry/subnet/v1/subnet.proto
@@ -27,9 +27,14 @@ message SubnetRecord {
   // there is a Registry record (of type ReplicaVersionRecord) with
   // replica_version_${replica_version_id} as its key.
   //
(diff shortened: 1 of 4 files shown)


rs/protobuf/def/registry/subnet/v1/subnet.proto
```diff
@@ -20,7 +20,16 @@ message SubnetRecord {
...
+  // Replica version IDs are generally git commit IDs in the ic git repository,
+  // pointing to the sources from which the Replica was built.
+  //
+  // Must refer to an "elected" Replica version. Concretely, this means that
+  // there is a Registry record (of type ReplicaVersionRecord) with
+  // replica_version_${replica_version_id} as its key.
+  //
+  // For Cloud Engine subnets, this may be blank. In that case, the Replica
+  // version is determined by StandardEngineReplicaVersionRecord (see
+  // registry/standard_engine_replica_version/v1/standard_engine_replica_version.proto).
```
bjoernek, line 33: Two small things in this block, both fixed by one edit.

**The record is a precondition, not just a consequence.** "this may be blank. In that case, the Replica version is determined by `StandardEngineReplicaVersionRecord`" reads as though blankness is always allowed for Cloud Engines and the record merely explains what happens next. But the invariant only tolerates blank when the record actually exists (`rs/registry/canister/src/invariants/replica_version.rs:104-107`); blank with no record still fails.

**"Must" vs. "may be blank."** Line 26 states the elected-version rule absolutely, and line 30 carves out a case where the field holds something that is not an elected version. It reads as rule-then-exception so most people will reconcile it, but naming it as an exception removes the snag.

Suggested:

```proto
  // Must refer to an "elected" Replica version. Concretely, this means that
  // there is a Registry record (of type ReplicaVersionRecord) with
  // replica_version_${replica_version_id} as its key.
  //
  // Exception: for Cloud Engine subnets, this may be blank, but only if a
  // StandardEngineReplicaVersionRecord exists. In that case, that record
  // determines the Replica version (see
  // registry/standard_engine_replica_version/v1/standard_engine_replica_version.proto).
```

If you take this, the three generated copies under `rs/protobuf/src/gen/` need regenerating too.

pierugo-dfinity (approved):

blind-oracle (approved):

rs/protobuf/def/registry/subnet/v1/subnet.proto
daniel-wong-dfinity-org-twin: I made some changes that I think address your concerns.

> "Must"

I think that is a minor technicality, because > 99% of the time, there will be a StandardEngineReplicaVersionRecord. It's a bit like saying, "You can go to work, but only if office exists". Nobody would ever mention "but only if the office exists". It just goes without saying that if the office is detonated, "You can go to work" no longer applies. I mean, if that has to be mentioned, then, we better say, "You can go to work, but only if the office exists, the sun hasn't exploded, gravity hasn't failed, aliens didn't invade while you were asleep, ${etc}".

Therefore, instead of implying that StandardEngineReplicaVersionRecord is not reliable, I would rather say it like this.

bjoernek (approved):

mbjorkqvist (approved):

PR #10890, opened 2026-07-23
chore: hermetic_cc_toolchain: 4.1.0 -> 4.3.0

basvandijk:
Upgrade `hermetic_cc_toolchain` **4.1.0 → 4.3.0** ([release notes](https://github.com/uber/hermetic_cc_toolchain/releases/tag/v4.3.0)),
which bumps the pinned zig compiler **0.12.0 → 0.15.2**.

Needed to fix [the build determinism issues](https://github.com/dfinity/ic/actions/runs/30307561001/job/90118915465?pr=10579) of `e2fsprogs`: https://github.com/dfinity/ic/pull/10931.

## Adapting our patches

We carry four patches against `hermetic_cc_toolchain`. 4.2.0 rewrote
`toolchain/defs.bzl` and refactored `toolchain/zig-wrapper.zig`, and 4.3.0
touched the wrapper again, so all of them had to be re-ported:

| patch | status |
| --- | --- |
| `strip` (strip debug info off every target for determinism) | re-ported |
| `cache_dir` (per-bazel-config zig cache dir, works around [ziglang/zig#23993](https://github.com/ziglang/zig/issues/23993)) | re-ported |
| `lto` (drop rustc's deprecated `-Wl,-O1` before zig's lld) | re-ported |
| `no_ubsan` (`-fno-sanitize=undefined` in cc mode) | **dropped — upstreamed in 4.3.0** |

`no_ubsan` was added in this PR for the 4.2.0 bump (zig 0.14 turns on
non-trapping UBSan by default at `-O0`, so `__ubsan_handle_*` stayed undefined
when rustc drove the link with `-nodefaultlibs`). 4.3.0 fixes this upstream in
exactly the same way, so our patch is gone.

`lto` needed real work: 4.3.0 rewrote the argument loop in `parseArgs` to
translate `-u SYMBOL` into `-Wl,-u,SYMBOL`, so the `-Wl,-O<n>` drop moves inside
that loop. zig 0.15.2 still warns `ignoring deprecated linker optimization
setting`, so the patch is still needed.

## New fallout from zig 0.15: host shared libraries

zig 0.15.2's lld extends the `--no-allow-shlib-undefined` check (the default
when linking an executable) to *versioned* symbol references, comparing them
against the target glibc stubs. The two libraries we import from the host —
`libsystemd` and `libvirt` — are built against the host glibc, which is newer
than the glibc 2.31 our zig toolchains target, so every link against them failed:

```
ld.lld: error: undefined reference: dlopen@GLIBC_2.34
>>> referenced by .../libsystemd.so (disallowed by --no-allow-shlib-undefined)
```

That broke `//cpp:infogetty_bin` and
`//rs/ic_os/os_tools/guest_vm_runner:guest_vm_runner`. Both libraries are
system-provided, so the host glibc resolves those symbols at runtime; the fix
passes `-Wl,--allow-shlib-undefined` from the library declarations in
`third_party/`, so the check is relaxed only for the targets that link them.

## Fetched with `archive_override` for now

4.3.0 is not in the Bazel Central Registry yet
([bazel-central-registry#9836](https://github.com/bazelbuild/bazel-central-registry/pull/9836)
— green, but unmerged), so it is fetched from the release tarball with
`archive_override` (pinned by sha256). The BCR entry's `MODULE.bazel` is
byte-identical to the archive's, so this goes back to a
`single_version_override` with the same patch list once that PR lands.


bazel/hermetic_cc_toolchain_cache_dir.patch
```diff
@@ -1,28 +1,29 @@
...
      if cache_prefix == "":
          if host_os == "windows":
--            cache_prefix = "C:\\\\Temp\\\\zig-cache"
+-            local_app_data = repository_ctx.os.environ.get("LOCALAPPDATA", "")
+-            cache_prefix = "{}\\zig".format(local_app_data) if local_app_data else "C:\\\\Temp\\\\zig-cache"
 +            fail("windows is not supported")
-         elif host_os == "macos":
-             cache_prefix = "/var/tmp/zig-cache"
-         elif host_os == "linux":
+         else:
```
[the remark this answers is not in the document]
basvandijk: The original patch already failed on windows and since we don't support windows let's leave it in.

basvandijk:
We decided to wait for the zig-0.16 upgrade.

third_party/BUILD.libvirt.bazel
```diff
@@ -1,9 +1,25 @@
...
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
 
 # Use libvirt from the host environment
 cc_import(
-    name = "libvirt",
+    name = "libvirt-internal",
     hdrs = glob(["include/libvirt/*.h"]),
     shared_library = "lib/x86_64-linux-gnu/libvirt.so",
+    visibility = ["//visibility:private"],
+)
```
nmattia, line 10: could this actually be a `cc_library`? where is this used?

MODULE.bazel
```diff
@@ -19,15 +19,23 @@ bazel_dep(name = "buildifier_prebuilt", version = "8.2.0.2", dev_dependency = Tr
...
 bazel_dep(name = "platforms", version = "1.1.0")
-bazel_dep(name = "hermetic_cc_toolchain", version = "4.1.0")
-single_version_override(
+bazel_dep(name = "hermetic_cc_toolchain", version = "4.3.0")
+
+# 4.3.0 is not in the Bazel Central Registry yet
+# (https://github.com/bazelbuild/bazel-central-registry/pull/9836), so fetch the
+# release archive directly. The registry entry is byte-identical to the
+# archive's MODULE.bazel, so this can be turned back into a
+# single_version_override (same patches) once that PR lands.
```
nmattia:
```suggestion
# can be replaced with `single_version_override` once 4.3.0
# is released to BCR
```

third_party/BUILD.libvirt.bazel
basvandijk: As discussed offline, this uses the same trick as in `third_party/BUILD.libsystemd.bazel`, the `cc_library` is used to hide the depth of the include folder.

nmattia (approved):

PR #10892, opened 2026-07-24
feat: Export renamed consumed-cycles metrics under new names

mraszyk:
Introduce MetricsRegistry::register_alias, which re-exports an already-registered metric under an additional name without duplicating the underlying metric (both names share the same source of truth). Use it to export the three consumed-cycles metrics under their new names as a transitional step of renaming them: the old names can only be dropped once the new names have been rolled out to all subnets.

  replicated_state_consumed_cycles_since_replica_started
    -> replicated_state_consumed_cycles
  replicated_state_consumed_cycles_from_replica_start
    -> replicated_state_consumed_cycles_by_use_case
  replicated_state_consumed_cycles_from_replica_start_as_counters
    -> replicated_state_consumed_cycles_by_use_case_as_counters


schneiderstefan (approved):
Presumably, we only drop the old names once all dashboards (including the public one) is updated?

mraszyk:
> Presumably, we only drop the old names once all dashboards (including the public one) is updated?

Indeed, that's why I created a PR that duplicates the metrics so that we can only drop the legacy names once the new names are used everywhere.

PR #10896, opened 2026-07-24
fix: run infer-bazel-targets on namespace

basvandijk:
The `infer-bazel-targets`, which runs on `ubuntu-latest` is very unstable ([example](https://github.com/dfinity/ic/actions/runs/30082683528/job/89447672173?pr=10895)) and often very slow. This switches the job to run on Namespace (on the same runner profile which is [going to be used](https://github.com/dfinity/ic/pull/10579) for running our bazel build and test invocations) which runs the job in under 2 minutes.

These were the resources used when running `infer-bazel-targets` on the Namespace runner:
Screenshot 2026-07-24 at 11 43 01

We could potentially go one class below to 8x16 but considering the RBE setup is using 16x32 for the driver and 16x32 for the RBE workers to I rather use 16x32 for  `infer-bazel-targets` as well to keep things simple and consistent for now. We can always lower it later if needed.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -168,29 +168,40 @@ jobs:
     needs: [config]
     steps:
       # calculate which bazel targets may need to be rebuilt
+      - uses: ./.github/actions/netrc
+      - name: Set up Bazel cache
+        run: |
+          # Creates a bazelrc configuration fragment which tells bazel where the cache lives.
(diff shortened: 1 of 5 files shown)


basvandijk:
Maybe the stability issues are not caused by the runner, because [this](https://github.com/dfinity/ic/actions/runs/30090076943/job/89471165334?pr=10579) `infer-bazel-targets` run is running on a `namespace-profile-rbe-driver-amd64-linux-16x32` but it appears to hang for 12 minutes already. The following is its resource usage:

Screenshot 2026-07-24 at 13 46 22

PR #10897, opened 2026-07-24
feat: Deliver initial spent HTTPS outcalls cycles

eichhorl:
## Background
The new pay-as-you-go pricing for HTTP outcalls consists of four phases:
1. Base cost. This fee is charged for every request upfront.
2. Per-replica cost. The remaining cycles after charging the base cost are split evenly between the participating replicas. Each replica consumes some of their allowance as the HTTP request is processed. In the end, the amount of consumed cycles is gossiped as part of the response share.
3. Consensus cost. This fee is charged for including the aggregated HTTP response as part of a block. The cost is covered by the sum of cycles, that each replica didn't consume out of their per-replica allowance to produce its share. Combined (consensus cost + sum of cycles per included share), this represents the **_initial_** amount of cycles spent on the request. This initial cycles cost is delivered to execution alongside the response. For each participating replica, cycles remaining after charging the initial cost are refunded to the user.
4. Asynchronous refunds. Shares that did not contribute to the initial response (i.e. because they were received too late) are delivered asynchronously, such that their unused cycles can be refunded, as well.

## Proposed Changes
This PR introduces the infrastructure for points 3 and 4 above, and partly implements point 3:

1. We add the `initial_spent` amount of cycles to all aggregated HTTP responses, that carry full payload bodies (meaning a consensus cost must be charged):
a. `CanisterHttpResponseWithConsensus`: The result of a fully/non-replicated outcall carrying one payload
b. `FlexibleCanisterHttpResponses`: The ok result of a flexible outcall carrying between `min`/`max` payloads
c. `FlexibleCanisterHttpError::TooManyRejects`: An err result of a flexible outcall proving that collecting at least `min` ok responses is impossible. It contains the full error responses, as they are interesting to the caller.
2. For now, the `initial_spent` is determined when the response is created by the payload builder, by calculating the sum of the consensus cost, and the cost incurred by each share. Eventually, the response may only be created if the refunds derived from each share actually cover the consensus cost of the full response. This isn't implemented yet.
3. When validating a block, we verify that `initial_spent` was calculated correctly.
4. When turning a Canister HTTP block payload into `ConsensusMessage`s as part of batch delivery, we additionally derive information about the initial cycles spent on each response. Specifically this is the `initial_spent` amount from above, the callback ID and the nodes participating in the response. Instead of an `initial_spent` field, for divergences and flexible responses too large errors we derive the initial spent on the fly by summing up all shares. This is possible because these responses do not carry the fully payload and therefore do not incur a consensus cost.
5. This information is delivered to DSM as part of a new field in the batch. In the future, it will be used to update the "refund status" of the call context, and credit refunds to the caller.
6. The consensus cost functions and constants are implemented as part of the `pricing` crate, which already contains the per-replica cost. For consistency, we also move the functions and constants for the base fee there.


rs/messaging/src/state_machine.rs
```diff
@@ -127,6 +127,7 @@ impl StateMachine for StateMachineImpl {
                 BatchContent::Data {
                     batch_messages,
                     consensus_responses,
+                    canister_http_spent: _,
```
pierugo-dfinity, line 130: Maybe add a TODO that mentions we'll read that field in the future?

rs/interfaces/src/canister_http.rs
```diff
@@ -58,6 +58,14 @@ pub enum InvalidCanisterHttpPayloadReason {
         spent: Cycles,
         limit: Cycles,
     },
+    /// The collective initial spent cycles included in the payload do not match
+    /// the value recomputed from the request context's subnet size and the
+    /// signed per-replica receipts.
+    InitialSpentMismatch {
+        callback_id: CallbackId,
+        payload_spent: Cycles,
+        computed_spent: Cycles,
```
pierugo-dfinity: Nit: the other variants in that enum appears to use `received`/`actual` and `calculated`/`expected`. Maybe we can use the same wording here.

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -0,0 +1,295 @@
...
+    n * (CONSENSUS_PER_NODE_BYTE_FEE * n + CONSENSUS_BYTE_FEE)
+}
+
+/// Computes the collective initial spent cycles for a fully-replicated (or
+/// non-replicated) HTTP outcall response.
+///
+/// The spend is the sum of the per-replica spends claimed in the proof's payment
+/// receipts, plus the consensus cost `N * (10 * N + 600) * <response_size>` of
+/// including the aggregated response in a block.
+pub fn fully_replicated_initial_spent(
```
pierugo-dfinity: Nit: I think the name of this function is a bit confusing because it is also used for non-replicated outcalls. Maybe `non_flexible_initial_spent` would be more appropriate?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -418,14 +432,16 @@ pub(crate) fn find_flexible_result(
...
     pool_access: &dyn CanisterHttpPool,
 ) -> FlexibleFindResult {
     let mut entries_sorted_asc: Vec<_> = grouped_shares.iter().collect();
     entries_sorted_asc.sort_unstable_by_key(|(metadata, _)| metadata.content_size);
 
     let min_responses = min_responses as usize;
     let mut ok_responses: Vec<(CanisterHttpResponse, &CanisterHttpResponseShare)> = Vec::new();
-    let mut ok_responses_size = size_of::<CallbackId>();
+    // Account for the serialized `initial_spent` field alongside the callback id.
+    let mut ok_responses_size = size_of::<CallbackId>() + size_of::<Cycles>();
```
pierugo-dfinity: Maybe we could put that into a helper method of `FlexibleCanisterHttpResponses`?

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1143,7 +1313,7 @@ fn flexible_error_into_consensus_response(
 /// The function includes request id, which is also part of the hashed value.
 fn divergence_response_into_reject(
     response: CanisterHttpResponseDivergence,
-) -> Option<ConsensusResponse> {
+) -> Option<(CallbackId, ConsensusResponse)> {
```
pierugo-dfinity: Can't callers get the corresponding `CallbackId` through `response.callback`?

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -1105,6 +1167,63 @@ fn validate_payload_succeeds_for_valid_non_replicated_response() {
...
+fn validate_payload_fails_for_initial_spent_mismatch() {
+    // ARRANGE
+    let subnet_size = 4;
+    test_config_with_http_feature(true, subnet_size, |mut payload_builder, _| {
+        let delegated_node_id = node_test_id(1);
+        let callback_id = CallbackId::from(77);
+
+        let request_context = CanisterHttpRequestContext {
+            subnet_size: NumberOfNodes::from(subnet_size as u32),
+            ..request_context(Replication::NonReplicated(delegated_node_id))
```
pierugo-dfinity, line 1180: Maybe we could also have a unit test for the fully replicated case

pierugo-dfinity (approved):

rs/interfaces/src/canister_http.rs
eichhorl: Done

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Good point, done

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Done

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Nice!

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Added

PR #10906, opened 2026-07-26
feat(system-tests): write the JUnit XML report from the test driver

basvandijk:
## What

For system-tests, convert the `SystemGroupSummary` to a JUnit XML file and write it to [`$XML_OUTPUT_FILE`](https://bazel.build/reference/test-encyclopedia#initial-conditions).

## Why

It shows a nice test summary in [BuildBuddy](https://dash.dm1-idx1.dfinity.network/invocation/0cf0aeeb-869b-4c99-989e-e7715368ee41?target=%2F%2Frs%2Ftests%2Fexecution%3Ainter_canister_queries_test&targetStatus=5):
Screenshot 2026-07-26 at 17 17 17

The real reason for working on this was that Bazel declares `test.xml` as an output of every test action. When a test doesn't write `$XML_OUTPUT_FILE` itself, `StandaloneTestStrategy` runs a **second** spawn (`@bazel_tools//tools/test:test_xml_generator`) to derive it from `test.log`. The guard in Bazel 9.2.0 is:

```java
Path xmlOutputPath = resolvedPaths.getXmlOutputPath();
if (fileOutErr.getOutputPath().exists() && !xmlOutputPath.exists()) {
    Spawn xmlGeneratingSpawn = createXmlGeneratingSpawn(...);
    …dispatch via SpawnStrategyResolver…
}
```

Under remote execution that spawn is dispatched to RBE like any other, so it has to queue for a free worker to do well under a second of work. Measured on the `bazel-test-all-rbe` profile of [run 30158073133](https://github.com/dfinity/ic/actions/runs/30158073133), every one of the 152 executed system tests ran exactly two remote spawns:

| | queue | exec wall |
|---|---|---|
| spawn 1 (the test) | median 770s | median 144s |
| spawn 2 (XML generation) | **median 435s, p90 1726s** | median 0.80s, max 4.0s |

That's 100,705s of aggregate queue occupancy for ~2 minutes of real work.

There is no flag to turn this off — `--experimental_split_xml_generation` was removed in Bazel 9 (no such option in `bazel help test --long`, and no `split_xml_generation` string in `A-server.jar`), and the generator is a private implicit attribute (`$xml_generator_script`), so it can't be overridden per target. The only lever is for the test to produce `test.xml` itself.

## How

- `SystemGroupSummary::to_junit_xml()` renders the summary as JUnit XML: one `<testsuite>` for the group, one `<testcase>` per task, with `<failure>` / `<skipped>` bodies.
- `SystemGroupSummary::write_junit_xml_report()` writes it to `$XML_OUTPUT_FILE`, no-op when the variable is unset (not running under `bazel test`).
- Called from `SystemTestGroup::execute` right after `create_report`, before the failure `bail!`, so it runs for passing and failing tests alike. A write error only warns — it must never fail a test.
- `xml_escape()` escapes the five XML entities **and** replaces code points XML 1.0 cannot represent. Both are needed: task messages carry arbitrary output from the system under test, so an unescaped `<` or a stray `\x1b` from a coloured log line would make the report unparsable.

### Drive-by fix

`rs/tests/idx/test_driver_tests.rs` spawns the driver ~10 times as child processes, which inherit the environment. Once the driver started honouring `$XML_OUTPUT_FILE`, each child overwrote the `test.xml` that Bazel expects `//rs/tests/idx:test_e2e_scenarios` itself to produce — confirmed on the first run, where the file held one scenario's tasks instead of the 21 Rust test names, and the panic scenarios would have written a bogus `<failure>` into a passing test's XML. Fixed with a `driver_command()` helper that does `env_remove("XML_OUTPUT_FILE")`, which also removes the duplicated spawn boilerplate at all eight call sites.

### DIFF
--- a/rs/tests/driver/src/driver/group.rs
+++ b/rs/tests/driver/src/driver/group.rs
@@ -41,7 +41,7 @@ use clap::Parser;
 use itertools::Itertools;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
-use slog::{Logger, debug, info, trace, warn};
+use slog::{Logger, debug, info, trace};
 use std::path::PathBuf;
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -1558,20 +1558,19 @@ impl SystemTestGroup {
                     let event: log_events::LogEvent<_> = report.clone().into();
                     // Emit a json log event, to be consumed by log post-processing tools.
                     event.emit_log(group_ctx.log());
+                    // Write the JUnit XML report Bazel expects at $XML_OUTPUT_FILE in case the latter is set..
+                    if let Some(xml_output_file) = std::env::var_os("XML_OUTPUT_FILE") {
+                        let path = PathBuf::from(xml_output_file);
+                        if let Some(parent) = path.parent() {
+                            std::fs::create_dir_all(parent).unwrap_or_else(|_| {
+                                panic!("Failed to create the directory of {}", path.display())
+                            });
+                        }
+                        std::fs::write(&path, report.to_junit_xml()).unwrap_or_else(|_| {
+                            panic!("Failed to write the JUnit XML report to {}", path.display())
+                        });
+                    }
                     info!(group_ctx.log(), "Report:\n{}", report.pretty_print());
-                }
-
-                // Write the JUnit XML report Bazel expects at $XML_OUTPUT_FILE. This is
-                // independent of --no_summary_report: it's for Bazel, not for the reader
-                // of the log. Failing to write it must never fail the test, so just warn.
-                match report.write_junit_xml_report() {
-                    Ok(Some(path)) => info!(
-                        group_ctx.log(),
-                        "Wrote the JUnit XML report to {}",
-                        path.display()
-                    ),
-                    Ok(None) => (),
-                    Err(e) => warn!(group_ctx.log(), "{e:?}"),
                 }
 
                 if with_farm && !args.no_delete_farm_group {
--- a/rs/tests/driver/src/driver/group.rs
+++ b/rs/tests/driver/src/driver/group.rs
@@ -41,7 +41,7 @@ use clap::Parser;
 use itertools::Itertools;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
-use slog::{Logger, debug, info, trace};
+use slog::{Logger, debug, info, trace, warn};
 use std::path::PathBuf;
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -1559,6 +1559,19 @@ impl SystemTestGroup {
                     // Emit a json log event, to be consumed by log post-processing tools.
                     event.emit_log(group_ctx.log());
                     info!(group_ctx.log(), "Report:\n{}", report.pretty_print());
+                }
+
+                // Write the JUnit XML report Bazel expects at $XML_OUTPUT_FILE. This is
+                // independent of --no_summary_report: it's for Bazel, not for the reader
+                // of the log. Failing to write it must never fail the test, so just warn.
+                match report.write_junit_xml_report() {
+                    Ok(Some(path)) => info!(
+                        group_ctx.log(),
+                        "Wrote the JUnit XML report to {}",
+                        path.display()
+                    ),
+                    Ok(None) => (),
+                    Err(e) => warn!(group_ctx.log(), "{e:?}"),
                 }
 
                 if with_farm && !args.no_delete_farm_group {
--- a/rs/tests/driver/src/driver/report.rs
+++ b/rs/tests/driver/src/driver/report.rs
@@ -1,11 +1,9 @@
 use std::{
     collections::{BTreeMap, HashMap},
     fmt::{Display, Formatter, Result},
-    path::PathBuf,
     time::{Duration, SystemTime},
 };
 
-use anyhow::Context;
 use serde::{Deserialize, Serialize};
 
 use crate::driver::event::TaskId;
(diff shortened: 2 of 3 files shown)


rs/tests/driver/src/driver/report.rs
```diff
@@ -128,6 +228,48 @@ impl TaskReport {
...
+            }
+        };
+        format!(
+            "    <testcase name=\"{}\" classname=\"{}\" status=\"run\" duration=\"{:.3}\" time=\"{:.3}\">{}</testcase>\n",
+            xml_escape(&self.name),
+            classname,
+            self.runtime,
+            self.runtime,
+            body,
+        )
```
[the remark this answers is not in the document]
basvandijk: Good catch — fixed in 4727218cbb. `status` is now derived from the outcome:

```rust
let status = match outcome {
    TaskOutcome::Passed | TaskOutcome::Failed => "run",
    TaskOutcome::Skipped => "notrun",
};
```

I kept the attribute rather than dropping it: Bazel's own `@bazel_tools//tools/test:test_xml_generator` emits `status` as well, so consumers already expect it here — the inconsistency was the value, not the attribute's presence. `run`/`notrun` is GoogleTest's convention, which that generator follows.

I also tightened the unit test while in there. It was matching the `<testcase>` opening tag in two loose fragments and so wouldn't have caught this; it now asserts the full tag for both a passed and a skipped task.

rs/tests/driver/src/driver/group.rs
```diff
@@ -1561,6 +1561,19 @@ impl SystemTestGroup {
                     info!(group_ctx.log(), "Report:\n{}", report.pretty_print());
                 }
 
+                // Write the JUnit XML report Bazel expects at $XML_OUTPUT_FILE. This is
+                // independent of --no_summary_report: it's for Bazel, not for the reader
+                // of the log. Failing to write it must never fail the test, so just warn.
```
nmattia: what's the link with `--no_summary_report`?

rs/tests/driver/src/driver/report.rs
```diff
@@ -86,6 +84,75 @@ impl SystemGroupSummary {
...
+    /// `bazel test`.
+    ///
+    /// Writing this file matters beyond the report itself: `test.xml` is a
+    /// declared output of every Bazel test action, so when a test doesn't produce
+    /// it, Bazel runs a *second* spawn
+    /// (`@bazel_tools//tools/test:test_xml_generator`) to derive it from
+    /// `test.log`. Under remote execution that spawn has to queue for a free
+    /// worker -- often for minutes -- to do well under a second of work. Writing
+    /// the file ourselves makes Bazel skip that spawn entirely.
+    pub fn write_junit_xml_report(&self) -> anyhow::Result<Option<PathBuf>> {
```
nmattia: nit: how about passing the target file as an argument? the caller references (and expects) `XML_OUTPUT_FILE` which is hidden down here, and the logic here in `write_junit_xml_report` mixes env lookup + actually report creation; moving everything to the env out of here would simplify things and make it easier to follow

rs/tests/driver/src/driver/report.rs
```diff
@@ -100,6 +167,39 @@ impl SystemGroupSummary {
...
+
+/// Escapes `s` so it can be used both as XML character data and as an attribute
+/// value, and replaces the code points that XML 1.0 cannot represent at all
+/// (most control characters) with `?`.
+///
+/// Both halves are needed: task messages carry arbitrary output from the system
+/// under test, and a single unescaped `<` or a stray escape sequence from a
+/// coloured log line would make the whole report unparsable.
+fn xml_escape(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
```
nmattia: nit: I think `new()` would be fine here; we'll be allocating more anyway because every escape will grow the string

rs/tests/driver/src/driver/report.rs
```diff
@@ -128,6 +228,56 @@ impl TaskReport {
         res.append(&mut report_lines);
         res
     }
+
+    /// Renders this task as a JUnit `<testcase>` element, already escaped and
+    /// indented to sit inside a `<testsuite>`. `classname` is expected to be
```
nmattia: nit: the XML is machine-read so we can probably simplify and skip the indentation logic

rs/tests/driver/src/driver/report.rs
```diff
@@ -128,6 +228,56 @@ impl TaskReport {
...
+            TaskOutcome::Passed | TaskOutcome::Failed => "run",
+            TaskOutcome::Skipped => "notrun",
+        };
+        format!(
+            "    <testcase name=\"{}\" classname=\"{}\" status=\"{}\" duration=\"{:.3}\" time=\"{:.3}\">{}</testcase>\n",
+            xml_escape(&self.name),
+            classname,
+            status,
+            self.runtime,
+            self.runtime,
```
nmattia: what's the difference between `duration` and `time`?

rs/tests/idx/test_driver_tests.rs
```diff
@@ -59,11 +59,24 @@ fn create_unique_working_dir() -> PathBuf {
...
+/// `XML_OUTPUT_FILE` is dropped from the child's environment on purpose: the
+/// driver writes its JUnit XML report there, and these children are not the
+/// process Bazel launched as the test, so they would otherwise each overwrite
+/// the test.xml that Bazel expects *this* test to produce.
+fn driver_command(scenario_name: &str) -> Command {
     let binary_path = env::current_dir().unwrap().join(BINARY_PATH);
     let mut cmd = Command::new(binary_path);
-    cmd.env("TEST_SCENARIO_NAME", scenario_name).args([
+    cmd.env("TEST_SCENARIO_NAME", scenario_name)
+        .env_remove("XML_OUTPUT_FILE");
```
nmattia, line 72: does `testlib` read this as well? is that what's causing the issue?

rs/tests/driver/src/driver/group.rs
basvandijk: `--no_summary_report` is set for colocated system-tests. I actually now moved the XML_OUTPUT_FILE creation under the `if !args.no_summary_report` condition.

rs/tests/driver/src/driver/report.rs
basvandijk: I've now just inlined it into the original callsite

rs/tests/driver/src/driver/report.rs
basvandijk: Done.

rs/tests/idx/test_driver_tests.rs
basvandijk: No, but multiple different test will write to the same XML_OUTPUT_FILE which doesn't make sense.

rs/tests/driver/src/driver/report.rs
basvandijk: No difference. Apparentely different consumers use different fields. But I just removed the `duration` field and BuildBuddy still parses it correctly.

nmattia (approved):

PR #10910, opened 2026-07-27
feat: Add a `cooling_down` flag to `SubnetTopology`

mraszyk:
A "cooling down" subnet stops being sent new messages and lets the messages already in flight drain.
While a subnet is cooling down:

* It inducts all messages from its incoming stream slices.
* It inducts no ingress messages.
* No subnet routes messages to its (outgoing) stream (incl. the loopback stream) to the subnet that is cooling down, except for messages (responses) in subnet output queues that are always routed into streams.
* It keeps executing messages in its input queues.

There is no registry field backing the flag yet, so it is always `false` in production and only tests set it. The registry plumbing that can actually set the flag must not ship until this code is deployed everywhere, since it also introduces a new error code that requires a staged roll-out.


rs/replicated_state/src/replicated_state.rs
```diff
@@ -800,6 +800,17 @@ impl ReplicatedState {
         }
     }
 
+    /// Returns whether this subnet is cooling down. Defaults to `false` if the
+    /// network topology is not populated.
+    ///
+    /// See [`ic_replicated_state::SubnetTopology::cooling_down`] for the exact
+    /// semantics.
+    pub fn is_own_subnet_cooling_down(&self) -> bool {
```
alin-at-dfinity: Nit: The `own_subnet` part is redundant. For one, there is no `SubnetId` argument, so it can only be our own subnet.

Else, you can move it into `SystemMetadata`, to make it even more obvious that it refers to "this subnet" (and pollute less). Then callers would need to use `state.metadata.is_cooling_down()`, but that's perfectly fine IMHO.

rs/replicated_state/src/metadata_state.rs
```diff
@@ -408,6 +418,32 @@ pub struct SubnetTopology {
     pub chain_keys_held: BTreeSet<MasterPublicKeyId>,
     pub cost_schedule: CanisterCyclesCostSchedule,
     pub subnet_admins: BTreeSet<PrincipalId>,
+    /// Whether the subnet is "cooling down", i.e. quiescing: it stops accepting new
```
```suggestion
    pub subnet_admins: BTreeSet<PrincipalId>,

    /// Whether the subnet is "cooling down", i.e. quiescing: it stops accepting new
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -408,6 +418,32 @@ pub struct SubnetTopology {
...
+    /// messages and lets the subnet messages already in flight drain. While a subnet
```
alin-at-dfinity: While it is true that the subnet stops accepting new ingress messages, it will definitely induct canister messages (that were already in streams; or that are still being routed into streams by subnets with an older registry version).

I would even argue that inducting messages from streams is more central to what we're trying to achieve (empty the streams and subnet queues) than rejecting / not inducting ingress messages (which merely avoids generating any extra subnet messages while we're trying to consume them).

Edit: Actually not inducting ingress messages is just as important, as we're also trying to empty the ingress history of everything but `Processing` statuses.

With a view to their relative importance, I would personally structure the bullet points below like: (i) induct all messages from streams; (ii) don't route messages to streams (maybe except subnet responses?); (iii) consume subnet messages, as usual; and (iv) don't execute canister messages (to avoid creating more subnet messages).

Edit: So either before (i) or before (iii) there should be a "don't induct any more ingress messages".

rs/execution_environment/src/scheduler/scheduler_metrics.rs
```diff
@@ -152,6 +153,11 @@ impl SchedulerMetrics {
                 "The number of times inner_rounds()'s loop exited because \
                       max allowed instructions were consumed.",
             ),
+            round_inner_iteration_skipped_cooling_down: metrics_registry.int_counter(
+                "scheduler_round_inner_iteration_skipped_cooling_down",
```
alin-at-dfinity: Nit: Counter metrics should technically have a `_total` or `_count` suffix.

rs/execution_environment/src/execution_environment.rs
alin-at-dfinity, line 3435: It may be worth elaborating a bit on what effects this has. E.g. it is not immediately obvious that returning an error will cause the ingress message to not be inducted (as opposed to, e.g., simply rejected).

(It is actually obvious, now that I've thought about it, but it's still worth mentioning, since it took me a while.)

alin-at-dfinity, line 4791: Idle question, for my own edification: assuming that any such stop contexts exist, wouldn't it make sense to time them out by now? Or were we concerned that the caller would not expect a (or this specific) reject response?

rs/protobuf/def/state/metadata/v1/metadata.proto
```diff
@@ -33,6 +33,9 @@ message SubnetTopology {
   repeated types.v1.MasterPublicKeyId chain_keys_held = 6;
   registry.subnet.v1.CanisterCyclesCostSchedule canister_cycles_cost_schedule = 7;
   repeated types.v1.PrincipalId subnet_admins = 8;
+  // Whether the subnet is "cooling down", i.e. quiescing: it accepts no new
+  // ingress or inter-canister messages and executes no canister messages.
+  bool cooling_down = 9;
```
alin-at-dfinity, line 44: Not entirely accurate: we do induct XNet messages, don't we? We just no longer route them.

rs/ingress_manager/src/ingress_selector.rs
```diff
@@ -530,6 +530,16 @@ impl IngressManager {
         num_messages: usize,
         cycles_needed: &mut BTreeMap<CanisterId, Cycles>,
     ) -> ValidationResult<IngressPayloadValidationError> {
+        // A cooling down subnet inducts no ingress messages at all. Note that this is
+        // also checked by the ingress filter on the receiving nodes, so under normal
+        // circumstances such a message never even reaches the ingress pool; this check
+        // is what makes it binding, both when building and when validating a payload.
```
alin-at-dfinity: Couldn't we have a race condition where the "cooling down" flag is flipped while the ingress pool is non-empty?

If so, the ingress filter check is technically only an optimization. This is the only check than ensures correctness.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -461,8 +464,41 @@ impl StreamBuilderImpl {
...
+                    // the destination subnet is cooling down: it is illegal there
+                    // permanently, so a transient rejection would be misleading.
+                    let is_illegal_engine_msg = (is_engine_dst || is_engine_src)
+                        && (msg.deadline() == NO_DEADLINE || msg.cycles() > Cycles::zero());
+
+                    // A cooling down destination subnet must not be sent any messages.
+                    // Requests are rejected below, so that the caller is not left
+                    // waiting; responses are instead retained in the output queue (a
+                    // response cannot be rejected; and dropping it would leave the
+                    // originator hanging forever), along with everything behind them.
```
alin-at-dfinity: How about subnet responses to local canisters? We won't be able to empty the subnet output queues (and worse, we will then never deliver any such responses) unless we make an exception and route them into the loopback stream.

Regardless, we may benefit from tests covering subnet requests and responses specifically.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -461,8 +464,41 @@ impl StreamBuilderImpl {
...
+                            .is_some_and(|t| t.subnet_type == SubnetType::CloudEngine);
+                    let is_engine_src =
+                        !is_loopback_stream && own_subnet_type == SubnetType::CloudEngine;
+                    // A message that must never cross an engine boundary: a
+                    // guaranteed-response message, or one carrying cycles. Such a message
+                    // is always handled by the engine boundary arms below, whether or not
+                    // the destination subnet is cooling down: it is illegal there
+                    // permanently, so a transient rejection would be misleading.
+                    let is_illegal_engine_msg = (is_engine_dst || is_engine_src)
+                        && (msg.deadline() == NO_DEADLINE || msg.cycles() > Cycles::zero());
```
alin-at-dfinity: Nit: Not your code, but since you're moving it I would rewrite it for efficiency (my formatting, take it with a grain of salt):

```suggestion
                    // A message that must never cross an engine boundary: a
                    // guaranteed-response message, or one carrying cycles. Such a message
                    // is always handled by the engine boundary arms below, whether or not
                    // the destination subnet is cooling down: it is illegal there
                    // permanently, so a transient rejection would be misleading.
                    let is_illegal_engine_msg = (msg.deadline() == NO_DEADLINE || msg.cycles() > Cycles::zero())
                        && !is_loopback_stream
                        && (own_subnet_type == SubnetType::CloudEngine
                            || network_topology
                                .subnets()
                                .get(&dst_subnet_id)
                                .is_some_and(|s| s.subnet_type == SubnetType::CloudEngine));
```

That way, we don't always look up the destination subnet type (relatively expensive) whether the message would be illegal or not.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -547,6 +568,18 @@ impl StreamBuilderImpl {
...
+                        // is not left waiting for the subnet to come back. Responses were
+                        // already retained above, so this is the only kind of message that
+                        // can still be headed for a cooling down subnet.
+                        RequestOrResponse::Request(req) if is_cooling_down_dst => {
+                            self.observe_message_type_status(
+                                LABEL_VALUE_TYPE_REQUEST,
+                                LABEL_VALUE_STATUS_COOLING_DOWN,
+                            );
+                            cooling_down_requests_to_reject.push(req);
+                        }
```
alin-at-dfinity: I've already raised the question of whether rejecting requests to cooling down subnets is necessary (or productive) in the design doc.

So I'll just point out here that were we to not reject them, we could simplify the code by dropping this branch altogether (and the extra logic below) and the `matches!(msg, RequestOrResponse::Response(_))` above where we skip output queues to cooling down subnets.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -705,6 +738,16 @@ impl StreamBuilderImpl {
             );
         }
 
+        for req in cooling_down_requests_to_reject {
+            let dst_canister_id = req.receiver;
+            self.reject_local_request(
+                &mut state,
+                &req,
+                RejectCode::SysTransient,
+                format!("Canister {dst_canister_id} is deployed to a subnet that is cooling down"),
```
alin-at-dfinity: Nit:
```suggestion
                format!("Canister {dst_canister_id} is hosted by a subnet that is cooling down"),
```

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -793,6 +836,10 @@ impl StreamBuilderImpl {
                             .get(&dst_subnet_id)
                             .is_some_and(|t| t.subnet_type == SubnetType::CloudEngine);
                     let is_engine_src = !is_loopback_stream && own_is_engine;
+                    // Checked before the cool-down check below: a refund always
+                    // carries cycles, so one at an engine boundary is illegal there
+                    // permanently, not transiently, and must still raise the critical
+                    // error rather than being retained.
```
alin-at-dfinity: Nit: I believe that rather than raising a critical error, the primary aim of this check is to drop the refund. The critical error is secondary.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,500 @@ fn build_streams_drops_refunds_at_engine_boundary() {
     }
 }
 
+/// The local canister used by the cooling down tests, hosted by `LOCAL_SUBNET`.
+const COOLING_DOWN_LOCAL_CANISTER: CanisterId = CanisterId::from_u64(0);
```
alin-at-dfinity: This can easily be misread as "a canister on the local, cooling down subnet". I would just name it `LOCAL_CANISTER`.

If anything I would use `COOLING_DOWN_SUBNET` instead of the plain `REMOTE_SUBNET`.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,500 @@ fn build_streams_drops_refunds_at_engine_boundary() {
...
+/// The remote canister used by the cooling down tests, hosted by `REMOTE_SUBNET`
+/// (except in `new_loopback_cooling_down_fixture()`, where it too is local).
+const COOLING_DOWN_REMOTE_CANISTER: CanisterId = CanisterId::from_u64(1);
```
alin-at-dfinity: Maybe `COOLING_DOWN_CANISTER` better matches the "remote except in fixture X".

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,500 @@ fn build_streams_drops_refunds_at_engine_boundary() {
...
+/// makes any difference to a message headed for a cooling down (non-engine)
+/// subnet; contrast with `is_illegal_engine_msg` in `build_streams()`.
+fn deadline_and_cycles_matrix() -> impl Iterator<Item = (CoarseTime, Cycles)> {
+    [NO_DEADLINE, SOME_DEADLINE]
+        .into_iter()
+        .flat_map(|deadline| {
+            [Cycles::zero(), Cycles::new(100)]
+                .into_iter()
+                .map(move |cycles| (deadline, cycles))
+        })
```
alin-at-dfinity: It would probably be just as concise / verbose (and more test-like) to simply list the 4 combinations:

```suggestion
    [
        (NO_DEADLINE, Cycles::zero()),
        (NO_DEADLINE, Cycles::new(100)),
        (SOME_DEADLINE, Cycles::zero()),
        (SOME_DEADLINE, Cycles::new(100)),
    ].into_iter()
```

alin-at-dfinity (commented):
LGTM, modulo some nitpicks and a couple of bigger points:
 1. We probably need to always route subnet responses, or they'll prevent us from halting the subnet and moving on with the merge.
 2. (Something of a personal preference) I wouldn't bother rejecting requests when routing to a cooling down subnet. It's unnecessary, it requires extra logic and still leaves us with all the requests hidden behind a response.

rs/execution_environment/src/execution_environment.rs
mraszyk: > returning an error will cause the ingress message to not be inducted (as opposed to, e.g., simply rejected)

What's the difference between "not inducted" and "simply rejected" here?

rs/replicated_state/src/metadata_state.rs
mraszyk: Rephrased: https://github.com/dfinity/ic/pull/10910/commits/d02f44bf5fb6f7a50f6ff7a3b94bd991ea41daaa

rs/ingress_manager/src/ingress_selector.rs
mraszyk: Indeed: https://github.com/dfinity/ic/pull/10910/commits/800ae1688a0ad96e87e5e98eb0ff13661b6e96a6

rs/protobuf/def/state/metadata/v1/metadata.proto
mraszyk: https://github.com/dfinity/ic/pull/10910/commits/36d4768d8bc3c8b4ac348e52d84174914b5f0a77

rs/execution_environment/src/execution_environment.rs
alin-at-dfinity: I guess I meant that if this function returns an error, then the ingress message is not even inducted by the payload builder as opposed to being rejected later on by the DSM.

In my case, I wasn't expecting code in `rs/execution_environment/src/execution_environment.rs` to have effect on the ingress payload builder.

PR #10911, opened 2026-07-27
chore(core): Bump libflate

michael-weigelt:


schneiderstefan (approved):

rs/tests/execution/large_zipped_wasm_test.rs
```diff
@@ -0,0 +1,111 @@
...
+        let compressed_hash = Sha256::hash(&compressed_wasm);
+
+        // this fails if libflate is on an older version.
+        let _ = mgr
+            .install_code(&canister_id, &compressed_wasm)
+            .call_and_wait()
+            .await;
+
+        // wait to give the node some time to restart if it crashed (so the systest shows the assert_no_metrics_error).
+        std::thread::sleep(Duration::from_secs(30));
```
mraszyk: this looks a bit fragile

basvandijk (approved):

rs/tests/execution/large_zipped_wasm_test.rs
mraszyk: a StateMachine/Execution test would be more reliable to detect a regression since it would panic right away

PR #10920, opened 2026-07-27
chore: use new ic-cdk apis

nmattia:
This migrates our code out of the 0.18-deprecated modules and into the new `ic_cdk::stable` and `api::*` system APIs.

This allows us to drop some explicit casts.

### DIFF
--- a/rs/nns/sns-wasm/canister/canister.rs
+++ b/rs/nns/sns-wasm/canister/canister.rs
@@ -35,7 +35,7 @@ use ic_sns_wasm::{
 };
 use ic_types::CanisterId;
 use ic_types_cycles::Cycles;
-use std::{cell::RefCell, collections::HashMap, convert::TryInto};
+use std::{cell::RefCell, collections::HashMap};
 
 use ic_cdk::{
     api::{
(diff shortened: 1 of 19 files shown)


mraszyk (approved):

blind-oracle (approved):

rs/nns/sns-wasm/canister/canister.rs
```diff
@@ -179,10 +181,10 @@ impl CanisterApi for CanisterApiImpl {
     }
```
daniel-wong-dfinity-org-twin: How difficult would it be to change all the relevant types to u128 (from u64)? I think this is something that AI could be good at.

daniel-wong-dfinity-org-twin, line 181: Even if changing everything to u128 would be a pain, at least we can minimize the local amount of dangerous down casting, like so:


```suggestion
        let available = msg_cycles_available();

        if available < required_cycles as u128 {
            return Err(format!(
                "Message execution requires at least {required_cycles} cycles, but only {available} cycles were sent.",
            ));
        }

        // Convert available to u64 with saturation before returning it.
        let available = available.clamp(0, u64::MAX) as u64;
        Ok(available)
```

Saturation is wrong, but at least, it seems "less" wrong in this case than simply chopping off the top (most significant!) bits.

rs/nns/sns-wasm/src/sns_wasm.rs
```diff
@@ -831,7 +831,7 @@ where
             .map_err(validation_deploy_error)?;
 
         canister_api
-            .this_canister_has_enough_cycles(SNS_CREATION_FEE)
+            .this_canister_has_enough_cycles(u128::from(SNS_CREATION_FEE))
```
daniel-wong-dfinity-org-twin: If you wouldn't mind also changing the type of these constants to u128, that would be swell. It would reduce the churn in this PR, I think.

daniel-wong-dfinity-org-twin (approved):

PR #10928, opened 2026-07-28
chore(testnets): deploy cloud_engine in dm1 with mainnet versions

pietrodimarco-dfinity:
Supersedes/absorbs #10682: this branch was cut from `nim-cloud-testnet`, so its commits are included here and the PR targets `master`.

Keeps #10682's topology (1-node System/NNS subnet, 1-node Application subnet, 20 unassigned nodes across the 20 cloud-provider DCs, 1 API boundary node, 1 ic-gateway, p8s) and its empty `demo` / `whale` canisters. Adds mainnet software versions, restores the public-IPv4 support that #10682 removed, and adds a dispatchable workflow to recreate the testnet.

## Mainnet versions

* `guestos = "mainnet_nns_dev"` boots the GuestOS version that mainnet's NNS subnet runs instead of the tip of the branch. The `_dev` variant is required because this testnet brings up a *fresh* NNS whose public key has to be configured on the nodes, same as `mainnet_nns_recovery` in this file; `mainnet_nns_state` can use the non-dev image only because it restores mainnet's own NNS state.
* NNS canisters already come from mainnet in the `cloud_engine` target, because `system_test_nns` declares the mainnet-NNS variant as `<name>` and the HEAD one as `<name>_head_nns`. Worth knowing: the live testnet is the `_head_nns` variant (Farm group `cloud-engine-head-nns--1783472604708529`), so today it runs NNS canisters built from master. The header comment now says which target to deploy.
* Recreating the testnet is therefore how it tracks mainnet: both versions are resolved at build time from `mainnet-icos-revisions.json` and `mainnet-canister-revisions.json`, so freshness follows whichever branch is built.

## Placement and the public IPv4

`system_test` sets `ALLOCATE_TESTNET_TO_LOCAL_DC=1`, which confines the Farm group to the DC derived from the deploying machine's `NODE_NAME`, and that value cannot be overridden with `--test_env` (I tried; `DC` can, `ALLOCATE_TESTNET_TO_LOCAL_DC` cannot). Group-level and per-VM host features are combined, so a group cannot span two DCs. Hence no per-VM DC pinning: pass `--test_env=DC=dm1` (or deploy from a dm1 machine) and the whole testnet lands in dm1.

`IC_GW_IPV4` is restored from `master` and gives the ic-gateway a static address out of `23.142.184.224/28`, which Farm publishes as the testnet's A record. **Why it matters:** the running testnet publishes `A 10.11.11.215`, an RFC1918 address, next to a globally routed AAAA. It is reachable over IPv6 and from inside the DFINITY network, but not from IPv4-only clients, which is why control-panel's `deploy-frontend` workflow cannot deploy to it from GitHub-hosted runners.

**The public-IPv4 path is not yet usable end to end, and reviewers should know why.** Per `https://farm.dfinity.systems/dc`, `dm1-dmz` is a DC of its own with two hosts (`dm1-spm59`, `dm1-spm60`, 256 vCPUs each, both idle) whose `dmz` feature is *mandatory*, i.e. Farm places a VM there only if the VM requests it. Combined with the two constraints above, that means every VM of the group has to sit in `dm1-dmz` and request `dmz` - including the vector log VM, which is constructed in `rs/tests/driver/src/driver/vector_logging_task.rs` as `VectorVm::new().with_start_time(...)` with no way to pass host features. Options, none taken here: set `logs = False` for this testnet and pin p8s via `prometheus_vm_required_host_features`; teach the driver to pass host features to the vector and p8s VMs; or have Farm expose the DMZ hosts under `dc=dm1` / drop the mandatory flag. The gateway therefore asks for `dc=dm1-dmz` + `dmz` only when `IC_GW_IPV4` is set, so the default path is unaffected.

## Recreate workflow

`.github/workflows/cloud-engine-testnet-recreate.yml`, dispatch-only, on a dm1 `dind-large` runner in the ic-build container. It deploys with `--keepalive --no-group-ttl --no-delete-farm-group`, polls the demo domain until it answers, then kills the driver so the group survives, and only then deletes the previous group - so a failed deployment leaves the old testnet in place rather than none. Groups are selected by `metadata.testName == cloud_engine`, leaving any `cloud_engine_head_nns` group alone.

No `schedule:` on purpose: the consumer of the testnet (a CronJob in cmh1-dashboard-dev) owns the schedule and re-seeds its own canisters and state afterwards, so there is one cron source rather than two racing to recreate the same testnet.

## Verified by deploying this branch

Deployed from a devenv with `--test_env=DC=dm1 --test_env=DEMO_DOMAIN=opencloud --test_env=NUM_UNASSIGNED_NODES=4`:

* Builds: 4430 actions. The BUILD change is accepted by `configure_icos` (an unknown `guestos` value would `fail()` at load time).
* Allocation in `dc=dm1`: zero failures.
* Replicas run `feb8e5f205ad5577b2e11e56a6f94faa2d1c6116`, exactly the mainnet NNS-subnet GuestOS revision recorded in this branch's `mainnet-icos-revisions.json`, fetched as `@mainnet_nns_images_dev//:guest-img`.
* Mainnet NNS canisters installed in 24.9 s, followed by II (backend + frontend), the NNS dapp and the Subnet Rental canister.
* `DEMO_DOMAIN=opencloud` produced `opencloud.demo.farm.dfinity.systems` with a Let's Encrypt certificate, A + AAAA records, `/api/v2/status` returning `replica_health_status: healthy`, and it is reachable from the public internet over IPv6.
* The Application subnet canister ids are reproducible across recreations: two independent deployments both produced `demo = 5v3p4-iyaaa-aaaaa-qaaaa-cai` and `whale = 5s2ji-faaaa-aaaaa-qaaaq-cai`.

**Heads-up for consumers:** `5s2ji-faaaa-aaaaa-qaaaq-cai` is currently pinned as `APP__IC__ENGINE_CANISTER_ID` in cmh1-dashboard-dev, where it is control-panel's engine canister on the existing head-NNS testnet. With this testnet definition the whale canister takes that id, so the value has to be repointed at whatever the engine canister gets once it is installed after `demo` and `whale`. Since the ids are deterministic, pinning stays viable.

Both targets are tagged `manual`, so CI does not build them; the workflow above is how the deployment gets exercised.


pietrodimarco-dfinity:
Deployed from this branch and verified end to end, with `--test_env=DC=dm1-dmz --test_env=IC_GW_IPV4=23.142.184.230 --test_env=DEMO_DOMAIN=opencloud`:

* All VMs allocated on a DMZ host (`dm1-spm60.dm1.dfinity.network`) - replicas, API boundary node, ic-gateway and Prometheus. The same request fails with `No suitable host currently available that satisfies the group's VmAllocation!` without the `dmz` feature on every VM, because Farm marks it mandatory on those hosts.
* Mainnet NNS canisters installed, then II, the NNS dapp and the Subnet Rental canister.
* Farm published `A 23.142.184.230` and `AAAA 2602:fb2b:100:101:5053:e6ff:fed2:3b1a` for `opencloud.demo.farm.dfinity.systems`.
* `/api/v2/status` answers `replica_health_status: healthy` over forced IPv4.

The public IPv4 is what this change is for, so I verified it from the runners that need it. Before the change, both runner types failed to reach the testnet at all; after it:

| runner | IPv6 egress | `/api/v2/status` |
| --- | --- | --- |
| `namespace-profile-control-panel` | none | 200 via 23.142.184.230 |
| `ubuntu-latest` | none | 200 via 23.142.184.230 |

Neither runner has IPv6, which is why the previously published private A record made the testnet unreachable from CI regardless of the globally routed AAAA.

Two incidental findings while testing:

* The driver's per-test timeout of 10 minutes, which `--keepalive` does not lift, killed setup one minute after the gateway VM started. Hence the 90 minute timeout in this PR, following `mainnet_nns_state`.
* Application-subnet canister ids are reproducible: three independent deployments each produced `demo = 5v3p4-iyaaa-aaaaa-qaaaa-cai` and `whale = 5s2ji-faaaa-aaaaa-qaaaq-cai`. Consumers pinning a canister id can keep doing so, but note the whale canister now occupies the id that control-panel currently pins as its engine canister.

rs/tests/testnets/cloud_engine.rs
```diff
@@ -97,275 +140,224 @@ struct DcConfig {
...
 
 pub fn setup(env: TestEnv) {
-    let dm1_dmz_features = vec![HostFeature::DC(DM1_DMZ_DC.to_string()), HostFeature::DMZ];
+    // Every VM of this testnet asks for a DMZ host, because that is the only
+    // place a publicly routed IPv4 can be attached (see IC_GW_IPV4 and the DC
+    // note at the top of this file). `InternetComputer::with_required_host_features`
+    // covers the API boundary nodes but is not inherited by `add_subnet` or
+    // `with_unassigned_node`, so the subnets and the unassigned nodes ask
+    // individually.
+    let dmz = vec![HostFeature::DMZ];
```
basvandijk, line 349: IIRC only the `dm1-dmz` DC has public IPv4 addresses. `zh1-dmz` doesn't.

basvandijk (approved):

rs/tests/testnets/cloud_engine.rs
pietrodimarco-dfinity: Right — corrected the comment: the mandatory `dmz` host feature is only what lets a VM land on those hosts, and pinning the DC to dm1-dmz is what actually makes `IC_GW_IPV4` attachable, since zh1-dmz has no public IPv4 network.

NikolaMilosa (approved):

PR #10930, opened 2026-07-28
chore: migrate ic-cdk usage to new management canister API

nmattia:
Continues the migration to ic-cdk 0.20, which removes the APIs deprecated in 0.18. `api::management_canister` was deprecated in favor of the root-level `management_canister` and `bitcoin_canister` modules.

Couple notes:

* `cloner_canister` used `create_canister(arg, INITIAL_CYCLES_BALANCE)`, which attached exactly that amount. The new API has no "attach exactly N" form, so this uses `create_canister_with_extra_cycles`, which attaches the creation fee *plus* that amount. Created canisters therefore end up with the full 1T rather than 1T minus the fee, which is what the constant's comment describes anyway.

* The management_canister benchmark canister deliberately parameterizes the cycles attached to `http_request` over its Candid interface, but `management_canister::http_request` computes them itself. That one call is issued via `Call::unbounded_wait(...).with_cycles(...)` instead, in keeping with the `Call` usage already in that file.

* `HttpRequestArgs` has a new `is_replicated` field; `None` is set everywhere to keep the existing replicated-outcall behavior.

* The benchmark canister defines its own `InstallCodeArgs`, `UpdateSettingsArgs` and `HttpRequestArgs` Candid types, which now collide with the new names, so the imports extend that file's existing `HttpHeader as IcCdkHttpHeader` aliasing convention rather than renaming the canister's public types.

### DIFF
--- a/rs/pocket_ic_server/tests/test.rs
+++ b/rs/pocket_ic_server/tests/test.rs
@@ -2,8 +2,7 @@
 use crate::common::{send_signal_to_pic, start_server, start_server_helper};
 use candid::{Encode, Principal};
 use ic_agent::agent::CallResponse;
-use ic_cdk::api::management_canister::main::CanisterIdRecord;
-use ic_cdk::api::management_canister::provisional::ProvisionalCreateCanisterWithCyclesArgument;
+use ic_cdk::management_canister::CanisterIdRecord;
 use ic_management_canister_types_private::ProvisionalCreateCanisterWithCyclesArgs;
 use ic_registry_proto_data_provider::ProtoRegistryDataProvider;
 use ic_utils::interfaces::ManagementCanister;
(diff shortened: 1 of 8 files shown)


gregorydemay (approved):

rs/pocket_ic_server/tests/test.rs
```diff
@@ -1091,7 +1090,7 @@ fn provisional_create_canister_with_cycles() {
         .with_application_subnet()
         .build();
 
-    let arg = ProvisionalCreateCanisterWithCyclesArgument::default();
+    let arg = ic_cdk::management_canister::ProvisionalCreateCanisterWithCyclesArgs::default();
```
mraszyk: let's include this in the use-clause above

mraszyk (approved):

daniel-wong-dfinity-org-twin (approved):

PR #10931, opened 2026-07-28
fix: build e2fsprogs with the hermetic cc toolchain

basvandijk:
## Why

`@e2fsprogs//:e2fsprogs_impl` (mke2fs + fuse2fs) opted out of the hermetic zig toolchain via `disable_hermetic_cc_target` and linked `-static`, so the build container's `libc.a` was baked into both binaries — while the host toolchain is **not** part of Bazel's action key.

Every `ic-build` image rebuild that bumps `libc6-dev` therefore breaks the `build-determinism` job: `bazel-test-all` serves a remote cache hit produced under the *previous* image, while `build-ic` (`--config=local`, no remote cache) rebuilds under the *new* one. Most recently on #10579, where a container rebuild moved glibc `2.43-2ubuntu2` → `2.43-2ubuntu2.3`:

```
@@+http_archive+e2fsprogs//:e2fsprogs_impl  .../bin/mke2fs:  f4814316... != 59f9c266...
@@+http_archive+e2fsprogs//:e2fsprogs_impl  .../bin/fuse2fs: 5d244cc0... != f4ca6b16...
```

The cache-side digest was frozen across the image bump (same action key) while the locally-built one moved — the signature of an action-key-invisible host input rather than an embedded build path. It is latent on `master` too: it fires on the next `ci/container/TAG` rebuild that bumps glibc.

## What

Both halves of the comment that justified the opt-out are no longer true:

* **libfuse's meson build works fine under zig** as of `hermetic_cc_toolchain` 4.3.0 / zig 0.15.2, which the base branch (#10890) introduces. `@libfuse//:libfuse` builds clean with `//bazel:hermetic_cc` at its default, using the zig `c++` wrapper — verified in the generated `build_script.sh`.

* **The real blocker was `-static`.** zig ships no static glibc, so `zig cc -target x86_64-linux-gnu.2.31 -static` fails with `error: libc of the specified target requires dynamic linking`, which autotools reports as the opaque `configure: error: C compiler cannot create executables`. Dropping `-static` is what we want regardless: the host `libc.a` *was* the reproducibility bug. Linked against zig's glibc 2.31 stubs the binaries require only glibc ≥ 2.28 at runtime, exactly like `dosfstools` and `mtools`.

`disable_hermetic_cc_target` had no other users, so the rule and its transition go too. The `//bazel:hermetic_cc` flag stays — `build:afl --hermetic_cc=false` still uses it.

## Verification

Built in **both container images** — the exact pair that broke CI — each with its own output base at a different path, so this varies both the container glibc and the build path:

| build environment | mke2fs (exec config) | fuse2fs |
| --- | --- | --- |
| host, glibc `2.43-2ubuntu2.3` | `ca440c474f8d…` | `47aad05981…` |
| `ic-dev` old image, glibc `2.43-2ubuntu2` | `ca440c474f8d…` | `47aad05981…` |

Byte-identical. The exec configuration is the one `build-determinism` compares.

Reverting the change and rebuilding on the same host reproduces CI's failing digest exactly — `59f9c266373e64d064b22c9ccaa0dc1bbb03b153e90555d0a21ded9945cf9d3a`, the value `build-ic` produced in run 30307561001 — which confirms the host `libc.a` was the only variable.

Also checked:

* Both binaries run (`mke2fs 1.47.4`, `fuse2fs 1.47.4`) and contain no embedded build paths (no `sandbox`, `bazel-out`, or `build_tmpdir` strings).
* `//ic-os/guestos:partition-config.tzst` is unchanged at `2a45cc26…`, so the filesystem images these tools produce are unaffected.
* `bazel build //... --nobuild` and `//:buildifier` are clean.

Labelled `CI_ALL_BAZEL_TARGETS` so the full target set — including every `ext4_image` consumer — is exercised.

## Note

`fuse2fs` execs `/bin/fusermount3` at runtime (a consequence of `bindir: /bin` plus `utils: false` in the libfuse build, so `fusermount3` is never built) and picks up the host's copy. That is pre-existing and untouched here, but it is a remaining hermeticity gap in the same corner.


third_party/BUILD.e2fsprogs.bazel
```diff
@@ -35,9 +35,17 @@ configure_make(
...
         "--with-systemd-unit-dir=no",
     ],
+    # No -static here. zig has no static glibc to offer ("error: libc of the
+    # specified target requires dynamic linking"), and linking the *host's*
+    # libc.a is exactly what we want to get away from: it baked the build
+    # container's glibc into these binaries without being part of Bazel's action
+    # key, so one remote cache entry outlived every container image bump and
+    # broke the build-determinism check. Linked against zig's glibc 2.31 stubs
+    # the binaries end up requiring only glibc >= 2.28 at runtime, and they are
+    # build-time tools; dosfstools and mtools are built the same way.
```
nmattia: this can be shortened a bit. More importantly, do you know if e2fsprogs links in more than just libc? if so, we may still want to statically link those (or not, unsure)

third_party/BUILD.e2fsprogs.bazel
```diff
@@ -71,24 +79,16 @@ configure_make(
...
-# glibc version mismatch. So just disable the hermetic toolchain for e2fsprogs.
-disable_hermetic_cc_target(
-    name = "e2fsprogs",
-    target = ":e2fsprogs_impl",
-)
-
 filegroup(
     name = "fuse2fs",
-    srcs = [":e2fsprogs"],
+    srcs = [":e2fsprogs_impl"],
```
nmattia: best to simply rename the `configure_name` target

third_party/BUILD.e2fsprogs.bazel
basvandijk: I've removed the whole comment since it was talking about the old situation.

Regarding linking, Claude checked the built binaries — it's glibc only.

`readelf -d` on both outputs:

| binary | `NEEDED` |
|---|---|
| `fuse2fs` | `libpthread.so.0`, `libc.so.6`, `libdl.so.2` |
| `mke2fs` | `libpthread.so.0`, `libc.so.6` |

All three of those *are* glibc — `libpthread`/`libdl` have been empty forwarding stubs since glibc 2.34, with the symbols living in `libc.so.6`. Every undefined symbol in either binary is versioned against a `GLIBC_*` tag; there isn't a single non-glibc `UND` symbol. No `RPATH`/`RUNPATH` either.

Everything else is already folded in statically, so there's nothing left to static-link:

- **libfuse3** — 269 `fuse_*` symbols defined locally, i.e. `@libfuse`'s `libfuse3.a` is baked in. No `libfuse3.so.3`.
- **e2fsprogs' own libs** — 568 locally-defined `ext2fs_*` / `com_err` / `uuid_*` / `blkid_*` symbols, so `--disable-elf-shlibs` plus `--enable-libuuid`/`--enable-libblkid` are doing their job. No host `libuuid.so.1` / `libblkid.so.1`.
- Not linked at all: `libselinux`, `libmagic`, `libz`, `libm`, `libiconv` (iconv is in glibc).

Two things worth recording:

1. **glibc floor** is `GLIBC_2.25` for `mke2fs` and `GLIBC_2.28` for `fuse2fs` — set by the zig toolchain's glibc target, not by e2fsprogs.
2. **`fuse2fs` imports `dlopen`/`dlsym`/`dlclose`** for libfuse's optional module loader (`-o modules=…`). Not a link-time dependency, and we never pass that option, but it's the one path by which a host `.so` could enter at runtime.

third_party/BUILD.e2fsprogs.bazel
basvandijk: Done.

nmattia (approved):

PR #10933, opened 2026-07-28
feat(subnet-splitting): add subnet splitting status field to summary

pierugo-dfinity:
To prepare for subnet splitting, this PR adds a new field to summary blocks, empty for the moment for backwards compatibility purposes. This field indicates the status of a potential subnet split. 
- In most cases, this will be `NotScheduled`.
- The summary at which the subnet split happens (the summary height following the registry change, call it `H`) will have `Scheduled` together with the two new subnet IDs (the source being the current one, and the destination which will be taken [from the registry](https://sourcegraph.com/r/github.com/dfinity/ic@0f3c60f6a7ba51e3306c2305072a2959b00718ef/-/blob/rs/registry/canister/src/mutations/do_split_subnet.rs?L170)).
- At that point, replicas will skip CUP for height `H` and will instead create two CUPs for `H + 500` (one for each subnet) each containing a new summary block created on the fly that contain the respective DKG transcripts created by the NNS and will have status `PostSplit` with the respective `new_subnet_id` they are part of (derived by looking at the registry).

### DIFF
--- a/rs/types/types/src/consensus/dkg.rs
+++ b/rs/types/types/src/consensus/dkg.rs
@@ -792,6 +792,7 @@ pub enum DkgPayloadCreationError {
 }
 
 /// Reasons for why a dkg payload might be invalid.
+#[allow(clippy::large_enum_variant)]
 #[derive(PartialEq, Debug)]
 pub enum InvalidDkgPayloadReason {
     CryptoError(CryptoError),
--- a/rs/types/types/src/consensus/dkg.rs
+++ b/rs/types/types/src/consensus/dkg.rs
@@ -241,7 +241,6 @@ pub enum SubnetSplittingStatus {
     #[default]
     NotScheduled,
     /// The subnet is requested to be split at the height of the summary block.
-    /// Contains all the information necessary to determine the new subnet of the replica
     Scheduled(SplittingArgs),
     /// The subnet was split at the previous summary block.
     PostSplit(PostSplitArgs),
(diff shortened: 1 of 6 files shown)


rs/types/types/src/consensus/dkg.rs
```diff
@@ -690,7 +796,7 @@ pub enum DkgPayloadCreationError {
 pub enum InvalidDkgPayloadReason {
     CryptoError(CryptoError),
     DkgVerifyDealingError(DkgVerifyDealingError),
-    MismatchedDkgSummary(DkgSummary, DkgSummary),
+    MismatchedDkgSummary(Box<DkgSummary>, Box<DkgSummary>),
```
pierugo-dfinity, line 798: Now required to avoid Clippy complaining about the size of the enum variant being too large

rs/types/types/src/consensus/dkg.rs
```diff
@@ -215,6 +216,37 @@ impl std::hash::Hash for RemoteDkgAttempts {
...
+
+/// Represents the status of subnet splitting at the given summary height.
+#[derive(Copy, Clone, Serialize, Deserialize, Eq, PartialEq, Hash, Debug, Default)]
+#[cfg_attr(test, derive(ExhaustiveSet))]
+pub enum SubnetSplittingStatus {
+    /// The subnet hasn't been requested to be split.
+    #[default]
+    NotScheduled,
+    /// The subnet is requested to be split at the height of the summary block.
+    /// Contains all the information necessary to determine the new subnet of the replica
```
mraszyk: It is just a pair of subnet IDs, I guess more information is needed coming from somewhere else.

mraszyk (approved):

rs/types/types/src/consensus/dkg.rs
pierugo-dfinity: True, I think this comment line doesn't add much value anyways

PR #10936, opened 2026-07-28
feat(subnet-splitting): implement subnet splitting's halting logic

pierugo-dfinity:
This PR implements the halting logic of subnet splitting by adapting Consensus' `get_status` function, used by block making, block validation, and batch delivery.

On a `Scheduled` subnet split summary, replicas will halt as they are about to skip the whole interval and create CUPs for the next summary height directly.

On a `PostSplit` subnet split summary, the split is done. As a replica's subnet ID is determined at startup, the source subnet will have it correct and can continue creating blocks. Though for a short amount of time, the destination subnet will still have the configured source subnet ID while the summary indicates the destination subnet ID. In that case, do not create new blocks (i.e. halt) until the orchestrator observes this `PostSplit` CUP and restarts the replica (with the correct destination subnet ID). After they restarted, replicas' subnet ID will match the one in the summary and should create blocks.

Note: as the two subnets are still connected under the same P2P network after the split but have different halting conditions, it is expected to receive artifact invalidation during that time: the destination subnet will receive non-empty blocks from the source subnet.
Note 2: again, because the two subnets are still connected under the same P2P network after the split, the source subnet will broadcast certifications/certification shares for heights above the post-split summary. These should be ignored by the destination subnet: coming later in a [separate PR](https://github.com/dfinity/ic/pull/10970).

P.S.: the PR also finds the last summary block at one single place and passes it around during block making and validation, allowing to remove some error variants and avoid panics.


rs/consensus/src/consensus/status.rs
```diff
@@ -64,25 +86,63 @@ pub(crate) fn should_halt(
...
+            .subnet_splitting_status()
+        {
+            SubnetSplittingStatus::NotScheduled => false,
+            SubnetSplittingStatus::Scheduled(..) => true,
+            // After the split, don't produce any blocks until we are on the right subnet.
+            SubnetSplittingStatus::PostSplit(PostSplitArgs { new_subnet_id }) => {
+                subnet_id != new_subnet_id
+            }
+        }
+    });
```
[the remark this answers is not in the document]
pierugo-dfinity: I would accept the risk, this isn't the only place we are doing this, and a regression should be easily detected through the replica crashing with all the system tests we have.

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -190,6 +190,11 @@ impl BlockMaker {
         let height = parent.height().increment();
         let certified_height = self.state_reader.latest_certified_height();
 
+        let Some(last_summary_block) = pool.dkg_summary_block(parent.as_ref()) else {
+            warn!(self.log, "Couldn't find the summary block");
+            return None;
+        };
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10936/commits/bba8fec86e7e7c5eb7c1d948b6ad51dd664a8569

rs/consensus/src/consensus/status.rs
```diff
@@ -29,18 +35,33 @@ pub(crate) enum Status {
 ///   halt.
 pub(crate) fn get_status(
     height: Height,
+    last_summary_block: &Block,
     registry_client: &(impl RegistryClient + ?Sized),
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10936/commits/c9e6b6ca928309381f8749797e321447428bfb6f

rs/consensus/src/consensus/status.rs
```diff
@@ -64,25 +87,63 @@ pub(crate) fn should_halt(
...
+    let should_halt_due_to_subnet_splitting = last_summary_block.map(|summary_block| {
+        match summary_block
+            .payload
+            .as_ref()
+            .as_summary()
+            .dkg
+            .subnet_splitting_status()
+        {
+            SubnetSplittingStatus::NotScheduled => false,
+            SubnetSplittingStatus::Scheduled(..) => true,
```
pierugo-dfinity: Sorry, this should actually probably check `height >= last_summary_block.height` to ensure the summary block is finalized.

pierugo-dfinity: https://github.com/dfinity/ic/pull/10936/commits/4726a581b51bede1d0081bb54607280610758362

rs/consensus/src/consensus/validator.rs
eichhorl: I guess this was already wrong before, but maybe
```suggestion
    NonEmptyPayloadWhileHalting,
```

rs/consensus/src/consensus/status.rs
```diff
@@ -64,25 +87,63 @@ pub(crate) fn should_halt(
...
         .ok()
         .flatten()
         .warn_if_none(
             logger,
             format!(
-                "Failed to check if the registry version at height {height} instructs the subnet to halt!",
+                "Failed to check if the registry version at height {height} \
+                instructs the subnet to halt!",
             ),
         );
```
eichhorl: De we need both logs?

rs/consensus/src/consensus/notary.rs
```diff
@@ -349,6 +349,7 @@ fn get_adjusted_notary_delay_from_settings(
     let halting = || {
         status::should_halt(
             notarized_height,
+            None,
```
eichhorl, line 352: Doesn't passing `None` here mean `should_halt` will never return `Some(false)`? Should it return `Some(true)` so that the summary may be finalized faster?

rs/consensus/src/consensus/status.rs
```diff
@@ -135,139 +205,150 @@ mod tests {
...
+                        .with_dkg_interval_length(DKG_LENGTH)
                         .with_replica_version(replica_version.as_ref())
                         .with_halt_at_cup_height(halt_at_cup_height)
                         .build(),
                 ),
             ],
         );
 
-        pool.advance_round_normal_operation_n(10);
+        pool.advance_round_normal_operation_no_cup_n(CUP_HEIGHT.get());
```
eichhorl: Is it intended that we produce *no* cup at `CUP_HEIGHT`?

rs/consensus/src/consensus/status.rs
```diff
@@ -23,24 +29,40 @@ pub(crate) enum Status {
 ///
 /// Returns
 /// * [Status::Halting] when there is a pending upgrade or the registry instructs the subnet to halt
-/// * [Status::Halted] when a CUP height has been finalized and either an upgrade is in progress or
-///   the registry instructs the subnet to halt;
-/// * [Status::Running] when there is no upgrade and the registry doesn't instruct the subnet to
-///   halt.
+/// * [Status::Halted] when the subnet is splitting or when the validation context's certified
+///   height reached the CUP height and either an upgrade is in progress or the registry instructs
+///   the subnet to halt;
```
eichhorl: It seems we already return `Halting` if a split is scheduled, and `Halted` once the scheduled split is certified

rs/consensus/src/consensus/status.rs
```diff
@@ -135,139 +205,150 @@ mod tests {
...
+        subnet_id: SUBNET_0,
+        expected_status: Some(Status::Halting),
+    })]
+    #[case::subnet_splitting_finalized(TestCase{
+        certified_height: CUP_HEIGHT,
+        current_height: CUP_HEIGHT,
+        replica_version: ReplicaVersion::default(),
+        halt_at_cup_height: false,
+        subnet_splitting_status: Some(SubnetSplittingStatus::Scheduled(SplittingArgs { destination_subnet_id: SUBNET_0, source_subnet_id: SUBNET_1 })),
+        subnet_id: SUBNET_0,
```
eichhorl: Shouldn't `SUBNET_0` be the source subnet? Since that's the one we create a subnet record for

eichhorl (approved):

rs/consensus/src/consensus/validator.rs
pierugo-dfinity: Yep thanks https://github.com/dfinity/ic/pull/10936/commits/c9be5e97fdbc98ba07631952bd0922591142d7b1

rs/consensus/src/consensus/status.rs
pierugo-dfinity: True, removed the added log https://github.com/dfinity/ic/pull/10936/commits/d26b6aa8d13285aa083e8e0864daacefcf21c3b6

rs/consensus/src/consensus/notary.rs
pierugo-dfinity: I think that here, there is no easy way to find the summary block of interest. If the splitting summary is not finalized yet, then we do not know what block at that height to choose from (by definition we still haven't reached consensus on it). If it is finalized, then it is already too late to adapt the delay anyways.
The difference with upgrades (and more generally looking for the registry version) is that we can predict the registry version of the next interval by looking at the current summary's validation context. There is no such thing for the splitting status, who is only known for the current interval.

If what you actually suggested was to shortcut and return `Some(true)` whenever the passed `last_summary_block` is `None`, I think that would be fragile to future uses of the function.

rs/consensus/src/consensus/status.rs
pierugo-dfinity: Thanks. This was required for `case::post_subnet_splitting_old_subnet_id` as it would otherwise return `Halting` because the certified height is lower than the highest CUP ("Note: If 'height' is smaller than the height of the last CUP, this will return [None]."). Though in practice, the post-split summary will be created with a certified height equal to itself, so there is no real value in testing with a certified height lower than the summary height. Adapted in https://github.com/dfinity/ic/pull/10936/commits/0a43ceead04eeae4c5ba62932a9f95f9f08c28ac

rs/consensus/src/consensus/status.rs
pierugo-dfinity: Thanks, I forgot to adapt the docs after https://github.com/dfinity/ic/commit/4726a581b51bede1d0081bb54607280610758362: https://github.com/dfinity/ic/pull/10936/commits/5ceb053926bd73406b65d3489dd316466c360484

rs/consensus/src/consensus/status.rs
pierugo-dfinity: True, thanks! https://github.com/dfinity/ic/pull/10936/commits/fb258f50ca672bb57e583bf9f2a19915835d3326

rs/consensus/src/consensus/notary.rs
eichhorl: Yes, we would have to find the summary in the notarized parent chain. But we can also leave it as is for now, and see how much we slow down in practice

eichhorl (approved):

PR #10938, opened 2026-07-28
feat(subnet-splitting): freeze registry version until summary is reached

pierugo-dfinity:
Because Consensus passes the block’s validation context's registry version to other components (including DSM), it could happen that a subnet split has not yet happened from Consensus' point of view but it would have from the other components' point of view. While it does not necessarily have to lead to misbehavior, this PR proposes to "freeze" the registry version in the validation context from the moment a split is detected in the registry until the next summary block (which will have a `Scheduled` status).

More precisely, we will not bump the registry version in the validation context of any data block in a DKG interval to any version of the registry which has a flag indicating that the subnet splitting should happen. We only bump it when reaching the summary block (to exactly the registry version the split was scheduled at, not a more recent one).


rs/consensus/src/consensus/block_maker.rs
```diff
@@ -521,20 +523,71 @@ impl BlockMaker {
         }
     }
 
-    // Returns the registry version received from the NNS some specified amount of
-    // time ago. If the parent's context references higher version which is already
-    // available locally, we use that version.
-    pub(crate) fn get_stable_registry_version(&self, parent: &Block) -> Option<RegistryVersion> {
+    /// Returns the registry version received from the NNS some specified amount of
+    /// time ago. If the parent's context references higher version which is already
+    /// available locally, we use that version.
```
eichhorl: Maybe we can update the documentation too

rs/consensus/utils/src/subnet_splitting.rs
```diff
@@ -0,0 +1,174 @@
...
+            ic_protobuf::registry::subnet::v1::SubnetSplittingArgs {
+                destination_subnet_id: Some(subnet_id_into_protobuf(DESTINATION_SUBNET_ID)),
+            },
+        )));
+
+        let status = get_status(
+            registry.as_ref(),
+            SOURCE_SUBNET_ID,
+            REGISTRY_CUP_REGISTRY_VERSION.decrement(),
+            REGISTRY_CUP_REGISTRY_VERSION,
```
eichhorl: Maybe a test where the looked up version is above the scheduled at version
```suggestion
            REGISTRY_CUP_REGISTRY_VERSION.increment(),
```

rs/consensus/src/consensus/validator.rs
```diff
@@ -1271,6 +1276,29 @@ impl Validator {
...
+            .map_err(ValidationFailure::SubnetSplittingStatusError)?
+            {
+                subnet_splitting::Status::NotScheduled => {}
+                subnet_splitting::Status::Scheduled { .. } => {
+                    return Err(
+                        InvalidArtifactReason::RegistryVersionNotFrozenDuringSubnetSplitting {
+                            context_registry_version: proposal.context.registry_version,
+                        }
+                        .into(),
+                    );
```
eichhorl, line 1319: Should we add a test for this?

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -521,20 +523,71 @@ impl BlockMaker {
...
+                last_summary_block_registry_version,
+                version,
+            )
+            .inspect_err(|err| {
+                warn!(
+                    self.log,
+                    "Failed to get subnet splitting status at registry version {version}: {err}"
+                )
+            })
+            .ok()?;
```
eichhorl: Note that returning `None` means we won't propose a block. Instead, would it make sense to try a different version, or continue with the parent version?

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -521,20 +523,71 @@ impl BlockMaker {
...
+
+            match subnet_splitting_status {
+                subnet_splitting::Status::NotScheduled => return Some(version),
+                subnet_splitting::Status::Scheduled { scheduled_at, .. } => {
```
eichhorl: Could it happen that `scheduled_at` is smaller than `parents_version`? In that case we would decrease the version

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -1366,4 +1434,151 @@ mod tests {
...
+            expected_stable_registry_version: RegistryVersion::new(MAX_REGISTRY_VERSION),
+        })]
+        #[case::version_frozen_before_splitting(TestCase {
+            splitting_registry_version: Some(RegistryVersion::new(MAX_REGISTRY_VERSION - 1)),
+            last_summary_block_registry_version: RegistryVersion::new(1),
+            next_summary_block_height: Height::new(4),
+            parent_height: Height::new(1),
+            expected_stable_registry_version: RegistryVersion::new(MAX_REGISTRY_VERSION - 2),
+        })]
+        #[case::version_frozen_before_splitting(TestCase {
```
eichhorl: Is it fine that the cases have the same name?

rs/consensus/utils/src/subnet_splitting.rs
```diff
@@ -0,0 +1,174 @@
...
+pub enum StatusError {
+    #[error("Error while getting CatchUpContents at registry version {0}: {1:?}")]
+    FailedToGetCatchUpContents(RegistryVersion, RegistryClientError),
+    #[error("CatchUpContents not found at registry version: {0}")]
+    CatchUpContentsMissingInRegistry(RegistryVersion),
+    #[error("Failed to deserialize CatchUpContents: {0}")]
+    CatchUpContentsDeserializationError(ProxyDecodeError),
+}
+
+pub fn get_status(
```
eichhorl, line 52: We could also add some documentation here

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -1366,4 +1434,151 @@ mod tests {
...
+            splitting_registry_version: Some(RegistryVersion::new(MAX_REGISTRY_VERSION - 2)),
+            last_summary_block_registry_version: RegistryVersion::new(1),
+            next_summary_block_height: Height::new(4),
+            parent_height: Height::new(1),
+            expected_stable_registry_version: RegistryVersion::new(MAX_REGISTRY_VERSION - 3),
+        })]
+        #[case::exact_version_during_splitting(TestCase {
+            splitting_registry_version: Some(RegistryVersion::new(MAX_REGISTRY_VERSION - 1)),
+            last_summary_block_registry_version: RegistryVersion::new(1),
```
eichhorl, line 1550: It seems that this is the same across all tests. We could try different versions for that and also the parent version

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -521,20 +523,71 @@ impl BlockMaker {
...
+                        self.log,
+                        "Subnet splitting scheduled at registry version {scheduled_at} \
+                        and height {next_summary_block_height}. Freezing registry version."
+                    );
+
+                    // Bump the registry version to the scheduled version of the split right when
+                    // we reach the height of the next summary block.
+                    if parents_height.increment() == next_summary_block_height {
+                        return Some(scheduled_at);
+                    }
```
eichhorl, line 583: The PR description says we bump the registry version "to exactly the registry version the split was scheduled at, not a more recent one", but technically this isn't enforced on the validator side (or wouldn't be the case if we implement the fallback above instead of stalling the subnet if the status can't be determined). Is that a problem?

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: Indeed thanks https://github.com/dfinity/ic/pull/10938/commits/70f6dead4e98e98e368823612eb3b0e11ca688fd

rs/consensus/utils/src/subnet_splitting.rs
pierugo-dfinity: Thanks, added all 9 possible combinations on the orders of  `last_summary_block_registry_version`, `scheduled_at` and `looked_up_registry_version` https://github.com/dfinity/ic/pull/10938/commits/a4ad3c49f8274e7cdddebd484ce49f130b02ee13

rs/consensus/src/consensus/validator.rs
pierugo-dfinity: Of course thanks https://github.com/dfinity/ic/pull/10938/commits/878b8c049f59471c0811a58d8b5183e4de280efc

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: Thanks that's a good point that I hadn't thought about. Reusing the parent version makes sense, but I'd even rather try any version above it that doesn't error out. We would then predictively stay stuck on the last registry version that doesn't fail (see latest changes).

Or do you think reusing the parent's version is better?

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: Let's imagine we are in this scenario where `parents_height` is precisely one below the next summary block and assume there is a subnet splitting CUP with `scheduled_at < parents_version`. The parent is a data block. It was obviously validated, meaning that `subnet_splitting::get_status(..., parents_version)` returned `NotScheduled` (otherwise validators would have rejected it with `InvalidArtifactReason::RegistryVersionNotFrozenDuringSubnetSplitting` or `ValidationFailure::SubnetSplittingStatusError`). The only way `get_status` can return `NotScheduled` if `scheduled_at < parents_version` is if `last_summary_block_registry_version >= scheduled_at`. But then, because the parent and the current block have the same last summary block, `subnet_splitting::get_status(..., version)` will also return `NotScheduled`, i.e. we will not enter this branch in the first place.

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: I renamed them as part of adding more tests/parameters, PTAL

rs/consensus/utils/src/subnet_splitting.rs
pierugo-dfinity: True, done https://github.com/dfinity/ic/pull/10938/commits/790e588a376523917c68b401dc496353efdf9aaf

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: Good point, I added some more tests PTAL

rs/consensus/src/consensus/block_maker.rs
pierugo-dfinity: That's true, I just added the validation logic, thank you for the catch. https://github.com/dfinity/ic/pull/10938/commits/ce11cd8b6d7a9e76e777221449968005d135caa9

The status can be undetermined only if either the blockmaker is dishonest or the failure is not replicated (the blockmaker knows the status but not us), so it is correct not to validate the block in both cases.

Note that if at the summary height, the registry version is smaller than the `scheduled_at`, then the block would still be validated (we would enter the `NotScheduled` arm) because maybe the blockmaker is honest but just does not have the splitting registry version locally and had to include a lower version. It could also be a malicious blockmaker that decides not to bump the registry version. In those cases, the subnet split would be delayed by one (or more) DKG intervals until an honest blockmaker includes the splitting registry version at a summary height. I do not think this is a problem, except that the subnet will stay stuck with the same registry version for some exceptionally long time.

rs/consensus/src/consensus/block_maker.rs
eichhorl: I think either is fine, but after https://github.com/dfinity/ic/pull/10938/commits/ce11cd8b6d7a9e76e777221449968005d135caa9, wouldn't such a summary block be rejected anyway?

pierugo-dfinity: I assume you wanted to reply to [this](https://github.com/dfinity/ic/pull/10938/#discussion_r3721665012). If the registry version in the proposed summary is lower than `scheduled_at`, then the block would still be validated because validators will not see the splitting CUP at that registry version.
Some more discussion points [here](https://github.com/dfinity/ic/pull/10938/#discussion_r3721778746)

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -521,20 +523,78 @@ impl BlockMaker {
...
+                    // Until then, we continue iterating back until finding a registry version that
+                    // *does not* contain the scheduled split. I.e., we "freeze" the registry
+                    // version at the last version that does not contain the split, until we reach
+                    // the next summary block.
+                    info!(
+                        every_n_seconds => 5,
+                        self.log,
+                        "Subnet splitting scheduled at registry version {scheduled_at} \
+                        and height {next_summary_block_height}. Freezing registry version."
+                    );
```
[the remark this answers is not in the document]
pierugo-dfinity: According to the [reference](https://doc.rust-lang.org/reference/expressions/literal-expr.html#string-continuation-escapes), omitting the back-slash would include the indentation. The back-slash is present precisely to avoid including it.

rs/consensus/src/consensus/validator.rs
```diff
@@ -3151,6 +3207,274 @@ pub mod test {
...
+                .get_mut()
+                .expect_latest_certified_height()
+                .return_const(context.certified_height);
+            time_source.set_time(context.time).unwrap();
+
+            let result = validator.check_block_validity(&PoolReader::new(&pool), &test_block);
+            if block_registry_version < SUBNET_SPLIT_REGISTRY_VERSION {
+                assert_matches!(result, Ok(()));
+            } else {
+                assert_matches!(
```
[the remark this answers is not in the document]
pierugo-dfinity: Indeed https://github.com/dfinity/ic/pull/10938/commits/1f28132c538c69632e4bd551f63257d490befd01

rs/consensus/src/consensus/validator.rs
```diff
@@ -3151,6 +3207,274 @@ pub mod test {
...
+                    )) if context_registry_version == context.registry_version
+                        && expected_registry_version == SUBNET_SPLIT_REGISTRY_VERSION
+                );
+            }
+        })
+    }
+
+    /// When the subnet splitting status cannot be determined, blocks that don't move the registry
+    /// version forward are still validated — letting the subnet keep running with a stuck registry
+    /// version — while blocks bumping it are held back until the status becomes available again.
```
eichhorl: This doesn't seem to match what the test does

rs/consensus/src/consensus/validator.rs
```diff
@@ -3151,6 +3207,274 @@ pub mod test {
...
+
+            if block_registry_version < SUBNET_SPLIT_REGISTRY_VERSION {
+                schedule_subnet_split(
+                    &data_provider,
+                    &registry_client,
+                    replica_config.subnet_id,
+                    &committee,
+                    SUBNET_SPLIT_REGISTRY_VERSION,
+                );
+            }
```
eichhorl: Does the summary actually contain `SUBNET_SPLIT_REGISTRY_VERSION.decrement()` in the `below_the_scheduled_version` case? IIUC, a record for the `block_registry_version` is only created once we call `schedule_subnet_split`, which happens after we already created the summary

rs/consensus/src/consensus/block_maker.rs
eichhorl: Yes, sorry that was the wrong thread

rs/consensus/src/consensus/block_maker.rs
eichhorl: Makes sense, thanks for explaining!

eichhorl (approved):

rs/consensus/src/consensus/validator.rs
pierugo-dfinity: Thank you, fixed https://github.com/dfinity/ic/pull/10938/commits/ecbb6068c31c0c0d152e63b2cbab6e1e32dddb6e

rs/consensus/src/consensus/validator.rs
pierugo-dfinity: Correct, now fixed, thank you! https://github.com/dfinity/ic/pull/10938/commits/d711a48dd83b4126b1d40c99775f74084ba78e87

PR #10940, opened 2026-07-28
chore: Delete the blessed version list from the registry

Bownairo:
Create a new migration that will remove the blessed version list record, if it exists.

### DIFF
--- a/rs/registry/canister/src/registry_lifecycle.rs
+++ b/rs/registry/canister/src/registry_lifecycle.rs
@@ -8,7 +8,7 @@ use ic_protobuf::registry::subnet::v1::SubnetListRecord;
 use ic_registry_keys::{
     make_node_operator_record_key, make_node_record_key, make_subnet_list_record_key,
 };
-use ic_registry_transport::{pb::v1::RegistryMutation, update};
+use ic_registry_transport::{delete, pb::v1::RegistryMutation, update};
 use ic_types::NodeId;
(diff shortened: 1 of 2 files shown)


rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -366,6 +372,22 @@ fn convert_type1dot1_nodes_to_type4dot5(registry: &Registry) -> Vec<RegistryMuta
...
+/// The blessed version list record is no longer used.
+/// If it still exists, delete it.
+fn remove_blessed_version_list_record(registry: &Registry) -> Vec<RegistryMutation> {
+    if registry
+        .get(
+            "blessed_replica_versions".as_bytes(),
+            registry.latest_version(),
+        )
+        .is_some()
+    {
```
daniel-wong-dfinity-org-twin, line 384: I really disagree with multi-line `if` conditions. This is usually solved by introducing a dummy variable. E.g.

```
let mut result = vec![];

let is_present = registry
    .get(...)
    .is_some();
if is_present {
    result.append(delete(...));
}

result
```

This feedback is not that important though, since this code is temporary.

daniel-wong-dfinity-org-twin (commented):
Please, add something to rs/registry/canister/unreleased_changelog.md under "Removed".

daniel-wong-dfinity-org-twin (approved):

PR #10941, opened 2026-07-28
chore: Remove migration and remnants of blessed versions

Bownairo:
Now that the record has been removed with https://github.com/dfinity/ic/pull/10940, drop the migration and the record type itself.

---

Long ago, we stopped requiring an extra proposal to add versions to the `blessed_replica_versions` list. The code continued to depend on this record for much longer, until the `eero/blessed-versions-*` saga of PRs moved everything to use `replica_version_*` records directly, instead. The record is now gone, and the last remaining reference can be removed.


daniel-wong-dfinity-org-twin (approved):
🙏 

You might want to say a bit more in the PR description. E.g. link to PR where the migration was added, and also mention what replaces this.

PR #10942, opened 2026-07-28
feat: Add a `version_id` field to `ReplicaVersionRecords`

Bownairo:
Currently the ID only exists as the key in the registry, which is nice for duplication, but not great for ergonomics. Passing around a `ReplicaVersionRecord` usually means also extracting the string ID from the registry key, and passing both around as a `(String, ReplicaVersionRecord)`.

With this we'll match `HostosVersionRecord`, and be able to simplify code around this type.


rs/protobuf/def/registry/replica_version/v1/replica_version.proto
```diff
@@ -13,6 +13,9 @@ message ReplicaVersionRecord {
   // The SEV-SNP measurements that belong to this release
   GuestLaunchMeasurements guest_launch_measurements = 9;
 
+  // The ID used to reference this version.
```
daniel-wong-dfinity-org-twin: Please, see [this].

[this]: https://sourcegraph.com/r/github.com/dfinity/ic@fc79106bd7662b690ccfce113ce66008fb17eb0a/-/blob/rs/protobuf/def/registry/subnet/v1/subnet.proto?L23-38

In general, (doc) comments should not be vacuous.

ofc, all the stuff about blank `replica_version_id` in `SubnetRecord` does not apply here. Still, that shows how to practice "no vacuous (doc) comments". Even if it is technically allowed that any string can be used as the ID of a Replica version, if in practice, they generally have some properties, then, this would be the place to document them.

rs/protobuf/def/registry/replica_version/v1/replica_version.proto
```diff
@@ -13,6 +13,9 @@ message ReplicaVersionRecord {
...
+  optional string version_id = 10;
```
daniel-wong-dfinity-org-twin: Either

```suggestion
  optional string id = 10;
```

or

```suggestion
  optional string replica_version_id = 10;
```

I prefer the latter, because we have fields elsewhere named `replica_version_id`. When names matches exactly, people be like, "wait, this cannot be a coincidence; they must correspond to each other", and that is a correct inference. Therefore, naming this `replica_version_id` communicates a highly salient fact that people might otherwise miss. When one datum refers to another, that is a super key fact in your design, so you want that fact to be as obvious as possible.

Ofc, the downside of the latter option is that it is a bit stutter-y/redundant. You could argue that it is "obvious" that `replica_version_id` corresponds to `ReplicaVersionRecord.id`, but for that to really work, we would need this to be an established pattern, and AFAIK, it isn't.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -584,6 +585,7 @@ mod tests {
 
         let key = make_replica_version_key(ReplicaVersion::default());
         let value = ReplicaVersionRecord {
+            version_id: Some(ReplicaVersion::default().to_string()),
```
daniel-wong-dfinity-org-twin: Not asking for change.

It really does not make sense for ReplicaVersion to have a default, but this issue is tangential to this PR. It would be like if you don't know someone's name, but you just walk up to them and be like, "Hi, John. What's your name?". Such a dialogue would be straight out of Monty Python!

Anyway, what I would have done in this file is something like

```
// within mod tests...

lazy_static! {
    static ref REPLICA_VERSION_ID: ReplicaVersion =
        ReplicaVersion::from_str("ee61d50749fa41e472400951ff6e75b84b5b1d14").unwrap();
}
```

(or, as is done elsewhere, "version".)

Yes, you need some arbitrary value for tests, but that is NOT what `Default` is for! `default` should return some predictable distinguished base (i.e. "falsy") value, like `false`, `0.0`, a collection of length 0. But we probably do not allow Replica version IDs to be of length 0 (or at least, we shouldn't allow it).

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
daniel-wong-dfinity-org-twin, line 1: This seems tangential, but ok.

rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -366,6 +375,27 @@ fn convert_type1dot1_nodes_to_type4dot5(registry: &Registry) -> Vec<RegistryMuta
     mutations
 }
 
+/// Backfill a new `version_id` field on `ReplicaVersionRecord`s.
+/// This ID is also used in the registry key. Adding it to the record is more
+/// convenient, as only a `ReplicaVersionRecord` needs to be passed around.
+fn add_version_id_to_replica_versions(registry: &Registry) -> Vec<RegistryMutation> {
```
daniel-wong-dfinity-org-twin, line 387: This would only generate no more than a handful of mutations, right? Maybe we can add a check at the end to be extra safe? E.g.

```
if mutations.len() > 5 {
    log(...);
    return vec![];
}
```

rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -366,6 +375,27 @@ fn convert_type1dot1_nodes_to_type4dot5(registry: &Registry) -> Vec<RegistryMuta
...
+    for (id, mut version) in
+        get_key_family_iter::<ReplicaVersionRecord>(registry, REPLICA_VERSION_KEY_PREFIX)
+    {
+        if version.version_id.is_none() {
+            version.version_id = Some(id.clone());
+            mutations.push(update(
+                make_replica_version_key(id),
+                version.encode_to_vec(),
+            ))
+        }
```
daniel-wong-dfinity-org-twin: A slight improvement:


```suggestion
        if version.version_id.is_some() {
            continue;
        };

        version.version_id = Some(id.clone());
        mutations.push(...);
```

"Return" as soon as detected. This generally results in a nice property: the "main idea" is on the lowest possible level of indentation. Here, the main idea is "populate version_id". Ofc, there is a caveat: if it is already populated, leave it alone. This is a SECONDARY idea; therefore, it should be relegated to deeper level(s) of indentation.

rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -366,6 +375,27 @@ fn convert_type1dot1_nodes_to_type4dot5(registry: &Registry) -> Vec<RegistryMuta
...
 }
 
+/// Backfill a new `version_id` field on `ReplicaVersionRecord`s.
+/// This ID is also used in the registry key. Adding it to the record is more
+/// convenient, as only a `ReplicaVersionRecord` needs to be passed around.
+fn add_version_id_to_replica_versions(registry: &Registry) -> Vec<RegistryMutation> {
+    let mut mutations = Vec::new();
+
+    for (id, mut version) in
+        get_key_family_iter::<ReplicaVersionRecord>(registry, REPLICA_VERSION_KEY_PREFIX)
```
daniel-wong-dfinity-org-twin, line 391: See, this could have been done as a `.iter` + `.filter_map` + `.collect` dot chain. But there is _absolutely nothing wrong_ with doing it as a `for` loop! People are far too in love with dot chains. It's that whole "everything looks like a nail, when you fall in love with hammers" mental blind spot.

rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -44,6 +47,12 @@ pub fn canister_post_upgrade(
             total_batches += 1;
         }
 
+        let mutations = add_version_id_to_replica_versions(registry);
```
daniel-wong-dfinity-org-twin, line 50: Nice. Can you remember to delete this after it has gone into production? We should probably automate that...

rs/registry/helpers/src/replica_version.rs
```diff
@@ -137,13 +143,19 @@ mod tests {
         let replica_versions_and_records = vec![
             (
                 "version1",
-                create_replica_record("12345", &[measurement1, measurement2]),
+                create_replica_record("version1", "12345", &[measurement1, measurement2]),
+            ),
```
daniel-wong-dfinity-org-twin, line 147: Instead of pairs, the elements can just be `ReplicaVersionRecord`s. That's kind of the whole point of this PR, right?!

rs/registry/canister/src/invariants/replica_version.rs
Bownairo: Haha I said the same thing this week, not realizing you had already commented here. Very much agreed!

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
Bownairo: Using the `version` twice below required a little adjusting.

rs/registry/canister/src/registry_lifecycle.rs
Bownairo: I've truncated to 10 mutations at a time (there are currently 8) so that we still make progress if something were to go wrong and we add a bunch of versions. How does that sound?

rs/registry/canister/src/registry_lifecycle.rs
Bownairo: I tend to prefer to read top down without jumping around too much. Plus I'm sure they compile down to the same thing, so we don't really need any control flow hacks.

It also feels a little more Rust-y to me to keep the blast radius of a check within a block. e.g. There's a box around the code that should only run if `is_none()` is true. If something were to happen to the `continue` skip, this code could get loose, and it's on the user to enforce this, not the compiler.

---

That said, it won't be around for long anyway, and it is your domain, so it is done. (But it is still fun to talk about 😁).

rs/registry/canister/src/registry_lifecycle.rs
Bownairo: I have it queued up in https://github.com/dfinity/ic/pull/10943, for once this rolls out.

rs/registry/helpers/src/replica_version.rs
Bownairo: That's the goal! I've left this part to https://github.com/dfinity/ic/pull/10943, since it is easier to work without the `Option` wrapped on the type. (We can remove it once every version is filled.)

rs/registry/canister/src/registry_lifecycle.rs
daniel-wong-dfinity-org-twin: I would rather all or nothing. The main point of my suggestions is not to limit the amount of mutations. The point is, for bugs that result in a large amount of mutation, my suggestion would stop ALL bad mutations. Admittedly, many bugs would not be detected by such a filter, but at least the most damaging ones would.

alin-at-dfinity (approved):
Core protocol changes LGTM.

pierugo-dfinity (commented):
Should we add an invariant that ensures the version in the key is the same as the one in the record?

blind-oracle (approved):

nmattia (approved):

pierugo-dfinity (approved):
When making that field mandatory in the follow-up PR, an upgraded replica with an out-of-date registry, or just anyone that would decode records predating this change would decode the missing field as an empty string. Would that be a problem?

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -81,16 +80,18 @@ pub(crate) fn check_replica_version_invariants(
         if let Some(Err(defects)) = r.guest_launch_measurements.map(|v| v.validate()) {
             panic!("guest_launch_measurements are not valid. Defects: {defects:?}");
         }
+
+        // Enforce that the stored version always matches the key
+        if let Some(replica_version_id) = r.replica_version_id {
+            assert_eq!(&replica_version_id, &version);
```
daniel-wong-dfinity-org-twin: I think you can remove the two `&`s ?

rs/registry/canister/src/registry_lifecycle.rs
```diff
@@ -53,6 +53,12 @@ pub fn canister_post_upgrade(
             total_batches += 1;
         }
 
+        let mutations = remove_blessed_version_list_record(registry);
+        if !mutations.is_empty() {
+            registry.maybe_apply_mutation_internal(mutations);
+            total_batches += 1;
+        }
+
```
daniel-wong-dfinity-org-twin, line 61: This seems to come from another (one of your) PR(s), right? That's fine, but you might need to merge or rebase or whatever.

rs/registry/canister/src/registry_lifecycle.rs
daniel-wong-dfinity-org-twin: The problem with that is that it does not scale well with the number of requirements. Also, that requires more surgery when requirements change, and requirements have a tendency to change. The only constant is change. I explain my rationale in greater detail [here].

Anyway, this thread is still resolved. Just replying to share more of my rationale.

[here]: https://www.tumblr.com/magic-bad/764948183732387840/nesting-bad?source=share

daniel-wong-dfinity-org-twin (approved):

Bownairo:
> When making that field mandatory in the follow-up PR, an upgraded replica with an out-of-date registry, or just anyone that would decode records predating this change would decode the missing field as an empty string. Would that be a problem?

Hmm, I think since there is no rush, we can let it sit in this transition state for a while. These records get cycled out pretty regularly. If a node is upgraded and still has an old registry after a few weeks, it is probably dead for other reasons. I don't think the empty string would be much of a problem, either way.

rs/registry/canister/src/registry_lifecycle.rs
Bownairo: This diff doesn't render on the tip of the branch for me, so I think it's just a GitHub oddity.

---
"Conversation":
Screenshot 2026-08-13 at 11 30 17 AM
"Files changed":
Screenshot 2026-08-13 at 11 29 59 AM

PR #10946, opened 2026-07-29
feat(cketh): move funded deposit addresses to a balance-sweep queue

gregorydemay:
When the periodic balance scan finds a registered ckERC20 deposit address holding at least a token's minimum, that funded address is moved out of the time-expiring watchlist into a dedicated **sweep queue** in `AutomaticDeposits`, keyed per account with one entry per funded token. This hands the deposit off to the future sweeper and stops re-scanning the address. A deposit address is per-account and token-agnostic, so one address can yield several sweep entries (one per funded token) in a single scan. Nothing drains the queue yet.

Highlights:
- **Event-sourced.** Each detection is recorded the moment funds are found as an `AutomaticDepositReceived` audit event — one event per scanned account, listing every funded token — durable across an ungraceful trap, replayed on upgrade to rebuild the queue, and surfaced via `get_events` like every other event, rather than only via a pre-upgrade snapshot.
- **Account-keyed queue.** The queue is keyed by the user account together with the deposit address derived for it (stored once, since the address is fully determined by the account), and each queued entry names the ERC-20 token contract so it is never conflated with that deposit address.
- **Deposit status in `deposit_erc20`.** The response now carries a `status`: `Scanning` (armed, still looking) or `AwaitingSweep` (detected, queued — one entry per funded token, with amount and detection block), so a user can follow the multi-minute flow. A detected address is **not re-armed** while its funds are queued, preventing duplicate detection/registration. The status variant is extensible for the sweeper's later states (DEFI-2924).
- **No new metrics.** Balance-scan and sweep observability (queue size, watchlist size, oldest unswept address, per-kind scan error counters, …) is deferred to a dedicated ticket (DEFI-2965).

Deferred: draining/consuming the sweep queue and minting (DEFI-2924); observability metrics (DEFI-2965); the 2-of-3 `NoReduction` RPC client for latest-block scans (DEFI-2964).

Note: adds an `AutomaticDepositReceived` variant to the `get_events` output and reshapes `DepositErc20Response` (both on the debug/query surface of an unreleased feature), so the Candid change is gated by the `CI_OVERRIDE_DIDC_CHECK` label.

See the design doc: `rs/ethereum/cketh/docs/deposit_from_cex.md`.


rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -308,6 +308,56 @@ mod deposit_erc20 {
...
+        // token).
+        let scanned_at = 4_500_000_u64;
+        ckerc20.refresh_latest_block(scanned_at);
+        ckerc20.run_balance_scan(&vec![1_000_000_000_u128; tokens]);
+
+        ckerc20
+            .cketh
+            .check_minter_metrics()
+            .assert_contains_metric_matching(format!(
+                r"cketh_minter_sweep_queue_size {tokens} \d+"
```
gregorydemay: 🧐 🟠 Medium — Coverage: the pre_upgrade/post_upgrade round-trip of a *populated* sweep queue is never exercised end-to-end, so the minicbor `Encode`/`Decode` annotations on the new `SweepQueueEntry` are untested. The unit test `should_reproduce_equal_deposits_across_snapshot_round_trip` only covers the in-memory `snapshot()`→`rebuild()` logic (no CBOR), and `arb_event_type` omits the `RegisteredDepositAddresses` variant entirely, so `event_encoding_roundtrip` doesn't cover it either. Spec criterion 4 asks for `State::is_equivalent_to` to hold across pre_upgrade snapshot + post_upgrade replay with the queue. `check_audit_log` does exactly that (emit_preupgrade CBOR-encode → replay CBOR-decode → `is_equivalent_to`, which compares `automatic_deposits` incl. the `sweep` map). Please extend this positive test with `ckerc20.cketh.check_audit_logs_and_upgrade_as_ref(Default::default())` after the move and re-assert `cketh_minter_sweep_queue_size {tokens}` survives the upgrade — a wrong `#[n(..)]`/`with` on `SweepQueueEntry` would otherwise ship undetected.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -195,10 +271,25 @@ impl Default for AutomaticDeposits {
...
                 MAX_ACTIVE_DEPOSIT_ADDRESSES,
             ),
+            sweep: BTreeMap::new(),
         }
     }
 }
 
+/// A funded deposit address moved out of the watchlist and awaiting sweeping, for
+/// one supported ERC-20 token.
+#[derive(Clone, Eq, PartialEq, Debug)]
```
gregorydemay: 🧐 🔵 Nit — Unused/inconsistent derive: `SweepEntry` derives `Eq`, but its only container `AutomaticDeposits` derives just `Clone, PartialEq, Debug` (no `Eq`), and its sibling value type `DepositRequest` (line 293) also omits `Eq`. Nothing requires `SweepEntry: Eq`, so it's dead capability and diverges from the sibling — drop `Eq` to match `DepositRequest`. (`SweepQueueEntry`'s `Eq` in event.rs is fine — it's required by `DepositAddressRegistry`'s `Eq`.)

gregorydemay (commented):
🧐 **VERDICT: CHANGES_REQUESTED** — 1 🟠 Medium, 1 🔵 Nit; CI red/pending (Cargo Lint failed on the pre-autofix commit; a fresh run is pending on the autofix head — verdict is gated on it going green).

<details>

**Design challenge.** The chosen shape — collect candidates into `BTreeMap<Account, Vec<(Address, Erc20Value)>>` during the scan, then per scanned account either `move_to_sweep` or `record_scan` — is minimal and clean. `move_to_sweep` correctly takes the address of record from the removed watchlist entry (not from the scan), and the holder-major/token-minor index math (`batch.addresses[i / tokens.len()]`) matches `balance_of_calls`. I did not find a better-or-similar alternative worth proposing.

**Local verification.** `cargo build -p ic-cketh-minter` and `cargo clippy --all-features -p ic-cketh-minter` (repo flags) both pass. Bazel `lib_tests`, `unit_tests`, and `integration_tests_tests/ckerc20_test` all pass locally. The old CI Cargo Lint failure was an import-ordering issue that the autofix commit already fixed; the Cargo Build "duplicate symbol" only reproduces with local `--all-targets` (canbench bench + lib test linked together) and is a pre-existing quirk, not a PR regression. **A new CI run is still pending — READY is blocked until it is green.**

**Findings** (both posted inline):
- 🟠 Medium (ckerc20.rs positive test): the pre_upgrade/post_upgrade round-trip of a *populated* sweep queue is untested, so `SweepQueueEntry`'s minicbor `#[n(..)]`/`with` annotations ship uncovered. Unit test covers only in-memory `snapshot()`→`rebuild()`; `arb_event_type` omits `RegisteredDepositAddresses` so the encoding proptest skips it too. Fix is ~2 lines: add `check_audit_logs_and_upgrade_as_ref(Default::default())` after the move and re-assert the metric.
- 🔵 Nit (automatic_deposits/mod.rs:281): `SweepEntry` derives an unused `Eq` that its sibling `DepositRequest` omits.

**Maintainability rundown.**
- Duplication: none found (new tests are `Vec<TestCase>` tables; no repeated multi-line blocks).
- Structural duplication vs siblings: `SweepEntry`/`SweepQueueEntry` intentionally mirror `DepositRequest`/`DepositAddressRegistration`; the `sweep` `BTreeMap` differs from the watchlist `TimedSizedMap` (no time index needed). Consistent with established pattern — cleared.
- Unused derives: `SweepEntry: Eq` (🔵 above). `SweepQueueEntry`'s derives all used (`Eq` required by `DepositAddressRegistry`'s `Eq`; `Decode`/`Encode` for CBOR). 
- Primitive-obsession params: none — `move_to_sweep`/`collect_candidates` use `Account`, `Address`, `BlockNumber`, `Erc20Value`, `Timestamp` throughout.
- Divergent invariant handling: `scan_count` increment is `saturating_add(1)` in *both* `record_scan` and `move_to_sweep`; liveness guard uses `get_entry(now, …)` in both — cleared.
- Silent fallbacks: none — `remove(...).expect("BUG: …")` is a justified assert right after a successful `get_entry`; the `.entry().or_default()` accumulation is expected. Cleared.
- Test-only code in production: none — helpers (`token`, `sweep_entry`) live in `tests.rs`. Cleared.
- Redundant/derivable params & caller-owned decisions: `collect_candidates` takes `tokens` only for `tokens.len()` (could derive from `batch`, but clearer as-is) — not flagged. Move-vs-record decision sits with the scan while `move_to_sweep` still guards empty/expired — appropriate. Cleared.

**Coverage / test pyramid.** Unit coverage is strong and correctly table-driven: `collect_candidates` (2 addresses × 2 tokens, incl. an attribution case that a wrong holder index would fail), `move_to_sweep` positive + 3-case no-op table, `TimedSizedMap::remove` (shared-bucket + bucket-drop + absent-key), snapshot round-trip incl. queue. Integration covers positive (moved, one entry per token, re-registration arms afresh) and negative (below-min → `sweep_queue_size 0`, schedule advances). Assertions use distinct balances (10/20) pinning per-token values — no constant-mock blind spots. The one gap is the end-to-end upgrade round-trip above.

**Docs.** Public items (`move_to_sweep`, `snapshot`, `rebuild`, `remove`, `SweepEntry`, `SweepQueueEntry`, `sweep_len`) carry terse `///` docs matching sibling style; the `snapshot` "(account, token) order" note matches the `BTreeMap` iteration order. The `DEFI-2923` TODO was removed as intended; no JIRA IDs remain in docs.

**Criterion 7 (candid-neutral / sweep not in `get_events`).** Acceptable for this PR: the queue round-trips via the internal stable-memory event log (which carries `sweep_queue`), while `get_events`/`.did` stay unchanged and `map_event` sets `sweep_queue: vec![]`. Since the feature is unreleased and draining is a later PR, keeping it internal is fine — but worth surfacing in the public audit trail once sweep actions become user-visible.

</details>

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -195,10 +271,25 @@ impl Default for AutomaticDeposits {
...
+pub struct SweepEntry {
```
gregorydemay: 🧐 🔵 G8: `SweepEntry` (and all four of its fields) is exposed `pub`, but it is used only inside this module — it is the private `sweep` map's value type, constructed and read solely within `automatic_deposits/mod.rs` (no reference in any other module or the integration-test crate). Unlike its sibling `DepositRequest`, which genuinely needs `pub` because `main.rs`/`state.rs`/tests consume it, `SweepEntry` has zero cross-module users, so the G11-consistency argument doesn't apply here. Drop `pub` from the struct and its fields (a bare `struct SweepEntry` with private fields compiles, since all access is in-module) to minimize the public surface. Ties in with the `Eq` nit already posted on line 281.

gregorydemay (commented):
🧐 **Clean-code pass (2nd review)** — 1 new 🔵 Nit (G8); consolidated verdict unchanged: **CHANGES_REQUESTED** (1 🟠 Medium + 3 🔵 Nits across both passes), gated on CI going green.

<details>
<summary>Clean-code pass details</summary>

Applied R.C. Martin's catalog to the diff only. One new finding, posted inline:
- 🔵 G8 (automatic_deposits/mod.rs:282): `SweepEntry` is fully `pub` but has zero cross-module users (unlike its sibling `DepositRequest`) — make it a private `struct` with private fields.

Considered and deliberately **not** flagged (reasoning stated so it isn't silently omitted):
- **F1 — `move_to_sweep(&mut self, now, account, block, candidates)` (4 args).** `now`/`account`/`block` deliberately mirror `record_scan` in the same file; G11 consistency outweighs a struct-ification of a 3+1 signature. Not flagged.
- **R2/N4 — `candidates: &[(Address, Erc20Value)]` (token, balance).** The first `Address` being a *token* is only clear from naming/docs, but this whole module uses `Address` for ERC-20 token contracts everywhere (`BalanceOfCall.token`, `min_deposit(&Address)`); a token newtype would be a cross-module change and would break G11. Not flagged.
- **G5 — `SweepEntry`↔`SweepQueueEntry` field mapping in `snapshot`/`rebuild`.** These are inverse (de)serialization boundaries, not duplicated knowledge, and mirror the existing registration mapping. Not a DRY violation.
- **G5/F — `collect_candidates(batch, tokens, balances)` passing `tokens` when `batch` implies it.** Deriving `tokens.len()` from `batch.calls.len()/batch.addresses.len()` would be less clear and introduce a divide-by-zero edge; explicit `tokens` is the safer, more readable choice.
- **R1 — `remove(...).expect("BUG: entry is live right after a successful get_entry")`.** Textbook-correct `expect` for a genuinely-impossible case with a proving message; a positive example, not a smell.
- **G36 / R3 — Law of Demeter & borrow-vs-clone.** No train wrecks in the new code; no gratuitous `.clone()` (all copies are `Copy` types).
- **T1/T5 boundaries.** New tests cover ==min / >min / <min / zero, empty/expired/absent no-ops, and `remove` shared-bucket + bucket-drop + absent-key — good boundary coverage. The one real test gap (populated-queue upgrade round-trip) is the 🟠 Medium from pass 1, not restated here.

**Boy-Scout positives** (no action needed): the stale `// TODO DEFI-2923` was removed (C1/C2), and `watchlist_snapshot`/`rebuild_watchlist` were renamed to `snapshot`/`rebuild` to match their now-broader responsibility (N1). `last_scanned_block: BlockNumber` (non-optional) on the sweep types is a nice illegal-states-unrepresentable improvement over `DepositRequest`'s `Option<BlockNumber>` (R2/G27).

</details>

rs/ethereum/cketh/minter/src/timed_sized_map/mod.rs
```diff
@@ -198,6 +198,19 @@ impl<K: Ord + Clone, V> TimedSizedMap<K, V> {
...
+    pub fn remove(&mut self, key: &K) -> Option<Entry<V>> {
+        let entry = self.entries.remove(key)?;
+        if let Some(bucket) = self.by_time.get_mut(&entry.expires_at) {
+            bucket.retain(|k| k != key);
+            if bucket.is_empty() {
+                self.by_time.remove(&entry.expires_at);
+            }
+        }
+        Some(entry)
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Done in 4df9559e4e: `remove` now `expect`s the entry's `by_time` bucket (with a BUG message) instead of silently skipping a missing one, matching the invariant-panic style of `iter_by_expiry`/`evict_expired`.

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Done in f2daf8632b: the positive integration test now drives a real pre_upgrade snapshot → CBOR encode → replay → post_upgrade round-trip via `check_audit_logs_and_upgrade_as_ref(Default::default())` after the address is moved into the sweep queue (this also asserts `is_equivalent_to` across the replay, comparing the `sweep` map), then re-asserts `cketh_minter_sweep_queue_size {tokens}` survives the upgrade. A wrong minicbor annotation on `SweepQueueEntry` would now fail here.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Done in 060a5d0099: dropped the unused `Eq` derive from `SweepEntry` to match its sibling `DepositRequest`.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Done in e761bb4363: `SweepEntry` and its fields are now private (bare `struct SweepEntry` with private fields). Confirmed no cross-module users (grep across `rs/`), and the child `tests` module still constructs it via `use super::*`. Build/clippy green.

gregorydemay:
🤖 On criterion 7 (candid-neutral / sweep queue not in `get_events`): confirming this is deliberate for this PR. The sweep queue is internal state — it round-trips across upgrades via the stable-memory event log (the internal `DepositAddressRegistry` carries `sweep_queue`), while `get_events`/the `.did` file stay untouched and `map_event` leaves the endpoint representation unchanged. Since the feature is unreleased and draining the queue is a later PR, keeping it internal is intended; surfacing sweep actions in the public audit trail is deferred to when they become user-visible. No code change.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -46,6 +47,9 @@ const MAX_ACTIVE_DEPOSIT_ADDRESSES: NonZeroUsize = NonZeroUsize::new(7_000).unwr
 #[derive(Clone, PartialEq, Debug)]
 pub struct AutomaticDeposits {
     watchlist: TimedSizedMap<Account, DepositRequest>,
+    /// Funded deposit addresses moved out of the watchlist, awaiting sweeping,
+    /// keyed per `(account, ERC-20 token contract)`.
+    sweep: BTreeMap<(Account, Address), SweepEntry>,
```
gregorydemay: Let's use a proper struct for this key, otherwise Adress is being overused between user-derived address and ERC-20 token address.

rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -190,9 +190,32 @@ pub struct DepositAddressRegistry {
...
     pub capacity: u64,
     /// Registered addresses in time-index order (ascending expiry, insertion
-    /// order within a shared expiry), as produced by `watchlist_snapshot`.
+    /// order within a shared expiry), as produced by `snapshot`.
     #[n(2)]
     pub registrations: Vec<DepositAddressRegistration>,
+    /// Funded deposit addresses moved out of the watchlist and awaiting sweeping,
+    /// one entry per `(account, token)`, in `(account, token)` order.
+    #[n(3)]
+    pub sweep_queue: Vec<SweepQueueEntry>,
```
gregorydemay: no this should not be part of `DepositAddressRegistry`, that is only emitted in pre_upgrade. It should be its own event, recorded when emitted like the other events.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Done in 05a57d0ef0: the sweep queue is now keyed by a named `SweepKey { account, token }` struct, so the ERC-20 token contract address is named `token` and no longer conflated with `SweepEntry.address` (the user-derived deposit address). Kept private; updated all constructions/lookups.

rs/ethereum/cketh/minter/src/state/event.rs
gregorydemay: 🤖 Done in 6f659bff86: the sweep move is now its own event, recorded the moment funds are detected (durable across an ungraceful trap), instead of riding on the pre_upgrade-only `DepositAddressRegistry` snapshot.

- New `EventType::MovedToSweepQueue(SweepMove)` (`#[n(26)]`); dropped `sweep_queue` from `DepositAddressRegistry`.
- `move_to_sweep` -> unconditional, idempotent `apply_sweep_move`, applied identically on live processing and replay; `snapshot`/`rebuild` now cover the watchlist only, and `rebuild` deliberately leaves the sweep map intact so the mid-stream `MovedToSweepQueue` events preceding the final snapshot are not wiped.
- The balance scan emits one `MovedToSweepQueue` per funded `(account, token)` via `process_event`; non-candidates still advance the schedule.
- Exposed via `get_events` (candid `EventPayload` + `.did`), so the `CI_OVERRIDE_DIDC_CHECK` label was added. The integration test now proves the queue survives a real pre/post-upgrade cycle via event replay (not the snapshot).

rs/ethereum/cketh/minter/src/state/audit.rs
```diff
@@ -166,7 +166,10 @@ pub fn apply_state_transition(state: &mut State, payload: &EventType) {
             );
         }
         EventType::RegisteredDepositAddresses(registry) => {
-            state.automatic_deposits.rebuild_watchlist(registry)
+            state.automatic_deposits.rebuild(registry)
+        }
+        EventType::MovedToSweepQueue(sweep_move) => {
```
gregorydemay: nit:

1. Rename event to AutomaticDepositReceived
2. rename `apply_sweep_move` to `record_automatic_deposit_received`

gregorydemay: 🤖 Done in b3e27f7660: renamed the event `EventType::MovedToSweepQueue` -> `AutomaticDepositReceived` (cbor index `#[n(26)]` unchanged; candid `EventPayload` variant + `.did` updated to match) and `AutomaticDeposits::apply_sweep_move` -> `record_automatic_deposit_received`. Also renamed the event payload `SweepMove` -> `AutomaticDeposit` so the variant reads coherently, cascading the producer/test helpers. The internal sweep-queue names (`sweep`/`SweepKey`/`SweepEntry`/`DetectedSweep`/`sweep_entries_for_account`) stay — they describe the queue, not the event. Pure rename, no behavior change.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -398,14 +398,37 @@ type DepositMode = variant {
...
-    // null if it has not been scanned yet. Surfaces the balance-scan progress.
-    last_scanned_block : opt nat;
-    // How many times this address' balance has been scanned so far.
-    scan_count : nat64;
+    // Where the deposit stands in the detect-and-sweep pipeline.
+    status : DepositStatus;
+};
+
+// The stage a ckERC20 deposit address is at. Extensible with Sweeping/Swept
+// once sweeping lands (DEFI-2924).
```
gregorydemay: no ticket in public docs

```suggestion
// The stage a ckERC20 deposit address is at.
```

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -398,14 +398,37 @@ type DepositMode = variant {
...
+    // entry per funded token.
+    AwaitingSweep : vec DetectedDeposit;
+};
+
+// A funded token detected at a deposit address and queued for sweeping.
+type DetectedDeposit = record {
+    // The ERC-20 token contract whose balance was found.
+    token : text;
+    // The balance detected for `token`.
+    amount : nat;
```
gregorydemay: nit rename
```suggestion
    scanned_balance : nat;
```

since the balance may change until sweep.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -88,7 +95,11 @@ impl AutomaticDeposits {
     /// `capacity` (no admission check). This makes the restored state equal to
     /// the one that produced the registry, which the event-log equivalence
     /// check relies on. Changing the limits across versions is future work.
-    pub fn rebuild_watchlist(&mut self, registry: &DepositAddressRegistry) {
+    ///
+    /// The sweep queue is deliberately left untouched: it is reconstructed from
+    /// the mid-stream `AutomaticDepositReceived` events that precede the final snapshot
+    /// event in the log, so clearing it here would wipe them.
+    pub fn rebuild(&mut self, registry: &DepositAddressRegistry) {
```
gregorydemay: why was the method's name changed? it's still specific to the wathclist.

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Done in 3067be996e: dropped the "Extensible with Sweeping/Swept once sweeping lands (DEFI-2924)" line so the published candid carries no internal ticket reference — the comment is now just `// The stage a ckERC20 deposit address is at.`

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Done in 3067be996e: renamed `amount` to `scanned_balance` in `DetectedDeposit` (the `.did`, the `endpoints.rs` struct, the producer in `automatic_deposits/mod.rs`, and the unit/integration tests), reflecting that the balance may change before the sweep. It also lines up with the `scanned_balance` field on the `AutomaticDepositReceived` event.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 You are right — reverted in 3067be996e back to `rebuild_watchlist` (and its sibling `snapshot` -> `watchlist_snapshot`). Both only touch the watchlist; the sweep queue is event-sourced (reconstructed from `AutomaticDepositReceived` events, not the registry), so the generic names were misleading. The registry that `watchlist_snapshot` produces carries only watchlist data, so the `watchlist_` prefix is accurate for the pair. Callers in `audit.rs`/`main.rs` and the unit tests updated accordingly.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -119,76 +135,48 @@ async fn scan<R: Runtime>(
...
+        for (account, candidates) in &scanned {
+            if candidates.is_empty() {
+                s.automatic_deposits.record_scan(now, account, latest_block);
+                continue;
+            }
+            // A funded address is event-sourced into the sweep queue (durable the moment funds are
+            // detected) instead of being advanced, so it is no longer re-scanned. Its watchlist
+            // entry gives the derived address and the scan count (this finding scan included); skip
+            // if the account is no longer live as of `now`.
+            let Some(entry) = s.automatic_deposits.get_entry(now, account) else {
```
mbjorkqvist: A detection can be discarded here with no event, no sweep entry and no log line.

`scanned` is keyed by `Account`, discarding the `DepositAccount` the scan already held, so recovering the address and `scan_count` needs a second watchlist lookup — and that lookup can fail. `evict_expired` is reachable from `insert_entry`, so any other user's `deposit_erc20` interleaving at one of the scan's outcall await points evicts this account's entry if its window closed after `now`. The scan then resumes, `get_entry(now, …)` returns `None`, and the funds the scan *just observed on-chain* are forgotten.

On impact, I want to be fair: the end state is indistinguishable from an ordinary, by-design outcome. `SCAN_GAP_SECS` sums to ~23.5h against a 24h window, so a deposit arriving in the last ~30 minutes is never scanned at all and the user has to re-arm to claim it — documented behaviour. The race lands the user in the same place. No loss, no double-credit, no state corruption. So this is not the correctness bug it first looks like.

What is defective is narrower: in the ordinary late-deposit case there is nothing to record because nothing was detected; here the minter knew and forgot silently. And whether to honour a detection whose armed window closed mid-scan is a legitimate policy choice — the problem is that the choice isn't expressed anywhere. It falls out of a `get_entry` call made for an unrelated reason (fetching the address and scan count).

Suggestion, in preference order:
1. Key `scanned` by `DepositAccount` — it already carries the address, so the lookup disappears and only `scan_count` needs the entry (defaulting when absent).
2. Failing that, `log!(INFO, …)` on this branch and say in the comment that a detection on a now-unarmed address is deliberately dropped.

Note for triage: DEFI-2924 does not cover this. Nothing ever enters the queue in this scenario, so no sweeper implementation can recover it.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -323,3 +311,89 @@ fn call_args(input: Vec<u8>, block: BlockNumber) -> evm_rpc_types::CallArgs {
...
+impl<'a> BalanceScan<'a> {
+    fn new(accounts: &'a [DepositAccount], tokens: &'a [Address]) -> Self {
+        Self { accounts, tokens }
+    }
+
+    /// Split into batches of at most `size` `balanceOf` calls, chunking whole accounts so an
+    /// account's per-token calls are never spread across batches. Panics if `size` is smaller than
+    /// the token count, which would make a single account's calls exceed one batch.
+    fn batch(&self, size: usize) -> BalanceScanBatch<'a> {
+        assert!(
```
mbjorkqvist: This assert converts a case the previous code handled gracefully into a trap on the timer path.

`plan_batches`/`addresses_per_chunk` guaranteed at least one whole address per batch even when a single address' token calls exceeded the cap (`.max(1)`), and `should_never_split_an_address_across_batches` covered exactly that with `num_tokens ∈ {1, 2, 7, MAX_CALLS_PER_BATCH, MAX_CALLS_PER_BATCH + 1}`. That coverage is now `{1, 2, 7}` and the case it protected became a panic. If the supported ckERC20 set ever passes 1000, the balance scan traps on every tick instead of sending oversized-but-still-cheap batches — and `MAX_CALLS_PER_BATCH`'s own doc comment argues an oversized batch stays cheap (32 bytes of return per call), which is the behaviour the assert now forbids.

Two follow-ons at line 337:
- Both `.max(1)` calls are dead given the assert.
- The assert is vacuous when `tokens.len() == 0` (`size >= 0`), so the type no longer rejects the empty-token case; only `scan`'s early return does.

If trapping is deliberate, restoring the `MAX_CALLS_PER_BATCH{,+1}` rows as `#[should_panic]` cases would at least keep the invariant pinned. Minor naming point while here: `batch()` returns an iterator *of* batches and `BalanceScanBatch` yields `BalanceScan` values — `batches()` / `BalanceScanBatches` would read more naturally.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -46,6 +49,10 @@ const MAX_ACTIVE_DEPOSIT_ADDRESSES: NonZeroUsize = NonZeroUsize::new(7_000).unwr
 #[derive(Clone, PartialEq, Debug)]
 pub struct AutomaticDeposits {
     watchlist: TimedSizedMap<Account, DepositRequest>,
+    /// Funded deposit addresses moved out of the watchlist, awaiting sweeping,
+    /// keyed by the funded [`DepositAccount`]; each holds one [`SweepEntry`] per
+    /// funded ERC-20 token.
+    sweep: BTreeMap<DepositAccount, Vec<SweepEntry>>,
```
mbjorkqvist, line 55: The sweep queue has no capacity check, no TTL, and nothing drains it (DEFI-2924) or measures it (DEFI-2965).

The important interaction: `MAX_ACTIVE_DEPOSIT_ADDRESSES` (7000) is the admission bound for tracked deposit addresses, and `record_automatic_deposit_received` removes the account from the watchlist — so moving an account into `sweep` *frees* its slot. Total tracked state therefore has no bound at all, whereas before this PR it did.

Each entry costs a real at-or-above-minimum token deposit (~$10) at a distinct account, so this isn't cheap to abuse and the funds are the user's own. But "unbounded and unbounded-in-time" is a different property from "bounded at 7000", and the queue is the thing that grows while its consumer doesn't exist yet.

I don't think this has to be solved in this PR, but I'd like it to be a conscious decision: either a cap (with a defined behaviour on overflow), or an explicit accepted-risk note tying the bound to DEFI-2924 landing before the feature is enabled on mainnet.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -152,17 +163,49 @@ impl AutomaticDeposits {
...
+            subaccount: deposit.subaccount,
+        };
+        self.watchlist.remove(&account);
+        let entries = self
+            .sweep
+            .entry(DepositAccount::new(account, deposit.address))
+            .or_default();
+        for detected in &deposit.deposits {
+            assert!(
+                !entries.iter().any(|e| e.erc20_token == detected.token),
```
mbjorkqvist, line 208: This assert lives in the replay path, so a duplicate doesn't just trap a scan tick — it makes `post_upgrade` trap.

I traced reachability and believe a duplicate `(account, token)` cannot occur today: `ckerc20_tokens` is a `DedupMultiKeyMap`, so contract addresses are distinct within one scan; and a funded account leaves the watchlist while `deposit_status` short-circuits to `AwaitingSweep`, so it can't be re-registered and re-scanned. Replay ordering is also fine — `rebuild_watchlist` deliberately leaves `sweep` alone, and the snapshot event is appended after the detections it follows.

So this is a shape question, not a bug. What makes it worth asking: 6f659bff86 described this as "unconditional, **idempotent** `apply_sweep_move`, applied identically on live processing and replay", and a later commit deliberately replaced that with trapping. An idempotent apply is the safer formulation for something `post_upgrade` replays, and it was already the design. Was there a specific failure mode that motivated trapping over skip-and-log? If the intent is to make a corrupt log loud, that's defensible — but then a comment saying "trapping here is a `post_upgrade` trap, and that is intended" would be worth having, since the blast radius isn't obvious from the call site.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -186,6 +229,46 @@ impl AutomaticDeposits {
...
+    /// has funds queued for sweeping (so it must be registered). Reports
+    /// [`DepositStatus::AwaitingSweep`] once funds have been detected and queued (one entry per
+    /// funded token, all sharing the deposit address), otherwise [`DepositStatus::Scanning`] while
+    /// the address is armed and being scanned as of `now`.
+    pub fn deposit_status(
+        &self,
+        now: Timestamp,
+        account: &Account,
+    ) -> Option<DepositErc20Response> {
+        if let Some((deposit_account, entries)) = self.sweep.get_key_value(account) {
```
mbjorkqvist, line 264: Because the sweep lookup short-circuits before the watchlist, a second token arriving at an already-queued address is never noticed.

Deposit addresses are per-account and token-agnostic, so this is a reachable user flow: withdraw USDC from a CEX → detected and queued → then withdraw USDT to the same address. `deposit_erc20` returns the same `AwaitingSweep([USDC])`, refuses to re-arm, and the address is off the watchlist, so the USDT is never scanned. Until DEFI-2924 drains the queue there is no way for the user to get it noticed at all.

Once sweeping lands this presumably becomes a transient of minutes, which is fine. But the design doc currently says only that the address is "not re-armed by further `deposit_erc20` calls … until it is swept" without saying what happens to funds that arrive during that interval. Worth stating explicitly — and worth confirming that DEFI-2924 re-arms (or re-queues) rather than just clearing the entry, since "swept and forgotten" would make the gap permanent.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -195,10 +278,76 @@ impl Default for AutomaticDeposits {
...
+
+/// A funded token awaiting sweeping at a [`DepositAccount`]'s deposit address.
+#[derive(Clone, PartialEq, Debug)]
+struct SweepEntry {
+    /// The ERC-20 token contract whose balance was found.
+    erc20_token: Address,
+    /// The block whose scan found the funds.
+    last_scanned_block: BlockNumber,
+    /// How many times the address was scanned, including the finding scan.
+    scan_count: u32,
```
mbjorkqvist, line 363: `scan_count` is write-only: set at line 198, never read outside the tests. `deposit_status` reads `erc20_token`, `scanned_balance` and `last_scanned_block` only.

Related shape point: `last_scanned_block` and `scan_count` are facts about the *scan*, not about the token, so both are copied onto every per-token entry for the same account. This looks like residue from the re-key — when the queue moved from `SweepKey { account, token }` to account-keyed, `address` was correctly hoisted onto the key, but these two weren't. A per-account value would model it once:

```rust
struct SweepState {
    detected_at_block: BlockNumber,
    scan_count: u32,          // or drop, per above
    entries: Vec<SweepEntry>, // { erc20_token, scanned_balance }
}
```

Minor naming note in the same area: this field is `last_scanned_block` here, `last_scanned_block` on the event, and `detected_at_block` on the endpoint — one value, two names. Its own doc here ("The block whose scan found the funds") argues for `detected_at_block` throughout.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -409,6 +409,14 @@ unique, deterministic deposit address, derived from the minter's threshold-ECDSA
   vector. With `fee = {from_subaccount, max_fee}`, the call is *sponsored*: the
   caller pays the sweep gas in ckETH and detection/sweep/crediting run on demand
   (step 0). Repeated calls are cheap lookups that re-arm the window.
+* The response carries the address plus a **status** so a caller can follow the
+  (multi-minute) detection progress: `Scanning { valid_until, last_scanned_block,
+  scan_count }` while the address is armed and no balance at or above the per-token
+  minimum has been seen yet, or `AwaitingSweep([{ token, amount, detected_at_block }, …])`
```
mbjorkqvist: Leftover from the `amount` → `scanned_balance` rename in 3067be996e: this still reads `{ token, amount, detected_at_block }`. The rename landed in the `.did`, `endpoints.rs`, the producer and the tests, but not here.

rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -190,11 +196,43 @@ pub struct DepositAddressRegistry {
     #[n(1)]
     pub capacity: u64,
     /// Registered addresses in time-index order (ascending expiry, insertion
-    /// order within a shared expiry), as produced by `watchlist_snapshot`.
+    /// order within a shared expiry), as produced by `snapshot`.
```
mbjorkqvist: Leftover from the `snapshot` → `watchlist_snapshot` revert in 3067be996e: this doc now points at a `snapshot` method that no longer exists. Should be `watchlist_snapshot`.

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -299,14 +321,147 @@ mod deposit_erc20 {
...
+                if scan_count == 1 && last_scanned_block == Some(candid::Nat::from(scanned_at))
+        );
+
+        // The balances were below every token's minimum, so nothing was moved to the sweep queue.
+        assert!(
+            ckerc20.cketh.get_all_events().iter().all(|event| !matches!(
+                event.payload,
+                EventPayload::AutomaticDepositReceived { .. }
+            )),
+            "below-minimum balances must not emit a AutomaticDepositReceived event"
```
mbjorkqvist: Nit: "a AutomaticDepositReceived" → "an AutomaticDepositReceived". (Copilot flagged this but suppressed it as low-confidence, so it was never posted.)

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -722,7 +717,8 @@ fn get_events(arg: GetEventsArg) -> GetEventsResult {
 
     fn map_event(Event { timestamp, payload }: Event) -> CandidEvent {
         use ic_cketh_minter::endpoints::events::{
-            DepositAddressRegistration as CandidDepositAddressRegistration, EventPayload as EP,
+            DepositAddressRegistration as CandidDepositAddressRegistration,
+            Erc20Balance as CandidDetectedErc20, EventPayload as EP,
```
mbjorkqvist: Nit: the alias `CandidDetectedErc20` doesn't match the type it renames (`Erc20Balance`), which makes the mapping at line 751 harder to follow than the sibling `CandidDepositAddressRegistration` above it. `CandidErc20Balance` would match the local convention.

Worth noting the underlying cause: there are now two distinct `Erc20Balance` types — `state::event::Erc20Balance` and `endpoints::events::Erc20Balance` — which is what forces an alias here at all.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -314,6 +314,8 @@ rust_test(
         ":minter",
         "//packages/ic-ethereum-types",
         "//rs/ethereum/cketh/test_utils",
+        "@crate_index//:assert_matches",
+        "@crate_index//:candid",
```
mbjorkqvist: Nit: `deposit_from_cex.rs` doesn't reference `candid` — the `Nat` comparisons in the new assertions resolve through trait impls without a direct dependency. Probably left over from an earlier revision that used `candid::Nat::from(…)`; `assert_matches` on the line above is genuinely needed.

mbjorkqvist (commented):
Thanks @gregorydemay!

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fixed in 9344ee55b4 — propagated the `amount` → `scanned_balance` rename to this doc; it now reads `{ token, scanned_balance, detected_at_block }`. Thanks for catching the leftover.

rs/ethereum/cketh/minter/src/state/event.rs
gregorydemay: 🤖 Fixed in 9344ee55b4 — the doc now points at `watchlist_snapshot`.

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Fixed in 9344ee55b4 — "a" → "an AutomaticDepositReceived".

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Fixed in 9344ee55b4 — renamed the alias to `CandidErc20Balance` to match the type it renames (`Erc20Balance`) and the sibling `CandidDepositAddressRegistration` convention.

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: 🤖 Fixed in 9344ee55b4 — dropped the now-unused `candid` dep from the `deposit_from_cex` target; `bazel build … --nobuild` still analyzes the target cleanly (`assert_matches` is kept, it is genuinely needed).

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Fixed in 5e332d1f79, and the underlying shape addressed in 49d1bdc15b. Thanks — the "the choice isn't expressed anywhere, it falls out of a `get_entry` call made for an unrelated reason" framing is exactly right, and it pointed at a better fix than just picking a policy.

**The policy, now explicit:** funds the scan observed on-chain are queued whether or not the address is still armed. The minter is the only party that saw them, so dropping the detection would lose it with no trace. That is stated at the call site rather than falling out of a lookup.

**The mechanism:** rather than option 1, the due entries are read once up front and carried through the scan. They already hold the address *and* the scan count, so the post-outcall lookup disappears entirely — no defaulting needed for `scan_count`. `addresses_to_scan_iter` now yields whole entries; its doc says why, and points at `watch_address_for_account` as the eviction path (`evict_expired` is reachable from `insert`), which is where I'd have looked first.

**Why a second commit:** I could not write a regression test for this. The funded branch calls `process_event` → `record_event` → `ic0::time`, which panics outside a canister, so no host-side test can reach it — which is also why the existing scan tests all use below-minimum balances and the funded path had zero unit coverage.

So 49d1bdc15b splits the scan at the state boundary. `scan_balances` does all the I/O and turns the due entries into one `ScanOutcome` per scanned account with no `State` in scope; `scan` reads, awaits fully, then applies the outcomes in a single `mutate_state`. That makes the funded path testable (three new tests, including one pinning that the detection's address and `scan_count` come from the pre-scan entry) and makes this class of bug unrepresentable rather than merely fixed: there is no watchlist for a post-outcall read to go stale against.

**On triage:** agreed it isn't DEFI-2924's. Two knock-on effects worth flagging against your other threads — a detection on an address whose window had already closed now permanently occupies a sweep slot (#discussion_r3705268244) and permanently blocks re-arming (#discussion_r3705268254). Both were already true for detections on live addresses; this widens the set slightly. Still open on my side.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Fixed in 6738d34843 — you were right on all three points, and on the history.

**The assert is gone.** `batches` holds at least one whole account and overshoots the requested size, which is what `addresses_per_chunk`'s `.max(1)` did before the size became a parameter. Weighing the two failure modes settles it: an oversized batch at worst fails at the provider and stalls its own addresses, while the assert stalls the *entire* scan on every tick. The doc on `batches` now states that `size` is a target rather than a cap, and why overshooting is the safer of the two.

**Both `.max(1)` calls are live again, and now pinned.** I checked the mutations individually — dropping either one fails the suite:
- The outer guard is covered by the restored `MAX_CALLS_PER_BATCH` / `MAX_CALLS_PER_BATCH + 1` rows. Worth noting the rows alone would not have caught this: the test also computed `batch_size = 2 * num_tokens`, so it never entered the overshoot regime. It now pairs each token count with an explicit batch size, and the two large rows use the real cap.
- The empty-token divide is covered by extending `should_build_no_calls_when_no_addresses_or_no_tokens` to batch a token-less scan. As you note, only `scan`'s early return rules this out in production; the test says so.

`should_reject_a_batch_smaller_than_the_token_count` is replaced by `should_overshoot_a_batch_smaller_than_the_token_count`, asserting the actual per-batch calls rather than a panic.

**Renamed** `batch` → `batches` and `BalanceScanBatch` → `BalanceScanBatches`.

Verified: `cargo fmt` clean, clippy clean, all 7 affected bazel targets pass.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: > The sweep queue has no capacity check, no TTL

Consumer will happen in another PR. I think it's ok to have no capacity/TTL for now: every entry in that list requires an Ethereum transaction with a minimum amount of roughly ($10) so that blowing up the heap would require a serious amount of 💵 . This is also similar to what we have in other queues, like `pending_withdrawal_requests` which is also not limited because every withdrawal request entry here requires a burn on the ledger with a certain minimum amount.

It does make sense to track the size of such queues via a metric, which I defer to DEFI-2965.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Documented in 7178590519 — keeping the trap, with the blast radius spelled out as you suggested.

**On what motivated the change:** 9329c1d626. One correction to the framing though — the behaviour it replaced was *upsert*, not skip-and-log: it did `entries.iter_mut().find(...)` and overwrote the existing `SweepEntry`. That is what 6f659bff86 meant by "idempotent" (replaying the same event twice lands on the same state). So the choice is three-way, and both non-trapping options silently pick a winner between two conflicting records of the same funds — upsert takes the last, skip takes the first — with `scanned_balance` being exactly what DEFI-2924 will act on. Neither seemed better than being loud.

**Why loud is in keeping here:** `replay_events`' own doc already states it panics when "one of the events in the log invalidates the minter's state invariants", and several sibling handlers reached from `apply_state_transition` do exactly that on the same replay path — `record_add_ckerc20_token` (three asserts), `record_erc20_withdrawal_request`, `record_skipped_block_for_contract`, `record_invalid_deposit`. An idempotent apply may well be the safer general formulation, but adopting it here alone would make this line the odd one out rather than the safe one.

**What the doc now says:** a `# Panics` section covering that `(account, token)` reaches the queue at most once, that `apply_state_transition` runs on replay so this traps `post_upgrade` and blocks upgrades until a repairing version ships, that this is deliberate, and a pointer to the sibling handlers.

**Reachability:** thanks for tracing it — I re-verified after the two commits above, since 5e332d1f79 now queues a detection even when the address is no longer armed. Still unreachable, and that change does not open a path: `TimerGuard` is held across every await so scans never overlap; within a scan the `DedupMultiKeyMap` gives distinct tokens and `scanned` is keyed by account, so one event per account; and across scans a queued account cannot re-enter the watchlist, since `deposit_erc20` (`main.rs:191-193`) returns early on any `deposit_status` and that checks `sweep` before the watchlist, so it never reaches `register_deposit_address`.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: good observation! I have the impression that this UX defect and the complexity of the current implementation (1 address spawns multiple balance scans, 1 per supported ERC-20, need to ensure scan for 1 address is done in same batch, etc.) originates from trying to scan everything for a single deposit address, when the actual granularity is finer: (deposit address, ERC-20). So what about the following approach:

1. Change `deposit_erc20` so that it takes as parameter also the ERC-20 to be deposited. In terms of UX this is similar I think to what CEX offer when you click on deposit (just doubled-check with Kraken):
    1. Select the token to be deposited, e.g. USDC
    2. (Select the network. Here there is no choice, it's Ethereum)
    3. Display deposit address (and QR code)
2. Refactor the processing to be per (deposit address, ERC-20). Also limit the number of parallel deposits that can be made by the same user (say at most 5 ERC-20).

I would still merge this PR and build on top that refactoring to reuse the infrastructure that we have. WDYT?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: The idea behind scan_count is to track the number of calls made so that the user can be charged appropriately upon minting. You are right that now there is no consumers for it. I like your `SweepState`! Based on my previous [comment](https://github.com/dfinity/ic/pull/10946#discussion_r3773474436) I would not do that change yet because I'm more and more convinced that the granularity of the deposit flow from the start should be (user account, ERC-20)

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -223,23 +225,27 @@ fn should_flag_only_deposits_at_or_above_the_per_token_minimum() {
...
-        2,
-        "only the 20 USDT and 15 USDC deposits clear the per-token minimum; the 1 USDT does not"
+    let deadline = Duration::from_secs(180);
+    assert_matches!(
+        setup.await_scan(setup.depositor(1), DEPOSIT_SUBACCOUNT, deadline).status,
+        DepositStatus::AwaitingSweep(detected)
+            if detected.len() == 1
+                && detected[0].token == SupportedToken::CkUsdt.contract().to_string()
+                && detected[0].scanned_balance == USDT_ABOVE_MINIMUM
+                && detected[0].detected_at_block > 0_u8
```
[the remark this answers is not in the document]
gregorydemay: 🤖 This one is a false positive — `candid::Nat` does implement `PartialOrd<u8>`, so both lines compile as written.

The impl is generated by macro rather than written out, which is likely why it isn't obvious from the source. In `candid-0.10.34/src/types/number.rs`:

```rust
define_ops!( Nat, usize u8 u16 u32 u64 u128 );
```

and `define_ops!` expands `define_ord!($t, $f)` for each scalar, which emits both directions:

```rust
impl PartialOrd<$scalar> for $res { ... }
impl PartialOrd<$res> for $scalar { ... }
```

So `Nat: PartialOrd<u8>` exists, alongside the matching `PartialEq` from `define_eq!` — which is what the neighbouring `scanned_balance == USDT_ABOVE_MINIMUM` assertions already rely on.

Confirmed empirically as well: `bazel test //rs/ethereum/cketh/minter:deposit_from_cex` builds and passes on the current head.

Related: this is the same mechanism noted earlier in this PR when the unused direct `candid` dependency was dropped from the `deposit_from_cex` target — the `Nat` comparisons resolve through trait impls without the test crate depending on `candid` directly.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Good call — I opened [DEFI-2976](https://dfinity.atlassian.net/browse/DEFI-2976) under the DEFI-2917 epic to track this refactor to per-(deposit address, ERC-20) granularity (token-parameterized `deposit_erc20` + per-user concurrency cap), building on top of this PR. Leaving this thread unresolved as a pointer.

mbjorkqvist: Sounds good to me!

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -186,6 +246,46 @@ impl AutomaticDeposits {
     pub fn watchlist_len(&self) -> usize {
         self.watchlist.len()
     }
+
+    pub fn sweep_len(&self) -> usize {
+        self.sweep.values().map(Vec::len).sum()
+    }
```
mbjorkqvist, line 252: nit: `sweep_len` is currently only used by tests (`cketh_minter_sweep_queue_size` used to call it, but was removed when the metrics changes were deferred). It will probably be used again as part of that work, so leaving it here should be fine, just wanted to bring it up in case you prefer to remove it now and re-introduce it later (if necessary).

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: I will leave it for now to avoid one additional round of CI.

PR #10947, opened 2026-07-29
test(cketh): end-to-end balance scan against a live anvil node

gregorydemay:
Adds a single end-to-end test proving the ckERC20 balance scan works against a real EVM, complementing the existing mock-based integration tests.

The test drives the full production path — minter → **real EVM RPC canister** → Ethereum — with no JSON-RPC mocking:

- A new PocketIC *live* harness in the cketh `test_utils` crate owns a local `anvil` node and installs the minter and the EVM RPC canister, configuring the latter with an `overrideProvider` that routes every provider to that node. In live mode the EVM RPC canister issues genuine HTTPS outcalls that reach anvil for real. The new harness is required because current integration tests infrastructure uses the `StateMachine` that cannot issue real HTTP request. Migrating the integration test infrastructure to PocketIC is deferred to DEFI-2262 (see #10949).
- Supported tokens (ckUSDC, ckUSDT) are registered directly through the minter's `add_ckerc20_token` endpoint by pointing its orchestrator id at a principal the harness controls, so no real orchestrator or spawned ledgers are needed.
- The harness places the two tokens at their real mainnet addresses on anvil (`anvil_setCode`) and credits the minter's derived deposit address above the candidate threshold (`anvil_setStorageAt`). The test then asserts the minter's periodic scan reads those real balances and flags the address as a deposit candidate for both tokens.

Stacked on top of #10873.


rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -0,0 +1,271 @@
...
+                        .unwrap_or(0.0) as u64,
+                };
+            }
+            assert!(
+                start.elapsed() <= deadline,
+                "the balance scan did not run within {deadline:?}"
+            );
+            std::thread::sleep(Duration::from_secs(2));
+        }
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch. Fixed by making the funding atomic w.r.t. the scan: `credit_deposit` now writes every token's balance *before* placing any code. The batcher is fail-loud — an `eth_call` reverts (and the address is *not* advanced) unless every token in the batch has code — so the first scan that advances the address necessarily runs after all balances are already in place, making `candidates_found` deterministic. (The metrics-polling wait was also replaced by polling `deposit_erc20`'s own `scan_count`.)

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -180,266 +177,43 @@ fn should_revert_the_whole_call_when_a_token_is_not_a_contract() {
...
+#[test]
+fn should_scan_real_erc20_balances_through_the_evm_rpc_canister() {
+    const DEPOSIT_SUBACCOUNT: [u8; 32] = [42; 32];
+    // 20 USDC/USDT (6 decimals), comfortably above each token's ~$10 candidate minimum.
+    const DEPOSIT_BALANCE: u128 = 20_000_000;
+
+    let setup = CkErc20LiveScanSetup::new_live();
+    let user = setup.caller();
+    let deposit = setup.register_deposit_address(user, DEPOSIT_SUBACCOUNT);
+    setup.credit_deposit(&deposit, DEPOSIT_BALANCE);
```
mbjorkqvist: IIUC, this deposits 20 USDC and 20 USDT using the same `DEPOSIT_BALANCE` value. Would it make sense to have different amounts for the two tokens, so that we could detect if the balances were swapped USDC<->USDT, and also have one of them be below `MIN_DEPOSITS` to also detect this error case (perhaps in a separate test)?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Good idea — done in 14eb493934.

The live test now drives three independent depositors: **20 USDT** and **15 USDC** (both candidates) and **1 USDT** (below the ~$10 minimum), so the per-token minimum is exercised end-to-end. The `candidates == 2` assertion now discriminates the below-minimum case — a regressed filter that counted the 1 USDT deposit would report 3.

On detecting a USDC↔USDT swap via the distinct amounts: the live harness currently observes only the candidate *count* from the scan logs, so that assertion is not yet possible without surfacing the scanned balance. Surfacing the per-candidate balance is coming in a follow-up observability PR; the distinct amounts are already in place here, so swap detection can be asserted on top with no rework. In the meantime, positional token↔balance attribution is covered at the byte level by `should_read_erc20_balances_across_tokens_and_holders`.

gregorydemay:
🤖 Thanks — addressed both points from the Copilot review in b0e768e301:

- **No HTTP timeout on the blocking anvil client** (`anvil.rs`): added an explicit per-request timeout via a shared `rpc_client()` builder, used by both `rpc_result` and `wait_until_ready`. A wedged node now fails fast instead of hanging a `send()` past the readiness deadline.
- **"HTTPS outcalls" wording** (`live_scan.rs`, `deposit_from_cex.rs`): reworded to say the EVM RPC canister makes genuine outcalls through the IC's HTTPS-outcalls *feature*, reaching the local anvil node over HTTP — dropping the misleading "HTTPS" wire description.

Live `deposit_from_cex` test still passes locally.

gregorydemay:
🤖 Addressed both points from the latest Copilot review in b508bce460:

- **Client reuse** — the blocking JSON-RPC client is now built once and stored on `Anvil`, so RPCs share a connection pool instead of constructing a fresh client (and TCP connection) per call.
- **`rpc_result` honoring its `Result`** — transport and decode failures (including an `RPC_TIMEOUT` timeout) now return `Err` tagged with the failing method instead of `unwrap()`-panicking, so callers like `eth_call_create` can distinguish a transient failure. Existing behavior is unchanged: a revert still surfaces via the JSON-RPC `error` body.

Live `deposit_from_cex` test still passes locally.

PR #10949, opened 2026-07-29
test(ledger-suite-orchestrator): migrate integration tests to PocketIC

gregorydemay:
Migrates the ledger-suite-orchestrator (LSO) integration tests from the internal `ic-state-machine-tests` harness to PocketIC, following the pattern already used for the EVM-RPC canister (dfinity/evm-rpc-canister#294). This is DEFI-2262: a pure refactor of the test harness — same tests, same coverage, same semantics.

`CkErc20Setup` in `rs/ethereum/cketh` embeds LSO's fixture and shares a `StateMachine` with it, so flipping LSO's fixture in place would break cketh in the same commit. To keep this change single-component and the tree green, this PR **adds** a PocketIC fixture in LSO's `test_utils` alongside the existing StateMachine one (precedent: `rs/ledger_suite/icp/test_utils`'s `state_machine_helpers`/`pocket_ic_helpers` coexisting) and migrates only `ledger-suite-orchestrator/tests/tests.rs` to it. The StateMachine fixture stays untouched at its current paths and is still exercised by `cketh`; it is deleted only in the last PR of this stack, once cketh no longer needs it.

## Summary of changes
- New `pocket_ic` module in `ledger-suite-orchestrator/test_utils` (`LedgerSuiteOrchestrator`, `AddErc20TokenFlow`, `ManagedCanistersAssert`, `UniversalCanister`, `new_pocket_ic()`), mirroring the existing StateMachine-based types and method names one-for-one so later PRs in the stack only need to swap import paths.
- `new_pocket_ic()` builds a system subnet (LSO tests assert exact cycle balances, which only hold on a non-charging subnet) with canister-execution rate limiting disabled (LSO repeatedly installs/upgrades ledger, index and archive canisters).
- Metrics assertions now go through `ic-metrics-assert`'s `PocketIcHttpQuery` (feature `pocket_ic`) instead of a hand-rolled `CanisterHttpQuery` impl.
- `tests/tests.rs` fully migrated to the new fixture.
- `Cargo.toml`/`BUILD.bazel` updated accordingly; `test_utils` keeps `ic-state-machine-tests` for now (still needed by the old fixture).

## Coverage
- All spec requirements from the PR are covered: PocketIC fixture added alongside the old one, `tests/tests.rs` fully migrated, old fixture/paths untouched, `cketh` still builds against the untouched fixture.

## 📚 PR stack

1. **#10949 — test(ledger-suite-orchestrator): migrate integration tests to PocketIC** 👈 you are here
2. #10950 — test(cketh): deduplicate signed-transaction literals in integration tests
3. #10955 — test(cketh): migrate integration tests to PocketIC
4. ~~#10956~~ — test(ledger-suite-orchestrator): drop the StateMachine test fixture (wrongly marked merged; its content is part of #10955)


rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub embedded_archive_wasm_hash: WasmHash,
+}
+
+impl Default for LedgerSuiteOrchestrator {
+    fn default() -> Self {
+        Self::new(Arc::new(new_pocket_ic()), crate::default_init_arg()).register_embedded_wasms()
+    }
+}
+
+impl AsRef<PocketIc> for LedgerSuiteOrchestrator {
```
gregorydemay: 🤖🧐 🟠 **Medium** — dead code (G9/G12). This `AsRef<PocketIc>` impl existed in the StateMachine fixture only to satisfy the `T: AsRef<StateMachine>` bounds on `stop_canister`/`out_of_band_upgrade` and the manual `http_query`. In the port those became `&PocketIc` parameters and `PocketIcHttpQuery`, so nothing calls this impl any more (`grep -rn "as_ref()" test_utils/src/pocket_ic tests/tests.rs` shows only `AsRef<[u8]>` uses; `rs/ethereum/cketh` does not use it either). Please drop it rather than carrying leftover scaffolding through to PR 4.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
```diff
@@ -0,0 +1,461 @@
...
+        }
+    }
+}
+
+pub struct ManagedCanistersAssert {
+    pub setup: LedgerSuiteOrchestrator,
+    pub canister_ids: ManagedCanisterIds,
+}
+
+impl AsRef<PocketIc> for ManagedCanistersAssert {
```
gregorydemay: 🤖🧐 🟠 **Medium** — dead code (G9/G12), same as the `AsRef<PocketIc>` impl on `LedgerSuiteOrchestrator`. `stop_ledger`/`stop_index`/`ledger_out_of_band_upgrade`/`index_out_of_band_upgrade` now pass `&self.setup.env` directly and the metrics path goes through `PocketIcHttpQuery`, so this impl has no remaining caller.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+        Self::new(Arc::new(new_pocket_ic()), init_arg)
+    }
+
+    pub fn new(env: Arc<PocketIc>, init_arg: InitArg) -> Self {
+        let ledger_suite_orchestrator_id = env.create_canister();
+        env.add_cycles(ledger_suite_orchestrator_id, u128::MAX);
+        Self {
+            env,
+            ledger_suite_orchestrator_id,
+            ledger_suite_orchestrator_wasm: crate::ledger_suite_orchestrator_wasm(),
```
gregorydemay: 🤖🧐 🔵 **Nit** — inconsistent import style (G11/R4). This module reaches for crate-root helpers via inline `crate::` paths in 13 places (`crate::default_init_arg()`, `crate::ledger_wasm()`, `crate::GIT_COMMIT_HASH`, `crate::MAX_TICKS`, …) while the sibling `flow.rs` imports the same kind of item with `use crate::MINTER_PRINCIPAL;`. A single `use crate::{...}` block at the top would match the file's own convention and, at line 308, restore the inline-capture panic message `"Failed to get result after {MAX_TICKS} ticks: …"` instead of the positional-argument rewrite.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+        self.env.tick(); //tick before upgrade to finish current timers which are reset afterwards
+        self.env.upgrade_canister(
+            self.ledger_suite_orchestrator_id,
+            self.ledger_suite_orchestrator_wasm.clone(),
+            Encode!(upgrade_arg).unwrap(),
+            None,
+        )
+    }
+
+    pub fn get_canister_status(&self) -> ic_cdk::management_canister::CanisterStatusResult {
```
gregorydemay: 🤖🧐 🔵 **Nit** — `ic_cdk::management_canister::CanisterStatusResult` is spelled out twice here (return type and `Decode!` arm) because the short name now collides with the `ic_management_canister_types::CanisterStatusResult` imported at the top. An aliased import (`use ic_cdk::management_canister::CanisterStatusResult as CdkCanisterStatusResult;`) reads better and keeps the boundary types explicit (R4).

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
```diff
@@ -0,0 +1,461 @@
...
+        .expect("failed to decode transfer response")
+    }
+
+    pub fn upgrade_ledger_to_change_archive_options(&self, archive_options: ChangeArchiveOptions) {
+        use ic_icrc1_ledger::{LedgerArgument, UpgradeArgs as LedgerUpgradeArgs};
+
+        let module_hash_before = self
+            .ledger_canister_status()
+            .module_hash
+            .clone()
```
gregorydemay: 🤖🧐 🔵 **Nit** — redundant `.clone()` (R3). `ledger_canister_status()` returns an owned `CanisterStatusResult`, so `module_hash` can be moved straight out of the temporary; the same applies at line 280. Every `.clone()` should have a reason.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
```diff
@@ -0,0 +1,461 @@
...
+            .collect()
+    }
+}
+
+impl PocketIcHttpQuery for ManagedCanistersAssert {
+    fn get_pocket_ic(&self) -> &PocketIc {
+        self.setup.get_pocket_ic()
+    }
+
+    fn get_canister_id(&self) -> ic_management_canister_types::CanisterId {
```
gregorydemay: 🤖🧐 🔵 **Nit** — `ic_management_canister_types::CanisterId` is fully qualified here while `mod.rs` imports the same type by name. Prefer a plain `use` for consistency (G11/R4).

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub fn check_metrics(self) -> MetricsAssert<Self> {
+        MetricsAssert::from_http_query(self)
+    }
+
+    pub fn trigger_creation_of_archive(self) -> Self {
+        const ARCHIVE_TRIGGER_THRESHOLD: usize = 10;
+
+        // The productive value for `trigger_threshold` is `2_000`,
+        // which would require `2_000` transfers to trigger the creation of an archive,
+        // which would take in the order of 20s (order of magnitude is 10ms per transfer with state machine tests).
```
gregorydemay: 🤖🧐 🔵 **Nit** — stale comment (C2). This line was copied verbatim from the StateMachine fixture and still says "10ms per transfer with state machine tests" inside the module that specifically does *not* use StateMachine. Either drop the parenthetical or re-anchor it to PocketIC.

rs/ethereum/ledger-suite-orchestrator/BUILD.bazel
```diff
@@ -143,6 +143,7 @@ rust_ic_test(
     data = [
         "ledger_suite_orchestrator_canister.wasm.gz",
         "ledger_suite_orchestrator_canister_getblocksdisabled.wasm.gz",
+        "//:pocket-ic-server",
```
gregorydemay, line 146: 🤖🧐 🔵 **Nit** — `//:pocket-ic-server` is declared in both `data` (here) and `deps` (line 168). One is enough: `rs/ledger_suite/icp/BUILD.bazel` lists it in `data` only, `rs/bitcoin/checker` and `rs/dogecoin/ckdoge/minter` in `deps` only. Since the target is a runtime binary reached via `POCKET_IC_BIN`, `data` alone is the more accurate declaration.

gregorydemay:
🤖🧐 **VERDICT: CHANGES_REQUESTED** — 0 blockers, 2 mediums, 6 nits; CI **pending** (22 pass / 7 pending / 0 red — `Bazel Test All`, `Bazel Test arm64-{linux,darwin}`, `Cargo Build/Lint Linux`, `Build IC`, `RBE Bazel Test All` still running).

Neither medium is a semantics problem: the port itself is faithful. Both are the same finding — two `AsRef<PocketIc>` impls that lost their only callers in the translation and would otherwise be carried through to PR 4.

<details>

### Verified by running

| Command | Result |
|---|---|
| `bazel test //rs/ethereum/ledger-suite-orchestrator:integration_tests --nocache_test_results` (in dev container) | **PASSED** in 16.4s, 1/1 |
| `cargo check --all-targets --all-features -p ic-ledger-suite-orchestrator-test-utils -p ic-ledger-suite-orchestrator` | clean |
| `gh pr checks 10949` | 22 pass, 7 pending, 10 skipping, 0 failing |

(Plain `bazel test` on the host fails in `jemalloc`'s `autoconf` step — pre-existing local-infra issue, unrelated to this PR.)

### Faithfulness of the port
- **Test set unchanged**: 21 `#[test]` before and after, identical function-name sets (`diff` of extracted `fn` names is empty). No test dropped, none weakened.
- **API mapping** matches the spec table throughout: `execute_ingress`→`update_call(_, Principal::anonymous(), …)`, `execute_ingress_as(pid, …)`→`update_call(c, pid.0, …)`, `query`→`query_call(_, anonymous, …)`, `send_ingress`/`await_ingress(id, MAX_TICKS)`→`submit_call`/`await_call`, `e.code()`/`e.description()`→`e.error_code`/`e.reject_message`, `create_canister_with_cycles(u128::MAX)`→`create_canister()`+`add_cycles(_, u128::MAX)`, `execute_ingress_as(ctrl, ic_00, "install_code", …)`→`update_call_with_effective_principal(management_canister, RawEffectivePrincipal::CanisterId(target), ctrl, …)`, `canister_status_as`→`canister_status(id, Some(sender))`.
- **Tick semantics preserved**: `advance_time_for_periodic_tasks` / `advance_time_for_upgrade` keep the identical 6-tick sequences; `wait_for` still bounds at `MAX_TICKS`; the "tick before upgrade to finish current timers" tick is retained in both upgrade paths.
- **Assertion strength preserved** where the shape changed: `assert_matches!(stop_res, Ok(WasmResult::Reply(_)))` → `.expect("failed to stop canister")` and `assert_reply(res.unwrap())` → `.expect(…)` are equivalent under pocket-ic's `Result<Vec<u8>, RejectResponse>` (a canister reject surfaces as `Err`, so the reject case still fails the test).
- **Metrics**: the hand-rolled `CanisterHttpQuery<UserError>` impls are replaced by `PocketIcHttpQuery`, whose blanket impl is exactly `query_call(canister_id, Principal::anonymous(), "http_request", request)` — same endpoint, same sender, same canister (orchestrator) for both `LedgerSuiteOrchestrator` and `ManagedCanistersAssert`.
- **Hard-coded canister ID** `ryjl3-…-aaaba-cai` → `xbze6-vd777-77777-aaaba-cai`: not vacuous. The value is compared against the index ID the orchestrator actually returns, and independently against the ledger's `ICRC106_INDEX_PRINCIPAL` metadata — two live assertions that cross-check each other. Same fragility class as the value it replaced.
- **Explicit controller** in `should_upgrade_canisters_managed_but_not_installed_by_orchestrator`: `pre_existing_controller` is captured from the temporary orchestrator that created those canisters, so it is genuinely a controller — this correctly replaces StateMachine's auto-pick now that PocketIC enforces the check.
- **`Nat` cycle arithmetic** in `should_top_up_spawned_canisters`: the deltas assert the same `TEN_TRILLIONS` increments; the added `.clone()`s are required because the balances are reused as the baseline for the second top-up.

### Scope / boundary
- The single commit touches only `Cargo.lock` + `rs/ethereum/ledger-suite-orchestrator/**`. ✔
- Old StateMachine fixture: `test_utils/src/flow.rs` and `test_utils/src/universal_canister.rs` are byte-identical to master; `test_utils/src/lib.rs` gains exactly one line (`pub mod pocket_ic;`). ✔
- Build config matches the spec: `integration_tests` drops `//rs/state_machine_tests`, gains `POCKET_IC_BIN` + `//packages/pocket-ic`; `test_utils` keeps `ic-state-machine-tests` and moves to `ic-metrics-assert_pocket_ic` (the `pocket_ic` feature is purely additive, so the old fixture's `CanisterHttpQuery` impl still compiles). ✔

- **Duplication**: ~915 lines of fixture are duplicated against `test_utils/src/{lib,flow,universal_canister}.rs`. This is the plan's deliberate, documented trade-off (`CkErc20Setup` shares an `Arc<StateMachine>` with LSO, so an in-place flip would break cketh in the same commit) with deletion scheduled for PR 4 — accepted, not counted against the verdict. I considered the alternative of abstracting both harnesses behind an env trait to avoid the copy, and it is clearly worse here: it would rewrite fixture code that cketh currently depends on, and the abstraction would itself be deleted in PR 4.
- **Structural duplication vs. the wider repo**: none beyond the intended mirror; the shape follows the established `rs/ledger_suite/icp/test_utils` precedent.
- **Unused derives**: none — no new types with derives are introduced.
- **Primitive-obsession parameters**: none introduced; the `CanisterId`/`PrincipalId` → `candid::Principal` change is the harness's own domain type, not a de-typing.
- **Divergent invariant handling**: cleared — `.expect(…)` is used uniformly on the new call paths; no site panics where a sibling swallows.
- **Silent fallbacks**: none — no `unwrap_or_default()`, `Result::ok()` discard or `let _ =` on a failure path. `PocketIc::install_canister` returns `()` and panics internally, so dropping its result is not a swallowed error.
- **Test-only code in production modules**: n/a, this is all `test_utils`.
- **Comment minimalism**: one stale comment carried over (flagged inline at `flow.rs:94`); no requirement-ID tags, no JIRA references, no commented-out code.

### Test pyramid
Not applicable as a coverage question — this is a harness migration with no production-behaviour change, so "a new test per behaviour change" is satisfied by the existing suite continuing to pass unmodified in content.

</details>

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
gregorydemay: 🤖 Fixed in 223a4482b3 — dropped the dead `impl AsRef<PocketIc> for LedgerSuiteOrchestrator`.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
gregorydemay: 🤖 Fixed in 223a4482b3 — dropped the dead `impl AsRef<PocketIc> for ManagedCanistersAssert`.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
gregorydemay: 🤖 Fixed in 759c6466e8 — consolidated the 13 inline `crate::` paths into a single `use crate::{...}` block, and `wait_for`'s panic message is back to inline-capture (`"...{MAX_TICKS} ticks..."`).

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/mod.rs
gregorydemay: 🤖 Fixed in 759c6466e8 — aliased `ic_cdk::management_canister::CanisterStatusResult` as `CdkCanisterStatusResult` at the import, `get_canister_status` now just uses the short alias.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
gregorydemay: 🤖 Fixed in 759c6466e8 — `flow.rs` now imports `ic_management_canister_types::CanisterId` instead of qualifying it inline.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
gregorydemay: 🤖 Fixed in d0af1f47a5 — dropped both redundant `.clone()`s, `module_hash` is moved straight out of the owned `ledger_canister_status()` temporary.

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
gregorydemay: 🤖 Fixed in 52b82fe587 — dropped the stale "state machine tests" parenthetical rather than guessing a PocketIC-specific number.

rs/ethereum/ledger-suite-orchestrator/BUILD.bazel
gregorydemay: 🤖 Fixed in de18dda2f9 — dropped `//:pocket-ic-server` from `deps`, kept it in `data` only. Re-ran `bazel test //rs/ethereum/ledger-suite-orchestrator:integration_tests` to confirm `POCKET_IC_BIN` still resolves and all 21 tests pass.

rs/ethereum/ledger-suite-orchestrator/tests/tests.rs
```diff
@@ -44,7 +43,7 @@ fn should_spawn_ledger_with_correct_init_args() {
         })
         .expect_new_ledger_and_index_canisters();
     let index_id = managed_canisters_assert.canister_ids.index.unwrap();
-    assert_eq!(index_id, "ryjl3-tyaaa-aaaaa-aaaba-cai".parse().unwrap());
+    assert_eq!(index_id, "xbze6-vd777-77777-aaaba-cai".parse().unwrap());
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Leaving this as-is. The two literals are not vacuous: they are independently cross-checked against each other (the `index_id` returned by the orchestrator on line 46, and the ledger's own `ICRC106_INDEX_PRINCIPAL` metadata on line 91-93 must both equal this value), so the assertion still fails if the wiring is wrong even though the concrete text is PocketIC-version-dependent. This is the same fragility class as the StateMachine-era literal it replaces (`ryjl3-tyaaa-aaaaa-aaaba-cai`), and the migration plan for this PR stack explicitly accepts pocket-ic-server-version-dependent constants (see PR 3's derived-key-constant regeneration). Asserting against the live `index_id` instead would make the test tautological for that half of the check.

gregorydemay:
🤖🧐 **VERDICT: READY** (re-review of `de18dda2f9`) — 0 blockers, 0 mediums, 0 nits outstanding; all 8 findings resolved. CI **pending, nothing red** (13 pass / 11 pending / 9 skipping / 0 failing) — ⚠️ **merge only once CI completes green**; I have not verified the pending jobs.

<details>
<summary>Re-review details</summary>

### Per-finding resolution

| # | Sev | Finding | Commit | Status |
|---|---|---|---|---|
| 1 | 🟠 | Dead `AsRef<PocketIc> for LedgerSuiteOrchestrator` | `223a4482b3` | ✅ impl removed; `PocketIc` import still earns its place (`Arc<PocketIc>` field, `get_pocket_ic`, `new_pocket_ic`, `out_of_band_upgrade`) |
| 2 | 🟠 | Dead `AsRef<PocketIc> for ManagedCanistersAssert` | `223a4482b3` | ✅ impl removed; `PocketIc` still used by `get_pocket_ic` |
| 3 | 🔵 | 13 inline `crate::` paths + positional panic message | `759c6466e8` | ✅ single `use crate::{...}` block; `wait_for` back to inline-capture `"…{MAX_TICKS} ticks…"` |
| 4 | 🔵 | `ic_cdk::…::CanisterStatusResult` spelled twice | `759c6466e8` | ✅ aliased `CdkCanisterStatusResult`, used in both the return type and the `Decode!` arm |
| 5 | 🔵 | Redundant `.clone()` on `module_hash` | `d0af1f47a5` | ✅ both dropped; `module_hash` moves out of the owned temporary |
| 6 | 🔵 | Fully-qualified `ic_management_canister_types::CanisterId` | `759c6466e8` | ✅ imported by name, matching `mod.rs` |
| 7 | 🔵 | Stale "state machine tests" comment | `52b82fe587` | ✅ parenthetical dropped; remaining comment reads cleanly and states no harness-specific timing |
| 8 | 🔵 | `//:pocket-ic-server` in both `data` and `deps` | `de18dda2f9` | ✅ kept in `data` only; `POCKET_IC_BIN` still resolves (test re-run confirms) |

Copilot's thread on `tests.rs:46` (fixed canister principal): the "leave as-is" reply matches what I independently validated in the first pass — the two literals are cross-checked against each other (orchestrator-returned index ID vs. the ledger's `ICRC106_INDEX_PRINCIPAL` metadata), so neither is vacuous, and the coupling is the same class the StateMachine fixture already had. No objection.

### Verified by running (on `de18dda2f9`)

| Command | Result |
|---|---|
| `bazel test //rs/ethereum/ledger-suite-orchestrator:integration_tests --nocache_test_results` (dev container) | **PASSED** in 15.2s, 1/1 |
| `cargo check --all-targets --all-features -p ic-ledger-suite-orchestrator-test-utils -p ic-ledger-suite-orchestrator` | clean, no warnings |
| `git diff --stat 522634fb44 de18dda2f9` | 10 files, all under `rs/ethereum/ledger-suite-orchestrator/` + `Cargo.lock` — component boundary still respected |
| old-fixture check | `test_utils/src/{flow,universal_canister}.rs` still byte-identical to master; `lib.rs` still +1 line (`pub mod pocket_ic;`) |

### Nothing new introduced
The five commits are strictly subtractive or rename-only — no call sites, senders, tick sequences, or assertions were touched. The faithfulness conclusions from the first pass (21/21 tests, identical function-name set, API mapping per spec, preserved tick semantics and assertion strength) still hold; the net effect on the diff is −14 lines.

One forward-looking note for PR 4: consolidating into a single `use crate::{...}` block makes the hoist *easier* than the scattered `crate::` prefixes did — if `pocket_ic/mod.rs` is merged into `lib.rs`, that one import line is deleted rather than 13 path prefixes rewritten. No action needed here.

### Outstanding
Only CI. Nothing is red, but `Bazel Test All`, `Bazel Test arm64-{linux,darwin}`, `Cargo Build/Lint Linux`, `Build IC`, `Lock Generate`, `Candid compatibility checks`, `Autofix`, `Bazel Run Fuzzers` and `RBE / Infer Bazel Targets` were still running when this verdict was written. Final approval and merge are the human's call.

</details>

rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs
```diff
@@ -0,0 +1,452 @@
...
+                Encode!(&InstallCodeArgs {
+                    mode: CanisterInstallMode::Upgrade(None),
+                    canister_id: self.ledger_canister_id(),
+                    wasm_module: crate::ledger_wasm().to_bytes(),
+                    arg: Encode!(&upgrade_args).unwrap(),
+                    sender_canister_version: None,
+                })
+                .unwrap(),
+            )
+            .expect("failed to change archive options");
```
gregorydemay, line 258: why not just use `PocketIc::upgrade_canister`?

gregorydemay: 🤖 Good catch — done in `e0c66218a1`. It is exactly equivalent: `PocketIc::upgrade_canister` builds the same `install_code` argument with the same `CanisterInstallMode::Upgrade(None)` (`packages/pocket-ic/src/nonblocking.rs:1340`), so the hand-assembled version bought nothing.

The manual form was a faithful port of the StateMachine fixture, which had to address the management canister itself to upgrade as the controller. I applied it to the second occurrence too — `out_of_band_upgrade` in `pocket_ic/mod.rs` had the same shape. Net −36/+9 lines, and it dropped four now-orphaned imports (`InstallCodeArgs`, `CanisterInstallMode`, `RawEffectivePrincipal`). `//rs/ethereum/ledger-suite-orchestrator:integration_tests` still passes.

mbjorkqvist (approved):

PR #10950, opened 2026-07-29
test(cketh): deduplicate signed-transaction and transaction-hash literals in integration tests

gregorydemay:
## Summary
Small preparatory cleanup for the cketh integration tests ahead of the PocketIC migration (DEFI-2262). No framework change here — tests still run on `ic-state-machine-tests` and stay green.

- Deduplicated the inline raw-transaction and transaction-hash string literals in `minter/tests/cketh.rs` that were byte-identical to the `DEFAULT_WITHDRAWAL_TRANSACTION` / `DEFAULT_WITHDRAWAL_TRANSACTION_HASH` constants already exported by `cketh/test_utils`.

Since PR 3 regenerates these signatures for the PocketIC fixture, keeping every occurrence tied to the single constant means the regeneration only needs to touch one place instead of risking a missed or inconsistent literal.

## 📚 PR stack

1. #10949 — test(ledger-suite-orchestrator): migrate integration tests to PocketIC
2. **#10950 — test(cketh): deduplicate signed-transaction literals in integration tests** 👈 you are here
3. #10955 — test(cketh): migrate integration tests to PocketIC
4. ~~#10956~~ — test(ledger-suite-orchestrator): drop the StateMachine test fixture (wrongly marked merged; its content is part of #10955)


rs/ethereum/cketh/minter/tests/cketh.rs
```diff
@@ -167,8 +167,7 @@ fn should_deposit_and_withdraw() {
                     effective_gas_price: Nat::from(4277923390_u64),
                     gas_used: Nat::from(21_000_u32),
```
gregorydemay, line 168: 🤖🧐 🔵 **Nit** — same G5 dedup, one step further: `effective_gas_price: Nat::from(4277923390_u64)` and `gas_used: Nat::from(21_000_u32)` duplicate the `EFFECTIVE_GAS_PRICE` (`4_277_923_390`) and `GAS_USED` (`0x5208` = `21_000`) constants that this file *already imports* and already uses at `:117` / `:1240`. Same pattern at `:585-586` and `:692-693` (6 literals total).

Strictly outside PR 2's stated scope — these are not signature-derived, so deduping them would not shrink PR 3's regeneration diff — so take it or leave it; it does not gate the verdict.

If you do take it, note that the `gas_limit: Nat::from(21_000_u32)` occurrences at `:110`, `:536`, `:676` are semantically a *limit*, not *gas used*, and should stay literal.

gregorydemay (commented):
🤖🧐 **VERDICT: READY** — 0 blockers, 0 mediums, 1 nit; CI: no failures, several jobs still pending (`Bazel Test All`, `Cargo Build/Lint Linux`, `RBE`, `arm64-darwin`, `Autofix`) — green CI remains the merge gate.

<details>

### Scope of this review
Delta commit `963b7708` only (this is a stacked PR based on `ic_DEFI-2262_1_lso-pocket-ic-fixture`; PR 1's changes underneath were not re-reviewed). One file touched, `rs/ethereum/cketh/minter/tests/cketh.rs`, +9/−12 — component boundary `rs/ethereum/cketh/**` respected.

### Challenge phase
For "dedup literals that are byte-identical to an already-exported constant", the minimal solution is exactly: add the import, replace the occurrences, let rustfmt re-wrap. The diff is that and nothing else. No alternative with a better or comparable trade-off exists — a helper fn or a local `const` would both be strictly worse than reusing the single source of truth in `test_utils`.

### Byte-identity verified (not eyeballed)
Extracted every removed literal from the commit and string-compared it against the constant definitions in `rs/ethereum/cketh/test_utils/src/lib.rs:77-80`:

- 2 × `DEFAULT_WITHDRAWAL_TRANSACTION` (238 chars) → exact match
- 4 × `DEFAULT_WITHDRAWAL_TRANSACTION_HASH` (66 chars) → exact match

Exactly 6 literals removed, 6 constant references added. Zero behaviour change.

### Completeness sweep (independent of the author's claim)
Case-insensitive grep over all of `rs/` for the five migration-relevant values — `DEFAULT_WITHDRAWAL_TRANSACTION(_HASH)`, `DEFAULT_CKERC20_WITHDRAWAL_TRANSACTION(_HASH)`, `MINTER_ADDRESS`. **Each now occurs exactly once in the tree: at its own `pub const` declaration.** No duplicates remain, so PR 3's regeneration only has to touch `test_utils/src/lib.rs`.

Correctly left untouched: the *distinct* resubmitted-transaction literals at `cketh.rs:631` (`0x02f873018084625900…`) and `ckerc20.rs:1316` (`0x02f8b0018084625900…`) — different signatures, single-occurrence each, not duplicates. `ckerc20.rs` is not in the delta at all.

Also checked but deliberately not flagged: the `r`/`s` signature components in `test_utils/src/response.rs:136,166` encode the same knowledge as the raw-tx literals, but in a different representation (decimal / hex-radix vs. RLP hex). Collapsing those is a decode, not a mechanical replace — out of scope here, and PR 3 already regenerates them.

- **Duplication**: the PR's purpose; verified eliminated for all in-scope values. One residual G5 instance found (`EFFECTIVE_GAS_PRICE` / `GAS_USED`) — posted inline as a 🔵 nit, out of stated scope.
  - Pre-existing structural duplication noted for the record, **no action requested**: the `assert_has_unique_events_in_order(&[…])` blocks in `should_deposit_and_withdraw` (`:135-172`) and `should_reimburse` (`:553-600`) are ~35 near-identical lines differing along one axis (`TransactionStatus::Success` vs `Failure`, plus the trailing `ReimbursedEthWithdrawal` event). Normally a parameterization candidate, but it predates this PR and restructuring it would *enlarge* PR 3's diff — the opposite of this PR's goal.
- **Unused derives**: n/a — no types added.
- **Primitive-obsession parameters**: n/a — no signatures changed.
- **Divergent invariant handling**: none found; the 6 sites all now go through the same constant, which is strictly *more* uniform than before.
- **Silent fallbacks**: none found; no error-handling paths touched.
- **Test-only code in production modules**: n/a.
- **Docs / comments**: no comments added or removed; no doc drift. Commit message and PR description match the diff exactly.

### Testing
Per the review rules, a behaviour change needs a test — there is no behaviour change here, so no new test is owed. Assertions keep identical values, so their ability to fail on a regression is unchanged (and is now anchored to the constant that PR 3 regenerates, which strictly *improves* it: a future regeneration can no longer leave a stale literal behind silently).

### What I ran
- `rustfmt --check --edition 2024 rs/ethereum/cketh/minter/tests/cketh.rs` → clean.
- `cargo check --all-targets --all-features -p ic-cketh-minter -p ic-cketh-test-utils` → passes.
- `gh pr checks 10950` → no red; pending jobs listed above.
- **Did not re-run** `bazel test //rs/ethereum/cketh/minter:integration_tests` (reported 2/2 green by the implementer). Justification: byte-identity of all 6 replacements was proven programmatically, so the compiled test binary is semantically identical to its parent — there is no concrete reason to distrust the earlier run, and `Bazel Test All` in CI covers it regardless.

</details>

rs/ethereum/cketh/minter/tests/cketh.rs
gregorydemay: 🤖 Acknowledged — leaving the `EFFECTIVE_GAS_PRICE`/`GAS_USED` dedup out of this PR as suggested: those literals are not signature-derived, so folding them in would not shrink the PR 3 regeneration diff, and `gas_limit` occurrences must stay literal to avoid conflating distinct semantics. Can be picked up as a follow-up after the migration stack lands.

mbjorkqvist (approved):
nit: The PR title only mentions *transaction* literals, but the change also includes *transaction hash* literals.

gregorydemay:
🤖 Thanks — fixed the title to `test(cketh): deduplicate signed-transaction and transaction-hash literals in integration tests`. You were right that it undersold the change: of the six replaced literals, four are `DEFAULT_WITHDRAWAL_TRANSACTION_HASH` and only two are the raw transaction.

PR #10953, opened 2026-07-29
chore: only depend on Universal VM or Prometheus images when necessary

basvandijk:
Each local system-test would always depend on both a big UVM and a big P8s image. These dependencies are only necessary when the test actually uses a UVM or P8s VM respectively (these VMs boot those images). So we now conditionally depend on these images.


Bownairo (approved):

gregorydemay (approved):

pierugo-dfinity (approved):

lwshang (approved):

daniel-wong-dfinity-org-twin:
I have attempted to deobfuscate the title. PTAL.

daniel-wong-dfinity-org-twin (commented):
Governance changes are fine. But the title and description of this PR require decryption to understand.

daniel-wong-dfinity-org-twin (approved):
woops. I forgot to select "Approve" a minute ago...

rs/tests/ckbtc/BUILD.bazel
daniel-wong-dfinity-org-twin, line 1: This comment is for the PR description, but to make this blocking, I'm attaching it to a file as a work around to how GH "works".

Can you briefly explain what these images are? Like, what is the special stuff in them that's often not needed? How would I know when they be (not) needed?

basvandijk: The UniversalVM and the PrometheusVM boot the `@farm_universal_vm_img` and `@farm_prometheus_vm_img` images respectively (they're defined [here](https://github.com/dfinity-lab/farm/blob/master/universal-vm/default.nix#L71-L76)). So your test needs those images when it deploys a UVM or a PrometheusVM. When it doesn't deploy these VMs it doesn't need those images.

nmattia (approved):

PR #10954, opened 2026-07-29
feat: Eagerly install rustc and friends in the dev container.

daniel-wong-dfinity-org-twin:
This fixes a problem for me: when I develop in VS Code with the rust-analyzer extension, the extension fails because there is no rustc & co. For whatever reason, the extension decides, "if there is no rustc, I refuse to install it". This works around that.


basvandijk:
For me https://github.com/dfinity/ic/pull/10822 fixed it. Doesn't that work for you?

ci/container/init.sh
```diff
@@ -37,4 +37,10 @@ if [ ! -d /tmp/zig-cache ]; then
     sudo chown "$(id -u)":"$(id -g)" /tmp/zig-cache
 fi
 
+# Installs rustc and friends (collectively, they form a "toolchain"). The
+# version of Rust is determined by the /ic/rust-toolchain.toml (and of course,
+# the target platform is this one, x86-64 + Linux). Installing pre-emptively
+# here avoids the rust-analyzer VS Code extension failing.
+rustup toolchain install --no-self-update 2>/dev/null || true
```
basvandijk, line 44: One issue with running this in `ci/container/init.sh` is that it will also [run on CI](https://github.com/dfinity/ic/blob/0ac65bf3e16dedcafbe6d2d43c35751bde4d1129/.github/workflows/ci-main.yml#L295) where we would rather not install the rust toolchain via rustup since we already have bazel.

tmu0 (approved):

PR #10955, opened 2026-07-29
test(cketh): migrate integration tests to PocketIC

gregorydemay:
## 📚 PR stack

1. ~~#10949~~ — test(ledger-suite-orchestrator): migrate integration tests to PocketIC (merged)
2. ~~#10950~~ — test(cketh): deduplicate signed-transaction and transaction-hash literals (merged)
3. **#10955 — test(cketh): migrate integration tests to PocketIC** 👈 you are here
4. ~~#10956~~ — test(ledger-suite-orchestrator): drop the StateMachine test fixture — **its content is part of this PR**, see below

Follow-up, based on this PR: #11123 (PocketIC server lifetime).

## Summary

Flips `rs/ethereum/cketh/test_utils` and the minter's `cketh.rs`/`ckerc20.rs` integration tests from `ic-state-machine-tests` onto the sync `pocket_ic` client, following the pattern established for the ledger-suite-orchestrator in #10949.

The minter is now initialized against the fiduciary subnet's `"key_1"` ECDSA key (StateMachine's `"master_ecdsa_public_key"` isn't available on PocketIC's fiduciary subnet), which changes the minter's derived Ethereum address. Every signature-derived constant is regenerated as a consequence: `MINTER_ADDRESS`, `DEFAULT_WITHDRAWAL_TRANSACTION(_HASH)`, `DEFAULT_CKERC20_WITHDRAWAL_TRANSACTION(_HASH)`, the r/s of the default ERC20 signed transaction, the per-account deposit address in `should_record_address_to_deposit`, and the resubmitted-transaction literal in `should_resubmit_new_transaction_with_same_max_fee_per_gas_when_price_increased`.

Canister-http mocking is rebuilt around PocketIC's `get_canister_http()` / `mock_canister_http_response()`, preserving the legacy `"Http body exceeds size limit of {n} bytes."` reject that the minter's response-size check matches on.

`tests/dump_stable_memory.rs` and `tests/deposit_from_cex_demo.rs` are untouched.

## Also included: dropping the StateMachine fixture from the orchestrator

This was meant to be a separate PR (#10956) reviewed on top of this one. It was marked merged by mistake — see [the note there](https://github.com/dfinity/ic/pull/10956#issuecomment-5271859102) — and since GitHub cannot reopen a merged PR, its content stays here rather than being shuffled again.

With the ckETH minter's tests on PocketIC, nothing in `rs/ethereum` uses the orchestrator's StateMachine-based fixture, so it is deleted along with the now-unused `ic-state-machine-tests`, `ic-types`, `ic-management-canister-types-private` and `assert_matches` dependencies. The environment-agnostic helpers the `pocket_ic` module and cketh still import (wasm loaders, init-arg builders, `usdc()`/`usdt()`, constants) stay at their existing paths. After this PR, no Rust target under `rs/ethereum` compiles against `ic-state-machine-tests`.

## Behavioural differences PocketIC forces, and how they are handled

PocketIC differs from StateMachine in two ways the tests are sensitive to. Both are handled in the fixture, in one place each, rather than worked around per test.

**Time advances kill the outcalls in flight.** PocketIC fails canister-http requests once they cross `CANISTER_HTTP_TIMEOUT_INTERVAL`; StateMachine left them pending indefinitely. The requests also stay listed until a round processes the failures, so a stub could bind to a request that was already dead and see its response discarded. Worse, a scraping cycle awaiting such a request keeps holding its `TimerGuard`, so the firing due at the new time is dropped as `AlreadyProcessing` and the query the test expects is never issued. `CkEthSetup::advance_time` therefore fails those outcalls up front and deterministically, doing what the advance would have done anyway. It is a no-op when nothing is in flight, because a fixture that has executed no round is expected to trigger the first scraping cycle itself.

**A stop request takes effect only in the following round.** The outcalls the minter's timers issue in that last running round become visible after it, so `stop_minter` ticks once before draining them.

The block number stubs also now match on `["finalized", false]`, which is what the scraping cycle actually queries — unconstrained, they could be consumed by the block height refresh timer's query for the latest block.

## Runtime

Measured against master with identical in-test instrumentation on both sides: **+5% (ckETH) and +11% (ckERC20)** of wall-clock at default parallelism (+15%/+20% single-threaded). Fixture-light tests are indistinguishable from master; the difference tracks the number of IC rounds a test drives, since PocketIC runs out of process. The dominant cost in both harnesses is tests that rebuild the whole fixture several times within one `#[test]`.


rs/ethereum/cketh/test_utils/src/mock.rs
```diff
@@ -110,60 +108,100 @@ impl JsonRpcRequestMatcher {
...
+            && self.provider.url() == request.url
             && has_expected_max_response_bytes
             && has_json_content_type_header
             && self.json_rpc_method == json_rpc_request.method
             && self
                 .match_request_params
                 .as_ref()
                 .map(|expected_params| expected_params == &json_rpc_request.params)
-                .unwrap_or(true)
+                .unwrap_or_else(|| {
```
gregorydemay: 🤖🧐 🟠 **Medium** — G5: this fallback re-implements `is_latest_block_refresh` (defined 10 lines below, and a third time in `lib.rs::drain_stray_latest_block_refresh_calls`). Three copies of one predicate, and they disagree on a malformed body: here `.expect("BUG: invalid JSON RPC request")`, in `is_latest_block_refresh` `Err(_) => false`, in `lib.rs` a bare `serde_json::from_slice(..).unwrap()`. Collapse to `.unwrap_or_else(|| !is_latest_block_refresh(request))` and make `is_latest_block_refresh` `pub(crate)` so `lib.rs` can reuse it.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +803,109 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+        }
+    }
 }
 
-fn assert_reply(result: WasmResult) -> Vec<u8> {
-    match result {
-        WasmResult::Reply(bytes) => bytes,
-        WasmResult::Reject(reject) => {
-            panic!("Expected a successful reply, got a reject: {reject}")
+fn drain_stray_latest_block_refresh_calls(env: &PocketIc) {
```
gregorydemay: 🤖🧐 🟠 **Medium** — G5 / divergent invariant handling: this is the third copy of the "is it the latest-block refresh poll" predicate (see `mock.rs` `matches` and `is_latest_block_refresh`), and the only one that parses the body as a raw `serde_json::Value` and `.unwrap()`s. Reuse `mock::is_latest_block_refresh` instead.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -549,41 +562,59 @@ impl CkEthSetup {
...
-        let stop_msg_id = self.env.stop_canister_non_blocking(self.minter_id);
-        self.stop_ongoing_https_outcalls();
-        let stop_res = self.env.await_ingress(stop_msg_id, 100);
-        assert_matches!(stop_res, Ok(WasmResult::Reply(_)));
+        let stop_msg_id = self.submit_stop_minter();
+        let stop_res = await_call(&self.env, stop_msg_id);
+        assert_matches!(stop_res, Ok(_));
     }
 
     pub fn stop_ongoing_https_outcalls(&self) {
```
gregorydemay, line 564: 🤖🧐 🟠 **Medium** — silent fallback: if `drain_pending_https_outcalls` never returns `true`, the loop simply ends and the function returns as if it had succeeded, despite its name promising the outcalls are stopped. The one caller (`tests/ckerc20.rs:113`) then goes straight into `await_call`, which will fail 1000 ticks later with an unrelated message. Panic here with the still-pending requests (`debug_http_outcalls`) so the real cause is visible.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +803,109 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+
+fn assert_reply(result: Result<Vec<u8>, RejectResponse>) -> Vec<u8> {
+    result.unwrap_or_else(|reject| panic!("Expected a successful reply, got a reject: {reject}"))
+}
+
+/// Like [`PocketIc::await_call`], but keeps rejecting any pending HTTP outcalls (typically the
+/// block height refresh timer's) between rounds, since `await_call`'s own internal ticking gives
+/// callers no chance to do so and can give up before a slow background timer or a large in-flight
+/// scrape (deterministic time slicing can spread these over many rounds) settles.
+pub(crate) fn await_call(
```
gregorydemay: 🤖🧐 🟠 **Medium** — this helper bundles two policies: "await the message" and "answer every pending canister-http request with HTTP 500". `stop_minter` needs the draining; `WithdrawalFlow::minter_response`, `Erc20WithdrawalFlow::expect_trap` and `DepositErc20Flow::expect_trap` do not — and for them the drain can silently swallow an outcall the test intended to stub, something the `await_ingress(id, MAX_TICKS)` it replaces never did. Splitting it (plain await vs. drain-then-await), or making the drain an explicit argument, keeps the destructive behaviour at the one call site that wants it.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -549,41 +562,59 @@ impl CkEthSetup {
...
-            .collect();
-        let mut payload = PayloadBuilder::new();
-        for callback_id in ongoing_https_outcalls {
-            payload = payload.http_response(callback_id, &server_error_response);
+        for _ in 0..MAX_TICKS {
+            if self.drain_pending_https_outcalls() {
+                return;
+            }
+            // The block height refresh timer can be mid-backoff after hitting ic-cdk-timers'
+            // per-timer concurrent-call cap (see `mock::tick_until_next_http_request`), with
```
gregorydemay: 🤖🧐 🟠 **Medium** — C2: `mock::tick_until_next_http_request` no longer exists (it was renamed to `JsonRpcRequestMatcher::poll_for_request`, and the jump-forward logic lives in `find_rpc_call_retrying`). This is the pointer to the PR's subtlest mechanism, so a dangling reference is costly. Retarget it.

rs/ethereum/cketh/test_utils/src/mock.rs
```diff
@@ -110,60 +108,100 @@ impl JsonRpcRequestMatcher {
...
+
+    // Like `find_rpc_call`, but for callers that know a matching request should eventually
+    // appear. The block height refresh timer's own outcall competes with the scrape's for
+    // ic-cdk-timers' per-timer concurrent-call cap (5 overlapping invocations): a big enough
+    // `advance_time` jump can make several of its intervals due at once, hit that cap, and make
+    // it defer itself by a full `REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL`. Nanosecond ticking can
+    // never cross that gap on its own, so jump forward to unstick it, then retry. This must not
+    // be used for negative expectations (no call should appear): jumping forward burns simulated
+    // time regardless of outcome, which can itself trigger unrelated periodic minter work.
+    fn find_rpc_call_retrying(&self, env: &PocketIc) -> Option<CanisterHttpRequest> {
```
gregorydemay: 🤖🧐 🟠 **Medium** — retry budget: `MAX_TICKS` (10) × `REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL` (30s) = 300s of simulated time, which exceeds `SCRAPING_ETH_LOGS_INTERVAL` (180s). So an expectation that is slow to appear can itself trigger an extra, unrelated log scrape, whose outcall the very next stub may then answer instead of the intended one — i.e. this can convert an ordering regression into a green run. It is deterministic (not a flakiness source), but please cap the total jump below `SCRAPING_ETH_LOGS_INTERVAL` (e.g. 5 iterations) so it cannot cross a scrape boundary.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +803,109 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+        }),
+        additional_responses: vec![],
+    });
+}
+
+fn assert_reply(result: Result<Vec<u8>, RejectResponse>) -> Vec<u8> {
+    result.unwrap_or_else(|reject| panic!("Expected a successful reply, got a reject: {reject}"))
+}
+
+/// Like [`PocketIc::await_call`], but keeps rejecting any pending HTTP outcalls (typically the
```
gregorydemay: 🤖🧐 🔵 **Nit** — C4 doc/code drift: the doc says "keeps rejecting any pending HTTP outcalls", but the body *replies* with a `CanisterHttpReply { status: 500 }`. A 500 reply and a canister-http reject take different paths through the minter (`is_response_too_large`, error classification), so the wording matters.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +803,109 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+/// block height refresh timer's) between rounds, since `await_call`'s own internal ticking gives
+/// callers no chance to do so and can give up before a slow background timer or a large in-flight
+/// scrape (deterministic time slicing can spread these over many rounds) settles.
+pub(crate) fn await_call(
+    env: &PocketIc,
+    message_id: RawMessageId,
+) -> Result<Vec<u8>, RejectResponse> {
+    const MAX_AWAIT_TICKS: usize = 100 * MAX_TICKS;
```
gregorydemay: 🤖🧐 🔵 **Nit** — G25/G16: `100 * MAX_TICKS` expresses 1000 as a multiple of an unrelated constant. Also worth calling out in the PR description that this is a 100× relaxation of the `await_ingress(id, MAX_TICKS = 10)` bound it replaces — a genuinely hung flow now burns 1000 ticks before failing. Just write the literal with a one-line rationale.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -510,29 +507,45 @@ impl CkEthSetup {
...
+                })
+                .unwrap(),
+            )
+            .expect("failed to submit stop_canister call")
     }
 
     /// Try to stop the minter without first stopping the ongoing HTTPS outcalls. Assert that the
-    /// `IngressStatus` is `Processing`.
+    /// call is still processing (i.e. blocked on the open call contexts for those outcalls).
     pub fn try_stop_minter_without_stopping_ongoing_https_outcalls(&self) {
```
gregorydemay, line 530: 🤖🧐 🔵 **Nit** — G12: the 10-iteration loop is dead on the passing path — the first `is_none()` returns immediately, so this only ever asserts "not completed after one tick"; and on the failing path it re-reads a terminal status 9 more times, which cannot change. Collapse to a single `tick()` + `assert!(self.env.ingress_status(..).is_none(), ..)`. (The "stop was actually picked up" signal that `IngressState::Processing` used to carry is preserved by the `tick_until_minter_canister_status(Stopping)` on the next line of `tests/cketh.rs:886`.)

rs/ethereum/cketh/test_utils/src/mock.rs
```diff
@@ -110,60 +108,100 @@ impl JsonRpcRequestMatcher {
...
+            }
+            for request in pending
+                .iter()
+                .filter(|request| is_latest_block_refresh(request))
+            {
+                env.mock_canister_http_response(MockCanisterHttpResponse {
+                    subnet_id: request.subnet_id,
+                    request_id: request.request_id,
+                    response: CanisterHttpResponse::CanisterHttpReject(CanisterHttpReject {
+                        reject_code: 2, // SysTransient
```
gregorydemay: 🤖🧐 🔵 **Nit** — G25: `reject_code: 2, // SysTransient` (and `1, // SysFatal` at :249, plus `2` again at `lib.rs:870`) are magic numbers documented by a comment. Name them once (`const SYS_FATAL: u64 = 1; const SYS_TRANSIENT: u64 = 2;`) rather than repeating the comment at each site.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -549,41 +562,59 @@ impl CkEthSetup {
...
+            // per-timer concurrent-call cap (see `mock::tick_until_next_http_request`), with
+            // nothing currently pending for us to drain; nanosecond ticking can't cross that gap,
+            // so jump forward to give it a chance to resurface before `await_call` gives up.
+            self.env
+                .advance_time(ic_cketh_minter::REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL);
         }
-        self.env.execute_payload(payload);
+    }
+
+    fn drain_pending_https_outcalls(&self) -> bool {
```
gregorydemay: 🤖🧐 🔵 **Nit** — G5: this `MockCanisterHttpResponse { .. CanisterHttpReply { status: 500, .. } }` block is byte-for-byte the one inside `await_call` (:891-901), and `reject_stray_http_outcall` is duplicated inline in `mock.rs::poll_for_request` (:120-131). Two small helpers (`reply_500`, `reject_transient`) remove all four copies.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +803,109 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
 
-fn install_evm_rpc(env: &StateMachine, evm_rpc_id: CanisterId) {
+fn install_evm_rpc(env: &PocketIc, evm_rpc_id: Principal) {
     let args = evm_rpc_types::InstallArgs::default();
-    env.install_existing_canister(evm_rpc_id, evm_rpc_wasm(), Encode!(&args).unwrap())
-        .unwrap();
+    env.install_canister(evm_rpc_id, evm_rpc_wasm(), Encode!(&args).unwrap(), None);
+}
+
+fn drain_startup_http_outcalls(env: &PocketIc) {
```
gregorydemay: 🤖🧐 🔵 **Nit** — this always burns the full 10 ticks even when nothing is pending, and never asserts it actually drained anything (same for `drain_stray_latest_block_refresh_calls`). An early exit once `get_canister_http()` is empty would keep the tick phase closer to the StateMachine baseline and make the intent explicit.

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -115,8 +111,8 @@ fn should_retry_to_add_usdc_when_minter_stopped() {
         .unwrap();
 
     ckerc20.cketh.stop_ongoing_https_outcalls();
-    let stop_res = ckerc20.env.await_ingress(stop_msg_id, 100);
-    assert_matches!(stop_res, Ok(WasmResult::Reply(_)));
+    let stop_res = ckerc20.env.await_call(stop_msg_id);
```
gregorydemay, line 117: 🤖🧐 🔵 **Nit** — G11: this is the one await that uses pocket-ic's built-in `await_call` while every other await in the suite goes through `ic_cketh_test_utils::await_call`. It is correct here (the preceding `stop_ongoing_https_outcalls()` clears the outcalls the built-in cannot answer), but the asymmetry is worth a word — otherwise the next reader copies the wrong one.

gregorydemay (commented):
🤖🧐 **VERDICT: CHANGES_REQUESTED** — 0 blockers, 6 mediums, 7 nits; CI pending (24 pass / 4 pending / 0 fail). Local: 2 consecutive uncached green runs of `//rs/ethereum/cketh/minter:integration_tests`.

> ⚠️ Posted as a comment rather than a formal “Request changes” review: GitHub refuses `--request-changes` on a PR owned by the same account. The verdict below is still **CHANGES_REQUESTED**.

<details>

### Bottom line
The port itself is excellent. The `#[test]` sets in `tests/cketh.rs` (26) and `tests/ckerc20.rs` (39) are **identical** to the PR-2 base — no test dropped, renamed, `#[ignore]`d or `#[should_panic]`ed — and every assertion change is accounted for by the documented type/constant mapping. No loosened asserts, no `println!`/`dbg!`/commented-out asserts survived the constant-regeneration pass. Senders are preserved exactly (`execute_ingress`/`query` → `Principal::anonymous()`, `execute_ingress_as`/`query_as` → `pid.0`). `mock.rs` keeps the raw-body `max_response_bytes` check with the verbatim legacy message and never invokes `cleanup_response` manually. Component boundary is clean: the delta touches only `rs/ethereum/cketh/**` + `Cargo.lock`; `dump_stable_memory.rs` and `deposit_from_cex_demo.rs` are untouched. Everything I flag is in the new fixture *plumbing*, not in the tests.

### The three declared deviations
1. **`find_rpc_call_retrying` (positive-only)** — the asymmetry is **sound**. `expect_no_matching_rpc_call` / `expect_no_refresh_gas_fee_estimate` still go through `find_rpc_call` → `poll_for_request`, whose detection window (10 × `tick()` + 1ns) is byte-for-byte the StateMachine `tick_until_next_http_request` it replaces — it has *not* silently shrunk. The mechanism is **deterministic**, not a flakiness source. I found no simpler in-repo precedent: `rs/bitcoin/checker`'s `tick_until_next_request` is the same plain 10-tick loop, but the checker has no competing periodic timer, so it doesn't face this. One concrete concern remains (see the inline note on `mock.rs:152`): the retry budget can burn 300s of simulated time, more than `SCRAPING_ETH_LOGS_INTERVAL` (180s), so it can cross a scrape boundary.
2. **Custom `await_call`** — justified: pocket-ic's own `await_call` ticks internally with no injection point, and `await_call_no_ticks` would deadlock in a sync client that must drive the rounds itself. It is bounded (1000 ticks) and panics with the message id. Two issues inline: the 100× bound relaxation is written as `100 * MAX_TICKS`, and the helper's implicit "500 every pending outcall" policy is only wanted by `stop_minter`, not by `minter_response`/`expect_trap`.
3. **`new_pocket_ic()` made `pub`** — **sound and effectively forced**. `CkErc20Setup::new_without_ckerc20_active` installs the minter, and `CkEthSetup::new` calls `minter_address()`, which needs the fiduciary `key_1`; LSO's system-subnet builder cannot serve it. The only semantic delta is that cycles are now charged, and neither `should_trap_when_ckerc20_feature_not_active` asserts a balance — they assert a trap. `pub` is the minimum surface needed by the two call sites.

- **Duplication** — **found**. The latest-block-refresh predicate exists in three places, two of them different implementations (`mock.rs::matches`, `mock.rs::is_latest_block_refresh`, `lib.rs::drain_stray_latest_block_refresh_calls`). The 500-reply block and the transient-reject block are each duplicated across two sites. Inline on `mock.rs:187`, `lib.rs:851`, `lib.rs:584`.
- **Divergent invariant handling** — **found**, same predicate: a malformed request body is `.expect(..)` in one copy, `Err(_) => false` in the second, bare `.unwrap()` in the third.
- **Silent fallbacks** — **found**: `stop_ongoing_https_outcalls` gives up after 10×10 attempts and returns as though it succeeded (`lib.rs:570`). The two `drain_*` helpers likewise never assert they drained anything.
- **Unused derives** — none; the PR introduces no new types.
- **Primitive-obsession parameters** — none of substance; the only primitives are the raw `reject_code` literals (nit, `mock.rs:125`).
- **Test-only code in production modules** — none.
- **Redundant / derivable parameters, caller-owned decisions** — one instance: `await_call`'s drain policy is decided implicitly for all three call sites (`lib.rs:885`).

### Coverage / test-pyramid
No new tests are warranted — this is a harness swap, and the existing suite is the regression oracle. The regenerated constants (`MINTER_ADDRESS`, both `DEFAULT_*_WITHDRAWAL_TRANSACTION(_HASH)`, both `response.rs` r/s pairs) are cross-checked by `test_utils:lib_tests::should_use_meaningful_constants` (encoding + keccak hash) and, for `MINTER_ADDRESS`, by the `assert_eq!` in `CkEthSetup::new`. Note that the per-account deposit address at `tests/ckerc20.rs:182` is a pure golden value with no independent oracle — unavoidable, just worth knowing.

### Not blocking, FYI
`drain_startup_http_outcalls` in `CkEthSetup::new` has no StateMachine counterpart: it *rejects* the genesis scrape and block-refresh outcalls rather than leaving them pending. I traced this and it does not weaken anything (a rejection leaves `last_observed_block_number` `None` exactly as a pending call did, and the scrape timer retries on the next interval), but it is a behavioural delta the PR description does not mention — worth a sentence there. Also: the description says 40 ckerc20 tests; there are 39.

### Verification run
- `bazel test //rs/ethereum/cketh/test_utils:lib_tests //rs/ethereum/cketh/minter:lib_tests //rs/ethereum/cketh/minter:integration_tests --nocache_test_results` → **4/4 PASSED** (ckerc20 143.9s, cketh 71.3s)
- second uncached run of `//rs/ethereum/cketh/minter:integration_tests` → **2/2 PASSED** (141.3s / 71.0s) — no timing drift between runs, consistent with the retry mechanism being deterministic
- `cargo check --all-targets --all-features -p ic-cketh-test-utils -p ic-cketh-minter` → clean
- `timeout = "long"` on the suite is still appropriate (bazel reports the declared size as *too big*, not too small).

CI is not yet green (4 jobs still pending, none failing), which is an independent gate on a READY verdict.

</details>

rs/ethereum/cketh/test_utils/src/mock.rs
gregorydemay: 🤖🧐 Fixed in c93b423a7e — collapsed the fallback to `!is_latest_block_refresh(request)` and made the predicate `pub(crate)`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — now reuses `mock::is_latest_block_refresh` instead of its own raw `serde_json` parse.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — panics with the still-pending requests (`debug_http_outcalls`) instead of returning silently.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — split into `await_call` (clears only stray latest-block-refresh outcalls, never a test-owned stub) and `await_call_draining_outcalls` (the old blanket-500 behaviour), the latter used only by `stop_minter`. Note: a fully drain-less `await_call` regressed ~24 unrelated tests in `cketh_test`/`ckerc20_test` (confirmed via a live run) because nothing was left to clear the stray latest-refresh backlog that `find_rpc_call_retrying` then has to work through — hence keeping that one categorically-safe rejection in the plain path.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — retargeted to `JsonRpcRequestMatcher::poll_for_request`/`find_rpc_call_retrying`.

rs/ethereum/cketh/test_utils/src/mock.rs
gregorydemay: 🤖🧐 Fixed in c93b423a7e — named `REJECT_CODE_SYS_FATAL`/`REJECT_CODE_SYS_TRANSIENT`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — reworded to describe the actual HTTP 500 reply.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — wrote the literal (1000) with a one-line rationale instead of `100 * MAX_TICKS`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — collapsed to a single tick + assert.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — extracted a shared `reply_500` helper (and `c93b423a7e` makes `poll_for_request` call the existing `reject_stray_http_outcall` instead of duplicating it).

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in b14c90c231 — both `drain_startup_http_outcalls` and `drain_stray_latest_block_refresh_calls` now exit early once `get_canister_http()` is empty.

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖🧐 Fixed in 3d407fdb3e — added a comment explaining why the built-in `await_call` is safe here.

rs/ethereum/cketh/test_utils/src/mock.rs
gregorydemay: 🤖🧐 Fixed in c93b423a7e, with a caveat worth recording. Capping strictly below 180s (5 attempts, 150s) regressed ~24 tests across both suites (verified via a live uncached run: they failed to find their target call within budget). The cap is at 6 attempts (180s) — exactly `SCRAPING_ETH_LOGS_INTERVAL`, not strictly under it. I'd rather flag this honestly than claim a stricter bound that isn't actually met: happy to dig further into why 5 wasn't enough if you'd like a tighter number, but 6 is the smallest value I could verify green twice.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -549,41 +562,59 @@ impl CkEthSetup {
     }
```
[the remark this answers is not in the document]
gregorydemay: 🤖🧐 Fixed in b14c90c231 — added an explicit `self.env.tick()` so the loop doesn't rely on `canister_status`'s own update-call processing to make progress. The other three spots you flagged as the same pattern (570, 837, 853) are covered separately: 570 (`stop_ongoing_https_outcalls`) now panics loudly instead of relying on incidental progress; 837/853 (`drain_startup_http_outcalls`/`drain_stray_latest_block_refresh_calls`) now exit early once the outcall queue is empty, both in b14c90c231.

rs/ethereum/cketh/test_utils/src/mock.rs
```diff
@@ -174,26 +212,25 @@ struct StubOnce {
...
-        if let Some((id, _)) = self.matcher.find_rpc_call(env) {
-            panic!("expect no request matching the stub {self:?} but found one {id}");
+    fn expect_no_matching_rpc_call(self, env: &PocketIc) {
+        if let Some(request) = self.matcher.find_rpc_call(env) {
+            panic!(
+                "expect no request matching the stub {self:?} but found one {}",
+                request.request_id
+            );
         }
     }
```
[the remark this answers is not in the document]
gregorydemay: 🤖🧐 Deliberately left as-is. This is the same design point gregorydemay's review validated in 'declared deviation #1': `poll_for_request`'s bounded ticking (10 x tick() + 1ns, no jump-forward) is byte-for-byte the StateMachine `tick_until_next_http_request` it replaces, for both positive and negative expectations. It hasn't grown any new side effect relative to the code it replaces, and 2 consecutive uncached full-suite runs show no flakiness from it. Happy to revisit if you can point to a concrete failure mode it introduces beyond what StateMachine already had.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +798,156 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+fn assert_reply(result: Result<Vec<u8>, RejectResponse>) -> Vec<u8> {
+    result.unwrap_or_else(|reject| panic!("Expected a successful reply, got a reject: {reject}"))
+}
+
+/// Polls `ingress_status` until `message_id` completes. Unlike [`PocketIc::await_call`], which
+/// ticks internally with no injection point, this gives the caller a chance to break a deadlock
+/// where the call is itself awaiting a pending HTTP outcall that only a test can answer.
+///
+/// Still clears stray latest-block-refresh outcalls between rounds (see
+/// `mock::is_latest_block_refresh`): no test-owned stub ever targets those (mock.rs's own
```
gregorydemay: 🤖🧐 🔵 **Nit** (new, from re-review) — the split is the right fix, but this sentence overstates the guarantee. I checked `matches`: the exclusion is categorical only for stubs that leave `match_request_params` unset (they hit the `!is_latest_block_refresh(request)` fallback). A stub built with an explicit `.with_request_params(json!(["latest", false]))` on `EthGetBlockByNumber` **does** match a refresh outcall — and two exist (`tests/cketh.rs:842`, `tests/ckerc20.rs:258`). They are safe here because of *ordering*, not exclusion: both are built and consumed synchronously by `expect_rpc_calls` before any await runs. Please state that, so a future test that interleaves such a stub with an await does not silently get its request rejected out from under it.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -768,41 +798,156 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+    env: &PocketIc,
+    message_id: RawMessageId,
+) -> Result<Vec<u8>, RejectResponse> {
+    poll_for_call(env, message_id, true)
+}
+
+fn poll_for_call(
+    env: &PocketIc,
+    message_id: RawMessageId,
+    drain_outcalls: bool,
```
gregorydemay: 🤖🧐 🔵 **Nit** (new, from re-review) — F3/G15: `drain_outcalls: bool` is a flag argument that makes the callee do two things. The two named public wrappers mitigate this well and the public API is now exactly right, so this is cosmetic — but a two-variant enum (`OutcallPolicy::{RejectRefreshOnly, DrainAll}`) would read better at the `match` site than a bare bool.

rs/ethereum/cketh/test_utils/src/mock.rs
```diff
@@ -110,60 +108,96 @@ impl JsonRpcRequestMatcher {
...
+    // ic-cdk-timers' per-timer concurrent-call cap (5 overlapping invocations): a big enough
+    // `advance_time` jump can make several of its intervals due at once, hit that cap, and make
+    // it defer itself by a full `REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL`. Nanosecond ticking can
+    // never cross that gap on its own, so jump forward to unstick it, then retry. This must not
+    // be used for negative expectations (no call should appear): jumping forward burns simulated
+    // time regardless of outcome, which can itself trigger unrelated periodic minter work.
+    //
+    // Capped at `SCRAPING_ETH_LOGS_INTERVAL` (3 minutes): retrying for longer than that could
+    // itself trigger an extra, unrelated log scrape, whose outcall the very next stub might then
+    // answer instead of the one this call is actually waiting for. A smaller cap (5 attempts,
```
gregorydemay: 🤖🧐 🔵 **Nit** (new, from re-review) — accepting the boundary-exact cap (see my summary for the reasoning), but one observation worth recording here for whoever touches this next: 6 × 30s lands on *exactly* `SCRAPING_ETH_LOGS_INTERVAL`, and the fact that ~24 tests fail at 150s but pass at 180s is strong evidence that what those tests are actually waiting for **is the next log scrape**, not merely an unstuck refresh timer. If so, the honest mechanism is "advance to the next scrape" and a single `advance_time(SCRAPING_ETH_LOGS_INTERVAL)` would express it directly. Worth a follow-up investigation rather than a change in this PR.

gregorydemay (commented):
🤖🧐 **VERDICT: CHANGES_REQUESTED** (CI-gated only) — 0 blockers, 0 mediums, 3 new nits; all 6 mediums + 7 nits from the first pass **resolved**. CI: 22 pass / 4 pending / **0 fail**. Local: 3rd consecutive uncached green run of the full triad.

> Substantively this is READY. The only thing standing between it and a READY verdict is that 4 CI jobs have not finished; `reviewer.md` makes non-green CI an independent hard gate. Re-run `gh pr checks` once they land — if they are green and no new commit has been pushed, treat this as READY for human approval.

<details>

### Correction to my own first review
I reported 39 ckerc20 tests against the PR's 40. **The implementer is right and I was wrong.** My extraction used `grep -A1 '#[test]'`, which misses the one test carrying an intervening `#[allow(deprecated)]` attribute (`tests/ckerc20.rs:1909`). Re-counted with an attribute-tolerant scan: **26 cketh + 40 ckerc20, in both the base and this PR, with identical function-name sets**. The "no test dropped" conclusion was never affected — the same flawed extraction was applied to both sides — but the number in my summary was wrong. Apologies for the noise.

### Per-finding resolution (13/13)
| # | Finding | Status |
|---|---|---|
| 🟠 `mock.rs:187` | latest-block-refresh predicate written 3× | **Resolved** — single `pub(crate) is_latest_block_refresh`, used by `matches`, `poll_for_request` and `lib.rs`; divergent malformed-body handling gone |
| 🟠 `lib.rs:851` | third impl on raw `serde_json::Value` | **Resolved** — delegates to `mock::is_latest_block_refresh` |
| 🟠 `lib.rs:570` | silent give-up | **Resolved** — now panics with `debug_http_outcalls`. This fix also *retroactively earns* the `ckerc20.rs:114` comment: the built-in `await_call` is safe there precisely because this now panics rather than lying |
| 🟠 `lib.rs:885` | `await_call` conflated await + drain | **Resolved** — split into `await_call` / `await_call_draining_outcalls`; blanket-500 now reaches only `stop_minter` |
| 🟠 `lib.rs:576` | stale `mock::tick_until_next_http_request` ref | **Resolved** — retargeted to `find_rpc_call_retrying`; symbol verified to exist |
| 🟠 `mock.rs:152` | retry budget 300s > scrape interval | **Addressed w/ pushback** — capped at 6 (=180s). Ruling below |
| 🔵 ×7 | doc drift, `100 * MAX_TICKS`, dead loop, magic reject codes, duplicated mock blocks, no early exit, built-in `await_call` | **All resolved** |

Also accepted: the Copilot-prompted `self.env.tick()` in `tick_until_minter_canister_status` (`lib.rs:558`). That is a genuine pre-existing bug — the loop never ticked and only made progress as a side effect of `canister_status` being an update call. Slight scope expansion, but the function's name promised it and it is strictly more robust.

### Ruling on pushback 1 — boundary-exact retry cap (6 × 30s = 180s = `SCRAPING_ETH_LOGS_INTERVAL`)
**Accepted.** I traced the equality case rather than taking it on faith. The cumulative 180s is only ever reached on a path where all six in-loop attempts have already failed; the loop exits early at k × 30s the moment a match appears. So the single path that both burns the full budget *and* succeeds is the trailing `find_rpc_call` after the sixth advance — every other outcome is a test failure with a clear panic. The stub-absorption race I worried about therefore survives only in that one narrow window, and only if the scrape timer is `>=`-triggered *and* its epoch aligns exactly with the loop start.

Against that: 5 attempts is *demonstrably* broken (~24 failures). A mechanism that fails two dozen tests is strictly worse than a theoretical boundary case, and "strictly under the interval" was my heuristic, not a correctness requirement. The code comment and the thread reply both state the tradeoff honestly rather than papering over it, which is the right call. No different mechanism is needed for this PR — but see my inline note on `mock.rs:147`: the fact that 150s fails and 180s passes is good evidence the tests are really waiting for *the next scrape*, which would make a single `advance_time(SCRAPING_ETH_LOGS_INTERVAL)` the honest expression of this. Follow-up, not a blocker.

### Ruling on pushback 2 — reject-only-refresh plain `await_call`
**Accepted; the policy separation I asked for is preserved.** The thing that mattered was that the blanket 500 stop reaching `minter_response`/`expect_trap`, and it no longer does — the residual action is confined to one request class.

But I verified the "categorically never test-owned" claim against the matchers and it is **not** universally true, so the doc needs one sentence softened (inline on `lib.rs:900`). The exclusion is categorical only for stubs with `match_request_params == None`, which hit the `!is_latest_block_refresh(request)` fallback. A stub with an explicit `.with_request_params(json!(["latest", false]))` on `EthGetBlockByNumber` **does** match a refresh outcall, and two such stubs exist (`tests/cketh.rs:842`, `tests/ckerc20.rs:258`). They are safe by *ordering* — both are built and consumed synchronously inside `expect_rpc_calls` before any await — not by exclusion. I also checked the adjacent hazard in `poll_for_request` and it is sound: the `any(matches)` early-return runs before the rejection sweep, so a per-provider stub is never starved by its own siblings' refresh requests.

### Nothing new broken
Re-diffed the three commits in full. No test touched beyond the one added explanatory comment; the 26+40 test sets are byte-identical to the base; no assertion, tick count, or constant changed in this round. Three new 🔵 nits only (doc precision on `lib.rs:900`, `bool` flag param on `lib.rs:925`, the follow-up observation on `mock.rs:147`) — none gate the verdict.

### Verification run
- `bazel test //rs/ethereum/cketh/test_utils:lib_tests //rs/ethereum/cketh/minter:lib_tests //rs/ethereum/cketh/minter:integration_tests --nocache_test_results` → **4/4 PASSED** (ckerc20 141.3s, cketh 69.1s)
- Timings across my three uncached runs: ckerc20 143.9 / 141.3 / 141.3s, cketh 71.3 / 71.0 / 69.1s — tight spread, consistent with the retry mechanism being deterministic rather than racing.
- `gh pr checks 10955` → 22 pass, 4 pending (`Build IC`, `Bazel Test All`, `Cargo Build Linux`, `Bazel Test All on RBE`), 0 failing.

</details>

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in 9ceecbf3d4 — reworded to state the real safety property (built and consumed synchronously before any await, not a matching exclusion), and named both exceptions (tests/cketh.rs:842, tests/ckerc20.rs:258) so a future interleaving bug is easy to spot.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖🧐 Fixed in 9ceecbf3d4 — replaced the bool with an `OutcallPolicy` enum (`RejectRefreshOnly` / `DrainAll`).

rs/ethereum/cketh/test_utils/src/mock.rs
gregorydemay: 🤖🧐 Agreed this is worth digging into, but leaving it as a follow-up per your own note — no change in this PR. If it does turn out these tests are specifically waiting on the next scrape, `advance_time(SCRAPING_ETH_LOGS_INTERVAL)` would indeed be the more honest mechanism than piggy-backing on the refresh-timer's cap-driven backoff.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -208,6 +208,7 @@ rust_ic_test_suite(
     ),
     data = [
         ":cketh_minter_debug.wasm.gz",
+        "//:pocket-ic-server",
```
mbjorkqvist, line 212: This test suite sets `timeout = "long"` a few lines above. However, [the default TTL](https://github.com/dfinity/ic/blob/777dc537a673a47169f34fae9ee0aa94c7ee3f2f/packages/pocket-ic/src/nonblocking.rs#L159) for a PocketIC server is 600s, so if a single test in this suite runs longer than 10min, the PocketIC server will be killed (before the `timeout = "long"` of 15min kills the test). [This PR](https://github.com/dfinity/ic/pull/11086/commits/a073ba80da42a2e04c8ff07286fd79c0ab39c3c2#diff-7df964536198fa7b37625c64c78ada2e17881020c3176f13d577453d2b297b10R358-R387) manually starts the server in a way that allows you to pass another value for the TTL (since `PocketICBuilder` doesn't expose it). There's [a related discussion](https://dfinity.slack.com/archives/C0527QACNJC/p1786116654910259) on our internal Slack also regarding the value.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: 🤖 Good catch — fixed in `191da42785`, and the risk was a bit different (and closer) than it looked.

The 600s you linked is passed as `hard_ttl`, not `ttl`. Per `StartServerParams`, `ttl` is an idle timeout ("stops gracefully if no request has been received for the duration of its TTL"), which activity keeps resetting, whereas `hard_ttl` "stops with a hard exit after the duration of its hard TTL **since its launch**" — traffic is irrelevant. And since `nonblocking.rs` starts the server with `reuse: true`, every fixture in a test binary shares one server, so that 10-minute clock runs against the **whole suite**, not a single test.

That makes the margin smaller than the `timeout = "long"` comparison suggests. I measured this suite recently: `ckerc20_test` is ~130s at default parallelism but **~487s single-threaded** — about 8 of the 10 minutes. Any run that ends up effectively serialized would lose the server mid-suite and fail every test after it, rather than failing the slow one.

Followed the approach from #11086: start the server explicitly with `hard_ttl` of 1h and pass its URL via `with_server_url`. That PR adds a `long_lived_server_url()` helper to this same crate (`sweeper_funding.rs`), so once it lands the two should be deduplicated into one helper — I kept the shape and name identical to make that trivial. Happy to converge on whatever value the Slack thread settles on.

Verified: `cketh_test` 27/27, `ckerc20_test` 40/40, `test_utils:lib_tests` 2/2, clippy `--all-targets` and buildifier clean.

gregorydemay: 🤖 Moved out of this PR into #11123, so it can be reviewed on its own.

To correct my earlier reply: the fix is unchanged, but it does not belong in the migration PR. #11123 starts the server explicitly with a 1h `hard_ttl` and passes its URL via `with_server_url`, and adds the `tokio` dependency that needs.

gregorydemay:
🤖 Applied your review from #10956 here, since that PR's content ended up in this one — `2615f601d4` drops `ic-http-types`, `proptest` (both from `Cargo.toml` and `BUILD.bazel`) and `ic-test-utilities-load-wasm` (`Cargo.toml` only, it was never in `BUILD.bazel`). Confirmed none of the three is referenced anywhere in the crate.

Also removed the note repeating what the stub parameters already say (`3b6dff3259`).

`ledger-suite-orchestrator:integration_tests` and `cketh_test` pass, buildifier clean.

PR #10957, opened 2026-07-29
fix(testnets): give the cloud_engine controller usable init args

pietrodimarco-dfinity:
The engine controller is installed alongside the NNS, and `cloud_engine` left its init args `None` — so both of the canister's own defaults applied, and both are mainnet-shaped:

| default | value | on a testnet |
|---|---|---|
| `DEFAULT_AUTHORIZED_CALLER` | `bct5z-vccu4-…-fqe` (mainnet backend principal) | no local canister may call the controller |
| `DEFAULT_INITIAL_DKG_SUBNET_ID` | `fuqsr-in2lc-…-nae` (a mainnet subnet) | subnet does not exist, so engine subnets cannot be bootstrapped |

Net effect: engine creation cannot work on any `cloud_engine` testnet. Confirmed against the live testnet, where `getEngineController()` resolves and the controller is installed, but its authorized caller is the mainnet principal rather than the consumer's engine canister.

## Changes

* `nns_dapp`: add `nns_dapp_customizations_with_engine_controller(authorized_caller, initial_dkg_subnet_id)`.
* `cloud_engine`: pass the engine canister (`5s2ji-faaaa-aaaaa-qaaaq-cai`, the `whale` reservation) as the authorized caller, and this testnet's own NNS subnet — read from the topology, not hardcoded — as the DKG source.

`5s2ji` is hardcoded because the controller is installed *before* the reservations are created, and the append-only reservation order fixes the id; the file already documents that contract.

The helper lives in `nns_dapp` rather than in the testnet because `nns_dapp` already depends on `ic-nns-test-utils`, where `EngineControllerInitArgs` is declared. Adding that dependency to `rs/tests/testnets` would have meant touching both `Cargo.toml` and `BUILD.bazel` — the mismatch that made #10928 red earlier today.

## Verification

`cargo check --locked -p testnets -p nns_dapp --all-targets` passes, and `rustfmt --edition 2024` is clean. Not yet exercised end to end: it takes effect on the next testnet build, and the proof will be an engine creation getting past `changeSubnetMembership`.

### DIFF
--- a/rs/tests/testnets/cloud_engine.rs
+++ b/rs/tests/testnets/cloud_engine.rs
@@ -110,6 +110,12 @@ const WHALE_CANISTER_CYCLES: u128 = 100_000_000_000_000_000;
 /// Cycles to fund the proxy application canister with. 300T cycles.
 const PROXY_CANISTER_CYCLES: u128 = 300_000_000_000_000;
 
+/// The reserved id the `whale` canister claims: control-panel's engine canister,
+/// and the only principal allowed to drive the engine controller here. Hardcoded
+/// because the reservation order below fixes it, and because the controller is
+/// installed with the NNS, before the reservations are created.
+const ENGINE_CANISTER_ID: &str = "5s2ji-faaaa-aaaaa-qaaaq-cai";
+
 /// Unclaimed ids reserved after the three above, so that a consumer can pin a new
 /// canister id without changing this testnet and redeploying it first. Raising this
 /// is safe — the new ids come after the existing ones; lowering it retires ids from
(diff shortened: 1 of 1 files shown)


rs/tests/testnets/cloud_engine.rs
```diff
@@ -110,6 +110,12 @@ const WHALE_CANISTER_CYCLES: u128 = 100_000_000_000_000_000;
 /// Cycles to fund the proxy application canister with. 300T cycles.
 const PROXY_CANISTER_CYCLES: u128 = 300_000_000_000_000;
 
+/// The reserved id the `whale` canister claims: control-panel's engine canister,
```
daniel-wong-dfinity-org-twin, line 113: Not sure what the "whale" canister is. Sounds like an informal name only known to a small clique.

rs/tests/testnets/cloud_engine.rs
```diff
@@ -456,10 +462,29 @@ pub fn setup(env: TestEnv) {
...
+    // mainnet's: only the mainnet backend principal may call it, and new engine
+    // subnets would be bootstrapped from a mainnet subnet that does not exist
+    // here. Both are set explicitly so engines can actually be created on this
+    // testnet — the caller is the engine canister that claims the `whale`
+    // reservation below, and the DKG source is this testnet's own NNS subnet.
+    //
+    // The init args are filled in via `Default::default()` and field assignment
+    // rather than by naming `EngineControllerInitArgs`, which would pull
+    // ic-nns-test-utils into this crate's Cargo.toml *and* BUILD.bazel for the
+    // sake of one struct literal.
```
daniel-wong-dfinity-org-twin, line 475: This sounds like the type of extra stuff that Claude would say.

The main thing that comments explain is what the code does. Comments can go into why of the how things are done (e.g. here), but I would only do that if someone would look at this, and be like, "wtf? I'm going to change this to do it the 'obvious' way". But in this case, I don't think there is necessary an "obvious" alternative that someone would just spontaneously "correct" this to, since, as this comment says, it would require editing the build files (i.e. Cargo.toml and BUILD.bazel). Even if someone does want to jump through those hoops, I do not see what the harm would be, why we should try to stop them with this comment. If they did that, then, rather than indirectly depending on other libraries, this would just depend on them DIRECTLY. That is arguably a good thing. E.g. at Google, we used to say "Depend on what you use" (and I don't see how you can really find fault with that!). In any case, not touching the build files does not actually change the "transative closure" of what you depend on. The exact same changes result in the exact same build cache invalidations.

PR #10965, opened 2026-07-30
feat(node-rewards): second temporary node provider reward reduction cohort

Dfinity-Bjoern:
## What

Adds a second temporary node provider reward reduction round — 50% from **2026-08-01** (inclusive) to **2026-11-01** (exclusive) — and generalizes the mechanism added in #10746 from a single hardcoded window to an append-only table of rounds.

No new motion proposal is needed: proposal 142724 established the mechanism, not the set of providers it applies to. Each round's cohort and window are set by this canister upgrade.

## Provider changes relative to the first round (proposal 142724)

**Remain penalized throughout (13)** — in both rounds, so reduced continuously from 2026-07-15 to 2026-11-01:

43rd Big Idea Films · 100 Count Holdings · Arceau NP · Bitmoon · BLP22 · BlockTech Ventures · Decentralized Entities Foundation · Extragone SA · Iancu Aurel · InfoObjects · Neptune Partners · Zarety · Zondax AG

**Recovered (6)** — absent from the new round, so their reduction ends with the first round's window on 2026-10-15 and they are rewarded in full from the second half of October on:

87m Neuron · Conic Ventures · Pindar Technology Limited · Power Meta Corporation · Wancloud limited · Zenith Code LLC

**Newly added (5)**:

| Provider | Principal |
|---|---|
| ACCUSET SOLUTIONS | `cp5ib-twnmx-h4dvd-isef2-tu44u-kb2ka-fise5-m4hta-hnxoq-k45mm-hqe` |
| Anonstake | `kos24-5xact-6aror-uofg2-tnvt6-dq3bk-c2c5z-jtptt-jbqvc-lmegy-qae` |
| MI Servers | `izmhk-lpjum-uo4oy-lviba-yctpc-arg4b-2ywim-vgoiu-gqaj2-gskmw-2qe` |
| Reist Telecom AG | `ma7dp-gz4tg-3c2wv-pgnsv-wna7u-czvhu-fpu47-t4dr6-gzxql-wr2m2-qae` |
| Starbase | `sixix-2nyqd-t2k2v-vlsyz-dssko-ls4hl-hyij4-y7mdp-ja6cj-nsmpf-yae` |

All principals were resolved against the on-chain `list_node_providers`; the 13 carry-over principals are unchanged from #10746.

## How

`REWARD_REDUCTION_START`/`END`/`REDUCED_NODE_PROVIDERS` are replaced by an append-only `const REWARD_REDUCTIONS: &[RewardReduction]`, where each entry carries its own UTC date window, multiplier, and provider set. Everything else about the mechanism is unchanged: it still runs at the canister boundary right after the versioned algorithm returns its `DailyResults`, still scales per-node `adjusted_rewards_xdr_permyriad` and recomputes the provider total from the scaled node values, and still leaves base rewards and the legacy v0 path untouched.

Key properties:

- **Overlapping rounds do not compound.** The smallest multiplier applicable on a given day wins, so a provider caught by both 50% rounds is reduced to 50% — not 25% — on the overlapping days.
- **The first round is left byte-identical.** Nothing in this canister persists computed rewards: every query recomputes the requested days from the stored raw metrics using the table compiled into the running Wasm. Editing or deleting a past entry would therefore silently rewrite the canister's account of days that have already been minted. The design notes in the code state that the table is append-only, and why.
- **No already-minted day is affected.** The new window starts in the future. More generally an entry's `start` must not predate the current, not-yet-minted reward period: governance advances its reward period past every day it has minted and never recomputes those days, so a reduction reaching further back would be a no-op in terms of tokens while still claiming a penalty that was never applied. This is documented alongside the table.
- **Parse-time validation.** Malformed dates, empty windows, and multipliers outside (0, 1) now fail loudly rather than silently paying out the wrong amount.
- **No interface change.** The reduction still rides on the existing `total_adjusted_rewards_xdr_permyriad` (and per-node adjusted) fields.

## Testing

13 unit tests, all passing. Beyond the tests carried over from #10746:

- overlapping windows resolve to 0.5 rather than 0.25, with the union of the two windows reduced and nothing outside it;
- the smallest multiplier wins when rounds differ (0.5 + 0.25 → 0.25, not 0.125);
- three tests against the real compiled-in table: a continuing provider is halved continuously Jul 15 → Nov 1, a recovered provider returns to full rewards on Oct 15, and a newly added provider is untouched before Aug 1;
- round membership is pinned at 19 / 18 / 13-in-both, with no duplicates within a round.

`cargo check --all-targets --all-features`, clippy with the repo's `-D warnings` set, `bazel build //rs/node_rewards/canister:{nrc,nrc-test}`, and all 5 affected Bazel test targets pass locally.

## Open question for review

The 13 continuing providers end up penalized for 3.5 months (2026-07-15 → 2026-11-01) rather than a fresh three, because the first round is preserved as-is. If the intent is instead "three months total, ending 2026-11-01" for them, the alternative is shortening the first round's `end` — which rewrites already-minted days and is what the append-only rule is there to prevent. Flagging it as a policy call rather than deciding it here.

### DIFF
--- a/rs/node_rewards/canister/src/canister/mod.rs
+++ b/rs/node_rewards/canister/src/canister/mod.rs
@@ -205,7 +205,7 @@ impl NodeRewardsCanister {
             )),
         }?;
 
-        // Apply the temporary node provider reward reduction mandated by NNS motion
+        // Apply the temporary node provider reward reductions introduced under NNS motion
         // proposal 142724. This is applied here, at the canister boundary, rather than inside the
         // versioned reward-calculation algorithm, so the algorithm stays pure and reproducible.
         apply_node_provider_reward_reduction(&mut daily_results, date);
@@ -215,13 +215,15 @@ impl NodeRewardsCanister {
 }
 
 // ================================================================================================
-// Temporary node provider reward reductions (NNS motions)
+// Temporary node provider reward reductions (NNS motion proposal 142724)
 // ================================================================================================
 //
-// NNS motions on node provider standards may mandate a temporary reduction of node provider
-// rewards for the providers that failed to meet those standards, e.g. that failed to respond to an
-// incident-response smoke test. Each such motion contributes one cohort to `REWARD_REDUCTIONS`
-// below: a set of node providers, a fixed UTC date window, and a multiplier.
+// The motion "Follow-up on Node Provider Standards, Incident Response Readiness" (proposal 142724)
+// established this mechanism: a temporary reduction of node provider rewards for the providers
+// that failed to meet the incident-response standards. The motion set the mechanism, not the
+// providers it applies to — each round's cohort and window are set by upgrading this canister,
+// and no further motion is needed to add one. Every round therefore appends one entry to
+// `REWARD_REDUCTIONS` below: a set of node providers, a fixed UTC date window, and a multiplier.
 //
 // Design notes:
 //   * Reductions are applied here, at the canister boundary, on top of the performance-based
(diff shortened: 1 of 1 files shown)


pietrodimarco-dfinity (approved):

robin-kunzler (commented):
lgtm, thanks!

PR #10966, opened 2026-07-30
feat: Consensus cost of flexible HTTP outcalls must be covered by allowance

eichhorl:
## Background

When delivering an HTTP outcalls response to the DSM, consensus includes an "initial spent" report. This is the sum of cycles spent by each individual replica to produce the share, plus the cost of reaching consensus after aggregating enough shares. The consensus cost scales with the response and subnet size. 

In https://github.com/dfinity/ic/pull/10967, we started to enforce that a non-flexible outcall response can only be included in a block, if its consensus cost is covered by the unspent allowance of all shares contributing to the response. If the cost is not covered, then the response must wait for more shares (thus, more unspent allowance) to arrive.

## Proposed Changes

With this PR we start to enforce the same principal also for responses of flexible outcalls. However, in contrast to non-flexible outcalls (which only contain one response body), flexible outcalls may contain anywhere between `min` and `max` responses (= `k`). One idea to enforce the consensus cost could be to determine the correct `k` responses which,
1. fit into the payload, and
2. have enough unspent allowance to fund themselves.

However, such a configuration is difficult to find, and also not guaranteed to exist (i.e. if the consensus cost of any `k` responses outweighs the unspent allowance of those same responses).

Therefore the approach proposed here is to add a new field `extra_shares` to all flexible outcall result types which already contain full responses (i.e. Ok results, and too-many-rejects errors). These `extra_shares` are treated as "free" in terms of their consensus cost (they are already covered by the base cost). This means, that given a set of (under-funded) `k` responses, we can add more "extra shares" in order to increase the unspent allowance of the response, without also increasing its consensus cost.

In other terms, responses for a request with (`min`, `max`) configuration are determined as follows:
1. Order all **Ok responses** to the outcall by their size (ascending)
2. Order **all shares** to the outcall by their unspent allowance (descending)
3. Select the smallest `max` responses to include in the payload from the first collection
4. See if the consensus cost of these responses is covered by their unspent allowances, if not, add extra shares from the second collection (excluding shares that were already selected in 3), until the consensus cost is covered.
5. If the consensus cost still isn't covered after adding extra shares, or the result doesn't fit into the payload, set `max = max-1`.
6. Repeat until `max < min`.

The same procedure can be applied for determining rejects for a too-many-rejects result, by setting `min = total_requests - min + 1`.

## Future Work

Currently, if the consensus cost of a response can never be covered, then the request will time out after 60 seconds. Instead, we will introduce a dedicated error that is returned, once such an event is detected.


rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -743,6 +769,26 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                    &mut seen_signers,
+                    context,
+                )
+                .map_err(CanisterHttpPayloadValidationError::InvalidArtifact)?;
+            }
+            // Defer signature verification.
+            sig_inputs.extend(response_share_sig_inputs(
+                shares_without_response,
+                context.registry_version,
+            ));
```
pierugo-dfinity, line 790: As you prefer.

This adds a level of indirection into `fn shares_without_delivered_responses` to really understand what shares there are and needs to be validated. Would it help to duplicate this into the match arms `TooManyRejects` and `ResponsesTooLarge` and validate there? In some sense, move the responsibility to each match arm as each of them know what to validate (responses + extra shares vs seen shares respectively)

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -556,90 +561,101 @@ pub(crate) fn find_flexible_result(
...
-                    .collect(),
-                initial_spent,
-            },
-            ok_responses_size,
-        );
+                responses: into_responses_with_proof(ok_candidates, selection.count),
+                extra_shares: selection.extra_shares.into_iter().cloned().collect(),
+                initial_spent: selection.initial_spent,
+            };
+            let group_size = group.count_bytes();
```
pierugo-dfinity, line 612: It looks like previously, `ok_responses_size` was the sum of all `responses`' (proofs') `count_bytes`, without including the `base_count_bytes`. The latter is now included, is this expected?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -656,6 +672,141 @@ pub(crate) fn find_flexible_result(
...
+    candidates: &[FlexibleCandidate<'a>],
+    all_shares: &[&'a CanisterHttpResponseShare],
+    callback_id: CallbackId,
+    min_to_select: usize,
+    max_to_select: usize,
+    min_responses: u32,
+    accumulated_size: usize,
+    max_payload_size: NumBytes,
+    context: &CanisterHttpRequestContext,
+) -> Option<FlexibleSelection<'a>> {
```
pierugo-dfinity, line 718: If I understand properly, this function returns the extra shares that should be included if the caller selects precisely the `count` first elements of `candidates`. I.e. it trusts the caller to do the right thing. Maybe this function could take ownership of `candidates` and directly return the first `count` of them, next to the `extra_shares`. That would avoid extra processing on the callers and have the selection more encapsulated to this function.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -656,6 +672,141 @@ pub(crate) fn find_flexible_result(
...
+                count: selected.len(),
+                extra_shares,
+                initial_spent: selected_spend + extra_spent,
+            });
+        }
+    }
+    None
+}
+
+/// Picks the fewest of `all_shares` required to cover the collective spend of
```
pierugo-dfinity, line 770: Do we pick the fewest number of shares to optimize for space in the block?

pierugo-dfinity (commented):
This is really cool

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: There will be another branch for the out of cycles error (and any future error case would most likely need it too), which is why I think the deduplication is worth it

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: It used to be initialized with the `base_count_bytes`, or do you mean something else?
https://github.com/dfinity/ic/blob/1e1a813088f35d1024f9591a0a55617d37f0fb9e/rs/https_outcalls/consensus/src/payload_builder/utils.rs#L540

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Unfortunately we can't really move the `candidates` into the function, because they are also needed for subsequent cases if response selection fails

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Yes, exactly

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity: Ok I see

rs/https_outcalls/consensus/src/payload_builder/utils.rs
pierugo-dfinity: Didn't see that :D All good then

rs/https_outcalls/consensus/src/payload_builder/utils.rs
pierugo-dfinity: Ah good point. In that case, we could move the third check (for `ResponsesTooLarge`) before the first two. I do not think this should lead to a change of behaviour, except that `ResponsesTooLarge` would take priority over `TooManyRejects` in case both errors happen.

But I'm also fine with the current implementation

pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: We should prioritize `TooManyRejects` in that case, so I will keep it as is for now

PR #10967, opened 2026-07-30
feat: Consensus cost of non-flexible HTTP outcalls must be covered by allowance

eichhorl:
## Background

When delivering an HTTP outcalls response to the DSM, consensus includes an "initial spent" report. This is the sum of cycles spent by each individual replica to produce the share, plus the cost of reaching consensus after aggregating enough shares. The consensus cost scales with the response and subnet size. 

## Proposed Changes

In order to guarantee that we only produce responses for which the consensus cost can be paid for by the caller (by means of the attached cycles), a refinement of this logic is required.

Initially, the cycles attached to the call are split into the per-replica-allowance (`= attached_cycles / committee_size`). On charging subnets, we already guarantee that no replica's share can claim to have used more than this allowance. When aggregating a set of shares, we can derive the amount of cycles that have not yet been consumed by the participating nodes: `unspent_allowance = sum(per_replica_allowance - share_i.spent)`.

With this PR we start to enforce that a response can only be produced, if the consensus cost is covered by this unspent allowance. If the cost is not covered, then the response must wait for more shares (thus, more unspent allowance) to arrive. If this never happens, then the request will produce a timeout response instead.

Note that this is only enforced for fully-replicated/non-replicated requests with pay-as-you-go pricing on a charging subnet (which cannot exist yet). 

Legacy pricing does not enforce the consensus cost to be covered (but raises a warning and metric in case it isn't).

On free subnets the allowance is zero, so the consensus cost isn't enforced either.

Additionally, with this PR the outcalls adapter will start to return the payment receipt (i.e. the amount of cycles spent) as calculated by the pay-as-you-go tracker. This allows for better observability of canisters that currently supply insufficient funding to their call, and have to be updated. At the same time, returning the pay-as-you-go spend is safe even on legacy pricing (which always reported 0 spending until now), because no cycles are refunded in the first place.

## Future Work

The same principal (consensus cost must be covered by unspent allowances of included shares) must also apply to flexible outcalls. This is significantly more complex (as a flexible outcall response may contain multiple response payloads), and will therefore be addressed in a follow up.


rs/https_outcalls/pricing/src/dark_launch.rs
```diff
@@ -112,7 +112,9 @@ impl BudgetTracker for DarkLaunchTracker {
     }
 
     fn create_payment_receipt(&self) -> CanisterHttpPaymentReceipt {
-        self.real.create_payment_receipt()
+        // Return the shadow's receipt for observability.
+        // This is safe because receipts are unused under legacy pricing.
+        self.shadow.create_payment_receipt()
```
mraszyk, line 117: I'm not sure about this change.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -5250,6 +5252,305 @@ fn flexible_error_too_many_rejects_invalid_signature() {
...
+    num_nodes: usize,
+    context: CanisterHttpRequestContext,
+    expected_responses: usize,
+) {
+    let cb_id = 0;
+    let threshold = num_nodes - get_faults_tolerated(num_nodes);
+    let (response, metadata) = test_response_and_metadata(cb_id);
+    // The consensus cost is nonzero, so a zero collective allowance can only
+    // cover it if the request is not subject to the limit at all.
+    assert!(!non_flexible_consensus_cost(num_nodes, metadata.content_size).is_zero());
```
mraszyk: I didn't get this => add a note to the doc comment that the allowance is expected to be zero.

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -509,6 +513,14 @@ impl CanisterHttpPayloadBuilderImpl {
                     expected,
                 });
             }
+            utils::check_initial_spent_within_limit(
+                response.initial_spent,
+                response.proof.signatures.len(),
+                callback_id,
+                request_context,
+                None,
```
mraszyk, line 521: I suppose an invalid artifact would alert, aren't logs/metrics useful in such a case?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -339,59 +418,81 @@ pub(crate) fn group_shares_by_callback_id<
...
 
 /// Finds a non-replicated HTTP outcall response from the designated node.
 ///
 /// Looks through the grouped shares for one signed by `designated_node_id`.
 /// If found, returns the assembled [`CanisterHttpResponseWithConsensus`].
+///
+/// As for fully-replicated responses, the response is held back while its
+/// collective spend exceeds the designated replica's allowance. Since no other
+/// replica contributes an allowance to a non-replicated outcall, such a request
+/// can only ever time out.
```
mraszyk, line 453: > such a request can only ever time out.

wouldn't it make sense to short-circuit in this special case (in a follow-up PR)?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -339,59 +418,81 @@ pub(crate) fn group_shares_by_callback_id<
 /// where at least `threshold` distinct replicas produced the same
 /// response hash. If found, returns the assembled
 /// [`CanisterHttpResponseWithConsensus`].
+///
+/// A response is only returned once the signing replicas left enough of their
+/// allowances unspent to cover the consensus cost of putting it into a block
+/// (see [`check_initial_spent_within_limit`]); otherwise it is held back until
+/// further shares have arrived, each of which contributes another allowance.
```
mraszyk, line 425: > otherwise it is held back until further shares have arrived

but hasn't some consensus cost already been incurred by exchanging the shares (e.g., by using bandwidth to exchange the actual per-replica responses)? how much do we refund upon timeout if we do not ever manage to cover the consensus cost after all replicas provide their spent cycles?

mraszyk (commented):
This change looks reasonable to me:
- I didn't get one change that I flagged explicitly - if it is unrelated, then maybe it could be moved to a separate PR;
- I also left a few questions for my understanding.

In any case, an extra pair of eyes that is more familiar with this part of the codebase would be useful.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: I agree that's not a good comment. The point is just to ensure that the setup generates a response that has an actual consensus cost (i.e. some content size greater zero), such that the test only passes if there is either enough unspent allowance, or the consensus cost is not enforced (legacy and free requests).

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Yes, the invalid artifact will alert and create a log message, so we wouldn't need it again here. But in this function we anyway only log the specific `Ok` case, where a legacy request would not be able to pay for its consensus cost, if it had been made with pay-as-you-go pricing and the same amount of cycles instead

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Yes, and for flexible outcalls such a short circuit will exist. Doing the same for (non-)replicated outcalls requires changing the response structure which is definitely enough work for a follow up.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: The cost for exchanging shares is mostly part of the per-replica fee, see [here](https://github.com/dfinity/ic/blob/91c7a2cb1c880dff8525e5630fcc91f96504a243/rs/https_outcalls/pricing/src/payg.rs#L116-L130), which cannot exceed the per replica allowance (on charging subnets). So any additional share included in the response can only increase the amount of unspent allowance.

Some (unavoidable) consensus cost is also charged already upfront, as part of the [base fee](https://github.com/dfinity/ic/blob/91c7a2cb1c880dff8525e5630fcc91f96504a243/rs/https_outcalls/pricing/src/fees.rs#L49-L56).

rs/https_outcalls/pricing/src/dark_launch.rs
eichhorl: Yes, this is the last part of the "proposed changes" in the description above. There is some more background here: https://github.com/dfinity/ic/pull/10519.

For pay-as-you-go pricing, when the HTTP request is made by a node, it uses a "BudgetTracker" that subtracts cycles from its per-replica allowance for each step (i.e. making the request, transforming the response, gossiping the response). If the tracker runs out of cycles at any of these steps, it will gossip a reject response instead. In the end, a node also gossips the amount of cycles that it spent this way.

For legacy pricing, the tracker doesn't return an error even if it runs out of cycles (to not break existing calls), and simply reports zero consumed cycles in the end.

Currently, we run both trackers in parallel so that we can compare which canisters would break with the new pricing. But until the change above, we would return the receipt of the legacy tracker (i.e. zero cycles spent).

It makes sense to change this now, meaning nodes will gossip the actual cycles usage out of their allowance, because now is the first time we actually start to observe it, with the metric and log introduced by this PR.

Doing this should still be fine for existing legacy requests, because in the end, the spent cycles that are reported by legacy requests are ignored by the DSM.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -5250,6 +5252,324 @@ fn flexible_error_too_many_rejects_invalid_signature() {
...
+
+            // Consensus is reached, but the signers' collective allowance does
+            // not cover the consensus cost yet.
+            let payload = build_and_validate_and_parse_payload(&payload_builder);
+            assert_eq!(payload.num_responses(), 0);
+
+            // The remaining shares tip the collective allowance over the cost.
+            {
+                let mut pool_access = canister_http_pool.write().unwrap();
+                add_received_shares_to_pool(pool_access.deref_mut(), shares[threshold..].to_vec());
```
pierugo-dfinity, line 5320:
```suggestion
                add_received_shares_to_pool(pool_access.deref_mut(), vec![shares[threshold]]);
```
should be sufficient right?

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -5250,6 +5252,324 @@ fn flexible_error_too_many_rejects_invalid_signature() {
...
+    let consensus_cost = non_flexible_consensus_cost(num_nodes, metadata.content_size);
+    let allowance = Cycles::new(consensus_cost.get().div_ceil(threshold as u128));
+    assert!(allowance * threshold >= consensus_cost);
+    assert!((allowance - Cycles::new(1)) * threshold < consensus_cost);
+
+    assert_responses_from_threshold_shares(
+        num_nodes,
+        with_payg_allowance(request_context(Replication::FullyReplicated), allowance),
+        1,
+    );
```
pierugo-dfinity, line 5411: Maybe also
```suggestion
    assert_responses_from_threshold_shares(
        num_nodes,
        with_payg_allowance(request_context(Replication::FullyReplicated), allowance),
        1,
    );
    assert_responses_from_threshold_shares(
        num_nodes,
        with_payg_allowance(request_context(Replication::FullyReplicated), allowance - 1),
        0,
    );
```
? Then it looks like this is very similar to what `fully_replicated_response_waits_for_shares_covering_consensus_cost` is doing

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -93,6 +97,81 @@ pub(crate) fn check_spent_within_limit(
...
+    num_replicas: usize,
+    callback_id: CallbackId,
+    context: &CanisterHttpRequestContext,
+    obs: Option<(&ReplicaLogger, &CanisterHttpPayloadBuilderMetrics)>,
+) -> Result<(), InvalidCanisterHttpPayloadReason> {
+    match per_replica_consensus_allowance(context) {
+        // Pay-as-you-go enforces that the collective spend must be covered
+        // by the collective allowance of contributing replicas.
+        ConsensusCostAllowance::PayAsYouGo(allowance)
+            if initial_spent > allowance * num_replicas =>
```
pierugo-dfinity, line 142: I was thinking whether this multiplication could overflow and raise false positives. It looks like the implementation of `Mul<usize> for Cycles` saturates the result so there is no way it can lead to a false positive with the current implementation, but I wonder whether it would be more defensive to check for `if initial_spent / num_replicas > allowance` instead.

pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: But that would be wrong if `initial_spend = num_replicas * allowance + 1` for instance. Since `allowance` is defined as `attached_cycles / committee_size`, the multiplication should always work.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Added: https://github.com/dfinity/ic/pull/10967/commits/d0a4bc5c1a337e399ab5b14beed51213f372ac55

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Only for f = 1 (which happens to be the case here)

rs/https_outcalls/consensus/src/payload_builder.rs
mraszyk: Claude says that we could just pass `Some((&self.log, &self.metrics))` - wouldn't it make sense to do so?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
mraszyk: Wouldn't it be easier to then only rely on the per-replica fee and not charge any additional cost that could make the request time out?

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: No we should not pass `Some((&self.log, &self.metrics))` here because the invalid artifact will already alert and create a log message. We also shouldn't log the canister compatibility case, because that is already logged during payload building, so it only adds more logs here that will make it harder to analyze.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Yes, that would be easier, but at the time of creating the share (and response), we do not know yet if the response will be included in the payload, so at that point we can only overcharge.

rs/https_outcalls/consensus/src/payload_builder.rs
mraszyk: > that is already logged during payload building, so it only adds more logs here that will make it harder to analyze

I see. Thanks for explaining!

rs/https_outcalls/consensus/src/payload_builder/utils.rs
mraszyk: If I understand correctly, we do not want to overcharge in case there are many concurrent http outcalls and not all of them actually make it into the payload which is a fair trade-off.

rs/https_outcalls/pricing/src/dark_launch.rs
mraszyk: Got it, sounds good.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: No, results of concurrent http outcalls will be included in subsequent payloads, if the current payload is already full. But the cost of an outcall depends on which response body (or rather its size), and in the flexible case how many response bodies, are included. We can only know this once we have enough responses to actually aggregate them. So trying to charge for it in advance will always be inaccurate.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
pierugo-dfinity: I misinterpreted the test case, that makes sense, thanks

PR #10970, opened 2026-07-30
feat(subnet-splitting): do not validate certifications of other subnet

pierugo-dfinity:
After a subnet split, nodes are still connected under the same P2P network and share the same pools of artifacts. It is only after the orchestrator detects that the split has happened (sees a `PostSplit` CUP for a different subnet) that it restarts the *destination* replicas and then the two subnets are strictly separated.

During that short window of time, the source subnet actually executes messages and it is important for the destination nodes not to create/validate any certifications/certification shares during that time.


rs/consensus/utils/src/lib.rs
```diff
@@ -325,6 +328,14 @@ pub fn active_high_threshold_committee(
     })
 }
 
+/// Return the current high transcript for the given height if it was found.
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10970/commits/3416609e9ad1c979388f920af89022bdd01e4f47

rs/consensus/certification/src/certifier.rs
```diff
@@ -479,6 +493,32 @@ impl CertifierImpl {
...
+        // shares). In particular because after a split, before replicas get restarted, they are
+        // still under the same P2P network and can gossip certifications for states of different
+        // subnets.
+        match self.should_skip_due_to_subnet_splitting(certification.height) {
+            Ok(true) => {
+                info!(
+                    every_n_seconds => 30,
+                    self.log,
+                    "Skipping the validation of a certification at height {} because a \
+                    subnet splitting is taking place",
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10970/commits/94f807908a358f8d06415305d5fee086267c2c64

rs/consensus/certification/src/certifier.rs
```diff
@@ -582,6 +648,24 @@ impl CertifierImpl {
             }
         }
     }
+
+    /// Checks if we should skip the creation and/or validation of certifications/shares
+    /// at the given height, due to an ongoing subnet splitting.
+    fn should_skip_due_to_subnet_splitting(&self, height: Height) -> Result<bool, String> {
```
eichhorl: Maybe we could avoid some duplication by moving the logging here, and returning a boolean directly

rs/consensus/certification/src/certifier.rs
```diff
@@ -479,6 +493,32 @@ impl CertifierImpl {
...
+                );
+                return None;
+            }
+            Ok(false) => {}
+            Err(err) => {
+                warn!(
+                    self.log,
+                    "Failed to check the subnet splitting status: {err}. \
+                    Skipping validation of the certificate"
+                );
```
eichhorl: Should this also have `every_n_seconds`?
```suggestion
                warn!(
                    every_n_seconds => 30,
                    self.log,
                    "Failed to check the subnet splitting status: {err}. \
                    Skipping validation of the certificate"
                );
```

rs/consensus/certification/src/certifier.rs
```diff
@@ -1557,4 +1646,305 @@ mod tests {
...
+
+    fn assert_for_all_subnet_splitting_statuses(
+        pool: &mut TestConsensusPool,
+        mut test: impl FnMut(SubnetSplittingStatus, Height),
+    ) {
+        for status in [
+            not_scheduled_splitting(),
+            scheduled_splitting(),
+            done_splitting_different_subnet(),
+            done_splitting_same_subnet(),
```
eichhorl: Maybe it would be easier to encode the expectation (should certifications be skipped) here, i.e.:
```suggestion
            (not_scheduled_splitting(), false),
            (scheduled_splitting(), true),
            (done_splitting_different_subnet(), true),
            (done_splitting_same_subnet(), false),
```
And just pass the expectation to `test`

rs/consensus/certification/src/certifier.rs
```diff
@@ -1557,4 +1646,305 @@ mod tests {
...
+                                    CertificationMessage::Certification(cert.clone())
+                                )),
+                                "Expected MoveToValidated when not splitting or splitting with same subnet ID"
+                            );
+                        }
+                    }
+                });
+            })
+        })
+    }
```
eichhorl, line 1930: We could also add a test for when the status is unknown

rs/consensus/certification/src/certifier.rs
pierugo-dfinity: Good idea! https://github.com/dfinity/ic/pull/10970/commits/0a14357d4dac5d790f92d4d42d06fdfd0616a780

rs/consensus/certification/src/certifier.rs
pierugo-dfinity: Done, thanks

rs/consensus/certification/src/certifier.rs
pierugo-dfinity: Yep good idea https://github.com/dfinity/ic/pull/10970/commits/c26eb4f2f652a1e5855dde6bfca252d4918d459c

rs/consensus/certification/src/certifier.rs
pierugo-dfinity: WDYT https://github.com/dfinity/ic/pull/10970/commits/3fdeff68df247a297e46b9e5778e812393cc94ed

rs/consensus/certification/src/certifier.rs
```diff
@@ -582,6 +604,59 @@ impl CertifierImpl {
...
+            None => {
+                warn!(
+                    every_n_seconds => 30,
+                    self.log,
+                    "Missing finalized summary block for height {height}. \
+                    Skipping creation/validation of certifications/shares"
+                );
+
+                true
+            }
```
eichhorl, line 621: Just pointing out that this case will most likely trigger quite regularly, after we introduced the change where we will try to validate certifications up to 20 heights in advance (so that we can rely on their state hashes instead of hashing ourselves during catchup). Specifically it will happen for heights above the summary, while we do not have the summary yet.

https://github.com/dfinity/ic/blob/3fdeff68df247a297e46b9e5778e812393cc94ed/rs/state_manager/src/lib.rs#L133-L136

But I think it should not make catching up significantly more difficult, because usually the bottleneck is execution (for which we need the summary block anyway).

eichhorl (approved):

rs/consensus/certification/src/certifier.rs
eichhorl: Actually nevermind:

https://github.com/dfinity/ic/blob/3fdeff68df247a297e46b9e5778e812393cc94ed/rs/state_manager/src/lib.rs#L3196-L3200

We take the minimum of tip+20, and the fast forward height (which is the latest certified-finalized height). So we will never request certifications for heights without blocks already today.

pierugo-dfinity: Thanks for catching this! Hmmm what about if execution is slow (below summary height `H`) and consensus is already at height `> H`? Wouldn't we stop validating certifications for heights we are currently executing (because the latest summary wouldn't cover those heights)? Technically, this situation should trigger a state sync, but we wouldn't if execution is very close to `H` and the CUP is held back.
Not sure this would necessarily lead to bad behaviour though.

pierugo-dfinity: @eichhorl

eichhorl: But in the case where the CUP is held back, it means that we should still have the previous CUP (thus summary block), and should therefore still be able to tell if there is a subnet split going on in the previous interval.

pierugo-dfinity: Of course, that makes sense. Thank you!

PR #10978, opened 2026-07-30
feat(migration-canister): export a metric for the oldest request in flight and add endpoint for listing requests

mraszyk:
Adds the gauge `migration_canister_oldest_request_in_flight_age_seconds` to the metrics served via `http_request`, reporting the age of the migration request that has been in flight the longest. The gauge is not set if there is no request in flight.

The age is derived from a new stable map `ACCEPTED_TIME` recording, for every request in `REQUESTS`, the time at which it was accepted. The timestamp is retained across state transitions (which remove and re-insert the request) and dropped once the request leaves `REQUESTS` for good, i.e., when the corresponding event is recorded in `HISTORY` (both on success and on failure). Requests that were already in flight before this bookkeeping was introduced are backfilled upon their next state transition.

The motivation for adding the gauge is ability to introduce alerts for request that are stuck for a long time.


schneiderstefan (approved):
May I suggest a follow-up in the k8s repo to add a new panel to the canister migration dashboard?

PR #10980, opened 2026-07-31
feat(subnet-splitting): implement `PostSplit` CUPs making & validation

pierugo-dfinity:
This PR implements the core logic behind subnet splitting in Consensus. It affects the following components, in suggested order of review:
- DKG summary payload builder
- CUP-shares maker
- CUP-shares validator
- CUP-shares aggregator
- CUP validator
- Bouncer

## DKG summary payload builder
When a subnet split is detected in the registry, the validation context's registry version is frozen until reaching the next summary, at which point it is [bumped](https://github.com/dfinity/ic/blob/aae9fb37f764efa1fa30c934fcfc5ca73424ea44/rs/consensus/src/consensus/block_maker.rs#L579-L583) to precisely the version it was scheduled. The payload builder thus fills the `subnet_splitting_status` field with `Scheduled` if there is a scheduled split at the validation context's registry version. Otherwise, it fills it with `NotScheduled`. It should never fill it with `PostSplit`, this will be the role of the CUP-shares maker.

## CUP-shares maker
Instead of unconditionally creating a CUP for finalized summary blocks, the CUP-shares maker now looks at the block's `subnet_splitting_status` field. If it is `Scheduled`, then it will *not* create a CUP for it. It will instead create a CUP for a "post-split" block for `H+500` which is not notarized/finalized like usual, but is instead computed locally (it is deterministic) and its hash is gossiped as part of the CUP share.
CUP-shares makers will create such shares based on the subnet they will take part in after the split. This means that there will be two distinct sets of CUP shares gossiped around, one for each resulting subnet. Each set of CUP shares contains the relevant catchup information for the new subnet: the DKG transcripts taken from the registry (not the ones from the original summary), the state hash and oldest registry version in use from the split state at the height of the original summary, and a dummy random beacon.

## CUP-shares validator
As soon as the `Scheduled` summary is reached, half the CUP-shares maker will create shares for a CUP A and the other half for a CUP B. The role of the CUP-shares validator will be to invalidate the CUP shares intended for the other subnet and validate the ones intended for its own subnet. To do so, it recreates the post-split block, state hash, etc and only validates shares which match these. The other subnet's CUP shares will be invalidated as the block, beacon and state hash will be different.

## CUP-shares aggregator
This component is not changed by much but its logic still had to be slightly rewritten to consider these new post-split CUPs.

## CUP validator
Similar to the CUP-shares validator, here as soon as enough shares for the two CUPs are gossiped, they will be aggregated and the two CUPs will be gossiped around. The role of the CUP validator will be to invalidate the CUP intended for the other subnet and validate the one intended for its own subnet. To do so, we look at the CUP's block's `subnet_splitting_status` as it contains the subnet the CUP is intended for (`PostSplit { new_subnet_id }`).

## Bouncer
Finally, the bouncer needs to tell P2P to start downloading CUP shares also for the next summary height since this is the height at which the post-split CUP will be created.

## Remaining changes
Follow-ups will adapt
- DKG key manager to preemptively load the post-split transcripts as soon as the `Scheduled` summary is reached and actually be able to sign the post-split CUP shares
- Batch delivery to tell DSM to actually split the state
- The orchestrator to restart destination replicas and update their `ReplicaConfig.subnet_id`


rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -206,87 +205,237 @@ impl CatchUpPackageMaker {
...
+            .get_cup_random_beacon(pool, &cup_block, cup_type)
+            .inspect_err(|err| warn!(self.log, "Could't get a random beacon for a CUP: {err}"))
+            .ok()?;
+
+        let high_threshold_transcript =
+            get_current_transcript_from_summary_block(&cup_block, &NiDkgTag::HighThreshold)
+                .or_else(|| {
+                    warn!(self.log, "Could't find transcript at height {height}");
+                    None
+                })?;
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/01be3269dd47930db15b27243f7e87d71670ab84

rs/types/types/src/consensus/catchup.rs
```diff
@@ -422,6 +422,17 @@ impl TryFrom<subnet_pb::SubnetSplittingArgs> for SubnetSplittingArgs {
     }
 }
 
+/// Types of [`CatchUpPackage`] created by the CUP maker
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum CatchUpPackageType {
+    Normal,
+    /// After deliverying a splitting block to the DSM, we immediately create a CUP at the start of
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/01be3269dd47930db15b27243f7e87d71670ab84

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -231,11 +271,30 @@ impl DkgKeyManager {
             // the next transcript before the previous removal (which would consider the
             // next transcript key irrelevant and remove it).
             self.delete_inactive_keys(pool_reader);
-            self.load_transcripts_from_summary(&summary.dkg);
-            self.last_dkg_summary_height = Some(summary_block.height);
+            self.load_transcripts_from_summary(&summary);
```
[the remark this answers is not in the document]
pierugo-dfinity: That's a good point. Addressed in https://github.com/dfinity/ic/pull/10980/commits/03857ee5cc9eeb9c95239f39f594ed06f2259274

rs/consensus/src/consensus/validator.rs
```diff
@@ -317,19 +340,59 @@ impl SignatureVerify for CatchUpPackage {
         membership: &Membership,
         crypto: &dyn ConsensusCrypto,
         _pool: &PoolReader<'_>,
-        _cfg: &ReplicaConfig,
+        cfg: &ReplicaConfig,
     ) -> ValidationResult<ValidatorError> {
+        let cup_registry_version = self.content.registry_version();
```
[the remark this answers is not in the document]
pierugo-dfinity: Thank you https://github.com/dfinity/ic/pull/10980/commits/ae301cbda3d6d317b72da466fc3eefb97c46dc0e

rs/consensus/src/consensus/priority.rs
```diff
@@ -53,8 +56,10 @@ fn compute_bouncer(
     }
     // Stash non-CUP artifacts, as long as they're too far ahead of the next pending CUP height.
     // This prevents nodes that have fallen behind from exceeding their validated pool bounds.
-    if !matches!(id.hash, ConsensusMessageHash::CatchUpPackage(_))
-        && height > next_cup_height + Height::new(ACCEPTABLE_NOTARIZATION_CUP_GAP)
+    if !matches!(
+        id.hash,
+        ConsensusMessageHash::CatchUpPackage(_) | ConsensusMessageHash::CatchUpPackageShare(_)
+    ) && height > next_cup_height + Height::new(ACCEPTABLE_NOTARIZATION_CUP_GAP)
```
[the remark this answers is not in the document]
pierugo-dfinity: Risk accepted. We anyways need to look-ahead for the `next_summary_height`. The alternative would be to `Wants` on  `height <= finalized_height || height == next_summary_height`, so I'd keep the current approach for simplicity.

rs/consensus/dkg/src/dkg_key_manager.rs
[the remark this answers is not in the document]
pierugo-dfinity: The DKG key manager part was moved to a separate [PR](https://github.com/dfinity/ic/pull/11205)

rs/consensus/src/consensus.rs
```diff
@@ -272,7 +273,7 @@ impl ConsensusImpl {
             validator: Validator::new(
                 replica_config.clone(),
                 membership.clone(),
-                Arc::clone(&registry_client),
+                registry_client.clone(),
```
eichhorl, line 276: I think the convention is actually `Arc::clone`, but we don't seem to follow it so no action required.

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -206,87 +205,237 @@ impl CatchUpPackageMaker {
...
+                    self.log,
+                    "Couldn't create a signature at height {height}: {err}"
+                )
+            })
+            .ok()?;
+
+        debug!(
+            self.log,
+            "Proposing a CatchUpPackageShare (type: {cup_type:?}) at height {height}"
+        );
```
eichhorl, line 351: I don't think `debug` logs show up anywhere

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -206,87 +205,237 @@ impl CatchUpPackageMaker {
...
+            return None;
+        }
+
+        // Skip if this node has already made a share
+        if pool
+            .get_catch_up_package_shares(height)
+            .any(|share| share.signature.signer == my_node_id)
+        {
+            return None;
+        }
```
eichhorl: Maybe we can still do this check more towards the top of the function

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -144,38 +158,23 @@ impl CatchUpPackageMaker {
...
-
-        // Skip if random beacon does not exist for the height
-        let random_beacon = pool.get_random_beacon(height)?;
+        let summary_height = start_block.height();
+        let cup_type = get_catch_up_package_type(
+            self.registry.as_ref(),
+            self.replica_config.node_id,
+            &start_block,
+        )
+        .inspect_err(|err| warn!(self.log, "Failed to get the catch up package type: {err}"))
```
eichhorl: Adding `every_n_seconds` would avoid flooding the logs, if this ever were to happen. Same for the logs below

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -794,4 +961,301 @@ mod tests {
...
+
+                let message_routing = FakeMessageRouting::new();
+                *message_routing.next_batch_height.write().unwrap() = Height::from(2);
+                let message_routing = Arc::new(message_routing);
+
+                let cup_maker = CatchUpPackageMaker::new(
+                    ReplicaConfig {
+                        node_id: NODE_5,
+                        subnet_id: SOURCE_SUBNET_ID,
+                    },
```
eichhorl:
```suggestion
                    state_manager,
                    replica_config,
                    ..
                } = DependenciesBuilder::multiple_subnets(pool_config, records)
                    .with_replica_config(ReplicaConfig {
                        node_id: NODE_5,
                        subnet_id: SOURCE_SUBNET_ID,
                    })
                    .build();

                let fake_state_hash = CryptoHashOfState::from(CryptoHash(vec![1, 2, 3]));
                state_manager
                    .get_mut()
                    .expect_get_state_hash_at()
                    .return_const(Ok(fake_state_hash.clone()));

                let message_routing = FakeMessageRouting::new();
                *message_routing.next_batch_height.write().unwrap() = Height::from(2);
                let message_routing = Arc::new(message_routing);

                let cup_maker = CatchUpPackageMaker::new(
                    replica_config,
```

rs/consensus/utils/src/subnet_splitting.rs
```diff
@@ -84,19 +89,122 @@ pub fn get_status(
...
+) -> Result<PostSplitAssignment, PostSplitAssignmentError> {
+    debug_assert!(matches!(
+        summary_block
+            .payload
+            .as_ref()
+            .as_summary()
+            .dkg
+            .subnet_splitting_status(),
+        SubnetSplittingStatus::Scheduled(..)
+    ));
```
eichhorl: Should we return an error in this case?

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -870,6 +933,42 @@ pub(crate) fn create_remote_dkg_config(
...
+pub fn get_post_split_dkg_summary(
+    new_subnet_id: SubnetId,
+    registry: &dyn RegistryClient,
+    last_summary_block: &Block,
+) -> Result<DkgSummary, String> {
+    let last_summary = &last_summary_block.payload.as_ref().as_summary().dkg;
+    debug_assert!(matches!(
+        last_summary.subnet_splitting_status(),
+        SubnetSplittingStatus::Scheduled(..)
+    ));
```
eichhorl: Should we return an error in this case?

rs/consensus/utils/src/subnet_splitting.rs
```diff
@@ -84,19 +89,122 @@ pub fn get_status(
...
+    pub other_subnet_id: SubnetId,
+}
+
+#[derive(Debug, Error)]
+pub enum PostSplitAssignmentError {
+    #[error("Error while getting the subnet id from the registry at version {0}: {1}")]
+    FailedToGetSubnetIdFromTheRegistry(RegistryVersion, RegistryClientError),
+    #[error("The node is unassigned at registry version {0}")]
+    Unassigned(RegistryVersion),
+    #[error("The node changed subnets during subnet splitting")]
```
eichhorl:
```suggestion
    #[error("The node changed to subnet {0} during the split, which is neither the source nor destination subnet")]
```

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -134,32 +144,26 @@ impl ShareAggregator {
...
+                break;
+            }
+            match self.aggregate_catch_up_package_shares_for_summary_block(pool, start_block) {
+                Ok(messages) if !messages.is_empty() => {
+                    return to_messages(messages);
+                }
+                Ok(_) => {}
+                Err(err) => {
+                    warn!(
+                        self.log,
```
eichhorl, line 160:
```suggestion
                    warn!(
                        every_n_seconds => ...,
                        self.log,
```

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +181,68 @@ impl ShareAggregator {
...
+                .ok_or_else(|| {
+                    String::from("Couldn't find the high threshold transcript in the summary block")
+                })?;
+        let threshold = transcript.threshold.get().get() as usize;
+        let dkg_id = transcript.dkg_id.clone();
+
+        let shares = pool
+            .get_catch_up_package_shares(block.height())
+            .map(|share| Signed {
+                content: CatchUpContent::from_share_content(share.content, block.clone()),
```
eichhorl: Should we keep the check that the hash in the share actually matches the block (that check was previously hidden in `get_block`)

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -541,6 +542,13 @@ pub(super) fn create_summary_payload(
         &vet_key_ids,
     )?;
 
+    let subnet_splitting_status = get_subnet_splitting_status(
+        subnet_id,
+        registry_client,
+        registry_version,
+        validation_context.registry_version,
+    )?;
+
```
eichhorl, line 551: If the DKG interval is smaller than `ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP`, could it happen that we create valid summary at the "post-split CUP height" the usual way?

Bownairo (approved):

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity: Of course, thanks https://github.com/dfinity/ic/pull/10980/commits/d74b05ebf8ec2c50339643e1570134fcf47aaea0

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity: Let me know what you think of [this](https://github.com/dfinity/ic/pull/10980/commits/2e9983c457cc71b051f981baeb0f3a4365d87abc). The only thing is that now there are two heights floating around: the `summary_height` and the `cup_height` which are sometimes different and each step in the function must carefully choose which one to use.
Also, any future modifications must ensure that `get_cup_height` is consistent with `get_cup_block`, which is error-prone. I added some guardrails.

rs/consensus/src/consensus.rs
pierugo-dfinity: I agree, I changed it really just for consistency with the other arguments (which are almost all `Arc`s)

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity: Right, I just adapted the existing log. I don't think it hurts if we ever decide to also include debug logs in system tests in the future? We already have quite a few `debug!`s around the repo.

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity: Thanks https://github.com/dfinity/ic/pull/10980/commits/913fea5576116ce46bf6d87b0830ec96774baa36

rs/consensus/utils/src/subnet_splitting.rs
pierugo-dfinity: Yes totally. Actually got rid of the debug-panic. https://github.com/dfinity/ic/pull/10980/commits/ae85fde71370313895c0ca67dfabf083c986068e

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: Yes too https://github.com/dfinity/ic/pull/10980/commits/29d6767f5ce4b33217277967d1e3bec125f11aaf

rs/consensus/utils/src/subnet_splitting.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/2323de5cb3178c01fbec8c13e549d1488822f475

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/5c5de9ee0ab59b4cb9735417b49ade82bb8de5da

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: Ah indeed. And if it did not match, we would panic... but this is arguably still better than aggregating a CUP for a block that was not actually agreed on.

I fee like this should be done in the validator (which does it), but this is a simple change so we can add it as a sanity-check. It's just that technically then we could argue to also sanity-check the random beacon, state hash, etc, i.e. redo the work of the validator

https://github.com/dfinity/ic/pull/10980/commits/832982b09520603e80ef138755caaef3fb14d47a

rs/consensus/dkg/src/payload_builder.rs
pierugo-dfinity: We could indeed, we should prevent that. Though I think this is a more general problem that can happen whenever we are halting, i.e. this could also happen on an upgrade:
- Empty blocks are created past the "upgrade height"
- If the interval is very small and the state manager hasn't computed the manifest yet, we reach the next summary height
- Here, we do not check if we should halt or not: we always include a payload (which will probably trigger a failed DKG interval because we do not have any dealings to work with)
- The validators would validate that summary

This is probably not as bad as for subnet-splitting, but is still unwanted. I think this is worth looking into but then maybe as a separate PR?

pierugo-dfinity: Hmm I'll think more about it, because it really does not harm much in regular upgrades, and we can't just "not make a block at the next summary height" because at that point there is still no guarantee that the halting height was finalized.

rs/consensus/src/consensus/share_aggregator.rs
eichhorl: My understanding was that under the situation described in the comment below, this would not just be a sanity check, because in some way we would have two "finalized" blocks at the post split height

rs/consensus/dkg/src/payload_builder.rs
eichhorl: I guess even with the new hash check we would still have the problem that some nodes might create shares for one summary, and some for the others

rs/consensus/src/consensus/catchup_package_maker.rs
eichhorl: Ah, I hadn't noticed that dependency, but the new code works for me 👍

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: > even with the new hash check [...]

Yes, and I think I'd prefer to keep the aggregator agnostic to that and move the responsibility of ensuring CUP-shares are issued for a single block at a given height to the maker & validator.

Though it looks hard to come up with additional custom logic in them to handle potentially duplicate "finalized" (as you say) blocks.
Maybe a very pragmatic solution would be that instead of jumping one interval, we can jump to the next multiple of `interval_length + 1` larger than `ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP`. I.e.
```rust
    cup_contents.height = last_summary.height
                           + (ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP + 1)
                              .div_ceil(last_summary.interval_length + 1)
                              .saturating_mul(last_summary.interval_length + 1);
```
instead of
```rust
    cup_contents.height = last_summary.height + last_summary.interval_length + 1;
```
It could be possible that nodes compute intermediary summaries (all will have `NotScheduled`) but no CUP will be created, because they would miss the state. And the blocks corresponding to these states will not be delivered (because we will be `Halting` at the expected batch height), so we should be fine. One potential problem is that registry version in the validation contexts of these intermediary blocks & summaries could increase. And then the post-split CUP will reset it to the version at which the split was scheduled. DSM will not be confused because we did not deliver those blocks anyways, but Consensus could be confused, in particular if there are any membership changes... Will continue to think about it early next week.

pierugo-dfinity: > Maybe a very pragmatic solution

So I went ahead with it (https://github.com/dfinity/ic/pull/10980/commits/45e7c5ccadcb41b2f8a5bc8f535a20cda056d466). This should avoid the situation of having two "finalized" heights.

> One potential problem is that registry version in the validation contexts of these intermediary blocks & summaries could increase. [...] Consensus could be confused, in particular if there are any membership changes

I also extended the registry version freezing logic to also freeze it on the chain that will be replaced (i.e. any child of a `Scheduled` summary block): https://github.com/dfinity/ic/pull/10980/commits/e6f091dc365049ee78ab3dec436e202e27b2b927, sorry for the big changes... I do not think it should be a big problem not to freeze it as long as we don't reach a new summary block, but it still prevents coming back to a previous registry version, which could have unintended behaviour.

eichhorl: Alternatively we could simply stall the subnet instead of making the second post-split summary, and note this down as a limitation for splitting subnets with a DKG interval length smaller than 70. After all, production subnets have more than that (and will equally stall if nothing gets finalized for 70 heights).

pierugo-dfinity: > and note this down as a limitation

Yeah ok that sounds easier. I can [revert](https://github.com/dfinity/ic/pull/10980/commits/c3e538cbd4ee6f8cf261402829bdfd41ca38f2db) then. Though implementing that longer jump revealed an independent issue fixed in https://github.com/dfinity/ic/pull/10980/commits/c50e59703bc59afe1ceae9e09fa46fa3354124dd.

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +182,76 @@ impl ShareAggregator {
...
+                })?;
+        let threshold = transcript.threshold.get().get() as usize;
+        let dkg_id = transcript.dkg_id.clone();
+
+        let shares = pool
+            .get_catch_up_package_shares(block.height())
+            .filter_map(|share| {
+                // The validator should already perform this check if implemented correctly, so this
+                // is just a sanity check
+                if ic_types::crypto::crypto_hash(&block) != share.content.block {
```
eichhorl: could we compute the hash outside of the loop?

rs/consensus/utils/src/pool_reader.rs
```diff
@@ -558,6 +558,15 @@ impl<'a> PoolReader<'a> {
             .dkg
             .get_next_start_height()
     }
+
+    pub fn get_next_summary_height(&self) -> Height {
```
eichhorl, line 563: Maybe we could add documentation that this is the next summary height according to the highest finalized summary

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -177,6 +182,76 @@ impl ShareAggregator {
...
+        let cups = aggregate_with_threshold(
+            &self.log,
+            self.crypto.as_aggregate(),
+            Box::new(|_| Some(dkg_id.clone())),
+            Box::new(|_| Some(threshold)),
+            shares,
+        );
+
+        for cup in &cups {
+            if let CatchUpPackageType::PostSplit { new_subnet_id } = cup_type {
```
eichhorl: We could also switch the `for` and `if` blocks

rs/consensus/src/consensus/validator.rs
```diff
@@ -4919,4 +5235,367 @@ pub mod test {
...
+                            DESTINATION_SUBNET_ID,
+                            SubnetRecordBuilder::from(&[NODE_3, NODE_4])
+                                .with_dkg_interval_length(DKG_INTERVAL_LENGTH)
+                                .build(),
+                        ),
+                    ],
+                )
+                .with_replica_config(ReplicaConfig {
+                    node_id: NODE_1,
+                    subnet_id: SOURCE_SUBNET_ID,
```
eichhorl: Does it make a difference if the validator is a different node on the destination subnet?

eichhorl (approved):

rs/consensus/src/consensus/share_aggregator.rs
eichhorl: Makes sense, but then we should also make sure to document the limitation, and how it behaves

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: Of course https://github.com/dfinity/ic/pull/10980/commits/297767b4da4f216ab5a87edb4065a3e67cb86b4b

rs/consensus/utils/src/pool_reader.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/e438bf856cae925c75c0d249857d0b8be4a9d6db

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/10980/commits/75e8a590b4181503c3911c191a161681d6aa34ba

rs/consensus/src/consensus/validator.rs
pierugo-dfinity: Duplicated all tests to also test with the destination validators: https://github.com/dfinity/ic/pull/10980/commits/ce0aed22b20c1c3380f99d9399efd398119acd0d

pierugo-dfinity: > different node

These should be more exhaustive https://github.com/dfinity/ic/pull/10980/commits/d528f82c2f843ae15fbad0cdb770c2df33899df1

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: Documented in https://github.com/dfinity/ic/pull/10980/commits/9d4f1194240c5f93f7311043b87fd6761c2331f0 (on `get_post_split_dkg_summary`, where the post-split height is chosen).

rs/consensus/dkg/src/payload_builder.rs
```diff
@@ -870,6 +935,74 @@ pub(crate) fn create_remote_dkg_config(
...
+        .get_cup_contents(new_subnet_id, registry_version)
+        .map_err(|err| {
+            format!("Failed to get the cup contents at registry version {registry_version}: {err}")
+        })?
+        .value
+        .ok_or_else(|| format!("Empty cup contents at registry version {registry_version}"))?;
+
+    // During subnet splitting we skip one DKG interval. The pre-split chain can never reach this
+    // height — see the function documentation for why, and for the limitation this implies.
+    cup_contents.height = last_summary.get_next_start_height().get();
```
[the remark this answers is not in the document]
pierugo-dfinity: Risk accepted

PR #10981, opened 2026-07-31
chore: enable bazel-test-all-rbe again but allow it to fail

basvandijk:
Reverts https://github.com/dfinity/ic/commit/c7ae711733440409618605018bbc5785b946b65c such that the `bazel-test-all-rbe` job is enabled again but configures it as `continue-on-error: true` to not get red checkmarks and slack notifications when the job fails on `master`.

### DIFF
--- a/.github/workflows/ci-rbe-evaluation.yml
+++ b/.github/workflows/ci-rbe-evaluation.yml
@@ -168,6 +168,7 @@ jobs:
 
   bazel-test-all-rbe:
     name: Bazel Test All on RBE
+    continue-on-error: true
     needs: [config, infer-bazel-targets]
     runs-on: namespace-profile-rbe-driver-amd64-linux-16x32
     container:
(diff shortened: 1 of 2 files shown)


mbjorkqvist (approved):

nmattia (approved):

.github/workflows/ci-rbe-evaluation.yml
```diff
@@ -168,6 +168,7 @@ jobs:
 
   bazel-test-all-rbe:
     name: Bazel Test All on RBE
+    continue-on-error: true
```
nmattia, line 171: should we mention that this is temporary?

basvandijk: This whole workflow is temporary. We'll remove it in a week once we integrate RBE properly.

blind-oracle (approved):

PR #10984, opened 2026-07-31
chore: document the namespace_requires_network execution property

basvandijk:


mbjorkqvist (approved):

rs/nns/integration_tests/BUILD.bazel
```diff
@@ -367,6 +367,8 @@ rust_ic_test(
     crate_root = "src/upgrade_canisters_with_golden_nns_state.rs",
     data = DEV_DATA,
     env = DEV_ENV,
+    # Bazel doesn't forward the "requires-network" tag to the Remote Execution API (REAPI)
+    # so we need this Namespace-specific execution property to open up the network on RBE:
```
daniel-wong-dfinity-org-twin: Instead of referring to them as "Namespace", how about "namespace.so" ? Because otherwise, it is a bit unclear that "Namespace" is a proper noun. I mean, yes, it is capitalized, but a reader might think the capital "N" is just a typo, because "namespace" is such a common word in our field, and many people have never heard of namespace.so. It's not a household name like "Google", or something. You won't find it in the dictionary.

daniel-wong-dfinity-org-twin (approved):

rs/nns/integration_tests/BUILD.bazel
daniel-wong-dfinity-org-twin, line 1: On the title of this PR, how about build instead of chore?

rs/nns/integration_tests/BUILD.bazel
basvandijk: Sounds good. WIll do. "Namespace" is not the best name for a company 🙈 .

blind-oracle (approved):

rs/nns/integration_tests/BUILD.bazel
basvandijk: Done.

rs/nns/integration_tests/BUILD.bazel
basvandijk: I tried but that's actually [not allowed](https://github.com/dfinity/ic/actions/runs/30632958862/job/91163481931?pr=10984).

PR #10987, opened 2026-07-31
feat: allow composite queries to call the management canister query methods

mraszyk:
Composite queries can now call the management canister methods that can be invoked by users as non-replicated query calls, i.e. the methods listed in `QueryMethod`: `fetch_canister_logs`, `canister_status`, `list_canisters` and `canister_metrics`. The calls are executed against the state of the own subnet on behalf of the calling canister, so they can only target canisters hosted by the own subnet and they are subject to the same access control as the corresponding queries sent by end users.

Implementation notes:
- The execution of the management canister query methods is factored out of the user query path into the new `subnet_query` module which is shared by both the user query path and the new composite query path.
- Requests to the management canister made by a composite query are no longer routed based on the method and the payload, but resolved to the own subnet and handled by the query handler.
- The instructions consumed while producing the reply (applies to `fetch_canister_logs` and `list_canisters`) are charged towards the instruction limit of the whole call graph, in addition to the base overhead charged for every query call.
- Query contexts calling the management canister are never cached: their results may depend on parts of the state that the query cache does not track, e.g. the cycles balance of a canister other than the queried one.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api/routing.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/routing.rs
@@ -51,24 +51,16 @@ fn route_canister_id(
         .ok_or(ResolveDestinationError::SubnetNotFound(canister_id, method))
 }
 
-/// Inspect the method name and payload of a request to ic:00 to figure out to
-/// which subnet it should be sent to.
+/// Inspect the method name and payload of a request to the management canister
+/// to figure out to which subnet it should be sent to.
 pub(super) fn resolve_destination(
     network_topology: &NetworkTopology,
     method_name: &str,
     payload: &[u8],
     own_subnet: SubnetId,
     caller: CanisterId,
-    is_non_replicated_composite_query: bool,
     logger: &ReplicaLogger,
 ) -> Result<PrincipalId, ResolveDestinationError> {
-    // Requests to ic:00 made by a composite query executed in the non-replicated
-    // mode are not routed to any other subnet: they are executed by the query
-    // handler against the state of the own subnet (or rejected by it if the
-    // method cannot be executed in the non-replicated mode).
-    if is_non_replicated_composite_query {
-        return Ok(own_subnet.get());
-    }
     // Figure out the destination subnet based on the method and the payload.
     let method = Ic00Method::from_str(method_name);
     match method {
@@ -676,7 +668,6 @@ mod tests {
                     &reshare_chain_key_request(key_id.clone(), subnet_test_id(1)),
                     subnet_test_id(2),
                     canister_test_id(1),
-                    false,
                     &logger,
                 )
                 .unwrap(),
@@ -696,7 +687,6 @@ mod tests {
                 &setup_initial_dkg_request(None),
                 own_subnet,
                 canister_test_id(1),
-                false,
                 &logger,
             )
             .unwrap(),
@@ -716,7 +706,6 @@ mod tests {
                 &setup_initial_dkg_request(Some(requested_subnet)),
                 own_subnet,
                 canister_test_id(1),
-                false,
                 &logger,
             )
             .unwrap(),
@@ -738,7 +727,6 @@ mod tests {
                 &setup_initial_dkg_request(None),
                 own_subnet,
                 canister_test_id(1),
-                false,
                 &logger,
             )
             .unwrap(),
@@ -761,7 +749,6 @@ mod tests {
                 &setup_initial_dkg_request(Some(requested_subnet)),
                 own_subnet,
                 canister_test_id(1),
-                false,
                 &logger,
             )
             .unwrap(),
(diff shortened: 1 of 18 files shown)


rs/execution_environment/src/canister_logs.rs
```diff
@@ -15,11 +15,10 @@ pub(crate) fn fetch_canister_logs(
     round_limits: &mut RoundLimits,
 ) -> Result<CanisterManagerResponse, CanisterManagerError> {
     let canister_id = canister.canister_id();
-    let (reply, record_count, content_size) = fetch_canister_logs_response(sender, canister, args)?;
+    let (reply, instructions) = fetch_canister_logs_reply(sender, canister, args)?;
```
schneiderstefan: This is confusing. So there is now a `fetch_canister_logs_reply`, in addition to `fetch_canister_logs_response`?

rs/execution_environment/src/query_handler/query_context.rs
```diff
@@ -890,6 +914,63 @@ impl<'a> QueryContext<'a> {
...
+        // before they even reach the query handler.
+        let method = match parse_query_method(&request.method_name) {
+            Ok(method) => method,
+            Err(err) => return reject(err),
+        };
+
+        // Charge the base overhead of a query call. This happens after rejecting
+        // calls to methods that are not management canister query methods, in the
+        // same way as calls to non-existing methods of a canister are not charged.
+        self.round_limits.instructions -= self.instruction_overhead_per_query_call;
```
schneiderstefan, line 942: Any risk of underflow here, and just below?

mraszyk: The instructions are signed integers so that the subtraction can actually go negative.

rs/embedders/src/wasmtime_embedder/system_api/routing.rs
```diff
@@ -204,18 +203,8 @@ pub(super) fn resolve_destination(
         }
         Ok(Ic00Method::SubnetInfo) => Ok(SubnetInfoArgs::decode(payload)?.subnet_id),
         Ok(Ic00Method::FetchCanisterLogs) => {
-            if is_composite_query {
-                Err(ResolveDestinationError::UserError(UserError::new(
-                    ic_error_types::ErrorCode::CanisterRejectedMessage,
-                    format!(
-                        "{} API cannot be called from a composite query",
```
schneiderstefan, line 211: This error is now gone even with the flag disabled. I don't think that's a issue per se, but worth being aware of.

rs/execution_environment/src/metrics.rs
```diff
@@ -300,24 +299,26 @@ impl QueryHandlerMetrics {
...
             Err(user_error) => &format!("{:?}", user_error.code()),
         };
 
         self.subnet_query_messages
             .with_label_values(&[method_name_label.as_str(), status_label])
             .observe(duration);
 
-        if let Ok(WasmResult::Reply(bytes)) = result {
+        if let Ok(bytes) = result {
             self.subnet_query_message_response_bytes
```
schneiderstefan, line 322: This metric is now slightly different, because it also includes rejects.

schneiderstefan (commented):
Can't claim to have found that one myself, but pasting here:

> ic:00 calls from a composite callback still go through resolve_destination
> 
>   hypervisor.rs:330 sets the flag as matches!(api_type, ApiType::CompositeQuery { .. }), which excludes CompositeReplyCallback and CompositeRejectCallback. Both variants flow through the same Hypervisor::execute from query_context.rs:665, so a call made from a reply/reject callback reaches sandbox_safe_system_state.rs:474 and is routed by method+payload.

rs/execution_environment/src/metrics.rs
mraszyk: The type `Result<WasmResult, UserError>` used before was unfortunate because subnet query message always used `WasmResult::Reply(...)` or `Err(...)`. This is cleaned up here.

rs/embedders/src/wasmtime_embedder/system_api/routing.rs
mraszyk: This error was only produced for fetch_canister_logs, but not for other query methods that could not be called from composite queries. Now that we enable calling mgmt canister from composite queries, this error never surfaces.

rs/execution_environment/src/canister_logs.rs
mraszyk: Fixed here: https://github.com/dfinity/ic/pull/10987/commits/60d7cf74f180ddc3e8b3e556b32e8ccd1ca4be58

mraszyk:
> Can't claim to have found that one myself, but pasting here:
> 
> > ic:00 calls from a composite callback still go through resolve_destination
> > hypervisor.rs:330 sets the flag as matches!(api_type, ApiType::CompositeQuery { .. }), which excludes CompositeReplyCallback and CompositeRejectCallback. Both variants flow through the same Hypervisor::execute from query_context.rs:665, so a call made from a reply/reject callback reaches sandbox_safe_system_state.rs:474 and is routed by method+payload.

Good point! Fixed here: https://github.com/dfinity/ic/pull/10987/commits/536c79f314914163c622709f1a4d2ac4ca7f731f

The happy path already worked, but this unifies the error message on the failure path.

schneiderstefan (approved):

PR #10990, opened 2026-08-01
fix(ckbtc): use retrieve_btc_min_amount/2 as fee rounding increment instead of hardcoded 50_000

Rachit2323:
## Bug

The minimum ckBTC withdrawal amount is rounded to the nearest hardcoded `50_000` satoshis. This was noted as TODO DEFI-2187 — the rounding should use `retrieve_btc_min_amount / 2` instead so it scales with the configured minimum.

On mainnet today this has no visible effect since `100_000 / 2 = 50_000` matches the hardcoded value. But if the minimum is ever changed, the rounding breaks.

## Fix

Replace hardcoded `50_000` with `(self.retrieve_btc_min_amount / 2).max(1)`.

## Tests

Added 3 unit tests covering: correct rounding with a small minimum, mainnet behavior unchanged, and no division by zero when minimum is 0.


rs/bitcoin/ckbtc/minter/src/fees/tests.rs
```diff
@@ -1,5 +1,58 @@
 use super::*;
 use crate::test_fixtures::bitcoin_fee_estimator;
+use crate::tx::FeeRate;
+
+#[test]
+fn test_fee_based_minimum_withdrawal_amount_rounds_to_half_of_min_amount() {
+    // Step 1: Prepare the world.
+    // Use a small retrieve_btc_min_amount so the increment (min/2 = 5_000)
+    // differs clearly from the old hardcoded 50_000.
```
gregorydemay: nit: please remove comments. AI is usually very chatty and then the diff becomes updating comments. This type of tests should be simple enough that comments are not needed.

rs/bitcoin/ckbtc/minter/tests/tests.rs
```diff
@@ -1851,34 +1851,37 @@ fn test_min_retrieval_amount_custom() {
 
     ckbtc.refresh_fee_percentiles();
     let retrieve_btc_min_amount = ckbtc.get_minter_info().retrieve_btc_min_amount;
-    assert_eq!(retrieve_btc_min_amount, min_amount);
+    // Testnet floor fee applies even before set_fee_percentiles: 3 * 6_172 + 12_345 = 30_861
+    assert_eq!(retrieve_btc_min_amount, 30_861);
```
gregorydemay: here and below please use variables instead of hard-coding the value and then explaining the logic in comment. Something like
```rust
assert_eq!(retrieve_btc_min_amount, 3*increment+min_amount)
```

If really hard-coding the value is essential then you can also add
```rust
assert_eq!(retrieve_btc_min_amount, 30_861);
```

gregorydemay (commented):
Thanks for the PR @Rachit2323 , generally looks good to me, only some small nits.

gregorydemay (dismissed):

gregorydemay (dismissed):

PR #10991, opened 2026-08-01
chore(bazel): replace hermetic_cc_toolchain (zig cc) with hermetic-llvm

basvandijk:
Replaces Uber's `hermetic_cc_toolchain` 4.3.0 (zig cc) with [hermetic-llvm](https://github.com/hermeticbuild/hermetic-llvm) — BCR module `llvm` **0.8.16**, a prebuilt clang/lld (LLVM 22.1.8) zero-sysroot cross toolchain — and removes every `/tmp/zig-cache` reference.

## Why

`hermetic_cc_toolchain` required a zig cache (`/tmp/zig-cache`) which proved problematic and had to be patched. A newer version fixed the need for the patch but it failed because of another reason on RBE.

Additionally `zig cc` required four local patches and a pile of plumbing that all disappear with a plain clang/lld toolchain:

| Zig workaround | Why it's gone |
|---|---|
| `hermetic_cc_toolchain_strip.patch` (out-of-band host `strip -S` for determinism) + `rust.strip_level_select(none)` | clang emits deterministic debug info; rules_rust's default `opt = -Cstrip=debuginfo` keeps symbols, sheds DWARF |
| `hermetic_cc_toolchain_cache_dir.patch` (zig cache race under RBE) + all `/tmp/zig-cache` mounts/mkdirs | clang/lld has no compile cache; runtimes are ordinary sandboxed Bazel actions |
| `hermetic_cc_toolchain_lto.patch` (zig's lld fork warned on rustc's `-Wl,-O1`) | upstream lld accepts `-O1` silently |
| `cc_rs.patch` (rewrote LLVM triples into zig naming) | standard triples are correct for clang |

## Toolchain wiring

- hermetic-llvm's extension has no `extra_target_settings`, so registration goes through wrapper `toolchain()`s in `//bazel` gated on the existing `//bazel:use_hermetic_cc`. **`--hermetic_cc=false` / `--config=afl` behave exactly as before** (verified with `--toolchain_resolution_debug`: wrappers rejected → falls through to `local_config_cc`).
- **macOS: hermetic for wasm32 (canister) targets only.** The zig `none_wasm` toolchain had no exec constraints, so canister builds on macs relied on it; the `macos_*→none_wasm32` wrappers keep that working (wasm needs no Apple SDK). Native macOS targets stay on the system Xcode toolchain: hermetic-llvm's macOS support (trimmed SDK sysroot, resource-dir handling) is not yet mature enough for the full darwin build — the first CI round failed in `tikv-jemalloc-sys`'s configure-based build script (`math.h` not found).
- **glibc floor stays pinned at `gnu.2.31`** (zig parity) via `//:linux-glibc-2.31` (a `@platforms//host`-parented platform) + `build:linux --platforms/--host_platform`. The 2.28 default was tried first per plan, but `gettid` (glibc 2.30) users in `rs/ic_os` hit the documented contingency.
- **Host-library `-L` leak class (found by CI):** cargo build scripts probing host libs via pkg-config inject `-L/usr/lib/x86_64-linux-gnu` ahead of the hermetic search paths, making `-lc` resolve to the *host* glibc while crt objects come from the hermetic one (`undefined __libc_csu_init/__libc_csu_fini`). zig was immune (it resolved libc internally). Fixed per crate: pcre2-sys and libusb1-sys now build their bundled C sources (more hermetic than before); devicemapper-sys / libcryptsetup-rs-sys / libsystemd-sys link the host library by explicit path via `cc_import` host-import repos (the existing `@libsystemd` pattern) with a pkg-config wrapper (`bazel/pkg_config_no_link_metadata.sh`) that strips `-L`/`-l` from probe output while keeping version checks and bindgen include paths; virt-sys / libcryptsetup-rs are probe-only and just get the wrapper.

## Missing-unwinder link class (found by CI, round 2)

hermetic-llvm links with `--unwindlib=none` and stubs out `libgcc_s`/`libunwind` with **empty archives**; the only provider of the `_Unwind_*` symbols rust std needs is the toolchain's static runtime (`cc_toolchain.static_runtime_lib` = LLVM's libc++/libc++abi/libunwind archives). rules_rust drops that runtime in two places — zig was immune because zig cc always injected its own unwinder. Both fixed by a patch on the rules_rust fork (`bazel/rules_rust_static_runtime.patch`; TODO: upstream to dfinity/rules_rust and drop on the next fork bump):

- **Every `rust_doc_test` failed to link** (`undefined symbol: _Unwind_Resume`) — ~34 of the 50 "Bazel Test All" failures and the single arm64-linux failure. rustdoc test links pass `include_link_flags = False`, which also skipped the static runtime's `-lstatic` flags; the doc-test writer additionally didn't rewrite the runtime's `-Lnative` search paths (which live under the toolchain's configuration root) into runfiles-relative form for the test-time link.
- **`rust_shared_library` silently linked no unwinder** — the actual cause of all 17 failing system tests (`canister_http_*`, `firewall_*`, `api_bn_decentralization_*`). rules_rust links the *dynamic* C++ runtime into dylib/cdylib targets, and hermetic-llvm's dynamic runtime consists of versioned files (`libunwind.so.1`) that `-lunwind` never matches, so lld silently fell back to the empty stub archive (shared-library links don't error on undefined symbols). At runtime: dlopen of `libnss_icos.so` failed with `undefined symbol: _Unwind_GetIP` → GuestOS couldn't resolve the `hostos` hostname (provided by that NSS plugin) → `nft -f` rejected the orchestrator-generated ruleset (it whitelists `hostos` by name: `ip6 saddr { hostos }`) → the `ip6 filter` table never appeared, which is exactly what those tests wait for. Diagnosed by SSHing into a keepalive testnet node: `reload_nftables.service` in a restart loop with `Error: Could not resolve hostname`. Fix: prefer the static runtime for dylib/cdylib links whenever the toolchain provides one (local/Xcode toolchains provide none and keep their dynamic-runtime behavior).

## Fallout fixed along the way

- `rs/canister_sandbox`'s `backtrace.c` includes `<libunwind.h>` (zig bundled it): new `//bazel:libunwind_include` (via `bazel/cc_headers_dir.bzl`) exposes `@llvm-project//libunwind` headers to the cc-rs build script. LLVM libunwind is statically linked by the toolchain runtime libs — verified no new runtime `.so` dependency.
- aws-lc-sys 0.42.0's memcmp compiler probe ignores `CFLAGS` and can't compile under a zero-sysroot toolchain. Bumped **aws-lc-rs 1.17.3 / aws-lc-sys 0.43.0**, where the probe is (correctly) skipped for clang — the miscompile it probes for is GCC-specific.
- rustc 1.97's `linker_messages` lint surfaced benign clang driver chatter on every link; the lint is allowed in bazelrc (linker errors still fail the build).
- **`libusb-1.0-0-dev` dropped from the dev container:** the only consumer (vsock_lib's rusb, for HSM attach) now builds the bundled libusb in both build systems — the `vendored` feature is enabled on the rusb `crate.spec` (Bazel) and in vsock_lib's Cargo.toml (cargo). Verified: vsock_host links libusb statically with no dynamic libusb `NEEDED`; `cargo check -p vsock_lib` passes.
- **Fuzzing simplification unlocked by clang:** `DEFAULT_SANITIZERS` dropped the zig-era `-Zexternal-clangrt` + hardcoded `-Clink-arg=…librustc-stable_rt.asan.a` pair — rustc's default mechanism links the same bundled ASan runtime itself under clang/lld, without a path that baked in the compilation mode, the rules_rust canonical repo name, the toolchain channel and the arch. Verified: libfuzzer target links exactly one ASan runtime (`nm`: one defined `__asan_init`) and runs 100 iterations; AFL target builds green (unaffected by design).
- `rules_cc` 0.2.18 → 0.2.22, `rules_go` 0.61.1 → 0.62.0 (matching MVS resolution with `llvm@0.8.16` in the graph).


ci/container/Dockerfile
```diff
@@ -37,8 +37,8 @@ RUN curl -fsSL https://github.com/bazelbuild/bazelisk/releases/download/v1.28.1/
 
 # Use lld as the default linker instead of GNU bfd ld (which is ~6x slower).
 # For our large Rust debug binaries lld links roughly as fast as eg mold and
-# with comparable memory use. Bazel uses its own hermetic (zig/lld) toolchain and
-# is unaffected; this only changes the system linker used by cargo builds.
+# with comparable memory use. Bazel uses its own hermetic (clang/lld) toolchain
+# and is unaffected; this only changes the system linker used by cargo builds.
```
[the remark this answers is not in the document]
basvandijk: Good catch — reworded the comment to note that `--hermetic_cc=false` builds (e.g. `--config=afl`) also use this system linker. Done in 90692935c5.

.github/workflows/ci-kickoff.yml
```diff
@@ -61,10 +61,8 @@ jobs:
   ci-rbe-evaluation:
     name: CI RBE Evaluation
     # for dfinity/ic: run on all events except for forked PRs
-    # NOTE: currently disabled due to the following error:
-    #
-    # > No available targets are compatible with triple “x86_64-unknown-linux5.10.0-gnu2.31.0”
-    if: false
+    if: |
+      github.repository == 'dfinity/ic' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```
basvandijk: TODO: remove before merge. This was just for testing to see if this fixes the hermetic_cc_toolchain issues we had on Namespace. The good news is the build via RBE went fine but we would want to enable RBE via a separate PR.

mraszyk (approved):

mbjorkqvist (approved):
🎉

blind-oracle (approved):

bazel/conf/.bazelrc.build
```diff
@@ -40,6 +47,16 @@ build --@rules_rust//:extra_rustc_flags=-Cdebug-assertions=on
...
 
+# The hermetic toolchain passes driver flags (e.g. --unwindlib=none) that the
+# clang++ driver reports as unused on some rustc link invocations; rustc 1.97's
+# linker_messages lint then surfaces that as a warning on every link. Allow the
+# lint to keep build logs clean (linker errors still fail the build; this only
+# hides linker chatter, matching rustc's pre-1.97 behavior). A driver-level
+# -Wno-unused-command-line-argument is not an option because wasm targets link
+# through rust-lld directly, which rejects that clang driver flag.
+build --@rules_rust//:extra_rustc_flag=-Alinker_messages
+build --@rules_rust//:extra_exec_rustc_flag=-Alinker_messages
```
nmattia: this will hide all linker warning which isn't ideal, that's why we originally went with a patch: bazel/hermetic_cc_toolchain_lto.patch

could we do something similar here?

bazel/conf/.bazelrc.build
```diff
@@ -128,14 +145,3 @@ common --test_env=RUST_BACKTRACE=full
 
 # Give canceled actions some more time to cleanup
 common --local_termination_grace_seconds=90
-
-# Speed up compilation with zig cc
```
nmattia, line 132: out of curiosity, what's the performance like now?

bazel/BUILD.bazel
```diff
@@ -54,6 +55,60 @@ config_setting(
     },
 )
 
+# The hermetic toolchain's libunwind headers, exposed as a directory for the
+# cc-rs-compiled C code in //rs/canister_sandbox (which does
+# `#include <libunwind.h>`). zig cc shipped these headers in its default
```
nmattia: let's drop mentions of past solutions in the code

bazel/pkg_config_no_link_metadata.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# pkg-config wrapper that strips -L<dir> and -l<lib> tokens from the output.
+#
+# The rust `pkg_config` crate forces PKG_CONFIG_ALLOW_SYSTEM_LIBS=1, so a
```
nmattia: could you create a ticket upstream?

bazel/pkg_config_no_link_metadata.sh
```diff
@@ -0,0 +1,22 @@
...
+# __libc_csu_init/__libc_csu_fini, wrong GLIBC_ symbol versions).
+#
+# Crates whose annotation sets PKG_CONFIG to this wrapper keep the probe's
+# version checks, -I include paths (for bindgen) and -D defines, but emit no
+# link metadata; the actual library is linked by explicit path through a
+# cc_import dependency instead (see e.g. the devicemapper-sys annotation in
+# bazel/rust.MODULE.bazel and third_party/BUILD.devmapper.bazel).
+
+# `pipefail` is load-bearing: without it a pkg-config failure would be masked
+# by sed's exit 0.
```
nmattia:
```suggestion
```

bazel/rules_rust_static_runtime.patch
nmattia, line 1: can you add a short overview at the top of this patch?

bazel/rust.MODULE.bazel
```diff
@@ -6,6 +6,19 @@ bazel_dep(name = "rules_rust", version = "0.71.3")
...
+    # rules_rust drops the cc_toolchain's static C++ runtime
+    # (`static_runtime_lib`: hermetic-llvm's libc++/libc++abi/libunwind, the
+    # sole provider of the `_Unwind_*` symbols rust std needs since the
+    # toolchain links with `--unwindlib=none`) in two places this patch fixes:
+    # `rust_doc_test` links (every doc test fails to link with undefined
+    # `_Unwind_*`) and dylib/cdylib links (rules_rust links the — here empty —
+    # dynamic runtime instead, and since shared-library links don't fail on
+    # undefined symbols the breakage only surfaces at runtime: dlopen of e.g.
+    # libnss_icos.so fails, leaving GuestOS unable to resolve "hostos" and
+    # nft unable to load the orchestrator's firewall rules). TODO: upstream to
```
nmattia: can you remove the TODO and instead link to a ticket upstream?

bazel/rust.MODULE.bazel
```diff
@@ -1416,6 +1418,10 @@ crate.spec(
     version = "^0.19.0",
 )
 crate.spec(
+    # vendored: build the bundled libusb instead of linking the system one
```
nmattia, line 1413: very nice

bazel/rust.MODULE.bazel
```diff
@@ -2190,6 +2194,69 @@ crate.annotation(
     patch_args = ["-p1"],
     patches = ["@@//bazel:lmdb_rkv_sys.patch"],
 )
+crate.annotation(
+    # Build the bundled pcre2 from source instead of probing the system one
+    # via pkg-config. The probe injects the host library dir (e.g.
+    # /usr/lib/x86_64-linux-gnu) into the link line ahead of the hermetic
```
nmattia: I don't think this needs to be repeated everywhere; a short sentence pointing to the patch is enough (unless it's specific to this crate)

BUILD.bazel
```diff
@@ -361,6 +361,20 @@ platform(
     ],
 )
 
+# Host platform + a pinned glibc for the hermetic-llvm toolchain (parity with
+# the gnu.2.31 zig toolchains we used to register). Without this constraint
```
nmattia: let's drop references to old solutions

MODULE.bazel
```diff
@@ -1300,27 +1292,75 @@ os_info = use_repo_rule("//bazel:os_info.bzl", "os_info")
...
-    "@zig_sdk//toolchain:none_wasm",
+    # Linux executors: native + cross + wasm.
+    "//bazel:hermetic_cc_linux_x86_64_to_linux_x86_64",
+    "//bazel:hermetic_cc_linux_x86_64_to_linux_aarch64",
+    "//bazel:hermetic_cc_linux_x86_64_to_none_wasm32",
+    "//bazel:hermetic_cc_linux_aarch64_to_linux_aarch64",
+    "//bazel:hermetic_cc_linux_aarch64_to_linux_x86_64",
+    "//bazel:hermetic_cc_linux_aarch64_to_none_wasm32",
+
+    # macOS executors: wasm only. The zig toolchain's none_wasm had no exec
```
nmattia: let's drop references to the zig toolchain, it doesn't add anything

MODULE.bazel
```diff
@@ -1300,27 +1292,75 @@ os_info = use_repo_rule("//bazel:os_info.bzl", "os_info")
...
-
-    # wasm toolchains
-    "@zig_sdk//toolchain:none_wasm",
+    # Linux executors: native + cross + wasm.
+    "//bazel:hermetic_cc_linux_x86_64_to_linux_x86_64",
+    "//bazel:hermetic_cc_linux_x86_64_to_linux_aarch64",
+    "//bazel:hermetic_cc_linux_x86_64_to_none_wasm32",
+    "//bazel:hermetic_cc_linux_aarch64_to_linux_aarch64",
+    "//bazel:hermetic_cc_linux_aarch64_to_linux_x86_64",
+    "//bazel:hermetic_cc_linux_aarch64_to_none_wasm32",
```
nmattia: do we really need these?

nmattia (approved):

bazel/conf/.bazelrc.build
basvandijk: Building `//rs/replica:replica` *locally* took:
* on `master`: [22.6s](https://dash.dm1-idx1.dfinity.network/invocation/39134740-6115-4bd2-81f0-ce65591120e7#targets).
* on `bas/hermetic-llvm`: [14.7s](https://dash.dm1-idx1.dfinity.network/invocation/59dfea9d-c2e2-4413-9b2f-e3fa72b02ba4#targets).

However, strangely *on CI* building `//rs/replica:replica` took:
* on `master`: [73.9s](https://dash.dm1-idx1.dfinity.network/invocation/4a0635fa-42ce-4b18-9b34-2e3cdbfb067c?targetFilter=%2F%2Frs%2Freplica%3Areplica#targets).
* on `bas/hermetic-llvm`: [483.8s](https://dash.dm1-idx1.dfinity.network/invocation/4889c30f-2839-4128-8658-d5fd8d1bd68f?targetFilter=%2F%2Frs%2Freplica%3Areplica#targets)!

Not sure what is/was going on with that latter run. I'll keep an eye on the build times when I push more commits to this branch...

bazel/BUILD.bazel
basvandijk: Done.

bazel/pkg_config_no_link_metadata.sh
basvandijk: Will do...

bazel/pkg_config_no_link_metadata.sh
basvandijk: Done.

bazel/rules_rust_static_runtime.patch
basvandijk: Done.

bazel/rust.MODULE.bazel
basvandijk: Will do...

bazel/rust.MODULE.bazel
basvandijk: Done.

MODULE.bazel
basvandijk: Done.

MODULE.bazel
basvandijk: Good point. I don't think so. I removed them.

BUILD.bazel
basvandijk: Done.

bazel/conf/.bazelrc.build
basvandijk: Good idea — done in 12f84ae40c. It turned out to be exactly one flag: clang warns `argument unused during compilation: '--unwindlib=none'` on rustc links (rustc passes `-nodefaultlibs`, so the driver never consults the unwinder selection; on regular cc links the flag is used and needed, so it can't be dropped). Instead of allowing the whole `linker_messages` lint, the toolchain now wraps just that flag in clang's `--start-no-unused-arguments`/`--end-no-unused-arguments` markers via a small patch on the `llvm` module (`bazel/hermetic_llvm_unwindlib.patch`, applied with `single_version_override` — same pattern as the old `hermetic_cc_toolchain_lto.patch`). `-Alinker_messages` is removed, so every other linker message (including unused-argument warnings for any other flag) surfaces again.

Wasm links are unaffected because rules_rust doesn't pass cc-toolchain link args to rust-lld links at all. The warning is also tracked upstream (hermeticbuild/hermetic-llvm#304, surfacing through cc-crate build scripts); the patch header carries a TODO to upstream this fix and drop the patch on the next bump.

bazel/conf/.bazelrc.build
basvandijk: Good question — I measured this from a few angles, because the BuildBuddy dashboards for this PR's runs make it look much *worse* than it is (more on that below). Short version: **clean-build and per-action performance are at parity or slightly better than zig cc**, and there's no cache to warm anymore — clang/lld doesn't compile its standard library on demand, the runtimes are prebuilt archives (plus a handful of ordinary, remote-cacheable Bazel actions for the crt objects/libc++), so this whole block became unnecessary rather than replaced.

**Full clean builds (CI, "Build IC", zero cache hits):**

| | elapsed | executed actions |
|---|---|---|
| master/zig (July 27 `rules_rust` bump full rebuild) | 867.4s | 10,943 |
| this PR (first full rebuild) | 858.9s | 13,898 |

Same wall time, ~27% more actions (this PR also rebuilds all the C deps that the rules_rust bump didn't touch).

**Per-action timings** (idle 32-core devbox, single-action rebuilds via `--profile`, 2 reps each):

| action | master (zig) | this PR (hermetic-llvm) |
|---|---|---|
| `logger` rlib | 4.7 / 1.7s | 1.6 / 1.7s |
| `types` rlib | 16.6 / 16.7s | 17.2 / 17.0s |
| `replicated_state` rlib | 12.3 / 12.2s | 12.1 / 12.2s |
| `replica` bin compile+**link** | 27.1 / 17.3s | 15.0 / 14.5s |

rlib compiles are flag-identical between the two toolchains (verified with `aquery` — only `-Cstrip` differs, a no-op with `debuginfo=0`), so parity there is expected; the final link is ~15% faster through clang/lld than through the zig wrapper. Binary size is unchanged (124.8 → 125.3 MB). Two structural wins on top:

- every rustc sandbox stages **3,319 inputs instead of 7,860** — zig shipped 4,562 toolchain files into every action, hermetic-llvm ships 38;
- the `aws-lc-sys` build script — the single biggest critical-path action on cold builds — dropped from **280s (zig) to ~200s (clang)**.

**About the scary dashboard numbers:** BuildBuddy's per-target "duration" measures *when a target finished relative to the start of the build*, including all time waiting on its dep tree. Nearly every commit on this PR touches toolchain config (module patches, bazelrc, `crate_index` repins), which invalidates all Rust and/or C actions, so each CI run rebuilds ~10k actions and leaf targets like `//rs/replica:replica` show 400–620s. Master does the same when *its* dep tree is invalidated (410s on Aug 3's run); conversely the one PR run whose parent run had already warmed the cache showed `replica` at 69.9s — same as master's typical 73.9s. So expect exactly one full rebuild on master (and per developer machine) after merge, then steady state is unchanged.

bazel/pkg_config_no_link_metadata.sh
basvandijk: This has been replaced by patching the `pkg-config` crate instead.

basvandijk:
## Would `bazel/rules_rust_static_runtime.patch` still be needed with rules_rs?

[hermetic-llvm's README](https://github.com/hermeticbuild/hermetic-llvm/blob/main/README.md#usage-with-rust) recommends using [rules_rs](https://github.com/hermeticbuild/rules_rs) for Rust + hermetic-llvm interop, so I looked into whether switching to it would obsolete this PR's rules_rust patch.

**TL;DR: yes — with rules_rs the patch would no longer be required.** rules_rs isn't an independent ruleset: it's a layer on top of a pinned **hermeticbuild/rules_rust fork** (currently [`hermeticbuild/rules_rust@0105631`](https://github.com/hermeticbuild/rules_rust/commit/010563126a081e065d9e57825f855d89559c6793)), and that fork already fixes both problem classes the patch addresses — just by different means. One of the two fixes would move from "automatic" to a per-target attribute we'd need to set.

### Recap: why the patch exists

hermetic-llvm links with `--unwindlib=none` and stubs `libgcc_s`/`libunwind` as empty archives, so the *only* provider of the `_Unwind_*` symbols rust std needs is the cc_toolchain's `static_runtime_lib` (LLVM's libc++/libc++abi/libunwind archives). Upstream rules_rust 0.71.3 drops that runtime in two places:

1. **Doc tests** pass `include_link_flags = False`, skipping the static runtime's `-lstatic` flags (every `rust_doc_test` fails with ``undefined symbol: _Unwind_Resume``), and the doc-test script writer doesn't rewrite the runtime's `-Lnative` paths into runfiles-relative form.
2. **dylib/cdylib links** use the *dynamic* runtime via `-ldylib=unwind`, which never matches hermetic-llvm's versioned `libunwind.so.1` — lld silently links no unwinder and the breakage only surfaces at `dlopen` time (the `libnss_icos.so` / GuestOS failure).

### How the hermeticbuild fork fixes the same things

- **Doc tests**: their `rustdoc.bzl` passes `include_link_flags = is_test` instead of hardcoded `False`, so the static runtime's link flags flow through `construct_arguments` naturally. And `rust_doc_test` was redesigned to run rustdoc as a **build action** (a `RustdocTest` action capturing stdout/stderr/exit code) rather than a script that relinks at test runtime — which eliminates the path-stripping problem (this patch's third hunk) by construction. They even carry a regression test, [`rustdoc_musl_unwind_link_flags_test`](https://github.com/hermeticbuild/rules_rs/blob/main/test/rustdoc_musl_unwind_test.bzl), asserting the libunwind link arg is present on the doc-test action.
- **dylib/cdylib**: the fork still defaults to the dynamic runtime, but on non-Windows it links it by **full path** via `-Clink-arg=<path>` instead of `-ldylib=<name>` (their comment: `--as-needed` could drop libunwind before libstd references it). A full path matches `libunwind.so.1` fine, so the "silently missing unwinder" link-time failure mode is gone. It also adds those libs to the rpath.

### The one caveat for us

The fork's dylib fix links the runtime *dynamically* — the resulting `.so` gets a `NEEDED libunwind.so.1` entry, and the rpath points into the Bazel execroot. For shared libraries deployed outside Bazel (exactly `libnss_icos.so` on GuestOS, where hermetic-llvm's `libunwind.so.1` doesn't exist on the image), that would still break at runtime. The fork's answer is an opt-in attribute on `rust_shared_library`:

```python
cc_runtime_linkage = "static"  # embed libc++/libc++abi/libunwind → self-contained .so
```

So instead of this patch's "prefer static whenever the toolchain has one" global default, we'd set this on the handful of deployed `rust_shared_library` targets. Arguably cleaner (explicit, no fork patch), but it's a behavioral difference to remember — forgetting it on a new deployed cdylib reproduces the failure, though now at dlopen with a *missing library* (loud, obvious) rather than a missing symbol.

### Bigger picture

Dropping the patch wouldn't be the hard part of that switch. Moving to rules_rs means replacing the `archive_override` on dfinity/rules_rust in `bazel/rust.MODULE.bazel` with hermeticbuild's fork (worth auditing what else the dfinity fork carries), and — the big one — replacing crate_universe with rules_rs's `crates.from_cargo`, which consumes `Cargo.lock` directly and would obsolete the `Cargo.Bazel.*.lock` files and the `bazel-pin.sh` workflow entirely. rules_rs's `rules_rust` extension does accept extra patches on the fork, so there's an escape hatch if something else comes up.

Short of that migration, the existing TODO (upstream this patch to the dfinity/rules_rust fork and drop it on the next fork bump) remains the low-effort path.

bazel/rules_rust_static_runtime.patch
basvandijk, line 1: Note for reviewers:
See [this comment](https://github.com/dfinity/ic/pull/10991#issuecomment-5172666820) about the `rules_rust_static_runtime.patch` patch in relation to `rules_rs`.

PR #11003, opened 2026-08-03
feat: drop derivative crate

nmattia:
The crate's last update was in 2021 and is now considered unmaintained: https://rustsec.org/advisories/RUSTSEC-2024-0388.html

Instead of moving to an alternative, and because of the few sites where it was used, we write the instance by hand.


basvandijk (approved):

rs/sns/treasury_manager/src/lib.rs
```diff
@@ -161,15 +157,26 @@ pub struct Error {
...
-    write!(f, "{principal}")
+/// Renders a [`Principal`] in its textual form (e.g. `aaaaa-aa`) when the
+/// enclosing type is formatted with `Debug`.
+///
+/// The `Debug` derived by `Principal` itself prints the raw bytes, which is
+/// unreadable in logs. Wrap the field to get the textual form instead:
+///
+/// ```ignore
+/// f.debug_struct("Account")
+///     .field("owner", &PrincipalAsString(&self.owner))
```
daniel-wong-dfinity-org-twin, line 166: Would this likely

rs/sns/treasury_manager/src/lib.rs
```diff
@@ -161,15 +157,26 @@ pub struct Error {
...
-fn fmt_principal_as_string(
-    principal: &Principal,
-    f: &mut std::fmt::Formatter,
-) -> Result<(), std::fmt::Error> {
-    write!(f, "{principal}")
+/// Renders a [`Principal`] in its textual form (e.g. `aaaaa-aa`) when the
+/// enclosing type is formatted with `Debug`.
+///
+/// The `Debug` derived by `Principal` itself prints the raw bytes, which is
+/// unreadable in logs. Wrap the field to get the textual form instead:
```
daniel-wong-dfinity-org-twin:
```suggestion
/// The `Debug` implementation of this uses the textual format (just like
/// `Display`), whereas the `Debug` implementation on `Principal` is
/// hostile to humans.
```

(Reflow as necessary.)

rs/sns/treasury_manager/src/lib.rs
```diff
@@ -194,6 +200,36 @@ pub enum ErrorKind {
     Generic { generic_error_name: String },
 }
 
+// Hand-written so that `canister_id` is rendered as text. The exhaustive
+// patterns make the compiler flag any variant or field added later.
```
daniel-wong-dfinity-org-twin:
```suggestion
// Hand-written so that `canister_id` is rendered as text.
```

Exhaustiveness is separate from hand-writing, so if you want to mention it, it should be placed directly above `match`

 But I think it is safe to leave it unsaid, because I think we all know that when we see

```
match some_enum {
```

we know that will raise a large "FIX ME!" alarm bell when a new variant is added to the enum.

daniel-wong-dfinity-org-twin (approved):

PR #11010, opened 2026-08-03
feat: Add replicated state metrics for ingress history, queues, call contexts and refunds

mraszyk:
Adds Prometheus metrics for a number of replicated state quantities that were not observable so far, but are necessary to confirm that a subnet can be merged with another one by only copying over canister states and canister snapshots:

 * `replicated_state_ingress_history_length_by_state`, the number of ingress history entries by ingress state. In order to avoid scanning the whole ingress history (up to `INGRESS_HISTORY_MAX_MESSAGES` entries) on every round, `IngressHistoryState` maintains the counts incrementally, in a new transient field, just like `memory_usage` (recomputed on deserialization and on subnet splitting; validated against a full recount in debug builds).
 * `execution_subnet_input_queue_messages` and `execution_subnet_output_queue_messages` for the subnet (i.e. management canister) queues, which the existing `execution_input_queue_messages` does not cover.
 * `replicated_state_subnet_call_contexts`, the number of in-progress subnet calls by type (canister HTTP request, stop canister, etc.).
 * `replicated_state_pending_refunds`, the number of pending anonymous refunds in the subnet-level refund pool.


rs/replicated_state/src/metadata_state.rs
```diff
@@ -1781,6 +1783,64 @@ impl Default for IngressHistoryState {
...
+/// The number of ingress history entries in each `IngressState`.
+///
+/// `IngressStatus::Unknown` does not describe an entry at all: it is the stand-in
+/// for a message with no ingress history entry, so recording one is
+/// `debug_assert`ed against in `IngressHistoryState::insert()`. It is still given
+/// a count of its own rather than being ignored, as a release build backstop: this
+/// way the counts always add up to `IngressHistoryState::len()` and a non-zero
+/// `unknown` count reveals that something did record one.
+#[derive(Clone, Copy, Eq, PartialEq, Debug, Default)]
+pub struct IngressStateCounts {
```
alin-at-dfinity: Personal preference: In general, for such structs I went with names like `<Struct>Stats`, so in this case I'd have used `IngressHistoryStateStats` or, more concisely, `IngressHistoryStats`.

(It now only aggregates counts by state, but we may add other stats in the future. Although I suppose you could also move `memory_usage` into it and update it in `on_insert()` / `on_remove()`. And combine the `debug_asserts`.)

rs/execution_environment/src/scheduler/tests/metrics.rs
```diff
@@ -582,6 +585,257 @@ fn replicated_state_metrics_some_canisters_not_in_routing_table() {
...
+                u64::from(Some(*value) == one),
+            )
+        })
+        .collect();
+    assert_eq!(
+        expected,
+        fetch_int_gauge_vec(&observe_replicated_state(state), name),
+        "unexpected value of `{name}`"
+    );
+}
```
alin-at-dfinity: FWIW, there's an `ic_test_utilities_metrics::nonzero_values(MetricVec)` helper function that you could use for more expressive asserts, like [here](https://sourcegraph.com/r/github.com/dfinity/ic@e1a7d3a6e387890623a4384f0713802043a46887/-/blob/rs/messaging/src/routing/stream_builder/tests.rs?L427-444). (I would consider it enough to check once for the existence of all expected labels, e.g. by leaving out the `nonzero_values()`, no need to do it on every assert.)

Feel free to leave it as is, though.

rs/execution_environment/src/scheduler/tests/metrics.rs
```diff
@@ -582,6 +585,257 @@ fn replicated_state_metrics_some_canisters_not_in_routing_table() {
...
+            .originator(remote_canister)
+            .respondent(subnet_canister)
+            .originator_reply_callback(CallbackId::from(1))
+            .build(),
+    ));
+    assert_gauge(1, &state, OUTPUT);
+
+    // And popping the response drops it back to 0.
+    assert_eq!(1, state.output_into_iter().count());
+    assert_gauge(0, &state, OUTPUT);
```
alin-at-dfinity: I had a colleague at Google who kept drilling into me throughout repeated code reviews that tests should be as verbose as they need to in order to be directly readable without having to delve into fixtures and helper functions. E.g. in this case he would have requested me to replace this with an `assert_eq!(Some(0), fetch_int_gauge(&observe(state), OUTPUT)`.

I'm merely suggesting it. (o:

alin-at-dfinity (approved):

rs/replicated_state/src/metadata_state.rs
mraszyk: Only did the rename: https://github.com/dfinity/ic/pull/11010/commits/a0ee5d4a75614691bbdc5716684c1e7487b362b5

rs/execution_environment/src/scheduler/tests/metrics.rs
mraszyk: Sure, got rid of the helper: https://github.com/dfinity/ic/pull/11010/commits/3ba557997c95f0afde5e2dfbe253cc9fcaae1bda

rs/execution_environment/src/scheduler/tests/metrics.rs
mraszyk: Good point, thanks! I simplified both helpers: https://github.com/dfinity/ic/pull/11010/commits/4b9f83d3bd7d7b0fd9c8fe096844d0fa57150d40

PR #11012, opened 2026-08-03
feat: Cap the amount of cycles that move into the per-replica HTTP outcalls allowance

eichhorl:
... and other smaller fixes.

## Background
The new pay-as-you-go pricing for HTTP outcalls consists of three parts:
1. Base fee. This fee is charged for every request upfront.
2. Per-replica fee. The remaining cycles after charging the base fee are split evenly between the participating replicas. Each replica consumes some of their allowance as the HTTP request is processed. In the end, the amount of consumed cycles is gossiped as part of the response share.
3. Consensus cost. This fee is charged for including the aggregated HTTP response as part of a block. The cost is covered by the sum of cycles, that each replica didn't consume out of their per-replica allowance to produce its share. 

If the call runs out of cycles at any point in time, it will fail. Therefore, in order to determine the correct amount of cycles to attach to the call, the system API offers a function `http_request_fee_v2` which may be used to estimate the expected cost based on some parameters. 

## Proposed Changes
1. Until now `http_request_fee_v2` did not have a parameter to choose the "replication" of the outcall (i.e. fully-/non-/flexible-replication), which does have an effect on the cost. We introduce this missing parameter.
2. The existing `http_request_fee_v2` implementation duplicated the same calculation and constants of the newer pricing crate. We move the implementation of `http_request_fee_v2` into the pricing crate, to ensure the functions don't diverge.
3. The existing `http_request_fee_v2` was charging for gossiping the response during fully replicated calls (see the original `650` consensus cost literal, vs. `600` now). This is incorrect since responses aren't gossiped in this case, therefore, the unification with the pricing crate removes this extra charge.
4. The caller of `http_request_fee_v2` is of course free to disregard the returned estimate, and attach as many cycles to the outcall as they please (say, their entire life savings). This is dangerous because in the current model, each malicious replica can burn the entirety of their allocated allowance share. To reduce this risk, we use the same functionality called by `http_request_fee_v2` (but with parameters chosen to be as big as possible), to derive a maximum cost. When deriving the per-replica allowance, we now deduct at most the maximum possible cost from the caller's payment. The remaining balance is refunded when a response is delivered, like it is in legacy pricing.
5. Previously, the per-replica allowance was defined as `(payment - base_cost) / committee_size`. Any remainder of this division was silently lost (i.e. charged to the caller but not observed by metrics). With this PR, any remainder is refunded back to the caller instead.

## Future Work
We can use the function calculating the maximum outcall cost to tighten the bound on cycles that each replica on a free subnet may use to answer a request (currently set to 1T).


rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -3585,26 +3587,36 @@ fn execute_flexible_canister_http_request() {
...
         );
+        // Whatever the payment covers beyond the base fee and the allowances stays
+        // in the payment, to be refunded along with the response.
+        let expected_payment = match cost_schedule {
+            CanisterCyclesCostSchedule::Free => payment,
+            CanisterCyclesCostSchedule::Normal => {
+                refundable_payment - expected_allowance * committee_size.max(1)
+            }
+        };
+        assert_ne!(expected_payment, Cycles::new(0));
```
pierugo-dfinity, line 3654: This hints the reader (me, at least) that this is a property that should always hold, since we are asserting on it. Though it is only true because of the current choice of parameters, right? The assertion would fail if the refundable payment is actually divisible by the committee size

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,228 @@ pub fn base_fee(
...
+        subnet_cycles_config.cost_schedule,
+    )
+}
+
+/// The [`base_fee`], on a subnet of `subnet_size` nodes.
+fn base_fee_amount(
+    request_size: NumBytes,
+    replication_kind: ReplicationKind,
+    subnet_size: NumberOfNodes,
+) -> u128 {
```
pierugo-dfinity: What about returning `Cycles` directly?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,228 @@ pub fn base_fee(
...
+/// large transformed response.
+///
+/// The result is rounded up to a multiple of the number of participating replicas,
+/// so that splitting it evenly among them always yields allowances that add up to at
+/// least the worst case.
+pub fn max_usage_fee(
+    replication: &Replication,
+    max_response_bytes: Option<NumBytes>,
+    subnet_size: NumberOfNodes,
+) -> Cycles {
```
pierugo-dfinity, line 293: Should this return `CompoundCycles`, similar to other public functions in this module, like `base_fee` and `total_fee`? Actually, this isn't the only public function that directly return `Cycles`, so I'm wondering if we should be consistent on that aspect?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,228 @@ pub fn base_fee(
...
+            consensus_fee(transformed_bytes, subnet_size),
+        ),
+        // Every replica gossips its own response, and up to `max_responses` of them
+        // are delivered — those beyond `min_responses` at an extra fee.
+        ReplicationKind::Flexible {
+            min_responses,
+            max_responses,
+            ..
+        } => {
+            let responses = max_responses as u128;
```
pierugo-dfinity, line 264:
```suggestion
```
I think it's easier to understand when using `max_responses` directly

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,228 @@ pub fn base_fee(
...
+            (
+                gossip_usage_fee(transformed_response_size, subnet_size),
+                consensus_fee(
+                    responses.saturating_mul(
```
pierugo-dfinity:
```suggestion
                    min_responses.saturating_mul(
```
Just for my understanding. Why charge the fee for putting responses into a block (== consensus fee) for the *maximum* number of responses, and an extra fee for each one one on top of the *minimum*? I would have expected to charge for the *minimum* number of responses and then an extra fee for each additional response.

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,228 @@ pub fn base_fee(
...
+                ) + flexible_extra_response_fee(
+                    responses.saturating_sub(min_responses as u128),
+                    subnet_size,
+                ),
+            )
+        }
+    };
+
+    let per_replica_fee = network_usage_fee(raw_response_size, http_roundtrip_time)
+        .saturating_add(transform_usage_fee(transform_instructions))
```
pierugo-dfinity: Does `transform_instructions` equal 0 when there is no transform (in the case of non-replicated/flexible outcalls)?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -292,4 +478,267 @@ mod tests {
...
+    #[test]
+    fn total_fee_flexible_charges_every_delegated_replica_and_response() {
+        // 3 delegated replicas, 2 of whose responses suffice, so all 3 perform the
+        // outcall and gossip their own response, and up to 3 responses are delivered
+        // — one of them beyond `min_responses`.
+        //   base fee    = 13 * (1_000_000 + 50*100 + 90_000*13 + 2_000*13*2 + 100_000*2)
+        //                                                          = 31_551_000
+        //   per replica = 3 * (650_002 + 50*500*13)                =  2_925_006
+        //   consensus   = 9_490 * 3 * (181 + 500)                  = 19_388_070
+        //   extra resp. = 13 * (2_000*13 + 100_000)                =  1_638_000
```
pierugo-dfinity:
```suggestion
        //   extra resp. = (3 - 2) * 13 * (2_000*13 + 100_000)      =  1_638_000
```

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -292,4 +478,267 @@ mod tests {
...
+
+    #[test]
+    fn max_usage_fee_flexible_covers_all_delegated_replicas_and_responses() {
+        // A flexible outcall with a committee of 3 (2 of which suffice) and a 1 KB
+        // response limit at N = 13: all 3 replicas perform it and gossip their own
+        // maximally large response, and all 3 responses are delivered, one of them
+        // beyond `min_responses`.
+        //   per replica    = 50_000 + latency + transform + gossip
+        //   consensus      = 9_490 * 3 * (181 + 2_000_000)   = 56_945_153_070
+        //   extra response = 13 * (2_000 * 13 + 100_000)     =      1_638_000
```
pierugo-dfinity:
```suggestion
        //   extra response = (3 - 2) * 13 * (2_000 * 13 + 100_000)     =      1_638_000
```

rs/types/types/src/canister_http.rs
```diff
@@ -206,30 +215,67 @@ impl Replication {
...
+    /// The response counts for a flexible request that does not specify its own:
+    /// Every node attempts the outcall and `floor(2/3 * N) + 1` responses are required.
+    pub fn default_flexible_counts(subnet_size: NumberOfNodes) -> ReplicationCounts {
+        let n = subnet_size.get();
+        ReplicationCounts {
+            total_requests: n,
+            min_responses: (2 * n) / 3 + 1,
+            max_responses: n,
+        }
+    }
```
pierugo-dfinity, line 257: Should this belong to `impl ReplicationCounts`?

pierugo-dfinity (commented):
Would be happy if someone else had a second look on the execution layer parts, as I'm less comfortable with that part of the code 🙏 Maybe @mraszyk

rs/https_outcalls/pricing/src/fees.rs
eichhorl: This function `usage_fee` is only used ahead of time to estimate the total fee (when a canister calls it) or the maximum fee (when the replica calls it to determine the allowance). This means, we don't yet know the _actual_ number of responses that will be included (and thus determine the consensus cost). Therefore, we can only assume the worst case which is why we set `let responses = max_responses` above.

rs/execution_environment/src/execution_environment/tests.rs
eichhorl: We are asserting that the `expected_payment` is not zero, as a sanity check that the test setup is correct, and this path of the code is indeed tested.

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Yes, but note that non-replicated/flexible outcalls may still have a transform function

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: I see, thanks. Should then this maybe be reflected more precisely in the documentation? Currently, `total_fee` reads that it returns the total fee for an outcall that *consumed* the given resources

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: Thanks 👍

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Good idea, done: https://github.com/dfinity/ic/pull/11012/commits/ae03cec3aed6b71a0ec76042ad90438b37c74a50

rs/https_outcalls/pricing/src/fees.rs
eichhorl: `total_fee` and `base_fee` need to return `CompoundCycles`, because there the `real` amount is genuinely zero on free subnets (and different to the `nominal` value).

`max_usage_fee` describes the maximum nominal cost on both free and charging subnets. So arguably, maybe it should return `NominalCycles`, but those have no constructor and would make the math at the caller more difficult.

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Done 👍 https://github.com/dfinity/ic/pull/11012/commits/ae03cec3aed6b71a0ec76042ad90438b37c74a50

rs/types/types/src/canister_http.rs
eichhorl: this `impl` would have to be in `ic-management-canister-types-private`, where we don't have `NumberOfNodes`, so I think it's fine here

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: I see, thanks for the explanation!

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: Ok now I understand why you aliased it

rs/types/types/src/canister_http.rs
pierugo-dfinity: Ah ok rip

rs/cycles_account_manager/src/cycles_account_manager.rs
```diff
@@ -1294,27 +1296,39 @@ impl CyclesAccountManager {
         ic_https_outcalls_pricing::fees::base_fee(request_size, replication, subnet_cycles_config)
     }
 
+    /// Returns the most an HTTP outcall with the given `replication` and
```
mraszyk:
```suggestion
    /// Returns the maximum amount of cycles an HTTP outcall with the given `replication` and
```

rs/embedders/src/wasmtime_embedder/system_api.rs
```diff
@@ -1974,16 +1975,74 @@ impl SystemApiImpl {
...
+            }
+        }
+    }
+}
+
+const MAX_COST_HTTP_REQUEST_V2_PARAMS_SIZE: usize = 144;
+
+/// How much work decoding [`CostHttpRequestV2Params`] may spend skipping values it
+/// has no type for — i.e. the `reserved` payload of an `outcall_type` variant.
+const MAX_COST_HTTP_REQUEST_V2_SKIPPING_QUOTA: usize = 1;
```
mraszyk, line 2050: How was this value derived?

rs/execution_environment/tests/hypervisor.rs
```diff
@@ -9434,6 +9539,7 @@ fn cost_http_request_v2_fails_with_too_big_candid() {
     let transformed_response_bytes = 800_000;
     let transform_instructions = 500_000_000;
     let garbage = "Some garbage to DoS the System API by making Candid decoding more expensive"
+        .repeat(10)
```
mraszyk: why is this needed now?

rs/execution_environment/tests/hypervisor.rs
```diff
@@ -9403,18 +9429,96 @@ fn invoke_cost_http_request_v2() {
...
+        Some(CostHttpRequestOutcallType::Flexible(None)),
+        ReplicationKind::default_flexible(NumberOfNodes::from(subnet_size as u32)),
+    );
+}
+
+#[test]
+fn cost_http_request_v2_accepts_maximal_params() {
+    // The largest params a caller can send: every value at its maximum, with the
+    // `outcall_type` variant that encodes largest. This is exactly
+    // `MAX_COST_HTTP_REQUEST_V2_PARAMS_SIZE` bytes, so it has to pass the size check
```
mraszyk: we could assert that this encodes to exactly `MAX_COST_HTTP_REQUEST_V2_PARAMS_SIZE` bytes if we claim it here

rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -381,4 +378,59 @@ mod tests {
...
+                    "{replication:?}, {max_response_bytes:?}"
+                );
+                assert_eq!(
+                    tracker.subtract_gossip_usage(NumBytes::from(MAX_CANISTER_HTTP_RESPONSE_BYTES)),
+                    Ok(()),
+                    "{replication:?}, {max_response_bytes:?}"
+                );
+                // There should still be allowance left for the consensus cost.
+                assert!(
+                    tracker.spent < allowance.get(),
```
mraszyk: Shouldn't we ensure there's enough cycles for the consensus fee, too?

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -3618,6 +3630,168 @@ fn execute_flexible_canister_http_request() {
     }
 }
 
+#[test]
+fn execute_canister_http_request_refunds_truncated_allowance_remainder() {
+    // The payment beyond the base fee does not generally divide evenly into
+    // per-replica allowances, and only whole allowances are ever refunded through
+    // the refund status. The truncated remainder is therefore not taken out of
+    // the payment, so that it is refunded along with the response instead of
+    // disappearing from the cycles accounting.
```
mraszyk: I didn't quite get that (in particular, "only whole allowances are ever refunded through the refund status") - I suppose this is a regression test so maybe the comment could state that instead of explaining the current implementation.

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -3618,6 +3630,168 @@ fn execute_flexible_canister_http_request() {
...
+        .get(&CallbackId::from(0))
+        .unwrap();
+    let committee_size = match &http_request_context.replication {
+        Replication::Flexible { committee, .. } => committee.len(),
+        other => panic!("expected flexible replication, got {other:?}"),
+    };
+    let base_fee = test.http_request_base_fee(
+        http_request_context.variable_parts_size(),
+        &http_request_context.replication,
+    );
```
mraszyk, line 3791: Can't we use this in the previous test, too, instead of the complex closure?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -54,50 +75,233 @@ pub fn base_fee(
...
-        Cycles::new(n * per_replica),
-        subnet_cycles_config.cost_schedule,
-    )
+    Cycles::new(n * per_replica)
+}
+
+/// The per-replica part of the base fee of a gossiping (flexible or
+/// non-replicated) outcall of `request_bytes` bytes that requires
+/// `min_responses` responses, on a subnet of `n` nodes.
+fn gossipping_base_fee_per_replica(request_bytes: u128, n: u128, min_responses: u128) -> u128 {
```
mraszyk: I'd drop `per_replica` from the name and doc comment since it is not "per replica", but global.

eichhorl: Technically, the returned value is "per replica" in the sense that it should still be multiplied by `n`. But I agree the naming is more confusing than helpful. I renamed it and moved the multiplication with `n` into the function

rs/execution_environment/src/execution_environment/tests.rs
eichhorl: Yes, done 👍

rs/execution_environment/src/execution_environment/tests.rs
eichhorl: Good point, I tried to clarify

rs/https_outcalls/pricing/src/payg.rs
eichhorl: Good point, done! This required extracting the "max consensus fee" into a separate function.

rs/execution_environment/tests/hypervisor.rs
eichhorl: Done 👍

rs/execution_environment/tests/hypervisor.rs
eichhorl: The maximum possible size of the parameters increased by supporting the different replication types. I changed it now to derive the length of the garbage based on `MAX_COST_HTTP_REQUEST_V2_PARAMS_SIZE` instead.

rs/embedders/src/wasmtime_embedder/system_api.rs
eichhorl: Empirically, this is that lowest value that works (claude is telling me that a well-formed request will use `null` for the reserved payload). Added the explanation to the docs.

rs/cycles_account_manager/src/cycles_account_manager.rs
eichhorl: Done :+1:

rs/embedders/src/wasmtime_embedder/system_api.rs
mraszyk: What fails for the value of 0?

rs/https_outcalls/pricing/src/fees.rs
mraszyk: I see, I missed `Cycles::new(n * per_replica)`, I saw multiplications by `n` in `ReplicationKind::FullyReplicated` and concluded that it can't be per-replica...

rs/embedders/src/wasmtime_embedder/system_api.rs
eichhorl: When encoding or decoding parameters that set the reserved `fully_replicated` field to `()` or `candid::Reserved`, I get the following:
```
thread 'wasmtime_embedder::system_api::test::test_cost_http_request_v2_skipping_quota_admits_null_and_stops_amplification' (71) panicked at rs/embedders/src/wasmtime_embedder/system_api.rs:4811:9:
assertion failed: `Err(Custom(input: 4449444c036c06bba8d4840478a182a2910578989188980678edadc7940801a4d9e6b90a789595df9c0c786e026b01f0afad85097f01000100000000000000010000000000000001000000000000000100_01000000000000000100000000000000
table: type table0 = record {
  1_083_511_867 : nat64;
  1_378_386_209 : nat64;
  1_661_077_656 : nat64;
  2_190_595_821 : table1;
  2_805_574_820 : nat64;
  3_281_504_917 : nat64;
}
type table1 = opt table2
type table2 = variant { 2_427_148_272 }
wire_type: null, expect_type: null

Caused by:
    0: Fail to decode argument 0 from table0 to record {
         request_bytes : nat64;
         transformed_response_bytes : nat64;
         transform_instructions : nat64;
         outcall_type : opt variant {
           flexible : opt record {
             total_requests : nat32;
             max_responses : nat32;
             min_responses : nat32;
           };
           fully_replicated : reserved;
           non_replicated : reserved;
         };
         raw_response_bytes : nat64;
         http_roundtrip_time_ms : nat64;
       }
    1: Skipping cost exceeds the limit
```

mraszyk: Apparently, Candid charges the cost of 1 even if we do expect a field of type "reserved" and we only pass `()`, i.e., no garbage data. Let's then keep the skipping quota of 1.

mraszyk (commented):
LGTM I didn't review all the formulas in the file rs/https_outcalls/pricing/src/fees.rs so I don't approve

pierugo-dfinity (approved):

PR #11015, opened 2026-08-04
fix: `debug_assert` against recording `IngressStatus::Unknown`

mraszyk:
`IngressStatus::Unknown` stands for the absence of an ingress history entry, so it must never be recorded as the status of a message. Hence, this PR adds a `debug_assert` for preventing `IngressStatus::Unknown` to be inserted into the ingress history.

Two test fixtures did record one:

* `test_traverse_ingress_history` covered the canonical state encoding of `Unknown`, which is unreachable for the same reason. Drop the entry along with the corresponding expected traversal.

* `valid_transitions()` listed `Unknown` both as an origin and as a target status. Represent the origin as an `Option<IngressStatus>`, where `None` stands for "no ingress history entry" (which is what `Unknown` means), and drop it from the targets. `test_invalid_transitions` (testing all but `valid_transitions()`) thus now covers `Unknown` as an invalid target from every origin, including from an empty ingress history.

Note. The `IngressStatus::is_valid_state_transition()` checks done by the callers of `IngressHistoryState::insert()` only catch this if an entry already exists, as any transition away from `Unknown` (i.e. from "no entry") is allowed. This PR doesn't tighten the validation as it could crash a subnet in production (unlike `debug_assert`).


rs/replicated_state/src/metadata_state/tests.rs
```diff
@@ -1494,6 +1494,21 @@ fn test_status_done(i: u64) -> IngressStatus {
     }
 }
 
+#[test]
+#[should_panic(expected = "Attempted to record `IngressStatus::Unknown`")]
+fn ingress_history_insert_unknown_status_panics() {
```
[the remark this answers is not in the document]
mraszyk: A test failing upon an unexpected configuration is better than silently losing a test upon an unexpected configuration.

rs/state_manager/src/tree_hash.rs
```diff
@@ -397,17 +400,17 @@ mod tests {
         // BACKWARD COMPATIBILITY CODE FOR OLD CERTIFICATION VERSIONS THAT
         // NEED TO BE SUPPORTED.
         let expected_hashes = [
-            "47C3A071B293B4723FCACB17F2FD2FD75F68C010E333007ACC0EF425D92765FB",
```
schneiderstefan: The fact that all of these hashes changed is a massive red flag.

mraszyk: That's why I created a separate PR for this change. The hashes change because the ingress history to be hashed changes: it no longer contains `Unknown` values which are invalid.

mraszyk: The hashes no longer change on this PR.

schneiderstefan (approved):

PR #11021, opened 2026-08-04
refactor: Restrict which CompoundCycles specializations can be consumed directly

mraszyk:
Introduce the marker trait `CyclesUseCaseNonRefundableKind`, the counterpart of `CyclesUseCaseRefundableKind`, and implement it for every use case that is charged directly rather than prepaid.

Restrict the public `CyclesAccountManager::consume_cycles` and `CyclesAccountManager::consume_with_threshold` to that trait, so refundable use cases can no longer be consumed through the generic API and instead have to go through a dedicated method (e.g. `prepay_execution_cycles` or `consume_cycles_for_final_instructions`). This makes it an explicit choice of the code author whether a charge is prepaid or final.

Both methods now delegate to private `*_impl` counterparts that keep the unrestricted `CyclesUseCaseKind` bound.

### DIFF
--- a/rs/types/cycles/src/cycles_use_case.rs
+++ b/rs/types/cycles/src/cycles_use_case.rs
@@ -113,8 +113,8 @@ pub trait CyclesUseCaseRefundableKind: CyclesUseCaseKind {}
 /// i.e. the ones that are charged directly, without a prepayment that would later
 /// be refunded via a call to `SystemState::refund_cycles`.
 ///
-/// This trait is disjoint from `CyclesUseCaseRefundableKind`: every use case
-/// implements exactly one of the two.
+/// Intended to be disjoint from `CyclesUseCaseRefundableKind`: each built-in use case
+/// in this crate implements exactly one of the two traits.
 pub trait CyclesUseCaseNonRefundableKind: CyclesUseCaseKind {}
 
 /*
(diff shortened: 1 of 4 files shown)


rs/types/cycles/src/cycles_use_case.rs
```diff
@@ -109,6 +109,14 @@ pub trait CyclesUseCaseKind: Copy + Clone + Debug {
 /// i.e. the ones that are prepaid and expected to be refunded if not fully consumed.
 pub trait CyclesUseCaseRefundableKind: CyclesUseCaseKind {}
 
+/// Marker trait to identify which use cases of `CyclesUseCase` are non-refundable,
+/// i.e. the ones that are charged directly, without a prepayment that would later
+/// be refunded via a call to `SystemState::refund_cycles`.
+///
+/// Intended to be disjoint from `CyclesUseCaseRefundableKind`: each built-in use case
```
schneiderstefan, line 116: There are ways of enforcing exclusivity by defining a trait type that is either `Refundable` or `NonRefundable`. Not sure if it's worth doing though. In nightly rust there is apparently just a one-liner declaring that two traits cannot be implemented at the same time.

schneiderstefan (approved):

rs/types/cycles/src/cycles_use_case.rs
mraszyk: Let's leave it for a follow-up.

PR #11027, opened 2026-08-04
feat(ic-admin): Added upgrade options.

daniel-wong-dfinity-org-twin:

[prev]: https://github.com/dfinity/ic/pull/10979
[next]: https://github.com/dfinity/ic/pull/11030

### DIFF
--- a/rs/registry/admin/bin/main.rs
+++ b/rs/registry/admin/bin/main.rs
@@ -1570,11 +1570,7 @@ fn assemble_canister_upgrade_options(
         return None;
     }
 
-    let skip_pre_upgrade = if skip_pre_upgrade {
-        Some(true)
-    } else {
-        None
-    };
+    let skip_pre_upgrade = if skip_pre_upgrade { Some(true) } else { None };
 
     let wasm_memory_persistence = wasm_memory_persistence.map(|wasm_memory_persistence| {
         WasmMemoryPersistenceProto::from(&wasm_memory_persistence) as i32
--- a/rs/registry/admin/bin/main.rs
+++ b/rs/registry/admin/bin/main.rs
@@ -1556,7 +1556,7 @@ impl ProposalAction for ProposeToChangeNnsCanisterCmd {
 }
 
 /// Constructs a CanisterUpgradeOptions from its constituents.
-/// 
+///
 /// (Presumably, the pieces were passed via CLI flags.)
 ///
 /// Returns None if neither flag was passed.
--- a/rs/registry/admin/bin/main.rs
+++ b/rs/registry/admin/bin/main.rs
@@ -1455,7 +1455,7 @@ struct ProposeToChangeNnsCanisterCmd {
 
     #[clap(long)]
     /// Whether to skip the canister's pre_upgrade hook. Only valid when mode is upgrade.
-    skip_pre_upgrade: Option<bool>,
+    skip_pre_upgrade: bool,
 
     #[clap(long)]
     /// Whether to retain (keep) or drop (replace) the canister's main memory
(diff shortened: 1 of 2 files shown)


yhabib (approved):

rs/registry/admin/bin/main.rs
```diff
@@ -1520,19 +1539,49 @@ impl ProposalAction for ProposeToChangeNnsCanisterCmd {
...
+            canister_upgrade_options,
         };
 
         ProposalActionRequest::InstallCode(install_code)
     }
 }
 
+/// Constructs a CanisterUpgradeOptions from its constituents.
+///
+/// (Presumably, the pieces were passed via CLI flags.)
```
mraszyk: What does this comment mean to say?

rs/registry/admin/bin/main.rs
```diff
@@ -1520,19 +1539,49 @@ impl ProposalAction for ProposeToChangeNnsCanisterCmd {
...
+///
+/// Returns None if skip_pre_upgrade is false and wasm_memory_persistence is
+/// None, i.e. there is nothing to say.
+fn assemble_canister_upgrade_options(
+    skip_pre_upgrade: bool,
```
mraszyk, line 1567: Why is this not `Option<bool>` analogously to the second argument?

rs/registry/admin/bin/main.rs
```diff
@@ -1520,19 +1539,49 @@ impl ProposalAction for ProposeToChangeNnsCanisterCmd {
...
+    wasm_memory_persistence: Option<WasmMemoryPersistence>,
+) -> Option<GovernanceCanisterUpgradeOptions> {
+    let has_option = skip_pre_upgrade || wasm_memory_persistence.is_some();
+    if !has_option {
+        return None;
+    }
+
+    let skip_pre_upgrade = if skip_pre_upgrade { Some(true) } else { None };
```
mraszyk: If we go with `skip_pre_upgrade: bool`, then why not `Some(skip_pre_upgrade)`?

rs/registry/admin/bin/main.rs
```diff
@@ -1520,19 +1539,49 @@ impl ProposalAction for ProposeToChangeNnsCanisterCmd {
...
+/// Returns None if skip_pre_upgrade is false and wasm_memory_persistence is
+/// None, i.e. there is nothing to say.
+fn assemble_canister_upgrade_options(
+    skip_pre_upgrade: bool,
+    wasm_memory_persistence: Option<WasmMemoryPersistence>,
+) -> Option<GovernanceCanisterUpgradeOptions> {
+    let has_option = skip_pre_upgrade || wasm_memory_persistence.is_some();
+    if !has_option {
+        return None;
+    }
```
mraszyk, line 1575: Why ever return `None`?

mraszyk (approved):
Approving `rs/types/management_canister_types/src/lib.rs` owned by the core protocol team, leaving some comments on the remaining code not owned by our team.

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: The flag is raw bool. That way, you do not have to say `--skip-pre-upgrade true`. Instead, you just say `--skip-pre-upgrade`.

Originally, the flag was `Option<bool>`, but I thought `--skip-pre-upgrade true` looks stupid. I also I figured that nobody would ever care to do `--skip-pre-upgrade false`, since `Some(false)` would end up doing the same thing as `None`.

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: that would also be ok, but I like this better because when false is passed, that means that there was no `--skip_pre_upgrade` in the command, and `None` better reflects such omission. `Some(false)` gives the false impression that an affirmative explicit "no" was given.

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: Update: `None` is required when mode is not upgrade.

<strike>
Similar reasoning vs. why false -> None for skip_pre_upgrade: it better reflects the LACK of flags. It would also be fine to always construct a CanisterUpgradeOptions.
</strike>

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: Updated. I _**think**_ it's less confusing now.

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: Added comments.

rs/registry/admin/bin/main.rs
daniel-wong-dfinity-org-twin: Added comment.

PR #11029, opened 2026-08-04
feat: Make snapshots immutable after loading

schneiderstefan:
This is done for consistency with snapshots that have been taken directly on-chain, which are also immutable, and separates the (comparably short) upload phase from the phase where the snapshot is usable.

### DIFF
--- a/rs/execution_environment/tests/canister_snapshots.rs
+++ b/rs/execution_environment/tests/canister_snapshots.rs
@@ -548,6 +548,80 @@ fn upload_and_load_snapshot_with_wasm_memory() {
     ));
 }
 
+// Once a metadata-upload snapshot has been loaded onto a canister it becomes
+// immutable and rejects further `upload_canister_snapshot_data`, no matter
+// whether it was loaded onto the canister owning it or onto another one.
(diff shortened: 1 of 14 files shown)


mraszyk (approved):

rs/execution_environment/tests/canister_snapshots.rs
```diff
@@ -548,6 +548,80 @@ fn upload_and_load_snapshot_with_wasm_memory() {
...
+    let target_canister_id = if load_onto_other_canister {
+        // Loading onto another canister copies the snapshot's memories, which
+        // requires them to have been flushed to disk first.
+        env.checkpointed_tick();
+        env.create_canister(None)
+    } else {
+        canister_id
+    };
+    let load_args = LoadCanisterSnapshotArgs::new(target_canister_id, snapshot_id, None);
+    env.load_canister_snapshot(load_args).unwrap();
```
mraszyk, line 600: We could also test that the snapshot can be loaded multiple times, e.g., to the same and a different canister, i.e., that the restored flag does prevent a subsequent restore.

schneiderstefan: Added

PR #11035, opened 2026-08-05
fix(ic-os): serialize podman's SHM lock creation to stop flaky exit-125 builds

basvandijk:
## Problem

ic-os container builds intermittently fail with, e.g. [this `Bazel Test All on RBE` run](https://github.com/dfinity/ic/actions/runs/30928640846/job/92057749525):

```
ERROR: ic-os/setupos/envs/prod/BUILD.bazel:11:25: Action ic-os/setupos/envs/prod/rootfs-tree.tar failed: (Exit 1)
  Command: 'podman --root /tmp/containers/icosbuildxUSY/root --runroot ... build -t e182b64e ...'
  Exit code: 125
  Error: failed to get new shm lock manager: failed to create 2048 locks in /libpod_rootless_lock_1001: file exists
```

Podman keeps its container lock manager in a shared memory segment — one per uid when rootless (`/dev/shm/libpod_rootless_lock_<uid>`), a single unsuffixed one as root (`/dev/shm/libpod_lock`). It **opens** that segment if it exists and only **creates** it `O_EXCL` if the open returns `ENOENT`, with nothing serializing the gap between those two steps. ic-os actions run concurrently and unsandboxed (`--strategy_regexp=ic-os[:/].*=local`), so on a machine where the segment does not exist yet — a fresh CI container — the first burst of actions all see `ENOENT` and race to create it. The losers exit 125.

This is why it only bites early in a job, and why it is rare (~1 failure in ~76 podman invocations in the run above; adjacent runs were clean).

## Fix

Materialize the segment exactly once under a lock of our own, before the build starts, double-checked so the lock is only taken when the segment is actually missing. Once it exists podman only ever opens it, so every later action pays a single `test -e` — no lock, no extra podman, and the builds themselves are never serialized.

`ic-os/bootloader/build-bootloader-tree.sh` drives podman outside the wrapper and can be in the same first burst, so its genrule now runs under `proc_wrapper` too — same pre-warm, one source of truth.

Root is deliberately **not** exempt: rootful podman keeps the same kind of segment (verified root-owned, 82488 bytes, byte-identical in size to the rootless one — `/run/libpod` is its *tmp dir*, not its lock), and `ci/container/container-run.sh:154-158` runs the build as root whenever the host uid has no matching container user.

## Verification

Measured with 5 rounds of 16 concurrent invocations from a cold segment, counting actions that reach podman while the segment is still absent — the race-exposed state:

| | before | after |
|---|---|---|
| rootless | **80/80** | **0/80** |
| root | **80/80** | **0/80** |

Note the guarantee is structural (creation can only happen inside the lock) rather than statistical: the raw `EEXIST` is too timing-sensitive to demonstrate on demand — it fired once, then not again in hundreds of attempts — so the invariant is what's being tested, not failure counts.

Also verified:

- `bazel build //ic-os/guestos/envs/dev:rootfs-tree.tar` from a cold segment — exit 0, action ran locally, pre-warm confirmed to have executed
- Three `rootfs-tree.tar` targets concurrently — exit 0, 4600 actions
- Fast path with the segment present — branch skipped, no lock file, no extra podman
- `shfmt`, `shellcheck -S warning`, `bash -n` clean; no affected test targets exist

## Reviewer notes / limitations

- The bootloader side is now handled by running `build-bootloader-tree.sh` under `proc_wrapper` (per @frankdavid's review comment) rather than duplicating the pre-warm, matching the pattern `ic-os/defs.bzl` already uses for `verity_sign`. Verified by forcing local execution of `//ic-os/bootloader:build-bootloader-tree` from a cold `/dev/shm`: the action runs the wrapper, builds successfully, and creates both the segment and the pre-warm lock file.
- Behavior verified against **podman 5.7.0**; the dev container installs podman unpinned (`ci/container/files/packages.common:38`), so CI's version may differ. The open-then-create behavior is the load-bearing assumption.
- Not addressed here, found while investigating: `podman create` in `build_container_filesystem_tar.py` never removes its container, so its libpod lock is never returned (~16 of 2048 per full ic-os build — pre-existing, irrelevant for CI's fresh containers, a slow drift on long-lived devboxes); and the cleanup `podman unshare` fails ~76x per job and silently falls back to `sudo rm -rf`, cause still unknown because `2>/dev/null` hides it.


basvandijk:
Both review comments addressed in 9c5494313b — thanks, the point was valid.

On the stated trigger (`flock` unavailable): that one can't really happen. `flock` ships in `util-linux`, which is `Priority: required, Essential: yes` on Ubuntu, so it is present in the `ghcr.io/dfinity/library/ubuntu` base the `ic-build` image is built from, and in the bare `ubuntu:26.04` that `repro-check.yml` uses. No container change needed.

But the underlying mechanism is real via a different trigger: `/dev/shm` is world-writable, so another uid can squat our fixed lock path. Reproduced by creating `/dev/shm/icos-podman-lock-init.$EUID.lck` as root mode 600 and running the wrapper:

```
./proc_wrapper.sh: line 49: /dev/shm/icos-podman-lock-init.1000.lck: Permission denied
WRAPPER EXIT=1
```

So the block was not best-effort as its comment claimed — a failed `9>` redirection (or a failed `flock`) aborted the action under `set -e`. Fixed by wrapping the subshell in `|| true`, so any failure degrades to building unprotected, exactly as before this series. Verified all three paths now exit 0: squatted lock file, `flock` failing (127 shim), and the normal pre-warm — with the segment still created in the normal case.

I did not add a `flock -w` timeout: the lock is only ever held across one `podman info`, and if that hung, every action would hang in its own podman anyway, so bounding it buys nothing.

ic-os/bootloader/build-bootloader-tree.sh
frankdavid, line 1: Did you consider just running this under the proc_wrapper?

basvandijk: I was assuming it was left out because of some reason but let me give that a shot...

basvandijk:
> Did you consider just running this under the proc_wrapper?

Good call — done in dca33059ed. It's strictly better than what I had: the pre-warm now lives in one place, and it follows the pattern `ic-os/defs.bzl` already uses for `verity_sign` (genrule `cmd` prefixed with `$(location //toolchains/sysimage:proc_wrapper)`, wrapper added to `tools`). Net effect is −19 lines of duplicated shell for +10 lines of BUILD. A bonus: the script no longer depends on `/tmp/containers` having been created by somebody else, since the wrapper does that.

Verified by forcing local execution from a cold `/dev/shm`:

```
$ rm -f /dev/shm/libpod_rootless_lock_$(id -u) /dev/shm/icos-podman-lock-init.$(id -u).lck
$ bazel build //ic-os/bootloader:build-bootloader-tree --noremote_accept_cached
WARNING: could not unshare podman runroot, forcing     <- proc_wrapper's cleanup, so the wrapper ran
INFO: 4 processes: 3 internal, 1 local.                <- really executed, not a cache hit
INFO: Build completed successfully, 4 total actions
segment:  YES                                          <- created by the pre-warm
lockfile: YES
```

Also `bazel build //... --nobuild` (3260 targets) and `//:buildifier` are clean.

**Correction to my earlier reviewer note**, which I've also fixed in the PR description: I previously wrote that this genrule cannot run on my dev box (`insufficient UIDs or GIDs ... /etc/gshadow`) and that I had confirmed it pre-existing by stashing the change. That was wrong on both counts — the stashed run was a *cache hit* (2 actions, nothing executed), so it proved nothing, and the target in fact builds fine here. The earlier failure was transient, most likely contention from other podman processes I had running against the same `/dev/shm` at the time. So the bootloader path is now genuinely verified at runtime rather than assumed unverifiable.

One thing I deliberately did not do: the script still makes its own `TMP_DIR` under `/tmp/containers` and cleans it with `sudo rm -rf`, even though the wrapper now exports `TMPDIR` and `PODMAN_STORAGE_DIR` and cleans up after itself. Switching the script over to `$PODMAN_STORAGE_DIR` would delete that duplication too — happy to fold it in if you'd like, but it's a behaviour change on a path this PR otherwise leaves alone.

ic-os/bootloader/build-bootloader-tree.sh
basvandijk: Done in https://github.com/dfinity/ic/commit/dca33059edb4a2b157d7f3e2cf0f8d338cad0688.

frankdavid (approved):

basvandijk:
Closing in favour of https://github.com/dfinity/ic/pull/11041.

PR #11038, opened 2026-08-05
chore: Remove impl_hash and REPLICA_BINARY_HASH

frankdavid:
These fields were never populated, nor read. `impl_hash` was marked with `skip_serializing_if = "Option::is_none"` so it didn't even show up in the serialized output.


pierugo-dfinity:
> These fields were never populated

I think it is during [replica startup](https://sourcegraph.com/r/github.com/dfinity/ic@956b33ee836fd9ff974df9432bc7f0abf332474d/-/blob/rs/replica/bin/replica/main.rs?L262-265)

PR #11039, opened 2026-08-05
fix: check the cycles balance and charge executed instructions on DTS resume

mraszyk:
`InstallCodeHelper` records the cycles balance of the clean canister state it was built from and compares it when resuming a paused `install_code` execution, analogously to `ResponseHelper` and `CallOrTaskHelper`.

Moreover, if resuming a paused execution fails, then the instructions already executed by the paused Wasm execution are charged now: those instructions consumed round instructions, but the instruction limits are only updated when the Wasm execution finishes and hence such instructions used to be free. For `install_code`, only the instructions of the steps preceding the paused Wasm execution (e.g., compilation) used to be charged; for update calls, replicated queries, canister tasks, and response callbacks, no instructions used to be charged at all; for cleanup callbacks, the instructions of the preceding response callback used to be charged, but not those of the cleanup callback itself.

Finally, `dts_replicated_execution_resume_fails_due_to_cycles_change` moves to the general execution tests as `dts_resume_fails_due_to_cycles_decrease`, decreases the cycles balance using `remove_cycles`, and covers response and cleanup callbacks and the heartbeat in addition to update calls and replicated queries; the new test `dts_install_code_resume_fails_due_to_cycles_decrease` covers `install_code`. The tests assert that the canister is charged exactly the cost of all the executed instructions.


rs/test_utilities/execution_environment/src/lib.rs
```diff
@@ -292,6 +292,11 @@ pub struct ExecutionTest {
     subnet_available_callbacks: i64,
     // The number of instructions executed so far per canister.
     executed_instructions: HashMap<CanisterId, NumInstructions>,
+    // The number of round instructions consumed so far by executing slices of
+    // canister messages and tasks per canister. Unlike `executed_instructions`,
+    // this is accumulated for every executed slice and thus also accounts for
+    // slices of executions that were paused and did not finish (yet).
+    round_instructions: HashMap<CanisterId, NumInstructions>,
```
alin-at-dfinity: Do we still need the separate `executed_instructions` anymore? Or could we replace them with the new definition?

alin-at-dfinity: "Round instructions" does not immediately convey the fact that this also tracks instructions consumed by slices of not yet completed message executions. I don't have a really good alternative name suggestion (which is why it would be nice if we could replace `executed_instructions` with this), but maybe something like "rolling executed instructions" or "slice executed instructions" / "instructions executed by slices"?

rs/test_utilities/execution_environment/src/lib.rs
```diff
@@ -292,6 +292,11 @@ pub struct ExecutionTest {
     subnet_available_callbacks: i64,
     // The number of instructions executed so far per canister.
```
alin-at-dfinity:
```suggestion
    // The number of instructions executed so far by completed message executions,
    // per canister.
```

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -6002,3 +6006,246 @@ fn list_canisters_via_ingress_fails_at_execution() {
...
+/// Decreases the cycles balance of the given canister, which must have a paused
+/// execution, and resumes that execution, which must fail.
+///
+/// Asserts that the failed execution is charged for exactly the instructions it
+/// had already executed: the instructions it reports as used match the round
+/// instructions consumed by its slices, and the cycles charged to the canister
+/// match the cost of exactly those instructions.
+///
+/// `before` must be taken right before the first slice of the paused execution.
+fn resume_paused_execution_after_cycles_decrease(
```
alin-at-dfinity: Seems to me like every invocation of `resume_paused_execution_after_cycles_decrease()` looks like

```rust
        let before = ExecutionAccounting::take(&test, a_id);
        test.execute_slice(a_id);
        resume_paused_execution_after_cycles_decrease(&mut test, a_id, before);
```

Meaning that you could have this helper funciton call `ExecutionAccounting::take()` and `execute_slice()` itself and simplify the API.

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -6002,3 +6006,246 @@ fn list_canisters_via_ingress_fails_at_execution() {
...
+        .execution_cost(
+            executed_instructions,
+            test.get_own_subnet_cycles_config(),
+            WasmExecutionMode::Wasm32,
+        )
+        .nominal();
+    assert_eq!(
+        (after.execution_cost - before.execution_cost).nominal(),
+        expected_cost
+    );
```
alin-at-dfinity: As long as you have (and check above) `executed_instructions`, this assert is redundant: `execution_cost` is updated at the same time as `executed_instructions`, with the cost of the instructions added to `executed_instructions`.

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -6002,3 +6006,246 @@ fn list_canisters_via_ingress_fails_at_execution() {
...
+
+        let canister_id = test.universal_canister().unwrap();
+        let (ingress_id, _) = test.ingress_raw(
+            canister_id,
+            "update",
+            wasm()
+                .set_heartbeat(
+                    wasm()
+                        .instruction_counter_is_at_least(1_000_000)
+                        .stable_grow(1)
```
alin-at-dfinity: In order to prove that the `stable_grow()` is not applied, shouldn't we execute it first? Or does `instruction_counter_is_at_least()` simply ensure that once everything else is executed, we spin until we reach that many instructions? (Even then, it may be clearer to put `stable_grow()` first.)

rs/execution_environment/src/execution/response.rs
```diff
@@ -126,6 +127,15 @@ struct ResponseHelper {
...
     response_sender: CanisterId,
+    // Instructions already executed by a Wasm execution that has been paused and
+    // has not finished yet. The instructions charged for this message are derived
+    // from `output.num_instructions_left` of a *finished* Wasm execution, while a
+    // paused slice only reports its executed instructions to the round limits.
+    // Hence such instructions are tracked here in order to be charged if the
+    // execution fails before the Wasm execution finishes. The counter is reset by
+    // `reset_executed_wasm_instructions` once a Wasm execution has finished and is
+    // reflected in the instruction limits of the next one (the cleanup callback).
+    executed_wasm_instructions: NumInstructions,
```
alin-at-dfinity: This comment is really hard to parse.

```suggestion
    /// Instructions already executed by this callback, if previously paused.
    ///
    /// *Finished* message executions are charged based on
    /// `output.num_instructions_left`. Paused executions only update the round
    /// limits. If resuming a paused execution fails, `output.num_instructions_left`
    /// is not updated. Hence, we track executed instructions, to make it possible
    /// to charge for failed resumptions.
    ///
    /// Reset by `reset_executed_wasm_instructions` once a Wasm execution has
    /// finished and updated the instruction limits correspondingly.
```

rs/execution_environment/src/execution/response.rs
```diff
@@ -295,7 +306,13 @@ impl ResponseHelper {
 
     /// Returns a struct with all the necessary information to replay the
     /// initial steps in subsequent rounds.
-    fn pause(self, round_limits: &mut RoundLimits) -> PausedResponseHelper {
+    /// The given `executed_wasm_instructions` are the instructions executed by
+    /// the Wasm slice that is being paused.
```
alin-at-dfinity: Then maybe name it something like `slice_executed_instructions`?

rs/execution_environment/src/execution/response.rs
```diff
@@ -1104,6 +1151,9 @@ fn execute_response_cleanup(
     execution_parameters
         .instruction_limits
         .update(instructions_left);
+    // The instructions executed by the response callback are reflected in the
+    // instruction limits updated above.
+    helper.reset_executed_wasm_instructions();
```
alin-at-dfinity: I'm not convinced that hiding `self.executed_wasm_instructions = 0` behind a function call (with a verbose doc comment) helps here. I'd just do it inline.

alin-at-dfinity (commented):
This is taking me quite a bit longer than expected, so here's a first batch of (not hugely consequential) comments.

rs/execution_environment/src/execution/response.rs
```diff
@@ -861,6 +899,11 @@ impl PausedExecution for PausedCleanupExecution {
         //
         // Note that we don't apply changes from the response callback execution
         // because the cleanup callback runs only if the response callback fails.
+        // If resuming fails, then the instructions already executed by the paused
+        // Wasm execution are still charged: they have been executed and hence
+        // consumed round instructions, but the paused Wasm execution never finishes
+        // and hence yields no `num_instructions_left` to derive them from.
+        let executed_wasm_instructions = self.helper.executed_wasm_instructions;
```
alin-at-dfinity, line 893: Nit: I'd move this comment below, where we actually compute `instructions_left` based on this.

rs/execution_environment/src/execution/install_code.rs
```diff
@@ -124,12 +127,23 @@ pub(crate) struct InstallCodeHelper {
...
+    // The cycles balance of the clean canister state this helper was built from.
+    initial_cycles_balance: Cycles,
+    // Instructions already executed by a Wasm execution that has been paused and
+    // has not finished yet. Such instructions are not reflected in
+    // `execution_parameters` because the instruction limits are updated only
+    // when the Wasm execution finishes, so they are tracked here in order to be
+    // charged if the execution fails before the Wasm execution finishes. The
+    // counter is reset by `reset_executed_wasm_instructions` once a Wasm execution
+    // has finished and is reflected in the instruction limits of the next one.
+    executed_wasm_instructions: NumInstructions,
```
alin-at-dfinity: Same comments here (and everywhere below in `install_code.rs`) as in `response.rs`.

rs/execution_environment/src/execution/call_or_task.rs
```diff
@@ -330,6 +331,13 @@ struct CallOrTaskHelper {
     canister: CanisterState,
     call_context_id: CallContextId,
     initial_cycles_balance: Cycles,
+    // Instructions already executed by a Wasm execution that has been paused and
+    // has not finished yet. The instructions charged for this message are derived
+    // from `output.num_instructions_left` of a *finished* Wasm execution (see
+    // `finish`), while a paused slice only reports its executed instructions to
+    // the round limits. Hence such instructions are tracked here in order to be
+    // charged if the execution fails before the Wasm execution finishes.
+    executed_wasm_instructions: NumInstructions,
```
alin-at-dfinity: And here (regarding applying the comments in `response.rs`).

alin-at-dfinity (approved):
Turns out I didn't have all that much left to say anyway, except "the same comments as in `result.rs`, where applicable, if you agree with them".

LGTM

rs/test_utilities/execution_environment/src/lib.rs
mraszyk: I unified the two definitions now.

mraszyk: I unified this with the original `executed_instructions` so I didn't introduce a new name.

PR #11040, opened 2026-08-05
feat: Introduce `OutOfCycles` error for flexible outcalls

eichhorl:
A flexible outcall whose committee has spent too much of its per-replica allowances to cover the consensus cost of delivering any response used to sit pending until it timed out. It is now reported to the caller as `FlexibleHttpGlobalError::OutOfCycles` as soon as that is provable.

The proof is the signed receipts seen so far: what is left of the committee's collective allowance is at most their unspent allowances plus a full allowance for every member not seen yet, and delivering a response costs at least `min_flexible_consensus_cost` of the same receipts.

For an `ok` response, at least `min_responses` ok shares need to be included, whereas for a reject response at least `total_requests - min_responses + 1` reject shares must be included. This determines how many "unseen" shares we need to consider in order to build the out of cycles proof.


rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -668,7 +735,25 @@ pub(crate) fn find_flexible_result(
...
+        committee.len(),
+        min_responses as u32,
+        callback_id,
+        context,
+    ) {
+        let error = FlexibleCanisterHttpError::OutOfCycles {
+            callback_id,
+            all_seen_shares: all_shares.into_iter().cloned().collect(),
+            min_cost: proof.min_cost,
+            unspent_allowance: proof.unspent_allowance,
```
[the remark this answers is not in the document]
eichhorl: Flexible shares always come with a response

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -668,7 +735,25 @@ pub(crate) fn find_flexible_result(
...
 
-    // 4. Not enough data yet
+    // 4. Can delivering a response ever be paid for?
+    if let Ok(proof) = check_out_of_cycles(
+        all_shares.iter().copied(),
+        committee.len(),
+        min_responses as u32,
+        callback_id,
+        context,
+    ) {
```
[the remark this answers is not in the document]
eichhorl: I think this is fine for simplicity

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -171,6 +173,68 @@ pub(crate) fn check_initial_spent_within_limit(
     }
 }
 
+pub(crate) struct OutOfCyclesProof {
```
pierugo-dfinity, line 178: Maybe add a comment that explains that the "proof" is that `unspent_allowance` is smaller than `min_cost`?

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -171,6 +173,68 @@ pub(crate) fn check_initial_spent_within_limit(
...
+        });
+    };
+
+    let mut unspent_allowance = Cycles::zero();
+    let mut seen = 0;
+    for share in seen_shares {
+        unspent_allowance += allowance - share.content.spent();
+        seen += 1;
+    }
+    unspent_allowance += allowance * committee_size.saturating_sub(seen);
```
pierugo-dfinity: This could be simplified with something like
```rust
let unspent_allowance = allowance * committee_size - seen_shares.iter().map(|share| share.content.spent()).sum();
```

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -169,6 +169,78 @@ fn extra_response_fee(subnet_size: NumberOfNodes, extra_responses: u128) -> Cycl
...
+///
+/// The cheapest response delivers as few bodies as it may: every committee member
+/// that has not been seen yet may still produce an empty body, so only the bodies
+/// that must come out of `seen_shares` are counted at their actual size.
+pub fn min_flexible_consensus_cost<'a>(
+    seen_shares: impl Iterator<Item = &'a CanisterHttpResponseShare>,
+    subnet_size: NumberOfNodes,
+    committee_size: usize,
+    min_responses: u32,
+) -> Option<Cycles> {
```
pierugo-dfinity, line 183: Can this function ever return `None`?

rs/types/types/src/batch/canister_http.rs
```diff
@@ -62,29 +62,43 @@ pub enum FlexibleCanisterHttpError {
         /// The total amount of cycles spent by the subnet to produce this response.
         initial_spent: Cycles,
     },
+    OutOfCycles {
```
pierugo-dfinity, line 65: Who is paying for the block space used by this new error message if the canister has not attached enough cycles?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -352,6 +431,127 @@ mod tests {
...
+        assert_eq!(
+            min_flexible_consensus_cost(seen.iter(), n, 4, 2),
+            Some(Cycles::new(6_102_070 + 1_638_000))
+        );
+    }
+
+    #[test]
+    fn min_flexible_consensus_cost_is_none_when_no_result_can_be_delivered() {
+        // A committee of 4 that has all responded successfully, but needs 10
+        // responses: neither the 10 successes nor a reject can still turn up, so
```
pierugo-dfinity, line 514: Is this scenario realistic?

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -929,6 +930,44 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                    // The shares must prove that the committee's remaining
+                    // allowances can no longer pay for a response...
+                    let expected = utils::check_out_of_cycles(
+                        all_seen_shares.iter(),
+                        flex_committee.len(),
+                        min_responses as u32,
+                        callback_id,
+                        context,
+                    )
+                    .map_err(CanisterHttpPayloadValidationError::InvalidArtifact)?;
```
pierugo-dfinity, line 948: Maybe could we add some `debug_assert`s that in the `Err` case, `.min_cost <= .unspent_allowance` and in the `Ok` case, `.min_cost > .unspent_allowance`?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -352,6 +431,127 @@ mod tests {
...
+        // which have to come from those seen, and pays the fee for the one response
+        // it delivers beyond `min_responses`:
+        //   size_term = (181 + 50) * 2 + 181       = 643
+        //   cost      = 9_490 * 643                = 6_102_070
+        //   fee       = 13 * (2_000 * 13 + 100_000) = 1_638_000
+        let n = NumberOfNodes::from(13);
+        let seen = [
+            reject_share(0, 50, 0),
+            reject_share(1, 50, 0),
+            reject_share(2, 50, 0),
```
pierugo-dfinity: Could be interesting to have different content sizes to make sure we select the 2 smallest responses

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Done 👍

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Good idea, done 👍

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Not the way it is currently called by the code, but syntactically, this case exists. The alternative would be to unwrap or return a default, but then we would have to argue why that is fine.

rs/types/types/src/batch/canister_http.rs
eichhorl: The number of signatures delivered for an outcall should be equal to the number of `total_requests` for all types of responses (once asynchronous refunds are implemented). Since this is constant, it should be covered by the base fee.

rs/https_outcalls/pricing/src/fees.rs
eichhorl: No, this test case is the trade off for not having to explain why the `None` case should not exist in our code

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Added for the error case, I don't think it adds much for the ok case

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Good point, done

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: Okok

rs/types/types/src/batch/canister_http.rs
pierugo-dfinity: That makes sense!

pierugo-dfinity (approved):

PR #11041, opened 2026-08-05
fix: Use unique locks for each podman invocation

Bownairo:
Podman's default lock backend can race on startup with the error:
```
Error: failed to get new shm lock manager: failed to create 2048 locks in /libpod_rootless_lock_1001: file exists
```

The file is checked and then created without locking, so two startups can race and fail.

---

Avoid the race by making the lock unique for each runner. We use the fallback `file` lock and create a unique home for it for each runner.


basvandijk (approved):

mbjorkqvist (approved):

basvandijk:
Force-merging this in since it fixes a recurring issue on RBE @ Namespace.

PR #11045, opened 2026-08-06
fix: Insert HTTP outcalls pay-as-you-go refunds into the refund pool

eichhorl:
## Background
Under the new HTTP outcalls pay-as-you-go pricing, the caller's payment is split into a per-replica allowance which is stored in the context. Each replica consumes part of this allowance to produce their share of the response. Any unspent cycles out of this allowance are transferred back to the caller (initial refund). A late replica which didn't participate in the response, may still issue its refund later, after the call already received a response. Both the initial and late refunds are credited to the canister directly.

For this purpose, there are two collections holding HTTP outcalls in the call context manager: one holding in progress requests (waiting for the response and initial refund), and one holding delivered requests (waiting for late asynchronous refunds).

## Problem
During a subnet split, some canisters may move to a different subnet. Currently, all HTTP contexts stay on the original subnet (A'). Most (if not all) of the in-progress contexts will likely time out because:
1. The committee of a fully-replicated request changed completely, and can no longer collect the required shares.
2. The dedicated node of a non-replicated request may now be on a different subnet.
3. The chosen (min, max, total_requests) of a flexible outcall no longer make sense on a subnet half the size.
4. The selected committee of a flexible outcall was split into two subnets.
5. The transform function belongs to a canister that no longer exists on the subnet.

For legacy pricing, this is generally fine, since any refund is part of the timeout response, which can still be routed to the calling canister, even if it moved to a different subnet.

However, under pay-as-you-go pricing, any refunds can no longer be credited if the canister moved to a different subnet.

## Proposed Changes
Instead of crediting refunds to canisters directly, insert them into the "refund pool" such that they may be routed to the target canister, even if that canister moved to a different subnet.


rs/messaging/src/canister_http_spent.rs
```diff
@@ -288,52 +288,35 @@ fn apply_capped(
...
-                lost_by_deleted_canisters += CompoundCycles::<HTTPOutcalls>::new(
-                    accounting.refund,
-                    CanisterCyclesCostSchedule::Normal,
-                )
-                .nominal();
-            }
+        state.add_refund(sender, accounting.refund);
+        if !accounting.consumed.is_zero()
+            && let Some(canister) = state.canister_state_make_mut(&sender)
+        {
```
alin-at-dfinity, line 400: I suppose you could also credit the cycles if the canister is there; and produce a refund only if it isn't (because crediting is immediate and requires less work). But it's just an idea, I'm fine with leaving it as is.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -1105,6 +1105,17 @@ impl ReplicatedState {
         Ok(())
     }
 
+    /// Pools `amount` cycles to be refunded to `receiver`, wherever it is hosted.
+    ///
+    /// Message Routing routes the pooled refunds (via the loopback stream, if the
+    /// recipient is local) and credits them on induction, accounting for them as
+    /// lost if the recipient no longer exists.
+    ///
+    /// No-op if `amount` is zero.
```
alin-at-dfinity: Nit: I believe we should drop this bit. It's an implementation detail, at most relevant for tests.

```suggestion
```

rs/types/types/src/messages/inter_canister.rs
```diff
@@ -558,7 +558,9 @@ impl Hash for Response {
 }
 
 /// XNet message type (like `Request` and `Response`) for guaranteed delivery of
-/// refunds for best-effort calls.
+/// cycles that are refunded outside of a response: e.g. the payment of a dropped
+/// best-effort call; or the unspent part of an HTTP outcall's payment, which is
+/// only settled after the response was already delivered.
```
alin-at-dfinity, line 563: Grammar Nazi nitpicking: I believe you mean "the part [...] that is only settled after". "Which is only settled after" implies that the refund is always settled after.

```suggestion
/// cycles that are refunded outside of a response: e.g. the payment of a dropped
/// best-effort call; or the unspent part of an HTTP outcall's payment that was
/// only settled after the response was already delivered.
```

alin-at-dfinity (approved):

rs/messaging/src/canister_http_spent.rs
eichhorl: > crediting is immediate and requires less work

I think a small delay is fine. Unless you think the extra work is significant I would prefer to use a single mechanism for simplicity

PR #11051, opened 2026-08-06
chore: bump libflate in cargo

schneiderstefan:
A previous PR bumped the version of libflate in the bazel build script but forgot to do the same in cargo. This commit aligns the build scripts again. This has no impact on code actually running in production, but the discrepancy confused IDEs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2780,15 +2780,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
-name = "core2"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -17135,25 +17126,25 @@ dependencies = [
...
  "crc32fast",
  "dary_heap",
  "libflate_lz77",
+ "no_std_io2",
 ]
 
 [[package]]
 name = "libflate_lz77"
-version = "2.2.0"
+version = "2.3.0"
```
alin-at-dfinity, line 17142: Curious that `Cargo.toml` specifies `2.3.1` but `cargo` decided to go with `2.3.0`.

alin-at-dfinity (approved):

Cargo.lock
alin-at-dfinity: Never mind, this is a dependency of `libflate`, not `libflate` itself.

PR #11052, opened 2026-08-06
feat: require SEV launch measurements for new GuestOS versions

r-birkner:
Electing a GuestOS version now requires SEV-SNP launch measurements. Without a
measurement, nodes running the version cannot be attested, so a version that
lacks one should never become electable in the first place.

Note for release tooling: a `ReviseElectedGuestosVersions` proposal submitted
without `--guest-launch-measurements-path` will now be rejected.


schneiderstefan (approved):

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -48,6 +51,18 @@ async fn assert_failed_with_reason(gov: &Canister<'_>, proposal_id: ProposalId,
...
+    GuestLaunchMeasurements {
+        guest_launch_measurements: vec![GuestLaunchMeasurement {
+            measurement: vec![0x42; 48],
+            metadata: Some(GuestLaunchMeasurementMetadata {
+                kernel_cmdline: Some("foo=bar".to_string()),
+                vcpu_type: None,
+            }),
+        }],
+    }
+}
```
daniel-wong-dfinity-org-twin: Can you convert this (0-argument `new` function) into a [`lazy_static!`][lazy_static]?

[lazy_static]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md#testing

daniel-wong-dfinity-org-twin: Is there some constant somewhere for the length of this?

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -77,8 +92,10 @@ fn test_submit_and_accept_update_elected_replica_versions_proposal() {
                     .as_ref()
                     .map(|_| vec!["http://release_package.tar.zst".to_string()])
                     .unwrap_or_default(),
+                guest_launch_measurements: elect
+                    .as_ref()
+                    .map(|_| guest_launch_measurements_for_test()),
```
daniel-wong-dfinity-org-twin: map that doesn't actually use the argument is kind of sus. I mean, I have no better idea. This is just highly highly unusual. Yes, existing code already does this. That code is also highly sus.

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -77,8 +92,10 @@ fn test_submit_and_accept_update_elected_replica_versions_proposal() {
...
                 replica_version_to_elect: elect,
```
daniel-wong-dfinity-org-twin, line 93: I would have put this first, since this is the thing that's really important. [Lead with the lede][lede].

[lede]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md#basic-principles

Ofc, the code already put this in a weird position -> do not change in this PR.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
daniel-wong-dfinity-org-twin, line 1: I think the usual pattern here is to inline tests.

I guess Claude noticed [this], but [Consistency is king]. Or did you choose to do this in a separate file?

[this]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md#testing

[Consistency is king]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md#basic-principles

Don't get me wrong: I love separate file (because it maximizes build caching), but consistency is king. My plan is to do a sweep, and move tests to separate files. But for now, let's be consistent.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    GuestLaunchMeasurements {
+        guest_launch_measurements: vec![GuestLaunchMeasurement {
+            measurement: vec![0x42; 48],
+            metadata: Some(GuestLaunchMeasurementMetadata {
+                kernel_cmdline: Some("foo=bar".to_string()),
+                vcpu_type: None,
+            }),
+        }],
+    }
+}
```
daniel-wong-dfinity-org-twin: lazy_static

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+/// Launch measurements are one of the parameters that electing a version
+/// requires, so leaving them out makes the payload incomplete.
+#[test]
+fn test_electing_a_version_without_launch_measurements_is_rejected() {
+    let payload = new_elect_payload(None);
+    let result = payload.validate();
+    let defect = result.unwrap_err().to_lowercase();
+    for key_word in [
+        "all parameters",
+        "missing parameters: [\"guest_launch_measurements\"]",
```
daniel-wong-dfinity-org-twin: raw string instead of escaping double quotes.


```suggestion
        r#"missing parameters: ["guest_launch_measurements"]"#,
```

slash to escape is one of the worst inventions of all time.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+        assert!(defect.contains(key_word), "{key_word} not in {defect}");
+    }
+}
+
+#[test]
+fn test_electing_a_version_with_invalid_launch_measurement_is_rejected() {
+    // An SEV-SNP measurement is 48 bytes long, not 3.
+    let payload = new_elect_payload(Some(GuestLaunchMeasurements {
+        guest_launch_measurements: vec![GuestLaunchMeasurement {
+            measurement: vec![0x42; 3],
```
daniel-wong-dfinity-org-twin: Is this the bomb? If so, call it out:


```suggestion
            // This is the bomb. It is supposed to be of length 48.
            measurement: vec![0x42; 3],
```

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[test]
+fn test_electing_a_version_with_launch_measurements_is_valid() {
+    let payload = new_elect_payload(Some(guest_launch_measurements()));
+    let result = payload.validate();
+    assert_eq!(result, Ok(()));
+}
+
+/// Launch measurements are one of the parameters that electing a version
+/// requires, so leaving them out makes the payload incomplete.
```
daniel-wong-dfinity-org-twin: This seems redundant vs. the name. If so, consider dropping.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    guest_launch_measurements: Option<GuestLaunchMeasurements>,
+) -> ReviseElectedGuestosVersionsPayload {
+    ReviseElectedGuestosVersionsPayload {
+        replica_version_to_elect: Some(REPLICA_VERSION_ID.to_string()),
+        release_package_sha256_hex: Some(RELEASE_PACKAGE_SHA256_HEX.to_string()),
+        release_package_urls: vec![RELEASE_PACKAGE_URL.to_string()],
+        guest_launch_measurements,
+        replica_versions_to_unelect: vec![],
+    }
+}
```
daniel-wong-dfinity-org-twin: This is fine, but it could also be a lazy_static. E.g.

```
lazy_static! {
    static ref REVISE_ELECTED_GUESTOS_VERSIONS_PAYLOAD ...
}

#[test]
fn test_stuff() {
    let mut payload = REVISE_ELECTED_GUESTOS_VERSIONS_PAYLOAD.clone();
    payload.guest_launch_measurements =  whatever;

    let result = payload.validate();

    assert...
}
```

Constants are good. lazy_static solves the problem of Rust's `const`s being too strict for practical use in many cases (a common disease in the Rust language, which itself is named after a disease...).

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    let result = payload.validate();
+
+    let defect = result.unwrap_err().to_lowercase();
+    for key_word in ["guest_launch_measurements", "48 bytes"] {
+        assert!(defect.contains(key_word), "{key_word} not in {defect}");
+    }
+}
+
+/// Launch measurements are only required when a version is elected. Unelecting
+/// does not involve any GuestOS image.
```
daniel-wong-dfinity-org-twin: I mean, you could just have the name explain this.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    let result = payload.validate();
+
+    assert_eq!(result, Ok(()));
+}
+
+/// Launch measurements only mean something for a version that is being
+/// elected. On an unelect-only payload they would be silently dropped, so they
+/// are reported as a defect instead.
+#[test]
+fn test_unelecting_a_version_with_launch_measurements_is_rejected() {
```
daniel-wong-dfinity-org-twin: This seems to directly contradict the name of the previous test. How about something like this instead:


```suggestion
fn test_launch_measurements_when_not_electing_is_rejected() {
```

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    let result = payload.validate();
+
+    let defect = result.unwrap_err().to_lowercase();
+    for key_word in [
+        "all parameters",
+        "replica_version_to_elect",
+        "release_package_sha256_hex",
+        "release_package_urls",
```
daniel-wong-dfinity-org-twin: I would think that something about launch measurements would be mentioned, since their presence is the defect.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+#[test]
+fn test_incomplete_election_parameters_are_rejected() {
+    // The release package hash and URLs are missing.
+    let payload = ReviseElectedGuestosVersionsPayload {
+        replica_version_to_elect: Some(REPLICA_VERSION_ID.to_string()),
+        guest_launch_measurements: Some(guest_launch_measurements()),
+        ..Default::default()
```
daniel-wong-dfinity-org-twin: This is a very short test, so this is not so important, but try to point out the defect in the input. Here, it's hard to see, there is nothing you can directly point at, because the missing fields are taken care of by this line. Still, you could do something like


```suggestion
        // release_package_* fields are missing.
        ..Default::default()
```

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,68 @@ pub struct ReviseElectedGuestosVersionsPayload {
 impl ReviseElectedGuestosVersionsPayload {
     pub fn is_electing_a_version(&self) -> Result<bool, String> {
```
daniel-wong-dfinity-org-twin, line 109: I know the code was already like this (so, do not fix it), but it is totally wacky for [`is_whatever`][is] to return `Result` instead of `bool`. Yes, I understand, it's a `Result<bool, ...>`, but that is not what people expect from `is_whatever`, at least I certainly do not.

[is]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md#naming

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,68 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
         }
 
-        Err("All parameters to elect a version have to be either set or unset.".into())
+        // Leave breadcrumbs: which parameters were missing.
+        let mut unset_params = Vec::new();
+        for (name, is_set) in elect_params {
+            if !is_set {
+                unset_params.push(name);
+            }
+        }
```
daniel-wong-dfinity-org-twin, line 143: You are a good person. Rare.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,68 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
     pub fn is_unelecting_a_version(&self) -> bool {
         !self.replica_versions_to_unelect.is_empty()
     }
 
     pub fn validate(&self) -> Result<(), String> {
-        if self.is_electing_a_version()? || self.is_unelecting_a_version() {
-            Ok(())
-        } else {
-            Err("At least one version has to be elected or unelected.".into())
+        if !self.is_electing_a_version()? && !self.is_unelecting_a_version() {
```
daniel-wong-dfinity-org-twin:
```suggestion
        let is_making_a_change = self.is_electing_a_version()? || self.is_unelecting_a_version()
        if !is_making_a_change {
```

rs/registry/canister/unreleased_changelog.md
daniel-wong-dfinity-org-twin, line 1: Seems verbose. I think all you need to say is

> Guest launch measurements is now required (when electing a new Guestos version).

I think this more strongly [leads with the led/avoids burying the lede][lede].

[lede]: https://sourcegraph.com/r/github.com/dfinity/ic@master/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?L248#basic-principles

rs/nns/integration_tests/src/upgrades_handler.rs
r-birkner: We don't have one.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions/tests.rs
r-birkner: I don't know who noticed it. It wasn't for sure me, but I inlined it.

rs/nns/integration_tests/src/upgrades_handler.rs
daniel-wong-dfinity-org-twin: Can you add one in a later PR? Seems like it would be used in production code, for validation, no? Or if there is no validation, then, this doesn't need to be of length 48, right?

Plus, it's hard for me to remember whether it's `[element; len]` or `[len; element]`. If you had a constant named `MEASURMENT_LEN`, then, at least, I would see the intent.

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -48,6 +52,20 @@ async fn assert_failed_with_reason(gov: &Canister<'_>, proposal_id: ProposalId,
     );
 }
 
+lazy_static! {
```
daniel-wong-dfinity-org-twin: Move this to where consts would be: between `use`s and `fn`s + `struct`s.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,257 @@ pub struct ReviseElectedGuestosVersionsPayload {
 impl ReviseElectedGuestosVersionsPayload {
     pub fn is_electing_a_version(&self) -> Result<bool, String> {
```
daniel-wong-dfinity-org-twin, line 109: Tangent: let's document this, since we are touching it anyway. Suggestion:

> Returns `true` if replica_version_to_elect is Some (otherwise, `false` is returned).
>
> Returns Err if self is invalid due to related fields being inconsistent with replica_version_to_elect. More precisely, when replica_version_to_elect is Some, then the following fields must also be Some:
>
> 1. ...
>
> Otherwise (replica_version_to_elect is None) those fields must also be None.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,257 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
+        let payload = ELECT_PAYLOAD.clone();
+        let result = payload.validate();
+        assert_eq!(result, Ok(()));
+    }
+
+    #[test]
+    fn test_electing_a_version_without_launch_measurements_is_rejected() {
+        // Step 1: Prepare the world.
+        let mut payload = ELECT_PAYLOAD.clone();
+        payload.guest_launch_measurements = None;
```
daniel-wong-dfinity-org-twin, line 226: This is completely fine, but if you wanted to, you could do it another way:

```
let payload = ReviseElectedGuestosVersionsPayload {
    field: defective_value,
    ..ELECT_PAYLOAD.clone()
};
```

This is more verbose, but avoids `mut`. Totally up to you.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,257 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
+    /// does not involve any GuestOS image.
+    #[test]
+    fn test_unelecting_a_version_without_launch_measurements_is_valid() {
+        let payload = ReviseElectedGuestosVersionsPayload {
+            replica_version_to_elect: None,
+            release_package_sha256_hex: None,
+            release_package_urls: vec![],
+            guest_launch_measurements: None,
+            replica_versions_to_unelect: vec![REPLICA_VERSION_ID.to_string()],
+        };
```
daniel-wong-dfinity-org-twin, line 292:
```suggestion
        let payload = ReviseElectedGuestosVersionsPayload {
            replica_version_to_elect: None,
            // None is Ok (in fact, required), because not electing a version.
            guest_launch_measurements: None,

            // Since not electing a version, we must at least unelect one to
            // avoid being a no-op, which is not allowed.
            replica_versions_to_unelect: vec![REPLICA_VERSION_ID.to_string()],

            release_package_sha256_hex: None,
            release_package_urls: vec![],
        };
```

Since guest_launch_measurements is the whole point here, it should be more prominent, not be last.

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,257 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
+
+    /// Launch measurements only mean something for a version that is being
+    /// elected. On an unelect-only payload they would be silently dropped, so they
+    /// are reported as a defect instead.
+    #[test]
+    fn test_unelecting_a_version_with_launch_measurements_is_rejected() {
+        // Step 1: Prepare the world.
+        let payload = ReviseElectedGuestosVersionsPayload {
+            guest_launch_measurements: Some(GUEST_LAUNCH_MEASUREMENTS.clone()),
+            replica_versions_to_unelect: vec![REPLICA_VERSION_ID.to_string()],
```
daniel-wong-dfinity-org-twin, line 307:
```suggestion
            replica_version_to_elect: None
            // The previous test shows that None does NOT explode. This test shows
            // that Some DOES explode.
            guest_launch_measurements: Some(GUEST_LAUNCH_MEASUREMENTS.clone()),

            replica_versions_to_unelect: vec![REPLICA_VERSION_ID.to_string()],
```

rs/registry/canister/src/mutations/do_revise_elected_replica_versions.rs
```diff
@@ -104,31 +105,257 @@ pub struct ReviseElectedGuestosVersionsPayload {
...
+            "replica_version_to_elect",
+            "release_package_sha256_hex",
+            "release_package_urls",
+        ] {
+            assert!(defect.contains(key_word), "{key_word} not in {defect}");
+        }
+    }
+
+    #[test]
+    fn test_neither_electing_nor_unelecting_is_rejected() {
```
daniel-wong-dfinity-org-twin: Do as  you see fit.

```suggestion
    fn test_no_op_is_rejected() {
```


```suggestion
    fn test_must_make_some_change() {
```

daniel-wong-dfinity-org-twin (approved):

PR #10995, opened 2026-08-03
feat: migrate non-governance canisters to new ic_cdk call interface

nmattia:
This migrates most canisters outside of the production NNS/governance canisters to the new ic-cdk cross-canister call interface introduced in 0.18. This tries to keep the old behavior wherever possible.

PR #11055, opened 2026-08-06
feat: migrate remaining ic-cdk ::call to Call

nmattia:
This migrates the governance & boundary node canisters to the new `ic-cdk` `Call` interface. This is the last big step before we can fully migrate to `ic-cdk` v0.20.

This builds on https://github.com/dfinity/ic/pull/10995.

### DIFF
--- a/rs/nervous_system/root/src/change_canister.rs
+++ b/rs/nervous_system/root/src/change_canister.rs
@@ -10,12 +10,12 @@
 //! Of course, after calling the Management canister's `install_code` method,
 //! this start the canister back up again, and unlocks it.
 
-#![allow(deprecated)]
 use crate::{LOG_PREFIX, private::perform_offline_canister_maintenance};
 use candid::{CandidType, Deserialize, Encode, Principal};
 use dfn_core::api::CanisterId;
 #[cfg(target_arch = "wasm32")]
 use dfn_core::println;
+use ic_cdk::call::{Call, CallFailed, RejectCode};
 use ic_crypto_sha2::Sha256;
 use ic_management_canister_types_private::{
     CanisterInstallModeV2, ChunkHash, IC_00, InstallChunkedCodeArgs, InstallCodeArgsV2,
@@ -265,7 +265,9 @@ pub async fn change_canister(request: ChangeCanisterRequest) -> Result<(), Strin
     // This handles errors coming from install_code (which is more or less a
     // thin wrapper around the Management canister method), such as the
     // DESTINATION_INVALID (3) reject code.
-    result.map_err(|(rejection_code, message)| {
+    result.map_err(|err| {
+        let (rejection_code, message) = map_call_error(err);
+
         format!(
             "Attempt to call install_code with request {request_str} failed with code \
              {rejection_code:?}: {message}"
@@ -273,8 +275,22 @@ pub async fn change_canister(request: ChangeCanisterRequest) -> Result<(), Strin
     })
 }
 
+/// Translates a failed call into the `(reject code, message)` pair that the error message
+/// above reports. Reports the same codes that the deprecated `ic_cdk::api::call` did.
+fn map_call_error(err: CallFailed) -> (RejectCode, String) {
+    match err {
+        CallFailed::CallRejected(rejected) => (
+            rejected.reject_code().unwrap_or(RejectCode::SysUnknown),
+            rejected.reject_message().to_string(),
+        ),
+
+        // The call never left this canister, so it may well succeed if retried.
+        err => (RejectCode::SysTransient, err.to_string()),
+    }
+}
+
 /// Calls a function of the management canister to install the requested code.
-async fn install_code(request: ChangeCanisterRequest) -> ic_cdk::api::call::CallResult<()> {
+async fn install_code(request: ChangeCanisterRequest) -> Result<(), CallFailed> {
     let ChangeCanisterRequest {
         mode,
         canister_id,
@@ -308,12 +324,13 @@ async fn install_code(request: ChangeCanisterRequest) -> ic_cdk::api::call::Call
             arg,
             sender_canister_version,
         };
-        ic_cdk::api::call::call(
+        Call::unbounded_wait(
             Principal::try_from(IC_00.get().as_slice()).unwrap(),
             "install_chunked_code",
-            (&argument,),
         )
+        .with_arg(&argument)
         .await
+        .map(|_reply| ())
     } else {
         let argument = InstallCodeArgsV2 {
             mode,
@@ -322,12 +339,13 @@ async fn install_code(request: ChangeCanisterRequest) -> ic_cdk::api::call::Call
             arg,
             sender_canister_version,
         };
-        ic_cdk::api::call::call(
+        Call::unbounded_wait(
             Principal::try_from(IC_00.get().as_slice()).unwrap(),
             "install_code",
-            (&argument,),
         )
+        .with_arg(&argument)
         .await
+        .map(|_reply| ())
     }
 }
 
--- a/rs/nervous_system/root/src/change_canister.rs
+++ b/rs/nervous_system/root/src/change_canister.rs
@@ -15,7 +15,7 @@ use candid::{CandidType, Deserialize, Encode, Principal};
 use dfn_core::api::CanisterId;
 #[cfg(target_arch = "wasm32")]
 use dfn_core::println;
-use ic_cdk::call::{Call, CallFailed, RejectCode};
+use ic_cdk::call::{Call, CallFailed};
 use ic_crypto_sha2::Sha256;
 use ic_management_canister_types_private::{
     CanisterInstallModeV2, ChunkHash, IC_00, InstallChunkedCodeArgs, InstallCodeArgsV2,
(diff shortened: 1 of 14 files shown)


blind-oracle (approved):

schneiderstefan (approved):

rs/nervous_system/root/src/change_canister.rs
```diff
@@ -265,16 +265,32 @@ pub async fn change_canister(request: ChangeCanisterRequest) -> Result<(), Strin
...
+/// Translates a failed call into the `(reject code, message)` pair that the error message
+/// above reports. Reports the same codes that the deprecated `ic_cdk::api::call` did.
+fn map_call_error(err: CallFailed) -> (RejectCode, String) {
+    match err {
+        CallFailed::CallRejected(rejected) => (
+            rejected.reject_code().unwrap_or(RejectCode::SysUnknown),
+            rejected.reject_message().to_string(),
+        ),
+
+        // The call never left this canister, so it may well succeed if retried.
```
daniel-wong-dfinity-org-twin: > never left this canister

But, how do you know that? I mean, I think you are assuming that anything that isn't CallRejected can only ever be SysTransient. That might be true today, but this code will still compile tomorrow even if this is no longer true, because `err` matches **_anything_**, right?

Stepping back a bit, I guess this is only used in one place to help with converting a failed result to a string, right? I mean, I guess you could just do something approximately like `format!("... {err:?} ... ")` instead, right? Maybe we can skip this function entirely?

Plus, the name of this function is super generic, and therefore, doesn't really explain its actual behavior. "map"? In what way?? "map" is a literal synonym for "apply some function". What function?

rs/nervous_system/runtime/src/lib.rs
```diff
@@ -106,6 +106,23 @@ impl Runtime for DfnRuntime {
...
+    match err {
+        IcCdkCallError::CallRejected(rejected) => (
+            rejected.raw_reject_code() as i32,
+            rejected.reject_message().to_string(),
+        ),
+        // A response that cannot be decoded means the callee misbehaved.
+        IcCdkCallError::CandidDecodeFailed(err) => {
+            (RejectCode::CanisterError as i32, err.to_string())
+        }
+        // The call never left this canister, so it may well succeed if retried.
```
daniel-wong-dfinity-org-twin: Ditto: err matches ANYTHING else, including things that do not exist yet.

rs/nervous_system/runtime/src/lib.rs
```diff
@@ -106,6 +106,23 @@ impl Runtime for DfnRuntime {
 
 pub struct CdkRuntime;
 
+/// Translates a failed call into the `(reject code, message)` pair that
+/// [`Runtime`] reports to its callers.
+fn map_call_error(err: IcCdkCallError) -> (i32, String) {
```
daniel-wong-dfinity-org-twin: Let's maybe name this `into_reject_code_and_message`.

This converter should maybe live in a shared library. Perhaps, it can live in rs/nervous_system/ic_cdk_shims or something?

rs/nns/cmc/src/main.rs
```diff
@@ -2265,28 +2257,31 @@ async fn do_create_canister(
...
-            u128::from(cycles),
-        )
-        .await;
+            })
+            .with_cycles(u128::from(cycles))
+            .await
+            .map_err(IcCdkCallError::from)
+            .and_then(|response| {
+                response
+                    .candid::<CanisterIdRecord>()
```
daniel-wong-dfinity-org-twin, line 2270: This manual decoding is something that we really need to get rid of (ofc, that is out of scope for this PR), because even if this was wrong (e.g. `A` instead of `(A,)`), the compiler would not care. Only an integration test would reveal such a mistake, and I really do not know if this is covered by any integration test...

rs/nns/sns-wasm/canister/canister.rs
```diff
@@ -102,16 +102,13 @@ impl CanisterApi for CanisterApiImpl {
...
-        )
-        .await;
-
-        response.map_err(handle_call_error(format!(
-            "Failed to delete canister {canister}"
-        )))
+        Call::unbounded_wait(CanisterId::ic_00().get().0, "delete_canister")
+            .with_arg(CanisterIdRecord::from(canister))
+            .await
+            .map(|_reply| ())
```
daniel-wong-dfinity-org-twin: I think for this to be really behavior preserving, you need to do


```suggestion
            .map(|response| response.candid_tuple::<()>())
```

right?

Ditto elsewhere.

rs/nns/sns-wasm/canister/canister.rs
```diff
@@ -193,43 +192,53 @@ impl CanisterApi for CanisterApiImpl {
...
+            )))
+    }
+}
 
-        response.map_err(handle_call_error(format!(
-            "Failed to send cycles to canister {target}"
-        )))
+/// Translates a failed call into the `(reject code, message)` pair that the error messages
+/// below report.
+fn map_call_error(err: IcCdkCallError) -> (i32, String) {
```
daniel-wong-dfinity-org-twin: Ditto.

rs/nns/sns-wasm/canister/canister.rs
```diff
@@ -193,43 +192,53 @@ impl CanisterApi for CanisterApiImpl {
...
 }
 
-/// This handles the errors returned from ic_cdk::call (and related methods)
-fn handle_call_error(prefix: String) -> impl FnOnce((RejectionCode, String)) -> String {
-    move |(code, msg)| {
-        let err = format!("{}: error code {}: {}", prefix, code as i32, msg);
+/// This handles the errors returned from Call::unbounded_wait (and related methods)
+fn handle_call_error<E: Into<IcCdkCallError>>(prefix: String) -> impl FnOnce(E) -> String {
+    move |err| {
+        let (code, msg) = map_call_error(err.into());
```
daniel-wong-dfinity-org-twin:
```suggestion
fn handle_call_error<E>(prefix: String) -> impl FnOnce(E) -> String
where
    E: Into<IcCdkCallError>
{
    move |err| {
        let (code, msg) = map_call_error(err.into());
```

The main thing is, use `where`. This is to minimize clutter before the parameter list.

rs/rust_canisters/ecdsa/src/main.rs
```diff
@@ -30,27 +30,26 @@ async fn get_sig(options: Options) {
...
+                        .map(ByteBuf::from)
+                        .collect()
+                ),
+                key_id: EcdsaKeyId {
+                    curve: EcdsaCurve::Secp256k1,
+                    name: options.key_name,
+                },
+            })
+            .unwrap(),
+        )
```
daniel-wong-dfinity-org-twin: Wouldn't `.with_arg(SignWithECDSAArgs { ... })` be equivalent?

rs/sns/governance/canister/canister.rs
```diff
@@ -124,6 +124,27 @@ impl CanisterEnv {
     }
 }
 
+/// Translates a failed call into the `(error code, message)` pair that
+/// [`Environment::call_canister`] reports to its callers.
+fn map_call_error(err: IcCdkCallError) -> (Option<i32>, String) {
```
daniel-wong-dfinity-org-twin: Ditto.

rs/sns/swap/src/swap.rs
```diff
@@ -108,11 +107,26 @@ impl From<(Option<i32>, String)> for CanisterCallError {
...
+                rejected.reject_message().to_string(),
+            ),
+
+            // A response that cannot be decoded means the callee misbehaved.
+            IcCdkCallError::CandidDecodeFailed(err) => {
+                (RejectCode::CanisterError as i32, err.to_string())
+            }
+
+            // The call never left this canister, so it may well succeed if retried.
+            err => (RejectCode::SysTransient as i32, err.to_string()),
```
daniel-wong-dfinity-org-twin: Ditto.

rs/nervous_system/runtime/src/lib.rs
nmattia: Yep, it's the started name I've used in this and other PRs. I'll consolidate everything after this one goes in, I'm tracking this. `into_reject_code_and_message` works for me.

rs/nns/cmc/src/lib.rs
```diff
@@ -67,7 +67,11 @@ pub fn caller() -> PrincipalId {
 
 /// Translates a failed call into the `(reject code, message)` pair that [`call_protobuf`]
 /// and this canister's error messages report.
-pub fn map_call_error(err: IcCdkCallError) -> (i32, String) {
+///
+/// The match is deliberately exhaustive (rather than using a catch-all arm) so
+/// that a new [`IcCdkCallError`] variant forces us to revisit this mapping
+/// instead of silently classifying it as [`RejectCode::SysTransient`].
+pub fn into_reject_code_and_message(err: IcCdkCallError) -> (i32, String) {
```
daniel-wong-dfinity-org-twin, line 74: Why not spin this out to a library that can be shared?

rs/nns/sns-wasm/canister/canister.rs
```diff
@@ -205,7 +209,11 @@ impl CanisterApi for CanisterApiImpl {
 
 /// Translates a failed call into the `(reject code, message)` pair that the error messages
 /// below report.
-fn map_call_error(err: IcCdkCallError) -> (i32, String) {
+///
+/// The match is deliberately exhaustive (rather than using a catch-all arm) so
+/// that a new [`IcCdkCallError`] variant forces us to revisit this mapping
+/// instead of silently classifying it as [`RejectCode::SysTransient`].
+fn into_reject_code_and_message(err: IcCdkCallError) -> (i32, String) {
```
daniel-wong-dfinity-org-twin, line 216: Ditto.

rs/sns/governance/canister/canister.rs
```diff
@@ -126,7 +126,11 @@ impl CanisterEnv {
 
 /// Translates a failed call into the `(error code, message)` pair that
 /// [`Environment::call_canister`] reports to its callers.
-fn map_call_error(err: IcCdkCallError) -> (Option<i32>, String) {
+///
+/// The match is deliberately exhaustive (rather than using a catch-all arm) so
+/// that a new [`IcCdkCallError`] variant forces us to revisit this mapping
+/// instead of silently classifying it as [`RejectCode::SysTransient`].
+fn into_reject_code_and_message(err: IcCdkCallError) -> (Option<i32>, String) {
```
daniel-wong-dfinity-org-twin, line 133: Ditto.

rs/nns/cmc/src/lib.rs
nmattia: Yep as mentioned [here](https://github.com/dfinity/ic/pull/11055#discussion_r3735906970) I'll do a final pass to extract this. So far I've copy pasted this helper wherever necessary as the migration is undergoing but will unite everything.

daniel-wong-dfinity-org-twin (approved):

PR #11057, opened 2026-08-06
feat: require launch measurements for SEV-enabled subnets

r-birkner:
An SEV-enabled subnet may now only run a GuestOS version that has `guest_launch_measurements`. This is enforced as a registry invariant, so any mutation that would leave an SEV-enabled subnet on a version without launch measurements is rejected.

Note: A CloudEngine may leave `replica_version_id` blank, in which case it runs the versions of the `StandardEngineReplicaVersionRecord`.


r-birkner:
Note for Reviewer @daniel-wong-dfinity-org: There is also [this PR](https://github.com/dfinity/ic/pull/11052) that addresses the same issue in a different place. By adding this one, we are protected at the version election and the deployment. But we can also drop this if you think it is not needed.

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -275,10 +286,12 @@ pub(crate) fn get_subnet_records_map(
...
 }
 
-/// All nodes of a subnet must support SEV in order for SEV to be enabled on the subnet.
+/// SEV-enabled subnets must consist of SEV-supporting nodes only, and must run a
+/// GuestOS version that has launch measurements.
 fn check_sev_subnet_invariants(
     subnet_id: SubnetId, // only used for error messages, so we can report which subnet is non-compliant
     subnet_members: HashSet<NodeId>,
+    replica_version_id: &str,
     snapshot: &RegistrySnapshot,
```
daniel-wong-dfinity-org-twin: I would say that this is on the verge of having "too many" parameters... I think one way to cut it down is just to pass subnet_record + snapshot. In the not too distant future, subnet_record will have ID in it (ask Eero).

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -315,6 +328,57 @@ fn check_sev_subnet_invariants(
...
+        .filter(|replica_version_id| {
+            let replica_version_record = get_value_from_snapshot::<ReplicaVersionRecord>(
+                snapshot,
+                make_replica_version_key(replica_version_id),
+            );
+
+            let Some(replica_version_record) = replica_version_record else {
+                // A version in use without a ReplicaVersionRecord is reported
+                // by the replica version invariants, so there is no need to report it here too.
+                return false;
```
daniel-wong-dfinity-org-twin: If there is no ReplicaVersionRecord, then, guest launch measurements are certainly missing!

[Do NOT sweep problem under the rug][rug].

[rug]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?L76&view=raw

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -315,6 +328,57 @@ fn check_sev_subnet_invariants(
...
+    subnet_id: SubnetId, // only used for error messages, so we can report which subnet is non-compliant
+    subnet_replica_version_id: &str,
+    snapshot: &RegistrySnapshot,
+) -> Result<(), InvariantCheckError> {
+    let replica_version_ids = if subnet_replica_version_id.is_empty() {
+        // A subnet normally runs the single version that its SubnetRecord names. A
+        // CloudEngine is allowed to leave that field blank, and then runs the versions
+        // of the StandardEngineReplicaVersionRecord instead (both the old and the new
+        // one, because the engines are being upgraded from one to the other).
+        get_all_standard_engine_replica_versions(snapshot)
```
daniel-wong-dfinity-org-twin: So, this would get run on every engine?? That's super repetitive. Perhaps, we can just say that replica versions referenced from StandardEngineReplicaVersion must always support SEV? Then, we can skip this check for engines with blank replica_version_id entirely.

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+fn setup_sev_enabled_subnet(
+    system_subnet_id: SubnetId,
+    test_subnet_id: SubnetId,
+) -> (RegistrySnapshot, SubnetRecord) {
+    let (mut snapshot, mut test_subnet_record) =
+        setup_minimal_registry_snapshot_for_check_subnet_invariants(
+            system_subnet_id,
+            test_subnet_id,
+            4,    // num_nodes_in_test_subnet
+            true, // with_chip_id
```
daniel-wong-dfinity-org-twin, line 184: Did Claude do this?? If so, my CLAUDE.md files are working!

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+         launch measurements:",
+    );
+}
+
+/// Only SEV-enabled subnets are constrained. Versions without launch
+/// measurements have been elected in the past, and the subnets running them
+/// must stay compliant.
+#[test]
+fn test_subnet_without_sev_can_run_version_without_launch_measurements() {
+    // Step 1: Prepare the world.
```
daniel-wong-dfinity-org-twin, line 306: Example of test phases rule being followed.

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+            // A SEV-SNP measurement is 48 bytes long. The value itself does not matter here.
+            measurement: vec![0x42; 48],
+            metadata: None,
+        }],
+    });
+
+    snapshot.insert(
+        make_replica_version_key(replica_version_id).into_bytes(),
+        ReplicaVersionRecord {
+            release_package_sha256_hex: "".to_string(),
```
daniel-wong-dfinity-org-twin: What about C0FFEEF00D?

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+            release_package_urls: vec![],
```
daniel-wong-dfinity-org-twin: How about example.com?

Even if it makes no difference to the test, if it is not difficult to have realistic test data, do it. One (admittedly small) benefit of this is that people sometimes use test code as examples. Ideally, tests should be realistic anyway.

If you want, you can add a comment that these values do not affect test results, but are there just for realism.

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+
+/// Elects the given GuestOS version, with launch measurements or without.
+fn elect_replica_version(
+    snapshot: &mut RegistrySnapshot,
+    replica_version_id: &str,
+    with_launch_measurements: bool,
+) {
+    let guest_launch_measurements = with_launch_measurements.then(|| GuestLaunchMeasurements {
+        guest_launch_measurements: vec![GuestLaunchMeasurement {
+            // A SEV-SNP measurement is 48 bytes long. The value itself does not matter here.
```
daniel-wong-dfinity-org-twin: This should really be a constant. If you plan to do this in another PR, fine. Just a reminder.

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+    (snapshot, test_subnet_record)
+}
+
+/// Elects the given GuestOS version, with launch measurements or without.
+fn elect_replica_version(
+    snapshot: &mut RegistrySnapshot,
+    replica_version_id: &str,
+    with_launch_measurements: bool,
+) {
+    let guest_launch_measurements = with_launch_measurements.then(|| GuestLaunchMeasurements {
```
daniel-wong-dfinity-org-twin: `.then(` is esoteric. E.g. it is only occurs [43 times][then] in the ic repo, and some of those occurrences are comments. Also, it saves only a very small amount of code. Furthermore, the word "then" really does NOT hint that an `Option` is returned. Therefore, it fails to [pull its own weight][weight]. Therefore, it should NOT be used. Instead,

[then]: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/dfinity/ic%24+%22.then(%22&patternType=regexp&case=yes&sm=0

[weight]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?L197&view=raw

```
if with_launch_measurements {
    Some(...)
} else {
    None
}
```

is completely accessible to all "good" Rust programmers, not just the ones who have memorized the obscure corners of the standard library.

rs/registry/canister/src/invariants/subnet/tests.rs
```diff
@@ -164,6 +167,217 @@ fn all_nodes_have_a_chip_id() {
...
+        .encode_to_vec(),
+    );
+}
+
+/// Points the given subnet at the given GuestOS version.
+fn deploy_replica_version_to_subnet(
+    snapshot: &mut RegistrySnapshot,
+    subnet_id: SubnetId,
+    subnet_record: &mut SubnetRecord,
+    replica_version_id: &str,
```
daniel-wong-dfinity-org-twin: It is confusing that the comment uses the term "GuestOS version" while the function and parameter names contain `replica_version`. If you want to ALSO mention "GuestOS version", ok, but please, also mention "Replica". E.g. "GuestOS/Replica version".

daniel-wong-dfinity-org-twin (commented):
Do we have some engines without SEV? If all engines already have SEV, I'd suggest that we simply require that StandardEngineReplicaVersion only refers to SEV Replica versions (i.e. those that have guest launch measurements).

rs/registry/canister/src/invariants/subnet/tests.rs
r-birkner: Yes

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -290,10 +283,15 @@ pub(crate) fn get_subnet_records_map(
 /// GuestOS version that has launch measurements.
 fn check_sev_subnet_invariants(
     subnet_id: SubnetId, // only used for error messages, so we can report which subnet is non-compliant
```
daniel-wong-dfinity-org-twin: > only used for error messages

Consider moving this last then. [Lead with the lede][lede].

[lede]: https://github.com/dfinity/ic/blob/eee7120754735b73a14d9024d536bae7860d0159/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L49

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -328,52 +326,29 @@ fn check_sev_subnet_invariants(
...
 
     Ok(())
 }
 
-/// An SEV-enabled subnet must run only GuestOS versions that have launch measurements; otherwise
+/// An SEV-enabled subnet must run only a GuestOS version that has launch measurements; otherwise
 /// its nodes cannot be attested, which defeats the purpose of enabling SEV.
 fn check_sev_subnet_launch_measurements(
-    subnet_id: SubnetId, // only used for error messages, so we can report which subnet is non-compliant
+    subnet_id: SubnetId,
```
daniel-wong-dfinity-org-twin: Isn't the comment correct? If so, ditto.

rs/registry/canister/src/invariants/subnet.rs
```diff
@@ -328,52 +326,29 @@ fn check_sev_subnet_invariants(
...
-                // by the replica version invariants, so there is no need to report it here too.
-                return false;
-            };
-
-            replica_version_record.guest_launch_measurements.is_none()
-        })
-        .collect::<Vec<String>>();
-
-    if !versions_missing_launch_measurements.is_empty() {
+    if !has_launch_measurements(subnet_replica_version_id, snapshot) {
```
daniel-wong-dfinity-org-twin: So, I guess it is assumed that subnet_id refers to an SEV subnet? Can you mention this in the doc comments?

I would think that what this function does is (approximately):

```
// IIUC, your code does NOT do this.
if !is_an_sev_subnet() {
    return Ok(());
}

// Your code DOES INDEED do this though.
if !replica_version_has_guest_launch_measurements() {
    return Err(...);
}

Ok(())
```

Otherwise, it seems like this is an almost pointlessly thin wrapper around has_launch_measurements. But do as you see fit.

rs/registry/canister/unreleased_changelog.md
```diff
@@ -11,9 +11,14 @@ on the process that this file is part of, see
 
 * Invariant requiring that SEV-enabled subnets may only run a GuestOS version that has
   `guest_launch_measurements`.
+* Invariant requiring that both versions of the `StandardEngineReplicaVersionRecord` have
+  `guest_launch_measurements`.
+* Add a `replica_version_id` to `ReplicaVersionRecord`s, and backfill with a data migration.
```
daniel-wong-dfinity-org-twin: This seems to come from another PR that you merged in? GH is so confusing...

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/unreleased_changelog.md
r-birkner: Cleaned up.

rs/registry/canister/src/invariants/subnet.rs
r-birkner: You are right. Unnecessary. Just inlined it into the calling function and that's it.

PR #11058, opened 2026-08-06
fix: Fix baremetal test after #10457

frankdavid:
#10457 introduced a templated guestos unit.


Bownairo (approved):
🙌 Thank you!

My grep wasn't good enough and this isn't included in `CI_ALL_BAZEL_TARGETS`. Maybe we should run these tests too when that's set?

PR #11060, opened 2026-08-07
feat(cketh): read a native ETH balance via the EVM RPC canister

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding), first of a seven-PR stack.

## Why

Funding the sweeper address with gas requires knowing how much gas it already holds. The EVM RPC canister exposes no endpoint for a native ETH balance, and its Rust client offers no getter for one, so the minter currently has no way to ask.

## What

Reads the balance through the EVM RPC canister's generic JSON-RPC passthrough, which forwards a payload to every provider and agrees on one answer under the configured consensus strategy. That strategy is a threshold of the providers — 3 of 4 on mainnet, 2 of 4 on Sepolia — and it is the only agreement accepted: there is no client-side reduction, so a result the canister reports as inconsistent stays an error rather than being resolved by picking a winner.

Because the canister deserializes each response's `result` field, what the minter receives is the quantity itself rather than any surrounding JSON. It is therefore decoded exactly: quotes, padding, leading zeros and sign characters are the provider's own malformation and are rejected rather than repaired.

A failed read is an error, never a zero. This is the decision the rest of the stack depends on: confusing "could not read the balance" with "no gas left" would burn ckETH to top up an address that is already funded, which is pure loss. The request builder and the result decoder are pure functions so both sides of that guarantee are pinned directly, including a test asserting that no error input can decode to a zero balance.

The route was also proven end to end against a live EVM RPC canister and a local anvil node, reaching 3-of-4 consensus at both `latest` and `finalized`.

## Stack

| # | PR | Status |
|---|----|--------|
| 1 | Read a native ETH balance via the EVM RPC canister | **this PR** |
| 2 | #11065 — Burn ckETH from the minter's own fee subaccount | ready for review |
| 3 | #11072 — Add the SweeperFunding withdrawal-request variant | ready for review |
| 4 | #11083 — Burn-first accounting for sweeper fee funding | ready for review |
| 5 | #11086 — Sweeper fee-funding task, with an end-to-end test | Copilot re-review pending, CI green incl. long tests |
| 6 | #11094 — Sweeper funding observability and the prepaid-gas gate | open |
| 7 | #11097 — Adversarial end-to-end coverage of sweeper fee funding | open |


rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,117 @@
...
+//! forwards an arbitrary JSON-RPC payload to every provider, parses each response's `result`
+//! field, and reduces those under the configured consensus strategy. Because the reduction
+//! happens on the parsed `result` — not on the raw response body — differences in how
+//! providers format the surrounding JSON envelope are irrelevant; only the hex quantity is
+//! compared.
+//!
+//! The first consumer is the sweeper address' ETH balance, which *is* the prepaid-sweep-gas
+//! counter (`docs/deposit_from_cex.md`, "Fund the transaction fees without touching the ckETH
+//! backing"): sweeping may only spend ETH that has already been covered by a ckETH burn, and the
+//! sweeper's on-chain balance is what makes that reconcilable.
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in b107e16ec3 — now `rs/ethereum/cketh/docs/deposit_from_cex.md`.

You're right that the short form resolves relative to nothing a reader can follow: there is no `docs/` directory beside either crate, so it only worked if you already knew where the file lived. `balance_scan/batcher` already used the repo-relative form, so this also makes the two consistent.

The same correction applies to five other module docs later in this stack; I've made it in those commits too, so they will already be right when those PRs go up.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,117 @@
...
+pub fn decode_balance(result: &str) -> Result<Wei, DecodeBalanceError> {
+    // `multi_request` hands back the JSON string's *contents*, but tolerate a quoted form too so
+    // a caller passing a raw JSON value through does not silently fail.
+    let quantity = result.trim().trim_matches('"');
+    if !quantity.starts_with("0x") || quantity.len() <= 2 {
+        return Err(DecodeBalanceError::NotAQuantity(result.to_string()));
+    }
+    Wei::from_str_hex(quantity).map_err(|_| {
+        // `from_str_hex` rejects both non-hex digits and values wider than 32 bytes; only the
+        // latter is possible here, since the digits were already accepted above.
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in b107e16ec3.

The tell that the comment was wrong is in the closure it sits in: it re-checks `is_ascii_hexdigit()`, which would be dead code if the digits really had been validated above. Reworded to say the checks establish only the shape of the quantity, which is what they actually do.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,117 @@
...
+    let payload = eth_get_balance_request(address, &block);
+    let result = read_state(rpc_client)
+        .multi_request(payload)
+        .with_cycles(MIN_ATTACHED_CYCLES)
+        .try_send()
+        .await
+        // A balance is compared for exact equality across providers: at a finalized block every
+        // honest provider must return the same quantity, so anything else is a disagreement
+        // worth surfacing rather than papering over by picking one answer.
+        .reduce_with_strategy(StrictMajorityByKey::new(|balance: &String| balance.clone()))
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 9f8e8730f7. This was the most valuable comment of the review — thank you.

Two things compounded here. `StrictMajorityByKey` returns the largest ballot whenever it beats the runner-up, so a 2/1/1 split across four providers wins on two votes, as you say. On top of that, it only ever runs on results the EVM RPC canister has already declared inconsistent — and the canister is configured `Threshold { total: 4, min: 3 }` on mainnet. So the reduction was salvaging an answer from precisely the case the threshold exists to reject.

Rather than add a true-majority strategy, the reduction is now `NoReduction`: the canister's consensus is the policy, and a disagreement stays an error. Failing is cheap here — the read is retried and funding stalls rather than misfires — which is the right direction for a number that gates irreversible burns.

This also subsumes the suppressed comment from the previous round about the key being the raw provider string. An intermediate commit (7e97e4e28c) keyed the majority on the decoded amount so that renderings differing only in hex case would agree; with no client-side reduction there is no key left to compare on, so that commit is superseded. The rendering tolerance it was protecting is still exercised directly against the decoder.

One thing worth separating out: the weakness is not confined to this PR. `StrictMajorityByKey` is also used for fee-history reduction in `tx/mod.rs`, and its name promises more than it delivers — which is how it came to look like a safe default here in the first place. Changing it would alter behaviour beyond this stack, so it is tracked separately as [DEFI-2969](https://dfinity.atlassian.net/browse/DEFI-2969) rather than fixed as a drive-by.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// Reads an `eth_getBalance` result — a hex quantity such as `0x1bc16d674ec80000` — as wei.
+pub fn decode_balance(result: &str) -> Result<Wei, DecodeBalanceError> {
+    // `multi_request` hands back the JSON string's *contents*, but tolerate a quoted form too so
+    // a caller passing a raw JSON value through does not silently fail.
+    let quantity = result.trim().trim_matches('"');
+    if !quantity.starts_with("0x") || quantity.len() <= 2 {
+        return Err(DecodeBalanceError::NotAQuantity(result.to_string()));
+    }
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 9f8e8730f7. Both cases confirmed.

`trim_matches('"')` strips quote characters from each end independently, so `"0x0` became `0x0` and decoded to zero; the quoted form is now parsed as JSON instead. Non-minimal quantities like `0x00` were accepted too, and are now rejected as not being what the protocol specifies.

The direction of this failure is what makes it worth being strict about: a wrong zero reads as "the sweeper is empty" and buys a burn that was never needed, whereas a wrongly rejected answer only delays a retry. My test for that guarantee only covered `""`, `"0x"`, `"null"` and `"latest"`, which is exactly why it missed both of these — it now also covers one-sided quotes, doubled quotes, `0x00` and `0x0000000000000000`.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/// Reads an `eth_getBalance` result — a hex quantity such as `0x1bc16d674ec80000` — as wei.
+pub fn decode_balance(result: &str) -> Result<Wei, DecodeBalanceError> {
+    let trimmed = result.trim();
+    // `multi_request` hands back the JSON string's *contents*, but tolerate a quoted form too so a
+    // caller passing a raw JSON value through does not silently fail. Unquoted by parsing as JSON
+    // rather than by stripping quote characters: stripping also accepts one-sided and repeated
+    // quotes, so `"0x0` would read as a balance of zero.
+    let quantity = serde_json::from_str::<String>(trimmed).unwrap_or_else(|_| trimmed.to_string());
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 66a825c237 — and you're right about the premise, which is the part that mattered.

I traced it to the canister: `multi_request` returns `RawJson`, which is `#[serde(transparent)]` over `String`, so serde deserializes the JSON string and hands back its *contents*. My comment justified the tolerance as letting "a caller passing a raw JSON value through" work, but no such caller can exist on this path. So the tolerance was not accommodating a caller, it was repairing malformed provider answers — into zero, which is the one value that costs money to get wrong here.

Now decoded exactly: no trimming, no unquoting, and quotes or padding are rejected as the provider's own malformation. The quoted and whitespace assertions moved from the "decodes" test to the rejection list, alongside `0x00` and the sign forms.

The module doc also claimed the reduction compared only the hex quantity and that envelope differences were irrelevant. With the client-side reduction gone that described nothing, so it now states what the canister actually does and what this module accepts.

---

On the suppressed comment from this round (line 108, `0x+0` parsing as `Wei::ZERO`): that one does not reproduce. `ethnum`'s parser accepts an optional sign only at position 0, *before* the radix prefix, so `0x+0` reaches the digit loop with `+` and fails; `+0x0` fails the `0x` prefix check. Verified rather than reasoned:

```
"0x+0"    -> Err(NotAQuantity("0x+0"))
"+0x0"    -> Err(NotAQuantity("+0x0"))
"  0x0  " -> Ok(0)      <- the real defect
"\"0x0\""  -> Ok(0)      <- the real defect
```

The underlying instinct was sound, though, so I took it anyway: the digits are now validated *before* parsing rather than inferred from a parse failure. That closes the class by construction instead of relying on the parser's ordering, and leaves "too large" as the only rejection the parser itself can still produce. Both sign forms are pinned as tests.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,130 @@
...
+        // No client-side reduction: the answer is whatever the EVM RPC canister's own consensus
+        // strategy agreed on, and an inconsistent result stays an error.
+        //
+        // Deliberately *not* `StrictMajorityByKey`, despite the name. That strategy returns the
+        // largest ballot whenever it beats the runner-up, so a 2/1/1 split across four providers
+        // wins on two votes — and it only ever runs on results the canister has already declared
+        // inconsistent, i.e. precisely when the configured threshold was not met. Picking a winner
+        // there would quietly settle for fewer providers than the threshold demands, on the number
+        // that decides whether ckETH gets burned.
+        .reduce_with_strategy(NoReduction)
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch on the guarantee being network-dependent — addressed in 451a521169, though the overclaim turned out to live in the PR description rather than the code.

The code said "the configured threshold", which is accurate but vague enough to invite the stronger reading. It now names the numbers in both the module doc and the comment at the call site: 3 of 4 on mainnet, 2 of 4 on Sepolia, so it reads as "the configured threshold agreed" rather than "a majority agreed". The PR description said "a strict majority is required" — left over from before the switch to `NoReduction` — and has been rewritten.

On your first option, a 3-of-4 client for this read specifically: I did consider it and decided against, for three reasons worth recording.

- `rpc_client` is shared by every RPC read the minter makes, so this would be the only call with its own consensus configuration.
- Sepolia is 2-of-4 precisely because its providers are less reliable. Tightening it would make testnet funding flakier, and a stalled funding check is the failure mode this stack works hardest to avoid.
- The asset it would protect is ckSepoliaETH, which is not worth anything. The stricter bar buys nothing on the one network where the numbers differ.

The counter-argument I can see is that Sepolia then does not rehearse the mainnet consensus path exactly. That applies to every RPC read in the minter rather than this one, so if we want it, it belongs in the shared client configuration rather than here.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,137 @@
...
+        // majority agreed". Deliberately the same threshold as every other read the minter makes:
+        // holding this one call to a stricter bar would need its own client, and would buy nothing
+        // on the network where the difference applies, since ckSepoliaETH is not worth anything.
+        //
+        // Deliberately *not* `StrictMajorityByKey`, despite the name. That strategy returns the
+        // largest ballot whenever it beats the runner-up, so a 2/1/1 split across four providers
+        // wins on two votes — and it only ever runs on results the canister has already declared
+        // inconsistent, i.e. precisely when the configured threshold was not met. Picking a winner
+        // there would quietly settle for fewer providers than the threshold demands, on the number
+        // that decides whether ckETH gets burned.
```
gregorydemay: nit: I would remove that whole comment.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,137 @@
...
+    // A JSON-RPC quantity is minimal-length hex: at least one digit, no leading zero unless the
+    // value *is* zero. Checking the digits here rather than inferring them from a parse failure
+    // also keeps `Wei::from_str_hex`'s tolerance for a leading `+` out of reach.
+    if digits.is_empty()
+        || !digits.chars().all(|c| c.is_ascii_hexdigit())
+        || (digits.len() > 1 && digits.starts_with('0'))
+    {
+        return Err(DecodeBalanceError::NotAQuantity(result.to_string()));
+    }
+    // The digits are known good, so the only rejection left is a value too wide for 32 bytes.
```
gregorydemay: nit: why not just return `Wei::from_str_hex(result)` (error mapping is ok)? What do the added checks bring? If they bring anything why not add them to `Wei::from_str_hex` directly?

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// broken provider or a bug on our side — never a balance of zero.
+    Decode(DecodeBalanceError),
+}
+
+/// The native ETH balance of `address` at `block`.
+///
+/// Fails rather than defaulting to zero on any error: a zero balance and "we could not read the
+/// balance" must not be confused, since the caller uses this to decide whether spending is
+/// already covered by a burn.
+pub async fn eth_get_balance(address: &Address, block: BlockTag) -> Result<Wei, GetBalanceError> {
```
gregorydemay: nit: let caller of the method decide which `.reduce_with_strategy(NoReduction)` to attach. I think it would be better to match what the signature on the EVM RPC canister would look like, something like

```rust
pub async fn eth_get_balance(address: &Address, block: BlockTag) -> MultiRpcResult<Wei>
```

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
```diff
@@ -0,0 +1,137 @@
...
+        BlockTag::Finalized => "finalized".to_string(),
+        BlockTag::Safe => "safe".to_string(),
+        BlockTag::Earliest => "earliest".to_string(),
+        BlockTag::Pending => "pending".to_string(),
+        BlockTag::Number(number) => {
+            // A JSON-RPC quantity is minimal-length hex, so strip leading zeros — but never all
+            // of them: block zero is `0x0`, not `0x`.
+            let digits = hex::encode(number.clone().into_be_bytes());
+            let trimmed = digits.trim_start_matches('0');
+            format!("0x{}", if trimmed.is_empty() { "0" } else { trimmed })
```
gregorydemay: Nit: note that `BlockNumber` implements `LowerHex` so one could do something like
```rust
        BlockTag::Number(number) => {
            format!("{:#x}", BlockNumber::from(number.clone()))
        }
```

gregorydemay (approved):
Thanks @mbjorkqvist ! Couple of nits but otherwise LGTM!

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
mbjorkqvist: Done in 0e152a22b4!

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
mbjorkqvist: You're right, and it's now just `Wei::from_str_hex(result)` mapped to an error, in 0e152a22b4.

What the checks caught that the parser doesn't: `ethnum` accepts an optional `+` before the radix prefix (`parse.rs:57-63`), so `+0x0` parsed as zero, and it accepts leading zeros, so `0x00` did too. I had treated both as provider malformation worth rejecting. But that reasoning doesn't survive being asked what it buys: both forms *say* the balance is zero, and zero is what they mean, so rejecting them trades a correct answer for a retry.

They don't belong in `Wei::from_str_hex` either — it parses signature `r`/`s` components and memo amounts, where a leading zero is perfectly ordinary. JSON-RPC quantity minimality is not an integer-parsing rule.

What is still rejected is anything that doesn't state a quantity at all: no `0x` prefix, non-hex digits, embedded spaces, and the quoted or padded forms the canister cannot produce. That keeps the one property the funding task depends on — a failed read must not become a zero balance — and `should_never_read_an_error_as_a_zero_balance` still pins it.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
mbjorkqvist: Taken, in 0e152a22b4. The signature is now

```rust
pub async fn eth_get_balance(address: &Address, block: BlockTag) -> Result<MultiRpcResult<Wei>, IcError>
```

The outer `Result` stays because `try_send` yields one and `ToReducedWithStrategy` is implemented on exactly `Result<MultiRpcResult<T>, IcError>` — so the caller writes `.reduce_with_strategy(NoReduction)` with no adapter in between, the same shape as `balance_scan` and `deposit`. Each provider's string is decoded to `Wei` through `MultiRpcResult::and_then`, so a provider that answers with something other than a quantity fails as that provider's error and consensus is compared on amounts rather than renderings.

Two consequences worth naming. `GetBalanceError` and `DecodeBalanceError` are gone: a decode failure is now an `RpcError::ValidationError`, which is what the canister would have reported for the same thing, and the caller had no use for the distinction — it logs and retries either way. And the choice of how much agreement is enough now sits at the call site in #11086, which is where the comment explaining it moved too.

rs/ethereum/cketh/minter/src/eth_rpc_client/get_balance.rs
mbjorkqvist: Applied in 0e152a22b4, with one adjustment:

```rust
BlockTag::Number(number) => format!("0x{:x}", BlockNumber::from(number.clone())),
```

`{:#x}` doesn't work here. `CheckedAmountOf`'s impl is `write!(f, "{:x}", self.0)` (`checked_amount.rs:231`), which formats the inner value with a fresh spec and so drops the alternate flag — `{:#x}` prints `ff`, not `0xff`. Verified with a standalone repro rather than assumed. Writing the `0x` out gets the same result and still removes the `hex::encode` plus `trim_start_matches` dance, since `{:x}` is already minimal-length and renders block zero as `0`. `should_render_a_block_number_as_a_minimal_length_quantity` covers it unchanged.

PR #11061, opened 2026-08-07
chore(system-tests): lower DEFAULT_MEMORY_KIB_PER_VM from 24GiB to 4GiB

basvandijk:
# What

Lowers the default memory of a system-test VM from 24GiB to 4GiB and drops the custom overrides in certain system-tests for the `SystemTestBackend::Local` backend that made the same change.

# Why

We would like to pack more system-tests on a single RBE worker (64GB RAM). It would therefor be better if the VMs used less memory. Some system-tests already lowered the VM RAM to 4GiB. This change lowers it by default.

All system-tests (with `CI_ALL_BAZEL_TARGETS`) run fine and don't flake with this change across multiple runs.


pierugo-dfinity (approved):

nmattia (approved):
Very nice!

PR #11062, opened 2026-08-07
feat: prettify bytes ourselves

nmattia:
This drops the [pretty-bytes](https://crates.io/crates/pretty-bytes) dependency (last update 9 years ago). Instead we now make the bytes pretty ourselves.

This cleans up our dependency graph and closes the following (low severity) vulnerability: https://github.com/advisories/GHSA-g98v-hv3f-hcfr


basvandijk (approved):

rs/sns/integration_tests/test_canisters/sns_governance_mem_test_canister.rs
```diff
@@ -131,8 +130,30 @@ fn proposal_action_iterator() -> impl Iterator<Item = u64> {
...
+    const UNITS: [&str; 9] = ["B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
+
+    let bytes = bytes as f64;
+    if bytes < 1_f64 {
+        return format!("{} B", bytes);
+    }
+
+    let delimiter = 1000_f64;
+    let exponent = std::cmp::min(
+        (bytes.ln() / delimiter.ln()).floor() as i32,
```
daniel-wong-dfinity-org-twin, line 148: > (bytes.ln() / delimiter.ln())

This is a strange way to say "divide the log base 10 by 3". At the very least, let's comment it as such.

rs/sns/integration_tests/test_canisters/sns_governance_mem_test_canister.rs
```diff
@@ -131,8 +130,30 @@ fn proposal_action_iterator() -> impl Iterator<Item = u64> {
...
+        (UNITS.len() - 1) as i32,
+    );
+
+    // Round to 2 decimals, then re-parse to drop any trailing zeros.
+    let value = format!("{:.2}", bytes / delimiter.powi(exponent))
+        .parse::<f64>()
+        .unwrap();
+    let unit = UNITS[exponent as usize];
+
+    format!("{} {}", value, unit)
```
daniel-wong-dfinity-org-twin: I think a more straightforward way is this:


```suggestion
    let value = format!("{:.2}", bytes / delimiter.powi(exponent))
        .trim_end_matches('0')
        .trim_end_matches('.');

    let unit = UNITS[exponent as usize];

    format!("{} {}", value, unit)
```

To me (and I think most non-Einsteinian humans), it is not clear how the round-trip method works, just like how it is surprising that (the result of) `0.1 + 0.2` gets formatted as "0.30000000000000004".

rs/sns/integration_tests/test_canisters/sns_governance_mem_test_canister.rs
```diff
@@ -131,8 +130,30 @@ fn proposal_action_iterator() -> impl Iterator<Item = u64> {
...
+/// Uses SI units (base 1000) and rounds to at most 2 decimal places, dropping
+/// any trailing zeros (e.g. 4_000_000 -> "4 MB", not "4.00 MB").
 fn pretty_bytes(bytes: usize) -> String {
-    converter::convert(bytes as f64)
+    const UNITS: [&str; 9] = ["B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
+
+    let bytes = bytes as f64;
+    if bytes < 1_f64 {
+        return format!("{} B", bytes);
+    }
```
daniel-wong-dfinity-org-twin: This is a strange way to implement special handling for 0.

```suggestion
    // This avoids later doing 0_f64.ln(), which (in math) has NO result.
    if bytes == 0 {
        return format!("0 B", bytes);
    }
    let bytes = bytes as f64;
```

Ideally, `0` would occur in the code.

If I want to discuss 2, I do not refer to it as "the only even prime". Rather, I say "2".

daniel-wong-dfinity-org-twin (approved):

PR #11064, opened 2026-08-07
refactor(consensus): introduce a builder pattern for dependencies in consensus' unit tests

pierugo-dfinity:
Consensus' (fake) dependencies in unit tests are built with functions like `dependencies`, `dependencies_with_subnet_params`, or `dependencies_with_subnet_records_with_raw_state_manager` that live in `ic-consensus-mocks`. Each time a developer wants to consistently modify a dependency, they should add such a function and then wire it through the others. This becomes hard to maintain and a Builder pattern is more than appropriate for this use-case. This PR removes these functions and replaces all call sites to build their dependencies through the builder. The relation is as follows:
- Old `dependencies_with_subnet_params(...)` becomes `DependenciesBuilder::single_subnet(...).build()`
- Old `dependencies(...)` becomes `DependenciesBuilder::new(...).build()`, which is a simple shortcut to `single_subnet`
- Old `dependencies_with_subnet_records_with_raw_state_manager(...)` becomes `DependenciesBuilder::single_subnet(...).without_mocked_state_manager().build()`

The two above constructors then fall down to the `multiple_subnets` constructor, which is for now otherwise unused (as Consensus usually stays scoped only to a single subnet) but will soon be for incoming subnet-splitting unit tests.

One particularly frequent setup is
```rust
let committee: Vec<_> = (0..no_nodes).map(node_test_id).collect();
dependencies_with_subnet_params(
    pool_config,
    subnet_test_id(0),
    vec![(
        1,
        SubnetRecordBuilder::from(&committee)
            .with_dkg_interval_length(interval_length)
            .build(),
    )],
);
```
which uses `...with_subnet_params` only just to modify the DKG interval length. `DependenciesBuilder` thus provides `.with_dkg_interval_length` such that such a setup boils down to a simple
```rust
DependenciesBuilder::new(pool_config, no_nodes)
    .with_dkg_interval_length(interval_length)
    .build()
```
I replaced such setups with this even when the subnet was not `subnet_test_id(0)` or the initial registry version was not 1 when it clearly did not matter. Otherwise, I kept them intact.


pierugo-dfinity:
@frankdavid Maybe this PR could be of interest to you, following our discussion

rs/consensus/chain_key/src/lib.rs
```diff
@@ -804,11 +802,13 @@ mod tests {
...
                 registry_data_provider,
                 ..
-            } = dependencies_with_subnet_records_with_raw_state_manager(
+            } = DependenciesBuilder::single_subnet(
                 pool_config,
                 subnet_id,
                 vec![(1, subnet_record_builder.build())],
-            );
+            )
+            .without_mocked_state_manager()
```
eichhorl: I suppose the state manager is still mocked but doesn't have expectations, so maybe

```suggestion
            .without_state_manager_expectations()
```

Then also `with_state_manager_expectations` for the opposite

rs/consensus/src/consensus/share_aggregator.rs
```diff
@@ -186,13 +186,13 @@ fn to_messages<T: ConsensusMessageHashable>(artifacts: Vec<T>) -> Vec<ConsensusM
...
     use super::*;
-    use ic_consensus_mocks::{Dependencies, dependencies, dependencies_with_subnet_params};
+    use ic_consensus_mocks::{Dependencies, DependenciesBuilder};
     use ic_interfaces::consensus_pool::ConsensusPool;
     use ic_logger::replica_logger::no_op_logger;
     use ic_test_utilities::message_routing::FakeMessageRouting;
     use ic_test_utilities_consensus::fake::{FakeContentSigner, FakeSigner};
-    use ic_test_utilities_registry::SubnetRecordBuilder;
-    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id};
+
```
eichhorl:
```suggestion
```

rs/consensus/dkg/src/lib.rs
eichhorl, line 794: Is it still node Id 1?

rs/consensus/src/consensus/catchup_package_maker.rs
```diff
@@ -312,8 +309,8 @@ mod tests {
             fake_signature_request_context_with_registry_version,
         },
     };
-    use ic_test_utilities_registry::SubnetRecordBuilder;
-    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id};
+
```
eichhorl:
```suggestion
```

eichhorl (approved):

rs/consensus/chain_key/src/lib.rs
pierugo-dfinity: True, thanks https://github.com/dfinity/ic/pull/11064/commits/343edfa15f0dd4fe687c606cd5ecbfa17fd33b38

rs/consensus/src/consensus/share_aggregator.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/11064/commits/49e5b9af1cbede0d08d6b0e040ba786b6154d984

rs/consensus/src/consensus/catchup_package_maker.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/11064/commits/49e5b9af1cbede0d08d6b0e040ba786b6154d984

rs/consensus/dkg/src/lib.rs
pierugo-dfinity: A-ha indeed. I reverted this to keep using `node_test_id(1)` staying consistent with the other unit tests https://github.com/dfinity/ic/pull/11064/commits/54eee4e67bf20d42ca6990943b438e4d8644c988

PR #11065, opened 2026-08-07
feat(cketh): burn ckETH from the minter's own fee subaccount

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding). #11060 has merged, so this now targets `master`.

## Why

Sweep gas has to be paid for without touching the ETH that backs ckETH 1:1. It is prepaid instead: ckETH is burned from the minter's own fee subaccount — the account the ckETH ledger credits transaction fees to — before any ETH moves.

The minter cannot do that today. Its existing burn names the minter's *default* subaccount as the ICRC-2 spender, and the ledger waives the allowance check only when the spender equals the `from` account compared as a **full** account. `{minter, None}` is a different account from `{minter, 0fee}`, so burning from the fee subaccount that way is rejected with `InsufficientAllowance`.

## What

Adds a burn that names the same subaccount as both `from` and spender, so no allowance is involved, alongside the constant identifying the fee subaccount. The existing user-withdrawal burn keeps its behaviour and signature; both now share a private helper, so the error mapping is not duplicated.

## Coverage

Deliberately none in this PR.

An earlier revision added a ckETH integration test, but it built the transfer arguments itself and called the ledger directly, so what it pinned was ledger behaviour rather than the new burn. That rule now lives where it applies to every ledger, in #11139 ([DEFI-2978](https://dfinity.atlassian.net/browse/DEFI-2978)) — which promptly turned up that the ICP ledger rejects such a burn while ICRC accepts it ([DEFI-2979](https://dfinity.atlassian.net/browse/DEFI-2979)), a divergence a ckETH-local test could never have shown.

The new burn is unreachable until #11086 wires it into the funding task, whose end-to-end test drives it through the minter and asserts both the supply drop and the fee-account debit. So it ships untested-and-uncalled for a few PRs, then is covered through the real path.

## Stack

| # | PR | Status |
|---|----|--------|
| 1 | #11060 — Read a native ETH balance via the EVM RPC canister | merged |
| 2 | Burn ckETH from the minter's own fee subaccount | **this PR** |
| 3 | #11072 — Add the SweeperFunding withdrawal-request variant | ready for review |
| 4 | #11083 — Burn-first accounting for sweeper fee funding | ready for review |
| 5 | #11086 — Sweeper fee-funding task, with an end-to-end test | draft |
| 6 | #11094 — Sweeper funding observability and the prepaid-gas gate | draft |
| 7 | #11097 — Adversarial end-to-end coverage of sweeper fee funding | draft |

Related, outside the stack: #11139 covers the ICRC-2 rule this PR relies on, for the ICP ledger and both ICRC token types.

[DEFI-2978]: https://dfinity.atlassian.net/browse/DEFI-2978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/ethereum/cketh/minter/tests/fee_account.rs
```diff
@@ -0,0 +1,183 @@
+use candid::{Decode, Encode, Nat};
```
gregorydemay: Why create a new integration test file and a new integration test setup? Let's reuse what we have if possible.

rs/ethereum/cketh/minter/tests/cketh.rs
```diff
@@ -1344,3 +1344,87 @@ mod cketh_evm_rpc {
         }
     }
 }
+
+/// The fee subaccount is the minter's own, so burning from it is a self-spend: the ledger accepts
+/// `{minter, 0fee}` as the spender without an allowance. These two tests pin that, since the
+/// sweeper funding in `ledger_client` depends on it.
```
gregorydemay: nit: 

1. isn't this testing more the ledger than the minter?
2. Maybe better to combine both tests to have a single `CkEthSetup`

gregorydemay (approved):

rs/ethereum/cketh/minter/tests/fee_account.rs
mbjorkqvist: Your comment started a chain that ends with this PR going away, so it is worth laying out where everything landed.

Reusing `CkEthSetup` made the test cheap enough that its real weakness became visible: it never exercised `burn_from_own_subaccount`. It built the `TransferFromArgs` itself and called the ledger directly as the minter, so what it pinned was ledger behaviour — that a spender may spend from an account only when it names that account's own subaccount — and not a line of the minter code this PR adds.

That rule now lives where it applies to every ledger, in #11139 ([DEFI-2978](https://dfinity.atlassian.net/browse/DEFI-2978)). Moving it there paid for itself immediately: the ICP ledger *rejects* such a burn, because `Operation::Burn` checks an allowance for any spender while exempting the self-spend only when consuming one. Both behaviours are now pinned per ledger, so closing that gap has to flip an assertion rather than pass unnoticed ([DEFI-2979](https://dfinity.atlassian.net/browse/DEFI-2979)).

Without that test, what remained here was 33 lines nothing reaches: `CKETH_FEE_SUBACCOUNT` is unused in this branch, and `burn_from_own_subaccount` has no caller until PR 5. So rather than merge dead code, each piece moves to where it is first needed:

- **`CKETH_FEE_SUBACCOUNT` → #11072**, whose `state/transactions/tests.rs` is its first use.
- **The whole `ledger_client.rs` change → #11086**, alongside its only caller in `sweeper.rs` and the end-to-end test that drives the burn through the minter and asserts the supply drop and fee-account debit.

I also considered keeping this PR as just the `burn_from_with_spender` extraction. Decided against: with the new burn gone, that parameter has one caller always passing `None`, and its justification would live in a PR you cannot see from here. Better to review the seam once, with both callers visible.

Nothing loses coverage in the move. The refactor is already exercised by the existing withdrawal tests, which assert the burn block the real ledger recorded — `spender: Some({minter, None})` with `fee: None` and `created_at_time: None`, in `cketh.rs` and three places in `ckerc20.rs` — plus the three error paths (`InsufficientAllowance`, `InsufficientFunds`, `AmountTooLow`) that run through the same `match`.

I will close this PR once you have seen this. Thanks for pushing on the test reuse — the ICP divergence would not have surfaced otherwise.

gregorydemay: Sounds good to me, thanks!

PR #11069, opened 2026-08-07
feat: migrate to ic-cdk 0.20

nmattia:
This builds on recent work and completes the migration to ic-cdk 0.20 (0.20.2 to be precise).

This introduces a dependency on the new (extracted) crate `ic-cdk-management-canister`.

Important note: the `GetUtxosRequest` exported by `bitcoin_canister` does _not_ implement `Clone` or `Ord` anymore, though we rely on it for BTreeMap ordering. To avoid breaking any sort of compatibility, we wrap this new type in a newtype and have a custom implementation with the same behavior as the old type.


nmattia:
This requires https://github.com/dfinity/ic/pull/11055 to merge first

rs/recovery/subnet_splitting/tests/integration_tests.rs
```diff
@@ -196,16 +196,25 @@ fn load_metrics_e2e_test() {
...
         // orientation is determined and checked for consistency by the exact `assert_eq_oriented`
         // checks below, and these `assert_near` checks pass in either orientation.
-        assert_near!(states_sizes_bytes.source, 4987773, 0.1);
-        assert_near!(states_sizes_bytes.destination, 4220431, 0.1);
-        assert_near!(instructions_executed.source, 145689972, 0.1);
-        assert_near!(instructions_executed.destination, 144345860, 0.1);
+        assert_near!(states_sizes_bytes.source, 5009842, 0.5);
+        assert_near!(states_sizes_bytes.destination, 4235402, 0.5);
+        assert_near!(instructions_executed.source, 145699596, 0.1);
+        assert_near!(instructions_executed.destination, 144352276, 0.1);
```
alin-at-dfinity: I'm guessing that the "varies across [...] platforms" applies here, as the change between the previous and the new values is well within 10% in every case, far from requiring a change to a 50% margin for state sizes.

alin-at-dfinity (approved):
Core protocol changes LGTM.

michael-weigelt (approved):

blind-oracle (approved):

rs/bitcoin/ckbtc/minter/src/lib.rs
```diff
@@ -7,8 +7,8 @@ use crate::updates::update_balance::UpdateBalanceError;
 use async_trait::async_trait;
 use candid::{CandidType, Deserialize, Principal};
 use canlog::log;
-use ic_cdk::bitcoin_canister;
-use ic_cdk::management_canister::SignWithEcdsaArgs;
+use ic_cdk_bitcoin_canister as bitcoin_canister;
```
gregorydemay, line 10: my impression is that `ic-cdk-bitcoin-canister` is not needed at all in ckbtc, we can use `ic-btc-interface`, which we already depend upon.

rs/dogecoin/ckdoge/minter/BUILD.bazel
```diff
@@ -28,6 +28,8 @@ rust_library(
         "@crate_index//:ciborium",
         "@crate_index//:hex",
         "@crate_index//:ic-cdk",
+        "@crate_index//:ic-cdk-bitcoin-canister",
```
gregorydemay, line 31: also here it would be better to just use `ic-btc-interface` (does not depend on cdk)

gregorydemay (commented):
Thanks @nmattia !

NikolaMilosa (approved):

basvandijk (approved):

rs/nns/handlers/root/impl/src/canister_management.rs
```diff
@@ -160,10 +160,11 @@ async fn try_to_create_and_install_canister(
...
         sender_canister_version: Some(ic_cdk::api::canister_version()),
     };
-    let install_res: Result<(), (RejectionCode, String)> =
-        call(CanisterId::ic_00().get().0, "install_code", (install_args,)).await;
-
-    install_res.map_err(|(code, msg)| format!("error code {}: {}", code as i32, msg))?;
+    Call::unbounded_wait(CanisterId::ic_00().get().0, "install_code")
+        .with_arg(install_args)
+        .await
+        .map(|_reply| ())
```
daniel-wong-dfinity-org-twin: Ditto.

(Again, I do not hate the new behavior, but it seems easy enough to get close to the original behavior, so unless the old behavior was broken, let's just stick with it.)

rs/nns/handlers/root/impl/src/canister_management.rs
```diff
@@ -333,6 +334,26 @@ pub async fn create_canister_and_install_code(
...
+        IcCdkCallError::CallRejected(rejected) => (
+            rejected.raw_reject_code() as i32,
+            rejected.reject_message().to_string(),
+        ),
+        // A response that cannot be decoded means the callee misbehaved.
+        IcCdkCallError::CandidDecodeFailed(err) => {
+            (RejectCode::CanisterError as i32, err.to_string())
+        }
+        // The call never left this canister, so it may well succeed if retried.
+        err => (RejectCode::SysTransient as i32, err.to_string()),
```
daniel-wong-dfinity-org-twin: Ditto.

rs/nns/handlers/root/impl/src/canister_management.rs
```diff
@@ -333,6 +334,26 @@ pub async fn create_canister_and_install_code(
     CreateCanisterAndInstallCodeResponse::from(result)
 }
 
+/// Formats a failed call as the `error code {code}: {message}` string that
+/// [`try_to_create_and_install_canister`] reports. Reports the same numeric codes
+/// that the deprecated `ic_cdk::api::call` did.
+fn format_call_error(err: IcCdkCallError) -> String {
```
daniel-wong-dfinity-org-twin, line 341: I think you said you would take care of this in the future, but just a reminder: let's avoid (near) copy & paste.

rs/nns/handlers/root/impl/src/root_proposals.rs
```diff
@@ -467,15 +464,25 @@ async fn get_nns_subnet_id() -> Result<SubnetId, String> {
...
+        Call::unbounded_wait(REGISTRY_CANISTER_ID.get().0, "get_subnet_for_canister")
+            .with_arg(request)
+            .await
+            .map_err(IcCdkCallError::from)
+            .and_then(|response| {
+                response
+                    .candid::<Result<SubnetForCanister, String>>()
+                    .map_err(IcCdkCallError::from)
+            })
+            .map_err(|err| {
```
daniel-wong-dfinity-org-twin, line 477: This seems like another spot where we might be able to use a shared converter.

rs/nns/handlers/root/interface/src/client.rs
```diff
@@ -45,31 +44,54 @@ impl NnsRootCanisterClient for NnsRootCanisterClientImpl {
...
+                    .candid::<CanisterStatusResult>()
+                    .map_err(IcCdkCallError::from)
+            })
+            .map_err(map_call_error)
     }
 }
 
+/// Translates a failed call into the `(reject code, message)` pair that
+/// [`NnsRootCanisterClientImpl`] reports to its callers.
+fn map_call_error(err: IcCdkCallError) -> (Option<i32>, String) {
```
daniel-wong-dfinity-org-twin: Ditto.

rs/nns/handlers/root/interface/src/client.rs
```diff
@@ -45,31 +44,54 @@ impl NnsRootCanisterClient for NnsRootCanisterClientImpl {
...
+        IcCdkCallError::CallRejected(rejected) => (
+            rejected.raw_reject_code() as i32,
+            rejected.reject_message().to_string(),
+        ),
+        // A response that cannot be decoded means the callee misbehaved.
+        IcCdkCallError::CandidDecodeFailed(err) => {
+            (RejectCode::CanisterError as i32, err.to_string())
+        }
+        // The call never left this canister, so it may well succeed if retried.
+        err => (RejectCode::SysTransient as i32, err.to_string()),
```
daniel-wong-dfinity-org-twin: Ditto.

daniel-wong-dfinity-org-twin (approved):

rs/nns/handlers/root/impl/src/canister_management.rs
nmattia: Refers to: https://github.com/dfinity/ic/pull/11055#discussion_r3735612079

rs/bitcoin/ckbtc/minter/src/lib.rs
nmattia: let's handle this separately!

rs/dogecoin/ckdoge/minter/BUILD.bazel
nmattia: let's handle this separately!

rs/bitcoin/ckbtc/minter/src/lib.rs
gregorydemay: it's actually needed because of the various cost methods. Keep as is.

rs/dogecoin/ckdoge/minter/BUILD.bazel
gregorydemay: it's actually needed because of the various cost methods. Keep as is.

gregorydemay (approved):

PR #11072, opened 2026-08-07
feat(cketh): add the SweeperFunding withdrawal-request variant

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding). Now targets `master`: #11060 has merged, and #11065 is being closed with its contents folded into the PRs that use them — the fee-subaccount constant into this one, the burn helper into #11086 alongside its caller.

## Why

Sweeper fee funding is mechanically an ordinary ckETH withdrawal — same nonce sequence, same threshold-ECDSA signing, same fee-bumped resubmission — so it becomes a third `WithdrawalRequest` variant rather than a parallel pipeline.

It differs in exactly one respect, and that difference is what the whole feature turns on: the ckETH burned for funding is **never** re-minted. A funding request must therefore never reach the reimbursement machinery.

## What

Three places enforce that, all of which would otherwise fail only at runtime:

- `maybe_reimburse` is the double-minting guard, and `record_reimbursement_request` asserts membership has been cleared before minting. Funding is kept out of the set on insert, and the corresponding assertion on removal is made conditional. Both are production assertions, so a missed branch traps the canister.
- The conversion to a reimbursement index becomes fallible, deliberately: a fallible conversion makes the compiler prove at every call site that funding cannot produce an index, rather than relying on a panicking arm that traps if a site is missed. The two callers construct their index inside the reimbursable arms instead.
- Nothing is recorded to pay back if a funding transaction finalizes with a failure receipt — an outcome that requires code at the destination, so it cannot arise for a bare transfer to the sweeper, and is logged as unexpected rather than handled. The ckETH stays burned while no ETH reaches the sweeper — only the gas the failed transaction still paid leaves the main address — so the rest of the burn over-backs ckETH instead of anything being lost, and neither a reimbursement path nor a second event type is needed.

Everything else follows ckETH: the 21'000 gas limit of a plain value transfer to a code-less address, a resubmission strategy ceilinged at the burned amount — so a climbing gas price shrinks the ETH delivered to the sweeper rather than spending more than was burned — and a fee carved out of that same amount, so balance accounting needs no change.

Since reimbursement is the only difference, the variant carries an `EthWithdrawalRequest` — the same payload a user's ckETH withdrawal carries — rather than a near-copy of it, and the spec drops the one piece of funding accounting that had no counterpart in the withdrawal flow: burned-but-unspent amounts are no longer credited against the next funding's burn. They stay as backing, just like the unspent gas a user's withdrawal leaves behind.

Funding appears in the withdrawal status endpoint, the dashboard and the event log rather than being hidden: it moves ckETH-denominated value and is a public, auditable action. A user query never matches one, since the sender is the minter itself.

The new event takes tag 27; 26 went to `AutomaticDepositReceived` on master while this branch was open, and the tags are the durable CBOR encoding, so they cannot collide.

## Stack

| # | PR | Status |
|---|----|--------|
| 1 | Add the SweeperFunding withdrawal-request variant | **this PR** |
| 2 | #11083 — Burn-first accounting for sweeper fee funding | ready for review |
| 3 | #11086 — Sweeper fee-funding task, with an end-to-end test | draft |
| 4 | #11094 — Sweeper funding observability and the prepaid-gas gate | draft |
| 5 | #11097 — Adversarial end-to-end coverage of sweeper fee funding | draft |

### DIFF
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -273,8 +273,6 @@ type TxFinalizedStatus = variant {
     };
     // Transaction failed and will be reimbursed,
     PendingReimbursement : EthTransaction;
-    // Transaction failed and will not be reimbursed.
-    Failed : EthTransaction;
 };
 
 // Retrieve the status of a withdrawal request.
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -273,6 +273,8 @@ type TxFinalizedStatus = variant {
     };
     // Transaction failed and will be reimbursed,
     PendingReimbursement : EthTransaction;
+    // Transaction failed and will not be reimbursed.
+    Failed : EthTransaction;
 };
 
 // Retrieve the status of a withdrawal request.
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -575,7 +575,7 @@ type Event = record {
             ledger_burn_index : nat;
             from : principal;
             from_subaccount : opt blob;
-            created_at: nat64;
+            created_at: opt nat64;
         };
         CreatedTransaction : record {
             withdrawal_id : nat;
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -570,6 +570,16 @@ type Event = record {
             from_subaccount : opt blob;
             created_at: opt nat64;
         };
+        // The minter burned ckETH from its fee subaccount to prepay gas for sweeping,
+        // and is sending that ETH to its dedicated sweeper address.
+        AcceptedSweeperFundingRequest : record {
+            withdrawal_amount : nat;
+            destination : text;
+            ledger_burn_index : nat;
+            from : principal;
+            from_subaccount : opt blob;
+            created_at: nat64;
+        };
         CreatedTransaction : record {
             withdrawal_id : nat;
             transaction : UnsignedTransaction;
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -242,7 +242,6 @@ type MinterInfo = record {
     evm_rpc_id : opt principal;
 };
 
-
 type GasFeeEstimate = record {
     // Maximum amount of Wei per gas unit that the transaction is willing to pay in total.
     // This covers the base fee determined by the network and the `max_priority_fee_per_gas`.
@@ -570,8 +569,6 @@ type Event = record {
             from_subaccount : opt blob;
             created_at: opt nat64;
         };
-        // The minter burned ckETH from its fee subaccount to prepay gas for sweeping,
-        // and is sending that ETH to its dedicated sweeper address.
         AcceptedSweeperFundingRequest : record {
             withdrawal_amount : nat;
             destination : text;
--- a/rs/ethereum/cketh/minter/src/dashboard/tests.rs
+++ b/rs/ethereum/cketh/minter/src/dashboard/tests.rs
@@ -884,12 +884,16 @@ fn should_display_reimbursed_requests() {
                                 },
                             );
                         }
+                        WithdrawalRequest::SweeperFunding(_) => {
+                            unreachable!("sweeper funding is never reimbursed")
+                        }
                     }
                 } else {
                     apply_state_transition(
                         &mut state,
                         &EventType::QuarantinedReimbursement {
-                            index: ReimbursementIndex::from(&req),
+                            index: ReimbursementIndex::try_from(&req)
+                                .expect("BUG: this test's fixtures are all user withdrawals"),
                         },
                     )
                 }
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -366,15 +366,11 @@ impl State {
             .get_processed_withdrawal_request(withdrawal_id)
             .expect("BUG: missing withdrawal request");
         let charged_tx_fee = match withdrawal_request {
-            WithdrawalRequest::CkEth(req) => req
+            WithdrawalRequest::CkEth(req) | WithdrawalRequest::SweeperFunding(req) => req
                 .withdrawal_amount
                 .checked_sub(tx.transaction().amount)
                 .expect("BUG: withdrawal amount MUST always be at least the transaction amount"),
             WithdrawalRequest::CkErc20(req) => req.max_transaction_fee,
-            WithdrawalRequest::SweeperFunding(req) => req
-                .withdrawal_amount
-                .checked_sub(tx.transaction().amount)
-                .expect("BUG: funded amount MUST always be at least the transaction amount"),
         };
         let unspent_tx_fee = charged_tx_fee.checked_sub(tx_fee).expect(
             "BUG: charged transaction fee MUST always be at least the effective transaction fee",
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -371,10 +371,6 @@ impl State {
                 .checked_sub(tx.transaction().amount)
                 .expect("BUG: withdrawal amount MUST always be at least the transaction amount"),
             WithdrawalRequest::CkErc20(req) => req.max_transaction_fee,
-            // As for ckETH, the fee was carved out of the burned amount. The accounting below then
-            // debits `eth_balance` by only the effective fee when the transaction failed, which is
-            // exactly right for funding too: the ETH stayed at the main address while the ckETH was
-            // burned, leaving ckETH over-backed rather than under-backed.
             WithdrawalRequest::SweeperFunding(req) => req
                 .withdrawal_amount
                 .checked_sub(tx.transaction().amount)
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -371,6 +371,14 @@ impl State {
                 .checked_sub(tx.transaction().amount)
                 .expect("BUG: withdrawal amount MUST always be at least the transaction amount"),
             WithdrawalRequest::CkErc20(req) => req.max_transaction_fee,
+            // As for ckETH, the fee was carved out of the burned amount. The accounting below then
+            // debits `eth_balance` by only the effective fee when the transaction failed, which is
+            // exactly right for funding too: the ETH stayed at the main address while the ckETH was
+            // burned, leaving ckETH over-backed rather than under-backed.
+            WithdrawalRequest::SweeperFunding(req) => req
+                .withdrawal_amount
+                .checked_sub(tx.transaction().amount)
+                .expect("BUG: funded amount MUST always be at least the transaction amount"),
         };
         let unspent_tx_fee = charged_tx_fee.checked_sub(tx_fee).expect(
             "BUG: charged transaction fee MUST always be at least the effective transaction fee",
--- a/rs/ethereum/cketh/minter/src/state/event.rs
+++ b/rs/ethereum/cketh/minter/src/state/event.rs
@@ -5,7 +5,7 @@ use crate::lifecycle::{init::InitArg, upgrade::UpgradeArg};
 use crate::numeric::{BlockNumber, Erc20Value, LedgerBurnIndex, LedgerMintIndex};
 use crate::state::transactions::{
     Erc20WithdrawalRequest, EthWithdrawalRequest, Reimbursed, ReimbursementIndex,
-    ReimbursementRequest,
+    ReimbursementRequest, SweeperFundingRequest,
 };
 use crate::timed_sized_map::Timestamp;
 use crate::tx::{Eip1559TransactionRequest, SignedEip1559TransactionRequest};
@@ -182,6 +182,9 @@ pub enum EventType {
     /// durable even across an ungraceful trap (unlike the pre-upgrade snapshot).
     #[n(26)]
     AutomaticDepositReceived(#[n(0)] AutomaticDeposit),
+    /// The minter burned ckETH from its fee subaccount to top up the sweeper address with gas.
+    #[n(27)]
+    AcceptedSweeperFundingRequest(#[n(0)] SweeperFundingRequest),
 }
 
 /// Full snapshot of the ckERC20 deposit address registry. Carries the limits in
--- a/rs/ethereum/cketh/minter/src/state/event.rs
+++ b/rs/ethereum/cketh/minter/src/state/event.rs
@@ -5,7 +5,7 @@ use crate::lifecycle::{init::InitArg, upgrade::UpgradeArg};
 use crate::numeric::{BlockNumber, Erc20Value, LedgerBurnIndex, LedgerMintIndex};
 use crate::state::transactions::{
     Erc20WithdrawalRequest, EthWithdrawalRequest, Reimbursed, ReimbursementIndex,
-    ReimbursementRequest, SweeperFundingRequest,
+    ReimbursementRequest,
 };
 use crate::timed_sized_map::Timestamp;
 use crate::tx::{Eip1559TransactionRequest, SignedEip1559TransactionRequest};
@@ -184,7 +184,7 @@ pub enum EventType {
     AutomaticDepositReceived(#[n(0)] AutomaticDeposit),
     /// The minter burned ckETH from its fee subaccount to top up the sweeper address with gas.
     #[n(27)]
-    AcceptedSweeperFundingRequest(#[n(0)] SweeperFundingRequest),
+    AcceptedSweeperFundingRequest(#[n(0)] EthWithdrawalRequest),
 }
 
 /// Full snapshot of the ckERC20 deposit address registry. Carries the limits in
--- a/rs/ethereum/cketh/minter/src/state/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/tests.rs
@@ -1348,6 +1348,49 @@ fn state_equivalence() {
     );
 }
 
+mod sweeper_funding {
+    use super::*;
+    use crate::eth_logs::LedgerSubaccount;
+    use crate::numeric::{LedgerBurnIndex, Wei};
+    use crate::state::audit::apply_state_transition;
+    use crate::state::transactions::{EthWithdrawalRequest, WithdrawalRequest};
+    use crate::test_fixtures::initial_state;
+    use assert_matches::assert_matches;
+
+    #[test]
+    fn should_record_an_accepted_funding_as_a_funding_request() {
+        let mut state = initial_state();
+        let funding = EthWithdrawalRequest {
+            withdrawal_amount: Wei::new(10_000_000_000_000_000),
+            destination: "0x5353535353535353535353535353535353535353"
+                .parse()
+                .unwrap(),
+            ledger_burn_index: LedgerBurnIndex::new(0),
+            from: "k2t6j-2nvnp-4zjm3-25dtz-6xhaa-c7boj-5gayf-oj3xs-i43lp-teztq-6ae"
+                .parse()
+                .unwrap(),
+            from_subaccount: LedgerSubaccount::from_bytes(crate::CKETH_FEE_SUBACCOUNT),
+            created_at: Some(1699527697000000000),
+        };
+
+        apply_state_transition(
+            &mut state,
+            &EventType::AcceptedSweeperFundingRequest(funding),
+        );
+
+        let request = state
+            .eth_transactions
+            .withdrawal_requests_iter()
+            .next()
+            .expect("BUG: the funding request was not recorded");
+        assert_matches!(request, WithdrawalRequest::SweeperFunding(_));
+        assert!(
+            !request.is_reimbursable(),
+            "recorded as an ordinary withdrawal, a failed funding would be reimbursed"
+        );
+    }
+}
+
 mod eth_balance {
     use super::*;
     use crate::eth_rpc_client::responses::{TransactionReceipt, TransactionStatus};
--- a/rs/ethereum/cketh/minter/src/state/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/tests.rs
@@ -1524,6 +1524,108 @@ mod eth_balance {
         );
     }
 
+    /// Funding takes its own arm in the balance accounting, so the ckETH and ckERC20 cases above
+    /// cannot reach it. What is asserted is the same shape as a ckETH withdrawal — a success debits
+    /// the transferred ETH plus the fee actually paid, a failure debits only that fee — because
+    /// funding is an ordinary withdrawal to the accounting. What differs is that nothing is ever
+    /// reimbursed, which is why the failing case must still leave the fee counters moving.
+    #[test]
+    fn should_update_after_successful_and_failed_sweeper_funding() {
+        let mut state_before_funding = initial_state();
+        apply_state_transition(
+            &mut state_before_funding,
+            &EventType::AcceptedDeposit(received_eth_event()),
+        );
+        let eth_balance_before_funding = state_before_funding.eth_balance.clone();
+
+        let funding_amount = Wei::new(10_000_000_000_000_000);
+        let funding_request = SweeperFundingRequest {
+            withdrawal_amount: funding_amount,
+            destination: "0xb44B5e756A894775FC32EDdf3314Bb1B1944dC34"
+                .parse()
+                .unwrap(),
+            ledger_burn_index: LedgerBurnIndex::new(0),
+            from: "k2t6j-2nvnp-4zjm3-25dtz-6xhaa-c7boj-5gayf-oj3xs-i43lp-teztq-6ae"
+                .parse()
+                .unwrap(),
+            from_subaccount: None,
+            created_at: 1699527697000000000,
+        };
+        let funding_flow = WithdrawalFlow {
+            tx_fee: GasFeeEstimate {
+                base_fee_per_gas: WeiPerGas::from(0xbc9998d1_u64),
+                max_priority_fee_per_gas: WeiPerGas::from(1_500_000_000_u64),
+            },
+            gas_limit: GasAmount::from(21_000_u32),
+            effective_gas_price: WeiPerGas::from(0x1176e9eb9_u64),
+            tx_status: TransactionStatus::Success,
+            ..WithdrawalFlow::for_request(funding_request)
+        };
+
+        let mut state_after_successful_funding = state_before_funding.clone();
+        let receipt_succeeded = funding_flow
+            .clone()
+            .apply(&mut state_after_successful_funding);
+        let after_success = state_after_successful_funding.eth_balance.clone();
+
+        // Asserted as the identity the accounting has to satisfy rather than as a fixed number: the
+        // funding ceiling covers both the ETH delivered and the fee, so whatever part of the fee
+        // went unspent is exactly what stays with the minter.
+        let unspent = after_success
+            .total_unspent_tx_fees
+            .checked_sub(eth_balance_before_funding.total_unspent_tx_fees)
+            .unwrap();
+        assert_eq!(
+            after_success.eth_balance,
+            eth_balance_before_funding
+                .eth_balance
+                .checked_sub(
+                    funding_amount
+                        .checked_sub(unspent)
+                        .expect("the unspent fee is part of the funding")
+                )
+                .unwrap(),
+            "a successful funding debits what it moved plus the fee it paid"
+        );
+        assert_eq!(
+            after_success.total_effective_tx_fees,
+            eth_balance_before_funding
+                .total_effective_tx_fees
+                .checked_add(receipt_succeeded.effective_transaction_fee())
+                .unwrap(),
+            "the fee actually paid must be counted"
+        );
+        assert!(
+            unspent > Wei::ZERO,
+            "this fixture over-provisions the fee, so some of it must be recorded as unspent"
+        );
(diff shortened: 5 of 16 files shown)


rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -744,6 +850,19 @@ impl EthTransactions {
                     );
                 }
             }
+            WithdrawalRequest::SweeperFunding(request) => {
+                if receipt.status == TransactionStatus::Failure {
```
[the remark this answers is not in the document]
mbjorkqvist: Confirmed and fixed in 3c5f94d720. Good find — this was a real defect, and a public-facing one.

Traced the flow to be sure: the funding burns ckETH at index `N`, its transaction finalizes with receipt status `Failure`, `record_finalized_transaction` takes the funding arm so no reimbursement request is created and `N` was never in `maybe_reimburse` — and then `retrieve_eth_status(N)` finds a finalized transaction with no reimbursed entry and a `Failure` status, returning `PendingReimbursement`. It stays that way permanently, because only recording a reimbursement moves that status on and nothing ever will. Both `retrieve_eth_status` and `withdrawal_status` are reachable for a funding index — there is no sender filter — so anyone can observe it.

`TxFinalizedStatus` had no variant meaning "failed, and will not be reimbursed", so one was added:

```
// Transaction failed and will not be reimbursed.
Failed : EthTransaction;
```

`processed_transaction_status` now picks between the two by asking the request whether it is reimbursable, rather than assuming failure implies reimbursement. Adding a case to a returned variant trips the Candid subtype rule, so this rides the same `CI_OVERRIDE_DIDC_CHECK` the event variant already needed.

On the missing coverage, also fixed — three tests rather than one, since the interesting part is the branch:

- `should_report_a_failed_funding_as_failed_not_pending_reimbursement`
- `should_report_a_successful_funding_as_success`
- `should_still_report_a_failed_user_withdrawal_as_pending_reimbursement`

The third is what gives the first its value: without it, an inverted branch would still pass.

Worth recording how reachable the failure itself is today, since it affects how you read the severity. A bare 21'000-gas transfer to a code-less address cannot revert, which is exactly why the sweeper is a plain EOA — so in production this is currently latent. It becomes live once the sweeper address can carry code, which EIP-7702 delegation introduces later in this feature: a delegated account executes its delegate on a plain transfer, and at 21'000 gas that runs out of gas and the transaction fails. PR 7 of this stack also forces it deliberately in a test.

Latent or not, it was worth fixing here rather than later: the burn is real, the ckETH is gone and will not be replaced, so telling an auditor reconciling supply that a reimbursement is pending is materially wrong.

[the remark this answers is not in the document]
mbjorkqvist: Addendum to my reply above, since it left out the part that determines how bad this actually is.

Saying the burned ckETH "will not be replaced" is true but incomplete: the ETH never moved either. On a failed funding `state.rs` records `transferred = 0` and only the effective gas as spent, so `burned_not_yet_spent` becomes ≈ the full funding amount, and `burn_required_for` then lets the *next* funding move that amount while burning (almost) nothing. ckETH ends up momentarily over-backed, never under-backed, and the only real cost is the failed transaction's 21'000 base gas.

So the wrong status was the actual defect here, not a stranded balance — nobody needs to intervene to recover funds. What does warrant attention is a *persistent* failure: each attempt still burns at least the ledger minimum, so the fee account would drain at that rate per funding interval while no gas is delivered. That belongs to observability rather than to this PR, and the alert condition is now recorded on DEFI-2965 along with the rest of the funding metrics.

mbjorkqvist (commented):
### On the suppressed comment (`state/transactions/mod.rs:645`, the `cketh_oldest_incomplete_eth_withdrawal_request_age_seconds` gauge)

The mechanics are exactly as described, and worth confirming: `oldest_incomplete_withdrawal_timestamp` chains `withdrawal_requests_iter()` with `maybe_reimburse_requests_iter()`, so once a funding's transaction is created it is in neither collection and contributes nothing to that gauge while it sits created or sent.

I am deliberately not changing it, for three reasons.

**The gauge is about user withdrawals, and it is alerted on as such.** Its name and help text say "ETH withdrawal request", and there is a stuck-withdrawal runbook attached to the alert (DEFI-2756). A minter-internal gas top-up appearing there would route an operator to a runbook about a user's stuck funds, for something no user is waiting on.

**The user-facing alert still fires when a wedged funding actually hurts.** A funding that never finalizes holds its nonce, which head-of-line blocks every later withdrawal. Those withdrawals are in `maybe_reimburse` or still pending, so they age and the alert fires — on the user-visible symptom, which is what it exists for. The only case it misses is a wedged funding with no user withdrawals behind it, where nothing user-facing is wrong yet.

**That remaining case has its own metric, in PR 6 of this stack.** `cketh_minter_sweeper_in_flight_funding_age_seconds` tracks a funding from acceptance through finalization, covering precisely the created and sent phases. Its own doc comment records why it cannot be folded into the balance-age gauge: the funding task refreshes the balance observation *before* consulting the in-flight guard, so that age resets every tick regardless.

The alternative you suggest — deriving incomplete requests from pending plus processed-without-a-finalized-transaction — would also change what the gauge reports for user withdrawals, on a metric that already has an alert and a runbook pointed at it. That is a behaviour change to production alerting, and it does not belong in the PR that introduces the variant.

Worth naming the one real cost: between this PR merging and PR 6 merging, a wedged funding has no age metric at all. That window is transient given the merge order, and the alert conditions for all six funding metrics are now recorded on DEFI-2965.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -274,6 +273,8 @@ type TxFinalizedStatus = variant {
     };
     // Transaction failed and will be reimbursed,
     PendingReimbursement : EthTransaction;
+    // Transaction failed and will not be reimbursed.
+    Failed : EthTransaction;
```
gregorydemay: That's a breaking change:

1. TxFinalizedStatus -> RetrieveEthStatus return type of the endpoint `retrieve_eth_status`

I think that an ETH transaction to an address controlled **by the minter** cannot fail. Generally the only way an ETH transaction can fail (meaning it's included in a block but its execution was reverted) is that the destination address has some smart contract code installed so that the 21k gas is not sufficient. This is not the case here. Something definitely to double check.

mbjorkqvist: You are right on both counts. Reverted in 51c3ccf106 — the variant, the `.did` line, the status branch and its tests are gone, and I dropped `[override-didc-check]` from the title, since needing that tag was the tell I should have followed up on rather than worked around.

On the breaking change: `TxFinalizedStatus` is reachable from `retrieve_eth_status` and from the withdrawal-status path, and adding a case to a variant in return position is not a Candid subtype — a client generated against the old interface cannot decode the new tag.

On whether a funding transaction can fail, I checked rather than assumed, and it cannot:

- The destination is `sweeper_address()`, derived from the minter's master threshold-ECDSA key (`deposit_address/mod.rs:48`), so it is an EOA the minter controls with no code installed.
- The transaction is a plain value transfer — `data: Vec::new()`, empty access list, the 21'000 intrinsic gas — so there is no execution to revert. A gas limit below intrinsic cost means the transaction is never mined, rather than mined and failed.
- An EOA acquires code only through an EIP-7702 authorization signed by its own key, which only the minter holds, and landing CREATE2 code on one chosen address is infeasible.

So the branch was dead code bought with a breaking interface change. A failed funding now reports the pre-existing pending-reimbursement status. That is imprecise for a request nothing will reimburse, so the site carries a comment saying why it is tolerable and what would invalidate it — funding routed through a contract, where a revert becomes possible. Worth noting because the demo already contains `CkSweeperViaHelper.sol`, so that is not a hypothetical direction.

Visibility is unchanged: `record_finalized_transaction` already logged this case, and I only sharpened the wording to `UNEXPECTED: … which should be impossible for a transfer to an address the minter controls`, since at INFO it otherwise reads like a routine outcome. INFO is the loudest sink the minter has.

One alternative I considered and did not take: making funding requests reimbursable, crediting `{minter, 0fee}` where the ckETH was burned from. That would make the status honest without any new API surface, but it changes the burn-first invariant in #11083, since a reimbursement credits ckETH back. Happy to go that way instead if you would rather not rest on an impossibility argument.

`is_reimbursable()` stays, incidentally — it drives the `maybe_reimburse` bookkeeping and predates the reverted commit.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -371,6 +371,10 @@ impl State {
                 .checked_sub(tx.transaction().amount)
                 .expect("BUG: withdrawal amount MUST always be at least the transaction amount"),
             WithdrawalRequest::CkErc20(req) => req.max_transaction_fee,
+            WithdrawalRequest::SweeperFunding(req) => req
```
mbjorkqvist: Addressing the suppressed comment from the latest Copilot round, which flagged that this arm has no coverage: correct, and fixed in f8b70daa3b with `should_update_after_successful_and_failed_sweeper_funding` next to the ckETH and ckERC20 balance tests.

One detail of the report is worth correcting for anyone reading later: nothing in `state/tests.rs` calls `State::record_finalized_transaction` directly — the existing balance tests reach it by applying events, which is the idiom there. So the gap was not a missing call style but a missing *case*: the ckETH and ckERC20 tests cannot enter this arm, so an incorrect debit, an uncounted fee, or the `expect` above trapping on a funding whose ceiling is below its transaction amount would all have gone unnoticed.

The new test asserts the identity the accounting has to satisfy rather than fixed numbers — the funding ceiling covers the ETH delivered plus the fee, so whatever part of the fee goes unspent is exactly what stays with the minter — and then asserts the failing case debits only the fee actually paid, with both fee counters unchanged between the two outcomes. Numbers copied from the ckETH fixture would have hidden the first arithmetic mistake I made here, which is what pushed me to the identity form.

For what it is worth, richer funding-accounting tests do exist later in the stack (`should_account_for_a_successful_sweeper_funding` and `should_keep_a_failed_sweeper_funding_as_prepaid_gas`), but they cannot move here: they depend on the burn-first accounting added in #11083 and on the `cketh_burned` field added in #11086. This test is deliberately scoped to what this PR introduces.

rs/ethereum/cketh/minter/src/dashboard/tests.rs
```diff
@@ -884,12 +884,16 @@ fn should_display_reimbursed_requests() {
                                 },
                             );
                         }
```
gregorydemay, line 886: Here for a lack of a better place: I would clean up the PR description:

> A failed funding transaction [...].

Maybe clarify what "failed" here means because as discussed there cannot be a failed funding transaction on Ethereum.

> An earlier revision reported a failed funding as its own[...]

I would remove any intermediate state from the PR description.

rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -182,6 +182,9 @@ pub enum EventType {
     /// durable even across an ungraceful trap (unlike the pre-upgrade snapshot).
     #[n(26)]
     AutomaticDepositReceived(#[n(0)] AutomaticDeposit),
+    /// The minter burned ckETH from its fee subaccount to top up the sweeper address with gas.
+    #[n(27)]
+    AcceptedSweeperFundingRequest(#[n(0)] SweeperFundingRequest),
```
gregorydemay: Why not reuse `EthWithdrawalRequest` since a funding request corresponds 1:1 to an ETH withdrawal request (the fields in `SweeperFundingRequest` and in `EthWithdrawalRequest` are the same)? We could have

```rust
    #[n(27)]

```

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -36,20 +38,23 @@ pub enum WithdrawalSearchParameter {
 pub enum WithdrawalRequest {
     CkEth(EthWithdrawalRequest),
     CkErc20(Erc20WithdrawalRequest),
+    SweeperFunding(SweeperFundingRequest),
```
gregorydemay: At the risk of contradicting myself and the current spec, I'm wondering if the extra-complexity of handling an ETH transaction differently between a funding transaction and a user withdrawal is worth it/make sense. In case the funding transaction would fail on Ethereum (which as we discussed should actually not happen,) why not reimburse the minter at the same account/subaccount as the funds were originally burnt from? I would not add that machinery if it was not already there but it's not clear to me why the Ethereum transactions (funding vs user) should be handled differently? Happy to hear your opinion on this, WDYT?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -744,6 +850,25 @@ impl EthTransactions {
                     );
                 }
             }
+            WithdrawalRequest::SweeperFunding(request) => {
+                if receipt.status == TransactionStatus::Failure {
+                    // Funding is a plain value transfer to an address derived from the minter's
+                    // own key, so there is no code for it to revert in: reaching this means an
+                    // assumption broke. Logged rather than trapped, since the accounting holds
+                    // either way — the burn simply stays unspent.
+                    log!(
```
gregorydemay, line 806: should we maybe also have a metric (fine as a separate PR) to track the number of failed funding transactions so that we could have an alert on it?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -881,6 +1006,12 @@ impl EthTransactions {
                 );
             }
             if tx.transaction_status() == &TransactionStatus::Failure {
+                // A sweeper funding request is never reimbursed, so this status is imprecise for
+                // one. Tolerated rather than given a status of its own, which would mean adding a
```
gregorydemay: Actually, it's not "is imprecise", it's "unreachable", no?

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -2118,6 +2122,222 @@ mod eth_transactions {
...
+                    TransactionNonce::ZERO,
+                    gas_fee_estimate(),
+                    SWEEPER_FUNDING_GAS_LIMIT,
+                    EthereumNetwork::Mainnet,
+                ),
+                Err(CreateTransactionError::InsufficientTransactionFee {
+                    cketh_ledger_burn_index,
+                    allowed_max_transaction_fee,
+                    ..
+                }) if cketh_ledger_burn_index == LedgerBurnIndex::new(15)
```
gregorydemay: nit: It's not really clear where this `15` value comes from. Something like `funding.ledger_burn_index` would make it clearer.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -2118,6 +2122,222 @@ mod eth_transactions {
...
+            CreateTransactionError, NotReimbursable, ReimbursementIndex, SweeperFundingRequest,
+            create_transaction,
+        };
+        use crate::tx::GasFeeEstimate;
+        use assert_matches::assert_matches;
+        use ic_ethereum_types::Address;
+        use maplit::{btreemap, btreeset};
+        use std::str::FromStr;
+
+        const SWEEPER_FUNDING_GAS_LIMIT: GasAmount = GasAmount::new(21_000);
```
gregorydemay: nit: why not reuse `CKETH_WITHDRAWAL_TRANSACTION_GAS_LIMIT`?

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -2118,6 +2122,222 @@ mod eth_transactions {
...
+            let request: WithdrawalRequest = sweeper_funding_request().into();
+
+            assert_eq!(
+                crate::withdraw::estimate_gas_limit(&request),
+                crate::withdraw::CKETH_WITHDRAWAL_TRANSACTION_GAS_LIMIT,
+            );
+        }
+
+        #[test]
+        fn should_survive_an_event_encoding_roundtrip() {
```
gregorydemay: nit: this type of tests are better suited as proptest with an arbitrary event, see `arb_withdrawal_request`. I would actually expect that there is a minicbor coding roundtrip with an aribtrary event (not just `AcceptedSweeperFundingRequest`)

gregorydemay: PS: I think this test is already covered with `event_encoding_roundtrip`

rs/ethereum/cketh/minter/src/state/tests.rs
```diff
@@ -1501,6 +1524,108 @@ mod eth_balance {
...
+
+        let mut state_after_successful_funding = state_before_funding.clone();
+        let receipt_succeeded = funding_flow
+            .clone()
+            .apply(&mut state_after_successful_funding);
+        let after_success = state_after_successful_funding.eth_balance.clone();
+
+        // Asserted as the identity the accounting has to satisfy rather than as a fixed number: the
+        // funding ceiling covers both the ETH delivered and the fee, so whatever part of the fee
+        // went unspent is exactly what stays with the minter.
```
gregorydemay, line 1629: The first part of the test (success case) exactly matches `should_update_after_successful_and_failed_withdrawal`, why not reuse it?

gregorydemay (commented):
Thanks @mbjorkqvist ! Generally looks good to me, I have an idea to simplify things further (see comment below) and please do challenge it!

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
mbjorkqvist: Good question! The way I see it, there are pros and cons to both approaches.

With the current implementation, we burn the amount we want to top the sweeper up with, and the transaction fee is then carved out of that same amount, exactly as for a user withdrawal. So for a 0.1 ETH top-up we burn 0.1 ckETH, provision `21'000 × (2 x base + priority)` for gas, and the sweeper actually receives 0.0995. We typically pay `21'000 × (base + priority)`, so 0.00021 ETH of the provisioned fee is never spent, and that lands in `burned_not_yet_spent`. Next time we fund the sweeper we take that into account and burn 0.00021 less than we otherwise would. If a funding transaction were to fail, we'd have 0.0997 sitting in `burned_not_yet_spent` (everything except the gas we actually paid), and the next funding would burn almost nothing before trying again. For failed user withdrawals we refund the transferred amount but never the fee, so that part isn't symmetric either.

If we also refunded failed funding transactions:

1. We'd get rid of the special-casing: `is_reimbursable()` (false only for funding), the `if is_reimbursable` guard around `maybe_reimburse.insert`, the matching conditional assertion in `record_finalized_transaction`, the fallible `ReimbursementIndex` conversion, and the funding-only log arm. Funding would just take the same path as a ckETH withdrawal.
2. The recovered ckETH would end up in `{minter, 0x…0fee}` where you can query it with `icrc1_balance_of`, instead of showing up as `burned_not_yet_spent() == 0.0997`, which the minter recomputes from the audit log on every upgrade (`replay_events`).
3. The credit is only spendable by another funding, whereas ckETH in the fee account can pay for anything that account funds. If we ever change how sweeping works, or turn funding off, the credit is stranded.
4. It's cheap: a ckETH mint, no gas, no tECDSA, no nonce, fee-free, and `process_reimbursement` already mints to an arbitrary `to`/`to_subaccount`.
5. The path would be in use already if the failure ever becomes reachable, e.g. EIP-7702 delegation or sweeping via a contract.

The drawbacks:

1. Right now the invariant can't break arithmetically: `record_burn` only adds, so `cumulative_burned` only grows and `burned_not_yet_spent()` is a subtraction. A refund would be the first thing that has to reduce it, across an `await` — `process_reimbursement` mints first, then records. If it traps in between (`QuarantinedReimbursement`), the 0.0997 is back in the fee account while `cumulative_burned` still counts it, so one burn pays for two transfers and ckETH is under-backed.
2. It means minting for a state that can't happen on mainnet, and minting is the only thing that can under-back ckETH. A bug there wouldn't be user-triggerable, but it would be a new way to create ckETH that isn't backed.
3. The failure mode would be quiet. Nothing traps if we miss the decrement, we just report 0.0997 of prepaid gas that isn't there and find out later. The rest of the module deliberately fails loudly: `burned_not_yet_spent()` panics instead of saturating, `mark_funding_in_flight` traps instead of overwriting ("trapping leaves ckETH over-backed, the safe direction").
4. It adds a third place where the accounting changes. Today it's the burn and finalization, both synchronous with the event; a refund arrives on a timer, possibly much later.
5. In the normal case it's a round trip: mint 0.0997 back, then burn it again at the next funding. Two more ledger transactions and one more failure window, to put the money back where it immediately goes again.
6. It doesn't close the one gap that's actually there: if we trap between `burn_from_own_subaccount` returning and recording `AcceptedSweeperFundingRequest`, there's no request to reimburse, `cumulative_burned` never sees the burn, and ckETH ends up over-backed by it. That one needs reconciliation against the ledger (burns from `{minter, 0x…0fee}` vs the events) and crediting the orphan, no mint involved.

gregorydemay: It seems you're not super convinced, and the following point 

> For failed user withdrawals we refund the transferred amount but never the fee, so that part isn't symmetric either.

is a good one and something I forgot. If we wanted to do this properly for funding requests then we would need to reimburse the amount+unspent transaction fees (the rest is used by the protocol for the failed transaction) so that special treatment would indeed be required, making the whole argument a bit moot.

Also given that minting is always security critical it's also good to not open another door, in particular given that we don't need it.

In short, let's keep this part as is. Thanks for the discussion!

rs/ethereum/cketh/minter/src/dashboard/tests.rs
mbjorkqvist: Both done, in the description rather than the code.

Dropped the section describing the earlier revision entirely — you're right that the intermediate state doesn't belong there, and the reverted attempt is on record in the `.did` thread above for anyone who wants the history.

For "failed", the bullet now says what it means and that it cannot happen: nothing is recorded to pay back if a funding transaction finalizes with a *failure receipt*, which requires code at the destination and so cannot arise for a bare transfer to the sweeper. It's logged as unexpected rather than handled, and the burn stays available as prepaid gas.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
mbjorkqvist: Good idea, added as `cketh_minter_sweeper_funding_failed_total` in b633db9157 on #11094, where the rest of the funding metrics live.

Two small things worth knowing about it. It is incremented where the receipt status is already matched, next to the transferred amount, so the two cannot drift apart — deriving it from `transferred == 0` would have been wrong, since a funding whose fee ate the whole ceiling delivers zero on success too. And it saturates rather than checked-adds: an alert that silently stops counting is worse than one that wraps.

Expected to stay at zero, which is exactly what makes it alertable.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
mbjorkqvist: You're right, and the old wording buried the stronger claim. Reworded in 2479790dd2 to lead with it:

```
// Unreachable for a funding: the destination is derived from the minter's own
// key, so a bare transfer there has no code to revert in. Were it reached, the
// status would be wrong, since nothing reimburses a funding — tolerable only
// because it cannot happen, and not worth a status of its own, which would mean
// adding a variant to `retrieve_eth_status`' return type and breaking existing
// clients. Revisit if funding ever goes through a contract, where a revert becomes
// possible.
```

"Unreachable" means on mainnet, not untested: the adversarial test later in the stack does reach it, by planting reverting code at the sweeper address.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
mbjorkqvist: Agreed, fixed in 2479790dd2 — the test now binds the request's own index and matches on that, so the assertion can't drift from the fixture.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
mbjorkqvist: Done in 2479790dd2. It was worse than a duplicated literal: the test used its own `const SWEEPER_FUNDING_GAS_LIMIT = 21_000` while `should_use_the_plain_transfer_gas_limit` right below asserts that `estimate_gas_limit` returns `CKETH_WITHDRAWAL_TRANSACTION_GAS_LIMIT` — so the neighbouring tests were checking a copy of the value under test. Both now use the constant.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
mbjorkqvist: You're right, and I checked before deleting: `arb_event_type` includes `arb_sweeper_funding_request().prop_map(EventType::AcceptedSweeperFundingRequest)`, so `event_encoding_roundtrip` already covers the new event with arbitrary values, which beats my single fixed instance. Removed in 2479790dd2.

rs/ethereum/cketh/minter/src/state/tests.rs
mbjorkqvist: Fair, deduplicated in 2479790dd2 by extracting the part that was actually shared: `apply_eth_transfer_both_ways` runs the 21'000-gas flow against a starting state twice, once finalizing successfully and once with a failure receipt, and hands back both resulting states and both receipts. `should_update_after_successful_and_failed_withdrawal` and the funding test now both use it.

I kept the assertions separate rather than merging the tests. The ckETH one pins exact numbers, whereas the funding one asserts the identity the accounting has to satisfy — the ceiling covers the ETH delivered plus the fee, so whatever part of the fee goes unspent is what stays with the minter. That distinction earned its keep: my first version of the funding test copied the ckETH test's debit constant, which already includes the fee, and it failed for that reason.

rs/ethereum/cketh/minter/src/state/event.rs
mbjorkqvist: The fields are identical today, but they stop being so in the next PR of the stack: `SweeperFundingRequest` gains a `cketh_burned` field, which records what the funding actually burned as opposed to the ceiling it may spend. The two differ whenever earlier burns went unspent, and they nearly always do — we provision `2·base + priority` per gas and pay `base + priority`, so roughly 40% of every provisioned fee is left over and offsets the next burn. So a funding of 0.085 ETH typically burns a bit less than that, and the accounting has to know both numbers.

It also has to be in the *event*, not derived: `SweeperFundingAccounting` is rebuilt by folding the audit log on every upgrade, and the burn depends on the credit at the time it was decided. If the event only carried the amount, replay would have to recompute the burn under whatever rules exist at replay time, so changing the offsetting logic later would silently rewrite history. (`created_at` also differs, mandatory here versus `Option<u64>` for a withdrawal.)

That said, your point about not declaring the same six fields twice stands, and there is a version that keeps it:

```rust
pub struct SweeperFundingRequest {
    #[n(0)] pub withdrawal: EthWithdrawalRequest,
    #[n(1)] pub cketh_burned: Wei,
}
```

That declares the shared shape once, keeps the burn in the event, and would let two match arms collapse that are currently identical for `CkEth` and `SweeperFunding` — the `ResubmissionStrategy` (both `ReduceEthAmount { withdrawal_amount }`) and `charged_tx_fee` in `record_finalized_transaction`. The costs are one nesting level in the CBOR encoding (fine, nothing is released yet), an extra hop in every field access across four PRs of the stack, and `created_at` becoming optional again unless we assert on construction.

I lean towards keeping the separate struct, mostly because after the reimbursement discussion the distinct type is what carries "this one is never reimbursed" — but it is a close call and the composition version is a genuine improvement on the duplication. WDYT?

gregorydemay (approved):

rs/ethereum/cketh/minter/src/state/event.rs
mbjorkqvist: Agreed offline to drop `cketh_burned` and reuse `EthWithdrawalRequest`, which is what e67b6c7365 does. The consequence is that burned-but-unspent funds are no longer offset against the next funding burn — the surplus just stays as backing, exactly like the unspent gas of a user withdrawal — so the field it needed disappears with it, and funding accounting no longer has any piece without a counterpart in the withdrawal flow. `R14` in the design doc is updated accordingly in ed2177c4ca.

Sharing the payload also let nine arms that were duplicating the ckETH behaviour merge with it. The variant stays, so `is_reimbursable() == false` and the fallible reimbursement index still make the compiler prove that funding never reaches the reimbursement machinery.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
mbjorkqvist: Agreed, and thanks for the discussion. One correction to the numbers in my reply above: they assumed that burned-but-unspent funds are offset against the next burn, and that offsetting is gone as of e67b6c7365 / ed2177c4ca. So the unspent part of a funding's fee — and the whole burn of a failed one, still unreachable — simply stays as ckETH backing instead of reducing what the next funding burns. The conclusion is unchanged: nothing is minted back.

rs/ethereum/cketh/minter/src/state/event.rs
mbjorkqvist: One more consequence of sharing the payload, and it bit: `record_withdrawal_request` takes `impl Into<WithdrawalRequest>`, and `EthWithdrawalRequest`'s own conversion yields `CkEth`, so the audit handler for `AcceptedSweeperFundingRequest` was recording every funding as an ordinary user withdrawal — reimbursable, and invisible to everything that keys off the variant. Fixed in cd5926a792 by naming the variant at that call site, with a test that replays the event and asserts both the variant and that it cannot be reimbursed.

CI was green on it because balance accounting treats the two variants identically; it only surfaced when the accounting in #11083 started reporting zero spend for a finalized funding. Worth remembering for the rest of the stack: with a shared payload, `.into()` is the wrong tool at every funding construction site.

PR #10919, opened 2026-07-27
chore: update rules_rust

nmattia:
This drops our local rules_rust patch which we used to hardcode strip values, and instead use the new `strip_level_select` tag:

https://github.com/bazelbuild/rules_rust/pull/4167

PR #11074, opened 2026-08-07
chore(bazel): move C++ libraries from rust `deps` to `link_deps`

basvandijk:
Since the rules_rust update in #10919, `rust_library`/`rust_binary`/`rust_test` print a deprecation warning for every `cc_library` found in their `deps` — e.g. [this run](https://github.com/dfinity/ic/actions/runs/31178310153/job/92865651208):

```
WARNING: Target @@+new_local_repository+libsystemd//:libsystemd in 'deps' of
@@rules_rust++crate+crate_index__libsystemd-sys-0.9.3//:libsystemd_sys is a C++ library.
Only Rust targets are allowed in 'deps'. Please use 'link_deps' for manual FFI linkage.
Support for C++ libraries in 'deps' is deprecated and will be removed in a future release.
```

Support is slated for removal, so this is a build break waiting on the next rules_rust bump, not just log noise.

## Changes

- Switch the `devicemapper-sys`, `libcryptsetup-rs-sys` and `libsystemd-sys` crate annotations from `deps` to `link_deps`. Behaviour-preserving: for a target that only provides `CcInfo`, `transform_link_deps` builds exactly the same `DepVariantInfo` that `transform_deps` does.
- `guest_vm_runner` hit the same warning by listing `@libvirt` directly in the `deps` of its two `rust_binary`s and one `rust_test`. Rather than repeat the attribute move three times, attach the library to the `virt-sys` crate instead — it is a pkg-config `-sys` crate (`links = "virt"`) just like the three above, so this is the pattern `bazel/pkg-config.patch` already documents: *"The `-l` libs resolve against cc_import dependencies that link the host library by explicit path"*. This needs a new `//third_party:libvirt` alias, since apparent repo names don't resolve from inside the generated crate repositories.
- Repin `Cargo.Bazel.json.lock` (checksum + the four `extra_deps` → `extra_link_deps` entries; `Cargo.Bazel.toml.lock` is untouched).

## Verification

`bazel query` confirms `//rs/ic_os/os_tools/guest_vm_runner`, `//rs/ic_os/os_tools/guest_disk` and `//rs/ic_os/device` are the only in-repo consumers of these four crates. Rebuilding all three with the analysis cache discarded: **6 warnings before, 0 after**, and the `NEEDED` entries of the resulting `guest_vm_runner` and `guest_disk` binaries are byte-identical (`libvirt.so.0`, `libsystemd.so.0`, `libcryptsetup.so.12` all still linked).

`bazel build //... --nobuild` is clean and `//rs/ic_os/os_tools/guest_vm_runner:guest_vm_runner_test` passes. The two remaining tests over these targets (`guest_disk_test`, `upgrade_device_mapper_test`) are `manual` because they need root; they build, and are exercised via `//rs/tests/node:root_tests`.

### DIFF
--- a/third_party/BUILD.bazel
+++ b/third_party/BUILD.bazel
@@ -1,5 +1,5 @@
 # Aliases into the host-import repos (see BUILD.devmapper.bazel and
-# BUILD.cryptsetup.bazel), so that crate annotations in
+# BUILD.cryptsetup.bazel), so that the `link_deps` of the crate annotations in
 # bazel/rust.MODULE.bazel can reference them with stable root-repo labels
 # (apparent repo names like @devmapper don't resolve from inside the
 # generated crate repositories).
@@ -12,6 +12,12 @@ alias(
 alias(
     name = "devmapper",
     actual = "@devmapper//:devmapper",
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "libvirt",
+    actual = "@libvirt//:libvirt",
     visibility = ["//visibility:public"],
 )
 
--- a/third_party/BUILD.bazel
+++ b/third_party/BUILD.bazel
@@ -1,8 +1,8 @@
-# Aliases into the host-import repos (see BUILD.devmapper.bazel and
-# BUILD.cryptsetup.bazel), so that the `link_deps` of the crate annotations in
-# bazel/rust.MODULE.bazel can reference them with stable root-repo labels
-# (apparent repo names like @devmapper don't resolve from inside the
-# generated crate repositories).
+# Aliases into the host-import repos (see BUILD.cryptsetup.bazel,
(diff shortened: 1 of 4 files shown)


blind-oracle (approved):

third_party/BUILD.bazel
```diff
@@ -15,6 +15,12 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+alias(
```
nmattia, line 18: is this actually used? I think you're using the canonical name anyway: `@@//third_party:libvirt`

nmattia (approved):
LGTM modulo the libvirt alias question!

third_party/BUILD.bazel
basvandijk: `@@//third_party:libvirt` is this alias. The canonical name is `@libvirt//:libvirt`.

PR #11078, opened 2026-08-08
feat(sns-governance): expose latest reward-event shares on neurons

aodl:
Context -> https://forum.dfinity.org/t/exposing-sns-voting-participation-shares-after-reward-settlement/74916

## What

SNS Governance already calculates exact per-neuron reward shares while settling each reward event, but currently discards them after using them to allocate native maturity.

This change stores each participating neuron's positive share together with the existing reward-event end timestamp in a new optional `Neuron.latest_reward_event_participation` field. The field is exposed through the existing `get_neuron` and paginated `list_neurons` APIs.

Participation is recorded independently of the native reward configuration, so it remains available when SNS voting reward rates are zero. Existing maturity allocation, proposal settlement, ballot clearing, reward timing, neuron visibility, and pagination remain unchanged.

## Why

This makes the canonical SNS voting-participation weighting available after settlement, allowing generic SNS applications to implement alternative participation-based reward schemes. For example, an app that has paying customers can share its profits with its SNS neurons.


aodl:
cc @daniel-wong-dfinity-org

borovan:
## Consumer review from `ic-query`

Thanks for putting this together. This is a strong fit for [`ic-query`'s existing SNS reward-checkpoint work](https://github.com/dragginzgame/ic-query/blob/main/docs/design/0.26/0.26-design.md): exposing the canonical voting-power weight directly is much better than reconstructing it from ballots or inferring it from maturity changes, and it remains useful when native reward rates are zero.

### Main request: add an event-level total and capability marker

Before downstream consumers build on this contract, would you consider adding an optional canonical total to `RewardEvent`, for example:

```protobuf
optional Uint128 total_reward_shares = <unused tag>;
```

For every reward event created by a Governance version supporting this feature, it should be populated, including `Some(0)`. Legacy reward events would naturally decode it as absent.

This would solve two important consumer problems:

1. **Rolling-upgrade capability detection.** If every neuron's new field is absent, a client currently cannot distinguish an older SNS Governance Wasm from a new event with legitimately zero positive participation. This matters especially because the feature is intended to work when native reward rates are zero.
2. **Snapshot validation and normalization.** A paginated consumer could check that the sum of all neuron records tagged with the current event timestamp exactly equals Governance's canonical total. It also supplies an authoritative denominator instead of requiring consumers to normalize over whichever rows they happened to enumerate.

An equivalent event-level summary/capability wrapper would work too; the important properties are explicit feature availability and a canonical total populated even for zero.

### Other feedback

- **Please include an explicit human privacy/threat-model review.** The field does not reveal `Yes` versus `No`, but it does reveal that a neuron participated and its aggregate voting-power magnitude, potentially indefinitely when that neuron does not participate again. That is intrinsic to the use case, but it should be weighed explicitly against the post-settlement ballot-privacy concerns discussed in the community.
- **Define malformed inner-option behavior.** The API permits `Some(RewardEventParticipation { reward_shares: None })`, even though Governance always writes the value. Consumers should treat a matching-event record with missing shares as invalid evidence, not as zero. Similarly, a matching-event value of zero should be invalid if only positive participants are persisted.
- **Add an end-to-end value-above-`u64` test.** The conversion helper covers `u128`, but it would be valuable to settle multiple proposals whose voting-power sum exceeds `u64::MAX`, then verify the exact high/low value through both `list_neurons` and `get_neuron`. That protects the full calculation → storage → API path.
- **Record worst-case state/response growth.** The per-neuron addition looks modest and reuses the existing reward loop, but a maximum-neuron state-size estimate and representative response-size measurement would make the operational cost explicit.

### How `ic-query` would use it

`ic-query` already performs strict `list_neurons` exhaustion bracketed by complete reward-event, nervous-system-parameter, and running-version responses. With this field it can produce direct, single-checkpoint participation evidence:

- preserve the raw event timestamp and `high`/`low` limbs;
- expose an exact decimal string in JSON rather than a potentially lossy JSON number;
- classify absent or older tags as zero shares for the bracketed event;
- reject malformed matching-event records;
- sum using checked `u128` arithmetic and compare with the proposed event-level total;
- expose exact numerator/denominator values without floating-point percentages.

This would remove the need for two maturity checkpoints, immediate-event maturity reconciliation, and maturity-conversion policy assumptions when the goal is external participation weighting. It still would not infer a payout beneficiary from neuron permissions; beneficiary authorization is a separate application concern.

Overall, the per-neuron implementation looks useful and well targeted. The event-level total/capability marker is the main addition I would want before treating the result as a robust generic allocation contract.

borovan:
### Narrow follow-up on the privacy point above

After re-reading the current patch, I think the concern can be addressed without introducing a new commitment system or substantially expanding this PR.

The important distinction is between:

1. storing `latest_reward_event_participation` internally so reward calculations can use it; and
2. unconditionally projecting that field into the existing public `get_neuron` and bulk `list_neurons` responses.

The second part creates a public, neuron-level activity dataset containing both participation and reward-share magnitude. Because neuron IDs are stable, repeated snapshots can be correlated over time. Splitting does not make this a reliable measure of holder-level participation: it mainly makes correlation easier for unsplit neurons than for holders who deliberately reorganize their stake.

My narrow recommendation for this PR is therefore:

- Keep the internal stable-state field and reward-event accounting.
- Do not expose the per-neuron participation field through anonymous `list_neurons` in this PR.
- Prefer not exposing it through anonymous `get_neuron` either.
- Add the event-level `total_reward_shares` value suggested above, so external tooling has the denominator needed for aggregate reward calculations.
- Treat selective per-neuron disclosure as a separate API decision.

If per-neuron access is required in this PR, a bounded compromise would be:

- omit it from `list_neurons`;
- return it from `get_neuron` only when the caller is authorized for that neuron using an existing permission such as `ManagePrincipals`; and
- return it only when its event timestamp matches the latest completed reward event, so an old positive record does not become an indefinitely retained activity marker.

That policy would apply uniformly to every neuron, without stake thresholds or holder-specific exceptions.

This also needs a clear scope boundary in the API documentation: this is an SNS reward-accounting feature and should not establish a general governance API pattern. Any equivalent NNS exposure should require its own design, privacy, and governance review.

At minimum, I would add tests proving that:

- anonymous `list_neurons` cannot retrieve per-neuron reward participation;
- unauthorized `get_neuron` callers cannot retrieve it;
- authorized access works when supported;
- stale participation records are redacted; and
- the event-level total and per-neuron `u128` values round-trip correctly.

This seems like the smallest change that preserves the useful accounting work while avoiding a permanent anonymous per-neuron activity feed.

borovan:
### How SNS could become a precedent for NNS

  There is no automatic SNS-to-NNS path. A hypothetical progression would require several separate decisions:

  1. The SNS field is accepted as a useful public transparency feature.
  2. Dashboards and tools begin depending on it.
  3. Public per-neuron reward data becomes treated as an established design pattern.
  4. Someone proposes adding the equivalent feature to NNS neurons.
  5. The NNS implementation copies the SNS disclosure model without applying the NNS’s existing private-neuron protections.

  6. Public collectors build a long-term database of NNS reward participation.

  That would require a separate NNS code change, review, canister upgrade proposal, and NNS vote. This PR cannot make it happen by itself.

borovan:
imo fix the privacy issues or it's going to be a problem

aodl:
> That would require a separate NNS code change, review, canister upgrade proposal, and NNS vote. This PR cannot make it happen by itself.

Yes. That's correct, this is tightly scoped to the SNS framework

> imo fix the privacy issues or it's going to be a problem

Please could you articulate your concern regarding privacy?

This PR doesn't change the status quo regarding privacy in any way, but I'd be happy to discuss this with you further if you disagree. 

Thanks for your interest.

aodl:
> Dear @aodl,
> 
> In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).
> 
> If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/635) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.
> 
> — The DFINITY Foundation

Done

aodl:
> return it from get_neuron only when the caller is authorized for that neuron using an existing permission such as ManagePrincipals;

There is no such permission guard on readability. It's all unconditionally public.

If you'd like the public/private toggle that the NNS has but for SNSs, I would suggest taking that to a new PR. Note that that would actually be a bigger change than you expected. If it were well implemented though, I'd probably support it.

aodl:
@daniel-wong-dfinity-org, @basvandijk, I'm not sure why the CLA task is still blocking. I actioned the request yesterday

borovan:
https://forum.dfinity.org/t/beyond-the-sns-why-toko-could-be-its-natural-successor/74992/2 - I think we're going to replace the SNS anyway

daniel-wong-dfinity-org-twin:
The PR description probably does not need to be that long. Can you shorten it?

daniel-wong-dfinity-org-twin:
> to allocate external, non-inflationary rewards without independently reconstructing ballots.

How about instead of this, "to implement alternative participation-based reward schemes. For example, an app that has paying customers can share its profits with its SNS neurons."

rs/sns/governance/canister/governance.did
```diff
@@ -584,6 +584,16 @@ type FolloweesForTopic = record {
   topic : opt Topic;
 };
 
+type Uint128 = record {
+  high : nat64;
+  low : nat64;
+};
```
daniel-wong-dfinity-org-twin: I don't like how this is an ad hoc solution to a general problem.

Here's another idea: use `nat` instead of this (one of the few features of Candid that I actually really like).

In Rust, `candid::Nat` is just a wrapper around `num_bigint::BigUint`.

For storage (i.e. Protocol Buffers), convert `BigUint` using `to_bytes_be` (alternatively, little endian), which can then be stored in a field of type `bytes` in Protocol Buffers.

WDYT?

rs/sns/governance/canister/governance.did
```diff
@@ -584,6 +584,16 @@ type FolloweesForTopic = record {
...
+
+type RewardEventParticipation = record {
+  reward_event_end_timestamp_seconds : nat64;
```
daniel-wong-dfinity-org-twin: Make this opt, in case we want to delete this later, replace it with something else. That is our policy. (This is not clear from what exists in our .did files, because we came up with it later).

rs/sns/governance/src/governance/reward_event_participation_tests.rs
daniel-wong-dfinity-org-twin, line 1: I haven't reviewed this yet, because I think there are some important design decisions we should figure out first.

rs/sns/governance/src/pb/conversions.rs
```diff
@@ -5,6 +5,24 @@ use crate::{
...
+}
+
+impl From<pb_api::Uint128> for pb::Uint128 {
+    fn from(item: pb_api::Uint128) -> Self {
+        Self {
+            high: item.high,
+            low: item.low,
+        }
+    }
+}
```
daniel-wong-dfinity-org-twin: If we end up going with my suggestion to use `nat` (instead of your `Uint128`). This would go away, I think?

rs/sns/governance/src/pb/conversions.rs
```diff
@@ -200,6 +224,24 @@ impl From<pb_api::Neuron> for pb::Neuron {
     }
 }
 
+impl From<pb::neuron::RewardEventParticipation> for pb_api::neuron::RewardEventParticipation {
+    fn from(item: pb::neuron::RewardEventParticipation) -> Self {
+        Self {
+            reward_event_end_timestamp_seconds: item.reward_event_end_timestamp_seconds,
+            reward_shares: item.reward_shares.map(pb_api::Uint128::from),
```
daniel-wong-dfinity-org-twin: This would be where blob gets converted (back) into `BigUint`/`candid::Nat` via `from_bytes_be` (or little endian).

rs/sns/governance/src/pb/conversions.rs
```diff
@@ -200,6 +224,24 @@ impl From<pb_api::Neuron> for pb::Neuron {
...
+        }
+    }
+}
+
+impl From<pb_api::neuron::RewardEventParticipation> for pb::neuron::RewardEventParticipation {
+    fn from(item: pb_api::neuron::RewardEventParticipation) -> Self {
+        Self {
+            reward_event_end_timestamp_seconds: item.reward_event_end_timestamp_seconds,
+            reward_shares: item.reward_shares.map(pb::Uint128::from),
```
daniel-wong-dfinity-org-twin: Ditto: this is where `BigUint` would become a blob via `to_bytes_be` (or little endian).

rs/sns/governance/src/governance.rs
```diff
@@ -702,6 +702,21 @@ fn spawn_in_canister_env(future: impl Future<Output = ()> + Sized + 'static) {
...
+fn try_convert_reward_shares_to_u128(reward_shares: Decimal) -> Result<u128, String> {
+    if reward_shares.is_sign_negative() {
+        return Err(format!(
+            "Reward shares must be non-negative: {reward_shares}."
+        ));
+    }
+
+    if reward_shares.fract() != Decimal::ZERO {
+        return Err(format!("Reward shares must be integral: {reward_shares}."));
+    }
```
daniel-wong-dfinity-org-twin: Why? I mean, if in practice, reward_shares would always be a (non-negative) integer, ok, but I really can't remember whether that's how it works, but since Decimal is used, I guess `fract` might return a nonzero value.


FWIW, I am pretty sure that in NNS, we once upon a time had (exchange rate) proposals whose weight was 0.1, so if you only voted on one of those, you would definitely get non-integer reward shares. Maybe, this type of thing never happened in SNS. If SNS never had this, then, ok.

rs/sns/governance/src/governance.rs
```diff
@@ -702,6 +702,21 @@ fn spawn_in_canister_env(future: impl Future<Output = ()> + Sized + 'static) {
     }
 }
 
+fn try_convert_reward_shares_to_u128(reward_shares: Decimal) -> Result<u128, String> {
```
daniel-wong-dfinity-org-twin: I really don't want a failure here to derail rewards. This is maybe another good reason to use some other type(s), not u128 et. al. I think the natural choice is Decimal itself, since then, this conversion goes away entirely. I mean, there was probably some reason (which I can no longer remember) that we chose Decimal before.

rs/sns/governance/src/governance.rs
```diff
@@ -702,6 +702,21 @@ fn spawn_in_canister_env(future: impl Future<Output = ()> + Sized + 'static) {
...
+            "Reward shares must be non-negative: {reward_shares}."
+        ));
+    }
+
+    if reward_shares.fract() != Decimal::ZERO {
+        return Err(format!("Reward shares must be integral: {reward_shares}."));
+    }
+
+    u128::try_from(reward_shares)
+        .map_err(|err| format!("Cannot represent reward shares {reward_shares} as u128: {err}."))
```
daniel-wong-dfinity-org-twin: Look, there are three failure paths here. That is a bit scary, and afaict, unnecessary.

rs/sns/governance/src/governance.rs
```diff
@@ -5972,6 +5991,22 @@ impl Governance {
                     }
                 };
 
+                if neuron_reward_shares > dec!(0) {
+                    let reward_shares = try_convert_reward_shares_to_u128(neuron_reward_shares)
+                        .unwrap_or_else(|err| {
+                            panic!(
```
daniel-wong-dfinity-org-twin: See, I really do not want this to reduce the reliability of voting rewards. Sorry if I'm beating a dead horse here... Even if there is no way to avoid a fallible conversion, I'm pretty sure that can at least skip instead of panic.

rs/sns/governance/src/types.rs
```diff
@@ -2178,6 +2178,21 @@ impl From<u64> for ProposalId {
     }
 }
 
+impl From<u128> for Uint128 {
+    fn from(value: u128) -> Self {
+        Self {
+            high: (value >> 64) as u64,
+            low: value as u64,
```
daniel-wong-dfinity-org-twin: I don't think it is well known what happens when you convert from u128 down to u64 when the value is too large to be represented as a u64. Even if you do some experiments to confirm, no experiment can definitively tell you that something has undefined behavior.

(ofc, if you get a seg fault, then, that's a pretty strong sign you have hit UB, but if you don't get seg fault, you could still be in the wonderful world of UB land, but it just hasn't revealed itself to you in a spectacular way.)

Therefore, what I would rather do is _explicitly_ mask out the upper bits. E.g.

```
// In binary, this looks like 000...111... with 64x 0s a the top and 64x 1s at the bottom.
const LOWER_MASK: u128 = (1_u128 << 64) - 1;

low: (value & MASK) as u64,
```

Ofc, to the machine, masking is superfluous. To human readers, I think it helps. At least, it helps me.

rs/sns/governance/canister/governance.did
aodl: Thanks `nat`'s much cleaner, sounds good. I can address this after we decide what to do about the Decimal (I'll have a think before responding to your other comment about using Decimal)

rs/sns/governance/src/governance.rs
aodl: `reward_shares` can only be an integer. I wonder if it would make more sense to accumulate `reward_shares` as a `u128` instead of as a `Decimal`, and only convert the numerator/denominator to `Decimal` when doing that [fractional calculation](https://github.com/dfinity/ic/blob/ce4b27fb0587960d2629d086c2aa0cfa52244c58/rs/sns/governance/src/governance.rs#L5977) (this is the only reason decimal seems to have been used). Then we can stick with your suggestion for `nat` above.

What do you think? I'm leaning this way because `Decimal` deserialization would be a little bit less consumer-friendly

rs/sns/governance/src/governance.rs
aodl: Fair point. There's no per-proposal weighting in the SNS like the NNS has, so `reward_shares` can only be integers. The only reason I can see that `Decimal` is used is to support the calculation [`rewards_purse_e8s * (neuron_reward_shares / total_reward_shares);`](https://github.com/dfinity/ic/blob/ce4b27fb0587960d2629d086c2aa0cfa52244c58/rs/sns/governance/src/governance.rs#L5977). I wonder if it would make more sense to ditch the `Decimal` altogether? (see comment above)

rs/sns/governance/src/types.rs
aodl: Yep, I believe the narrowing cast is well-defined in Rust (it truncates the upper bits), but this conversion can go away entirely with the nat/protobuf-bytes change.

rs/sns/governance/src/governance.rs
daniel-wong-dfinity-org-twin: Ahh. Maybe that's why: because eventually, this gets divided by the total reward shares.

I would just log if fract returns nonzero, truncate, and comment why we expect this to be 0.

rs/sns/governance/src/governance.rs
daniel-wong-dfinity-org-twin: I like Decimal, because you can write `pot * (my_shares / total_shares)` instead of `pot * my_shares / total_shares`. The first way makes it clear that `pot` is being multiplied by some number in the "unit interval". I mean, this is a slight preference (I do not super hate integers), but since the code is already in the preferred way, I wouldn't change it.

rs/sns/governance/src/types.rs
daniel-wong-dfinity-org-twin: Yes, but that was not my point. The point is, do people actually KNOW this with CERTAINTY? My guess is "no". At least, I don't. Under such uncertainty, I would default to, "don't chance it".

daniel-wong-dfinity-org-twin:
Sorry for the slowness. Swamped.

aodl:
> Sorry for the slowness. Swamped.

No worries, I've been busy too. I've implemented the changes we discussed now though. Please let me know if there's anything further needed or anything I can do to make this easier to review

rs/sns/governance/canister/governance.did
aodl: Done: the public API now uses candid `nat`. The existing reward calculation remains Decimal and the integral participation value is stored as `BigUint::to_bytes_be()` in protobuf

rs/sns/governance/canister/governance.did
aodl: Done: `reward_event_end_timestamp_seconds` is now `opt nat64` in the candid API

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: The design changes are in now and I simplified the focused tests to match the final implementation. Ready for another pass when you get a chance.

rs/sns/governance/src/pb/conversions.rs
aodl: Done: `Uint128` and its conversions have been removed

rs/sns/governance/src/pb/conversions.rs
aodl: Done: this now uses `BigUint::from_bytes_be` and exposes the result as `candid::Nat`

rs/sns/governance/src/governance.rs
aodl: Done: the existing Decimal reward calculation is unchanged. If the fractional part is unexpectedly non-zero, it's logged and only the exposed participation value is truncated; native rewards continue using the original Decimal

rs/sns/governance/src/governance.rs
aodl: Done: there is no longer a panic or fallible conversion in the participation-recording path

rs/sns/governance/src/types.rs
aodl: Thanks, the Uint128 representation and these conversions have been removed entirely with the nat/protobuf-bytes change you suggested

daniel-wong-dfinity-org-twin:
👀 

BTW, I will be going on vacation starting after work on Friday. If we cannot finish this before then, perhaps, Björn A. can take this over the finish line.

rs/sns/governance/api/src/ic_sns_governance.pb.v1.rs
```diff
@@ -245,13 +234,16 @@ pub mod neuron {
...
+    /// Participation in a reward event. Governance currently populates both optional fields.
+    /// Consumers should use the shares only when both fields are present and the timestamp matches
+    /// `latest_reward_event.end_timestamp_seconds`.
     #[derive(Default, candid::CandidType, candid::Deserialize, Debug, Clone, PartialEq)]
     pub struct RewardEventParticipation {
         /// The end timestamp of the reward event that calculated these shares.
-        pub reward_event_end_timestamp_seconds: u64,
-        /// The sum of the neuron's canonical ballot voting power over all
+        pub reward_event_end_timestamp_seconds: Option<u64>,
+        /// The non-negative sum of the neuron's canonical ballot voting power over all
```
daniel-wong-dfinity-org-twin:
```suggestion
        /// The sum of the neuron's voting power (weighted by proposal) over all
```

I would like to call it "reward shares", but I don't think that term is used outside the code. VP is the closest thing. Not sure what you mean by "canonical". AFAIK, other than weighting by proposal, there is only one kind of VP.

rs/sns/governance/api/src/ic_sns_governance.pb.v1.rs
```diff
@@ -149,10 +139,9 @@ pub struct Neuron {
     /// The neuron's followees, specified as a map of proposal topics IDs to followees neuron IDs.
     pub topic_followees: Option<neuron::TopicFollowees>,
     /// The neuron's positive reward shares from its most recent participating reward event,
-    /// tagged with that event's end timestamp. Consumers must compare this timestamp with
-    /// `latest_reward_event.end_timestamp_seconds`. An absent value or a different timestamp means
-    /// that the neuron had zero shares in the target event. A neuron that did not participate in a
-    /// newer event may retain its older tagged value.
+    /// tagged with that event's end timestamp. An absent participation or a timestamp that differs
+    /// from `latest_reward_event.end_timestamp_seconds` means zero shares for that event. A neuron
+    /// may retain participation from an older event.
```
daniel-wong-dfinity-org-twin:
```suggestion
    /// might have a stale value here, from a previous voting reward event.
```

"Retain" makes it sound like we intentionally keep it around, but it's really the opposite: we don't bother to clean up old values (but we could).

rs/sns/governance/src/governance.rs
```diff
@@ -5992,18 +5978,30 @@ impl Governance {
...
+                    let truncated_reward_shares = neuron_reward_shares.trunc();
+                    // After truncation, the Decimal has scale zero, so its mantissa is the exact
+                    // integer value exposed as a Candid nat.
+                    let reward_shares =
+                        BigUint::from(truncated_reward_shares.mantissa().unsigned_abs());
 
                     neuron.latest_reward_event_participation = Some(RewardEventParticipation {
                         reward_event_end_timestamp_seconds,
-                        reward_shares: Some(Uint128::from(reward_shares)),
+                        reward_shares: reward_shares.to_bytes_be(),
```
daniel-wong-dfinity-org-twin:
```suggestion
                    let reward_shares =
                        BigUint::from(reward_shares
                            // Truncates. Other than that, this is not lossy,
                            // because reward_shares is the sum of u64s.
                            .as_i128()
                            // Converts to u128. We know that sign is not lost here,
                            // because we are in the > 0 branch.
                            .unsigned_abs()
                        )
                        .to_bytes_be();

                    neuron.latest_reward_event_participation = Some(RewardEventParticipation {
                        reward_event_end_timestamp_seconds,
                        reward_shares,
```

rs/sns/integration_tests/src/neuron.rs
```diff
@@ -1617,18 +1618,19 @@ async fn couple_of_neurons_who_voted_get_rewards() {
             .neurons
             .get(&neuron.id.as_ref().unwrap().to_string())
             .unwrap();
+        let public_neuron = pb_api::Neuron::from(neuron.clone());
```
daniel-wong-dfinity-org-twin:
```suggeestion
        let api_neuron = pb_api::Neuron::from(neuron.clone());
```

In NNS, "public neuron" means something else, so `public_neuron` is confusing, even if not technically ambiguous.

rs/sns/integration_tests/src/neuron.rs
```diff
@@ -1617,18 +1618,19 @@ async fn couple_of_neurons_who_voted_get_rewards() {
...
             assert_eq!(
                 participation.reward_event_end_timestamp_seconds,
-                reward_event_end_timestamp_seconds,
-            );
-            assert_eq!(
-                u128::from(participation.reward_shares.unwrap()),
-                u128::from(weight),
+                Some(reward_event_end_timestamp_seconds),
             );
+            assert_eq!(participation.reward_shares, Some(candid::Nat::from(weight)),);
```
daniel-wong-dfinity-org-twin: I think this could be done as one assert_eq, and if so that would be preferrable.

```suggestion
            assert_eq!(
                participation,
                RewardEventParticipation {
                    reward_event_end_timestampseconds: Some(reward_event_end_timestamp_seconds),
                    reward_shares: Some(candid::Nat::from(weight)),
                },
                "{:#?}",
                neuron,
            );
```

rs/sns/governance/src/governance/assorted_governance_tests.rs
```diff
@@ -4329,6 +4330,17 @@ async fn test_split_neuron_succeeds() {
...
+    let original_participation = neuron::RewardEventParticipation {
+        reward_event_end_timestamp_seconds: 123,
+        reward_shares: BigUint::from(99_u128).to_bytes_be(),
+    };
+    setup
+        .governance
+        .proto
+        .neurons
+        .get_mut(&setup.neuron_id.to_string())
+        .expect("Missing orig neuron!")
```
daniel-wong-dfinity-org-twin: I think we usually just do `.unwrap()` in tests.

rs/sns/integration_tests/src/neuron.rs
```diff
@@ -1616,6 +1618,20 @@ async fn couple_of_neurons_who_voted_get_rewards() {
...
+            let participation = public_neuron
+                .latest_reward_event_participation
+                .as_ref()
+                .unwrap();
+            assert_eq!(
+                participation.reward_event_end_timestamp_seconds,
+                Some(reward_event_end_timestamp_seconds),
+            );
+            assert_eq!(participation.reward_shares, Some(candid::Nat::from(weight)),);
+        }
```
daniel-wong-dfinity-org-twin: I think this could be done as one assert_eq, and if so that would be preferrable.

```suggestion
            let expected_participation = if weight == 0 {
                None
            } else {
                Some(RewardEventParticipation {
                    reward_event_end_timestampseconds: Some(reward_event_end_timestamp_seconds),
                    reward_shares: Some(candid::Nat::from(weight)),
                })
            };
            assert_eq!(
                pb_api::Neuron::from(neuron.clone()).latest_reward_event_participation,
                expected_participation,
                "{:#?}",
                neuron,
            );
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
+use super::{test_helpers::*, *};
+use ic_nervous_system_canisters::cmc::FakeCmc;
+use ic_nervous_system_common::ONE_DAY_SECONDS;
+use ic_sns_governance_api::pb::v1 as pb_api;
+use maplit::{btreemap, btreeset};
+use num_bigint::BigUint;
+use prost::Message;
+
+use crate::pb::v1::{Motion, VotingRewardsParameters};
+use crate::types::test_helpers::NativeEnvironment;
```
daniel-wong-dfinity-org-twin: Consolidate, and do not separate using a blank line. I think this is our usual way, although, to be fair, we are not hyper consistent on this.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        wait_for_quiet_state: Some(WaitForQuietState::default()),
+        is_eligible_for_rewards: true,
+        ..Default::default()
+    }
+}
+
+fn governance_with_neurons(neurons: Vec<Neuron>) -> (Governance, u64) {
+    let round_duration_seconds = ONE_DAY_SECONDS;
+    let mut environment = NativeEnvironment::new(Some(CanisterId::from_u64(1)));
+    environment.now = 11 * round_duration_seconds;
```
daniel-wong-dfinity-org-twin: This is an unrealistically small value. I mean, it probably doesn't matter, but when it is easy to use realistic values in test, do it.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let mut governance = Governance::new(
+        proto.try_into().unwrap(),
+        Box::new(environment),
+        Box::new(DoNothingLedger {}),
+        Box::new(DoNothingLedger {}),
+        Box::new(FakeCmc::new()),
+    );
+    governance.env.set_time_warp(TimeWarp {
+        delta_s: i64::try_from(round_duration_seconds).unwrap(),
+    });
```
daniel-wong-dfinity-org-twin: This seems pointless? I mean, you are already controlling time via environment, right? Not harmful either though.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+
+    (governance, round_duration_seconds)
```
daniel-wong-dfinity-org-twin: Not sure why return round duration if it's fixed to the usual value. Better to have a global

```
// 1 day is the usual value, at least for SNSs that take after NNS.
const ROUND_DURATION_SECONDS: u64 = ONE_DAY_SECONDS;
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+fn reward_shares(neuron: &Neuron) -> Option<BigUint> {
```
daniel-wong-dfinity-org-twin: I have a slight preference for inlining this. This seems like "too much" indirection to me.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    };
+    let mut ballots = voting_powers
+        .iter()
+        .map(|(neuron_id, voting_power)| {
+            (
+                neuron_id.to_string(),
+                Ballot {
+                    vote: Vote::Unspecified as i32,
+                    voting_power: *voting_power,
+                    cast_timestamp_seconds: 1,
```
daniel-wong-dfinity-org-twin: I guess it does not matter, but Unspecified and nonzero cast_timestamp_seconds seems inconsistent...

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    Governance::cast_vote_and_cascade_follow(
+        &ProposalId { id: proposal_id },
+        alice_id,
+        alice_vote,
+        motion_function_id,
+        &function_followee_index,
+        &btreemap! {},
```
daniel-wong-dfinity-org-twin: Can you add a comment at the end of the line with the parameter's name (or a description thereof)?

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        neurons,
+        1,
```
daniel-wong-dfinity-org-twin: Ditto.

Ditto elsewhere.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+                Ballot {
+                    vote: Vote::Unspecified as i32,
+                    voting_power: *voting_power,
+                    cast_timestamp_seconds: 1,
+                },
+            )
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    Governance::cast_vote_and_cascade_follow(
```
daniel-wong-dfinity-org-twin, line 118:
```suggestion
    // Alice casts her vote. Follower follows.
    Governance::cast_vote_and_cascade_follow(
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        alice_vote,
+        motion_function_id,
+        &function_followee_index,
+        &btreemap! {},
+        neurons,
+        1,
+        &mut ballots,
+        Topic::Governance,
+    );
+    Governance::cast_vote_and_cascade_follow(
```
daniel-wong-dfinity-org-twin, line 131:
```suggestion

    // Bob casts his vote.
    Governance::cast_vote_and_cascade_follow(
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let reward_event = governance.proto.latest_reward_event.as_ref().unwrap();
+    assert_eq!(reward_event.distributed_e8s_equivalent, 0);
+    assert_eq!(
+        reward_event.settled_proposals,
+        vec![
+            ProposalId { id: 1 },
+            ProposalId { id: 2 },
+            ProposalId { id: 3 },
+        ]
+    );
```
daniel-wong-dfinity-org-twin, line 236: If this could be one assert_eq, that would be preferred.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let event_timestamp_seconds = reward_event.end_timestamp_seconds.unwrap();
+
+    for (neuron_id, expected_shares) in [
+        (
+            &alice_id,
+            Some(BigUint::from(u64::MAX) + BigUint::from(1_u8)),
```
daniel-wong-dfinity-org-twin: Wouldn't it make more sense to express this as 2 * max + 2?

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        Vote::Yes,
+    );
+    let proposal_3_ballots = cascaded_ballots(
+        3,
+        &governance.proto.neurons,
+        &btreemap! {
+            alice_id.clone() => 2,
+            bob_id.clone() => 0,
+            carol_id.clone() => 0,
+            follower_id.clone() => 0,
```
daniel-wong-dfinity-org-twin: 0 seems unnecessarily strange and unrealistic.

It is also weird that alice goes from massive voting power down to 2. Not sure why?

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+            .proposals
+            .get(&3)
+            .unwrap()
+            .ballots
+            .is_empty()
+    );
+}
+
+#[test]
+fn test_replaces_only_positive_participants_and_retains_older_event_tags() {
```
daniel-wong-dfinity-org-twin: I don't think we need to assert that neurons that didn't participate have an old record. The important thing is that they do NOT have a new record (either, the don't have a participation record at all, or it's an old one).

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let alice_id = neuron_id(1);
+    let bob_id = neuron_id(2);
+    let (mut governance, round_duration_seconds) =
+        governance_with_neurons(vec![neuron(&alice_id), neuron(&bob_id)]);
+    governance.proto.proposals.insert(
+        1,
+        proposal_data(
+            1,
+            btreemap! {
+                alice_id.to_string() => Ballot { vote: Vote::Yes as i32, voting_power: 10, ..Default::default() },
```
daniel-wong-dfinity-org-twin: Default is just to avoid the timestamp field? Explicit seems better.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+            .ballots
+            .is_empty()
+    );
+}
+
+#[test]
+fn test_replaces_only_positive_participants_and_retains_older_event_tags() {
+    // Step 1: Prepare two neurons and an event in which only Alice participates.
+    let alice_id = neuron_id(1);
+    let bob_id = neuron_id(2);
```
daniel-wong-dfinity-org-twin: These could be global constants. Would probably simplify some things/avoid repetition.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+            },
+        ),
+    );
+    governance.distribute_rewards(Tokens::from_e8s(0));
+    let event_1_timestamp_seconds = governance
+        .latest_reward_event()
+        .end_timestamp_seconds
+        .unwrap();
+
+    // Step 2: Advance one event and settle a proposal in which only Bob participates.
```
daniel-wong-dfinity-org-twin:
```suggestion
    // Step 2: Run code under test. Here, another voting reward round
    // occurs, except this time, only Bob participates.
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+            .get(&3)
+            .unwrap()
+            .ballots
+            .is_empty()
+    );
+}
+
+#[test]
+fn test_replaces_only_positive_participants_and_retains_older_event_tags() {
+    // Step 1: Prepare two neurons and an event in which only Alice participates.
```
daniel-wong-dfinity-org-twin:
```suggestion
    // Step 1: Prepare the world: two neurons and an voting reward round
    // in which only Alice participates.
```

It might be a little bit more interesting if Bob participates in the first round, because there might be a bug where we fail to clobber his first participation record.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+            2,
+            btreemap! {
+                alice_id.to_string() => Ballot { vote: Vote::Unspecified as i32, voting_power: 30, ..Default::default() },
+                bob_id.to_string() => Ballot { vote: Vote::No as i32, voting_power: 40, ..Default::default() },
+            },
+        ),
+    );
+    governance.distribute_rewards(Tokens::from_e8s(0));
+
+    // Step 3: Alice retains event 1; Bob is tagged with event 2.
```
daniel-wong-dfinity-org-twin:
```suggestion
    // Step 3: Verify results: Bob's participation gets updated. Alice's does not.
```

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let event_2_timestamp_seconds = governance
+        .latest_reward_event()
+        .end_timestamp_seconds
+        .unwrap();
+    assert_ne!(event_1_timestamp_seconds, event_2_timestamp_seconds);
```
daniel-wong-dfinity-org-twin: I mean, you could be more precise: t2 = t1 + round_duration

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    let bob = governance.proto.neurons.get(&bob_id.to_string()).unwrap();
+    assert_eq!(reward_shares(alice), Some(BigUint::from(10_u8)));
+    assert_eq!(
+        alice
+            .latest_reward_event_participation
+            .as_ref()
+            .unwrap()
+            .reward_event_end_timestamp_seconds,
+        event_1_timestamp_seconds,
+    );
```
daniel-wong-dfinity-org-twin: one assert_eq

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+    assert_eq!(reward_shares(bob), Some(BigUint::from(40_u8)));
+    assert_eq!(
+        bob.latest_reward_event_participation
+            .as_ref()
+            .unwrap()
+            .reward_event_end_timestamp_seconds,
+        event_2_timestamp_seconds,
+    );
```
daniel-wong-dfinity-org-twin: Ditto.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
```diff
@@ -0,0 +1,455 @@
...
+        Some(candid::Nat(reward_shares)),
+    );
+    assert_eq!(
+        Neuron::from(public_alice).latest_reward_event_participation,
+        Some(participation),
+    );
+}
+
+#[test]
+fn test_legacy_neuron_decodes_without_participation() {
```
daniel-wong-dfinity-org-twin: This seems unnecessary. I mean, this is basically just testing Prost, right? But we are not touching Prost here. I mean, this is not "actually" harmful either. FWIW, there are already tests that make sure .proto files are evolved in a safe way.

daniel-wong-dfinity-org-twin (commented):
Thanks again for this contribution, and for taking my nat suggestion.

I think all my suggestions in this round are minor and straightforward, so we should be able to approve quickly from here.

rs/sns/governance/api/src/ic_sns_governance.pb.v1.rs
aodl: Thanks, I dropped “canonical”. I left out “weighted by proposal” because SNS currently sums the eligible ballots’ voting power directly (there isn’t a separate per-proposal reward weight like the NNS has)

rs/sns/integration_tests/src/neuron.rs
aodl: renamed to api_neuron

rs/sns/governance/src/governance/assorted_governance_tests.rs
aodl: changed to .unwrap()

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the test now constructs the expected API participation and compares it with one assert_eq!

rs/sns/governance/src/governance.rs
aodl: Thanks, after syncing with current master Decimal::as_i128() is available, so I used the suggested conversion.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the test now constructs the expected participation and compares it with a single assert_eq!

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the test now asserts event_2_timestamp_seconds == event_1_timestamp_seconds + ROUND_DURATION_SECONDS

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: removed the legacy Prost compatibility test and the now-unused Prost import

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: updated the Step 3 comment to make it explicit that Bob’s participation is updated and Alice’s is not

rs/sns/integration_tests/src/neuron.rs
aodl: the test now constructs the expected participation and compares it with a single assert_eq!

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: consolidated the imports and removed the unnecessary blank-line separation

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the test helper now uses a realistic 100 * E8 neuron stake

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: removed the redundant setup time warp and initialize the reward-event timing directly in the test environment

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: added a shared ROUND_DURATION_SECONDS = ONE_DAY_SECONDS constant, and the helper no longer returns the round duration

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: removed the helper and inlined the participation comparison

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: Unspecified ballots now use cast_timestamp_seconds = 0; the timestamp is only set when a vote is actually cast

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: added parameter-name comments to the positional arguments to make the call sites clearer

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: applied the same parameter-name annotations to the corresponding call sites elsewhere in the test.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: added the comment clarifying that Alice casts her vote and the follower follows

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: added the comment clarifying that Bob casts his vote her

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: each neuron’s complete expected RewardEventParticipation is now compared with a single assert_eq!

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the expected value is now expressed directly from its contributions: twice maximum_reward_share_contribution plus the third proposal’s voting power. I also replaced the artificial 2 with a realistic voting-power value per your later comment

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: replaced the artificial zero/small voting powers with realistic nonzero values, including a realistic third contribution for Alice

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the test no longer requires Alice’s old participation record to remain; it only verifies that she is not tagged with the second reward event.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: the relevant ballots now set vote, voting_power, and cast_timestamp_seconds explicitly instead of relying on Default

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: moved the repeated neuron IDs, round duration, and realistic stake into shared test constants/statics to reduce repetition

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: updated the Step 2 comment to describe the second reward round and that only Bob participates.

rs/sns/governance/src/governance/reward_event_participation_tests.rs
aodl: Bob now participates in the first round as well, then participates again in the second round, so the test verifies that his previous participation record is overwritten.

aodl:
> Thanks again for this contribution, and for taking my nat suggestion.
> 
> I think all my suggestions in this round are minor and straightforward, so we should be able to approve quickly from here.

No worries, thanks for all of your feedback and making the time to review this. Please let me know if there's anything more I can do to get this over the line. I've addressed all of your feedback points above.

cc @bjoernek

daniel-wong-dfinity-org-twin:
back. will take a look at this today or so. Please, bear with me, since I need to clear a big backlog from when I was gone. Sorry for the delay.

rs/sns/governance/src/governance.rs
daniel-wong-dfinity-org-twin: Thanks.

Can you also include the suggested comments?

rs/sns/governance/src/governance/reward_event_participation_tests.rs
daniel-wong-dfinity-org-twin: What I meant was, you should use something like 1767225600 (Jan 1, 2026 midnight UTC). Whereas, 11 * round_duration_seconds is some time in the early 1970s, decades before ICP genesis.

daniel-wong-dfinity-org-twin (approved):
Phew! Sorry this has been kind of a saga.

rs/sns/governance/src/governance.rs
aodl: Added comments explaining the truncation and why unsigned_abs() is safe in the positive-shares branch

aodl:
> Phew! Sorry this has been kind of a saga.

Thanks for all your feedback! Hope you had a good holiday

PR #11080, opened 2026-08-10
fix: Enforce limit on claimed HTTP outcalls response size

eichhorl:
## Background
In HTTP outcalls, only the HTTP response bodies of flexible or non-replicated requests are gossiped. The signed metadata shares (containing the response size and hash) is always gossiped. When receiving such a response via gossip (in case of non-replicated & flexible requests) the size of the response is verified against the one claimed by the metadata. Additionally, a maximum size of `max_response_bytes + candid_overhead` is enforced, where `candid_overhead = 1KiB`.

## Problem.
When receiving an aggregated response as part of a block, the (claimed) response size isn't verified. This is relevant in the context of the new out-of-cycles error, introduced for the new (disabled) pricing in https://github.com/dfinity/ic/pull/11040. By claiming an oversized response size, a malicious replica could potentially blow up the minimum consensus cost required to deliver the response, and force an out-of-cycles error.

## Proposed Changes
With this PR we also enforce the size limit of shares received as part of blocks. An included shares that exceed this limit will cause the block to be invalidated.

For consistency, we enforce this limit across all response types: fully-/non-replicated responses, divergences, flexible ok/err responses.


rs/https_outcalls/consensus/src/payload_builder/proptests.rs
```diff
@@ -492,12 +492,12 @@ fn prop_flexible_kind(max_size: usize, subnet_size: usize) -> impl Strategy<Valu
 fn prop_content(max_size: usize) -> impl Strategy<Value = CanisterHttpResponseContent> {
     prop_oneof![
         (0..=max_size).prop_map(|size| CanisterHttpResponseContent::Success(vec![0; size])),
-        (0..=max_size).prop_map(
-            |size| CanisterHttpResponseContent::Reject(CanisterHttpReject {
+        (0..=max_size.min(MAXIMUM_CANISTER_HTTP_ERROR_MESSAGE_BYTES)).prop_map(|size| {
```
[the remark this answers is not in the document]
eichhorl: done

rs/interfaces/src/canister_http.rs
```diff
@@ -58,6 +58,15 @@ pub enum InvalidCanisterHttpPayloadReason {
...
         limit: Cycles,
     },
+    /// A share claims a `content_size` larger than the largest response the replica
+    /// could have produced, i.e. [`CanisterHttpRequestContext::max_http_outcall_content_size`].
+    ///
+    /// [`CanisterHttpRequestContext::max_http_outcall_content_size`]: ic_types::canister_http::CanisterHttpRequestContext::max_http_outcall_content_size
+    ContentSizeExceedsLimit {
+        callback_id: CallbackId,
+        content_size: u32,
+        limit: u64,
```
alin-at-dfinity, line 68: Why is `limit` a `u64` if `content_size` is only a `u32`?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -408,29 +393,16 @@ impl CanisterHttpPoolManagerImpl {
                     self.requested_id_cache.borrow_mut().remove(&response.id);
                     self.metrics.shares_signed.inc();
 
-                    if matches!(
-                        context.replication,
-                        Replication::NonReplicated(_) | Replication::Flexible { .. }
-                    ) {
-                        if let Err(err) =
-                            validate_response_size(&response, context.max_response_bytes)
-                        {
```
pierugo-dfinity, line 417: Do you know why we weren't validating the response size for `FullyReplicated`?

alin-at-dfinity (approved):

pierugo-dfinity (approved):

rs/interfaces/src/canister_http.rs
eichhorl: `u64` is the return type of the function computing the limit, which is forced by the `max_response_bytes` API argument of `CanisterHttpRequestArgs` being `u64`:

https://github.com/dfinity/ic/blob/3faa347cbaf5cf6424b8034cb0ba1591e0052391/rs/types/management_canister_types/src/http.rs#L99-L101

Technically, it has to be smaller than 2MB, but internally we also use the `NumBytes` type, which is again `u64`...

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: It's isn't really necessary for `FullyReplicated`, because we anyway need at least `2f+1` shares that agree on the same metadata, and at most `f` nodes can create a metadata with a content size that is too large.

PR #11083, opened 2026-08-10
feat(cketh): burn-first accounting for sweeper fee funding

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding), first of the four remaining PRs. Targets `master`, now that #11072 has merged.

## Why

Sweep gas is prepaid: ckETH is burned from the minter's fee subaccount before the ETH moves, so that at every instant

```
cumulative ckETH burned for sweeping >= cumulative ETH debited from the main address for sweeping
```

Nothing yet keeps track of either side, so nothing can check it. This PR adds that bookkeeping, plus the bounds deciding when a top-up is due.

## What

The accounting is a fold over events the minter already persists — the accepted funding request and the finalized transaction's receipt — so it is reconstructed exactly on replay and needs no event type of its own. It is deliberately not serializable, which keeps that property honest.

Writing it surfaced something worth stating plainly: a surplus arises on **every** funding, not only failed ones. The value transferred is the burn minus the transaction's *max* fee, while only the *effective* fee is ever spent, so the unused fee allowance stays at the main address. Nothing reuses it — each funding burns for its own transfer, and the surplus simply stays as ckETH backing, exactly like the unspent gas of a user withdrawal. A failed funding is the extreme case of the same thing.

That surplus is tracked rather than read back from the chain because it sits at the **main** address, not the sweeper's. The sweeper's on-chain balance answers "how much prepaid gas is in place", which is a different question from "how far has burn run ahead of spend" — the quantity the invariant is about, and the one an operator needs in order to check it.

Spending more than was burned would mean ckETH is under-backed, so that traps rather than saturating to zero — and it is checked eagerly at each finalized funding, so a violation surfaces at the transition that caused it rather than whenever someone next reads the surplus.

## The bounds

Proposal-configurable, validated as a **pair** rather than individually: a target at or below the low-water mark would make a funding immediately due again and loop, and headroom below the minimum burn would make every cycle burn more ckETH than the ETH it moves. Both are rejected wholesale rather than partially applied.

The same check runs when only the *minimum withdrawal amount* changes, since the invariant relates two independently configurable amounts and raising one alone would silently invalidate bounds that were valid when set.

Defaults are 0.02 / 0.1 ETH — deliberately provisional, sized so a funding covers many sweeps and its own fee stays a small fraction of the amount moved. They are meant to be calibrated during the Sepolia rollout once real sweep gas costs are known.

## Stack

| # | PR | Status |
|---|----|--------|
| 1 | Burn-first accounting for sweeper fee funding | **this PR** |
| 2 | #11086 — Sweeper fee-funding task, with an end-to-end test | draft |
| 3 | #11094 — Sweeper funding observability and the prepaid-gas gate | draft |
| 4 | #11097 — Adversarial end-to-end coverage of sweeper fee funding | draft |

### DIFF
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -153,10 +153,12 @@ type UpgradeArg = record {
     ethereum_sweeper_contract_address : opt text;
 
     // Sweeper balance (in wei) below which the minter tops it up with gas for sweeping.
-    // Must stay strictly below sweeper_funding_target, otherwise funding would loop.
+    // Validated as a pair with sweeper_funding_target: the target must exceed this by at least
+    // minimum_withdrawal_amount, since that difference is the smallest amount a funding ever moves.
+    // Setting only one of the two keeps the other's current value, and the pair is validated again.
     sweeper_funding_low_water_mark : opt nat;
 
-    // Sweeper balance (in wei) the minter tops up to.
+    // Sweeper balance (in wei) the minter tops up to. See the constraint above.
     sweeper_funding_target : opt nat;
 };
 
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -152,14 +152,6 @@ type UpgradeArg = record {
     // Change the sweeper smart contract address.
     ethereum_sweeper_contract_address : opt text;
 
-    // Sweeper balance (in wei) below which the minter tops it up with gas for sweeping.
-    // Validated as a pair with sweeper_funding_target: the target must exceed this by at least
-    // minimum_withdrawal_amount, since that difference is the smallest amount a funding ever moves.
-    // Setting only one of the two keeps the other's current value, and the pair is validated again.
-    sweeper_funding_low_water_mark : opt nat;
-
-    // Sweeper balance (in wei) the minter tops up to. See the constraint above.
-    sweeper_funding_target : opt nat;
 };
 
 type MinterArg = variant { UpgradeArg : UpgradeArg; InitArg : InitArg };
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -151,6 +151,13 @@ type UpgradeArg = record {
 
     // Change the sweeper smart contract address.
     ethereum_sweeper_contract_address : opt text;
+
+    // Sweeper balance (in wei) below which the minter tops it up with gas for sweeping.
+    // Must stay strictly below sweeper_funding_target, otherwise funding would loop.
+    sweeper_funding_low_water_mark : opt nat;
+
+    // Sweeper balance (in wei) the minter tops up to.
+    sweeper_funding_target : opt nat;
 };
 
 type MinterArg = variant { UpgradeArg : UpgradeArg; InitArg : InitArg };
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -151,7 +151,6 @@ type UpgradeArg = record {
 
     // Change the sweeper smart contract address.
     ethereum_sweeper_contract_address : opt text;
-
 };
 
 type MinterArg = variant { UpgradeArg : UpgradeArg; InitArg : InitArg };
--- a/rs/ethereum/cketh/minter/src/lifecycle/init.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/init.rs
@@ -120,6 +120,8 @@ impl TryFrom<InitArg> for State {
             log_scrapings,
             automatic_deposits: AutomaticDeposits::default(),
             sweeper_contract_address,
+            sweeper_funding: Default::default(),
+            sweeper_funding_config: Default::default(),
         };
         state.validate_config()?;
         Ok(state)
--- a/rs/ethereum/cketh/minter/src/lifecycle/init.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/init.rs
@@ -121,7 +121,6 @@ impl TryFrom<InitArg> for State {
             automatic_deposits: AutomaticDeposits::default(),
             sweeper_contract_address,
             sweeper_funding: Default::default(),
-            sweeper_funding_config: Default::default(),
         };
         state.validate_config()?;
         Ok(state)
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -32,14 +32,6 @@ pub struct UpgradeArg {
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
     #[n(10)]
     pub ethereum_sweeper_contract_address: Option<String>,
-    /// Sweeper balance below which the minter tops it up with gas. Validated as a pair with
-    /// `sweeper_funding_target`: the target must exceed this by at least the minimum withdrawal
-    /// amount, that difference being the smallest amount a funding ever moves.
-    #[cbor(n(11), with = "icrc_cbor::nat::option")]
-    pub sweeper_funding_low_water_mark: Option<Nat>,
-    /// Sweeper balance the minter tops up to. See the constraint on the low-water mark.
-    #[cbor(n(12), with = "icrc_cbor::nat::option")]
-    pub sweeper_funding_target: Option<Nat>,
 }
 
 pub fn post_upgrade(upgrade_args: Option<UpgradeArg>) {
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -32,10 +32,12 @@ pub struct UpgradeArg {
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
     #[n(10)]
     pub ethereum_sweeper_contract_address: Option<String>,
-    /// Sweeper balance below which the minter tops it up with gas.
+    /// Sweeper balance below which the minter tops it up with gas. Validated as a pair with
+    /// `sweeper_funding_target`: the target must exceed this by at least the minimum withdrawal
+    /// amount, that difference being the smallest amount a funding ever moves.
     #[cbor(n(11), with = "icrc_cbor::nat::option")]
     pub sweeper_funding_low_water_mark: Option<Nat>,
-    /// Sweeper balance the minter tops up to. Must exceed the low-water mark.
+    /// Sweeper balance the minter tops up to. See the constraint on the low-water mark.
     #[cbor(n(12), with = "icrc_cbor::nat::option")]
     pub sweeper_funding_target: Option<Nat>,
 }
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -32,6 +32,12 @@ pub struct UpgradeArg {
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
     #[n(10)]
     pub ethereum_sweeper_contract_address: Option<String>,
+    /// Sweeper balance below which the minter tops it up with gas.
+    #[cbor(n(11), with = "icrc_cbor::nat::option")]
+    pub sweeper_funding_low_water_mark: Option<Nat>,
+    /// Sweeper balance the minter tops up to. Must exceed the low-water mark.
+    #[cbor(n(12), with = "icrc_cbor::nat::option")]
+    pub sweeper_funding_target: Option<Nat>,
 }
 
 pub fn post_upgrade(upgrade_args: Option<UpgradeArg>) {
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -122,6 +122,7 @@ pub struct State {
     /// Address of the sweeper smart contract on Ethereum, which the minter
     /// delegates to when sweeping funded deposit addresses.
     pub sweeper_contract_address: Option<Address>,
+
     /// Burn-first accounting for sweeper fee funding, folded from the audit events.
     pub sweeper_funding: SweeperFundingAccounting,
 }
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -13,6 +13,7 @@ use crate::numeric::{
 };
 use crate::state::automatic_deposits::{AutomaticDeposits, ScanProgress};
 use crate::state::eth_logs_scraping::{LogScrapingId, LogScrapings};
+use crate::state::sweeper_funding::{SweeperFundingAccounting, SweeperFundingConfig};
 use crate::state::transactions::{Erc20WithdrawalRequest, TransactionCallData, WithdrawalRequest};
 use crate::timed_sized_map::{Entry, Timestamp};
 use crate::tx::GasFeeEstimate;
@@ -32,6 +33,7 @@ pub mod audit;
 pub mod automatic_deposits;
 pub mod eth_logs_scraping;
 pub mod event;
+pub mod sweeper_funding;
 pub mod transactions;
 
 #[cfg(test)]
@@ -120,6 +122,10 @@ pub struct State {
     /// Address of the sweeper smart contract on Ethereum, which the minter
     /// delegates to when sweeping funded deposit addresses.
     pub sweeper_contract_address: Option<Address>,
+    /// Burn-first accounting for sweeper fee funding, folded from the audit events.
+    pub sweeper_funding: SweeperFundingAccounting,
+    /// When to top the sweeper address up, and to what (proposal-configurable).
+    pub sweeper_funding_config: SweeperFundingConfig,
 }
 
 #[derive(Eq, PartialEq, Debug)]
@@ -131,6 +137,7 @@ pub enum InvalidStateError {
     InvalidMinimumWithdrawalAmount(String),
     InvalidLastScrapedBlockNumber(String),
     InvalidLastErc20ScrapedBlockNumber(String),
+    InvalidSweeperFundingConfig(String),
 }
 
 #[derive(Clone, Eq, PartialEq, Debug)]
@@ -181,6 +188,16 @@ impl State {
             EthereumNetwork::Mainnet => Wei::new(2_000_000_000_000),
             EthereumNetwork::Sepolia => Wei::new(10_000_000_000),
         };
+        self.sweeper_funding_config
+            .validate(self.cketh_minimum_withdrawal_amount)
+            .map_err(|e| {
+                InvalidStateError::InvalidSweeperFundingConfig(format!(
+                    "ERROR: sweeper funding bounds are invalid for the minimum withdrawal \
+                     amount: {e:?}. Raise sweeper_funding_target (upgrade argument) so the \
+                     headroom above the low-water mark covers the minimum withdrawal amount. \
+                     Note InitArg cannot set the bounds, so at install the defaults apply."
+                ))
+            })?;
         if self.cketh_minimum_withdrawal_amount < cketh_ledger_transfer_fee {
             return Err(InvalidStateError::InvalidMinimumWithdrawalAmount(
                 "minimum_withdrawal_amount must cover ledger transaction fee, \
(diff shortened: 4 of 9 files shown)


rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,173 @@
...
+                 meaning ckETH is under-backed",
+            )
+    }
+
+    /// How much ckETH a funding of `amount` still has to burn, given the credit from earlier burns
+    /// that were never spent. Zero when the credit already covers it.
+    pub fn burn_required_for(&self, amount: Wei) -> Wei {
+        amount
+            .checked_sub(self.burned_not_yet_spent())
+            .unwrap_or(Wei::ZERO)
```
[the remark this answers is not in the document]
mbjorkqvist: Correct diagnosis, and it is exactly the change the next PR in the stack makes — documented in 49890c31d5 so the next reader does not have to re-derive it.

`SweeperFundingRequest` gains a `cketh_burned` field distinct from `withdrawal_amount`, for precisely the reason you give: with a single amount, reducing the burn shrinks the transfer by the same amount, so the credit is never consumed and grows forever instead. The accounting then records `cketh_burned` as the burn while `withdrawal_amount` stays the spend ceiling. This was found the hard way when the funding task was first written, and it is why the field exists.

Worth confirming the state of this PR though: `burn_required_for` has no production caller here, only unit tests — the funding task arrives with the field. So nothing computes a wrong burn at this commit; the helper is simply not reachable yet. That is the seam between this PR and the next, and it is the kind of thing splitting a feature into reviewable pieces produces.

I considered moving the function into the next PR so that this one contains nothing unusable, and decided against it: `burn_required_for` is what expresses the offsetting rule, which is the substance of this PR's accounting. Splitting it out would leave the accounting incomplete here and no clearer there. Instead its doc now states the requirement and that the field arrives with the task.

rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
```diff
@@ -30,6 +30,12 @@ pub struct UpgradeArg {
     pub deposit_with_subaccount_helper_contract_address: Option<String>,
     #[cbor(n(9), with = "icrc_cbor::nat::option")]
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
+    /// Sweeper balance below which the minter tops it up with gas.
+    #[cbor(n(10), with = "icrc_cbor::nat::option")]
+    pub sweeper_funding_low_water_mark: Option<Nat>,
+    /// Sweeper balance the minter tops up to. Must exceed the low-water mark.
+    #[cbor(n(11), with = "icrc_cbor::nat::option")]
+    pub sweeper_funding_target: Option<Nat>,
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 49890c31d5. You are right that a proposal following the documented rule could still be rejected — the doc gave the strict-inequality half of the check and omitted the headroom half.

Both `UpgradeArg` fields now state the complete constraint: the target must exceed the low-water mark by at least the minimum withdrawal amount, that difference being the smallest amount a funding ever moves. The low-water-mark field carries the full rule and the target refers to it, rather than repeating it in a way the two could drift apart.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -143,6 +143,13 @@ type UpgradeArg = record {
 
     // Change the last scraped block number of the deposit with subaccount helper smart contract.
     last_deposit_with_subaccount_scraped_block_number : opt nat;
+
+    // Sweeper balance (in wei) below which the minter tops it up with gas for sweeping.
+    // Must stay strictly below sweeper_funding_target, otherwise funding would loop.
+    sweeper_funding_low_water_mark : opt nat;
+
+    // Sweeper balance (in wei) the minter tops up to.
+    sweeper_funding_target : opt nat;
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 49890c31d5, together with the same omission on `UpgradeArg`.

The Candid comment now states the full constraint and one further thing a proposal author needs: setting only one of the two bounds keeps the other's current value, and the pair is validated again as a whole. That matters because it means a proposal changing one bound can be rejected on account of a value it did not set.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -151,6 +151,15 @@ type UpgradeArg = record {
...
     ethereum_sweeper_contract_address : opt text;
+
+    // Sweeper balance (in wei) below which the minter tops it up with gas for sweeping.
+    // Validated as a pair with sweeper_funding_target: the target must exceed this by at least
+    // minimum_withdrawal_amount, since that difference is the smallest amount a funding ever moves.
+    // Setting only one of the two keeps the other's current value, and the pair is validated again.
+    sweeper_funding_low_water_mark : opt nat;
+
+    // Sweeper balance (in wei) the minter tops up to. See the constraint above.
+    sweeper_funding_target : opt nat;
```
gregorydemay: understanding question: why do we need those 2 fields to be configurable? We could express both as a function of `minimum_withdrawal_amount`, no?
Example:

1. sweeper_funding_target to be 10* minimum_withdrawal_amount (about $100 on prod minter)
2. sweeper_funding_low_water_mark to be sweeper_funding_target/2

mbjorkqvist: Good point, makes sense! Done in b8ffc077ed: the target is 10x `minimum_withdrawal_amount`, and refilling starts at half the target. On mainnet that gives 0.3 / 0.15 ETH against the 0.03 minimum, where the fixed defaults were 0.1 / 0.02, so 3x more ETH parked at the sweeper. We could still adjust the constants if necessary.

rs/ethereum/cketh/minter/src/lifecycle/init.rs
```diff
@@ -120,6 +120,8 @@ impl TryFrom<InitArg> for State {
             log_scrapings,
             automatic_deposits: AutomaticDeposits::default(),
             sweeper_contract_address,
+            sweeper_funding: Default::default(),
+            sweeper_funding_config: Default::default(),
```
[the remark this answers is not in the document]
mbjorkqvist: Resolved by b8ffc077ed, which landed after the commit this review looked at (`f41cdcf394`).

The finding was right about the old shape: nothing persisted the bounds a deployment was installed with, so replaying a historical `Init` event would have adopted whatever the defaults said at the time of the upgrade. Rather than persist the pair, the pair is gone — the bounds are now a pure function of `minimum_withdrawal_amount`, which the `Init` and `Upgrade` events do carry, so replay reconstructs them exactly and there is no stored configuration left to drift. `init.rs` now seeds only the counters, which are folded from the audit events anyway, and the observation cache.

One residual, stated plainly: the multiplier (target = 10x the minimum, refill at half of it) is a code constant, so changing it does move every deployment's bounds at that upgrade. That is a deliberate, uniform change reviewed as part of a release, rather than deployments installed at different times ending up with silently different configurations — which was the hazard here.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -120,6 +122,9 @@ pub struct State {
     /// Address of the sweeper smart contract on Ethereum, which the minter
     /// delegates to when sweeping funded deposit addresses.
     pub sweeper_contract_address: Option<Address>,
+
+    /// Burn-first accounting for sweeper fee funding, folded from the audit events.
```
gregorydemay: nit: 
```suggestion
    /// Burn-first accounting for sweeper fee funding.
```

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,127 @@
+//! Burn-first accounting for sweeper fee funding, per
+//! `rs/ethereum/cketh/docs/deposit_from_cex.md`, "Fund the transaction fees without touching the
```
gregorydemay: nit: I would remove the mention of the design doc
```suggestion
//! Burn-first accounting for sweeper fee funding, "Fund the transaction fees without touching the
```

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,127 @@
...
+//! address, not the sweeper's, so it is tracked here rather than read back on chain.
+
+#[cfg(test)]
+mod tests;
+
+use crate::numeric::Wei;
+
+/// How much ckETH has been burned for sweeping and how much of it has actually been spent.
+///
+/// Not CBOR-serializable: rebuilt from the audit events on every upgrade, never persisted.
```
gregorydemay: nit:

```suggestion
```

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,127 @@
...
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct SweeperFundingAccounting {
+    cumulative_burned: Wei,
+    cumulative_transferred: Wei,
```
gregorydemay, line 25: There I would add some docs to detail the relation with cumulative_burned, how it evolves when a transaction is finalized, etc.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+/// When to top the sweeper address up, and to what. Derived from the minimum withdrawal amount, so
+/// that the gap between the two — the smallest amount a funding moves — clears the ledger minimum by
+/// construction.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub struct SweeperFundingConfig {
+    /// Fund the sweeper address once its ETH balance falls below this.
+    pub low_water_mark: Wei,
```
gregorydemay, line 95: Shouldn't we also have a field in `SweeperFundingAccounting` to track the balance of the sweeper address? This will be debited when the sweeper address does a sweep.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /// How much ETH to move to bring `sweeper_balance` up to the target, or `None` when the
+    /// balance is still above the low-water mark and no funding is due.
+    pub fn amount_due(&self, sweeper_balance: Wei) -> Option<Wei> {
+        if sweeper_balance >= self.low_water_mark {
+            return None;
+        }
+        // Non-zero: the balance is below the low-water mark, which is half the target.
+        self.target.checked_sub(sweeper_balance)
```
gregorydemay: nit: maybe instead of this comment it should be better to have an assert/expect since there is the invariant:

```rust
low_water_mark <= target
```

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    #[test]
+    fn should_keep_the_unspent_fee_as_surplus_after_a_successful_funding() {
+        let mut accounting = SweeperFundingAccounting::default();
+        accounting.record_burn(Wei::new(BURN));
+        accounting.record_finalized_funding(Wei::new(BURN - FEE), Wei::new(FEE / 2));
+
+        assert_eq!(
+            accounting.burned_not_yet_spent(),
+            Wei::new(FEE - FEE / 2),
```
gregorydemay: nit:

```suggestion
            Wei::new(FEE / 2),
```

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        assert_eq!(
+            config.target,
+            Wei::new(MINIMUM_BURN * SWEEPER_FUNDING_TARGET_IN_MINIMUM_WITHDRAWAL_AMOUNTS as u128)
+        );
+        assert_eq!(
+            config.low_water_mark,
+            config.target.checked_div_floor(2_u8).unwrap()
+        );
+    }
```
gregorydemay: I'm not sure of the worth of this test, it seems to just copy the implementation of the constructor.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[test]
+    fn should_fund_up_to_the_target() {
+        let config = config_for(MINIMUM_BURN);
+        let just_below = config.low_water_mark.checked_sub(Wei::ONE).unwrap();
+
+        assert_eq!(
+            config.amount_due(just_below),
+            Some(config.target.checked_sub(just_below).unwrap()),
+            "top up the shortfall to the target, not a fixed amount"
+        );
```
gregorydemay: That seems to just copy the implementation. How about the following proptest:

1. arbitrary balance amount between 0 and low_water_mark-1
2. assert balance+amount_due == target

rs/ethereum/cketh/minter/src/state/tests.rs
```diff
@@ -298,6 +298,51 @@ mod upgrade {
...
+        );
+    }
+
+    #[test]
+    fn should_install_with_a_minimum_the_fixed_bounds_would_have_refused() {
+        use crate::lifecycle::init::InitArg;
+        use crate::state::State;
+        use crate::test_fixtures::valid_init_arg;
+
+        let minimum = 90_000_000_000_000_000_u128; // above the 0.08 ETH the fixed bounds left
```
gregorydemay: I don't follow this test, why testing this particular value?

rs/ethereum/cketh/minter/src/state/tests.rs
```diff
@@ -1609,6 +1652,121 @@ mod eth_balance {
...
+                .unwrap(),
+            ledger_burn_index: LedgerBurnIndex::new(0),
+            from: "k2t6j-2nvnp-4zjm3-25dtz-6xhaa-c7boj-5gayf-oj3xs-i43lp-teztq-6ae"
+                .parse()
+                .unwrap(),
+            from_subaccount: crate::eth_logs::LedgerSubaccount::from_bytes(
+                crate::CKETH_FEE_SUBACCOUNT,
+            ),
+            created_at: Some(1699527697000000000),
+        };
```
gregorydemay: nit: IntelliJ warns for duplicate code, let's extract a helper method (in test fixtures)

rs/ethereum/cketh/minter/src/state/tests.rs
```diff
@@ -1609,6 +1652,121 @@ mod eth_balance {
...
+        WithdrawalFlow {
+            tx_status: TransactionStatus::Failure,
+            ..WithdrawalFlow::for_request(WithdrawalRequest::SweeperFunding(funding.clone()))
+        }
+        .apply(&mut state);
+
+        let spent = state.sweeper_funding.cumulative_spent();
```
gregorydemay, line 1719: shouldn't we check that the spend is exactly the effective transaction fee (from the tx receipt)?

gregorydemay (commented):
Thanks @mbjorkqvist ! Couple of nits but generally LGTM!

rs/ethereum/cketh/minter/src/state.rs
mbjorkqvist: Done in b1b4b726eb.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Done in b1b4b726eb.

One thing to flag while we are here: `eth_rpc_client/get_balance.rs` (merged in #11060) and `balance_scan/batcher/mod.rs` cite the same document the same way, so this file is now the odd one out. Happy to drop those references too if you would rather have none in the crate — just not in this PR.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Removed in b1b4b726eb.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Done in b1b4b726eb: each field now says at which transition it moves and by what — the burn at acceptance, the transferred amount and the fee at finalization — and that the latter two together are the spend, which never overtakes the burn.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: The balance is tracked, just not as a counter: #11094 caches it as an observation, refreshed by the funding task on every run — it reads the balance anyway to decide whether funding is due — and consulted through a gate that fails closed on absent, stale or insufficient.

I would rather not put a counter here. It would have to be debited at submission by the gas the sweep may spend, corrected to the effective fee at finalization, and raised again on each resubmission. That is the state machine `EthTransactions` already implements, but tied to the main address' nonce, keyed by ledger burn index, and debiting `eth_balance` at finalization, so sweeps want it generalised and instantiated a second time rather than shared — otherwise they land back on the withdrawal lane that `R17` keeps them off.

A counter here also could not be folded from the audit log: the events that debit it belong to S2 ([DEFI-2926](https://dfinity.atlassian.net/browse/DEFI-2926)) and do not exist yet, the starting balance is not in the log, and anyone can send ETH to an EOA. This struct is deliberately only what the log reproduces exactly, and only the main address' side: ETH counts as spent when it leaves main, whether the sweeper has spent it yet or not.

For S2 I would use `observed balance - provisioned gas of sweeps in flight`, kept with the sweep lane. An on-demand `eth_getBalance` does not avoid the second term — a read cannot see a sweep that has not been included — so overlapping sweeps need earmarks either way.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Done in b1b4b726eb — the comment is gone and the subtraction now carries the expect.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
mbjorkqvist: Done in b1b4b726eb.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
mbjorkqvist: Agreed, it did. Removed in b1b4b726eb; `should_leave_headroom_above_the_minimum_withdrawal_amount` is the one with content, since it asserts the relation the bounds exist for rather than the arithmetic that produces them.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
mbjorkqvist: Also agreed — done as you suggested in b1b4b726eb. It generates a balance across the whole target range, keeps the ones below the low-water mark, and asserts `balance + amount_due == target`.

rs/ethereum/cketh/minter/src/state/tests.rs
mbjorkqvist: Fair — the value only meant something against the bounds this PR deletes. 0.09 ETH used to exceed the 0.08 ETH of headroom the fixed defaults left, so an install with that minimum failed outright, and the test was there to show that derived bounds scale with it instead.

That is a motivation only visible in the history, which makes it a poor test to keep. Deleted in b1b4b726eb. The coverage that matters is `should_leave_headroom_above_the_minimum_withdrawal_amount`, which runs the derivation over minimums from 1 wei up to a thousand times mainnet's and asserts the relation holds for each.

---

Follow-up on this thread: deleting that test made me check what the state level still covers, and it turned up a real hole — not from the deletion, but from deriving the bounds in the first place.

`validate_config` gained a branch that refuses a minimum whose target would not fit, which is what lets `sweeper_funding_config()` unwrap. Only the helper returning `None` was covered, so the rejection itself was untested: removing the branch left the suite green. 74d0ea169a adds the case to `should_fail_when_upgrade_args_invalid`, next to the existing zero and below-the-ledger-fee ones, and I checked it fails without the branch.

For the record on what the level change cost: nothing else. `should_track_the_minimum_withdrawal_amount_in_the_funding_bounds` upgrades to the same 0.09 ETH the deleted test installed with and expects success, so a minimum the old fixed bounds would have refused is still asserted to be accepted — through `upgrade` rather than `try_from`, both of which funnel into `validate_config`.

rs/ethereum/cketh/minter/src/state/tests.rs
mbjorkqvist: Done in b1b4b726eb — `sweeper_funding_request(amount)` in `test_fixtures`, used by both.

rs/ethereum/cketh/minter/src/state/tests.rs
mbjorkqvist: Yes, better. Done in b1b4b726eb: the test keeps the receipt and asserts the spend equals `receipt.effective_transaction_fee()`, which also states the point — a failed funding moved no ETH, so the gas it paid is the whole spend.

gregorydemay (approved):
Thanks @mbjorkqvist !

PR #11085, opened 2026-08-10
feat: Non-flexible out of cycles error

eichhorl:
In https://github.com/dfinity/ic/pull/11040, we introduced an out-of-cycles error for flexible outcalls. This PR does the same for non- and fully-replicated outcalls.

A non-replicated outcall request will receive such an error, if the unspent allowance of its only share does not cover the consensus cost of including that share into a block.

A fully replicated outcall request will receive an out-of-cycles error, if the current set of known shares doesn't cover the consensus cost, and no share could be received that would add enough unspent allowance to change that. To do this, we also consider share which we haven't seen yet, which potentially may be for a response of size zero, or refund the entirety of their allowance.

Note that, unlike flexible outcalls, responses to non- and fully-replicated outcalls do not have structured ok and err types. Errors can only be expressed as part of the `Payload::Reject` message.


rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -241,6 +245,36 @@ fn min_delivery_cost(
...
+    }
+    let unseen = committee_size.saturating_sub(seen);
+
+    votes
+        .into_iter()
+        .map(|(metadata, votes)| (votes, metadata.content_size))
+        // A response nobody has voted for yet would have to come from the replicas that
+        // have not been seen, and may have an empty body.
+        .chain(std::iter::once((0, 0)))
+        .filter(|(votes, _)| votes + unseen >= threshold)
```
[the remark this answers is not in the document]
eichhorl: Yes, if the honest committee is split into two camps (on camp out-of-cycles, the other success), then the malicious nodes can decide that the out-of-cycles error will be delivered. The fact that they also could have chosen the success camp doesn't really change that.

rs/protobuf/def/types/v1/canister_http.proto
```diff
@@ -127,6 +134,7 @@ message CanisterHttpResponseMessage {
     CanisterHttpResponseDivergence divergence_response = 3;
     FlexibleCanisterHttpResponses flexible_responses = 4;
     FlexibleCanisterHttpError flexible_error = 5;
+    CanisterHttpOutOfCycles out_of_cycles = 6;
```
alin-at-dfinity, line 137: Just double-checking: This proto is never persisted in a checkpoint, only used in block payloads, right?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -556,6 +590,56 @@ mod tests {
...
 
+    #[test]
+    fn min_non_flexible_consensus_cost_is_zero_while_another_response_can_still_win() {
+        // A committee of 13 tolerating 4 faults, so 9 have to agree.
+        let n = NumberOfNodes::from(13);
+        let (committee, threshold) = (13, 9);
+        let votes: Vec<_> = (0..13).map(|i| share(i, 1_000, 0)).collect();
+
+        // The 4 replicas that have not voted for it are exactly enough, together with
+        // the 4 still unseen, to lift some other response to 9 instead.
```
alin-at-dfinity: I'm not sure I understand this: aren't we dealing with 4 votes for the same response and 9 nodes that haven't yet voted (are unseen)?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -556,6 +590,56 @@ mod tests {
...
+        assert_eq!(
+            min_non_flexible_consensus_cost(votes[..4].iter(), n, committee, threshold),
+            Some(Cycles::zero())
+        );
+        // One more vote and they are not: 9_490 * (181 + ...) -- no per-response
+        // overhead here, a single body is delivered: 9_490 * 1_000.
+        assert_eq!(
+            min_non_flexible_consensus_cost(votes[..5].iter(), n, committee, threshold),
+            Some(Cycles::new(9_490_000))
+        );
```
alin-at-dfinity: This is just for my own edification: what this test is saying that the consensus cost of 4 replicas voting for some non-empty response and 9 replicas voting for an empty response is zero (presumably because it only includes the latter 9 empty shares); but as soon as 5 replicas voted for the same non-empty response the cost of a successful response must be `9_490_000` (since it can only be whatever the former 5 voted on).

But how about a split response (or timeout, or whatever)? Couldn't that cost less than `9_490_000`? Or is this out of cycles error intended as a short circuit saying "you can no longer get an actual response out of this, only some form of error, so we're giving up"?

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1455,6 +1590,23 @@ fn flexible_error_into_consensus_response(
...
+    let total_spent: Cycles = shares.iter().map(|share| share.content.spent()).sum();
+    format!(
+        "Out of cycles: {} of the assigned replicas reported a collective spend of {} cycles, \
+         leaving {} cycles of the attached payment (after base fee deduction). \
+         Delivering a response would cost at least {} cycles.",
+        shares.len(),
+        total_spent,
+        unspent_allowance,
+        min_cost,
+    )
```
alin-at-dfinity, line 1604: Nit: You could also inline the local variables for better readability:

```suggestion
    format!(
        "Out of cycles: {} of the assigned replicas reported a collective spend of {total_spent} cycles, \
         leaving {unspent_allowance} cycles of the attached payment (after base fee deduction). \
         Delivering a response would cost at least {min_cost} cycles.",
        shares.len(),
    )
```

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -643,6 +676,91 @@ impl CanisterHttpPayloadBuilderImpl {
...
+            // agree on it for it to be delivered.
+            let (committee, threshold) = match &context.replication {
+                Replication::FullyReplicated => {
+                    let CanisterHttpCommittee {
+                        committee,
+                        threshold,
+                        ..
+                    } = self
+                        .membership
+                        .get_canister_http_committee(context.registry_version)
```
alin-at-dfinity, line 702: Idly wondering if it isn't worth memoizing the committee `BTreeSet` for every seen registry version during each call to `validate_canister_http_payload_impl` / `get_canister_http_payload_impl`.

Recalling the recent forum question about the 3000 concurrent HTTP outcalls limit, recomputing the committee (potentially repeatedly) for every open context / response may (or may not) result in significant overhead.

rs/https_outcalls/consensus/src/payload_builder/parse.rs
alin-at-dfinity: FWIW, this doc comment is no longer accurate.

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -3462,18 +3356,32 @@ fn into_messages_emits_initial_spend_reports() {
         min_responses: 1,
     };
 
+    let oom_callback = CallbackId::from(600);
+    let (_, oom_metadata) = test_response_and_metadata(oom_callback.get());
```
alin-at-dfinity: `oom` -> `ooc`?

alin-at-dfinity (approved):

rs/protobuf/def/types/v1/canister_http.proto
eichhorl: Correct, more importantly it is only part of data block payloads (and not summary block payloads).

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Thanks, that's a really bad comment. You are exactly right, I tried to clarify here: https://github.com/dfinity/ic/pull/11085/changes/a0a90f8119eebe5699dfaba08aa9cb6622ff0429#diff-e01a149a7f42e12ef1c1509e8af1093e60a2dd57b949cb40691db2c716117b8fR599-R608

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Yes, this function simply tells us the minimum cost that delivering a success response will have, given the currently seen and unseen shares. But of course we still need to wait until we actually receive enough shares to create the success response (it's just that we might be able to tell that we are out of cycles already before receiving all shares).

But indeed, other things could go wrong before the success response can be created. The context may time out, or we receive shares for too many different responses such that no majority will be achieved ever (in which case we produce a "divergence" response).

Since we check for the out of cycles case last, it should have the lowest priority among the three errors (timeout, divergence, out-of-cycles).

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Good idea, done 👍

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Yes, good point, I planned to look into performance a bit more after the feature works, so I will add that to the list.

rs/https_outcalls/consensus/src/payload_builder/parse.rs
eichhorl: Good catch, fixed

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Thanks!

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -201,64 +202,143 @@ pub(crate) struct OutOfCyclesProof {
...
+    let mut seen_signers = BTreeSet::new();
+    grouped_shares
+        .values()
+        .flatten()
+        .filter(|share| {
+            let signer = share.signature.signer;
+            committee.contains(&signer) && seen_signers.insert(signer)
+        })
+        .copied()
+        .collect()
```
pierugo-dfinity, line 313: Shouldn't we have already invalidated the duplicate shares in the past if a signer sent out multiple shares for the same request?

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity, line 584: Sorry, not your changes. 

Can the `DivergenceDuplicateSigner` and `SignersNotMembers` error cases collapse to simply calling `validate_response_share`? In particular, is it anywhere checked that the shares' `callback_id` all match the request's?

pierugo-dfinity, line 537: Sorry, not your changes.

Is it checked anywhere that there are no duplicate signers in `response.proof.signatures`?

eichhorl: There may be duplicates on the wire, but they are folded together (last one wins) when deserializing (`response.proof.signatures` is a `BTreeMap<NodeId, CanisterHttpResponseSignature>`).

eichhorl: Yes, there is even a ticket for it 😄 https://dfinity.atlassian.net/browse/CON-1730

> In particular, is it anywhere checked that the shares' callback_id all match the request's?

Yes, just below we check that it is a single group, and then below that there exists a context for that group.
```rust
// A divergence proof must contain shares for exactly one callback id.
let grouped_shares = group_shares_by_callback_id(response.shares.iter());
if grouped_shares.len() != 1 {
    return invalid_artifact(
        InvalidCanisterHttpPayloadReason::DivergenceProofContainsMultipleCallbackIds,
    );
}
```

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Yes, technically the pool should not have duplicate shares or shares from nodes that aren't part of the committee. Historically the payload builder still does these checks, so now they are also done for flexible outcalls and here.

We need to flatten the grouped shares anyway, so I think we might as well keep the check for now. But I agree we should check the performance cost of all the committee checks and potentially remove them everywhere in a follow up.

pierugo-dfinity: Right, works for me

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity: > Yes, there is even a ticket for it 😄 

👍 

> Yes, just below we check that it is a single group, and then below that there exists a context for that group.

Ah yes, that makes sense, thank you

pierugo-dfinity: Ah right ok 👍

pierugo-dfinity (approved):

PR #11086, opened 2026-08-10
feat(cketh): sweeper fee-funding task, with an end-to-end test

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding), first of the three remaining PRs. Targets `master`, now that #11083 has merged.

Also carries the fee-subaccount burn helper, folded in from #11065 — that PR was closed rather than merged, since on its own it was a constant and a function nothing called. It lands here instead, next to its only caller and the end-to-end test that drives it through the minter. The ICRC-2 rule it depends on — that a spender may spend from an account only when it names that account's own subaccount — is covered for every ledger in #11139.

## What

The periodic task that keeps the dedicated sweeper address funded: derive the address, decide whether a top-up is due, burn ckETH from the minter's fee subaccount, and only then queue the transfer. Everything after the burn is the existing withdrawal pipeline. A failed burn halts funding — and eventually sweeping — rather than moving ETH nothing has paid for.

## How it knows the sweeper is low

Not by reading the chain. The minter already records what it sent: the ETH that finalized fundings delivered is a **lower bound** on the sweeper's balance, and nothing else debits that address until sweeping lands (S2), at which point the bound subtracts the gas submitted sweeps provisioned. ETH sent there by anyone else only pushes the true balance above the bound.

A bound is the right input for the decision, because it errs in the safe direction: at or above the low-water mark it proves no funding is needed; below it, we may top up when we need not, which costs a burn we are backing anyway and self-corrects, since our own transfer raises the bound. It also means the funding task makes no HTTPS outcall at all — no provider agreement to choose, no block height to pick, no unreadable-balance path to handle, and nothing to go stale.

## What a funding burns

The amount it moves, and nothing else: the fee is carved out of that same amount, exactly as for a user withdrawal, so a top-up of 0.3 ETH — the target ten minimum withdrawals imply on mainnet — burns 0.3 ckETH and the sweeper receives 0.3 minus the fee. Whatever part of the provisioned fee goes unpaid stays as backing rather than being credited against the next funding — the simplification agreed for this iteration, which leaves funding accounted for the same way withdrawals already are.

The amount always clears the ledger's own minimum without a floor of its own, since a funding moves at least the configured headroom and validation keeps that at or above the minimum withdrawal amount. A real burn is what gives the funding the ledger index the whole withdrawal pipeline is keyed by.

Only one funding is allowed in flight at a time, which the planner reads off the withdrawal pipeline — a funding request that is pending, or has a created or sent transaction, is outstanding. Deriving it rather than mirroring it in the accounting means the two cannot disagree, so there is no inconsistent state to detect. The rule itself is prudence rather than a correctness requirement — two fundings would each be covered by their own burn — but it keeps a single funding on the withdrawal nonce lane and the accounting easy to follow.

## A production guard the adversarial test forced

Finalizing a funding debits the minter's ETH balance counter, which is credited only by deposits. In steady state every debit is covered by its own burn, so the counter cannot go negative — but on a fresh deployment it starts at zero while the main address may already hold ETH, and nothing previously stopped a funding from reaching for it. The debit at finalization would then underflow and trap, in the withdrawal timer, which would be stuck permanently and head-of-line block every user withdrawal behind it. It does not need anything to go wrong: a perfectly successful funding triggers it.

Planning now refuses to fund beyond the deposit-backed balance and logs why. Waiting is the safe direction — the sweeper simply stays unfunded until deposits cover it.

This was invisible until the adversarial test in the last PR of the stack waited for finalization; both live tests previously finished while the transfer was still in flight, which is also why the assertions could not see it.

## Live tests that no longer wait

The funding path sits behind the 6-minute withdrawal timer, so the end-to-end test used to spend that time on the wall clock. It now buys the ticks with `advance_time`, which works on a live instance: auto-progress sets the time once and thereafter advances *by* elapsed deltas, so an explicit jump sticks. **372s to 18.6s**, and the target is no longer tagged `long_test`, so it runs in the ordinary pipeline. The harness also sets the certified time before going live, the same race #11299 just fixed for the balance-scan harness.

Two rules the harness documents at the constants that enforce them: a jump fires a due interval timer *once* however far it jumps, so N ticks need N jumps; and nothing may be in flight when time moves, because `CANISTER_HTTP_TIMEOUT_INTERVAL` is 60 seconds and PocketIC uses the real payload builder — so the harness lets outcalls drain after each jump.

## Cost

Replaying the recorded mainnet event log through the funding accounting is measured at 1.177B instructions in `bench_post_upgrade`, against 1.187B on `master` — so it costs nothing measurable next to what master's own changes have moved in the meantime. `canbench/results.yml` records the figure, regenerated after each merge rather than reconciled by hand.

## Stack

| # | PR | Status |
|---|----|--------|
| — | #11083 — Burn-first accounting for sweeper fee funding | merged |
| 1 | Sweeper fee-funding task, with an end-to-end test | **this PR** |
| 2 | #11094 — Sweeper funding observability and the prepaid-gas gate | draft |
| 3 | #11097 — Adversarial end-to-end coverage of sweeper fee funding | draft |


rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -108,6 +111,15 @@ fn setup_timers() {
     ic_cdk_timers::set_timer_interval(BALANCE_SCAN_INTERVAL, async || {
         balance_scan().await;
     });
+    // Checked shortly after install as well as on the interval, so an upgrade that finds the
+    // sweeper address low does not have to wait a whole interval to act on it. Deliberately not at
+    // zero: see INITIAL_SWEEPER_FUNDING_DELAY.
+    ic_cdk_timers::set_timer(INITIAL_SWEEPER_FUNDING_DELAY, async {
+        fund_sweeper_address().await;
+    });
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 25a99e0c08. You are right, and the distinction you are drawing is the one my commit message glossed over: a delay narrows the window, it does not order the two.

The first check now runs from the *same timer callback* as the key fetch, immediately after it:

```rust
ic_cdk_timers::set_timer(Duration::from_secs(0), async {
    let _ = lazy_call_ecdsa_public_key().await;
    fund_sweeper_address().await;
});
```

The ordering is structural rather than timed, and `INITIAL_SWEEPER_FUNDING_DELAY` is gone. This also keeps the property the delay was introduced for in the first place: the two stay on one task, so there is never a second concurrent `ecdsa_public_key` call — which trips an ic-cdk invariant and traps the minter. That was the defect the delay was papering over, and sequencing addresses both at once.

On your second scenario — a key fetch that traps, leaving periodic funding to skip until something else populates the key — that remains true, and I decided against guarding it. `lazy_call_ecdsa_public_key` traps rather than returning an error, so it cannot be retried in place, and re-arming a short retry from the funding task would be protecting a state where the feature cannot work anyway: without the master public key there is no sweeper address to fund, so sweeping is blocked by the missing key rather than by the funding schedule. Recorded in the commit message rather than left implicit.

One knock-on worth mentioning: the mocked fixture no longer advances IC time by 60 seconds to reach the check. That jump was itself a nuisance — it perturbs the other periodic timers, which is why the fixture deliberately does not re-settle after `upgrade_minter` — so removing it takes out a source of test coupling. I ran the three mocked suites locally to confirm the fixture still settles the check without it (`cketh_test`, `ckerc20_test`, `fee_account_test`, all passing), since this changes install-time timer ordering for every mocked test.

mbjorkqvist (commented):
### On the suppressed comment (`tests/sweeper_funding.rs:17`, seeding the fee account after install)

Confirmed and fixed in 3a2e7972ec. Both failure modes you describe are real, and there is a connection worth recording: **the ordering fix in the previous round made this race more likely, not less.** The old 60-second delay had been accidentally giving the test time to seed the fee account before the task looked; sequencing the check straight after the key fetch removed that grace period.

The fee account is now seeded through the ledger's `initial_balances`, so the balance exists before the minter is installed at all and the task cannot precede it — your suggestion, and the right shape, since the ledger is installed first. `new_live_with_empty_fee_account` covers the one test that wants it empty.

Verified by running both live suites rather than reasoning about it, and that turned up a second instance of the same class which the fix had exposed:

`should_not_fund_a_sweeper_above_the_low_water_mark` (in PR 7 of this stack) **failed** — supply 1.0 → 0.9 ETH, a burn where the test asserts none. With the fee account funded from install and the check firing immediately, the install-time check funded the sweeper before the test could top it up. And there is no second chance inside the test window: the next scheduled check is 24 hours away.

That test now arranges itself in the only order that is deterministic:

1. start with an **empty** fee account, so the install-time check decides a funding is due, fails to burn, and changes nothing — the address is undiscoverable until the minter caches its key, so this is the only window;
2. read the sweeper address, set its balance above the low-water mark, then fund the fee account;
3. upgrade the minter to re-arm the timers, so a check runs again inside the test;
4. assert no burn — which now means it *declined* with a funded fee account and a real opportunity to act.

Both suites pass: `sweeper_funding` 370 s, `sweeper_funding_hardening` 492 s. The other two hardening tests needed no change — one deliberately uses the empty fee account, and the revert test's `set_code` has the full 6-minute withdrawal-timer window before the transfer is sent, so its burn landing early is expected rather than racy.

rs/ethereum/cketh/test_utils/src/sweeper_funding.rs
```diff
@@ -0,0 +1,459 @@
...
+        setup
+            .anvil
+            .set_balance(&setup.minter_address, MINTER_ETH_BALANCE);
+        setup.await_deposit_credited(Duration::from_secs(300));
+        // Funded here rather than at install: with an empty fee account the install-time check
+        // cannot burn, whichever way it and the scrape interleave. Safe to do now because the next
+        // scheduled check is a whole interval away, so nothing is watching until the upgrade below
+        // re-arms the timers — which makes that run the first one able to fund.
+        if fee_account_balance > 0 {
+            setup.mint_cketh(setup.fee_account(), fee_account_balance);
```
mbjorkqvist: Good catch, and the analysis is right — fixed in 0cd713dc.

The harness seeded the fee account through the ledger's `initial_balances`, so the install-time check could burn the moment the scraper credited the deposit. The test reads `supply_before` after the harness returns, so on that interleaving the in-flight guard blocks the post-upgrade funding, the earlier transfer still lands (so `await_eth_received` succeeds), and the burn it measures is zero — exactly the failure you describe.

I took the first of your two options, since removing the ordering dependence is better than measuring around it. The ledger now installs with every balance empty, and the fee account is credited after the deposit is credited and before the upgrade re-arms the timers. Nothing is watching in that window — the next scheduled check is a whole interval away — so the install-time check has nothing to burn however it and the scrape interleave, and the post-upgrade run is deterministically the first that can fund. `mint_cketh` mints from the minting account, so the credit is fee-free and the resulting balance is identical to what `initial_balances` produced.

This is the same race the adversarial tests later in the stack already avoid by starting from an empty fee account — I hadn't carried that reasoning back to this test.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -317,6 +318,36 @@ rust_test(
     ],
 )
 
+rust_test(
+    name = "sweeper_funding",
+    # Waits on the minter's 6-minute withdrawal timer for the funding transfer to be sent.
+    size = "large",
+    timeout = "long",
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in 514e14f7 — the target is now `timeout = "eternal"`.

The arithmetic is slightly worse than described: the waits are 5 minutes for the deposit to be credited, **3** for the sweeper address (not 2 for the funding decision), and 13 for the receipt — 21 minutes against a 15-minute budget. Typical runs finish in about 6, so this only bites a slow one, but it bites in the worst way: Bazel kills the test before any internal deadline fires, so the diagnostics those deadlines exist to produce are replaced by a bare timeout.

You're also right that it contradicted the harness. `long_lived_server_url` sizes the PocketIC hard TTL at 4,200s specifically to sit above `eternal`, and its comment says so — the two disagreed, and the comment was the correct half. The hardening target added later in the stack was already `eternal`; this one was the outlier.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -822,6 +822,17 @@ impl Default for EthBalance {
...
+#[cfg(test)]
+impl EthBalance {
+    /// An `EthBalance` holding `eth_balance` of deposit-backed ETH.
+    pub fn with_eth_balance(eth_balance: Wei) -> Self {
+        Self {
+            eth_balance,
+            ..Default::default()
+        }
+    }
+}
```
gregorydemay: nit: I would move this to the test file.

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/// The minter's dedicated sweeper address, or `None` while the master public key is still unknown.
+///
+/// Reads the cached key rather than fetching it: a second concurrent `ecdsa_public_key` call traps
+/// the canister, so the first run is delayed until the install-time fetch has cached it.
+pub fn sweeper_address_from_state(state: &State) -> Option<Address> {
+    let (master_public_key, chain_code) = state.public_key_and_chain_code()?;
+    Some(sweeper_address(&master_public_key, &chain_code))
+}
```
gregorydemay: nit: I would move this as a method on `State`

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -0,0 +1,190 @@
...
+            DEBUG,
+            "[fund_sweeper]: SKIPPING: the ECDSA public key is not available yet"
+        );
+        return;
+    };
+
+    let block_height = read_state(|s| s.ethereum_block_height.clone()).into();
+    let sweeper_balance =
+        match crate::eth_rpc_client::get_balance::eth_get_balance(&sweeper, block_height)
+            .await
```
gregorydemay: There might be a slight mismatch here:
* Note that in production this will take the last fetched finalized block (`ethereum_block_height`). 
* The method `eth_rpc_client::get_balance::eth_get_balance` was until now used to notice users' deposits on minter-controlled addresses using Latest block height and was planning on using an RPC client with a lower threshold (2 out-of 3, instead of 3 out of 4), PR in draft (#11243).


I see 2 alternatives:

1. refactor the method `eth_get_balance` to take an additional parameter being a reference to the client, so that it does not instantiate it.
2. Do we need RPC calls at all? We could track on the heap a lower bound on what the balance of the sweeper address should be, since the minter controls the spending (there could be anonymous donations :-))

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -12,7 +12,22 @@
...
+/// A funding whose burn is recorded but whose transfer has not finalized yet. Kept so that only one
+/// funding is ever outstanding, and so a wedged one is visible instead of merely absent.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub struct InFlightFunding {
+    pub ledger_burn_index: LedgerBurnIndex,
+    /// The ETH the transfer will move.
+    pub amount: Wei,
+    /// IC time the funding was accepted, so a wedged one shows up as a climbing age. Optional
+    /// because the request's own field is: the minter always sets it, but the type allows its
+    /// absence and inventing a timestamp here would report an age that is not real.
```
gregorydemay: nit:

```suggestion
    /// IC time the funding was accepted.
```

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -47,9 +65,68 @@ impl SweeperFundingAccounting {
...
+            Some(earmarked) if earmarked.ledger_burn_index == ledger_burn_index => {
+                self.in_flight = None
+            }
+            earmarked => log!(
+                INFO,
+                "[record_finalized_funding]: UNEXPECTED: finalizing sweeper funding \
+                 {ledger_burn_index}, but the earmarked funding is {:?}; leaving it earmarked, \
+                 which holds off further funding until it settles",
+                earmarked.map(|funding| funding.ledger_burn_index),
+            ),
```
gregorydemay: shouldn't we just panic in that case? This arm would the sign of a BUG.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -47,9 +65,68 @@ impl SweeperFundingAccounting {
...
+        if let Some(earmarked) = self.in_flight {
+            log!(
+                INFO,
+                "[mark_funding_in_flight]: UNEXPECTED: sweeper funding {ledger_burn_index} of {} \
+                 was accepted while funding {} of {} is still in flight; earmarking the new one",
+                amount,
+                earmarked.ledger_burn_index,
+                earmarked.amount,
+            );
+        }
```
gregorydemay: This would be the sign of a BUG, I would panic.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
```diff
@@ -0,0 +1,49 @@
+//! End-to-end proof of sweeper fee funding against a real EVM with nothing mocked: real ckETH
+//! ledger, real EVM RPC canister, threshold-ECDSA signature, local anvil. The harness lives in
+//! [`ic_cketh_test_utils::sweeper_funding`].
+//!
+//! Genuinely waits minutes: the transfer is only sent by the minter's 6-minute withdrawal timer.
```
gregorydemay: Let's hope not, there is PocketIc advance_time exactly for this.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use ic_cketh_test_utils::sweeper_funding::SweeperFundingSetup;
+use std::time::Duration;
+
+const FUNDING_DEADLINE: Duration = Duration::from_secs(13 * 60);
+
+#[test]
+fn should_fund_the_sweeper_address_by_burning_cketh_from_the_fee_account() {
+    let setup = SweeperFundingSetup::new_live();
```
gregorydemay: why would we need yet another setup? Let's reuse what we have.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -330,6 +331,38 @@ rust_test(
     ],
 )
 
+rust_test(
+    name = "sweeper_funding",
```
gregorydemay: why not have this test in deposit_from_cex? It does not make sense IMO to have a new test setup for this and another test file. It's part of that feature so I would move it there and it should considerably reduce the diff.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -330,6 +331,38 @@ rust_test(
...
+    ],
+    env = {
+        "ANVIL_BIN": "$(rootpath //:anvil)",
+        "CARGO_MANIFEST_DIR": "rs/ethereum/cketh/minter",
+        "CKETH_MINTER_WASM_PATH": "$(rootpath :cketh_minter_debug.wasm.gz)",
+        "EVM_RPC_CANISTER_WASM_PATH": "$(rootpath @evm_rpc.wasm.gz//file)",
+        "LEDGER_CANISTER_WASM_PATH": "$(rootpath //rs/ledger_suite/icrc1/ledger:ledger_canister_u256.wasm.gz)",
+        "POCKET_IC_BIN": "$(rootpath //rs/pocket_ic_server:pocket-ic-server)",
+    },
+    tags = ["long_test"],  # ~6 minutes, nearly all of it waiting on that timer.
```
gregorydemay: That shouldn't be necessary, use advance_time

rs/ethereum/cketh/test_utils/src/anvil.rs
```diff
@@ -173,6 +211,49 @@ impl Anvil {
...
+
+        self.set_code(helper, &code);
+        let hash = self.send_transaction(from, Some(helper), &[]);
+        assert!(
+            status_ok(&self.await_receipt(&hash)),
+            "emitting the deposit log reverted"
+        );
+        // The minter reads at `finalized`, which trails `latest`.
+        self.mine(3);
+    }
```
gregorydemay: I think it would be easier/more realistic to just do a ckETH deposit flow. This seems a bit like dark magic.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -1,12 +1,28 @@
...
+        accounting: &mut SweeperFundingAccounting,
+        index: u64,
+        amount: Wei,
+        burn: Wei,
+    ) -> LedgerBurnIndex {
+        let index = LedgerBurnIndex::new(index);
+        accounting.record_burn(burn);
+        accounting.mark_funding_in_flight(index, amount, CREATED_AT);
+        index
+    }
```
gregorydemay: why not apply state transitions? This would use the productive code. See for example the tests on the dashboard like `should_display_pending_deposits_sorted_by_decreasing_block_number`

rs/ethereum/cketh/minter/tests/cketh.rs
```diff
@@ -871,13 +873,19 @@ fn should_be_able_to_stop_canister_during_scraping() {
...
-        cketh.env.get_canister_http().len(),
+        pending_outcalls_for(&cketh.env, &JsonRpcMethod::EthGetLogs),
         4,
         "Expected HTTPS outcalls since scraping is still in progress: {}",
         debug_http_outcalls(&cketh.env)
     );
 
+    // The install-time funding check's balance read is an open call context of its own, and this
+    // test asserts the canister reaches Stopped with nothing left in flight.
+    cketh.settle_initial_sweeper_funding_check();
```
gregorydemay: maybe it would be better to answer all pending calls that are **not** for `EthGetLogs`. Like this, this test does not need to know about funding.

rs/ethereum/cketh/minter/tests/cketh.rs
```diff
@@ -1376,3 +1384,16 @@ mod cketh_evm_rpc {
...
+/// failures wherever the constant is used.
+#[test]
+fn should_derive_the_expected_minter_address() {
+    let cketh = CkEthSetup::default();
+
+    assert_eq!(
+        Address::from_str(MINTER_ADDRESS).unwrap(),
+        Address::from_str(&cketh.minter_address()).unwrap()
+    );
+}
```
gregorydemay: duplicates `should_derive_minter_address`

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -156,6 +157,10 @@ impl CkEthSetup {
         install_evm_rpc(&env, &canisters, &backend);
         install_minter(&env, &canisters, &backend);
 
+        // Deliberately no call to the minter here, not even to check its address against
+        // `MINTER_ADDRESS`: `minter_address` is an update call, and a fixture that has executed a
+        // round lets the install-time timers fire before the test drives them, which the flows rely
+        // on doing themselves. `should_derive_the_expected_minter_address` checks the constant.
```
gregorydemay:
```suggestion
```

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -490,6 +525,9 @@ impl CkEthSetup {
             )
             .unwrap();
         self.start_minter();
+        // Deliberately not settled again after an upgrade: advancing time to it perturbs the other
+        // periodic timers and breaks unrelated tests, and `stop_minter` drains pending outcalls
+        // anyway. Assertions counting pending outcalls should filter by method.
```
gregorydemay:
```suggestion
  ```

File: rs/ethereum/cketh/minter/src/sweeper/tests.rs [hunk 20]
```diff
@@ -0,0 +1,202 @@
...
+
+        let first = match plan_funding(&state, Wei::ZERO) {
+            FundingDecision::Fund(amount) => amount,
+            other => panic!("the first funding must be due, got {other:?}"),
+        };
+        assert_eq!(first, Wei::new(TARGET));
+        state.sweeper_funding.record_burn(first);
+        state
+            .sweeper_funding
+            .mark_funding_in_flight(LedgerBurnIndex::new(1), first, CREATED_AT);
```
gregorydemay: nit: here and below it would be better to apply the state transaction method with the event.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+        let fee = Wei::new(1_000_000_000_000_000);
+        state.sweeper_funding.record_finalized_funding(
+            LedgerBurnIndex::new(1),
+            first.checked_sub(fee).unwrap(),
+            fee,
+        );
+
+        // The sweeper now holds roughly the target, so nothing is due — but the reason is
+        // "not due", not "still in flight".
```
gregorydemay:
```suggestion
        // The sweeper now holds roughly the target, so nothing is due.
```

rs/ethereum/cketh/minter/src/sweeper/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+        );
+        // Once it drains again, funding resumes.
+        assert!(matches!(
+            plan_funding(&state, Wei::ZERO),
+            FundingDecision::Fund(_)
+        ));
+    }
+
+    /// Whatever sequence the guard permits, the invariant must survive it. This is the assertion the
+    /// buggy sequence failed: `record_finalized_funding` traps when spend exceeds burn.
```
gregorydemay, line 226:
```suggestion
```

rs/ethereum/cketh/minter/src/sweeper/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+                LedgerBurnIndex::new(index),
+                amount.checked_sub(fee).unwrap(),
+                fee,
+            );
+            assert!(
+                state.sweeper_funding.cumulative_burned()
+                    >= state.sweeper_funding.cumulative_spent()
+            );
+        }
+    }
```
gregorydemay, line 241: isn't this the same as `should_preserve_the_invariant_across_consecutive_fundings`?

rs/ethereum/cketh/minter/src/sweeper/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+            Wei::new(TARGET).checked_sub(fee).unwrap(),
+            fee,
+        );
+        accounting.record_finalized_funding(
+            LedgerBurnIndex::new(1),
+            Wei::new(TARGET).checked_sub(fee).unwrap(),
+            fee,
+        );
+        assert!(accounting.cumulative_burned() >= accounting.cumulative_spent());
+    }
```
gregorydemay, line 286: why would we test this scenario? I think the guard makes sense and it's good to have it, but why would we have test for artificial scenarios that don't exist in the production code?

gregorydemay (commented):
Comments are mainly about the tests, where I think we could significantly reduce the PR diff. 1 comment about the use of eth_getBalance.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Rather than argue the panic either way, the state this arm guards is gone: 67b2f06cd4 derives the outstanding funding from the withdrawal pipeline instead of mirroring it in the accounting, so `mark_funding_in_flight` and the `in_flight` field no longer exist and there is no arm left to panic or log in.

Two things pushed me that way. The pipeline already knows — `pending_withdrawal_requests`, then the `created_tx`/`sent_tx` entries — so the earmark was a second copy of one fact, and a second copy is what made an inconsistent state representable at all. And the reaction was a poor trade in either direction: `ic-cdk-timers` reschedules in the dispatching message, so a trap would not have stopped the daily retry, and the burn commits one message before the trap, so a panic would have left ckETH burned with no event recording it — invisible to the counters and to `/metrics`, since a trap rolls back the heap.

`plan_funding` now asks `state.withdrawal_transactions.outstanding_sweeper_funding()`, and the guard test asserts refusal at pending, created and sent, and resumption after finalization — so it fails if the derivation breaks, which the old test could not.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Also gone with the derivation in 67b2f06cd4: the index was only there to release the earmark, so `record_finalized_funding(transferred, transaction_fee)` no longer takes it and the pairing match is deleted, both arms.

Nothing is lost by not checking: the caller has already looked the request up by `withdrawal_id` and matched it as `SweeperFunding`, so the pairing is established by the lookup. The eager `burned_not_yet_spent()` check stays, and that one does panic — under-backed ckETH is unrecoverable, so trapping is right there. Trap on solvency, not on bookkeeping.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Moot now: `InFlightFunding` no longer exists, so neither does the field. The age the metric reports comes from the request's own `created_at` (67b2f06cd4).

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: Done in 67b2f06cd4: the `concurrent_fundings` tests now move a funding through the pipeline by applying the events the minter records — `AcceptedSweeperFundingRequest`, `CreatedTransaction`, `SignedTransaction`, `FinalizedTransaction` — through `apply_state_transition`, rather than poking the accounting.

Worth saying why beyond style: the audit-handler bug I fixed in #11072 (a funding recorded as a `CkEth` withdrawal, because `EthWithdrawalRequest`'s own `Into` yields that variant) was invisible precisely because the tests bypassed the event path.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
mbjorkqvist: The helper this was on is gone — the funding is accepted through the event now, in the tests under `sweeper/tests.rs`.

The remaining tests in this file still call `record_burn`/`record_finalized_funding` directly, deliberately: they have no `State`, they assert arithmetic (overflow, the surplus after a failed transfer, the under-backing panic), and routing them through events would duplicate the sequencing tests that now exist next door. There is a module doc pointing at where the sequencing is covered.

rs/ethereum/cketh/minter/src/state.rs
mbjorkqvist: Done in 94a7b54a74 — it lives in `state::tests` now, which can build the struct directly since the fields are its own.

rs/ethereum/cketh/minter/src/sweeper.rs
mbjorkqvist: Done in 835ca9c640: `State::sweeper_address()`. The dashboard reads it off the state too (in #11094).

rs/ethereum/cketh/test_utils/src/lib.rs
mbjorkqvist: Removed in 94a7b54a74.

rs/ethereum/cketh/test_utils/src/lib.rs
mbjorkqvist: Removed in 94a7b54a74 — I read the empty suggestion as "drop the comment"; say the word if you meant something else.

rs/ethereum/cketh/minter/tests/cketh.rs
mbjorkqvist: Good suggestion, and it went one better in the end: the funding task no longer makes an outcall at all, so an install schedules nothing but scraping calls and this test needs no settling whatsoever. `tests/cketh.rs` is back to its state on master — no diff from this PR — and the helper I had added for it is gone too.

That follows from taking your option 2 on the `eth_getBalance` thread: the balance is a tracked bound now, not a chain read.

rs/ethereum/cketh/minter/tests/cketh.rs
mbjorkqvist: You're right — removed in af7d4e73c5. `should_derive_minter_address` already pins the constant; mine was left over from taking the check out of `CkEthSetup::new`.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: Done in 94a7b54a74.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: Done in 94a7b54a74.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: Agreed, they overlapped. Removed the arithmetic one in 94a7b54a74 and kept the sibling that drives `plan_funding` through the pipeline.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: It is reachable only by bypassing the guard, so I take the point. What it pins after 67b2f06cd4 is defined behaviour rather than an impossible state: with the outstanding funding derived, two of them is simply what the pipeline holds, the oldest is the one reported, and the invariant survives because each funding burned for its own transfer.

That felt worth one cheap test, since it is the property the guard's absence would violate. Happy to drop it if you would still rather not have it.

rs/ethereum/cketh/minter/src/sweeper.rs
mbjorkqvist: Good catch on both counts, and thinking it through I would take your **option 2**, not option 1 — the balance does not need reading at all.

The minter already knows a lower bound on it, from data it records: `cumulative_transferred` is the ETH finalized fundings delivered to the sweeper, and nothing debits it yet. Donations only push the true balance above that bound, which is the harmless direction. When sweeping lands (S2), the bound becomes `cumulative_transferred - provisioned gas of submitted sweeps + unspent refunds at finalization` — debiting the provisioned maximum at submission, the same discipline `WithdrawalTransactions` already applies to withdrawals, so a lost finalization leaves the bound too low rather than too high.

A bound is the right input for both decisions. Funding: if the bound is at or above the low-water mark then so is the balance, so we correctly do not fund; below it we may fund when we need not, which is safe and self-correcting since our own transfer raises the bound. The prepaid-gas gate in #11094: `bound >= required` is conservative the same way, and better than what I had — "never spend ETH no burn has covered" then holds by construction rather than by a staleness rule.

What decided it is that adopting it deletes machinery instead of adding it: the `eth_getBalance` call goes from the funding task (so the block-height and threshold coupling you raised disappears, and with it any dependency on #11243), the install-time outcall goes — which takes the harness settling and your `cketh.rs:887` comment with it — and in #11094 the observation cache, the staleness limit, `NeverObserved`/`Stale` and the NaN metric plumbing all go.

Costs, for the record: the bound starts at zero, so a sweeper address that already holds ETH gets one unnecessary funding (a fresh address on mainnet, so zero); the dashboard reports a bound rather than an observation; and `R14`'s "reconcilable on-chain with one `eth_getBalance`" wants rewording, since we would track it and reconcile only when we choose to. If we later want divergence alerting, a monitoring-only read can go back in #11094 without any decision depending on it.

Doing this in #11086 and #11094 now — shipping the observation machinery and then removing it would be worse.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
mbjorkqvist: You were right, and I was wrong to doubt it — I had read the client's doc string ("update the time of the IC to the real time") as meaning auto-progress would undo a jump. The server does not: it sets the time once at start and then advances *by* elapsed deltas, so `advance_time` is additive and sticks. `ingress_expiry` follows the instance's time too, so later calls stay valid.

Adopted in 15a0be5. `:sweeper_funding` goes from **372s to 18.6s**, and the module doc no longer claims the harness must wait minutes.

Two rules it has to follow, both documented at the constants that enforce them, in case they save you time elsewhere:

* **One jump fires a due interval timer once**, however far you jump. A single 60-minute jump sent the funding transaction and then stopped — finalization is the *next* run of that timer. N ticks need N separate jumps.
* **Never jump while an outcall is in flight.** `CANISTER_HTTP_TIMEOUT_INTERVAL` is 60s (`rs/types/types/src/canister_http.rs:81`) and PocketIC uses the real payload builder, so any outcall in flight dies with `SysTransient: Canister http request timed out`. The harness drains for ~5s after each jump; below 2s it fails reliably.

rs/ethereum/cketh/minter/BUILD.bazel
mbjorkqvist: Correct, and gone. With ticks bought rather than waited out, both targets are now `size = "medium"` with no `timeout` override and **no `long_test` tag** — so they run in the ordinary PR pipeline instead of on demand.

Kept `medium` rather than the `small` Bazel suggests: 60s is tight for a real anvil plus a live instance on a loaded runner, and there is a comment in `BUILD.bazel` saying so. Shout if you would rather keep `long_test` — it does mean two live anvil tests on every PR.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
mbjorkqvist: Fair, and it shrank a lot in the meantime for a different reason: the funding task no longer reads the balance off the chain (your option 2 on the `eth_getBalance` thread), so the harness lost the mock-provider plumbing, the settling helpers and `set_eth_balance`/`mine` along with it.

What is left is genuinely funding-specific — a fee account with ckETH, the sweeper's address, and waiting for a burn and a finalized transfer — but it does still install the ledger, the EVM RPC canister and the minter itself, which `CkEthSetup::new(EthereumBackend::Anvil(..))` already does. I would like to rebuild it on that rather than relocate the tests: it removes the duplication you are pointing at, and it inherits a property `live_scan.rs` documents and this harness does not — build the fixture on a non-live instance and only then go live, since building against an auto-progressing instance "reproducibly failed under CPU contention".

Happy to do that here if you want it before merging, or as the first commit of a follow-up. What I would avoid is folding these into `deposit_from_cex`: see the reply on the BUILD comment.

rs/ethereum/cketh/minter/BUILD.bazel
mbjorkqvist: I would rather not, for two reasons that are about the fixture rather than the file count.

`LiveBalanceScanSetup` builds `CkErc20Setup::with_cketh(..).add_supported_erc20_tokens()` — orchestrator, two ledgers, two index canisters, two minter upgrades — none of which funding touches. And `deposit_from_cex` is fast (~70s) because it only ever waits on `BALANCE_SCAN_INTERVAL` (30s); funding sits behind `PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL` (6 min), so before the `advance_time` work it would have turned that target into a seven-minute one. Less of an issue now that ticks are cheap, but the fixture mismatch stands.

The diff reduction you are after is real though, and I think rebuilding this harness on `CkEthSetup` gets most of it — see the reply on `tests/sweeper_funding.rs:14`. Sharing a fixture also means sharing its fixes: `make_live` has the same asynchronous-time-set race that #11299 just fixed for the balance-scan harness, so I have applied the same `set_certified_time` there for now, and a rebuild would inherit it along with `CkEthSetup::advance_time`'s drain-before-jump.

rs/ethereum/cketh/test_utils/src/anvil.rs
mbjorkqvist: Agreed that it is obscure, and it turns out the real thing is within reach: `EthDepositHelper.sol` is in-repo under the `helper_contracts` target, and two existing test targets already compile Solidity at test time with the vendored `//:solc` (that is how the demo test builds MockUSDT). So the bytecode trick can be replaced by deploying the real deposit helper and calling it.

I would rather not do it in the same change as the harness rebuild above, since both touch construction and I would like the failure modes separable — but I am happy to take it as the next commit after that, or now if you prefer it before merging.

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -0,0 +1,174 @@
...
+        log!(
+            DEBUG,
+            "[fund_sweeper]: SKIPPING: the ECDSA public key is not available yet"
+        );
+        return;
+    };
+
+    // What the minter's own records say reached the sweeper address, rather than a chain read: the
+    // bound errs low, so it can only delay a funding, never authorise one against gas that is not
+    // there.
```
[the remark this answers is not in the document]
mbjorkqvist: You are right, and the docstring on `plan_funding` below already had it the other way round from this comment — which is how the slip survived. Fixed in 704b74394e: erring low can only make a funding look due sooner than it is, never hide one that is.

The "never authorise gas that is not there" half belongs to the other reader of the same bound, the sweeping side, where erring low withholds gas rather than over-committing it. That distinction now lives where both readers meet, in the design doc.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -348,8 +349,13 @@ with an `icrc1_balance_of` of `1_762_128_000_000_000_000` wei ≈ 1.76 ckETH as
...
    resubmission machinery as withdrawals.
 
-* The sweeper address' balance *is* the `prepaid_sweep_gas` counter, reconcilable
-  on-chain with one `eth_getBalance`. Sweep gas draws it down; burned ckETH is
+* The sweeper address' balance is the `prepaid_sweep_gas` counter, and the minter
+  tracks a lower bound on it from its own events — what finalized fundings
+  delivered, less the gas submitted sweeps provisioned — and may reconcile that
+  bound against the chain whenever it chooses. The bound errs low: ETH anyone else
+  sends to the address only pushes the true balance above it, so the bound can only
+  delay a funding or a sweep, never authorise one against gas that is not there.
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed — the two effects were collapsed into one sentence. Fixed in 704b74394e; the paragraph now separates them: a funding may be triggered earlier than strictly needed, never skipped, and a sweep may be held back, never authorised against gas that is not there.

rs/ethereum/cketh/minter/BUILD.bazel
mbjorkqvist: Copilot suppressed a comment on this line asking for the explicit timeout back, and while I would rather not, the observation under it was fair: the harness' own waits summed to more than the 300 seconds `size = "medium"` grants — 300 for the deposit to be credited, then 120 for the funding decision, then the settle windows — so a genuinely hung minter would have been killed by Bazel before any of those could report its logs.

Fixed from the other end in 704b74394e: every wait now shares one 60-second deadline, sized against that budget, which puts the worst case around 170 seconds including setup. The timeout stays gone.

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -7,62 +7,47 @@
...
-                    "[fund_sweeper]: SKIPPING: failed to read the balance of {sweeper}: {e:?}"
-                );
-                return;
-            }
-        };
+    // What the minter's own records say reached the sweeper address, rather than a chain read.
+    // Erring low can only make a funding look due sooner than it is, never hide one that is.
+    let sweeper_balance = read_state(|s| s.sweeper_funding.sweeper_balance_lower_bound());
 
     let amount = match read_state(|s| plan_funding(s, sweeper_balance)) {
```
gregorydemay: nit: since plan_funding already needs a `&State` there is no need for the second parameter `sweeper_balance`, since it can be read from state.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
gregorydemay: > I would like to rebuild it on that rather than relocate the tests: it removes the duplication you are pointing at, and it inherits a property live_scan.rs documents and this harness does not — build the fixture on a non-live instance and only then go live, since building against an auto-progressing instance "reproducibly failed under CPU contention".

I don't follow the reasoning, at the end we want to test the feature end-to-end (as much as possible) and so we will need one test file to do that, and the sweeper funding task is part of that feature. If there is some problems with the current test harness, then let's fix those.

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: > Less of an issue now that ticks are cheap, but the fixture mismatch stands.

I don't agree with this part: there is not fixture mismatch, we want to test the full feature. As part of that feature, the ckETH ledger is also needed because the ledger fees are what is used to pay for the sweeper transaction fees.

rs/ethereum/cketh/test_utils/src/anvil.rs
gregorydemay: I still very much want to reuse the existing infrastructure that we have and then do the ckETH deposit flow, which we already have.

gregorydemay (commented):
LGTM! The main question is about the test infrastructure setup

rs/ethereum/cketh/minter/src/sweeper.rs
mbjorkqvist: Taken, in 057cad0ac8: `plan_funding(state)` reads the bound itself, and the task takes one snapshot for the decision and for the log lines that report it.

The tests were the interesting part. They had been passing a balance the state did not agree with, and two of them leaned on that: they funded the whole target and then asserted a funding was due again, which only holds if you pass zero in — with the bound derived, a topped-up sweeper is `NotDue`. They now fund partially, so a lifted guard is distinguishable from a sweeper that needs nothing, and a new test covers the second case. The repeated-funding test arranges its fundings rather than planning them, since nothing draws the bound down until sweeping spends the gas.

rs/ethereum/cketh/minter/tests/sweeper_funding.rs
mbjorkqvist: Agreed, and fixed rather than deferred — the second harness is gone.

`live_scan.rs` and `sweeper_funding.rs` were the same wrapper twice: a fixture, an owned anvil node and a switch to live outcalls, differing only in what they built and what they knew how to arrange. They are now one `LiveSetup<S>` in `live.rs`, generic over the fixture, so everything a live test needs regardless of feature sits in one place — buying minter time, depositing through the helper contract, the dashboard and canister-log readers, the anvil accessors. The balance scan keeps its token seeding and scan waiter as a layer on `LiveSetup<CkErc20Setup>`, funding its fee-account seeding on `LiveSetup<CkEthSetup>`. 797 lines across two modules became 700 in one, and all the live tests are in `deposit_from_cex.rs`.

A few things came out of sharing the fixture, all commented where they sit. Its anvil now starts mainnet-like for both flavours, since the minter signs for `EthereumNetwork::Mainnet` — which had not mattered while nothing sent a transaction. And a live instance wants three small habits: an ingress message cannot be awaited by driving the instance with explicit ticks (which `CkEthSetup::minter_address` and `stop_minter` both do), the upgrade goes through the client for the same reason, and two live instances in one test binary starve each other, hence `RUST_TEST_THREADS=1`.

rs/ethereum/cketh/minter/BUILD.bazel
mbjorkqvist: You are right, and I withdraw the fixture-mismatch argument: the ckETH ledger is part of the feature rather than adjacent to it. On mainnet its `fee_collector_account` is the minter's `0x…0fee` subaccount, which is exactly what a funding burns from, so the fixture with the ledger is the more faithful one.

There is one target now: `deposit_from_cex`, `size = "large"`, about three and a half minutes for eight tests of which four are live, run one at a time.

rs/ethereum/cketh/test_utils/src/anvil.rs
mbjorkqvist: Done, and it took a few other shortcuts with it. The harness now deploys `DepositHelperWithSubaccount.sol` — the production contract, compiled at test time by the vendored solc, as `deposit_from_cex_demo` already does — against the address the minter derives, adds it by upgrade the way mainnet gained it, and deposits through it. Gone along with the fabricated log: `anvil_setBalance` putting 100 ETH at the minter's address that no deposit had paid in, and the fee account being minted its ckETH directly. The ETH is now there because a depositor sent it, and the event is the one the contract emits. The happy-path test also got a little faster, 18.7s to about 16s.

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -0,0 +1,184 @@
...
+        required: Wei,
+    },
+    NotDue,
+}
+
+/// Decides whether a funding is due at `sweeper_balance`, which callers supply as a *lower* bound on
+/// the sweeper address' real balance. Erring low can only make a funding look due when it is not,
+/// which over-provisions gas the minter has burned for; the reverse would spend ETH against gas that
+/// is not there.
+///
```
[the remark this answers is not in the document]
mbjorkqvist: Right, and my own doing: that was meant to be a replacement rather than an insertion, so the doc described both the old parameter and its absence. Removed in cb9fdf3bdc.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -732,9 +740,24 @@ impl CkEthSetup {
 
 /// Builds the PocketIC instance for [`CkEthSetup::new`]: the fiduciary subnet every ckETH fixture
 /// needs for the secp256k1 `key_1` used by the minter. Always a non-live (manual-round) instance,
-/// even for [`EthereumBackend::Anvil`]: [`crate::live_scan`] builds its whole fixture here first and
+/// even for [`EthereumBackend::Anvil`]: [`crate::live`] builds its whole fixture here first and
 /// only switches to live outcalls once construction is complete, so `await_call` ticks
 /// deterministically for every setup call in between.
```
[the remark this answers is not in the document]
mbjorkqvist: Correct — `switch_to_live` went in between `new_env`'s doc comment and `new_env` itself, so it inherited documentation for something it does not do. The paragraph is back above `new_env` in cb9fdf3bdc.

rs/ethereum/cketh/minter/canbench/results.yml
```diff
@@ -2,7 +2,7 @@ benches:
   bench_post_upgrade:
     total:
       calls: 1
-      instructions: 1187292943
+      instructions: 1176834138
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch: the description was quoting a comparison against a `master` baseline that has since moved. The recorded figure is the current one — `results.yml` is regenerated with the `_update` target after each merge rather than reconciled by hand — and it now reads 1.177B here against 1.187B on `master`, so the funding accounting costs nothing measurable next to what master itself moved. The description says that instead of the old +2.13%.

rs/ethereum/cketh/test_utils/src/live.rs
```diff
@@ -0,0 +1,649 @@
...
+        String::from_utf8_lossy(&response.body).to_string()
+    }
+
+    /// The value rendered in a dashboard row, e.g. `sweeper-cketh-burned`.
+    pub fn dashboard_row(&self, id: &str) -> Option<String> {
+        let dashboard = self.dashboard_html();
+        let row = dashboard.split(&format!(r#"id="{id}""#)).nth(1)?;
+        let cell = row.split("<td>").nth(1)?.split("</td>").next()?;
+        Some(cell.trim().to_string())
+    }
```
gregorydemay: I would not expect those here, we can better test the dashboard via unit tests. It seems those methods are not used.

gregorydemay:
🤖 **Cleanup pass on `live.rs`: dead code and duplication**

Checked every public item against all of `rs/`. The only consumer is `minter/tests/deposit_from_cex.rs` (two tests).

**Never called anywhere:**

| Item | |
|---|---|
| `dashboard_html` / `dashboard_row` | live.rs:544, 565 |
| `await_minter_log` | live.rs:527 |
| `advance_ticks` | live.rs:483 |
| `new_funding_with_empty_fee_account` | live.rs:284 |
| `await_fee_account_credited` | live.rs:314 |
| `CkEthSetup::upgrade_minter_without_changes` | lib.rs:486 |

`dashboard_row`'s doc example (`sweeper-cketh-burned`) names a row the dashboard template doesn't have — the only sweeper row is `sweeper-contract-address`. And `new_funding_with_empty_fee_account` is the only reason `new_funding_with_fee_account_balance` takes a parameter; deleting it collapses that constructor.

**`pub` with no caller outside the module:** `FEE_ACCOUNT_BALANCE`, `AWAIT_DEADLINE`, `upgrade_minter_with`, `sweeper_address`, `minter_logs`, `deposit`. Same in `anvil.rs`, where the new `deploy_deposit_helper` / `deposit_eth` are `pub` but `eth_balance` / `mine` are `pub(crate)` — all four have one caller, `live.rs`.

**Three copies of the same polling loop.** `await_scan` (245), `await_sweeper_address` (510), `await_minter_log` (527) each spell out `Instant::now()` → check → `assert!(elapsed <= deadline)` → sleep. One helper on the shared impl covers all three (`LiveSetup<CkErc20Setup>` satisfies `AsRef<CkEthSetup>`):

```rust
fn poll_until<T>(&self, deadline: Duration, what: impl Fn(&Self) -> String,
                 mut observe: impl FnMut(&Self) -> Option<T>) -> T
```

It also folds in the two other repeats: the literal 2-second sleep in all three (while `POLL_INTERVAL` sits unused next to them at 106), and the third copy of `minter_logs().join("\n")` in the failure messages.

**Two deadline conventions.** `await_scan` takes a `deadline` argument; everything else reads `AWAIT_DEADLINE`. The argument only ever gets `Duration::from_secs(180)`, hardcoded in the test — a `SCAN_DEADLINE` const drops the parameter and puts both numbers side by side.

**`deposit_helper: Option<Address>` should be a type distinction, not a runtime one.** `deposit` (364) is on the generic impl, so it exists on `LiveSetup<CkErc20Setup>` where it can only ever hit `.expect("this fixture was built without a deposit helper")`. `with_deposit_helper` has exactly one caller, the funding constructor. Moving both into `impl LiveSetup<CkEthSetup>` makes the panic unreachable by construction.

**`let _ = self.cketh_total_supply();`** in `settle` (446) is a call for its side effect with a comment explaining it keeps the instance awake — `self.keep_instance_awake()` would say that in the code.

rs/ethereum/cketh/minter/src/sweeper.rs
```diff
@@ -0,0 +1,184 @@
...
+        return FundingDecision::AlreadyInFlight {
+            ledger_burn_index: outstanding.ledger_burn_index,
+            amount: outstanding.withdrawal_amount,
+        };
+    }
+    // For the fresh deployment, where `eth_balance` starts at zero while the minter may already
+    // hold ETH no deposit credited: funding that would reach past what deposits paid in underflows
+    // the debit at finalization, and traps in the withdrawal timer. Waiting is free.
+    let available = state.eth_balance().eth_balance();
+    if available < amount {
```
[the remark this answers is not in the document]
mbjorkqvist: I worked through this one and came out the other side: the reservation would refuse fundings without preventing anything, because the check cannot fire at all while the books agree.

`eth_balance` has exactly one credit site and one debit site — `state.rs:418`, an ETH deposit, and `state.rs:457`, a finalized withdrawal. Per request the debit is at most what was burned for it: `tx.amount + effective_fee` on success, `effective_fee` alone on failure, against a `charged_tx_fee` of `withdrawal_amount - tx.amount` (or `max_transaction_fee` for ckERC20), and reimbursement returns only `tx.amount`, i.e. never more than the part that was not debited. Sweeps do not touch the counter at all. Put together, `eth_balance` stays at or above the ckETH supply in any state the accounting can reach.

A funding burns ckETH out of that supply — it comes from the fee subaccount — so `amount <= fee account balance <= supply <= eth_balance`, whatever else is pending, created or sent. Subtracting outstanding commitments would only decline fundings that are in fact covered.

What the check does catch is ckETH that no deposit backs, and that is not hypothetical: the harness used to mint straight into the fee account, and that is the state in which the funding debited ETH the minter never received and trapped the withdrawal timer at finalization — which is how this guard came to exist. The harness deposits for real now, so the check should be unreachable from here, and I have corrected its comment, which had claimed the fresh deployment as its case: a fresh deployment has an empty fee account too, so nothing is funded there. If a debit ever stops being covered by its own burn, the reservation becomes the right answer.

gregorydemay (approved):
Thanks @mbjorkqvist !

rs/ethereum/cketh/test_utils/src/live.rs
mbjorkqvist: Both halves taken. The two readers went from this PR in 2dc391e4fb along with the five other uncalled items, and after your point about unit tests they are now gone from the stack entirely. The module summary here still listed them, which is fixed in 057df0eb9f.

On that second half: those live tests were not testing the dashboard: they read it because it was the only queryable view of the funding state I had, and cheaper than polling `retrieve_eth_status`, which is an update call. Three of the four reads had a machine-readable equivalent once #11094 exports metrics — `sweeper-prepaid-gas` is `cketh_minter_sweeper_gas_balance`, and `sweeper-eth-spent` / `sweeper-burned-not-yet-spent` are the counter and the gauge beside it — so they read those now.

The fourth was the burn index of the funding in flight, and it existed only to ask `retrieve_eth_status` whether the failed funding sat in pending reimbursement. That assertion is gone rather than ported: the state is the ordinary one, imprecise here because nothing will ever settle it, and the test already pins what matters in stronger terms — no ETH at the sweeper, the supply staying reduced, spend recorded, surplus equal to burn minus spend. With it went the index, both readers and the status query, so nothing in the stack parses HTML any more.

PR #11087, opened 2026-08-10
refactor: Remove ReplicaVersion::default()

frankdavid:
The replica version is the git commit hash that the replica (and the entire GuestOS) is built from. The value is read from `/opt/ic/share/version.txt` in the GuestOS and passed to the orchestrator and the replica via CLI args. Previously, we used a global static variable to store the version. The code could access it via `ReplicaVersion::default()`. The implementation read the global variable which was usually set at the start of the program (e.g., in main). The implementation fell back to the `CARGO_PKG_VERSION` when the global was unset (in most tests). Since `ReplicaVersion` is defined in `ic_types`, the implementation used the crate version which has been 0.9.0 since Dec 2023. The replica version was stored in a global variable, so tests could not mock this value. In addition, prod code that transitively depended on `ReplicaVersion::default()` may have not known about it and thus implicitly used the confusing 0.9.0 value (which does not even match the commit hash format).

This is a major refactor which:
- explicitly wires `ReplicaVersion` to components that previously used `ReplicaVersion::default()` -> any prod code that uses the replica version must explicitly pass the version
- adds a `test_replica_version()` to be used strictly by tests only
- removes `ReplicaVersion::default()` since the default value was confusing and error-prone


daniel-wong-dfinity-org-twin:
Wow. I mean, I guess we all agree that `impl Default for ReplicaVersion` is evil, but I was not expecting anyone to actually do the surgery needed to get rid of that, since it is fairly pervasive. I mean, I would be a little bit afraid of rocking the boat on this one. OTOH, once you simply delete `impl Default for ReplicaVersion`, the bazel + rustc COMPREHENSIVELY shows you all the code that needs to be reworked, and the rework would be "easy" to verify, so I guess there is no reason to be afraid of this change...

rs/types/types_test_utils/src/ids.rs
```diff
@@ -153,6 +154,10 @@ pub fn node_test_id(i: u64) -> NodeId {
     NodeId::from(PrincipalId::new_node_test_id(i))
 }
 
+pub fn test_replica_version() -> ReplicaVersion {
```
daniel-wong-dfinity-org-twin, line 157: This is NOT a request for change. Rather, it is just a statement of general principle.

I would rather use `lazy_static` over a 0-argument fn, but in this file, there's already `user_anonymous_id`. Sigh. I guess we can sweep through later and convert to `lazy_static`.

rs/types/types_test_utils/src/ids.rs
```diff
@@ -153,6 +154,10 @@ pub fn node_test_id(i: u64) -> NodeId {
...
+    ReplicaVersion::from_str("test_replica_version").unwrap()
```
daniel-wong-dfinity-org-twin: How about a 40 char hexadecimal string like `cafebabe0000ffff0000ffff0000ffff0000ffff`? I mean, clearly, this is NOT a production value.

rs/nns/integration_tests/src/subnet_handler.rs
```diff
@@ -40,7 +39,7 @@ fn test_submit_and_accept_update_subnet_proposal() {
                 max_block_payload_size: 4 * 1024 * 1024,
                 unit_delay_millis: 500,
                 initial_notary_delay_millis: INITIAL_NOTARY_DELAY.as_millis() as u64,
-                replica_version_id: ReplicaVersion::default().into(),
+                replica_version_id: "replica_test_version".to_string(),
```
daniel-wong-dfinity-org-twin: Why not also use test_replica_version here?

daniel-wong-dfinity-org-twin (commented):
I guess this is not ready for review, since it is still in draft, and more commits are coming in, but I opened this, and started looking, so let me send some of my early incomplete thoughts...

rs/nns/integration_tests/src/subnet_handler.rs
frankdavid: Optimally, it would be nice to not use the same constant everywhere because it gives false assumptions (for example, in production a Block coming from the wire may have a different version inside than the replica's current version but then in the test both would be the same because both are wired up to `test_replica_version()`). So I tried using different constants to avoid these false equalities. However, it quickly gets annoying as many of our test mocks/factories/setups just hardcode the version and if two different factories use two different versions, the code under test will not like that (code where the requirement is legitimately that the replica versions in two components must be the same). It would have take way too much time to use different constants but then use the same constant when the test requires so. For simplicity and the sake of not wasting more time on this, I use `test_replica_version()` in most places but I still tried to use different constants in a few places where there is no requirement for the version to be the same as in some other place. In a perfect world, the test cases would explicitly set up the mocks such that versions match if and only if they have to match for the test case to pass.

blind-oracle (approved):

rs/nns/integration_tests/src/subnet_handler.rs
daniel-wong-dfinity-org-twin: SG, but then, that makes me think we should not have test_replica_version at all. Instead, all tests come up with their own local test value. WDYT?

frankdavid: Yup, we should do that. But it just takes a lot of time to set that up and I don't think anybody will every want to invest time into that. The reason is that tests often use shared mocks/setup functions. Optimally, we'd always pass the replica version to the setup function but it'd require changing a lot of setup functions. So instead, setup functions hardcode some replica version but then different setups will have different replica versions and test cases that use multiple such setup functions will have a mismatch in places where there shouldn't be a mismatch (e.g. consensus making blocks using a different replica version than what's the elected registry version). I actually tried this approach but there were just too many test failures to fix one by one so I had to introduce a shared constant.

My preferred solution would be automatic compile-time dependency injection using some library that automatically wires up the components, but introducing one is a lot of work and there'd certainly be opposers who prefer the current manual DI.

rs/types/types_test_utils/src/ids.rs
frankdavid: I'm not a huge fan of lazy_static because it's kind of a hack (a custom type that has Deref implemented on it, the destructor never runs). Whereas a value returned from a function has a clear lifetime. For performance/identity reasons, reusing a static can have many benefits but I find using a method cleaner in a test if the underlying code is not const.

rs/types/types_test_utils/src/ids.rs
frankdavid: sure

rs/consensus/dkg/src/lib.rs
```diff
@@ -1527,6 +1544,8 @@ mod tests {
                     let state_manager_2 = dependencies_2.state_manager.clone();
                     let subnet_id_1 = dependencies_1.replica_config.subnet_id;
                     let subnet_id_2 = dependencies_2.replica_config.subnet_id;
+                    let replica_config_1 = dependencies_1.replica_config;
+                    let replica_config_2 = dependencies_2.replica_config;
```
Pace222:
```suggestion
                    let replica_version_1 = dependencies_1.replica_config.replica_version;
                    let replica_version_2 = dependencies_2.replica_config.replica_version;
```
I would keep it consistent

rs/consensus/dkg/src/lib.rs
```diff
@@ -2111,15 +2133,23 @@ mod tests {
                     dkg_tag: NiDkgTag::LowThreshold,
                     target_subnet: NiDkgTargetSubnet::Remote(target_id),
                 };
-                let remote_message = create_dealing(1, remote_dkg_id);
+                let remote_message = create_dealing_with_replica_version(
+                    1,
+                    remote_dkg_id,
+                    deps.replica_config.replica_version.clone(),
+                );
```
Pace222: Why isn't `create_dealing` sufficient?

rs/consensus/dkg/src/utils.rs
```diff
@@ -291,6 +291,10 @@ mod tests {
         parent: &Block,
         dkg_payload: DkgDataPayload,
     ) -> BlockProposal {
+        let replica_version = dkg_payload
+            .messages
+            .first()
+            .map_or_else(test_replica_version, |m| m.content.version.clone());
```
Pace222: Just for simplicity, I think I would use the `parent`'s version instead

rs/consensus/src/consensus/status.rs
```diff
@@ -41,6 +41,7 @@ pub(crate) fn get_status(
     subnet_id: SubnetId,
     pool: &PoolReader<'_>,
     logger: &ReplicaLogger,
+    replica_version: &ReplicaVersion,
```
Pace222, line 43: nit: I would try to keep the `logger` argument last

rs/consensus/src/consensus/batch_delivery.rs
```diff
@@ -53,6 +53,7 @@ pub fn deliver_batches(
     registry_client: &dyn RegistryClient,
     subnet_id: SubnetId,
     log: &ReplicaLogger,
+    replica_version: &ReplicaVersion,
```
Pace222: I would try to keep the `log` argument last (second-to-last here actually)

rs/consensus/src/consensus/batch_delivery.rs
```diff
@@ -122,7 +125,6 @@ pub(crate) fn deliver_batches_with_result_processor(
             );
             break;
         };
-        let replica_version = block.version().clone();
```
Pace222, line 125: I think we could use the same `replica_version` everywhere in this function. I.e. use the block's version instead of our own replica version in `get_status`, dropping the `current_replica_version` argument. That would be a change of behaviour though, but they should always be the same.
Maybe cc @eichhorl to have a second thought

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -787,6 +795,14 @@ mod tests {
             )
             .build();
 
+            // The block-making schedule depends on the random state set up by dependencies.
+            // For this test, we simulate the blockmaker running on node with ID 1.
+
+            // TODO: once we have an easier way to pass parameters to
+            // dependencies_with_subnet_params, pass the NodeID, so it's consistently applied to
+            // all test components.
+            replica_config.node_id = node_test_id(1);
```
Pace222:
```suggestion
            .with_replica_config(...)
            .build();
```

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -1057,7 +1076,7 @@ mod tests {
     #[test]
     fn test_halting_due_to_protocol_upgrade() {
         test_halting(
-            ReplicaVersion::try_from("0xBEEF").unwrap(),
+            ic_test_utilities_types::ids::test_replica_version(),
```
Pace222: I'm surprised this tests passes with this. Shouldn't we use a different replica version than the one currently running?

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -1204,7 +1226,7 @@ mod tests {
             let proposal = proposal.unwrap();
             let block = proposal.content.as_ref();
             // blocks still uses default version, not the new version.
```
Pace222: Outdated comment

rs/consensus/src/consensus/notary.rs
```diff
@@ -299,6 +306,7 @@ fn get_adjusted_notary_delay_from_settings(
     membership: &Membership,
     rank: Rank,
     logger: &ReplicaLogger,
+    replica_version: &ReplicaVersion,
```
Pace222, line 308: nit: I would try to keep the `logger` argument last

rs/consensus/src/consensus/validator.rs
```diff
@@ -2537,12 +2546,13 @@ pub mod test {
             // Put a random tape share in the unvalidated pool
             let pool_reader = PoolReader::new(&pool);
             let beacon_1 = pool_reader.get_random_beacon(Height::from(1)).unwrap();
-            let beacon_2 = RandomBeacon::from_parent(&beacon_1);
+            let mut beacon_2 = RandomBeacon::from_parent(&beacon_1);
+            beacon_2.content.version = replica_config.replica_version.clone();
```
Pace222: Why is this line necessary?

rs/consensus/src/consensus/validator.rs
```diff
@@ -4401,6 +4434,13 @@ pub mod test {
             block.signature.signer = incorrect_signer;
             pool.insert_unvalidated(block);
 
+            dbg!(
+                pool.validated()
+                    .block_proposal()
+                    .get_all()
+                    .collect::<Vec<_>>()
+            );
+
```
Pace222:
```suggestion
```

rs/consensus/tests/payload.rs
```diff
@@ -126,6 +131,7 @@ fn consensus_produces_expected_batches() {
                 1,
                 SubnetRecordBuilder::from(&[node_test_id(0)])
```
Pace222:
```suggestion
                SubnetRecordBuilder::from(&[replica_config.node_id])
```
Sorry, not your changes. But for consistency

rs/consensus/utils/src/lib.rs
```diff
@@ -434,12 +434,6 @@ fn get_subnet_splitting_status_at_given_summary(
 ///
 /// # Returns
 ///
```
Pace222:
```suggestion
```
(GitHub is broken, but I was suggesting to remove the entire now-outdated comment block)

rs/crypto/prng/tests/tests.rs
```diff
@@ -163,6 +132,7 @@ fn fake_random_beacon(height: u64) -> RandomBeacon {
         content: RandomBeaconContent::new(
             Height::from(height),
             CryptoHashOf::new(CryptoHash(vec![])),
+            ReplicaVersion::from_str("0.8.0").unwrap(),
```
Pace222, line 137: Maybe add a comment explaining this

rs/execution_environment/src/scheduler/tests/metrics.rs
```diff
@@ -627,7 +627,7 @@ fn threshold_signature_agreements_metric_is_updated() {
     let schnorr_key_id = make_schnorr_key_id(0);
     let master_schnorr_key_id = MasterPublicKeyId::Schnorr(schnorr_key_id.clone());
     let mut test = SchedulerTestBuilder::new()
-        .with_replica_version(ReplicaVersion::default())
+        .with_replica_version(ReplicaVersion::from_str("foobar").unwrap())
```
Pace222: Does this need to be a special case?

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -304,7 +304,7 @@ mod tests {
             content_hash: CryptoHashOf::new(CryptoHash(vec![])),
             content_size,
             is_reject: false,
-            replica_version: ReplicaVersion::default(),
+            replica_version: ReplicaVersion::try_from("test_replica_version").unwrap(),
```
Pace222:
```suggestion
            replica_version: test_replica_version(),
```

rs/messaging/src/state_machine/tests.rs
```diff
@@ -820,7 +821,7 @@ fn test_online_split(new_subnet_id: SubnetId, other_subnet_id: SubnetId) -> Repl
             .checked_add(Duration::from_secs(1))
             .unwrap(),
         blockmaker_metrics: BlockmakerMetrics::new_for_test(),
-        replica_version: ReplicaVersion::default(),
+        replica_version: ReplicaVersion::from_str("foo").unwrap(),
```
Pace222, line 824: Does this need to be a special case?

rs/prep/src/bin/prep.rs
```diff
@@ -181,7 +181,9 @@ fn main() -> Result<()> {
             Some(fetch_replica_version_sha256(replica_version_id.clone())?);
     }
 
-    let replica_version = valid_args.replica_version_id.unwrap_or_default();
+    let replica_version = valid_args
+        .replica_version_id
+        .unwrap_or_else(|| ReplicaVersion::from_str("unknown").unwrap());
```
Pace222: I am not sure where this script is used, but would it make sense to panic if `replica_version_id` is not given? (Or just remove the `Option` wrapper in `CliArgs`)

rs/replay/src/player.rs
```diff
@@ -243,7 +245,7 @@ impl Player {
             // Use the replica version from the finalized tip in the pool.
             PoolReader::new(pool).get_finalized_tip().version().clone()
         } else {
-            Default::default()
+            ReplicaVersion::from_str("unknown").unwrap()
```
Pace222: I don't think it's worth continuing the program with this value. Could you try to replace it with a panic and see if any test fails?

rs/replay/src/player.rs
```diff
@@ -270,9 +272,6 @@ impl Player {
         _async_log_guard: AsyncGuard,
     ) -> Self {
         println!("Setting default replica version {replica_version}");
```
Pace222:
```suggestion
```

rs/replay/src/validator.rs
```diff
@@ -125,7 +126,11 @@ impl ReplayValidator {
             Height::from(0),
         ));
         let node_id = NodeId::from(PrincipalId::new_node_test_id(1));
-        let replica_cfg = ReplicaConfig::new(node_id, subnet_id);
+        let replica_cfg = ReplicaConfig {
+            node_id,
+            subnet_id,
+            replica_version: replica_version.clone(),
```
Pace222:
```suggestion
            replica_version,
```

rs/replay/src/player.rs
```diff
@@ -711,6 +713,7 @@ impl Player {
                 &*self.registry,
                 self.subnet_id,
                 &self.log,
+                replica_version,
```
Pace222: I think you could just read from `self.replica_version` instead of taking an additional argument.

Same for `message_routing` (sorry not your change)

rs/replica/setup_ic_network/src/lib.rs
```diff
@@ -661,7 +669,11 @@ fn start_consensus(
             Arc::new(Mutex::new(canister_http_adapter_client)),
             Arc::clone(&consensus_crypto),
             Arc::clone(&consensus_pool_cache),
-            ReplicaConfig { subnet_id, node_id },
+            ReplicaConfig {
+                subnet_id,
+                node_id,
+                replica_version: replica_version.clone(),
+            },
```
Pace222:
```suggestion
            replica_config,
```

rs/replica/setup_ic_network/src/lib.rs
```diff
@@ -614,6 +621,7 @@ fn start_consensus(
         ic_consensus_dkg::DkgImpl::new(
             node_id,
             subnet_id,
+            replica_config.replica_version.clone(),
```
Pace222:
```suggestion
            replica_version,
```
for consistency

rs/test_utilities/artifact_pool/src/consensus_pool.rs
```diff
@@ -311,12 +316,17 @@ impl TestConsensusPool {
 
     pub fn make_next_beacon(&self) -> RandomBeacon {
         let beacon = self.validated().random_beacon().get_highest().unwrap();
-        RandomBeacon::from_parent(&beacon)
+        let mut beacon = RandomBeacon::from_parent(&beacon);
+        beacon.content.version = self.replica_version.clone();
```
Pace222: Is this necessary if `from_parent` uses the parent's version?

rs/test_utilities/consensus/src/lib.rs
```diff
@@ -228,6 +228,7 @@ pub fn make_genesis(summary: DkgSummary) -> CatchUpPackage {
         .unwrap()
         .dkg_id
         .clone();
+    let replica_version = ic_test_utilities_types::ids::test_replica_version();
```
Pace222: You could import it

rs/types/types/src/replica_version.rs
```diff
@@ -5,40 +5,17 @@ use std::convert::TryFrom;
...
-        ReplicaVersion {
-            version_id: DEFAULT_VERSION_ID
-                .get_or_init(|| env!("CARGO_PKG_VERSION").to_string())
-                .clone(),
-        }
-    }
-}
-
 #[derive(Eq, PartialEq, Debug)]
 pub struct DefaultVersionAlreadySetError;
```
Pace222:
```suggestion
```
We should be able to delete this too

rs/types/types/src/replica_version.rs
```diff
@@ -122,31 +99,9 @@ mod test {
...
-        assert_eq!(
-            ReplicaVersion::default().as_ref(),
-            env!("CARGO_PKG_VERSION")
-        );
-        let version = ReplicaVersion::try_from("1.2.1").unwrap();
-        assert!(ReplicaVersion::set_default_version(version).is_err());
-    }
-
-    use rusty_fork::rusty_fork_test;
-    rusty_fork_test! {
```
Pace222, line 143: You should be able to remove the dependency to `rusty_fork` from that crate

Pace222 (commented):
Two suggestions maybe worth exploring:
- Similar to the `FakeContent<T>` trait, maybe we could introduce a trait that exposes a method `fake`. Each artifact would implement it and pass `test_replica_version`. Now all tests would replace `ArtifactName::new(..., test_replica_version())` with `ArtifactName::fake(...)`. I haven't looked at all artifacts, so I am not sure this is 100% feasible, but it would give the flexibility to mock certain fields at a single place.
- More involved, but we could introduce a new
```rust
pub struct Versioned<T> {
    pub version: RepilcaVersion,
    pub content: T,
}
```
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;struct and replace all artifact aliases from `Signed<T, S>` to `Signed<Versioned<T>, S>`. This would definitely involve even more plumbing, but we could enforce versioning of artifacts at the type-level and collapse duplicate implementations of checking version matches to a single place.

rs/https_outcalls/pricing/src/fees.rs
frankdavid: We don't really need the dependency here since the replica versino does not have to be consistent with other parts of the code (e.g. mocks). I'll just use foobar_replica_version constant to make it clear.

rs/execution_environment/src/scheduler/tests/metrics.rs
frankdavid: See this comment: https://github.com/dfinity/ic/pull/11087#discussion_r3751039194

Optimally, we should special case every single test and only make two versions equal when they have to be to make it explicit when two code paths expect matching versions and when they don't. It's quite error prone to assume that all the versions will always be the same. Unfortunately, it'd be too much effort to properly fix all mocks to properly do this, so reusing the same constant is a necessary bad. However, in a few tests where it was clear that they pass even with a different value, I wanted to make it clear and do not depend on the constant (same in other places).

rs/messaging/src/state_machine/tests.rs
frankdavid: same as above

rs/consensus/src/consensus/block_maker.rs
frankdavid: Was an oversight from my part, reverted.

rs/replay/src/player.rs
frankdavid: Reorganized the code and added a panic above.

rs/execution_environment/src/scheduler/tests/metrics.rs
frankdavid: Actually, not in this case.

rs/crypto/prng/tests/tests.rs
frankdavid: Added comment - version doesn't matter

rs/messaging/src/state_machine/tests.rs
frankdavid: It doesn't have to depend on test_replica_version, any value works. Whenever we can, we should just use different constants.

rs/consensus/dkg/src/lib.rs
frankdavid: It is (it remained from a previous version but it's no longer necessary).

rs/consensus/src/consensus/validator.rs
frankdavid: We can remove it. Theoretically, it could happen that the beacon's child has a different replica version but since everything is wired up based on the replica_config in ValidatorAndDependencies, it doesn't make a difference. In a previous version, I tried to decouple the version values but it was too much work.

rs/prep/src/bin/prep.rs
frankdavid: Done

rs/test_utilities/artifact_pool/src/consensus_pool.rs
frankdavid: The parent could theoretically have a different replica version but in practice it does not make a difference in the test. Would be interesting to have test cases where we make it different.

frankdavid:
Thanks, @Pace222 for your comments and suggestions. I addressed the simpler ones. As for the larger refactorings, I think they are cool ideas, but I'd prefer if the consensus team could think about the right approach and implement the changes as necessary.

rs/replay/src/player.rs
pierugo-dfinity:
```suggestion
        consensus_pool: ConsensusPoolImpl,
```

rs/test_utilities/artifact_pool/src/consensus_pool.rs
```diff
@@ -294,6 +298,7 @@ impl TestConsensusPool {
         block.context.time += monotonic_block_increment;
 
         block.context.registry_version = registry_version;
+        block.version = self.replica_version.clone();
```
pierugo-dfinity, line 301:
```suggestion
```
Looks superfluous

frankdavid: I think this one actually makes sense. When we build a block on top of an existing block, the replica version of the new block is not (necessarily) the same as the base. It's completely legal to make a block with RV=2 while the parent has RV=1. The pool should not automatically assume that the versions will be the same but rather add its own replica_version to the created block.

pierugo-dfinity: Ok true fair point

basvandijk (approved):

rs/nns/integration_tests/src/upgrades_handler.rs
```diff
@@ -139,14 +139,20 @@ fn test_submit_and_accept_update_elected_replica_versions_proposal() {
...
-            default_version,
+            &default_version,
             version_to_elect_and_unelect1,
             version_to_elect_and_unelect2,
             unassigned_nodes_version,
         ] {
-            assert!(is_elected_version(&nns_canisters.registry, version).await);
+            assert!(
+                is_elected_version(&nns_canisters.registry, version).await,
+                "Expected {version} to be elected"
```
daniel-wong-dfinity-org-twin, line 149: 😎

I wonder if Claude was trying to obey ["leave breadcrumbs"][bread] here 🤔

[bread]: https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?L343&view=raw

daniel-wong-dfinity-org-twin (approved):
approving for Governance-owned files.

UPDATE: Strike out the following, because I already said this earlier.

<strike>
For constants like test_replica_version_id, I would instead do

```
lazy_static! {
    pub static ref TEST_REPLICA_VERSION_ID ...
}
```

unless, ofc, that would be inconsistent with the crate where it is defined.

The benefit is that this makes it clear that this thing is fixed. A 0-argument function CAN return different values every time (e.g. random number generators). A lazy_static + all caps says, "No, really. This is FIXED!". That's why (when possible), we use `const` + all caps instead of 0-argument functions. It gives a helpful hint to the reader.
</strike>

daniel-wong-dfinity-org-twin:
Also, for the record, `impl Default for ReplicaVersion` is evil. If there is a not high risk way to get rid of it, great. I would be afraid to do that, but maybe, with AI, it is actually highly feasible.

rs/types/types_test_utils/src/ids.rs
daniel-wong-dfinity-org-twin: yes, lazy_static is pretty magical, but you do not have to think about it very much.

It's not about performance. I think it is clearer. 0-argument functions often return different values. lazy_static + all caps gives an important hint to the reader: this is really really FIXED!

> if the underlying code is not const.

I don't think this is a good reason. const is "purely technical". What matters is that CONCEPTUALLY, the thing is constant. If so, then, lazy_static can help us get around the technical limitation.

Another way is just to implement const new.

rs/replay/src/player.rs
```diff
@@ -222,35 +223,36 @@ impl Player {
...
+        if !cfg.artifact_pool.consensus_pool_path.exists() {
+            panic!(
+                "No consensus pool found at {:?}",
+                cfg.artifact_pool.consensus_pool_path
             );
-            Some(consensus_pool)
-        } else {
-            None
-        };
+        }
```
pierugo-dfinity: @eichhorl Can you think of a use-case for `ic-replay` without a consensus pool? 

With the suggested changes, `Player::consensus_pool` will always be `Some`, but there is quite some deeper amount of code that properly handles a missing pool. Not sure what to set the `replica_version` to in that case though...

rs/replica/setup_ic_network/src/lib.rs
```diff
@@ -661,7 +669,7 @@ fn start_consensus(
             Arc::new(Mutex::new(canister_http_adapter_client)),
             Arc::clone(&consensus_crypto),
             Arc::clone(&consensus_pool_cache),
-            ReplicaConfig { subnet_id, node_id },
+            replica_config.clone(),
```
pierugo-dfinity:
```suggestion
            replica_config,
```

pierugo-dfinity (commented):
Would approve but I'd gladly first let Leo have a look at the two locations I pinged him at 🙏 And thanks for the plumbing!

rs/replica/bin/replica/main.rs
```diff
@@ -182,16 +182,20 @@ fn main() -> io::Result<()> {
         .as_ref()
         .map(|proto| CatchUpPackage::try_from(proto).expect("deserializing CUP failed"));
 
-    // Set the replica version and report as metric
-    setup::set_replica_version(&replica_args, &logger);
+    const UNKNOWN_REPLICA_VERSION: &str = "unknown_replica_version";
+    let replica_version = replica_args.as_ref().map_or_else(
+        |_| ReplicaVersion::try_from(UNKNOWN_REPLICA_VERSION).unwrap(),
```
eichhorl, line 187: Why is this needed? I.e. why not panic?

rs/consensus/dkg/src/lib.rs
```diff
@@ -78,6 +79,7 @@ impl DkgImpl {
     pub fn new(
         node_id: NodeId,
         subnet_id: SubnetId,
+        replica_version: ReplicaVersion,
```
eichhorl: I guess this could also take the `ReplicaConfig` instead

rs/types/types/src/replica_version.rs
```diff
@@ -5,43 +5,17 @@ use std::convert::TryFrom;
...
 use std::str::FromStr;
+use std::sync::Arc;
 
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize, Serialize)]
-pub struct ReplicaVersion {
-    version_id: String,
-}
-
-static DEFAULT_VERSION_ID: OnceCell<String> = OnceCell::new();
 pub static REPLICA_BINARY_HASH: OnceCell<String> = OnceCell::new();
```
eichhorl, line 10: why is this needed?

rs/test_utilities/consensus/src/fake.rs
```diff
@@ -254,24 +260,25 @@ impl FakeContentSigner<&RandomBeacon> for RandomBeaconShare {
...
             content: beacon,
             signature,
         }
     }
 }
 
-impl FakeContentSigner<Height> for RandomTapeShare {
-    fn fake(height: Height, signer: NodeId) -> RandomTapeShare {
+impl FakeContentSigner<RandomTapeContent> for RandomTapeShare {
+    fn fake(content: RandomTapeContent, signer: NodeId) -> RandomTapeShare {
```
eichhorl: Why not keep it as `FakeContentSigner<Height>` and set the test replica version here?

rs/consensus/src/consensus/batch_delivery.rs
eichhorl: I agree they should be the same, but it introduces a dependency on the validator to continue ensuring that. No strong preference though

rs/replay/src/player.rs
eichhorl: I think there could be a scenario where we might want to make multiple changes to a recovery state over multiple calls to ic-replay. In that case, only the first call replays the pool, every subsequent call only applies extra batches on top of the checkpoint that the previous call created.

There are also some sub-commands that don't need a pool.

If we have a pool, then we could probably also use `ic_artifact_pool::get_replica_version`. But if there is no pool, and a replica version is needed, then it could be a new CLI argument.

rs/consensus/src/consensus/batch_delivery.rs
pierugo-dfinity: I think it makes sense in this component to behave based on the information of the block. The block is finalized at that point, so I think it is reasonable to trust that the validator has done its job. Then it also guarantees that the halting decision is consistent across all nodes (i.e. we trust the blockchain == consensus, instead of our own view). (+ it simplifies the code)

rs/replay/src/player.rs
frankdavid: https://github.com/dfinity/ic/pull/11087/changes/226e9cd15cc508e93c768eeb8ef1d7786bd453a1 

Is this correct?

rs/registry/admin/bin/create_subnet.rs
```diff
@@ -414,7 +408,7 @@ mod tests {
             max_block_payload_size: None,
             unit_delay_millis: None,
             initial_notary_delay_millis: None,
-            replica_version_id: None,
+            replica_version_id: ReplicaVersion::from_str("").unwrap(),
             dkg_interval_length: None,
```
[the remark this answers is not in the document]
frankdavid: It's only a test

rs/replica/bin/replica/main.rs
frankdavid: The code is written with the assumption that replica_args has Result type and thus can be an error. I didn't want to change this assumption.

basvandijk (approved):

rs/types/types/src/replica_version.rs
frankdavid: We expose the hash over the status endpoint. It was there already before, I just moved a few lines around.

rs/replay/src/player.rs
pierugo-dfinity: I think I would keep `Player::replica_version` non-optional and then panic at `player.rs` new line 251 if the the `replica_version` argument is not given.

rs/consensus/src/consensus/batch_delivery.rs
frankdavid: https://github.com/dfinity/ic/pull/11087/changes/8687733911c374aa48b717c2472bfad7a9389862

NikolaMilosa (approved):

rs/replay/src/player.rs
frankdavid: https://github.com/dfinity/ic/commit/2b96c59a12c4c27e6b2174ec7db78b835fed5f7c

rs/consensus/src/consensus/batch_delivery.rs
pierugo-dfinity: nit: let's just reuse the existing style and assign the `block.version().clone()` once to a unique `replica_version` variable that is used everywhere

pierugo-dfinity (approved):

rs/replay/src/player.rs
pierugo-dfinity: Thanks

eichhorl (approved):

rs/registry/admin/bin/create_subnet.rs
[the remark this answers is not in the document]
Bownairo: 👋
https://github.com/dfinity/ic/pull/11136/changes#diff-ff661c771f4abe23055a169df9ce5e054e4b48bd82200b055073d59b980a6c67R411

PR #11092, opened 2026-08-10
fix(governance): don't panic when a node provider has no id

Rachit2323:
`validate_assign_noid_payload` was calling `.unwrap()` on `np.id`, which is an `Option`. If any node provider in governance state has `id = None`, this panics and blocks all `AddNodeOperator` proposal submissions.

Fixed by comparing the `Option` values directly instead of unwrapping them.


Rachit2323:
Addressed the checklist:
1. Added entry to unreleased_changelog.md
2. No breaking changes — existing APIs behave as before
3. No data migration needed
4. No security concern — defensive fix against a latent panic

rs/nns/governance/src/governance.rs
```diff
@@ -5024,7 +5024,7 @@ impl Governance {
 
         let is_registered = node_providers
             .iter()
-            .any(|np| np.id.unwrap() == decoded_payload.node_provider_principal_id.unwrap());
+            .any(|np| np.id == decoded_payload.node_provider_principal_id);
```
daniel-wong-dfinity-org-twin: This seems like a bug: if any element in node_providers has id set to None, then decoded_payload.node_provider_principle_id is allowed to be None.

If the goal here is to stop `np.id.unwrap()` from panicking (reasonable), I think the way to do that is more like this:

```suggestion
        if let Some(node_provider_id_of_node_operator) = decoded_payload.node_provider_principal_id else {
            return Err("The payload's node_provider_principal_id field was None".to_string());
        }

        let is_registered = node_providers
            .iter()
            .any(|np| {
                let Some(np) = np else {
                    log("This is not supposed to happen!"); // Do not take this line literally!
                    return false;
                };
                
                np.id == node_provider_id_of_node_operator
            });
```

rs/nns/governance/src/governance/tests/mod.rs
```diff
@@ -1399,6 +1399,61 @@ fn test_validate_execute_nns_function() {
...
+            ],
+            ..Default::default()
+        },
+        Arc::new(MockEnvironment::new(vec![], 100)),
+        Arc::new(StubIcpLedger {}),
+        Arc::new(StubCMC {}),
+        Box::new(MockRandomness::new()),
+    );
+
+    let make_valid = |principal_id| {
```
daniel-wong-dfinity-org-twin:
```suggestion
    let new_valid_assign_node_operator_proposal_action = |principal_id| {
```

rs/nns/governance/src/governance/tests/mod.rs
```diff
@@ -1399,6 +1399,61 @@ fn test_validate_execute_nns_function() {
...
+    let result_registered =
+        governance.validate_execute_nns_function(&make_valid(PrincipalId::new_node_test_id(1)));
+
+    // Step 3: Verify result(s).
+    // Unregistered provider → clean error, no panic.
+    let err = result_unregistered.unwrap_err();
+    assert!(err.error_message.contains("not registered"));
+    // Registered provider → ok.
+    assert_eq!(result_registered, Ok(()));
+}
```
daniel-wong-dfinity-org-twin, line 1460: I guess the main assert here is implicit: no panic. If so, please add a comment that says this.

rs/nns/governance/src/governance/tests/mod.rs
```diff
@@ -1399,6 +1399,61 @@ fn test_validate_execute_nns_function() {
...
+#[test]
+fn test_validate_assign_noid_tolerates_node_provider_with_none_id() {
+    // Step 1: Prepare the world.
+    // Mix a legacy entry (id = None) with a valid entry to ensure neither panics nor false match.
+    let governance = Governance::new(
+        api::Governance {
+            economics: Some(api::NetworkEconomics::with_default_values()),
+            node_providers: vec![
+                api::NodeProvider {
+                    id: None,
```
daniel-wong-dfinity-org-twin, line 1416:
```suggestion
                    // This used to cause a panic in the code under test,
                    // whereas, now, it just logs a warning.
                    id: None,
```

daniel-wong-dfinity-org-twin (commented):
LG.

rs/nns/governance/src/governance/tests/mod.rs
daniel-wong-dfinity-org-twin: I see no new comment here?

rs/nns/governance/src/governance.rs
```diff
@@ -5018,13 +5018,21 @@ impl Governance {
...
             return Err("The payload's node_provider_principal_id field was None".to_string());
-        }
+        };
 
         let is_registered = node_providers
             .iter()
-            .any(|np| np.id == decoded_payload.node_provider_principal_id);
+            .any(|np| {
+                let Some(np_id) = np.id else {
+                    return false;
```
daniel-wong-dfinity-org-twin: Can you log in this case?

(I believe I asked for this in my first review.)

rs/nns/governance/src/governance/tests/mod.rs
```diff
@@ -1441,10 +1443,12 @@ fn test_validate_assign_noid_tolerates_node_provider_with_none_id() {
...
 
     // Step 2: Run the code under test.
-    let result_unregistered =
-        governance.validate_execute_nns_function(&make_valid(PrincipalId::new_node_test_id(99)));
-    let result_registered =
-        governance.validate_execute_nns_function(&make_valid(PrincipalId::new_node_test_id(1)));
+    let result_unregistered = governance.validate_execute_nns_function(
+        &new_valid_assign_node_operator_proposal_action(PrincipalId::new_node_test_id(99)),
+    );
+    let result_registered = governance.validate_execute_nns_function(
```
daniel-wong-dfinity-org-twin:
```suggestion
    let unregistered_node_provider_result = governance.validate_execute_nns_function(
        &new_valid_assign_node_operator_proposal_action(PrincipalId::new_node_test_id(99)),
    );
    let registered_node_provider_result = governance.validate_execute_nns_function(
```

rs/nns/governance/src/governance/tests/mod.rs
daniel-wong-dfinity-org-twin: Ah. I now see "no panic". I was hoping for something more prominent, since the whole point of this PR is to not panic, but I guess it's not that important.

rs/nns/governance/src/governance/tests/mod.rs
```diff
@@ -1427,7 +1429,7 @@ fn test_validate_assign_noid_tolerates_node_provider_with_none_id() {
         Box::new(MockRandomness::new()),
     );
 
-    let make_valid = |principal_id| {
+    let new_valid_assign_node_operator_proposal_action = |principal_id| {
         let payload = Encode!(&AddNodeOperatorPayload {
             node_provider_principal_id: Some(principal_id),
```
daniel-wong-dfinity-org-twin:
```suggestion
    let new_valid_assign_node_operator_proposal_action = |node_provider_principal_id| {
        let payload = Encode!(&AddNodeOperatorPayload {
            node_provider_principal_id: Some(node_provider_principal_id),
```

Rationale: In general, principals are pretty generic. In this case, multiple principals are potentially involved, because a node operator is associated with a node provider. Therefore, we should indicate the role.

daniel-wong-dfinity-org-twin (commented):
Sorry. I know some of the feedback in this batch could have been mentioned earlier. I didn't spot them before. Please, bear with me. Thank you!

daniel-wong-dfinity-org-twin (approved):

rs/nns/governance/src/governance.rs
```diff
@@ -5018,13 +5018,22 @@ impl Governance {
...
+            let Some(np_id) = np.id else {
+                println!(
+                    "{}Skipping node provider with no id while checking registration.",
+                    LOG_PREFIX,
+                );
+                return false;
+            };
+
+            np_id == node_provider_id_of_node_operator
+        });
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: Good point. I think such a case would never happen IRL though. Do you think it is possible? If not, I think we need to cut off the adding even more complexity.

daniel-wong-dfinity-org-twin:
I believe this is now destined for master, the last step being checks ☝️

PR #11094, opened 2026-08-10
feat(cketh): sweeper funding observability

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding). Targets `master`; #11097 follows.

## Why

The funding task shipped in #11086, but nothing makes its accounting visible. An operator cannot see whether ckETH is still fully backed, nor whether funding has stopped happening — and a funding that gets stuck blocks every later one, so sweeping starves without anything else looking wrong.

## The metrics

Eight, all under `cketh_minter_sweeper_funding_`:

| metric | |
|---|---|
| `cketh_burned_total`, `eth_spent_total` | the two sides of the invariant: burned must never fall below spent |
| `finalized_total{status="success"\|"failure"}` | fundings that finalized, by outcome, so both throughput and a failure rate are expressible. Failures are expected to stay at zero — funding is a bare transfer to an address derived from the minter's own key, with no code to revert in — so any non-zero value wants a look |
| `burned_not_yet_spent` | how far burn runs ahead of spend |
| `gas_balance` | the prepaid-gas bound the funding decision reads |
| `low_water_mark`, `target` | the thresholds it compares that bound against |
| `in_flight_age_seconds` | how long the outstanding funding has been outstanding, 0 if none |

**The alert to write is `gas_balance < low_water_mark`, held for a couple of hours.** That is "the sweeper needs gas and is not getting it", and it catches every cause at once: a refusal for want of deposit-backed ETH, a burn failing on an empty fee account, a funding wedged in the pipeline, or the task not running at all. It stays quiet on an idle minter whose sweeper is still full, which is why it is this rather than an age-since-last-funding. Exporting the two thresholds is what keeps it from hardcoding a number that moves with `cketh_minimum_withdrawal_amount`.

`in_flight_age_seconds` is then the diagnostic rather than the alert: when the predicate above fires, a climbing age says the cause is a wedged funding, and the dashboard row says which one.

`burned_not_yet_spent` is worth one caution: for a funding that succeeded it is the same wei `cketh_minter_total_unspent_tx_fees` already counts, so summing the two double counts. Proposed alert conditions live on [DEFI-2965](https://dfinity.atlassian.net/browse/DEFI-2965).

## The dashboard

A sweeper-funding section with seven rows: the gas-paying address derived from the minter's key, the prepaid-gas bound, the low-water mark and target, ckETH burned and ETH spent, the funding in flight if there is one, and the unspent burn.

It renders an in-flight funding's acceptance time as a timestamp rather than an age, because `DashboardTemplate::from_state` must stay callable outside a canister — `ic_cdk::api::time()` traps in unit tests — and ages belong in metrics anyway.

A funding also travels the ordinary withdrawal pipeline, so it appears in the withdrawal table; it is marked there rather than reading as an unexplained ckETH withdrawal to an address nobody recognises.

[DEFI-2965]: https://dfinity.atlassian.net/browse/DEFI-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -159,6 +159,61 @@ impl SweeperFundingAccounting {
...
+    if age_nanos > MAX_SWEEPER_BALANCE_AGE_NANOS {
+        return Err(PrepaidGasUnavailable::Stale { age_nanos });
+    }
+    if observed.balance < required {
+        return Err(PrepaidGasUnavailable::Insufficient {
+            available: observed.balance,
+            required,
+        });
+    }
+    Ok(observed.balance)
```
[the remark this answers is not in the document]
mbjorkqvist: The observation about the function's semantics is correct — it is a per-request predicate against a snapshot, and two 6-wei sweeps would both pass against a 10-wei observation. I have made that explicit in the doc comment (cd36d1a1), since the previous wording invited exactly this reading.

I don't think the conclusion follows, though, on two counts.

**The backing invariant does not rest on this check.** `record_finalized_funding` adds the whole `transferred` amount to `cumulative_transferred` as soon as the funding transaction finalizes, and `cumulative_spent() = transferred + fees`. Every wei at the sweeper address is therefore already covered by a burn that preceded the ETH moving, so drawing that balance down — even over-eagerly — cannot make cumulative spend outrun cumulative burn. What over-authorising costs is a sweep the sweeper cannot pay for: wasted signatures, nonces and outcalls, not under-backed ckETH.

**There is no caller yet.** Nothing in production calls `check_prepaid_sweep_gas`; this PR adds the observation cache and the predicate, while sweeping itself is [S2] ([DEFI-2926](https://dfinity.atlassian.net/browse/DEFI-2926)), still in the backlog. The function takes `Option<ObservedSweeperBalance>` by value, so it cannot reserve anything by construction — reserving across several sweeps is the caller's job.

So rather than restructure a function whose caller does not exist, I have recorded the requirement on DEFI-2926: the sweeping task must subtract what it has already committed since the last observation, either by tracking outstanding reservations in state and including them in `required` or by reconciling each against its receipt. Flagging it here was useful — daily observations mean many sweeps can fall between two of them, so this would have been easy to get wrong.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -1128,6 +1128,58 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+                    "Cumulative ckETH burned from the fee account to prepay sweep gas.",
+                )?;
+                w.encode_counter(
+                    "cketh_minter_sweeper_funding_eth_spent_total",
+                    s.sweeper_funding.cumulative_spent().as_f64(),
+                    "Cumulative ETH debited from the minter's main address for sweeping.",
+                )?;
+                w.encode_gauge(
+                    "cketh_minter_sweeper_funding_burned_not_yet_spent",
+                    s.sweeper_funding.burned_not_yet_spent().as_f64(),
```
mbjorkqvist, line 1361: A note on how this gauge sits next to the pre-existing `cketh_minter_total_unspent_tx_fees`, since this is the PR that puts the funding numbers beside it.

For a user withdrawal the ckETH burned is `withdrawal_amount` while the ETH leaving the main address is `delivered + effective`, so `ETH at main address − ckETH supply == Σ (charged − effective) == total_unspent_tx_fees` — the figure `docs/cketh.adoc` describes as fees charged but in retrospect not needed. A funding behaves the same way, since it burns its full amount and never discounts a later burn, so the identity still holds with funding live.

What does not hold is reading the two side by side. For a funding that succeeds, `burned_not_yet_spent` grows by exactly `charged − effective` — the same wei `total_unspent_tx_fees` just counted — so summing them double counts. For one that fails they diverge instead: the whole undelivered burn stays as backing, while the unspent-fee figure only sees the fee difference. Composing them means picking the user-withdrawal part of the older counter, which nothing exposes separately today.

I would rather not add a derived over-backing gauge here for the sake of it. Flagging it so it is on the record when the alerts on DEFI-2965 are written: alert on burned versus spent, not on a sum of these two.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -1198,6 +1198,59 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+                     provisioned but never paid. Nothing draws it down -- the surplus stays as \
+                     ckETH backing -- so between fundings it only ratchets up.",
+                )?;
+                w.encode_gauge(
+                    "cketh_minter_sweeper_gas_balance",
+                    s.sweeper_funding.sweeper_balance_lower_bound().as_f64(),
+                    "Prepaid sweep gas: a lower bound on the sweeper address' ETH balance, from the \
+                     fundings the minter has recorded as finalized -- not an observation of the \
+                     chain, and never above the true balance.",
+                )?;
```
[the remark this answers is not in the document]
mbjorkqvist: Right on both counts, and the fix is the description: it still described the design this stack started with, where the funding task read the balance through the EVM RPC canister and cached it, and the gate refused on "never observed" or "stale". That was replaced by the tracked lower bound (the `eth_getBalance` thread on #11086), which removed the cache, the staleness limit and the age metric along with it. The description now says what the code does.

There is deliberately no balance-age metric in the new design: there is nothing to go stale, since the bound is derived from the minter's own events and cannot be older than the last funding it recorded. What can go wrong is a funding that never finalizes, which is what `cketh_minter_sweeper_in_flight_funding_age_seconds` is for.

rs/ethereum/cketh/minter/templates/dashboard.html
```diff
@@ -181,6 +181,40 @@ <h3>Block sync</h3>
...
+            <h3 id="sweeper-funding">Sweeper fee funding</h3>
+            <table>
+                <tbody>
+                    <tr id="sweeper-address">
+                        <th>Sweeper address</th>
+                        <td>{% if sweeper_funding.sweeper_address.is_some() %}{% call etherscan_address_link(sweeper_funding.sweeper_address.as_ref().unwrap()) %}{% endcall %}{% else %}unknown (ECDSA public key not yet available){% endif %}</td>
+                    </tr>
+                    <tr id="sweeper-prepaid-gas">
+                        <th>Prepaid gas (lower bound)</th>
+                        <td>{{ sweeper_funding.prepaid_gas_lower_bound }} Wei</td>
```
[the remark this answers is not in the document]
mbjorkqvist: Same stale description, now corrected — there is no observation to show a timestamp for.

On the conflation: with a tracked bound, `0 Wei` is not "never looked". It means no funding has been recorded as finalized yet, which is a fact about the minter's own history rather than a missing reading, and the row above it shows the low-water mark and target it is compared against. The distinction that mattered in the old design — "no gas" versus "never read the chain" — cannot arise when nothing reads the chain.

rs/ethereum/cketh/minter/src/dashboard/tests.rs
```diff
@@ -26,6 +26,61 @@ use ic_ethereum_types::Address;
...
+    state
+        .sweeper_funding
+        .record_finalized_funding(Wei::from(900_000_u64), Wei::from(50_000_u64));
+
+    let dashboard = DashboardTemplate::from_state(&state, DashboardPaginationParameters::default());
+
+    DashboardAssert::assert_that(dashboard)
+        .has_sweeper_cketh_burned("1_000_000 Wei")
+        .has_sweeper_eth_spent("950_000 Wei")
+        // 1_000_000 burned - 950_000 spent: the credit that offsets the next funding.
```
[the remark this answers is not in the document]
mbjorkqvist: Agreed, and fixed. That comment survived from a design where the surplus was discounted from the next burn; it now says the surplus stays as backing. The metric help text and the accounting docs already said so, which is what made the comment stand out.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -10,6 +10,26 @@ const FEE: u128 = 1_000_000_000_000_000; // 0.001 ETH
...
+        assert_eq!(
+            accounting.failed_fundings(),
+            0,
+            "a funding that delivered its ETH is not a failure"
+        );
+
+        // What the caller does on a failure receipt: no ETH transferred, and the count bumped.
+        accounting.record_burn(Wei::new(BURN));
+        accounting.record_failed_funding();
+        accounting.record_finalized_funding(Wei::ZERO, Wei::new(FEE));
```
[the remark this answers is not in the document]
mbjorkqvist: A real gap, thank you — the counter was only ever incremented by a test calling `record_failed_funding` directly, so the wiring in `record_finalized_transaction` could have been deleted with the suite still green.

Covered now by `should_record_a_funding_whose_transaction_failed_on_chain`, which drives the events rather than the accounting: it accepts a funding, creates and signs the transaction, finalizes it with a failure receipt, and asserts the counter went 0 to 1, that the bound stays at zero because nothing was delivered, that the gas the failed transaction still paid is recorded as spent, and that the burn stays put. The success paths in the same module assert the counter stays at zero.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -1259,6 +1259,59 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+                     zero: funding is a bare transfer to an address derived from the minter's own \
+                     key, which has no code to revert in.",
+                )?;
+                w.encode_gauge(
+                    "cketh_minter_sweeper_funding_burned_not_yet_spent",
+                    s.sweeper_funding.burned_not_yet_spent().as_f64(),
+                    "ckETH burned for sweeping but not yet spent, i.e. how far burn runs ahead of \
+                     spend: the burn of a funding still in flight, plus the fees earlier fundings \
+                     provisioned but never paid. Nothing draws it down -- the surplus stays as \
+                     ckETH backing -- so between fundings it only ratchets up.",
```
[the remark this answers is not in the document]
mbjorkqvist: Right — a failed funding leaves its whole undelivered amount here, not merely an unpaid fee, and that is the large jump an operator would see beside a failure. The help text says so now: "A funding that failed on chain leaves its whole undelivered amount here, so this jumps alongside a failure."

Worth noting what the gauge does *not* do, since it reads like it might: nothing ever draws that part down. A failed funding is never reimbursed, so its undelivered burn stays as ckETH backing for good, and only the in-flight component of the gauge falls again when that funding finalizes.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -104,9 +111,28 @@ impl SweeperFundingAccounting {
...
+    /// `transaction_fee` what it paid for gas, which it did either way.
+    ///
+    /// The outcome decides both the count and whether the transfer landed, so it is one call.
+    /// A failure is counted rather than trapped: the burn simply goes unspent, and the count is the
+    /// only signal that something believed impossible has happened.
+    pub fn record_finalized_funding(
+        &mut self,
+        status: &TransactionStatus,
+        amount: Wei,
+        transaction_fee: Wei,
```
gregorydemay: nit: I think it would be simpler to just past a reference to a `FinalizedEip1559Transaction` and let it pick what it needs. Now it's for example too easy to mix the order of the parameters `amount` and `transaction_fee` (same type, next to each other)

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -184,15 +222,19 @@ impl SweeperFundingAccounting {
...
+    /// exactly when those surfaces must stay up. The state transition traps instead.
     pub fn burned_not_yet_spent(&self) -> Wei {
-        self.cumulative_burned
-            .checked_sub(self.cumulative_spent())
-            .expect(
-                "BUG: more ETH spent on sweeping than ckETH burned for it, \
-                 meaning ckETH is under-backed",
-            )
+        self.checked_burned_not_yet_spent().unwrap_or(Wei::ZERO)
+    }
```
gregorydemay: nit: I think this method is a bit dangerous because when would it be ok for `self.checked_burned_not_yet_spent()` to return `None`? I think never. I understand that we may want to avoid panics for tracking purposes (metrics/dashboard), but maybe it would be better to just inline the method (2 productive call sites)

rs/ethereum/cketh/minter/src/sweeper/tests.rs
```diff
@@ -238,6 +249,44 @@ mod concurrent_fundings {
...
+        assert_eq!(
+            state.sweeper_funding.failed_fundings(),
+            0,
+            "fundings that delivered must not count as failures"
+        );
+    }
+
+    /// The failure branch of finalization, which the accounting test cannot reach on its own: the
+    /// counter, the undelivered transfer and the gas a failed transaction still pays are all wired
+    /// up here, in the state transition.
```
gregorydemay: nit:

```suggestion
```

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -487,6 +487,7 @@ impl State {
         self.withdrawal_transactions
             .record_finalized_transaction(*withdrawal_id, receipt.clone());
         self.update_balance_upon_withdrawal(withdrawal_id, receipt);
+        self.update_sweeper_funding_upon_withdrawal(withdrawal_id, receipt);
```
gregorydemay: I don't really understand the purpose of this change, it seems too easy to forget to call this method, whereas in `update_balance_upon_withdrawal` as it was the code forces you (we should have a match on `WithdrawalRequest`) to handle all variants.

gregorydemay (commented):
Generally, LGTM, just a couple of nits

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Done — it takes `&FinalizedEip1559Transaction` now and reads the status, the amount and the effective fee off it, so there is nothing left to transpose. It also lines up with `record_finalized_sweep` just below, which already worked this way.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
mbjorkqvist: Agreed — `None` is never a legitimate state, it is the invariant being violated. Inlined at both call sites, so each states its own policy where you can see it: the transition traps, and the getter saturates because the metrics and the dashboard read it from a query, which is exactly when those surfaces need to stay up.

rs/ethereum/cketh/minter/src/sweeper/tests.rs
mbjorkqvist: Removed.

rs/ethereum/cketh/minter/src/state.rs
mbjorkqvist: You are right — it is back inside `update_balance_upon_withdrawal`, behind a match on `WithdrawalRequest` so a new variant has to be dealt with.

Two things I had missed: the extracted method looked the request and the finalized transaction up again when the caller already held both, and being a separate call it could simply be forgotten. The only thing it bought was that `update_balance_upon_withdrawal`'s name says nothing about fundings, which does not weigh much against that.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -1322,10 +1322,62 @@ fn http_request(req: HttpRequest) -> HttpResponse {
...
+                    s.sweeper_funding.burned_not_yet_spent().as_f64(),
+                    "ckETH burned for sweeping and not yet spent, i.e. how far burn runs ahead of \
+                     spend. A funding that failed on chain leaves its whole undelivered amount \
+                     here, so this jumps alongside a failure.",
+                )?;
                 w.encode_gauge(
-                    "cketh_minter_stored_attestations",
-                    s.automatic_deposits.attestations_len() as f64,
-                    "Number of deposit address attestations the minter has signed and stored.",
+                    "cketh_minter_sweeper_funding_gas_balance",
```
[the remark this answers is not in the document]
mbjorkqvist: Correct, and thank you — restored in e5f72c9. The gauge came in with #11319 and reorganising this block dropped it; nothing in the suite asserts it, so everything stayed green while the attestations map grew unexported.

That gap is closed too: 17d5d32c adds a test asserting both `cketh_minter_stored_attestations` and `cketh_minter_stored_authorizations`. I checked it bites by deleting the gauge again and watching it fail. f7345a96 then moves the sweeper-funding block below the pair, so the two sit together as they did and this branch only ever inserts metrics into the existing order.

gregorydemay (approved):

PR #11097, opened 2026-08-11
test(cketh): adversarial end-to-end coverage of sweeper fee funding

mbjorkqvist:
Part of [DEFI-2933](https://dfinity.atlassian.net/browse/DEFI-2933) (sweeper fee funding), the last PR of the stack. Every PR below it has merged, so this one targets `master`.

## Why

The stack's existing live tests prove funding works. This one proves it fails safe. Each test drives a distinct way funding can go wrong through the real pipeline — real ckETH ledger, real EVM RPC canister, local anvil, threshold-ECDSA signature — and asserts the invariant the feature exists to protect: ETH never moves ahead of the ckETH burned to pay for it.

## What

Three adversarial scenarios:

- **An empty fee account.** The burn fails, so no ETH may move. Held across several withdrawal-timer ticks, since "the minter must not do this" can only be established as a bounded negative — and the test first waits for the log line proving the check ran and declined, so it cannot pass because nothing happened.
- **A sweeper already above the low-water mark.** No funding is due, so nothing is burned. Arranged by letting a real funding land and finalize, which is the only thing that moves the bound the decision reads — putting ETH at the address behind the minter's back deliberately would not.
- **A funding transaction that fails on chain.** The destination is given code, so a bare transfer runs out of gas: all 21'000 gas it carries is intrinsic, leaving nothing to execute with. The ckETH stays burned and is *not* reimbursed: no ETH reaches the sweeper, the failed transaction still pays its gas, and the burn minus that gas stays as backing — which is what keeps burned ahead of spent. The test waits for finalization, pins spend against the burn numerically, and then buys a tick longer than the reimbursement interval so that "never reimbursed" is not just true at the moment of finalization.

No live fee-spike test: that ceiling is pinned exactly by unit tests, and reproducing it here would mean driving anvil's base fee up over several ticks for little extra signal.

## What it found

Waiting for finalization — rather than asserting mid-flight, as the earlier live test did — surfaced a production defect that would trap the withdrawal timer permanently. The fix shipped in #11086, where it belongs; this PR is what made it visible.

## Harness work

These tests need to arrange a chain and then read the minter's own view of it, so the harness gains a fixture that leaves the fee account empty, wrappers over anvil's `setCode`/`getCode`, a reader for the minter's `/metrics` endpoint, a wait on a log line, and a wait for a funding to finalize whatever its outcome — the one already there insists on success, which is the opposite of what the failure test needs.

They join the live tests already in the `deposit_from_cex` target, which runs one test at a time and is already sized `large`; the three added here cost roughly a minute each, and the whole target now takes about four minutes.

## Stack

| # | PR | Status |
|---|----|--------|
| — | #11083 — Burn-first accounting for sweeper fee funding | merged |
| — | #11086 — Sweeper fee-funding task, with an end-to-end test | merged |
| — | #11094 — Sweeper funding observability | merged |
| 1 | Adversarial end-to-end coverage of sweeper fee funding | **this PR** |


rs/ethereum/cketh/minter/tests/sweeper_funding_hardening.rs
```diff
@@ -0,0 +1,198 @@
...
+    // The next scheduled check is a whole interval away, so re-arm the timers: from here the
+    // funding proceeds, and the point is what happens when its transaction fails.
+    setup.upgrade_minter();
+
+    let burned = await_burn(&setup, supply_before, Duration::from_secs(180));
+    assert!(burned > 0, "funding must burn ckETH up front");
+    // Read while the funding is still in flight: the dashboard clears the row once it finalizes.
+    let burn_index = setup
+        .in_flight_funding_burn_index()
+        .expect("the funding must be in flight once its burn has happened");
```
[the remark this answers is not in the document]
mbjorkqvist: Correct, and well spotted — fixed in 2096a99c.

The ordering is as you describe: `fund_sweeper_address` awaits the ledger burn and records `AcceptedSweeperFundingRequest` only after that call returns, while `await_burn` watches the ledger's total supply through an external query. So there is a real window in which the supply has dropped and the dashboard still shows no in-flight funding, and reading the index there would panic on an empty row.

It now polls, with the deadline as a named constant because it has to sit between two bounds rather than merely being long enough: generous for an inter-canister hop, but well under the roughly six minutes before the transaction finalizes and clears that row again. Polling indefinitely would have traded this race for the opposite one.

rs/ethereum/cketh/test_utils/src/sweeper_funding.rs
```diff
@@ -210,32 +211,172 @@ impl SweeperFundingSetup {
...
+                self.minter_logs().join("\n")
+            );
+            // Keeps the instance alive, and its timers with it.
+            let _ = self.cketh_total_supply();
+            std::thread::sleep(Duration::from_secs(5));
+        }
+    }
+
+    /// The minter's tECDSA-derived main address, which also proves key derivation works here.
+    /// Waits until the in-flight funding row clears, i.e. its transaction has finalized, mining
```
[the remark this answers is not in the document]
mbjorkqvist: Removed in 2096a99c. It was copied from the main-address accessor and described nothing this method does.

rs/ethereum/cketh/minter/tests/sweeper_funding_hardening.rs
```diff
@@ -0,0 +1,224 @@
...
+    setup.set_eth_balance(&sweeper, ABOVE_LOW_WATER_MARK);
+    // The minter reads at `finalized`, which trails `latest` by two blocks.
+    setup.mine(3);
+    setup.mint_cketh(setup.fee_account(), FEE_ACCOUNT_BALANCE);
+
+    // Captured before the timers are re-armed, not after: the post-upgrade check runs on a
+    // zero-delay timer, so a minter that wrongly funded could burn before these queries returned
+    // and both assertions below would then compare against an already-debited state — passing
+    // precisely when the behaviour they reject had happened.
+    let supply_before = setup.cketh_total_supply();
```
mbjorkqvist: Replying to the suppressed comment about this baseline being captured after `upgrade_minter()` (no comment node to thread onto, hence a fresh one here).

Right, and a fair catch — fixed in 532989e3.

Worth being precise about the failure mode, because it is not a flake in the current code: with a correct minter the sweeper is above the low-water mark, no funding is due, and nothing is burned, so the ordering never bites. It matters for a *broken* minter — if the post-upgrade check wrongly funded, its burn could land before those two queries returned, and then both "unchanged" assertions would compare against an already-debited state. The test would go green at precisely the moment it was supposed to fail.

The baselines are now read before `upgrade_minter()`, which is the one point where nothing can yet have happened: the fee account is funded, the sweeper is arranged, and the next scheduled check is a whole interval away. The revert test below already did it in that order — I fixed it there when addressing the in-flight race and did not carry the reasoning back up.

I left the first test as it is: its fee account is empty, so no burn can succeed at all, and its log assertion proves the task ran and declined *at the burn* rather than never running.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -331,6 +332,35 @@ rust_test(
     ],
 )
 
+rust_test(
+    name = "sweeper_funding_hardening",
+    # Three live scenarios, two of which watch across several withdrawal-timer ticks to show the
+    # minter does *not* act. Those ticks are bought by advancing the instance's clock, so the
+    # minutes of minter time they cover cost seconds of wall clock.
+    size = "medium",
```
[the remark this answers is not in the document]
mbjorkqvist: Correct, and the description was the stale half — the tag went away when the test stopped waiting out the withdrawal timer and started buying ticks with `advance_time`. The whole target now runs in about 50 seconds, so running it in ordinary CI is intended, and the description says so instead of claiming `long_test`.

rs/ethereum/cketh/minter/tests/sweeper_funding_hardening.rs
```diff
@@ -0,0 +1,250 @@
...
+    assert_ne!(
+        prepaid, "0 Wei",
+        "the bound must carry the funding that landed, or the decision above declined for the \
+         wrong reason"
+    );
+}
+
+/// A funding transaction that fails on chain is never reimbursed. No ETH reaches the sweeper — only
+/// the gas the failed transaction still pays leaves the main address — and the ckETH stays burned,
+/// so the burn minus that gas ends up as extra backing.
```
mbjorkqvist: Replying to the suppressed comment on this doc comment, which has no node to thread onto.

You were right and it is fixed: a failed transaction does move ETH out of the main address — its gas. The comment now says no ETH reaches the *sweeper*, that only the gas leaves, and that the burn minus that gas is what stays as backing. The same overstatement was in the funding task's log line and in #11086's description; both are corrected there.

rs/ethereum/cketh/minter/tests/sweeper_funding_hardening.rs
```diff
@@ -0,0 +1,250 @@
...
+        "the transfer failed, so no ETH may have reached the sweeper"
+    );
+    assert_eq!(
+        setup.cketh_total_supply(),
+        supply_before - burned,
+        "a failed funding must NOT be reimbursed — the supply must stay reduced"
+    );
+    // Read as numbers rather than asserted non-zero: "some burn remains" would also hold if
+    // finalization had recorded no gas at all, which is the other half of what this test claims.
+    let spent = wei_row(&setup, "sweeper-eth-spent");
```
mbjorkqvist: Two suppressed comments landed on the assertions here, both fair, both now acted on.

The first: `assert_ne!(surplus, "0 Wei")` also passes if finalization recorded no gas at all, which is half of what this test claims. The rows are read as numbers now — spend must be non-zero, and the surplus must equal the burn minus that spend, so a zero-fee regression fails here.

The second: "never reimbursed" was established by looking once at finalization, while reimbursement runs on its own three-minute timer. The test now buys a tick longer than that interval and asserts the supply is still reduced, so a funding wrongly queued for reimbursement is caught rather than minted after the test ends. Costs about four seconds.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -282,6 +292,202 @@ fn should_fund_the_sweeper_address_by_burning_cketh_from_the_fee_account() {
...
+    setup.upgrade_minter();
+    let sweeper = setup.await_sweeper_address();
+    setup.await_minter_log("[fund_sweeper]: SKIPPING: failed to burn");
+
+    // All 21'000 gas of a bare transfer is intrinsic, so the callee runs with none and the first
+    // opcode that costs anything halts it. That is what fails the transaction here — `PUSH1` costs
+    // 3 — and it is why the bytecode is not arbitrary: code beginning with the zero-cost `STOP`
+    // would run to completion and succeed. Spelled PUSH1 0, PUSH1 0, REVERT, so the call would also
+    // fail on its own terms if it were ever reached with gas to spare.
+    setup.set_code(&sweeper, &[0x60, 0x00, 0x60, 0x00, 0xfd]);
```
mbjorkqvist, line 414: Replying to the suppressed comment about "any non-empty code"; no comment node to thread onto, hence a fresh one here.

Right, and the overstatement was mine from an hour earlier. `STOP` is a zero-cost opcode, so runtime code beginning with `0x00` runs to completion on the nothing a bare transfer leaves the callee, and the transfer succeeds.

Fixed in 334a51e9. The comment now says what actually fails the transaction — the callee runs with no gas, so the first opcode that costs anything halts it, and `PUSH1` costs 3 — and therefore why the bytecode is not arbitrary.

rs/ethereum/cketh/test_utils/src/live.rs
```diff
@@ -758,6 +809,149 @@ impl<S: AsRef<CkEthSetup>> LiveSetup<S> {
...
+            self.env().advance_time(WITHDRAWAL_TICK);
+            self.settle(&mut |_| false);
+        }
+    }
+
+    /// Places `code` at `address`, so that a plain value transfer to it no longer succeeds: all
+    /// 21'000 gas of a bare transfer is intrinsic, leaving the callee none. Which makes this a
+    /// guarantee only for `code` whose first opcode costs something — code beginning with the
+    /// zero-cost `STOP` runs to completion on no gas at all and the transfer still succeeds.
+    pub fn set_code(&self, address: &Address, code: &[u8]) {
```
mbjorkqvist, line 877: Replying to the suppressed comment on this doc comment; no node to thread onto.

Agreed. The wrapper cannot promise that placing code makes a transfer fail, only that it does for code whose first opcode costs something. Fixed in 334a51e9: the doc states that condition and names `STOP` as the case that slips through, so nobody reaching for this reads it as a guarantee.

rs/ethereum/cketh/test_utils/src/live.rs
```diff
@@ -737,6 +776,123 @@ impl<S: AsRef<CkEthSetup>> LiveSetup<S> {
...
+    /// zero both when nothing is outstanding and when a funding was accepted less than a second of
+    /// instance time ago, which is every moment between the burn and the next tick.
+    pub fn await_funding_finalized(&self, max_ticks: u32) {
+        let spent_before = self.metric_value(SWEEPER_ETH_SPENT);
+        self.drive_until(
+            max_ticks,
+            |_| "the funding had not finalized".to_string(),
+            |setup| setup.metric_value(SWEEPER_ETH_SPENT) > spent_before,
+        );
+    }
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in a1ff9038, and the shape you suggested is what it uses: the helper takes the `ledger_burn_index` off the `AcceptedSweeperFundingRequest` and waits for the `FinalizedTransaction` carrying it as `withdrawal_id`.

The mechanism is as you describe — `process_retrieve_eth_requests` sends and finalizes in one pass (`withdraw.rs:193-194`), so nothing structural separates them. What stopped the baseline from actually being poisoned is the block tag: `await_eth_received` observes anvil's `latest`, while finalization gates on `finalized_transaction_count`, which asks for `BlockTag::Finalized` — two blocks behind under one-slot-per-epoch. At the first poll where the sweeper's balance goes non-zero the transfer's block is not finalized yet, and acting on it takes a further run of the withdrawal timer, which no settle window can produce because instance time does not move inside one.

Worth changing anyway, and not only for that margin: a baseline delta was the wrong shape regardless, and the event the counter stood in for is both more specific and the thing the test actually means. The old predicate would also have been satisfied by any transaction finalizing; this one is the funding or nothing.

PR #11101, opened 2026-08-11
fix: keep the aggregate canister log within its byte capacity

mraszyk:
`CanisterLog::append_delta_log` made free space in the aggregate log for the incoming delta log, but nothing bounded the delta itself, so the aggregate could end up holding more than its own byte capacity.

Trim the delta to the aggregate's capacity before appending: drop the delta's oldest records and, if the newest one still does not fit, truncate its content. The trimming happens before the existing index gap check, so that records dropped this way clear the aggregate as well and the record indices stay contiguous; the new next index is computed beforehand so that it advances regardless of what the trimming drops.

`Records::append` now documents and debug-asserts that precondition.

### DIFF
--- a/rs/types/types/src/canister_log.rs
+++ b/rs/types/types/src/canister_log.rs
@@ -115,11 +115,41 @@ impl Records {
     }
 
     /// Appends all records from `other` to `self`, making sure the size limit is respected.
+    ///
+    /// The caller must have trimmed `other` to `self.byte_capacity` beforehand (see
+    /// `trim_to_capacity`), otherwise `self` would exceed its own capacity: making
+    /// free space in `self` cannot compensate for an oversized `other`.
     fn append(&mut self, other: &mut Self) {
+        debug_assert!(other.bytes_used <= self.byte_capacity);
         self.make_free_space_within_limit(other.bytes_used);
         self.records.append(&mut other.records);
         self.bytes_used += other.bytes_used;
         other.clear();
+    }
+
+    /// Drops the oldest records so that the used bytes fit within `byte_capacity`.
+    ///
+    /// If the newest record alone does not fit, its content is truncated; should even
+    /// an empty record not fit (i.e. `byte_capacity` is below the fixed record size),
+    /// the buffer is left empty, just like `Records::from` would.
+    fn trim_to_capacity(&mut self, byte_capacity: usize) {
+        // Drop the oldest records, keeping the newest ones that fit.
+        while self.bytes_used > byte_capacity && self.records.len() > 1 {
+            self.pop_front();
+        }
+        // The single remaining record may still be too big on its own.
+        if self.bytes_used > byte_capacity
+            && let Some(record) = self.records.pop_front()
+        {
+            let record = truncate_content(byte_capacity, record);
+            let record_size = record.data_size();
+            if record_size <= byte_capacity {
+                self.records.push_front(record);
+                self.bytes_used = record_size;
+            } else {
+                self.bytes_used = 0;
+            }
+        }
     }
 
     /// Removes old records to make enough free space for new data within the limit.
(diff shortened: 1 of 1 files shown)


rs/types/types/src/canister_log.rs
```diff
@@ -115,13 +115,43 @@ impl Records {
     }
 
     /// Appends all records from `other` to `self`, making sure the size limit is respected.
+    ///
+    /// The caller must have trimmed `other` to `self.byte_capacity` beforehand (see
+    /// `trim_to_capacity`), otherwise `self` would exceed its own capacity: making
+    /// free space in `self` cannot compensate for an oversized `other`.
     fn append(&mut self, other: &mut Self) {
+        debug_assert!(other.bytes_used <= self.byte_capacity);
```
alin-at-dfinity:
```suggestion
    fn append(&mut self, other: &mut Self) {
        // The caller must have trimmed `other` to `self.byte_capacity` beforehand (see
        // `trim_to_capacity`).
        debug_assert!(other.bytes_used <= self.byte_capacity);
```

rs/types/types/src/canister_log.rs
```diff
@@ -260,20 +290,36 @@ impl CanisterLog {
...
-        // aggregate records to maintain index continuity.
-        if let Some(first) = delta_log.records.get().front()
-            && first.idx > self.next_idx
-        {
-            self.records.clear();
+        // Assume records sorted chronologically (with increasing idx): the new next
+        // index is the one following the delta's last record. Read here and not after
+        // the trimming below, as that may drop every single delta record (if the
+        // capacity is too small to hold even one), while the next index has to
+        // advance past dropped records just the same.
```
alin-at-dfinity:
```suggestion
        // Assume records sorted chronologically (with increasing idx): the new next
        // index is the one following the delta's last record. Read here and not after
        // the trimming below, as that may drop some or even all delta records.
```

rs/types/types/src/canister_log.rs
```diff
@@ -260,20 +290,36 @@ impl CanisterLog {
...
+            .records
+            .get()
+            .back()
+            .map_or(self.next_idx, |last| last.idx + 1);
+
+        // The delta log may have a bigger capacity than this (aggregate) log, in which
+        // case the delta records that don't fit here have to be dropped, oldest first;
+        // otherwise the aggregate would end up exceeding its own capacity. This must
+        // happen before the gap check below, so that the records dropped here are
+        // accounted for as a gap as well.
```
alin-at-dfinity, line 304:
```suggestion
        // The delta log may have a bigger capacity than this (aggregate) log, in which
        // case the delta records that don't fit here have to be dropped, oldest first.
```

rs/types/types/src/canister_log.rs
```diff
@@ -260,20 +290,36 @@ impl CanisterLog {
...
+        delta_log.records.trim_to_capacity(self.byte_capacity());
+
+        match delta_log.records.get().front() {
+            // The delta continues where the aggregate left off, keep both.
+            Some(first) if first.idx <= self.next_idx => {}
+            // Either records were evicted (by the delta itself or by the trimming
+            // above), leaving a gap between the aggregate's next expected index and
+            // the delta's first record, or the trimming dropped every single delta
+            // record — the delta was not empty here, see the early return above.
+            // Either way, drop the aggregate records to maintain index continuity.
```
alin-at-dfinity:
```suggestion
            // Records were evicted (by the delta itself or by the trimming
            // above), leaving a gap between the aggregate's next expected index and
            // the delta's first record. Drop the aggregate records to maintain
            // index continuity.
```

alin-at-dfinity (approved):
Only some shorter comment suggestions, feel free to take them or leave them.

rs/types/types/src/canister_log.rs
mraszyk: I'd keep this so that we emphasize that
> // [...] This must
> // happen before the gap check below, so that the records dropped here are
> // accounted for as a gap as well.

mraszyk: In any case, the legacy canister log buffer (and thus also this code) is supposed to be dropped soon anyway.

PR #11102, opened 2026-08-11
refactor: decode `Cycles` from little-endian bytes via `TryFrom`

mraszyk:
`Cycles` was decoded from its little-endian byte representation through an infallible `From<&Vec<u8>>` impl, which left it with no way to report a length mismatch to its caller. Replace it with `TryFrom<&Vec<u8>>`.

The two protobuf conversions built on it, `state.queues.v1.Cycles` and `canister_state_bits.v1.CyclesAccount`, become `TryFrom` impls returning `ProxyDecodeError::ValueOutOfRange`, so a length mismatch is now reported like any other proto decoding error. This matches how `NominalCycles` and `CompoundCycles` already decode in the same crate.

The `try_from_option_field` call sites (`Request::cycles_payment`, `Response::cycles_refund`, `CallContext::available_cycles`, `CanisterStateBits::cycles_balance`, `Refund::amount`) need no change: they already required a `TryFrom` whose error converts into `ProxyDecodeError` and were satisfied through the blanket impl. The callers that used the infallible impls directly now propagate the error instead: `CompoundCycles::real`, `Callback::cycles_sent`, the three `RefundStatus` fields, and four `CanisterStateBits` fields in `state_layout`.

The encoding is unchanged, so this only affects the decoding failure path.

### DIFF
--- a/rs/types/cycles/src/cycles.rs
+++ b/rs/types/cycles/src/cycles.rs
@@ -1,9 +1,10 @@
 use candid::{CandidType, Nat};
 use ic_heap_bytes::DeterministicHeapBytes;
+use ic_protobuf::proxy::ProxyDecodeError;
 use ic_protobuf::state::canister_state_bits::v1::CyclesAccount as pbCyclesAccount;
 use ic_protobuf::state::queues::v1::Cycles as PbCycles;
 use serde::{Deserialize, Serialize};
-use std::convert::TryInto;
+use std::array::TryFromSliceError;
 use std::iter::Sum;
 use std::{
     fmt,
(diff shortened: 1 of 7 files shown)


rs/types/cycles/src/cycles.rs
```diff
@@ -92,9 +93,13 @@ impl From<u64> for Cycles {
     }
 }
 
-impl From<&Vec<u8>> for Cycles {
-    fn from(bytes: &Vec<u8>) -> Self {
-        Self::new(u128::from_le_bytes(bytes.as_slice().try_into().unwrap()))
+/// Decodes `Cycles` from their little-endian representation, as produced by
+/// `From<Cycles> for Vec<u8>`. Fails if `bytes` is not exactly 16 bytes long.
+impl TryFrom<&[u8]> for Cycles {
```
alin-at-dfinity: Just wondering: could we instead implement something like `TryFrom<AsRef<[u8]>>`? Then we wouldn't have to explicitly call `as_slice()` in all tests.

alin-at-dfinity (approved):

rs/types/cycles/src/cycles.rs
mraszyk: Done: https://github.com/dfinity/ic/pull/11102/commits/dbb9ec621cc3024ea6e45c879df2885474a9ff16

PR #11104, opened 2026-08-11
fix: CanisterDidNotReply for failed calls in composite queries

mraszyk:
Requests to the management canister made by a composite query are executed by the query handler against the state of the own subnet, and thus they are not routed based on the method and the payload. However, that short-circuit was nested inside the `Ok(())` arm of `validate_sender_canister_version`, which fails with `CanisterMethodNotFound` for a method name that is not a management canister method at all. Such a request was hence rejected via `reject_subnet_output_request`, which pushes the reject response onto the canister's input queue. Input queues are never inducted while evaluating a query call graph, so the reject response was lost and the composite query failed with `CanisterDidNotReply` instead.

Skip validation together with routing for composite queries so that the request reaches the query handler, which rejects it with the same error as a query sent by an end user to the management canister ("Query method <name> not found."). This is safe: all management canister query methods have no sender canister version in their payload and every method that does carry one is rejected as not being a query method anyway.

Since `is_composite_query` also covers the reply and reject callbacks of a composite query, this fixes calls made from those callbacks as well; a regression test is added for that case.

Finally, calls with a subnet ID as the callee are now rejected with "Canister <subnet_id> not found" in composite queries instead of failing with `CanisterDidNotReply`.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
@@ -456,21 +456,7 @@ impl SystemStateModifications {
         let subnet_ids: BTreeSet<PrincipalId> =
             network_topology.subnets().keys().map(|s| s.get()).collect();
         for mut msg in self.requests {
-            if msg.receiver == IC_00 && is_composite_query {
-                // Requests to the management canister made by a composite query
-                // (including from its callbacks) are neither validated nor routed
-                // based on the method and the payload: they are executed by the query
-                // handler against the state of the own subnet (or rejected by it if
-                // the method cannot be executed in the non-replicated mode). Note
-                // that composite queries are always executed in the non-replicated
-                // mode.
-                //
-                // In particular, such a request must not be rejected here: the reject
-                // response would be pushed onto the canister's input queue, which is
-                // never inducted while evaluating a query call graph, and hence the
-                // composite query would fail with `CanisterDidNotReply`.
-                Self::push_message(system_state, time, msg, logger)?;
-            } else if msg.receiver == IC_00 {
+            if !is_composite_query && msg.receiver == IC_00 {
                 match Self::validate_sender_canister_version(&msg, system_state.canister_version())
                 {
                     Ok(()) => {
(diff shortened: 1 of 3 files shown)


rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -456,32 +456,36 @@ impl SystemStateModifications {
...
         for mut msg in self.requests {
-            if msg.receiver == IC_00 {
+            if msg.receiver == IC_00 && is_composite_query {
+                // Requests to the management canister made by a composite query
+                // (including from its callbacks) are neither validated nor routed
+                // based on the method and the payload: they are executed by the query
+                // handler against the state of the own subnet (or rejected by it if
+                // the method cannot be executed in the non-replicated mode). Note
+                // that composite queries are always executed in the non-replicated
+                // mode.
```
alin-at-dfinity: Nit:

```suggestion
                // the method cannot be executed in non-replicated mode). Note
                // that composite queries are always executed in non-replicated
                // mode.
```

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
```diff
@@ -507,6 +511,15 @@ impl SystemStateModifications {
                         )?;
                     }
                 }
+            } else if subnet_ids.contains(&msg.receiver.get()) && is_composite_query {
```
alin-at-dfinity: I would merge this branch with the first one. And pull `subnet_ids.contains(&msg.receiver.get())` into a variable, since it's now potentially evaluated on a couple of branches. Something like:

```rust
let receiver_is_subnet = subnet_ids.contains(&msg.receiver.get());
if is_composite_query && (msg.receiver == IC_00 || receiver_is_subnet) {
    // ...
    Self::push_message(...)
} else if msg.receiver == IC_00 {
    ...
} else if receiver_is_subnet {
    ...
} else {
    Self::push_message(...)
}
```

Or re-reading the above, you could even turn it around to:

```rust
if !is_composite_query && msg.receiver == IC_00 {
    ...
} else if !is_composite_query && subnet_ids.contains(&msg.receiver.get()) {
    ...
} else {
    // ... (composite query explanation) ...
    Self::push_message(...)
}
```

rs/execution_environment/src/query_handler/tests.rs
```diff
@@ -1586,19 +1593,155 @@ fn composite_query_call_to_management_canister_rejects_unknown_methods() {
...
+    composite_query_call_to_own_subnet_id_impl(false);
+}
+
+#[test]
+fn composite_query_call_to_own_subnet_id_on_nns_subnet() {
+    composite_query_call_to_own_subnet_id_impl(true);
+}
+
+#[test]
+fn composite_query_call_to_own_subnet_id_rejects_unknown_methods() {
```
alin-at-dfinity: Isn't this somewhat redundant? I.e. isn't it already rejected because it's aimed at at `own_subnet_id` before we even check the method name?

Feel free to leave it as is, though, if you're aiming for completeness.

alin-at-dfinity (approved):

rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs
mraszyk: Good point, done

rs/execution_environment/src/query_handler/tests.rs
mraszyk: Dropped the test after all.

PR #11110, opened 2026-08-12
test: Fix tests asserting on `ReplicatedStateMetrics`

alin-at-dfinity:
`ReplicatedStateMetrics` are observed asynchronously, from a worker thread triggered by `StateManager::commit_and_certify()`. A zero capacity channel between StateManager and the worker thread means that whenever the worker thread is busy (which is often true during excessively high load, such as when lots of tests run concurrently) some metric observations are skipped, so waiting for the worker thread to complete its current observation is not sufficient to ensure correctness.

Instead, make `WorkerThread::enqueue()` blocking, so no metric observations are ever skipped (particularly since the `state_manager_skipped_state_observations` metric shows that on busy subnets we skip about 5% of `ReplicatedStateMetrics` observations, which is not ideal). And make the `WorkerThread` channel size configurable, with `ReplicatedStateMetricsThread` using a size 1 channel, to make it less likely for the DSM main thread to block when enqueuing metrics observations in case the worker thread was not scheduled (which apparently happens in the replica under heavy load).

### DIFF
--- a/rs/state_manager/src/lib.rs
+++ b/rs/state_manager/src/lib.rs
@@ -199,7 +199,7 @@ pub struct StateManagerMetrics {
     fast_forward_height: IntGauge,
     no_state_clone_count: IntCounter,
     skip_optimization_missing_cert_count: IntCounter,
-    skipped_state_observations: IntCounter,
+    state_observation_blocked_duration: Histogram,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


schneiderstefan:
Wait, so now it is blocking even in production?

schneiderstefan (approved):

alin-at-dfinity:
> Wait, so now it is blocking even in production?

Yes, `WorkerThread::enqueue()` will block iff the previous workload (i.e. request to observe metrics) is still enqueued. We now have a size 1 channel between the DSM and the metrics thread, precisely in order to make it less likely.

But this is mostly an issue when CI is running tests, where you have dozens of processes (or more) and hundreds of threads (or more) running concurrently; and the DSM thread (or test main thread) isn't really doing much work or waiting for the next block for hundreds of milliseconds at a time; so it can happen that we execute a couple of "rounds" before the metrics thread is ever scheduled by the kernel.

I.e. we will end up actually blocking on CI (ensuring correctness); but we are exceedingly unlikely to ever block in production (although we'll have to wait for the metrics to actually tell us whether this is so; right now, wit the non-blocking implementation we are losing a meaningful number of metrics observations on some subnets, messing up some metrics rules / dashboards).

PR #11111, opened 2026-08-12
feat: Respect query limits from the registry

schneiderstefan:
A previous PR (https://github.com/dfinity/ic/pull/10882) made it possible to specify different query limits for a subnet via the registry.

This PR is the follow-up that actually respects these limits in the replica.


rs/execution_environment/src/query_handler.rs
```diff
@@ -246,9 +246,19 @@ impl InternalHttpQueryHandler {
...
+        // its own replica default when unset.
+        let resource_limits = state.get_ref().resource_limits();
+        let max_query_call_graph_instructions = resource_limits
+            .maximum_query_instructions_or(self.config.max_query_call_graph_instructions);
         let max_instructions_per_query = match max_instructions {
-            Some(max_ins) => max_ins.min(self.max_instructions_per_query),
-            None => self.max_instructions_per_query,
+            // A caller-provided limit (currently only the HTTP outcalls transform budget) is
+            // authoritative for that call.
+            Some(max_ins) => max_ins,
```
alin-at-dfinity: Clueless question: Are we OK with this potentially greatly exceeding the registry-configured limit?

Also, IMHO the abbreviation is more pain that it's worth:

```suggestion
            // A caller-provided limit (currently only the HTTP outcalls transform budget) is
            // authoritative for that call.
            Some(max_instructions) => max_instructions,
```

rs/execution_environment/src/query_handler/tests.rs
```diff
@@ -434,6 +435,112 @@ fn composite_query_callgraph_max_instructions_is_enforced() {
...
+        })
+        .build();
+
+    let mut canisters = vec![];
+    for _ in 0..NUM_CANISTERS {
+        canisters.push(test.universal_canister_with_cycles(CYCLES_BALANCE).unwrap());
+    }
+
+    // Generate call tree of depth 1: canister 0 calls into each of canisters 1..num_calls
+    // sequentially, staying under the depth limit but hitting the total-instruction limit.
```
alin-at-dfinity, line 494: Rant: I'm quite sure that the comment accurately describes what's going on, but I find the recursive implementation (and everything to do with the Universal Canister) incredibly hard to parse. Maybe that's just me.

rs/execution_environment/src/query_handler/tests.rs
```diff
@@ -434,6 +435,112 @@ fn composite_query_callgraph_max_instructions_is_enforced() {
...
+        Arc::new(test.state().clone()),
+        vec![],
+        /*certificate_delegation_metadata=*/ None,
+    );
+    assert_eq!(result, Ok(WasmResult::Reply(b"ok".to_vec())));
+}
+
+#[test]
+fn composite_query_callgraph_max_instructions_from_resource_limits_is_enforced() {
+    const NUM_CANISTERS: u64 = 20;
```
alin-at-dfinity: Nit: 6 or even 5 canisters should be sufficient.

alin-at-dfinity (approved):

daniel-wong-dfinity-org-twin (approved):

rs/execution_environment/src/query_handler.rs
schneiderstefan: I've done the renaming.

As for exceeding the limit, the main point of the new registry parameter is so that you can set it larger for engines where you are willing to take the negative consequences. So it's not really an issue if you can exceed tiny values in HTTP outcalls.

rs/execution_environment/src/query_handler/tests.rs
schneiderstefan: Done.

rs/execution_environment/src/query_handler/tests.rs
schneiderstefan: Yeah that's just because the universal canister is this stack based programming model. You can do anything, but it's not necessarily in a way that corresponds to your normal mental model.

PR #11113, opened 2026-08-12
feat: Async HTTP outcalls refunds

eichhorl:
## Background
The new pay-as-you-go pricing for HTTP outcalls consists of four phases:
1. Base cost. This fee is charged for every request upfront.
2. Per-replica cost. The remaining cycles after charging the base cost are split evenly between the participating replicas. Each replica consumes some of their allowance as the HTTP request is processed. In the end, the amount of consumed cycles is gossiped as part of the response share.
3. Consensus cost. This fee is charged for including the aggregated HTTP response as part of a block.
4. Asynchronous refunds. Shares that did not contribute to the initial response (i.e. because they were received too late) are delivered asynchronously, such that their unused cycles can be reported and refunded, as well.

In https://github.com/dfinity/ic/pull/10897, we implemented the infrastructure required for phases 3 and 4, and partly implemented phase 3: After a response for a "pay-as-you-go" context is delivered, the initial refund is applied, and the context is moved into a new collection `delivered_canister_http_contexts`, where it waits for further asynchronous receipts to be delivered. After two minutes since the creation of the original context, the "delivered context" times out, refunding the allowances of all replicas that haven't reported a receipt yet, in full.

## Proposed Changes
This PR implements the inclusion, validation and delivery of asynchronous refund receipts (phase 4):

### Client (pool manager):
- Shares whose context was moved to `delivered_canister_http_contexts` are no longer purged
- As before, we will *not* make *new* requests to the HTTP adapter for contexts that were already moved to `delivered_canister_http_contexts`. No work was done, so the replica's allowance will be refunded in full on timeout.
- If an *ongoing* HTTP adapter request finishes, we continue to create and gossip a share, as long as the context is still part of `delivered_canister_http_contexts`. This is to report the work that was done.
- We continue to accept and validate shares whose context was moved to `delivered_canister_http_contexts`

### Payload builder:
- We add a new field to the payload, holding the async receipts
- We will include async receipts for contexts that are part of `delivered_canister_http_request_contexts` and have not yet expired.
- We will only include an async receipt, if the issuing node has not refunded part of their allowance yet, i.e. as part of the initial response, or a previous async receipt. This is checked via the refund status of the context in the state, and all delivered async receipts in payloads above it.

## Future Work
For flexible and non-replicated outcalls, the gossiped receipts still contain the full response bodies (just like normal shares for these replication types). However, these responses will never be included into a block, so we should avoid wasting bandwidth by only transmitting the share (like we do for fully replicated outcalls).


rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -1103,6 +1165,53 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                .or_default()
+                .push(share);
+        }
+        for (callback_id, shares) in receipts_by_callback {
+            // Only an outcall that has already been responded to can be refunded asynchronously.
+            let context = delivered_http_contexts.get(&callback_id).ok_or(
+                CanisterHttpPayloadValidationError::InvalidArtifact(
+                    InvalidCanisterHttpPayloadReason::UnknownDeliveredCallbackId(callback_id),
+                ),
+            )?;
```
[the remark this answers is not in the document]
eichhorl: I think this is fine for simplicity

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -374,6 +385,52 @@ impl CanisterHttpPayloadBuilderImpl {
                     },
                 }
             }
+
+            // Collect the asynchronous receipts of the requests that have already
+            // been responded to.
+            for (callback_id, request) in delivered_canister_http_request_contexts {
+                if responses_included >= CANISTER_HTTP_MAX_RESPONSES_PER_BLOCK {
+                    break;
```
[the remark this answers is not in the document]
eichhorl: This is a general problem with the payload builder, which we can look at in a follow up.

rs/https_outcalls/consensus/src/payload_builder/parse.rs
```diff
@@ -101,33 +106,75 @@ pub(crate) fn payload_to_bytes(payload: CanisterHttpPayload, max_size: NumBytes)
...
 
     iterator_to_bytes(message_iterator, max_size)
 }
 
-pub(crate) fn parse_past_payload_ids(
+/// Relevant data of payloads between the certified height and the block being built.
+#[derive(Default)]
+pub struct PastPayloads {
+    /// The callback ids that have already been responded to.
+    pub delivered_ids: HashSet<CallbackId>,
```
pierugo-dfinity, line 127: I see this was already the case before, but would it be safer & more consistent to use a `BTreeSet` instead?

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -374,6 +385,52 @@ impl CanisterHttpPayloadBuilderImpl {
...
+                }
```
pierugo-dfinity, line 395: I think this could be removed as it is checked again later in the inner loop

rs/https_outcalls/consensus/src/payload_builder.rs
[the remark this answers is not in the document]
pierugo-dfinity: It does look like a simple check though

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -339,6 +339,23 @@ pub(crate) fn find_non_flexible_out_of_cycles(
...
+/// The receipts of the replicas that have contributed to an already responded to
+/// outcall but have not been refunded yet, at most one per replica.
+///
+/// `already_refunded` tells whether a replica's spend has already been accounted
+/// for, either by the response that was delivered or by an earlier asynchronous
+/// refund.
+pub(crate) fn find_async_receipts<'a>(
+    grouped_shares: &BTreeMap<CanisterHttpResponseMetadata, Vec<&'a CanisterHttpResponseShare>>,
+    committee: &BTreeSet<NodeId>,
+    already_refunded: impl Fn(&NodeId) -> bool,
```
pierugo-dfinity: Maybe
```rust
context.refund_status.refunding_nodes.contains(&signer)
    || refunded_nodes
        .get(&callback_id)
        .is_some_and(|nodes| nodes.contains(&signer))
```
could be in its own helper function to avoid this closure here and the deduplication in the validator?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -331,19 +332,23 @@ impl CanisterHttpPoolManagerImpl {
...
-        let active_contexts = &self
-            .latest_state()
-            .metadata
-            .subnet_call_context_manager
-            .canister_http_request_contexts;
+        let state = self.latest_state();
+        let subnet_call_context_manager = &state.metadata.subnet_call_context_manager;
+        let active_contexts = &subnet_call_context_manager.canister_http_request_contexts;
+        let delivered_contexts =
+            &subnet_call_context_manager.delivered_canister_http_request_contexts;
```
pierugo-dfinity, line 345: I don't think we differentiate between active and delivered in the rest of the function, so should just chain them in a single "known contexts"?

Oh, I guess you will differentiate in the future to gossip the responses only for active contexts and not for delivered ones, correct?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -418,10 +423,10 @@ impl CanisterHttpPoolManagerImpl {
...
 
         let state = self.latest_state();
-        let active_contexts = &state
-            .metadata
-            .subnet_call_context_manager
-            .canister_http_request_contexts;
+        let subnet_call_context_manager = &state.metadata.subnet_call_context_manager;
+        let active_contexts = &subnet_call_context_manager.canister_http_request_contexts;
+        let delivered_contexts =
+            &subnet_call_context_manager.delivered_canister_http_request_contexts;
```
pierugo-dfinity, line 438: Ditto, but here I'm not sure why we can't just merge them. But for consistency, I'm also happy with keeping them separate.

Here will you still invalidate NonReplicated/Flexible without a response, even when they are in the active contexts? Maybe the sender is just ahead of us and was already in the delivered contexts

Sorry, I am already thinking of whatever is your next PR lol

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -690,6 +705,25 @@ pub mod test {
...
+    ) -> ReplicatedState {
+        let mut replicated_state = ReplicatedState::new(subnet_test_id(0), SubnetType::System);
+        let contexts = &mut replicated_state.metadata.subnet_call_context_manager;
+        // Hand out callback ids up to the largest delivered one, so that shares for
+        // them are not mistaken for shares belonging to a future state.
+        if let Some((max_id, context)) = delivered.iter().next_back() {
+            for _ in 0..=max_id.get() {
+                contexts.push_context(SubnetCallContext::CanisterHttpRequest(context.clone()));
+            }
+        }
```
pierugo-dfinity, line 723: I'm not sure I understand what this is doing

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -3675,4 +3709,372 @@ pub mod test {
...
+    /// A share is signed even for an outcall that has already been responded to: the
+    /// work was done and paid for, so the receipt has to be published for the spend
+    /// to be settled asynchronously. It is gossiped exactly as it would have been
+    /// before the response was delivered.
+    #[test]
+    fn test_share_is_created_for_a_delivered_context() {
+        for replication in [
+            Replication::FullyReplicated,
+            Replication::NonReplicated(node_test_id(0)),
+        ] {
```
pierugo-dfinity, line 3954: Why not also test Flexible?

rs/https_outcalls/consensus/src/payload_builder.rs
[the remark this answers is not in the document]
eichhorl: Added: https://github.com/dfinity/ic/pull/11113/changes/bb52daf69e83c99bae07e986747e81434e2ce494

rs/https_outcalls/consensus/src/payload_builder/parse.rs
eichhorl: I think it's fine to use `HashSet`, technically the payload builder does not need to be deterministic, but we anyway do not care about the order of ids

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: I guess then we would have to break the outer loop in the second check? But in that case we would still iterate through the outer loop, until there is at least one iteration of the inner loop (which may never happen)

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Done 👍 https://github.com/dfinity/ic/pull/11113/changes/ca587ff33e7f6e7a23ecc7a19d4e68e3f1224efc

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Yes, in the future we will have to differentiate

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: You mean merging both maps into one? I guess we could, but potentially doing that in every call to `on_state_change` might be more work than one extra lookup for any delayed shares.

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Apparently this is how the callback ID needs to be incremented. Add a comment: https://github.com/dfinity/ic/pull/11113/changes/e6b71cde68c9b8ad7663ef6e3274005c54f364c8

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Good question: https://github.com/dfinity/ic/pull/11113/changes/be0f7be44778e6c415953b2cc428bf2cfb05331d

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity: Yes, I agree it would be best for performance to keep both checks.
But removing it does not affect correctness and (imo) makes it simpler and easier to reason about.

I'm happy to keep both checks maybe with a comment on the first one that explains the optimization of shortcutting early.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
pierugo-dfinity: Looks cool. I was also thinking of a simpler function that would take a `&CanisterHttpRequestContext` and a `&BTreeMap<CallbackId, HashSet<NodeId>>` and return the boolean, that would be called once during validation and once inside `find_async_receipts`.

Though also happy with the current style if you prefer.

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: I had neglected the cost of merging the two maps indeed, you should be correct.

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Ah yes I understand now.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -123,13 +124,13 @@ impl CanisterHttpPoolManagerImpl {
             .with_label_values(&["purge_shares"])
             .start_timer();
 
-        let active_callback_ids = self.active_callback_ids();
+        let known_callback_ids = self.known_callback_ids();
         let next_callback_id = self.next_callback_id();
```
pierugo-dfinity: Sorry not your changes.

Technically if we're very unlucky, the states returned by the two `get_latest_state` calls inside `known_callback_ids` and `next_callback_id` could be different, and so inconsistent. Maybe consider calling `get_latest_state` once.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -418,10 +423,10 @@ impl CanisterHttpPoolManagerImpl {
...
         let state = self.latest_state();
-        let active_contexts = &state
-            .metadata
-            .subnet_call_context_manager
-            .canister_http_request_contexts;
+        let subnet_call_context_manager = &state.metadata.subnet_call_context_manager;
+        let active_contexts = &subnet_call_context_manager.canister_http_request_contexts;
+        let delivered_contexts =
+            &subnet_call_context_manager.delivered_canister_http_request_contexts;
         let next_callback_id = self.next_callback_id();
```
pierugo-dfinity: Ditto

pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Added a comment: https://github.com/dfinity/ic/pull/11113/changes/4e4ee26f6ae99965f193585012dbaeb41475b8d4

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Good catch! I changed it so that we only get the latest state once in the beginning of `on_state_change`: https://github.com/dfinity/ic/pull/11113/changes/eda9f49c789eb52006d90efa861f6940caa66f5a

pierugo-dfinity: Nice

PR #11117, opened 2026-08-12
feat: Add a `cooling_down` flag to `SubnetTopology`

mraszyk:
A subnet that is "cooling down" inducts no ingress messages:

 * The ingress filter rejects them with the new `ErrorCode::SubnetCoolingDown` (`RejectCode::SysTransient`), so they don't make it into the ingress pool and the user gets a meaningful error. This is only an optimization, though: messages already in the pool when the subnet starts cooling down are not affected by it; and a malicious node may ignore it altogether.
 * They are excluded when building a block payload and a block containing one is invalid (`InvalidIngressPayloadReason::SubnetCoolingDown`). This is what actually guarantees that no such message ever makes it into a block.

There is no registry field backing the flag yet, so no subnet ever cools down in production; only tests set it. The registry plumbing that can actually set the flag must not ship until this code is deployed everywhere, since it also introduces a new error code that requires a staged roll-out.


rs/execution_environment/src/execution_environment.rs
alin-at-dfinity: As in the other PR, it may be worth spelling out that an error response will cause the ingress message to be immediately rejected by the receiving replica (as opposed to it being included in the ingress pool).

Also, with the addition of the `is_cooling_down()` check, this is no longer strictly about "asking the canister". I suppose it already wasn't entirely true, since we also implicitly errored out on stopped and stopping canisters.

alin-at-dfinity (approved):

PR #11121, opened 2026-08-12
fix(consensus): return Err instead of panic on missing DKG subnet record

Rachit2323:
Problem: In 4 places in the DKG (Distributed Key Generation) code, when the registry has no record for a subnet at a given version, the code calls panic!() instead of returning an error. This could crash replica nodes and potentially halt an entire subnet.

Fix: Added two new error variants (SubnetRecordNotFound, InvalidDkgConfig) and replaced all 4 panics with proper Err(...) returns so the error is handled gracefully instead of crashing.

Files changed:
- rs/types/types/src/consensus/dkg.rs — new error variants
- rs/consensus/dkg/src/payload_validator.rs — fix panic in block validation
- rs/consensus/dkg/src/payload_builder.rs — fix 3 panics in payload building

### DIFF
--- a/rs/consensus/dkg/src/payload_builder.rs
+++ b/rs/consensus/dkg/src/payload_builder.rs
@@ -2247,4 +2247,66 @@ mod tests {
         let selected = select_dealings_for_payload(&configs, &dealers_from_chain, &pool, 10);
         assert!(selected.is_empty());
     }
+
+    /// When the registry has no subnet record at the requested version,
+    /// `get_dkg_interval_length` must return `Err(SubnetRecordNotFound)` instead of panicking.
+    #[test]
+    fn test_get_dkg_interval_length_returns_err_on_missing_subnet_record() {
+        use ic_interfaces_registry_mocks::MockRegistryClient;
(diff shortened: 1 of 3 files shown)


alin-at-dfinity:
First, please note that even with errors instead of panics, the subnet would still stall, as no replica would be able to create or validate blocks.

Beyond that, on the face of it this is a sensible change: it would be nicer for replicas to continue running while the subnet stalls (easier to diagnose actual undesired behavior if the replicas are not all in a crash loop). However, we're not convinced that e.g. there doesn't exist downstream logic that expects to never be called unless there's a valid subnet record and DKG config. If any such logic starts panicking (or worse, misbehaving) because we haven't aborted the program, it may make it harder to diagnose the root cause; or actively produce incorrect behavior.

So all else being equal, I'm slightly in favor of retaining the existing panics. They should only trigger (assuming no bugs) after a subnet deletion, which is expected behavior (even if not ideal).

Rachit2323:
Makes sense — if the subnet stalls either way, keeping the panics for clearer diagnostics is reasonable.

PR #11124, opened 2026-08-13
test(cketh): build the live balance-scan harness on the shared fixtures

gregorydemay:
## Summary

The live-anvil balance-scan harness had grown into a second test fixture: it created and funded canisters, sequenced their installs, and registered ckERC20 tokens on its own, in parallel with the fixtures every other cketh integration test already uses. This PR removes that duplication — the harness now builds on the shared fixtures and keeps only what is genuinely specific to it: the local Ethereum node it owns, and the balance-scan assertions.

Collapsing the two fixtures meant removing the knobs the duplicate had accumulated:

- **Whether PocketIC runs live is decided once, after construction, not by the chain under test.** Every fixture — mocked or live — builds on an ordinary, non-live PocketIC instance, where setup calls tick deterministically; only `LiveBalanceScanSetup::new_live` switches the instance to auto-progress, and only once its whole fixture — including the ledger-suite orchestrator and the tokens it registers — is already built. Building against a live (wall-clock-paced) instance from the start made every setup call race a round deadline it didn't control, which reproducibly failed under CPU contention.
- **The harness owns the local Ethereum node** rather than borrowing a URL to it, so the type proves a node is really running behind that URL.
- **Callers can no longer supply a pre-built environment.** The chain under test already determines the canisters' init args, and every caller was passing exactly what would have been built anyway — the parameter only offered a way to get it wrong.
- **Canisters are created and installed in the same order for every chain under test**, as on master; only their init args differ. The harness' dedicated controller and its skipped ledger install turned out not to be load-bearing: the anonymous principal already controls everything PocketIC creates by default, and the skipped install was down to a Bazel data dependency that simply wasn't declared.
- **Token registration goes through a real ledger-suite orchestrator** instead of the harness impersonating one with placeholder ledgers — which is what testing the deposit flow will need next.
- **The set of tokens the balance scan credits with anvil bytecode is now exactly the set the orchestrator registered**, not a separately hardcoded list — `Holding` and the harness' token-facing methods take the orchestrator's own token type directly, so the two can no longer silently drift apart.

Behaviour of the mocked fixtures is otherwise unchanged: same canister creation order, same derived minter address, and the same cold-start property the deposit-flow tests depend on. One thing did change for every fixture, mocked included: canisters are now funded with `u64::MAX` cycles rather than `u128::MAX` — the amount that reproducibly crashes the live harness' replica on its first HTTPS outcall — deliberately funding every fixture the same way rather than carrying two cycle amounts. No cketh test asserts on cycle balances, so this is not expected to be observable.


rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -68,12 +68,14 @@ pub struct Holding {
...
-const ECDSA_KEY_NAME: &str = "key_1";
-
 /// A fixed non-anonymous principal used as the canisters' controller and as the minter's stand-in
 /// ledger-suite-orchestrator id, so this harness can register supported tokens itself.
+///
+/// The minter and EVM RPC canister are deliberately controlled by this fixed principal rather than
+/// the anonymous default every other ckETH fixture uses: with the anonymous default, the very first
+/// HTTPS outcall the minter issues after going live reproducibly crashed the PocketIC replica with a
+/// cycle-accounting assertion failure (`Invalid cycle change`). This harness keeps its own
+/// canister-creation/controller setup separate from [`crate::CkEthSetup::new`] for that reason.
```
gregorydemay: 🤖🧐 🟠 **Medium** — this comment very likely blames the wrong axis, and a wrong documented cause is worse than none (C4/G21).

The assertion is `SystemStateModifications::validate_cycle_change` (`rs/embedders/src/wasmtime_embedder/system_api/sandbox_safe_system_state.rs:160-191`): `expected` is built from `call_context_balance_taken` (cycles the canister accepted via `msg_cycles_accept`), `got` is the observed balance delta. The reported `expected Added(1_360_000_000), got Added(0)` is the exact signature of a **saturated cycles balance**:

- `impl AddAssign for Cycles` saturates — `rs/types/cycles/src/cycles.rs:138-141`
- `SystemState::add_cycles` is `self.cycles_balance += amount` — `rs/replicated_state/src/canister_state/system_state.rs:2006`
- the shared helper you converged onto does `env.add_cycles(id, u128::MAX)` — `lib.rs:788/790/792`

At `u128::MAX` the canister cannot observe *any* added cycles, so accepting 1.36 G cycles yields a zero delta and trips the check. The live harness's own `u128::from(u64::MAX)` (L109/L114) leaves headroom. That also explains why the mocked fixture never trips it despite the same `u128::MAX`: it is the live harness that adds the NNS subnet, and hence real cycle accounting.

Concretely: the axis that must stay separate is the **cycles amount**, not the controller. Cheap experiment — redo the convergence but keep `u128::from(u64::MAX)`. If it passes, please drop this paragraph; `controller()` then keeps only its real (and sufficient) justifications: the LSO stand-in id, and controller-only `fetch_canister_logs` at L233-235.

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -276,17 +268,17 @@ fn install_minter(
...
         ledger_id,
-        // anvil is a fresh chain with no finalized blocks, so track its "latest" head.
-        ethereum_block_height: ic_cketh_minter::endpoints::CandidBlockTag::Latest,
+        ethereum_block_height: backend.ethereum_block_height(),
         minimum_withdrawal_amount: Nat::from(CKETH_MINIMUM_WITHDRAWAL_AMOUNT),
         next_transaction_nonce: Nat::from(0_u8),
-        last_scraped_block_number: Nat::from(0_u8),
+        last_scraped_block_number: backend.last_scraped_block_number(),
         evm_rpc_id: Some(evm_rpc_id),
     };
```
gregorydemay: 🤖🧐 🟠 **Medium** — the duplication this PR set out to remove is still here.

This body still hand-builds the full 9-field `MinterInitArgs` and the install call, near-identically to `lib.rs:862-880`; the same holds for `install_evm_rpc` here (L257) vs `lib.rs:882-889`. After the refactor the only genuine deltas between the two pairs are:
- the install **sender** (`Some(controller())` vs `None`), and
- the two values now behind `EvmRpcBackend`.

So the PR added an abstraction (`EvmRpcBackend`) without deleting the copies that abstraction was meant to justify — net indirection for the same knowledge. The fix that actually pays for the enum: give the shared `install_minter`/`install_evm_rpc` the sender — e.g. carry `controller: Option<Principal>` on `CkEthCanisters`, which keeps both signatures at 3 args (F1) — and delete both local copies here.

Note this does **not** require converging canister creation/cycles: the live harness can keep its `create_canister_with_settings` + `u64::MAX` cycles and simply assemble a `CkEthCanisters` to hand to the shared installers. (G5)

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +774,118 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+}
+
+/// Where the EVM RPC canister's outcalls are answered, and the Ethereum chain state that follows
+/// from it.
+enum EvmRpcBackend<'a> {
+    /// Canned JSON-RPC mocks pinned to a historical mainnet snapshot.
+    Mocked,
+    /// A live anvil node reached over HTTP at `url`: a fresh chain with no finalized blocks yet.
+    Anvil(&'a str),
+}
```
gregorydemay: 🤖🧐 🔵 **Nit** — the enum's name covers only one of its three methods, and as it stands it deduplicates nothing.

`install_args()` says *where the EVM RPC canister points*; `ethereum_block_height()` and `last_scraped_block_number()` are **minter** init args describing *which chain is under test*. The doc line papers over that with "and the Ethereum chain state that follows from it", but the name `EvmRpcBackend` only names the first facet, so a reader meeting `backend.last_scraped_block_number()` in `install_minter` has to jump here to see why an "EVM RPC backend" has an opinion about the minter's scrape cursor. `EthereumBackend` / `ChainUnderTest` would cover all three.

More substantively: each of the three methods is called at exactly two sites, each with a *statically known* variant (`Mocked` in `lib.rs`, `Anvil` in `live_scan.rs`). `Mocked`'s values and `Anvil`'s values were never duplicated copies of one fact — they are two different facts, each stated once. So the enum replaces two inline literals with a match that every caller resolves statically, and threads a constant-valued parameter through `lib.rs`'s two installers. It earns its keep only once the installers themselves are shared — see the comment on `live_scan.rs::install_minter`. (G13/G15)

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +774,118 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+            EvmRpcBackend::Anvil(_) => CandidBlockTag::Latest,
+        }
+    }
+
+    fn last_scraped_block_number(&self) -> Nat {
+        match self {
+            EvmRpcBackend::Mocked => LAST_SCRAPED_BLOCK_NUMBER_AT_INSTALL.into(),
+            EvmRpcBackend::Anvil(_) => 0_u8.into(),
+        }
+    }
```
gregorydemay, line 882: 🤖🧐 🔵 **Nit** — asymmetric commenting: both `Anvil` arms explain themselves, both `Mocked` arms don't. A reader cannot tell *why* mocking the JSON-RPC responses implies `Finalized` and `LAST_SCRAPED_BLOCK_NUMBER_AT_INSTALL` — the real reason (the mocks replay a historical mainnet snapshot whose blocks are long finalized, and the fixture's canned responses are pinned to that scrape cursor) lives only in the variant doc above, and only half of it. One clause on each `Mocked` arm closes it. (C4)

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -276,17 +268,17 @@ fn install_minter(
     minter_id: Principal,
     ledger_id: Principal,
     evm_rpc_id: Principal,
+    backend: &EvmRpcBackend,
 ) {
     let args = MinterInitArgs {
         ethereum_network: EthereumNetwork::Mainnet,
-        ecdsa_key_name: ECDSA_KEY_NAME.to_string(),
+        ecdsa_key_name: "key_1".to_string(),
```
gregorydemay: 🤖🧐 🔵 **Nit** — this deletes a named constant in favour of a magic string, in a PR about deduplication. `ECDSA_KEY_NAME` is gone and `"key_1"` is now written literally here *and* at `lib.rs:864` — two occurrences where there was one named one. The doc it carried was usefully relocated to `pocket_ic_builder()`, but the name itself was worth keeping: hoist `const ECDSA_KEY_NAME: &str = "key_1";` into `lib.rs` and use it from both installers. (G25/G5)

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +774,118 @@ fn evm_rpc_wasm() -> Vec<u8> {
     )
 }
 
-fn install_minter(
-    env: &PocketIc,
-    ledger_id: Principal,
+/// The 3 canisters every ckETH fixture creates: the minter, its ckETH ledger and the EVM RPC
+/// canister it calls out to.
```
gregorydemay: 🤖🧐 🔵 **Nit** — "every ckETH fixture" overstates it: `CkErc20LiveScanSetup` never constructs this struct and creates its own three canisters, in a *different order* (ledger first, so its minter gets a different canister id — fine there, but it means the "minter first" invariant in `create_cketh_canisters` is not universal either). Suggest scoping the wording to `CkEthSetup`. (C4)

gregorydemay:
🤖🧐 **VERDICT: CHANGES_REQUESTED** — 0 blockers, 2 mediums, 4 nits; CI pending (`Build IC`, `Bazel Test All`, `Bazel Test All on RBE` still running; nothing red).

<details>

### Verified claims

- **`CkEthSetup::new` is behaviourally unchanged.** Compared operation for operation: same three `create_canister` calls in the same order (minter first, so the hard-coded canister id / Ethereum address still hold), same `u128::MAX` cycles, same ledger init args, same installs in the same sequence, no added `tick()`. `EvmRpcBackend::Mocked.install_args()` is byte-equal to the previous `InstallArgs::default()` — every `InstallArgs` field is an `Option` and the struct derives `Default`, so the explicit `override_provider: None` changes nothing. The fixture's cold-start property (no IC rounds executed in `new`) is preserved.
- **The live harness's ordering invariant survives.** `make_live(None)` still runs *before* `install_minter` (`live_scan.rs:119` → `:121`) and after `install_evm_rpc` (`:111`), and the comment explaining why (install schedules timers whose outcalls would stall holding task guards) is retained verbatim.
- **The module doc fix is correct.** `CkErc20Setup` has run on PocketIC since #10955; the new wording ("also runs on PocketIC but answers ... with canned canister-http mocks") states the real distinction.

### The three axes kept separate

- **Controller / settings / cycles** — keeping this separate is right, but for a reason the code does not state. See the inline comment on `live_scan.rs:74`: the `Invalid cycle change: expected Added(1360000000), got Added(0)` signature points at the shared helper's `add_cycles(id, u128::MAX)` saturating the balance, not at the anonymous controller. Worth a one-line re-test before the comment is frozen into the tree.
- **Ledger placeholder** — justified as written. Adding `LEDGER_CANISTER_WASM_PATH` to the `deposit_from_cex` target to install a ledger the balance-scan path never calls buys nothing.
- **ckERC20 activation via a bare `upgrade_canister`** — justified. Stop/start exists in the mocked fixture to quiesce mocked outcalls; live outcalls are real, so there is nothing to quiesce.

- **Duplication:** found — `install_minter`/`install_evm_rpc` remain duplicated across the two files after the refactor (🟠, inline). Also a minor regression: the `ECDSA_KEY_NAME` constant was replaced by a literal now written twice (🔵).
- **Structural duplication against the codebase:** the mirrored pair above is the only instance; no other sibling type is being paralleled.
- **Unused derives:** none — the PR adds no derives.
- **Primitive-obsession parameters:** cleared. `EvmRpcBackend::Anvil(&str)` matches the surrounding fixture style (`anvil.url()` already returns `&str`).
- **Divergent invariant handling:** one instance — the two fixtures fund canisters with different cycle amounts (`u128::MAX` vs `u64::MAX`) with no stated rule for which applies when. Folded into the 🟠 on `live_scan.rs:74`.
- **Silent fallbacks:** none found.
- **Test-only code in a production module:** N/A — the whole crate is `test_utils`.
- **Redundant / derivable parameter:** `backend` is constant (`Mocked`) at both `lib.rs` call sites; folded into the 🟠/🔵 pair on the duplication and the enum.
- **Decision the caller should not own:** the install *sender* — pushing it into the shared installers is what unlocks deleting the duplicate copies.
- **Docs/comments:** three findings (🟠 on the crash rationale, 🔵 on the uncommented `Mocked` arms, 🔵 on the `CkEthCanisters` doc). No JIRA refs, no metadata, no commented-out code.

### Testing

No behaviour change, so no new test is owed; the existing suites are the regression net and all pass locally:

- `//rs/ethereum/cketh/minter:integration_tests_tests/cketh_test` PASSED (69.4s)
- `//rs/ethereum/cketh/minter:integration_tests_tests/ckerc20_test` PASSED (132.5s)
- `//rs/ethereum/cketh/test_utils:lib_tests` PASSED
- `//rs/ethereum/cketh/minter:deposit_from_cex` PASSED (64.8s) — the live PocketIC + anvil harness this PR rewires
- `cargo clippy --all-targets --all-features -p ic-cketh-test-utils -p ic-cketh-minter -- -D warnings` clean

All run with `--nocache_test_results`.

</details>

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +774,118 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+        match self {
+            EvmRpcBackend::Mocked => LAST_SCRAPED_BLOCK_NUMBER_AT_INSTALL.into(),
+            EvmRpcBackend::Anvil(_) => 0_u8.into(),
+        }
+    }
+}
+
+fn install_minter(env: &PocketIc, canisters: &CkEthCanisters, backend: &EvmRpcBackend) {
     let args = MinterInitArgs {
         ecdsa_key_name: "key_1".parse().unwrap(),
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in c2053c6e8e — restored `ECDSA_KEY_NAME` as a shared constant and use `.to_string()` instead of the fallible `.parse().unwrap()`.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Confirmed and fixed in f683e8c657. Ran the suggested one-line experiment: converged everything else (controller/settings included) but bumped the live harness's cycles to `u128::MAX` — the exact same `Invalid cycle change` crash reproduced with `controller()` completely untouched, confirming cycles saturation (not the controller) is the cause. Rewrote `controller()`'s doc to state its real, sufficient justifications, and moved the cycles-headroom rationale to the `add_cycles` call sites.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in 5c6bda60ec — `CkEthCanisters` now carries `controller: Option<Principal>`, threaded through the shared `install_minter`/`install_evm_rpc` as the install/upgrade sender. The live harness keeps its own canister creation/cycles/controller setup, but now assembles a `CkEthCanisters` and calls the shared installers instead of duplicating them. Both signatures stay at 3 args.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 7e84561945 (rename to `EthereumBackend`, now that 5c6bda60ec made the installers genuinely shared) and 04d586b9c6 (Mocked-arm comments, addressing the sibling nit below).

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 04d586b9c6 — added a one-clause reason to each `Mocked` arm, matching the `Anvil` arms.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in c2053c6e8e — restored `const ECDSA_KEY_NAME: &str = "key_1";` in lib.rs, used by the (now single, shared) `install_minter`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 206bb5c905 — reworded to describe how each fixture actually uses `CkEthCanisters` (CkEthSetup::new creates+installs all 3 via create_cketh_canisters; the live harness creates its own 3, ledger first, under its own controller, and only assembles this struct for the shared installers).

gregorydemay:
🤖 Addressed all 6 review findings, one commit each:

- `5c6bda60ec` — Medium: thread the install sender through the shared `install_minter`/`install_evm_rpc` via `CkEthCanisters.controller`, deleting the duplicated copies in `live_scan.rs`.
- `c2053c6e8e` — Nit + Copilot: restored `ECDSA_KEY_NAME` as a single shared constant; dropped the fallible `.parse().unwrap()`.
- `f683e8c657` — Medium: corrected the wrongly-documented crash cause.
- `7e84561945` — Nit: renamed `EvmRpcBackend` → `EthereumBackend` now that it's genuinely shared.
- `04d586b9c6` — Nit: explained the `Mocked` arms of `ethereum_block_height`/`last_scraped_block_number`.
- `206bb5c905` — Nit: reworded `CkEthCanisters`'s doc to match how each fixture actually uses it.

**Cycles experiment (per the Medium-1 root-cause comment):** re-tested with `controller()`/`CanisterSettings` completely untouched, only bumping the live harness's cycles from `u128::from(u64::MAX)` to `u128::MAX` — the identical `Invalid cycle change: expected Added(1360000000), got Added(0)` crash reproduced. This confirms cycle-balance saturation, not the controller, was the cause. `controller()` is kept (it has two independent, sufficient justifications — the LSO stand-in id and controller-only `fetch_canister_logs`), but its doc no longer attributes the crash to it; that's now explained at the `add_cycles` call sites.

Verification (all synchronous, `--nocache_test_results` where noted):
- `rustfmt` — clean
- `cargo check --all-targets --all-features -p ic-cketh-test-utils -p ic-cketh-minter` — clean
- `cargo clippy` (pinned flags, `--all-targets`) — clean
- `bazel test` (via `./ci/container/container-run.sh`) for `cketh_test`, `ckerc20_test`, `test_utils:lib_tests`, `deposit_from_cex` — all 4 PASSED, `--nocache_test_results`

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -104,20 +100,33 @@ impl CkErc20LiveScanSetup {
         // A placeholder ckETH ledger: the minter stores its id at init but never calls it on the
         // balance-scan path, so it is left uninstalled.
         let ledger_id = env.create_canister();
+        // Unlike the mocked fixture's `u128::MAX` (never executed as real cycles since it never
+        // goes live), this must leave headroom below the saturating `Cycles` balance ceiling: a
+        // canister that has already saturated cannot observe *any* further cycle addition, and the
+        // first HTTPS outcall a live canister accepts cycles for trips a cycle-accounting assertion
+        // in the replica if its balance cannot move.
```
gregorydemay: 🤖🧐 🔵 **Nit** (follow-up to the resolved 🟠) — the load-bearing half of this comment is now proven and reads well. Two small things left, both about the parenthetical:

1. *"never executed as real cycles since it never goes live"* is an unverified causal claim, and it is the same shape as the original finding — asserting a mechanism the experiment did not test. Your experiment proved the cycles amount is the trigger **in this harness**; it did not establish *why* the mocked fixture survives `u128::MAX`. Live vs. non-live mode is not itself what makes cycle accounting real — PocketIC runs the same replica execution either way — so the likelier distinction is the NNS subnet this harness adds and the cost schedule that follows from it. I have not proven that either, which is exactly why I would drop the parenthetical rather than swap one guess for another: *"unlike the mocked fixture's `u128::MAX`, this must leave headroom …"* states everything the evidence supports.

2. Placement: the comment sits above `create_canister_with_settings` but describes the `add_cycles` on the next line, and the identical constraint on `minter_id` at L113 carries no marker. Moving it directly above L110 (or above both `add_cycles` calls as a single note) would put it where a reader changing an amount will actually look.

gregorydemay:
🤖🧐 **VERDICT: READY** (re-review) — 0 blockers, 0 mediums, 1 nit; CI 21 pass / 0 fail / 6 pending. **Merge gate:** `Bazel Test All`, `Bazel Test All on RBE`, `Build IC`, `Cargo Build/Lint Linux`, `Bazel Run Fuzzers` must go green before merge. Final approval is the human's — this is not an approval.

<details>
<summary>Re-review details</summary>

### Per-finding resolution — all six resolved

| Prev | Commit | Resolution |
|---|---|---|
| 🟠 duplication | `5c6bda60ec` | **Genuinely collapsed, not moved.** Both copies in `live_scan.rs` are deleted (−37 lines); it now assembles a `CkEthCanisters` and calls the shared installers. `CkEthCanisters.controller` is threaded into all *three* installers, so the sender is now handled identically everywhere instead of `install_ledger` hard-coding `None`. |
| 🟠 wrong crash cause | `f683e8c657` | Corrected, and the experiment (bump *only* the cycles, leave `controller()` untouched → identical crash) is the right way to have settled it. `controller()`'s doc now states only its two real justifications. One nit remains on the replacement wording — see inline. |
| 🔵 `ECDSA_KEY_NAME` | `c2053c6e8e` | Restored as a documented shared const; also drops the gratuitously fallible `"key_1".parse().unwrap()`. |
| 🔵 enum name | `7e84561945` | `EthereumBackend` covers all three methods; the new doc names both facets explicitly. |
| 🔵 uncommented `Mocked` arms | `04d586b9c6` | Both arms now carry a one-clause reason, symmetric with `Anvil`. |
| 🔵 `CkEthCanisters` doc | `206bb5c905` | Accurate, and usefully spells out how the two fixtures differ in creation order and in which canisters they install. |

### Does `EthereumBackend` earn its keep now? Yes.

My earlier objection was that each method resolved to a statically-known variant at both call sites, so nothing was deduplicated. That is gone by construction: `install_minter` and `install_evm_rpc` are now single shared functions reached with *both* variants, so `backend` is a genuine parameter and each method has exactly one call site consuming a dynamic value. This is the F3 "take an enum, not a bool" form doing real work. It should stay.

### Invariants re-checked after the deeper refactor

- **`make_live` before the minter install** — holds (`live_scan.rs:127` → `:129`), with the explanatory comment intact. `install_evm_rpc` moved from before to after `create_canister(minter)` (it now needs `minter_id` to build `CkEthCanisters`), but is still before `make_live`; canister creation *order* is unchanged, so both canister ids are unchanged, and the EVM RPC install schedules no outcall timers. Benign.
- **`CkEthSetup::new` unchanged operation-for-operation** — re-verified against the pre-PR base, not just the previous head: same three creates in the same order, same `u128::MAX`, same ledger args, same install sequence, no added `tick()`. The two new indirections are value-preserving: `canisters.controller` is `None` for `create_cketh_canisters`, matching the previous literal `None` at all three install sites; and `ecdsa_key_name` is a plain `String` (`lifecycle/init.rs:20`), so `ECDSA_KEY_NAME.to_string()` ≡ `"key_1".parse().unwrap()`. Cold-start property preserved.
- **Live harness install senders** — `Some(controller())` for both minter and EVM RPC, exactly as the deleted copies did.

### Nothing new introduced

No dead code or orphaned imports (`EthereumNetwork`, `MinterInitArgs`, `CKETH_MINIMUM_WITHDRAWAL_AMOUNT`, `ETH_HELPER_CONTRACT_ADDRESS`, `evm_rpc_wasm` all correctly dropped from `live_scan.rs`; `minter_wasm`/`Nat` correctly kept). No new derives. No silent fallbacks. `CkEthCanisters.controller` is named for the identity rather than the role it fills — cleared, since PocketIC requires an install sender to *be* a controller, so the two coincide by construction.

### Re-ran (all fresh, `--nocache_test_results`)

- `//rs/ethereum/cketh/minter:integration_tests_tests/cketh_test` PASSED 73.0s
- `//rs/ethereum/cketh/minter:integration_tests_tests/ckerc20_test` PASSED 134.5s
- `//rs/ethereum/cketh/test_utils:lib_tests` PASSED
- `//rs/ethereum/cketh/minter:deposit_from_cex` PASSED 74.8s
- `cargo clippy --all-targets --all-features -p ic-cketh-test-utils -p ic-cketh-minter -- -D warnings` clean

</details>

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in d14bc70713 — dropped the unverified parenthetical (live-vs-not doesn't itself explain why the mocked fixture survives `u128::MAX`) and moved the comment to sit directly above the `add_cycles` call it describes, noting it applies to `minter_id`'s identical call too.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +852,143 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+                ..Default::default()
+            }),
+        ),
+    }
+}
+
+fn create_cketh_canisters(env: &PocketIc, controller: Option<Principal>) -> CkEthCanisters {
+    // Create minter canister first to match canister ID and Ethereum address hardcoded in tests.
+    let minter_id = create_canister(env, controller);
+    env.add_cycles(minter_id, u128::from(u64::MAX));
```
gregorydemay: 🧐 🟠 **Medium** — unifying the cycle amount silently deleted the only record of *why* it is not `u128::MAX`. `d14bc70` deliberately placed a comment above this `add_cycles` explaining that the value must leave headroom below the saturating `Cycles` ceiling, because a canister already at the ceiling cannot observe a further addition and the first HTTPS outcall a live canister accepts cycles for then trips a cycle-accounting assertion in the replica. That constraint still holds and is still load-bearing for `deposit_from_cex`, but nothing in the tree says so any more: `u128::from(u64::MAX)` now reads as an arbitrary magic value repeated three times (G25/G32), and the obvious "simplification" back to `u128::MAX` would break only the anvil-backed target. Please restore the rationale once — ideally as a named constant carrying the doc, e.g. `/// … const CANISTER_CYCLES: u128 = u64::MAX as u128;` — so the three call sites share one authoritative explanation (G5). C2 also applies in reverse: the comment was not obsolete, the code it described is still here.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+        self.live = true;
+        self
+    }
+
+    fn build(self) -> CkEthSetup {
+        let env = self.env.unwrap_or_else(|| Arc::new(new_env(self.live)));
+        let controller = self.live.then(live_controller);
+        let canisters = create_cketh_canisters(&env, controller);
+        if !self.live {
+            // Tried installing the real ledger for the live harness too, to drop this
```
gregorydemay: 🧐 🟠 **Medium** — this comment leads with the incidental obstacle and buries the reason. The reason the live harness skips the ledger is the last sentence: the balance scan never calls it. The Bazel data-dependency gap is a fixable accident of the `deposit_from_cex` target, not a justification — as written a reader concludes "declare the ledger Wasm as a data dep and this conditional disappears", which inverts cause and effect and invites work that would buy nothing. Please reorder to lead with the reason and mention the obstacle second (one clause is enough). Separately, six lines of prose about Bazel sandboxing sit at a different abstraction level from the three-line install sequence they interrupt (G6/G34) — the "why live mode has no ledger" half reads better on `with_live_mode`'s doc, next to the other axes it switches, leaving a one-liner here.

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -253,68 +218,26 @@ fn candidates_in_log(line: &str) -> Option<u64> {
...
+            cketh.minter_id,
+            minter_wasm(),
+            Encode!(&MinterArg::UpgradeArg(upgrade)).unwrap(),
+            Some(live_controller()),
+        )
+        .expect("BUG: failed to activate the ckERC20 feature");
 }
 
-fn register_supported_tokens(env: &PocketIc, minter_id: Principal) {
+fn register_supported_tokens(cketh: &CkEthSetup) {
```
gregorydemay: 🧐 🟠 **Medium** — `register_supported_tokens` hand-rolls the `update_call` that `CkEthSetup::add_ckerc20_token(&self, from, erc20)` already wraps, so the harness reimplements a shared method it is now sitting directly on top of (G5). Given this PR's premise is reuse over relocated duplication, this is the one place it still does not hold. `cketh.add_ckerc20_token(live_controller(), &arg).expect("BUG: add_ckerc20_token was rejected");` is a drop-in.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+/// logs (controller-only by default) as that same principal.
+fn live_controller() -> Principal {
+    Principal::from_slice(&[0x0a; 10])
+}
+
+/// Builds the PocketIC instance for [`CkEthSetupBuilder::build`]: the fiduciary subnet every
+/// ckETH fixture needs for the secp256k1 `key_1` used by the minter, plus — in live mode — an NNS
+/// subnet (required by [`PocketIc::make_live`]) and going live immediately, before any canister of
+/// this fixture exists to schedule a timer whose outcall could stall waiting for an answer.
+fn new_env(live: bool) -> PocketIc {
```
gregorydemay: 🧐 🔵 **Nit** — `new_env(live: bool)` is a flag argument driving two materially different instances (F3/G15), and it reads badly at the call sites: `new_env(false)` on L815 tells the reader nothing. An enum would name both modes and make the asymmetry below self-documenting, e.g. `enum EnvMode { Mocked, Live }` / `fn new_env(mode: EnvMode)`. Low stakes — the function is private with two call sites — but it is cheap here.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+    if live {
+        let mut env = pocket_ic_builder().with_nns_subnet().build();
+        let _gateway = env.make_live(None);
+        env
+    } else {
```
gregorydemay: 🧐 🔵 **Nit** — the live branch does not set `IcpConfig { canister_execution_rate_limiting: Disabled }` that the mocked branch sets two lines below. This matches the pre-refactor `build_live`, so it is preserved behaviour rather than a regression, but now that both modes are built by one function the divergence is visible and unexplained: a reader cannot tell whether live mode wants the rate limiter on (real rounds, real timing) or whether the flag was simply never carried over. One clause saying which would settle it.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+}
+
+/// Builds the PocketIC instance for [`CkEthSetupBuilder::build`]: the fiduciary subnet every
+/// ckETH fixture needs for the secp256k1 `key_1` used by the minter, plus — in live mode — an NNS
+/// subnet (required by [`PocketIc::make_live`]) and going live immediately, before any canister of
+/// this fixture exists to schedule a timer whose outcall could stall waiting for an answer.
+fn new_env(live: bool) -> PocketIc {
+    if live {
+        let mut env = pocket_ic_builder().with_nns_subnet().build();
+        let _gateway = env.make_live(None);
```
gregorydemay: 🧐 🔵 **Nit** — `make_live` returns a `Url`, not an RAII guard, so naming the binding `_gateway` implies a lifetime contract that does not exist (the value is dropped immediately either way, and the gateway stays up). `env.make_live(None);` says exactly what happens (N1/N6, G16). Carried over from `build_live`, so purely a Boy-Scout item.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+    backend: EthereumBackend<'a>,
+    live: bool,
+}
+
+impl<'a> CkEthSetupBuilder<'a> {
+    /// Reuses an existing PocketIC instance instead of building one via [`new_env`], so this
+    /// fixture can be composed with others that need the same instance (e.g. `CkErc20Setup`'s
+    /// orchestrator). Not combined with [`Self::with_live_mode`] today: live mode's instance needs
+    /// the added NNS subnet and to already be live before any canister exists.
+    fn with_env(mut self, env: Arc<PocketIc>) -> Self {
```
gregorydemay: 🧐 🔵 **Nit** — the doc says `with_env` is "not combined with `with_live_mode` today", but nothing enforces it: if both are set, `build` takes the supplied env while `self.live` still selects the controller and skips the ledger install, yielding a live-flagged fixture on a non-live instance — a silent, hard-to-diagnose combination. Both knobs are private with one call site each, so the risk is low, but a `debug_assert!(!(self.live && self.env.is_some()))` in `build` would encode the invariant the comment currently only asserts (G27/G31).

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,103 @@ impl CkEthSetup {
...
+        self.env = Some(env);
+        self
+    }
+
+    fn with_ethereum_backend(mut self, backend: EthereumBackend<'a>) -> Self {
+        self.backend = backend;
+        self
+    }
+
+    /// Switches PocketIC to live mode: an NNS subnet in addition to the fiduciary one, going live
```
gregorydemay: 🧐 🔵 **Nit** — this doc enumerates what live mode switches ("an NNS subnet …, going live …, and a fixed non-anonymous controller"), reading as a complete list, but `live` still drives a fourth axis: skipping the ckETH ledger install. Correctly dropping the cycles item is right (it no longer forks), but the ledger axis should be added here — especially if the L761 comment moves up as suggested.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +852,143 @@ fn evm_rpc_wasm() -> Vec<u8> {
     )
 }
 
-fn install_minter(
-    env: &PocketIc,
-    ledger_id: Principal,
+/// The minter, its ckETH ledger and the EVM RPC canister it calls out to, plus the sender to
+/// install/upgrade them as (`None`: every other fixture's anonymous default; `Some`: the live
+/// harness' [`live_controller`]). Built by [`create_cketh_canisters`] and installed by
+/// [`install_ledger`]/[`install_minter`]/[`install_evm_rpc`], shared by every
```
gregorydemay: 🧐 🔵 **Nit** — "shared by every [`CkEthSetupBuilder::build`]" is a leftover from the `build_mocked`/`build_live` era: there is exactly one `build` now, so "every" reads as if several existed (C2/C4). "shared by both modes of [`CkEthSetupBuilder::build`]" or just "used by [`CkEthSetupBuilder::build`]".

gregorydemay (commented):
_(posted as a comment review: GitHub does not allow requesting changes on one's own PR — the verdict below is CHANGES_REQUESTED.)_

🧐 **VERDICT: CHANGES_REQUESTED** — 0 blockers, 3 mediums, 6 nits; CI pending (21 pass / 6 pending / 10 skipping, incl. `Bazel Test All` and `Bazel Test All on RBE`).

<details>

Scope reviewed: `33d34d1` and `651d03a` only; earlier commits on the branch were previously cleared.

### Verified locally (all four green, in-container, matching this tree)
`//rs/ethereum/cketh/minter:integration_tests_tests/cketh_test`, `//rs/ethereum/cketh/minter:integration_tests_tests/ckerc20_test`, `//rs/ethereum/cketh/test_utils:lib_tests`, `//rs/ethereum/cketh/minter:deposit_from_cex`.

### The three properties this refactor had to preserve — all hold

- **Cold start of the default path.** `make_live` occurs exactly once in the crate, in `new_env`'s `if live` branch; `live` is set only by `with_live_mode`, whose only caller is `LiveBalanceScanSetup::new_live`. `CkEthSetup::new`/`Default` cannot reach it, so the ~20 deposit-flow tests that depend on the fixture not driving rounds on its own are safe. Verified by reading the call graph, not from the commit message.
- **Creation/install order.** `build()` with `live == false` is `create_cketh_canisters` (minter → ledger → evm_rpc), then `install_ledger` → `install_evm_rpc` → `install_minter` — byte-for-byte the order of `origin/master`'s `CkEthSetup::new`. `MINTER_ADDRESS` derivation is therefore untouched. Live mode now shares that order rather than its old ledger-first one.
- **Unified cycle amount.** Nothing under `rs/ethereum` reads a canister cycle balance or depends on the `u128::MAX` ceiling (only production `with_cycles(MIN_ATTACHED_CYCLES)` call sites and the orchestrator's own unrelated `Cycles` use). The drop to `u64::MAX` is behaviourally inert — but the *rationale* for it was lost, see the Medium at `create_cketh_canisters`.

### Genuine reuse
`live_scan.rs` no longer re-implements any fixture step: no canister creation, no settings, no cycles, no install ordering. `LiveBalanceScanSetup` is a thin `{ cketh, anvil }` wrapper and everything left on it (`Holding`, `SupportedToken`, `credit_deposits`, `await_scan`, `balance_scan_candidates`, `deposit_erc20`, `register_deposit_address`, `depositor`) is genuinely balance-scan-specific. The one exception is `register_supported_tokens`, flagged inline.

### Maintainability rundown
- **Duplication (in-diff):** one instance — `register_supported_tokens` re-implements `CkEthSetup::add_ckerc20_token` (inline 🟠). The structural duplication this PR set out to remove is gone.
- **Structural duplication vs. the codebase:** none found; `CkEthSetupBuilder` has no sibling it mirrors.
- **Unused derives:** none — `#[derive(Clone, Copy)]` on `SupportedToken` is required by `for token in SupportedToken::ALL`.
- **Primitive-obsession parameters:** `new_env(live: bool)` (inline 🔵, F3/G15). No other bare primitive stands beside a domain newtype.
- **Divergent invariant handling:** one — `IcpConfig`'s rate-limiting flag set on the mocked branch only (inline 🔵). Preserved behaviour, but now visible and unexplained.
- **Silent fallbacks:** none. `balance_scan_candidates`' `.unwrap_or(0)` is a documented "no scan has logged yet" default, not a masked failure; every other path `expect`s with a `BUG:` message, which is right for a test fixture.
- **Dead code:** none — `build_mocked`/`build_live`/`CkErc20LiveScanSetup` are fully removed, with no dangling references anywhere in `rs/`.
- **Test-only code in production modules:** n/a, this is a `test_utils` crate.

### Docs
Every rustdoc intra-doc link in both files resolves to an item that still exists — I checked each one, including the ones rewritten in `651d03a` (`new_env`, `live_controller`, `create_cketh_canisters`, `install_ledger`/`install_minter`/`install_evm_rpc`, `PocketIc::make_live`, `crate::mock::MockJsonRpcProviders`, `crate::ckerc20::CkErc20Setup`, `Self::balance_scan_candidates`). No reference to the deleted `build_mocked`/`build_live` survives. Three smaller doc-vs-code slips are flagged inline.

**Cross-cutting: the PR description now drifts from the code.** It still describes `with_live_mode` as switching "`u64::MAX` cycles instead of `u128::MAX`" as one of its knobs, calls this "the one real fork in `with_live_mode`'s cycle amount", and says it is "documented at the `add_cycles` call sites". After `651d03a` the amount is unified for both modes, there is no fork, and there is no comment at those call sites. Please fold the second commit into the description before this leaves draft — as it stands the description is the only surviving explanation of a constraint the code no longer mentions, which is also why the inline Medium asks for that comment back.

### Testing
No new behaviour to cover — this is a pure fixture refactor whose contract is "the existing suites still pass", which they do. `deposit_from_cex` remains the sole consumer of the live path and is unchanged apart from the rename. No test earns removal, none duplicates another, and no assertion here is tautological or mock-constant-blind.

</details>

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in ad8ce46e58 — restored the rationale as a named `CANISTER_CYCLES` constant carrying the doc, used by all three `add_cycles` sites in `create_cketh_canisters`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 5d619c334d — reordered to lead with the reason (balance scan never calls the ledger), moved the Bazel-obstacle explanation to `with_live_mode`'s doc (also folds in nit 8 below), and left a one-line pointer at the call site.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in a114ffdec4 — `register_supported_tokens` now calls `cketh.add_ckerc20_token(live_controller(), &arg)` instead of hand-rolling the `update_call`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in d000a55ab7 — introduced `enum EnvMode { Mocked, Live }`, `new_env` now takes `EnvMode` instead of a bool.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Tried applying it to both modes and ran `deposit_from_cex` fresh (`--nocache_test_results`): still passes. Collapsed in 501c22f0d7 — no divergence left to explain.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 501c22f0d7 (same commit as the rate-limiting unification, since both touch `new_env`) — dropped the `_gateway` binding, now plain `env.make_live(None);`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 78aa809289 — added `debug_assert!(!(self.live && self.env.is_some()), ..)` in `build`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 5d619c334d (same commit as the medium above, since both touch `with_live_mode`'s doc) — added the ledger axis to the list.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 986fd3886c — reworded to "used by `CkEthSetupBuilder::build` for both modes".

gregorydemay:
🤖 Addressed all 9 findings from the latest review, one commit each (2 pairs combined where they touched the exact same lines):

- `ad8ce46e58` — Medium: restored the cycles-headroom rationale as a named `CANISTER_CYCLES` constant.
- `5d619c334d` — Medium + Nit: reordered the ledger conditional's comment to lead with the reason, moved the "why" to `with_live_mode`'s doc (also adds the ledger axis it was missing there).
- `a114ffdec4` — Medium: `register_supported_tokens` now calls `cketh.add_ckerc20_token(..)` instead of hand-rolling the `update_call`.
- `501c22f0d7` — Nit + Nit: unified the rate-limiting config across both `new_env` modes (verified — see below), and dropped the misleading `_gateway` binding.
- `d000a55ab7` — Nit: `new_env` now takes an `EnvMode` enum instead of a bool.
- `78aa809289` — Nit: `debug_assert!` encoding the `with_env`/`with_live_mode` mutual-exclusion invariant.
- `986fd3886c` — Nit: fixed `CkEthCanisters`'s doc, no more "every `CkEthSetupBuilder::build`".

**Rate-limiting experiment:** tried setting `canister_execution_rate_limiting: Disabled` for both modes and ran `deposit_from_cex` fresh (`--nocache_test_results`) — still passes, so this axis collapses too; no divergence left to explain.

PR description rewritten to match the current single-`build()` architecture (dropped the stale references to `with_live_mode`'s cycle-amount knob).

Verification (all synchronous, final pass `--nocache_test_results`):
- `rustfmt` — clean
- `cargo check --all-targets --all-features -p ic-cketh-test-utils -p ic-cketh-minter` — clean
- `cargo clippy` (pinned flags, `--all-targets`) — clean
- `bazel test` for `cketh_test`, `ckerc20_test`, `test_utils:lib_tests`, `deposit_from_cex` — all 4 PASSED

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,117 @@ impl CkEthSetup {
     }
 }
 
+struct CkEthSetupBuilder<'a> {
+    env: Option<Arc<PocketIc>>,
+    backend: EthereumBackend<'a>,
+    live: bool,
```
gregorydemay: 🧐 🔵 **Nit** (re-review, answering the question about this field) — the invariant `live ⇒ env.is_none()` is now *stated* in two places (the `with_env` doc, the new `debug_assert!`) but is still representable, so both statements have to be maintained by hand.

Swapping `live: bool` for a stored `mode: EnvMode` would be a lateral move: `build` would trade `if self.live` for `matches!(self.mode, EnvMode::Live)` at the two sites that pick the controller and skip the ledger install, and `env: Option<..>` would remain independently settable — the illegal pair survives. The shape that pays is collapsing *both* fields into one:

```rust
enum Env {
    /// Build a fresh mocked instance.
    New,
    /// Reuse a caller-supplied (mocked) instance, e.g. `CkErc20Setup`'s.
    Shared(Arc<PocketIc>),
    /// Build a fresh live instance: NNS subnet, live before any canister exists.
    Live,
}
```

`with_env`/`with_live_mode` then assign a variant, `EnvMode` and this field become one type, the `if self.live { EnvMode::Live } else { .. }` conversion in `build` disappears, and the combination the `debug_assert!` guards becomes unrepresentable rather than checked (R2/G27).

On the `debug_assert!` itself: it is genuinely live here — `bazel/conf/.bazelrc.build:46` forces `-Cdebug-assertions=on` even under `--compilation_mode=opt` — so it is not dead in CI, and as a fix for the previous nit it does the job. The only point is that with the enum it would not need to exist.

Non-gating: both knobs are private with one call site each.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,18 +723,117 @@ impl CkEthSetup {
...
-            ..Default::default()
-        })
-        .build()
+    new_env(EnvMode::Mocked)
+}
+
+/// A [`PocketIcBuilder`] with the fiduciary subnet every ckETH fixture needs for the secp256k1
+/// `key_1` used by the minter; callers add anything further (e.g. an NNS subnet for
+/// [`PocketIc::make_live`]) before `build()`.
+fn pocket_ic_builder() -> PocketIcBuilder {
```
gregorydemay: 🧐 🔵 **Nit** — `pocket_ic_builder()` has exactly one caller (`new_env`, L810), and it costs more than it saves: its doc's first clause restates `new_env`'s own first clause almost verbatim (the fiduciary subnet / secp256k1 `key_1` rationale is now written twice), and its second clause addresses "callers" in the plural that do not exist. Inlining `PocketIcBuilder::new().with_fiduciary_subnet()` into `new_env` removes the hop and leaves that rationale stated once (G12/C3). Purely a Boy-Scout item.

gregorydemay (commented):
🧐 **VERDICT: READY** — 0 blockers, 0 mediums, 2 nits (both new, both non-gating); CI green. All 9 findings from the previous round are genuinely resolved.

<details>

### Previous 9 findings

| # | Finding | Status |
|---|---|---|
| 🟠 1 | cycles rationale deleted | **Resolved** (`ad8ce46e58`) |
| 🟠 2 | ledger comment leads with the obstacle | **Resolved** (`5d619c334d`) |
| 🟠 3 | `register_supported_tokens` hand-rolls `add_ckerc20_token` | **Resolved** (`a114ffdec4`) |
| 🔵 4 | `new_env(bool)` flag argument | **Resolved** (`d000a55ab7`) |
| 🔵 5 | rate-limiting divergence unexplained | **Resolved by collapsing** (`501c22f0d7`) — verified, see below |
| 🔵 6 | `_gateway` binding | **Resolved** (`501c22f0d7`) |
| 🔵 7 | `with_env`/`with_live_mode` invariant unenforced | **Resolved** (`78aa809289`) |
| 🔵 8 | `with_live_mode` doc missing the ledger axis | **Resolved** (`5d619c334d`) |
| 🔵 9 | stale `CkEthCanisters` doc | **Resolved** (`986fd3886c`) |

Spot-checks on the three I flagged as needing evidence rather than inspection:

- **(1)** `CANISTER_CYCLES` carries the constraint accurately — `u128::MAX` crashes the live replica with `Invalid cycle change` on the minter's first HTTPS outcall because a canister at the saturating `Cycles` ceiling cannot observe a further addition — and is used at all three `add_cycles` sites, with no remaining literal.
- **(2)** `with_live_mode`'s doc now leads with *"the balance scan never calls it"* and demotes the Bazel data-dependency to a trailing clause; the install site keeps a one-line pointer.
- **(3)** `cketh.add_ckerc20_token(live_controller(), &arg)` is semantically identical to the code it replaced: same sender, same `Result<Vec<u8>, RejectResponse>`, same `.expect` on the reject. The reply stays undecoded exactly as before.

### Nit 5 — the behaviour change to the live instance, verified

`canister_execution_rate_limiting: Disabled` now applies to live mode too. Ran in-container, uncached (`--nocache_test_results`):

- `//rs/ethereum/cketh/minter:deposit_from_cex` — **4 uncached runs, all PASSED** (one standalone + `--runs_per_test=3`); 63.4s–70.5s, dev 2.9s. No flakiness signal.
- `integration_tests_tests/cketh_test` PASSED (73.0s), `integration_tests_tests/ckerc20_test` PASSED (138.2s), `test_utils:lib_tests` PASSED.

### Non-negotiables re-verified from the call graph

- **`make_live` unreachable from `CkEthSetup::new`/`Default`.** `make_live` appears once, inside `new_env` under `EnvMode::Live`. `EnvMode::Live` is produced only by `build()` when `self.live`, which only `with_live_mode()` sets, whose only caller is `LiveBalanceScanSetup::new_live()`. `Default` → `builder().build()` takes the `EnvMode::Mocked` path; `new(env)` short-circuits `new_env` entirely via `unwrap_or_else`. The cold-start property holds.
- **Mocked creation/install order matches `origin/master`'s `CkEthSetup::new`.** Minter created first (comment preserved), then ledger, then EVM RPC, `add_cycles` after each; installs ledger → EVM RPC → minter, sender `None` throughout. The single intentional delta is the cycle amount (`u128::MAX` → `u64::MAX`), which is the documented unification and is called out in the description.
- **`MINTER_ADDRESS` derivation untouched.** Same creation order and same `"key_1"` (`"key_1".parse().unwrap()` → `ECDSA_KEY_NAME.to_string()`, same value). `tests/cketh.rs:1186` asserts `minter_address() == MINTER_ADDRESS` and passes uncached.

- **Duplication:** none found. The last in-diff instance (finding 3) is gone. Structural duplication against the codebase: the sibling that motivated this PR (`CkErc20LiveScanSetup`) is deleted, not mirrored.
- **Unused derives:** cleared. `#[derive(Clone, Copy)]` on `EnvMode` is load-bearing — `new_env` matches `mode` by value twice, which would not compile without `Copy`. No other new derives.
- **Primitive-obsession parameters:** `new_env`'s bool parameter is gone. What remains is a private builder *field* (`live: bool`) — raised as a 🔵 below.
- **Divergent invariant handling:** none. The live/mocked fork is decided once in `build`; the `with_env`/`live` invariant has exactly one check.
- **Silent fallbacks:** none new. `self.env.unwrap_or_else(..)` is a builder default for an expected-absent input, not an invariant breach; `balance_scan_candidates`'s `.unwrap_or(0)` is pre-existing and documented.
- **Test-only code in production modules:** none — the diff is confined to `test_utils` plus one test file.

### Docs

All doc references resolve against the code; no leftovers from the `build_mocked`/`build_live` era remain. On the re-review question about length: `with_live_mode`'s doc is six lines covering four axes at roughly a clause each, and it is now the single definition of what live mode means — I judge that proportionate, not overgrown.

### PR description

Accurate against the code. The previously false claim about a cycle-amount knob "documented at the `add_cycles` call sites" is gone; every artefact the rewrite names (`CANISTER_CYCLES = u64::MAX as u128`, `new_env`/`EnvMode`, `create_cketh_canisters(env, controller: Option<Principal>)`, the rate-limiting unification, the reuse of `CkEthSetup::add_ckerc20_token`) matches what is in the tree, and the rate-limiting claim reproduces locally. Purpose-focused, no "Test plan" section. No 📚 stack section — and none to preserve: this PR bases directly on `master`, as does its sibling #11123.

### Testing

No new tests are owed: this is a fixture refactor whose behaviour is pinned by the ~70 existing integration tests that run through it, all of which pass uncached above.

</details>

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Agreed this is the cleaner shape — collapsing `env: Option<Arc<PocketIc>>` and `live: bool` into one three-variant enum would make the illegal pair unrepresentable and delete the `debug_assert!` rather than guard with it.

Not taking it in this PR. The change was asked for and reviewed as "remove the parallel live fixture", and that is done; reshaping the builder's own state is a different edit that would put the whole suite through another uncached verification round for a nit on code that is already correct. Recorded as a follow-up instead, so it is not lost.

Happy to be overruled — say so and it goes in.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fair — `pocket_ic_builder()` is a one-caller helper whose doc now restates `new_env`'s, so inlining it would lose nothing.

Leaving it for the same reason as the sibling nit: it is unrelated to this PR's purpose (removing the parallel live fixture) and would cost another full uncached verification round. Recorded as a follow-up.

gregorydemay: inlined

gregorydemay: 🤖 Done in 75f3d8e518 — inlined `PocketIcBuilder::new().with_fiduciary_subnet()` into `new_env` and deleted `pocket_ic_builder()` and its doc. Confirmed via `grep -rn pocket_ic_builder rs/` that it had no other caller or re-export anywhere in the tree.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -143,43 +145,23 @@ impl PocketIcHttpQuery for &CkEthSetup {
...
-            ))
-            .unwrap(),
-            None,
-        );
-        install_evm_rpc(&env, evm_rpc_id);
-        let minter_id = install_minter(&env, ledger_id, minter_id, evm_rpc_id);
+    /// Builds a fresh PocketIC instance (fiduciary subnet only, non-live, anonymous controller) and
+    /// installs the minter, its ckETH ledger and the EVM RPC canister on it against `backend`.
+    /// [`Default`] uses [`EthereumBackend::Mocked`]; [`live_scan`] passes
+    /// [`EthereumBackend::Anvil`] for the live balance-scan harness.
```
mbjorkqvist: nit: I assume the "anonymous controller" part references the installed canisters and not the PocketIC instance itself?

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -69,64 +66,30 @@ pub struct Holding {
...
     pub fn new_live() -> Self {
-        let anvil = Anvil::start();
-
-        let mut env = PocketIcBuilder::new()
-            .with_nns_subnet() // make_live requires an NNS subnet.
-            .with_fiduciary_subnet() // holds the secp256k1 `key_1` used by the minter.
-            .build();
+        let anvil = Arc::new(Anvil::start());
+        let cketh = CkEthSetup::new(EthereumBackend::Anvil(Arc::clone(&anvil)));
+        let ckerc20 = CkErc20Setup::with_cketh(cketh).add_supported_erc20_tokens();
```
mbjorkqvist, line 79: The hardcoded token list can now drift from what the fixture registers.

`add_supported_erc20_tokens()` registers whatever `ic_ledger_suite_orchestrator_test_utils::supported_erc20_tokens()` returns, but `SupportedToken::ALL` (and `credit_deposits`) still hardcode ckUSDC/ckUSDT. Before this PR the harness registered its own two tokens, so the two lists could not disagree.

If a third token is ever added to `supported_erc20_tokens()`, `credit_deposits` won't give it code — and per the comment on `credit_deposits` itself, "a token without code would revert the whole scan even for addresses that do not hold it". The symptom would be a 180s hang ending in `the deposit address was not scanned within 180s`, with nothing pointing at the cause.

Could `SupportedToken` derive its contracts from `self.ckerc20.supported_erc20_tokens` instead? That would also remove the duplicated `USDT_ERC20_CONTRACT_ADDRESS` constant. Failing that, asserting here that the two agree in length would at least fail loudly.

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -299,6 +299,10 @@ rust_test(
         "tests/deposit_from_cex_demo/MockUSDT.sol",
         # End-to-end balance scan on a live PocketIC + local anvil node.
         ":cketh_minter_debug.wasm.gz",
+        "//rs/ethereum/ledger-suite-orchestrator:ledger_suite_orchestrator_canister.wasm.gz",
```
mbjorkqvist, line 304: `size = "medium"` gives this test a 300s bazel timeout, and that budget now has to cover five extra canister installs that master's harness didn't do (orchestrator + 2 ledgers + 2 index canisters + archives) on top of PocketIC and anvil startup, while `deposit_from_cex.rs` still applies a 180s `deadline` per `await_scan`.

If we do end up having timeouts on CI and bumping to `size = "large"`, then we'd probably also need to mark the tests as long tests and not run them on PRs, but perhaps it makes sense to go with `medium` now and cross that bridge when we get there?

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +778,137 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
     evm_rpc_id: Principal,
-) -> Principal {
+}
+
+/// Cycles every canister this fixture creates is funded with. `u128::MAX` — the natural "as much as
+/// possible" choice — reproducibly crashes the live harness' PocketIC replica with a
+/// cycle-accounting assertion failure (`Invalid cycle change`) on the minter's first HTTPS outcall:
+/// a canister already at the saturating `Cycles` balance ceiling cannot observe any further
+/// addition. This amount leaves headroom below that ceiling instead.
+const CANISTER_CYCLES: u128 = u64::MAX as u128;
```
mbjorkqvist, line 797: The rationale here is specific to the live harness, but the constant applies to every ckETH fixture — the mocked ones go from `u128::MAX` to `u64::MAX` too.

That looks harmless (I couldn't find a cketh test that asserts on cycle balances), but it does make the PR description's "Behaviour of the mocked fixtures is unchanged" inaccurate. Either scope the lower amount to `EthereumBackend::Anvil`, or say in the doc that funding every fixture the same way is the point.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +778,137 @@ fn evm_rpc_wasm() -> Vec<u8> {
     )
 }
 
-fn install_minter(
-    env: &PocketIc,
-    ledger_id: Principal,
+/// The minter, its ckETH ledger and the EVM RPC canister it calls out to. Built by
+/// [`create_cketh_canisters`] and installed by
+/// [`install_ledger`]/[`install_minter`]/[`install_evm_rpc`], identically for every backend.
```
mbjorkqvist: nit: "identically for every backend" isn't quite right — `install_minter` and `install_evm_rpc` both take `&backend` and produce different init args (override provider, block tag, last scraped block). What is identical is the set of canisters, the creation order, and the controller. Perhaps "…installed in the same order for every backend; only the init args differ"?

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -791,36 +778,137 @@ fn evm_rpc_wasm() -> Vec<u8> {
...
+}
+
+/// The Ethereum chain under test: which JSON-RPC endpoint the EVM RPC canister's outcalls reach,
+/// and the corresponding minter init/upgrade assumptions about that chain's state (the block height
+/// to track, and where its log-scraping cursor starts).
+enum EthereumBackend {
+    /// Canned JSON-RPC mocks pinned to a historical mainnet snapshot.
+    Mocked,
+    /// A live anvil node, reached over HTTP at its own URL: a fresh chain with no finalized blocks
+    /// yet. Shared with the harness that started the node and keeps it running, hence the [`Arc`].
```
mbjorkqvist: Two small things on this doc:

- "Shared with the harness that started the node and keeps it running, hence the [`Arc`]" — the sharing doesn't outlive construction. `CkEthSetup::new` takes `backend` by value and doesn't store it, so this `Arc` clone is dropped when `new` returns and `LiveBalanceScanSetup` is the sole owner from then on. A `&Anvil` would give the same "a node is really running behind that URL" guarantee, at the cost of a lifetime parameter.
- "created and installed exactly as for `EthereumBackend::Mocked`" — same as the `CkEthCanisters` doc above: the init args do differ per backend.

rs/ethereum/cketh/test_utils/src/live_scan.rs
```diff
@@ -1,42 +1,39 @@
...
-//! IC's HTTPS-outcalls feature, and installs it with an `overrideProvider` that rewrites every
-//! provider URL to the harness' anvil node (reached over HTTP, mirroring the `evm_rpc_local`
-//! configuration of the EVM RPC canister). The minter therefore reads real Ethereum state from
-//! anvil, exercising the balance scan end to end: minter → EVM RPC canister → anvil.
+//! The whole fixture — minter, EVM RPC canister, ledger-suite-orchestrator, and the ckUSDC/ckUSDT
+//! ledger and index canisters the orchestrator spawns — is built on an ordinary (non-live) PocketIC
+//! instance, exactly as [`crate::ckerc20::CkErc20Setup`]'s mocked fixture is: `await_call` ticks
+//! deterministically, and every setup call completes in a bounded number of rounds. Only once
+//! construction is complete does [`LiveBalanceScanSetup::new_live`] switch the instance to
+//! auto-progress, so the EVM RPC canister's outcalls reach anvil for real from that point on.
```
mbjorkqvist, line 11: The comment this replaces — "Go live *before* installing the minter: its install schedules immediate refresh and balance-scan timers that issue HTTPS outcalls, which would stall (holding the task guards) if they fired while the outcalls could not be answered" — documented a hazard that the new order walks straight into, and the module doc doesn't say why that's now fine.

It is fine, as far as I can tell: `ProcessCanisterHttpInternal` re-scans `canister_http_request_contexts()` on every auto-progress round and filters only on already-dispatched request ids (`rs/pocket_ic_server/src/pocket_ic.rs:3792-3812`), so outcalls created while the instance was still non-live do get dispatched once `auto_progress()` starts. That's non-obvious enough to be worth a sentence here, since it's the thing that makes "build first, go live last" safe.

rs/ethereum/cketh/test_utils/src/ckerc20.rs
```diff
@@ -84,8 +90,14 @@ impl CkErc20Setup {
         ckerc20
     }
 
-    pub fn new_without_ckerc20_active(env: Arc<PocketIc>) -> Self {
-        let cketh = CkEthSetup::new(env.clone());
+    pub fn new_without_ckerc20_active() -> Self {
```
mbjorkqvist, line 93: nit: five entry points now, where `new()` and the `Default` impl are identical, and `new_without_ckerc20_active()` differs from `without_ckerc20_active()` only by a `CkEthSetup::default()`. Not worth churn on its own, but it could collapse while you're here.

Also on the doc just below: "its Ethereum backend decides how" reads like the backend still influences how the instance is built. After 7a5682b7d, `new_env()` is unconditional — the backend only decides the minter / EVM RPC init args.

rs/ethereum/cketh/test_utils/src/lib.rs
```diff
@@ -747,11 +729,12 @@ impl CkEthSetup {
...
 
-pub fn format_ethereum_address_to_eip_55(address: &str) -> String {
-    Address::from_str(address).unwrap().to_string()
-}
-
-pub fn new_pocket_ic() -> PocketIc {
+/// Builds the PocketIC instance for [`CkEthSetup::new`]: the fiduciary subnet every ckETH fixture
+/// needs for the secp256k1 `key_1` used by the minter. Always a non-live (manual-round) instance,
+/// even for [`EthereumBackend::Anvil`]: [`crate::live_scan`] builds its whole fixture here first and
+/// only switches to live outcalls once construction is complete, so `await_call` ticks
```
mbjorkqvist, line 735: This doc is accurate, but it now contradicts the PR description's first bullet: "The chain under test is the only thing that decides how the test environment is built. Whether PocketIC runs live follows from it…". After 7a5682b7d that's no longer the case — `new_env()` is unconditionally non-live for both backends, and liveness is flipped by `LiveBalanceScanSetup::new_live` calling `auto_progress()` once construction is done. Worth rewording the description before this lands, since that's the artifact that survives.

Same for "Behaviour of the mocked fixtures is unchanged" — see the `CANISTER_CYCLES` comment below.

(General note, since the review body can't be edited after the fact: the duplication removal itself reads as a clear win. `live_scan.rs` drops ~190 lines of parallel fixture and keeps only the anvil node and the scan assertions, canister creation order is preserved so the hardcoded minter address stays valid, and the Bazel env vars and `_u256` wasm variants match the existing `integration_tests` target exactly. The two comments I'd actually want addressed are the `SupportedToken::ALL` drift and the `size = "medium"` headroom; the rest are accuracy fixes.)

mbjorkqvist (approved):
Thanks for the cleanup, @gregorydemay!

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 31c0c76045 — reworded: the anonymous controller applies to the installed canisters, not the PocketIC instance itself.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in 407406798d — went with Greg's call (supersedes my "assert the lengths agree" suggestion): deleted `SupportedToken` entirely and pass the orchestrator's own `Erc20Token` through `Holding`/`deposit_erc20`/`register_deposit_address`/`await_scan`. A new `supported_erc20_tokens()` accessor exposes the registered list, and `credit_deposits` iterates it instead of a hardcoded const, so the set that gets code on anvil IS the registered set, by construction. `USDT_ERC20_CONTRACT_ADDRESS` is gone with it. Verified the EIP-55 case empirically rather than assuming: comparing `detected.erc20_contract_address` directly against `token.contract.address` (no re-normalizing) — the live suite passes 10/10 under contention.

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: 🤖 Agreed, staying with `medium` — no BUILD.bazel change. Measured after this round's changes: `deposit_from_cex` alone is ~72–73s (uncached, this environment) against the 300s cap, and 10/10 pass under `--runs_per_test=5 --local_test_jobs=5` contention across two full runs. `Bazel Test All on RBE` also passed with this test included on 3cb7de4fc9.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 31c0c76045 — went with your second option: the doc now says funding every fixture the same amount is deliberate, rather than scoping the lower amount back to `EthereumBackend::Anvil`.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 31c0c76045, using close to your suggested wording: "...installed by [...] in the same order for every backend; only the init args differ."

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Fixed in 31c0c76045 — kept the `Arc` (agreed a `&Anvil` lifetime parameter costs more than the wrong sentence) and fixed both: the doc now says this particular clone is dropped once `CkEthSetup::new` returns (the harness' own separate clone is what keeps anvil running), and that only the init args differ from `Mocked`, not the create/install sequence.

rs/ethereum/cketh/test_utils/src/live_scan.rs
gregorydemay: 🤖 Fixed in 31c0c76045 — added a paragraph to the module doc. Verified `ProcessCanisterHttpInternal` does what you describe by reading `rs/pocket_ic_server/src/pocket_ic.rs:3792-3812`: it re-scans `canister_http_request_contexts()` every round and dispatches whichever request ids aren't already in `canister_http.pending`, so outcalls issued while non-live are picked up for real on auto-progress's first round rather than dropped.

rs/ethereum/cketh/test_utils/src/ckerc20.rs
gregorydemay: 🤖 Doc half fixed in 31c0c76045 ("its Ethereum backend decides how" → the backend only picks init args now). Declining the five-entry-point collapse as out of scope for this PR, as you say it's optional — noted as a follow-up.

rs/ethereum/cketh/test_utils/src/lib.rs
gregorydemay: 🤖 Rewrote both. The PR description no longer claims the chain under test decides whether PocketIC runs live (it now describes construction-then-`auto_progress()`, matching 7a5682b7d5), and no longer claims mocked-fixture behaviour is fully unchanged — it now calls out the `u128::MAX`→`u64::MAX` cycles change explicitly, same as the `CANISTER_CYCLES` doc fix. Thanks for the summary note — appreciated.

gregorydemay:
🤖 Addressed all 9 open threads, in two commits:

- `31c0c76045` — doc accuracy fixes: `CkEthSetup::new` (anonymous controller scope), `CkEthCanisters` ("identically for every backend" → same order, different init args), `CANISTER_CYCLES` (states funding every fixture the same amount is deliberate), `EthereumBackend::Anvil` (Arc lifetime + init-args wording), `ckerc20.rs` ("its Ethereum backend decides how"), and the `live_scan.rs` module doc (added the missing "why build-first-go-live-last is safe" paragraph, mechanism verified by reading `rs/pocket_ic_server/src/pocket_ic.rs:3792-3812`).
- `407406798d` — the substantive fix: deleted `SupportedToken` entirely per Greg's call (supersedes my "assert lengths agree" suggestion). `Holding`, `deposit_erc20`, `register_deposit_address`, `await_scan` now take the orchestrator's own `Erc20Token` (the lifetime approach worked cleanly at every call site, no cloning needed), and a new `supported_erc20_tokens()` accessor plus `let [usdc, usdt] = ... else { panic!(..) }` in the test give the loud arity check. Verified the EIP-55 comparison empirically rather than assuming — direct string comparison against `token.contract.address` works, confirmed by the live suite passing 10/10 under contention.

`BUILD.bazel` `size = "medium"` — no change, replied with the measured numbers (~72–73s solo, 10/10 under contention across two runs, RBE passed on 3cb7de4fc9). `CkErc20Setup`'s five-entry-point collapse — declined as out of scope, noted as a follow-up.

PR description rewritten: no longer claims the chain under test decides liveness (it's decided after construction via `auto_progress()`), and no longer claims mocked-fixture behaviour is fully unchanged (calls out the `u128::MAX`→`u64::MAX` cycles change).

Verification (all synchronous, uncached): rustfmt, `cargo check --all-targets --all-features`, clippy (pinned flags + `--all-targets`) all clean. `cketh_test` 27/27, `ckerc20_test` 42/42, `test_utils:lib_tests` pass. `deposit_from_cex` under `--runs_per_test=5 --local_test_jobs=5`: 10/10 across two full runs, ~72s avg each.

PR #11128, opened 2026-08-13
feat(cketh): per-(deposit address, ERC-20) deposit granularity

gregorydemay:
## Why

Automatic ERC-20 deposit detection was keyed at the granularity of a **deposit address**: one `deposit_erc20` call armed an address, and the balance scan read every armed address against *every* supported ERC-20 in gas-capped batches. That per-address unit is the source of most of the machinery (keeping an address's per-token calls in one batch, advancing a whole address, dropping all of an address's tokens the moment any one funds) and a status wart (`AwaitingSweep` lumping every token together).

This PR makes the natural finer unit — **(deposit address, ERC-20 token)** — the unit of registration, scanning, status, and sweeping, which is also the standard CEX deposit UX: pick the token, the network is always Ethereum, then show the (shared) deposit address.

## Nothing here is live yet

The automatic-deposit feature as a whole — `deposit_erc20` and the balance scan behind it — has never shipped to mainnet. No deployed minter exposes the endpoint, so **no production event log contains any of its events** and no client calls it.

That is what makes the rest of this PR cheap, and it is worth keeping in mind while reviewing:

- The Candid changes are marked breaking, but they break no deployed caller.
- Event payloads are reshaped and their fields renamed **in place**, with no versioning or migration, because there is no recorded history to stay compatible with.
- Where an encoding *is* preserved (the newtype and field renames below leave the CBOR bytes untouched), that is for the sake of the replayable mainnet/Sepolia event-log tests, not for production data.

## What

- `deposit_erc20` takes an explicit `erc20_contract_address`, required to be a minter-supported ckERC20. The deposit address stays shared across a caller's tokens.
- Registration / watchlist / scan schedule / sweep queue are keyed by `(account, token)`; funding one token no longer removes its siblings. Per-account cap of 5 concurrently armed tokens.
- The balance scan reads only the registered `(address, token)` pairs — the cartesian batching (split-assert, whole-address advance) is gone; each pair is one `balanceOf` call.
- `AwaitingSweep` now carries a single detected deposit; the automatic-deposit events carry one entry per token.
- The `deposit_from_cex.md` spec is updated to match.

## Review follow-ups

- Every Candid field naming an ERC-20 contract is now `erc20_contract_address` rather than `token`, matching the existing `deposit_erc20` argument and the minted-token event. The internal event payloads follow suit, so the two layers no longer disagree.
- `deposit_erc20`'s failure handling now matches `withdraw_erc20`: an unparseable contract address traps, and an unsupported token returns the list of tokens the minter does support instead of an empty error.
- A deposit address derived by the minter and an ERC-20 contract address were both plain Ethereum addresses, so every place the deposit flow handled both was a silent-swap hazard that still compiled — including the balance-scan call where transposing them would have queried the wrong account and quietly read zero. The derived address now has its own type, which makes those transpositions compile errors.
- The `(account, token)` pair and the scan bookkeeping stored against it were named the wrong way round; they now say what they are, mirroring `WithdrawalRequest` on the withdrawal side.

Rebased onto master now that #10946 has merged. Tracks DEFI-2976.


rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -150,55 +169,56 @@ impl AutomaticDeposits {
...
-            });
-        }
+                scanned_balance: deposit.scanned_balance,
+            },
+        );
+        assert!(
+            previous.is_none(),
+            "BUG: sweep queue already has an entry for account {account:?} token {}",
+            deposit.token
+        );
```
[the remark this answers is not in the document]
gregorydemay: 🤖 On the IC a trap rolls back every state mutation of the current message (and aborts `post_upgrade` replay entirely), so the `insert` preceding the `assert!` is never observable — the overwrite is undone together with the trap. The method is therefore already "trap without mutating" in every context it runs in, which is exactly the invariant the `# Panics` doc relies on. Leaving as-is.

rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -274,6 +276,12 @@ pub struct DetectedDeposit {
 
 #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub enum DepositErc20Error {
+    /// The `erc20_contract_address` could not be parsed as an Ethereum address.
+    InvalidErc20ContractAddress(String),
+    /// The `erc20_contract_address` is not a ckERC20 token supported by the minter.
+    UnsupportedCkErc20Token,
+    /// The account already has the maximum number of ERC-20 tokens armed.
+    TooManyTokensForAccount,
     /// The maximum number of concurrently armed deposit addresses has been reached.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch — reworded in ec41407a1a. The global cap counts armed `(account, token)` pairs (the watchlist is keyed by `DepositKey`), and both the Rust doc and the `.did` comment now say so.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1114,8 +1132,11 @@ at $2'500 and the sweep-gas figures from the
...
 
-**Detection schedule** — per armed address, backing off over the 24h scanning window of
-`R15`; the basis for the outcall counts below. Scheduling is **block-based, not wall-clock**:
+**Detection schedule** — per armed `(address, token)` pair (each pair is one
+`balanceOf` sub-call), backing off over the 24h scanning window of
+`R15`; the basis for the outcall counts below. The per-account cap (≤ 5 armed
+tokens) bounds an account's contribution to this schedule. Scheduling is
+**block-based, not wall-clock**:
 elapsed time is measured as `elapsed_blocks × SECS_PER_BLOCK` (≈ 12 s/block) against a
 33-entry gap table (`SCAN_GAP_SECS` in `automatic_deposits/mod.rs::addresses_to_scan_iter`).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in ec41407a1a — updated the reference to `scan_targets_iter`.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -394,29 +394,42 @@ unique, deterministic deposit address, derived from the minter's threshold-ECDSA
   funds at a deposit address never depend on contract code — even without
   EIP-7702, any balance is recoverable by funding the address with gas and signing
   a normal transfer.
-* Endpoint `deposit_erc20(account, fee?) -> String` (EIP-55 checksummed).
+* Endpoint `deposit_erc20({ erc20_contract_address, account, fee? }) -> { address, status }`
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in ec41407a1a — the spec now shows `deposit_erc20({ erc20_contract_address, mode }) -> { address, status }`, and notes the depositing account is the caller principal plus the optional subaccount carried in `mode` (the future `Sponsored` `mode` variant carries the fee).

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -135,45 +122,42 @@ async fn scan<R: Runtime>(
...
-            // A funded address is event-sourced into the sweep queue (durable the moment funds are
+            };
+            // A funded pair is event-sourced into the sweep queue (durable the moment funds are
             // detected) instead of being advanced, so it is no longer re-scanned. Its watchlist
-            // entry gives the derived address and the scan count (this finding scan included); skip
-            // if the account is no longer live as of `now`.
-            let Some(entry) = s.automatic_deposits.get_entry(now, account) else {
+            // entry gives the scan count (this finding scan included); skip if the pair is no
+            // longer live as of `now`.
+            let Some(entry) = s.automatic_deposits.get_entry(now, &target.key()) else {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 This is now covered after the refactor. `scan_balances` returns `ScanOutcome::Detected(AutomaticDeposit { owner, subaccount, address, token, last_scanned_block, scan_count, scanned_balance })`, asserted field-for-field in `balance_scan/tests.rs::should_detect_a_funded_pair_from_pre_scan_targets_even_after_eviction` (which also proves the pre-scan target is the sole source, so a mid-scan eviction cannot drop the detection). Watchlist removal + `sweep_len()` increment on a funded pair are covered in `automatic_deposits/tests.rs::record_automatic_deposit_received_removes_the_pair_and_queues_it`, and the end-to-end sweep-queue move (event fields included) by `ckerc20.rs::should_move_funded_addresses_to_the_sweep_queue`.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -656,6 +666,7 @@ type Event = record {
             registrations : vec record {
                 owner : principal;
                 subaccount : opt Subaccount;
+                token : text;
```
gregorydemay: @claude : let's rename this field to erc20_contract_address

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -666,9 +677,10 @@ type Event = record {
             owner : principal;
             subaccount : opt Subaccount;
             address : text;
+            token : text;
```
gregorydemay: @claude : same, rename to `erc20_contract_address`

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -188,24 +199,21 @@ async fn deposit_erc20(arg: DepositErc20Arg) -> Result<DepositErc20Response, Dep
...
-        address: entry.value.address.to_string(),
-        valid_until: entry.expires_at.as_nanos(),
-        last_scanned_block: entry.value.last_scanned_block.map(|block| block.into()),
-        scan_count: entry.value.scan_count as u64,
-    }
+    mutate_state(|s| s.register_deposit_address(now, account, token))?;
+    Ok(
+        read_state(|s| s.automatic_deposits.deposit_status(now, &account, token))
+            .expect("BUG: a just-registered pair must report a Scanning status"),
+    )
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch on the race — fixed in a0be3de03a and 5accc99ca5. Two corrections to the mechanism, though.

**The `expect` cannot trap.** It unwraps an `Option`; it does not check for `Scanning`, despite what its message says. On the racy path `deposit_status` returns `Some(AwaitingSweep)`, which unwraps fine. And there is no await between `register_deposit_address` and that `read_state`, so `None` is unreachable there.

**The real failure is deferred and lands elsewhere.** `watch_deposit` never consulted the sweep queue, so the pre-await `deposit_status` check was the only thing keeping a pair out of both maps. When the await breaks that atomicity, the re-armed pair ends up in the watchlist *and* the sweep queue — and the `deposit_erc20` call still returns `AwaitingSweep`, looking perfectly healthy. The damage surfaces later, when a scan re-detects the still-unswept funds and trips `assert!(previous.is_none())` in `record_automatic_deposit_received`. Because `process_event` applies before recording, that trap rolls the message back, so the event log is *not* poisoned and upgrades are unaffected — but the scan tick traps and keeps trapping every interval, stalling detection for all users until the duplicate entry expires at 24h.

**Fix taken.** Rather than re-checking after the await, `deposit_erc20` now takes a per-principal guard for the whole call, as `withdraw_erc20` already does. Arming a pair requires a `deposit_erc20` call from that account's owner, so serializing per principal removes the only way the pair can change underneath a suspended call — closing the race at its cause rather than narrowing the window. `watch_deposit` also asserts the invariant now, as a backstop against a future caller reopening it.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -179,6 +179,26 @@ async fn minter_address() -> String {
 async fn deposit_erc20(arg: DepositErc20Arg) -> Result<DepositErc20Response, DepositErc20Error> {
     validate_ckerc20_active();
     let caller = validate_caller_not_anonymous();
+    // Held for the whole call, including across the ECDSA public key fetch below, so that the
+    // status check and the registration that follows it cannot be interleaved with another
+    // `deposit_erc20` from the same principal.
+    let _guard = deposit_erc20_guard(caller).unwrap_or_else(|e| {
```
mbjorkqvist, line 185: Parallel `deposit_erc20` calls from the same principal now trap rather than returning an error. Would it make sense to return e.g., `DepositErc20Error::TemporarilyUnavailable` in case obtaining the guard fails with `AlreadyProcessing`?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -36,55 +35,81 @@ const SECS_PER_BLOCK: u64 = 12;
...
         &mut self,
         now: Timestamp,
         account: Account,
-        address: Address,
-    ) -> Result<Entry<DepositRequest>, DepositErc20Error> {
+        token: Address,
+        address: DepositAddress,
+    ) -> Result<Entry<ScanProgress>, DepositErc20Error> {
+        let request = DepositRequest::new(account, token);
+        assert!(
```
mbjorkqvist, line 81: nit: There does seem to be a way to reach this `assert` that is not strictly a bug:

1. The pair is live when the balance scan captures its `ScanTarget`s.
2. The entry expires.
3. `deposit_erc20` sees `deposit_status == None` (expired, so no early return) and suspends on the ECDSA fetch (meaning the canister would have had to have been upgraded so that the ECDSA key is not cached?).
4. The scan applies `Detected` and the pair enters `sweep`.
5. `watch_deposit` asserts, and the user's call traps.

In this case, the outcome would be a rolled-back trap that self-heals (the retry returns `AwaitingSweep`).

rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -256,25 +259,29 @@ pub enum DepositStatus {
...
 #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub enum DepositErc20Error {
-    /// The maximum number of concurrently armed deposit addresses has been reached.
+    /// The `erc20_contract_address` is not a ckERC20 token supported by the minter.
+    TokenNotSupported { supported_tokens: Vec<CkErc20Token> },
+    /// The account already has the maximum number of ERC-20 tokens armed.
+    TooManyTokensForAccount,
+    /// The maximum number of concurrently armed deposits (`(account, token)` pairs) has been
+    /// reached.
     TooManyActiveAddresses,
```
mbjorkqvist: nit: Since `MAX_ACTIVE_DEPOSIT_ADDRESSES` was renamed to `MAX_ACTIVE_DEPOSITS`, would it make sense to rename `TooManyActiveAddresses` e.g., to `TooManyActiveDeposits` accordingly?

rs/ethereum/cketh/minter/src/balance_scan/tests.rs
```diff
@@ -415,120 +211,80 @@ async fn should_not_advance_addresses_when_the_chunk_fails() {
...
 
-    let outcomes = scan_balances(
-        &due,
-        &[token],
-        latest,
-        stub_client(vec![ok_balances(&[min, below_min])]),
-    )
-    .await;
+    // Capture the due targets, then wipe the watchlist — as a concurrent arming/expiry could between
+    // reading the targets and applying the scan's outcome.
```
mbjorkqvist: nit: Is there a missing word here?
```suggestion
    // Capture the due targets, then wipe the watchlist — as a concurrent arming/expiry could evict between
    // reading the targets and applying the scan's outcome.
```

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -8,18 +8,15 @@ use crate::guard::TimerGuard;
...
 use crate::state::{TaskType, mutate_state, read_state};
 use crate::timed_sized_map::Timestamp;
 use batcher::BalanceOfCall;
 use evm_rpc_client::{CandidResponseConverter, DoubleCycles, EvmRpcClient};
 use ic_canister_log::log;
 use ic_canister_runtime::Runtime;
 use ic_ethereum_types::Address;
-use icrc_ledger_types::icrc1::account::Account;
-use std::collections::BTreeMap;
-use std::slice::Chunks;
```
mbjorkqvist, line 22: (Commenting here since I can't add a comment further down in the file):

The doc (in particular, the last paragraph) for `MAX_CALLS_PER_BATCH` seems out of date now, with the addition of `scan_balances`, which caps the chunk size at 1000.

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -211,7 +254,7 @@ mod deposit_erc20 {
         );
 
         let (ckerc20, response2) = ckerc20
-            .call_minter_deposit_erc20(caller, Some([42_u8; 32]))
+            .call_minter_deposit_erc20(caller, Some([42_u8; 32]), token.clone())
```
mbjorkqvist: Nit:
```suggestion
            .call_minter_deposit_erc20(caller, Some(DEFAULT_USER_SUBACCOUNT), token.clone())
```
Since `DEFAULT_USER_SUBACCOUNT` is the literal `[42_u8; 32]`, would it make sense to pass the former here?

mbjorkqvist (approved):
Thanks @gregorydemay, just a few minor comments!

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 `deposit_erc20` traps here to stay consistent with `withdraw_erc20`/`withdraw_eth`, which both `ic_cdk::trap` on guard contention — exactly what the earlier "align deposit_erc20 error handling with withdraw_erc20" commit set up. Returning `TemporarilyUnavailable` would make `deposit_erc20` diverge from those (released) withdraw endpoints, so I've kept the trap here to preserve the alignment. Happy to switch all three endpoints to the graceful path in a separate PR if you'd prefer that everywhere.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Good catch. Fixed in 2479aa1: `deposit_erc20` now re-checks `deposit_status` after the ECDSA `await` and returns the pair's status if a concurrent scan queued it for sweeping. From there the call is synchronous, so no scan can interleave before the registration — the assert is now a genuine invariant.

rs/ethereum/cketh/minter/src/endpoints.rs
gregorydemay: 🤖 Done in 9dab822 — renamed to `TooManyActiveDeposits`.

rs/ethereum/cketh/minter/src/balance_scan/tests.rs
gregorydemay: 🤖 Done in 9dab822 — added the missing "evict".

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Done in 9dab822 — refreshed the last paragraph to describe `scan_balances`' per-chunk all-or-nothing behaviour, replacing the stale `balance_scan`/"addresses" wording.

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Done in 9dab822 — now using `DEFAULT_USER_SUBACCOUNT`. Applied it to the two other `[42u8; 32]` sites for the same default account as well, and left the `[7u8; 32]` "other account" as-is.

PR #11129, opened 2026-08-13
feat: Do not route canister messages to a cooling down subnet

mraszyk:
Messages (requests and responses) in canister output queues destined for a subnet that is cooling down are no longer routed into the stream to it. They are retained in the output queue until the destination subnet stops cooling down, rather than being rejected or dropped. This includes the loopback stream, i.e. a cooling down subnet does not route its canisters' messages to its loopback stream either.

### DIFF
--- a/rs/messaging/src/routing/stream_builder.rs
+++ b/rs/messaging/src/routing/stream_builder.rs
@@ -14,7 +14,7 @@ use ic_types::messages::{
     MAX_INTER_CANISTER_PAYLOAD_IN_BYTES, MAX_REJECT_MESSAGE_LEN_BYTES, NO_DEADLINE, Payload,
     RejectContext, Request, RequestOrResponse, Response, StreamMessage,
 };
-use ic_types::{CountBytes, SubnetId};
+use ic_types::{CanisterId, CountBytes, SubnetId};
 use ic_types_cycles::{CompoundCycles, Cycles};
 #[cfg(test)]
 use mockall::automock;
@@ -75,6 +75,7 @@ const LABEL_VALUE_STATUS_SUCCESS: &str = "success";
 const LABEL_VALUE_STATUS_CANISTER_NOT_FOUND: &str = "canister_not_found";
 const LABEL_VALUE_STATUS_PAYLOAD_TOO_LARGE: &str = "payload_too_large";
 const LABEL_VALUE_STATUS_ENGINE_NOT_ALLOWED: &str = "engine_not_allowed";
+const LABEL_VALUE_STATUS_RETAINED_COOLING_DOWN: &str = "retained_cooling_down";
 
 const CRITICAL_ERROR_PAYLOAD_TOO_LARGE: &str = "mr_stream_builder_payload_too_large";
 const CRITICAL_ERROR_RESPONSE_DESTINATION_NOT_FOUND: &str =
@@ -442,6 +443,10 @@ impl StreamBuilderImpl {
         let refund_limit = self.max_stream_messages / 2;
         self.route_refunds(&mut state, refund_limit, &network_topology, &mut streams);
 
+        // No canister can have the subnet's own principal as its canister ID, so this
+        // identifies the messages taken from the subnet's own output queues.
+        let own_subnet_as_canister_id = CanisterId::from(self.subnet_id);
+
         let mut requests_to_reject = Vec::new();
         let mut oversized_requests = Vec::new();
         let mut engine_requests_to_reject: Vec<Arc<Request>> = Vec::new();
@@ -458,11 +463,24 @@ impl StreamBuilderImpl {
             // Cheap to clone, `RequestOrResponse` wraps `Arcs`.
             let msg = msg.clone();
 
+            let is_from_subnet_queues = msg.sender() == own_subnet_as_canister_id;
+
             match network_topology.route(msg.receiver().get()) {
                 // Destination subnet found.
                 Some(dst_subnet_id) => {
-                    let dst_stream_entry = streams.entry(dst_subnet_id);
                     let is_loopback_stream = self.subnet_id == dst_subnet_id;
+
+                    // A cooling down destination subnet must not be sent any messages
+                    // from canister output queues. Retain the message (along with
+                    // everything behind it in the same queue) until the destination stops
+                    // cooling down, rather than rejecting or dropping it.
+                    if !is_from_subnet_queues && network_topology.is_cooling_down(&dst_subnet_id) {
+                        self.observe_message_status(&msg, LABEL_VALUE_STATUS_RETAINED_COOLING_DOWN);
+                        output_iter.exclude_queue();
+                        continue;
+                    }
+
+                    let dst_stream_entry = streams.entry(dst_subnet_id);
                     if !is_loopback_stream
                         && is_at_limit(
                             &dst_stream_entry,
--- a/rs/messaging/src/routing/stream_builder.rs
+++ b/rs/messaging/src/routing/stream_builder.rs
@@ -42,6 +42,9 @@ struct StreamBuilderMetrics {
     pub routed_payload_sizes: Histogram,
     /// Misrouted messages currently in streams, by remote subnet.
     pub stream_misrouted_messages: IntGaugeVec,
+    /// Canister output queues skipped because their destination subnet was
+    /// cooling down.
+    pub cooling_down_skipped_queues: IntCounter,
     /// Critical error for payloads above the maximum supported size.
     pub critical_error_payload_too_large: IntCounter,
     /// Critical error for responses dropped due to destination not found.
@@ -63,6 +66,7 @@ const METRIC_SIGNALS_END: &str = "mr_signals_end";
 const METRIC_ROUTED_MESSAGES: &str = "mr_routed_message_count";
 const METRIC_ROUTED_PAYLOAD_SIZES: &str = "mr_routed_payload_size_bytes";
 const METRIC_STREAM_MISROUTED_MESSAGES: &str = "mr_stream_misrouted_messages";
+const METRIC_COOLING_DOWN_SKIPPED_QUEUES: &str = "mr_cooling_down_skipped_queues";
 
 const LABEL_TYPE: &str = "type";
 const LABEL_STATUS: &str = "status";
@@ -75,7 +79,6 @@ const LABEL_VALUE_STATUS_SUCCESS: &str = "success";
 const LABEL_VALUE_STATUS_CANISTER_NOT_FOUND: &str = "canister_not_found";
 const LABEL_VALUE_STATUS_PAYLOAD_TOO_LARGE: &str = "payload_too_large";
 const LABEL_VALUE_STATUS_ENGINE_NOT_ALLOWED: &str = "engine_not_allowed";
-const LABEL_VALUE_STATUS_RETAINED_COOLING_DOWN: &str = "retained_cooling_down";
 
 const CRITICAL_ERROR_PAYLOAD_TOO_LARGE: &str = "mr_stream_builder_payload_too_large";
 const CRITICAL_ERROR_RESPONSE_DESTINATION_NOT_FOUND: &str =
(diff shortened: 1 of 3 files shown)


rs/replicated_state/src/metadata_state.rs
```diff
@@ -427,8 +427,14 @@ pub struct SubnetTopology {
...
-    /// Whether the subnet is "cooling down", i.e. quiescing: it inducts no ingress
-    /// messages.
+    /// Whether the subnet is "cooling down", i.e. quiescing. While a subnet is
+    /// cooling down:
+    ///
+    ///  * it inducts no ingress messages;
+    ///  * the messages in canister output queues destined for it are not routed
+    ///    into the outgoing stream to it -- not even on the cooling down subnet
+    ///    itself, into its loopback stream -- but retained in the output queue they
+    ///    came from until it stops cooling down.
```
alin-at-dfinity: It may be useful to give a very brief indication of why we're doing each of these things:

```suggestion
    /// Whether the subnet is "cooling down", i.e. quiescing. While a subnet is
    /// cooling down:
    ///
    ///  * it inducts no ingress messages, so the ingress history becomes free of
    ///    expiring message statuses;
    ///  * (on all subnets) no messages are routed to a cooling down subnet --
    ///    including into the cooling down subnet's own loopback stream -- but
    ///    retained in their respective output queues, so that the respective
    ///    streams can be emptied.
```

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -458,11 +463,24 @@ impl StreamBuilderImpl {
             // Cheap to clone, `RequestOrResponse` wraps `Arcs`.
             let msg = msg.clone();
 
+            let is_from_subnet_queues = msg.sender() == own_subnet_as_canister_id;
```
alin-at-dfinity, line 476: `is_subnet_response`?

It might also make the intent clearer: we must still route/deliver subnet responses, just not other messages.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -458,11 +463,24 @@ impl StreamBuilderImpl {
...
                 Some(dst_subnet_id) => {
-                    let dst_stream_entry = streams.entry(dst_subnet_id);
                     let is_loopback_stream = self.subnet_id == dst_subnet_id;
+
+                    // A cooling down destination subnet must not be sent any messages
+                    // from canister output queues. Retain the message (along with
+                    // everything behind it in the same queue) until the destination stops
+                    // cooling down, rather than rejecting or dropping it.
+                    if !is_from_subnet_queues && network_topology.is_cooling_down(&dst_subnet_id) {
+                        self.observe_message_status(&msg, LABEL_VALUE_STATUS_RETAINED_COOLING_DOWN);
```
alin-at-dfinity: This should be a different metric. `observe_message_status()` (which should have been more precisely named `observe_routed_message_status()`) so far only counts messages that have been consumed. Whether that's routed, rejected or silently dropped.

So we'd need a different counter for how many output queues (since it's a different metric, we may as well count queues rather than messages, inaccurately) we've skipped because the destination subnet was cooling down.

alin-at-dfinity (commented):
Haven't looked at the tests yet.

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11129/commits/2bf7ebae6f04643c654bc353a91a9cd7aecad38f

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,307 @@ fn build_streams_drops_refunds_at_engine_boundary() {
...
+const COOLING_DOWN_SUBNET: SubnetId = REMOTE_SUBNET;
+/// A third subnet, never cooling down, used to check that only the messages to
+/// cooling down subnets are held back.
+const OTHER_SUBNET: SubnetId = SUBNET_3;
+
+/// The sender of all messages in the tests below, hosted by `LOCAL_SUBNET`.
+const SENDER_CANISTER: CanisterId = CanisterId::from_u64(0);
+/// The destination canister, hosted by the cooling down subnet (which is
+/// `LOCAL_SUBNET` itself in `new_loopback_cooling_down_fixture()`).
+const DESTINATION_CANISTER: CanisterId = CanisterId::from_u64(1);
```
alin-at-dfinity: Nit: I'd give an indication in the name of the fact that this is a canister on a cooling down subnet. E.g. `COOLING_DOWN_CANISTER`.

Else, you can wrap all this new code in an inline `cooling_down` module, then the names become scoped to the module and can stay as generic as they are.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,307 @@ fn build_streams_drops_refunds_at_engine_boundary() {
...
+/// None of these dimensions makes any difference to a message headed for a cooling
+/// down subnet; contrast with `is_illegal_engine_msg` in `build_streams_impl()`.
+fn cooling_down_message_matrix(receiver: CanisterId) -> impl Iterator<Item = RequestOrResponse> {
+    [NO_DEADLINE, SOME_DEADLINE]
+        .into_iter()
+        .flat_map(|deadline| {
+            [Cycles::zero(), ONE_TRILLION_CYCLES]
+                .into_iter()
+                .map(move |cycles| (deadline, cycles))
+        })
```
alin-at-dfinity: As commented on your earlier PR, I'd rather list all combinations. As is, apart from the method name and doc comment it's really hard to understand what exactly is going on.

```suggestion
    [
        (NO_DEADLINE, Cycles::zero()),
        (NO_DEADLINE, ONE_TRILLION_CYCLES),
        (SOME_DEADLINE, Cycles::zero()),
        (SOME_DEADLINE, ONE_TRILLION_CYCLES),
    ].into_iter()
```

rs/messaging/src/routing/stream_builder.rs
mraszyk: > is_subnet_response?

That would be correct because subnet output queues can only contain responses. But I planned to leave subnet output queues for a follow-up PR.

> we must still route/deliver subnet responses, just not other messages.

I now put this into the design doc: messages in subnet output queues (responses as the mgmt canister never makes requests itself) on subnets that are not “cooling down” are not routed to streams to subnets that are “cooling down” (i.e., they stay in the subnet output queues)—this is enough because the (source) subnet that is “cooling down” is not being routed any new input subnet messages and thus it stops “bombarding” the target subnet which is also “cooling down” by its subnet output queue messages eventually;

I.e., we should still not route responses from subnets that are not cooling down. Otherwise, canisters on the subnet that is cooling down could keep calling the mgmt canister on a remote subnet that is not cooling down and if those responses are not held back, then the stream to the cooling down subnet never fully empties.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -458,11 +473,24 @@ impl StreamBuilderImpl {
...
                 // Destination subnet found.
                 Some(dst_subnet_id) => {
-                    let dst_stream_entry = streams.entry(dst_subnet_id);
                     let is_loopback_stream = self.subnet_id == dst_subnet_id;
+
+                    // A cooling down destination subnet must not be sent any messages
+                    // from canister output queues. Retain the message (along with
+                    // everything behind it in the same queue) until the destination stops
+                    // cooling down, rather than rejecting or dropping it.
+                    if !is_from_subnet_queues && network_topology.is_cooling_down(&dst_subnet_id) {
```
alin-at-dfinity, line 487: There is no test for subnet responses being routed to cooling down subnets. There's no need for a full matrix, this filter doesn't care about the message type, class or attached cycles. Just a subnet response to a remote / local cooling down subnet.

rs/messaging/src/routing/stream_builder/tests.rs
```diff
@@ -1151,6 +1151,307 @@ fn build_streams_drops_refunds_at_engine_boundary() {
...
+
+/// Tests that only the messages to the cooling down subnet are held back: a
+/// message from the same canister to a canister on a subnet that is not cooling
+/// down is routed in the very same round.
+#[test]
+fn build_streams_routes_messages_to_other_subnets_while_one_is_cooling_down() {
+    with_test_replica_logger(|log| {
+        let retained = request_from_sender(DESTINATION_CANISTER, NO_DEADLINE, Cycles::zero());
+        let routed =
+            request_from_sender_with_callback(OTHER_CANISTER, NO_DEADLINE, Cycles::zero(), 2);
```
alin-at-dfinity: FWIW, you don't need an explicit callback here. The request would be different from the one above anyway (and you could have a `debug_assert` to that effect) because the destination is different.

alin-at-dfinity (approved):

rs/messaging/src/routing/stream_builder/tests.rs
mraszyk: I kept it as nested `flat_map`s since the matrix now has 3 dimensions and I didn't want to enumerate 8 combinations manually.

rs/messaging/src/routing/stream_builder.rs
mraszyk: I plan to leave subnet output queues for a follow-up PR and thus such a test would be rewritten very soon again. Hence, I'd prefer to only add it in that follow-up PR. Since this condition is a conjunction with `.is_cooling_down(...)` which is always false in production at the moment, this code is unreachable in prod anyway so it should be fine to not cover subnet output queues by tests on this PR.

rs/messaging/src/routing/stream_builder/tests.rs
mraszyk: The callback ID is actually required, committed as [54c9cc5c17](https://github.com/dfinity/ic/pull/11129/commits/7bd6d6ebf071351fe89891a8b00cd6b106008917) (comment only).
I (Claude) verified it: switching `routed` to `request_from_sender(OTHER_CANISTER, ...)` (callback 1) makes the test panic at `canister_states_with_outputs()`'s `assert_eq!(req.sender_reply_callback, callback_id)` — left: 1, right: 2. The helper pushes each request through `push_output_request()`, which mints a fresh callback ID from the sender's `CallContextManager` regardless of destination, so the sender's second request must carry callback ID 2. The distinctness of the two messages isn't the issue; the helper's consistency assertion is. I added an inline comment saying so rather than a `debug_assert`, since the constraint is "must be 2", not "must differ".

rs/messaging/src/routing/stream_builder/tests.rs
mraszyk: Done both: https://github.com/dfinity/ic/pull/11129/commits/e2b816086e28ebcb266de1951f5123a9616d8119?diff=split&w=1

rs/messaging/src/routing/stream_builder.rs
alin-at-dfinity: Right, my bad. I totally forgot that this was only an issue for the cooling down subnet itself and its loopback stream.

But then what's with the `if !is_from_subnet_queues && is_cooling_down(&dst_subnet_id)` below? Why are we letting messages from subnet queues through? Or is this change intended to be strictly about "messages from canister output queues", with "messages from subnet output queues" intended to be handled in the follow-up PR?

If so, I suppose I found it somewhat confusing that you chose to explicitly deal with subnet queues in this PR. If you were going to split it into 2 PRs, I would have found it a lot more natural to ignore subnet queues in this PR altogether. I.e. simply say `if is_cooling_down(&dst_subnet_id) { exclude_queue() }`. Then, in the follow-up refine it to "except for routing subnet responses into the loopback stream"

rs/messaging/src/routing/stream_builder/tests.rs
alin-at-dfinity: Ugh, forgot about that. The other way to deal with this would have been to manually create the sender canister; call `push_output_request()` on it; and update the requests' callback IDs with the returned values.

Never mind.

rs/messaging/src/routing/stream_builder.rs
alin-at-dfinity: I didn't fully get that before.

SGTM

alin-at-dfinity (approved):

rs/messaging/src/routing/stream_builder.rs
mraszyk: > Or is this change intended to be strictly about "messages from canister output queues", with "messages from subnet output queues" intended to be handled in the follow-up PR?

Exactly, this PR should not _change behavior_ for subnet output queues at all.

> this was only an issue for the cooling down subnet itself and its loopback stream

not sure what you mean by this; if you refer to the case when we need to route subnet responses, then we need to do so unless source is not cooling down _and_ target is cooling down, i.e., whenever source is cooling down or target is not cooling down

alin-at-dfinity: You are right, I thought the cooling down subnet only needed to route subnet responses into its loopback stream; but of course, because its management canister is going away, we need to route its responses into any and all streams before it's gone forever.

PR #11131, opened 2026-08-13
perf(key-manager): optimize some clones

pierugo-dfinity:
This PR avoids/optimizes some clones in the DKG key manager:
- Simplify `load_transcripts_from_summary` to clone each transcript once and move each one individually to the respective thread that loads it, instead of cloning the entire summary each time and then iterating through it to find the corresponding transcript (which could even panic).
- In `delete_inactive_keys`, clone only the transcripts of interest, not the entire summaries, DKG (which can contain additional transcripts, like reshared ones and ones for remote subnets) and IDKG (useless here).

### DIFF
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -311,13 +311,13 @@ impl DkgKeyManager {
             let current_transcripts_with_load_deadlines = summary
                 .current_transcripts()
                 .values()
-                .map(|t| (current_interval_start, t.clone()));
+                .map(|t| (current_interval_start, t));
 
             // For next transcripts, we take the start of the next interval as a deadline.
             let next_transcripts_with_load_deadlines = summary
                 .next_transcripts()
                 .values()
-                .map(|t| (next_interval_start, t.clone()));
+                .map(|t| (next_interval_start, t));
 
             current_transcripts_with_load_deadlines.chain(next_transcripts_with_load_deadlines)
         };
@@ -334,6 +334,7 @@ impl DkgKeyManager {
 
             let crypto = self.crypto.clone();
             let logger = self.logger.clone();
+            let transcript = transcript.clone();
             let (tx, rx) = sync_channel(0);
             self.pending_transcript_loads
                 .insert(transcript.dkg_id.clone(), (deadline, rx));
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -303,50 +303,44 @@ impl DkgKeyManager {
     /// that CSP does not reload transcripts, which were successfully loaded
     /// before.
     fn load_transcripts_from_summary(&mut self, summary: &DkgSummary) {
-        let transcripts_to_load: Vec<_> = {
-            let current_interval_start = summary.height;
-            let next_interval_start = summary.get_next_start_height();
+        let current_interval_start = summary.height;
+        let next_interval_start = summary.get_next_start_height();
 
+        let transcripts_to_load = {
             // For current transcripts we take the current summary height as a deadline.
             let current_transcripts_with_load_deadlines = summary
                 .current_transcripts()
-                .iter()
-                .filter(|(_, t)| !self.pending_transcript_loads.contains_key(&t.dkg_id))
-                .map(|(_, t)| (current_interval_start, t.dkg_id.clone()));
+                .values()
+                .map(|t| (current_interval_start, t.clone()));
 
             // For next transcripts, we take the start of the next interval as a deadline.
             let next_transcripts_with_load_deadlines = summary
                 .next_transcripts()
-                .iter()
-                .filter(|(_, t)| !self.pending_transcript_loads.contains_key(&t.dkg_id))
-                .map(|(_, t)| (next_interval_start, t.dkg_id.clone()));
+                .values()
+                .map(|t| (next_interval_start, t.clone()));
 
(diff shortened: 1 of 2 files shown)


rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -303,50 +303,44 @@ impl DkgKeyManager {
...
 
+        let transcripts_to_load = {
             // For current transcripts we take the current summary height as a deadline.
             let current_transcripts_with_load_deadlines = summary
                 .current_transcripts()
-                .iter()
-                .filter(|(_, t)| !self.pending_transcript_loads.contains_key(&t.dkg_id))
-                .map(|(_, t)| (current_interval_start, t.dkg_id.clone()));
+                .values()
+                .map(|t| (current_interval_start, t.clone()));
```
alin-at-dfinity: I took a quick peek at the definition of `NiDkgTranscript` and it's quite the struct: 200 bytes plus a `BTreeSet` and a `BTreeMap`, both holding `PrincipalIds`.

I suppose we do need it cloned if we're going to pass it on to a worker thread. And I'm not at all clear about the actual cardinalities and the workload involved. But would it make sense to wrap the `NiDkgTranscripts` in `DkgSummary::current_transcripts` and `DkgSummary::next_transcripts` within `Arcs`? Then the cloning would be really cheap.

(I suppose you could also wrap the whole `DkgSummary` in an `Arc`. Then you wouldn't need all these changes. But my reading suggests that you would prefer to not have to iterate the summary looking for individual transcripts.)

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -303,50 +303,44 @@ impl DkgKeyManager {
...
+        let next_interval_start = summary.get_next_start_height();
 
+        let transcripts_to_load = {
             // For current transcripts we take the current summary height as a deadline.
             let current_transcripts_with_load_deadlines = summary
                 .current_transcripts()
-                .iter()
-                .filter(|(_, t)| !self.pending_transcript_loads.contains_key(&t.dkg_id))
-                .map(|(_, t)| (current_interval_start, t.dkg_id.clone()));
+                .values()
```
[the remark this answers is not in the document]
pierugo-dfinity: Good further optimization, [applied](https://github.com/dfinity/ic/pull/11131/commits/2589f61fe731db16f6847e75958e66aab54c896f).

rs/consensus/dkg/src/dkg_key_manager.rs
alin-at-dfinity, line 455: I've also poked at the implementation of `NiDkgAlgorithm::retain_only_active_keys` and it looks to me like it would work just as well with references to `NiDkgTranscripts` (it appears to be synchronous and only does read accesses to the transcripts).

alin-at-dfinity (approved):
A couple of clueless comments/questions, otherwise LGTM.

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: > But would it make sense to wrap the NiDkgTranscripts in DkgSummary::current_transcripts and DkgSummary::next_transcripts within Arcs?

Maybe, but I feel that it would make it harder to comprehend what a `DkgSummary` entail, specifically who is it shared with (whereas keeping them without `Arc`s ensures whoever owns a `DkgSummary` uniquely owns everything inside of it). Those transcripts are expected to be loaded (and thus cloned) about once per interval, so I do not think it would really justify that change.

> I suppose you could also wrap the whole DkgSummary in an Arc.

You mean creating an `Arc::new(summary)` and move a cheap copy to the thread? I do not think this is possible because even though we own the [`summary_block`](https://sourcegraph.com/r/github.com/dfinity/ic@a87fac145f0faa1f13fe4a8aa866dab7b9fbd52d/-/blob/rs/consensus/dkg/src/dkg_key_manager.rs?L218), we cannot own the included [`summary`](https://sourcegraph.com/r/github.com/dfinity/ic@a87fac145f0faa1f13fe4a8aa866dab7b9fbd52d/-/blob/rs/consensus/dkg/src/dkg_key_manager.rs?L220), because itself is behind an [`Arc`](https://sourcegraph.com/r/github.com/dfinity/ic@a87fac145f0faa1f13fe4a8aa866dab7b9fbd52d/-/blob/rs/types/types/src/consensus/payload.rs?L177) which would require moving out of it.

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: Yes, I also think this function could be optimized to work only with references. But even if it did, since it runs in a separate thread, we would also need to move the (owned) transcripts inside the closure.

rs/consensus/dkg/src/dkg_key_manager.rs
alin-at-dfinity: What I meant was that the summary block (or whoever, I wasn't necessarily aware who it was specifically that owned it) could hold the summary within an `Arc` and in some cases return cheap clones to said `Arc` instead of plain references. Similarly with transcripts, I guess.

Although I'm not clear whether the transcripts or summary are ever mutated by their owner (or really, what they stand for). If they are never mutated, then why does it matter who owns what?

(OTOH, having looked at a bit of the code, the rabbit hole runs deep. So switching anything to `Arcs` or even plain references, where applicable, would result in LOTS of code changes. So I fully understand why you're not particularly keen on it.)

But we should definitely take this discussion offline, we're not going to reach any major conclusions (or a meaningful understanding, in my case) on a GitHub comment thread.

PR #11132, opened 2026-08-13
feat: ic0_subnet_self_node_count

eichhorl:
Implement a new system API function allowing a canister to determine the size of its current subnet. This will be helpful for flexible outcalls, in which a canister can individually specify how many replicas should perform a request, and how many of them should produce a response. This only makes sense within the context of the own subnet size.

Specified here (draft): https://github.com/dfinity/developer-docs/pull/254


rs/interfaces/src/execution_environment.rs
```diff
@@ -1432,6 +1434,10 @@ pub trait SystemApi {
         size: usize,
         heap: &mut [u8],
     ) -> HypervisorResult<()>;
+
+    /// Used to look up the number of nodes currently on the subnet that the
+    /// calling canister is running on.
+    fn ic0_subnet_self_node_count(&self) -> HypervisorResult<u32>;
```
alin-at-dfinity, line 1440: Idle thoughg: Not the best place for this, but wondering whether we shouldn't have a more generic `ico_subnet_info()` method instead. It could, for now, only return the subnet size inside a struct, but it would allow us later to add fields such as subnet type or cost schedule or whatnot, without adding separate calls for each.

alin-at-dfinity (approved):

rs/interfaces/src/execution_environment.rs
eichhorl: As discussed, will leave it as is for now, but we can revisit if more use cases appear

PR #11133, opened 2026-08-13
feat: Lazy hashing of canister modules

michael-weigelt:


rs/canonical_state/src/lazy_tree_conversion.rs
```diff
@@ -911,7 +911,7 @@ impl<'a> CanisterFork<'a> {
                 }
                 METADATA_LABEL => canister_metadata_as_tree(execution_state, self.version),
                 MODULE_HASH_LABEL => {
-                    Blob(execution_state.wasm_binary.binary.module_hash_ref(), None)
+                    blob(move || execution_state.wasm_binary.binary.module_hash().to_vec())
```
michael-weigelt: I had to remove the `module_hash_ref` method because the data is in a mutex now and can't leave the mutex guard (a temporary value). 

However, `[u8;32]` implements `Copy` so this should be cheap anyway.

rs/types/wasm_types/src/lib.rs
```diff
@@ -62,22 +61,40 @@ pub enum ModuleLoadingStatus {
...
-    // The Wasm binary.
-    #[validate_eq(Ignore)]
+    /// The Wasm binary.
     module: ModuleStorage,
-    // The Sha256 hash of the binary.
-    module_hash: [u8; WASM_HASH_LENGTH],
+    /// The Sha256 hash of the binary.
+    /// This field is `None` when the hash is not yet computed (as an optimization).
+    /// This is the case after a call to `write` and before a subsequent call to `module_hash`.
+    module_hash: Mutex<Option<[u8; WASM_HASH_LENGTH]>>,
```
alin-at-dfinity: Do we actually need a `Mutex` here?

If not, a `RefCell` (or even better, a `OnceCell`) would get rid of a lot of `unwraps()` (and overhead). `OnceCell` would even give you back the efficient `module_hash_ref()`.

alin-at-dfinity (approved):

rs/types/wasm_types/src/lib.rs
michael-weigelt: `OnceCell` is not enough because we may write to this several times. `RefCell` is not `Sync`. Its Sync equivalent is an `RwLock`, which we considered but found that we don't have multiple readers so a `Mutex` will be simpler. 

If I could have used a simple `Cell` here, I would have. But I think we can't use that safely given that `CanisterModule`s are inside `Arc`s and shared across threads.

alin-at-dfinity: How about `OnceLock` / `LazyLock`? They are both `Sync` and both provide references to the content.

As for writing multiple times to them, since you have a mutable reference to the `CanisterModule` at that point, you can replace the lock itself, you don't need to replace its contents.

PR #11136, opened 2026-08-13
refactor(registry): Cleanup registry invariant helpers

Bownairo:
Cleanup invariant helpers to reduce duplicated code, improve search efficiency, and update error handling.


daniel-wong-dfinity-org-twin (commented):
Test? Let's see this trip wire explode 💥

rs/registry/canister/src/invariants/common.rs
```diff
@@ -112,14 +112,17 @@ pub(crate) fn get_all_chain_key_signing_subnet_list_records(
...
-) -> BTreeSet<HostosVersionRecord> {
-    let mut result = BTreeSet::new();
+) -> BTreeMap<String, HostosVersionRecord> {
+    let mut result = BTreeMap::new();
     for (k, v) in snapshot {
-        if k.starts_with(HOSTOS_VERSION_KEY_PREFIX.as_bytes()) {
-            let hostos_version_record = HostosVersionRecord::decode(v.as_slice()).unwrap();
-            result.insert(hostos_version_record);
+        if let Some(key) = str::from_utf8(k)
+            .unwrap()
```
daniel-wong-dfinity-org-twin: This function should not blow up just because there is a non-UTF-8 key, even though we will probably never have such a key in practice. Defensive programming means, avoid relying on things to the greatest "practical" extent. In this case, not making assumptions about keys that you are not interested in does not require "bending over backwards".

rs/registry/canister/src/invariants/common.rs
```diff
@@ -112,14 +112,17 @@ pub(crate) fn get_all_chain_key_signing_subnet_list_records(
...
 
 // Retrieve all HostOS version records
-pub(crate) fn get_all_hostos_version_records(
+pub(crate) fn get_all_hostos_version_records_with_keys(
     snapshot: &RegistrySnapshot,
-) -> BTreeSet<HostosVersionRecord> {
-    let mut result = BTreeSet::new();
+) -> BTreeMap<String, HostosVersionRecord> {
+    let mut result = BTreeMap::new();
     for (k, v) in snapshot {
```
daniel-wong-dfinity-org-twin:
```suggestion
    for (k, v) in snapshot.range(HOSTOS_VERSION_KEY_PREFIX..) {
        if !k.starst_with(HOSTOS_VERSION_KEY_PREFIX) {
            break;
        }

```

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
-    let node_versions = get_all_hostos_versions_of_nodes(snapshot);
-    all_versions.extend(node_versions);
-
-    // Get the current list of registered HostOS versions
-    let registered_versions = get_all_hostos_version_records(snapshot);
+    let versions_in_use = get_all_hostos_versions_of_nodes(snapshot);
 
-    all_versions.extend(registered_versions.into_iter().map(|v| v.hostos_version_id));
-    all_versions.dedup();
+    // Re-collect since we can't compare `BTreeSet<String>` with `BTreeSet<&String>` with `is_superset`.
```
daniel-wong-dfinity-org-twin, line 26:
```suggestion
    // Convert to BTreeSet<&String> from BTreeSet<String>.
```

I wouldn't mention why, because `is_superset` is not mentioned until later. Also, the reason/motivation seems unimportant. What the reader really wants to know here is, "WTF is even happening?", because it kind of looks like nothing is happening. The subtle thing that the reader might have missed is that this adds a `&` to `String` within `Set`.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
+    let versions_in_use: BTreeSet<_> = versions_in_use.iter().collect();
```
daniel-wong-dfinity-org-twin: The comment is maybe unnecessary if you just mention `&String` here, instead of leaving it unsaid. If the whole point is to go from `String` to `&String`, do not elide! Just because Rust allows it does not mean it's a good thing.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
-    for version in all_versions {
-        // Check that every referenced version exists, i.e. we can only set a
-        // Node's version to one that has already been added to the registry.
-        let r = get_hostos_version_record(snapshot, version);
+    // Get the current list of registered HostOS versions
+    let elected_versions = get_all_hostos_version_records_with_keys(snapshot);
+    let elected_set: BTreeSet<_> = elected_versions.keys().collect();
+    assert!(
+        elected_set.is_superset(&versions_in_use),
+        "Using a version that isn't elected. Elected versions: {elected_set:?}, in use: {versions_in_use:?}."
```
daniel-wong-dfinity-org-twin: It would be more helpful to see what's missing. It would also be helpful to know who is using the missing thing.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
+    );
+    assert!(
+        elected_set.iter().all(|v| !v.trim().is_empty()),
```
daniel-wong-dfinity-org-twin: Seems to me, if there is anything to trim, THAT right there is a problem unto itself.

> [Don't sweep problems under the rug.][rug]

[rug]: https://github.com/dfinity/ic/blob/fa09a04/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L76

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
     get_all_node_records(snapshot)
         .into_values()
         .filter_map(|node_record| node_record.hostos_version_id)
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::common::test_helpers::invariant_compliant_registry;
+
```
daniel-wong-dfinity-org-twin: I don't think grouping `use`s (using blank lines) is a local pattern.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -18,45 +17,76 @@ use ic_registry_keys::make_hostos_version_key;
...
+    use prost::Message;
+
+    #[test]
+    #[should_panic(
+        expected = "The registry key and internal `hostos_version_id` must be consistent."
+    )]
+    fn panic_with_inner_version_mismatch() {
+        let registry = invariant_compliant_registry(0);
+
+        let key = make_hostos_version_key("FOO");
```
daniel-wong-dfinity-org-twin: Feel free to use less neutral values, such as "will-explode" and "because-does-not-match".

What I kind of wish is that we would require 40 char hexadecimal (lower case), mostly, because it would a strong hint that we use git commit IDs. If we did that, you could still be "less neutral" by using "deadbeef" + "f" * 32, and such.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -53,9 +53,11 @@ pub(crate) fn check_replica_version_invariants(
     versions_in_use.append(&mut get_all_standard_engine_replica_versions(snapshot));
     versions_in_use.append(&mut get_all_api_boundary_node_versions(snapshot));
 
-    let elected_set: BTreeSet<_> = get_all_replica_version_records(snapshot)
-        .into_keys()
-        .collect();
+    // Re-collect since we can't compare `BTreeSet<String>` with `BTreeSet<&String>` with `is_superset`.
```
daniel-wong-dfinity-org-twin, line 61: Ditto.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -53,9 +53,11 @@ pub(crate) fn check_replica_version_invariants(
...
+    let versions_in_use: BTreeSet<_> = versions_in_use.iter().collect();
```
daniel-wong-dfinity-org-twin: Ditto.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -65,36 +67,32 @@ pub(crate) fn check_replica_version_invariants(
         "Elected an empty version ID."
     );
 
-    for version in elected_set {
-        let r = get_replica_version_record(snapshot, &version);
-
+    for (key, record) in elected_versions {
```
daniel-wong-dfinity-org-twin: Ditto.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -65,36 +67,32 @@ pub(crate) fn check_replica_version_invariants(
...
+        if let Some(Err(defects)) = record.guest_launch_measurements.map(|v| v.validate()) {
             panic!("guest_launch_measurements are not valid. Defects: {defects:?}");
         }
 
         // Enforce that the stored version always matches the key
-        if let Some(replica_version_id) = r.replica_version_id {
-            assert_eq!(replica_version_id, version);
+        if let Some(replica_version_id) = record.replica_version_id {
+            assert_eq!(
+                replica_version_id, key,
```
daniel-wong-dfinity-org-twin: Shouldn't we strengthen this to

```
assert_eq!(record.replica_version_id, Some(key));
```

? Alternatively, I think we can just remove `optional` from [this line].

[this line]: https://github.com/dfinity/ic/blob/369f5bfe950e061feee5e297c221c4b4d735531b/rs/protobuf/def/registry/replica_version/v1/replica_version.proto#L21

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -65,36 +67,32 @@ pub(crate) fn check_replica_version_invariants(
...
+                "The registry key and internal `replica_version_id` must be consistent."
```
daniel-wong-dfinity-org-twin: [Leave breadcrumbs][rule].

[rule]: https://github.com/dfinity/ic/blob/fa09a04/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L343

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -677,4 +675,22 @@ mod tests {
...
     }
+
+    #[test]
+    #[should_panic(
+        expected = "The registry key and internal `replica_version_id` must be consistent."
+    )]
+    fn panic_with_inner_version_mismatch() {
+        let registry = invariant_compliant_registry(0);
+
+        let key = make_replica_version_key("FOO");
```
daniel-wong-dfinity-org-twin: Ditto.

rs/registry/canister/src/invariants/common.rs
daniel-wong-dfinity-org-twin: This seems like it would yield a "big" performance improvement, since it ONLY scans keys that you want, not ALL of them.

range scan ≫ full scan

rs/registry/canister/src/invariants/common.rs
Bownairo: What do you suggest instead? Many other fns in this file follow the same pattern. I think the root is that the registry returns bytes, and the helpers use `String`s, so the conversion must happen at some point.

rs/registry/canister/src/invariants/hostos_version.rs
Bownairo: > Convert to BTreeSet<&String> from BTreeSet<String>.

Hmm, I see this and ask "why?", then spend a few minutes trying to figure that out. The conversion can be read from the code (I'll fix up the elision if `.iter()` over `into_iter()` isn't clear enough), but the motivation cannot.

rs/registry/canister/src/invariants/hostos_version.rs
Bownairo: > Rust allows it does not mean it's a good thing.

I'd argue, we don't really care what the thing is, as long as it is borrowed vs owned, (`.iter()` v. `into_iter()`), and we can avoid specifying to let the compiler decide. But it's already hairy here and I blame the colleciton impl. Will update.

rs/registry/canister/src/invariants/replica_version.rs
Bownairo: `assert_eq!` should take care of that. Is this what you mean?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2c26e8e1e012846e4cfc6aeef09ef522
```
assertion `left == right` failed: The values don't match
  left: "FOO"
 right: "BAR"
 ```

**Comment by Bownairo** [comment]
I'll be swapping this to a general cleanup PR, and then I will follow up with my new invariants and tests.

File: rs/registry/canister/src/invariants/common.rs (hunk 1 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/common.rs]: I've collected most of the explosions into one point, but haven't made changes here yet. WDYT?

File: rs/registry/canister/src/invariants/common.rs (hunk 2 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/common.rs]: I've made this the new focus of the PR, and will follow up with my invariants after.

File: rs/registry/canister/src/invariants/hostos_version.rs (hunk 5 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/hostos_version.rs]: I've updated the error to hold the difference, I can thread through the user of the missing value, but it would only be for the error message. WDYT?

File: rs/registry/canister/src/invariants/hostos_version.rs (hunk 6 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/hostos_version.rs]: I only matched the code from the replica version side. Since this PR is a refactor now, I'll remove it from both.

File: rs/registry/canister/src/invariants/hostos_version.rs (hunk 8 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/hostos_version.rs]: Will follow up with the next PR.

File: rs/registry/canister/src/invariants/hostos_version.rs (hunk 7 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/hostos_version.rs]: I think this is gone?

File: rs/registry/canister/src/invariants/replica_version.rs (hunk 12 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/replica_version.rs]: I'm waiting to make that change [here](https://github.com/dfinity/ic/pull/10943).

File: rs/registry/canister/src/invariants/replica_version.rs (hunk 14 above)
Comment (Bownairo) [reply rs/registry/canister/src/invariants/replica_version.rs]: Will follow up in the next PR.

File: rs/registry/canister/src/invariants/common.rs [hunk 15]
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+        .map(|(k, v)| (NodeId::from(k.parse::<PrincipalId>().unwrap()), v))
+        .collect()
 }
 
-/// Returns an all api boundary node ids record from the registry snapshot.
-pub(crate) fn get_api_boundary_node_ids_from_snapshot(
+pub(crate) fn get_value_from_snapshot<T: Message + Default>(
     snapshot: &RegistrySnapshot,
-) -> Result<BTreeSet<NodeId>, InvariantCheckError> {
+    key: String,
```
pierugo-dfinity: nit: this function only needs a slice

I see it has quite a lot of callers, which for most allocate anyways, so feel free to ignore

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
-        .into_keys()
-        .collect();
-    assert!(
-        elected_set.is_superset(&versions_in_use),
-        "Using a version that isn't elected. Elected versions: {elected_set:?}, in use: {versions_in_use:?}."
-    );
-    assert!(
-        elected_set.iter().all(|v| !v.trim().is_empty()),
-        "Elected an empty version ID."
-    );
```
pierugo-dfinity, line 68: Was removing that check intentional?

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
+    let elected_versions = get_all_replica_version_records(snapshot);
+    let elected_set: BTreeSet<&String> = elected_versions.keys().collect();
+    if !elected_set.is_superset(&versions_in_use) {
+        panic!(
+            "Using a version that isn't elected: {:?}.",
+            versions_in_use.difference(&elected_set)
+        );
+    }
+
+    for (key, record) in elected_versions {
```
pierugo-dfinity:
```suggestion
    for (version, record) in elected_versions {
```
maybe?

pierugo-dfinity (approved):

rs/registry/canister/src/invariants/replica_version.rs
Bownairo: Some more context here: https://github.com/dfinity/ic/pull/11136#discussion_r3845040441.

rs/registry/canister/src/invariants/common.rs
Bownairo: How's this look? https://github.com/dfinity/ic/pull/11136/commits/30978dfc40d598757f07f7845e67ad94275069db

rs/registry/canister/src/invariants/replica_version.rs
pierugo-dfinity: Ah yes makes sense with the added check in `TryFrom<&str> for ReplicaVersion`

rs/registry/canister/src/invariants/common.rs
pierugo-dfinity: Hmmm I was instead thinking of asking for a `&str` and prepending a `&` to all call-sites

rs/registry/canister/src/invariants/common.rs
daniel-wong-dfinity-org-twin: I think the way it worked previously was good: HOSTOS_VERSION_KEY_PREFIX.as_bytes(). Basically, the old way involved converting in the other direction.

rs/registry/canister/src/invariants/common.rs
Bownairo: It was slightly less code to take both and leave most of the call sites alone - but I'm happy with whatever, I'll make the change.

rs/registry/canister/src/invariants/hostos_version.rs
daniel-wong-dfinity-org-twin: threading through is probably not worth it. If the reader knows what method was called, and what was passed (e.g. because they are looking at a failed proposal), they can figure it out. My suggestion was more to avoid spamming the reader with uninteresting facts (i.e. other versions). Only the missing ones are of interest to the reader, i.e. desired - allowed.

rs/registry/canister/src/invariants/replica_version.rs
daniel-wong-dfinity-org-twin: Hmm. I'm not sure what I meant here anymore. Yes, assert_eq very often leaves sufficient breadcrumbs.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+    for (k, v) in snapshot.range::<[u8], _>((Included(prefix.as_bytes()), Unbounded)) {
+        let Some(k) = k.strip_prefix(prefix.as_bytes()) else {
+            break;
+        };
+
+        let key = str::from_utf8(k)
+            .map_err(|err| InvariantCheckError {
+                msg: format!("Failed to decode keys from the RegistrySnapshot: {err}"),
+                source: None,
+            })?
```
daniel-wong-dfinity-org-twin, line 66: Unlike before, this use of fallible from_utf8 is ok, because this Err only happens if there is a weird key in the range selected by the caller, not when there is a weird key _anywhere_ in the universe.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+            .to_string();
+        let value = T::decode(v.as_slice()).map_err(|err| InvariantCheckError {
+            msg: format!("Deserialize registry value failed with {err}"),
```
daniel-wong-dfinity-org-twin: More breadcrumb(s): key.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
     }
 }
 
-/// Returns all node records in the snapshot.
-pub(crate) fn get_all_node_records(snapshot: &RegistrySnapshot) -> BTreeMap<NodeId, NodeRecord> {
+fn scan_snapshot_for_prefix<T: Message + Default>(
```
daniel-wong-dfinity-org-twin: I like this 😁

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+        nodes.insert(key, value);
     }
 
-    nodes
+    Ok(nodes)
+}
+
+/// Returns all node records in the snapshot.
+pub(crate) fn get_all_node_records(snapshot: &RegistrySnapshot) -> BTreeMap<NodeId, NodeRecord> {
+    scan_snapshot_for_prefix::<NodeRecord>(snapshot, NODE_RECORD_KEY_PREFIX)
```
daniel-wong-dfinity-org-twin: Mentioning NodeRecord here is nice. I think Rust could infer it, but per my mantra: compiler allows it != good.

rs/registry/canister/src/invariants/common.rs
daniel-wong-dfinity-org-twin: Your name is fine, but I like this slightly better:

```suggestion
fn registry_snapshot_range<T: Message + Default>(
```

daniel-wong-dfinity-org-twin: A couple of ideas for further enhancements, but probably better left to a future PR:

```
// Idea 1:
trait Record : Message + Default {
    const KEY_PREFIX: &str;
}

impl Record for WidgetRecord {
    const KEY_PREFIX: &str = WIDGET_RECORD_KEY_PREFIX;
}

// Idea 2:
trait RegistrySnapshotTrait {
    fn range<RECORD>(&self) -> ... 
    where
        RECORD: Record
    ;
}

impl RegistrySnapshotTrait for RegistrySnapshot {
    fn range...
}
```

because then, it would look like this:

```
snapshot.range::<WidgetRecord>()
```

This is not as important as getting rid of horrible repetitive (slow!) loops though.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
-    result
+    scan_snapshot_for_prefix::<ChainKeyEnabledSubnetList>(
+        snapshot,
+        CHAIN_KEY_ENABLED_SUBNET_LIST_KEY_PREFIX,
+    )
+    .unwrap()
+    .into_iter()
+    // Preserve the prefix for crypto code, which depends on it downstream.
+    .map(|(k, v)| (format!("{CHAIN_KEY_ENABLED_SUBNET_LIST_KEY_PREFIX}{k}"), v))
+    .collect()
```
daniel-wong-dfinity-org-twin, line 108: This is unfortunate. Perhaps, in a future enhancement, we can re-work callers so that they do not require redundant prefix... Perhaps, TODO now?

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
-            let api_boundary_node_record = ApiBoundaryNodeRecord::decode(value.as_slice()).unwrap();
-            let node_id = NodeId::from(principal_id);
-            result.insert(node_id, api_boundary_node_record);
-        }
-    }
-    result
+    scan_snapshot_for_prefix::<ApiBoundaryNodeRecord>(snapshot, API_BOUNDARY_NODE_RECORD_KEY_PREFIX)
+        .unwrap()
+        .into_iter()
+        .map(|(k, v)| (NodeId::from(k.parse::<PrincipalId>().unwrap()), v))
```
daniel-wong-dfinity-org-twin, line 125: My Record trait idea could also include parsing the key suffix:

```
trait Record: ... {
    type Key;
    fn parse_key_suffix(&suffix) -> ...
}
```

Then, `range` could return (WidgetKey, WidgetRecord) pairs.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+        .collect()
 }
 
-/// Returns an all api boundary node ids record from the registry snapshot.
-pub(crate) fn get_api_boundary_node_ids_from_snapshot(
+pub(crate) fn get_value_from_snapshot<T: Message + Default, U: AsRef<str>>(
     snapshot: &RegistrySnapshot,
-) -> Result<BTreeSet<NodeId>, InvariantCheckError> {
+    key: U,
+) -> Result<Option<T>, InvariantCheckError> {
```
daniel-wong-dfinity-org-twin:
```suggestion
pub(crate) fn get_value_from_snapshot<RESULT, KEY>(
    snapshot: &RegistrySnapshot,
    key: KEY,
) -> Result<Option<RESULT>, InvariantCheckError>
where
    RESULT: Message + Default,
    KEY: AsRef<str>,
{
```

OTOH, why not use &str as the key type?

Another alternative:

```
key: impl AsRef<str>
```

This avoids have to do `get_value_from_snapshot::<WidgetRecord, _>(...)`. I mean, `_` is not terribile, but still.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
     snapshot
-        .keys()
-        .cloned()
-        .map(|key| {
-            String::from_utf8(key).map_err(|_| InvariantCheckError {
-                msg: "Failed to decode keys from the RegistrySnapshot".to_string(),
+        .get(key.as_ref().as_bytes())
+        .map(|v| {
+            T::decode(v.as_slice()).map_err(|err| InvariantCheckError {
+                msg: format!("Deserialize registry value failed with {err}"),
```
daniel-wong-dfinity-org-twin: breadcrumbs: key

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
+    scan_snapshot_for_prefix::<ApiBoundaryNodeRecord>(snapshot, API_BOUNDARY_NODE_RECORD_KEY_PREFIX)
+        .unwrap()
+        .into_iter()
+        .map(|(k, v)| (NodeId::from(k.parse::<PrincipalId>().unwrap()), v))
+        .collect()
 }
 
-/// Returns an all api boundary node ids record from the registry snapshot.
-pub(crate) fn get_api_boundary_node_ids_from_snapshot(
+pub(crate) fn get_value_from_snapshot<T: Message + Default, U: AsRef<str>>(
```
daniel-wong-dfinity-org-twin: Similar to my ideas for range, caller really should not have to call make_widget_key. Instead, adding more requirements to Record trait, calls could instead look like

```
snapshot.get::<WidgetRecord>(widget_key)
```

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,164 +47,119 @@ impl error::Error for InvariantCheckError {
...
-pub(crate) fn assert_sha256(s: &str) {
-    if s.bytes().any(|x| !x.is_ascii_hexdigit()) {
-        panic!("Hash contains at least one invalid character: `{s}`");
-    }
-
-    if s.len() != 64 {
-        panic!("Hash is an invalid length: `{s}`");
-    }
-}
-
```
daniel-wong-dfinity-org-twin, line 206: why? seems alright. Also, this seems orthogonal to other stuff you are doing here?

rs/registry/canister/src/invariants/firewall.rs
```diff
@@ -281,7 +281,9 @@ fn get_node_firewall_rules(
 fn get_firewall_rules(snapshot: &RegistrySnapshot, record_key: String) -> Option<FirewallRuleSet> {
     if snapshot.contains_key(record_key.as_bytes()) {
         Some(
-            get_value_from_snapshot(snapshot, record_key.clone())
+            get_value_from_snapshot(snapshot, &record_key)
+                .ok()
+                .flatten()
```
daniel-wong-dfinity-org-twin:
```suggestion
                .unwrap()
```

Otherwise the tasty tasty breadcrumbs in Err get thrown away!

rs/types/types/src/replica_version.rs
```diff
@@ -57,7 +57,7 @@ impl TryFrom<&str> for ReplicaVersion {
     type Error = ReplicaVersionParseError;
 
     fn try_from(version_str: &str) -> Result<Self, Self::Error> {
-        if !version_str.chars().all(is_valid_version_symbol) {
+        if version_str.is_empty() || !version_str.chars().all(is_valid_version_symbol) {
```
daniel-wong-dfinity-org-twin: Do as you see fit.

To me, `a || !b` is on the verge of being too complicated. IMHO, a good heuristic is "You can repeat a boolean operator (`&&`, `||`, `!`), but do not mix." One way to avoid that is to treat the fail conditions separately. E.g.

```
if ugly {
    ...
} else if smelly {
    ...
} else {
    ...
}
```

In fact, it often makes sense to treat different forms of badness differently, not try to combine them into one using `||`.

Another way to help the reader is with dummy variables. E.g.

```
let has_bad_char = chars.any(|c| !good(c));
let bad = version_str.is_empty() || had_bad_char;
if bad {
```

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -160,10 +155,12 @@ pub(crate) fn has_launch_measurements(
...
 ) -> bool {
-    get_value_from_snapshot::<ReplicaVersionRecord>(
+    get_value_from_snapshot::<ReplicaVersionRecord, _>(
         snapshot,
         make_replica_version_key(replica_version_id),
     )
+    .ok()
+    .flatten()
     .and_then(|replica_version_record| replica_version_record.guest_launch_measurements)
     .is_some()
```
daniel-wong-dfinity-org-twin:
```suggestion
    let result = get_value_from_snapshot::<ReplicaVersionRecord, _>(
        snapshot,
        make_replica_version_key(replica_version_id),
    );
    
    match result {
        Ok(Some(ok)) => ok.guest_launch_measurements.is_some()

        Ok(None) => {
            // This seems misleading. I mean yeah, sure, there are no
            // guest launch measurements, but there is an even bigger problem:
            // no record at all!
            log...
            false
        }

        Err(err) =>  {
            // Previously, this would have panic'ed, right? That seems like it might
            // be better, because otherwise, it gets reported as "no launch
            // measurement" when really, there was a decoding failure, or
            // something else.
            log...
            false
        }
    };
```

To me, only the first case is where a "genuine" `false` can be returned. That's why I put it first. The others seem to sweep problems under the rug. I mean, sure, the original behavior was already sweeping problems under the rug. I'm not asking you to solve that.

My suggestion has a couple of objectives:

1. Do not throw away breadcrumbs (via `.ok()`). My solution to this is log.

2. Seems wrong to conflate Err and Ok(None) (via `.flatten()`). Sure, they both lead to returning `false`, but for different reasons. There is no rule that says "different conditions that lead to the same result must share path". Seems to me, we should put them on different paths. I think it leads to clearer code. I mean, with my suggestion, it's difficult to get confused. That's good!

After staring at your impl for a long time, I thought I had found a bug: What if `get_value_from_snapshot` returns `Ok(Some(Record { guest_launch_measurements: None, ... }))`. Well, I have now convinced myself that your impl does the right thing in that case. Maybe I'm just dumb, but I think it's just genuinely difficult to tell that the code handles all cases correctly, because there is a ton of "conflation" going on. I think with my suggestion, such confusion is really not possible, because it conflates nothing. It just gives three distinct ways of reaching `false`, not one (`.is_some()`).

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
-fn get_replica_version_record(snapshot: &RegistrySnapshot, version: &str) -> ReplicaVersionRecord {
-    get_value_from_snapshot(snapshot, make_replica_version_key(version))
-        .unwrap_or_else(|| panic!("Could not find replica version: {version}"))
-}
-
 fn get_subnet_record(snapshot: &RegistrySnapshot, subnet_id: SubnetId) -> SubnetRecord {
     get_value_from_snapshot(snapshot, make_subnet_record_key(subnet_id))
+        .ok()
+        .flatten()
         .unwrap_or_else(|| panic!("Could not get subnet record for subnet: {subnet_id}"))
```
daniel-wong-dfinity-org-twin, line 114: Yeah, but why? This conflates two different things:

1. no record
2. decode fail

Why `.ok()` is evil:

1. It NEVER returns `Ok(...)`, contrary to the one word on the tin.
2. It throws away `Err`. Imagine putting an incendiary device inside of an airplane's black box. That would be insane! But that's what `.ok()` is!

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
+    let versions_in_use: BTreeSet<&String> = versions_in_use.iter().collect();
+
+    let elected_versions = get_all_replica_version_records(snapshot);
+    let elected_set: BTreeSet<&String> = elected_versions.keys().collect();
+    if !elected_set.is_superset(&versions_in_use) {
+        panic!(
+            "Using a version that isn't elected: {:?}.",
+            versions_in_use.difference(&elected_set).collect::<Vec<_>>()
+        );
+    }
```
daniel-wong-dfinity-org-twin, line 71: Yours is ok, but here is an alternative:


```suggestion
    let missing_version_ids = versions_in_use.difference(&elected_set).collect::<Vec<_>>();
    if !missing_version_ids.is_empty() {
        panic!(
            "Using a replica version that isn't elected: {:?}.",
            missing_version_ids,
        );
    }
```

Ditto for hostos_version.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
+
+    for (version, record) in elected_versions {
```
daniel-wong-dfinity-org-twin, line 73: > record

🙏

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -55,56 +55,51 @@ pub(crate) fn check_replica_version_invariants(
...
-
         // Check whether release package URLs (update image) and corresponding hash are well-formed.
         // As file-based URLs are only used in test-deployments, we disallow file:/// URLs.
         assert_valid_urls_and_hash(
-            &r.release_package_urls,
-            &r.release_package_sha256_hex,
+            &record.release_package_urls,
+            &record.release_package_sha256_hex,
             false, // allow_file_url
         );
```
daniel-wong-dfinity-org-twin, line 95: To ensure future comprehensiveness, consider

```suggestion
        let ReplicaVersionRecord {
            release_package_urls,
            release_package_sha256_hex,
            
            // IF YOU USE .. HERE, YOU WILL FACE A FIRING SQUAD!
        } = &record;

        // Check whether release package URLs (update image) and corresponding hash are well-formed.
        // As file-based URLs are only used in test-deployments, we disallow file:/// URLs.
        assert_valid_urls_and_hash(
            release_package_urls,
            release_package_sha256_hex,
            false, // allow_file_url
        );
```

That way, if some new field gets added, whoever adds that field is FORCED to visit this code and make an affirmative choice: Do I ignore this new thing that I just added, or add validation?

daniel-wong-dfinity-org-twin:
FYI: I'll be going on vacation for 2 weeks starting at the end of this week (back on the 21st). I've told Björn A. about this PR. Seems like he can help bring it over the finish line. OTOH, it seems like you can live without the PR for a while, since this is "just" a refactor?

rs/registry/canister/src/invariants/common.rs
Bownairo: This one was unique because we never returned the key (`BTreeSet<_>` instead of `BTreeMap<String, _>`). But either way, sounds like we're okay with https://github.com/dfinity/ic/pull/11136/changes#r3926423961.

rs/registry/canister/src/invariants/common.rs
Bownairo: Ooh, I like `impl AsRef<str>` the most, I was trying to avoid changing the call sites too much.

I'll leave it as `&str` for now, though, since Pierugo  got here first 🙂. https://github.com/dfinity/ic/pull/11136#discussion_r3911950301

rs/registry/canister/src/invariants/common.rs
Bownairo: Just inlined. Two ifs didn't seem worth it, but I can put it back.

rs/registry/canister/src/invariants/replica_version.rs
Bownairo: There's an extra collect on the happy path that we could avoid with some messy iterator magic, but that's probably fine.

Weighing `!is_superset()` against `!missing_version_ids.is_empty()` I think I like the first. I'll leave it as is unless you really think so.

rs/registry/canister/src/invariants/common.rs
Bownairo: Will leave this to a future PR.

rs/registry/canister/src/invariants/common.rs
Bownairo: Will leave this to a future PR.

rs/registry/canister/src/invariants/common.rs
Bownairo: Will leave this to a future PR.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -20,49 +20,44 @@ use ic_types::hostos_version::HostosVersion;
...
+    let elected_set: BTreeSet<&String> = elected_versions.keys().collect();
+    if !elected_set.is_superset(&versions_in_use) {
+        panic!(
+            "Using a HostOS version that isn't elected: {:?}.",
+            versions_in_use.difference(&elected_set).collect::<Vec<_>>()
+        );
+    }
 
-    for version in all_versions {
+    for (version, record) in elected_versions {
```
bjoernek, line 39: Was dropping the validation of `HostosVersionRecord.hostos_version_id` intentional?

On master, `all_versions` also included `registered_versions.map(|v| v.hostos_version_id)`, so `get_hostos_version_record` panicked with "Could not find HostOS version: X" whenever a record's internal `hostos_version_id` didn't match its key, and `HostosVersion::try_from` ran on that field too. Now that we iterate over the keys only, the field is entirely unchecked: a record stored at `hostos_version_eb3ab997...` with `hostos_version_id: "G@RBAGE"` would pass.

The replica version side kept its equivalent check (`assert_eq!(replica_version_id, version)`), so the two files are now inconsistent. Could we add it here as well, to keep the refactor behavior-preserving?

```rust
assert_eq!(
    record.hostos_version_id, version,
    "The registry key and internal `hostos_version_id` must be consistent."
);
```

Related: the doc comment above still promises that the check covers versions "contained in a HostosVersionRecord" and that "the corresponding HostosVersionRecord exists", neither of which holds any more.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,149 +47,115 @@ impl error::Error for InvariantCheckError {
     }
 }
 
-/// Returns all node records in the snapshot.
-pub(crate) fn get_all_node_records(snapshot: &RegistrySnapshot) -> BTreeMap<NodeId, NodeRecord> {
+fn registry_snapshot_range<T: Message + Default>(
```
bjoernek, line 50: All five callers of this immediately `.unwrap()` the `Result` (L82, L92, L103, L115, L123), and so do several `get_value_from_snapshot` callers (`common.rs:154`, `firewall.rs:285`, `replica_version.rs:111`, `replica_version.rs:171`). So for those paths the error type buys us a nicer panic message and nothing else, while the signature suggests the error is handled.

That's fine if panicking is the deliberate choice here (invariant checks trap by design, and the old code panicked too), but could we say so? Either a short note on this function that callers are expected to `.unwrap()` unless they can propagate, or make the ones that *can* propagate do it. The `?` users (crypto.rs, subnet.rs, unassigned_nodes_config.rs, standard_engine_replica_version.rs) show it's worth having; it's just unclear which category a new caller should pick.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,149 +47,115 @@ impl error::Error for InvariantCheckError {
...
+    snapshot: &RegistrySnapshot,
+    prefix: &str,
+) -> Result<BTreeMap<String, T>, InvariantCheckError> {
     let mut nodes = BTreeMap::new();
```
bjoernek, line 54: nit: leftover name from `get_all_node_records`. This helper is generic now, so `records` reads better.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,149 +47,115 @@ impl error::Error for InvariantCheckError {
...
-    for (k, v) in snapshot {
-        if let Some(id) = get_node_record_node_id(str::from_utf8(k).unwrap()) {
-            let record = NodeRecord::decode(v.as_slice()).unwrap();
-            nodes.insert(NodeId::from(id), record);
-        }
+
+    // Note: effectively .range(PREFIX..)
+    for (k, v) in snapshot.range::<[u8], _>((Included(prefix.as_bytes()), Unbounded)) {
```
bjoernek, line 57: nit: the explicit bounds (and the `ops::Bound` import) look unnecessary. `RangeFrom<&T>` implements `RangeBounds<T>`, so this should work directly and make the note above redundant:

```suggestion
    for (k, v) in snapshot.range::<[u8], _>(prefix.as_bytes()..) {
```

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,149 +47,115 @@ impl error::Error for InvariantCheckError {
...
+        let Some(k) = k.strip_prefix(prefix.as_bytes()) else {
+            break;
+        };
+
+        let key = str::from_utf8(k)
+            .map_err(|err| InvariantCheckError {
+                msg: format!("Failed to decode keys from the RegistrySnapshot: {err}"),
```
bjoernek, line 64: nit: no breadcrumbs on this path. The key can't be included verbatim (it isn't UTF-8 by definition), but the `prefix` we were scanning, or `String::from_utf8_lossy(k)`, would at least tell the reader which record type and roughly which entry blew up. Right now the message is the same for every prefix.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -46,149 +47,115 @@ impl error::Error for InvariantCheckError {
...
-    nodes
+    Ok(nodes)
+}
+
+/// Returns all node records in the snapshot.
+pub(crate) fn get_all_node_records(snapshot: &RegistrySnapshot) -> BTreeMap<NodeId, NodeRecord> {
+    registry_snapshot_range::<NodeRecord>(snapshot, NODE_RECORD_KEY_PREFIX)
+        .unwrap()
+        .into_iter()
+        .map(|(k, v)| (NodeId::from(k.parse::<PrincipalId>().unwrap()), v))
```
bjoernek, line 84: This changes behavior beyond the refactor: `get_node_record_node_id` used `PrincipalId::from_str(key).ok()`, so a key with an unparseable suffix was silently *skipped*. Now it panics, which means one bad key traps every invariant check and therefore blocks all registry mutations, permanently.

I think panicking is the better default (a malformed node record key is a real problem, not something to skip), so this is mostly a request to make it deliberate rather than incidental: worth a note here, and the same applies to `get_api_boundary_node_records_from_snapshot` on L125. In practice keys are always built via `make_node_record_key`, so the risk is low either way.

bjoernek:
One process question: there's no `rs/registry/canister/unreleased_changelog.md` entry. The PR is labelled `refactor`, but a couple of things are observable from the outside:

* `ReplicaVersion`/`HostosVersion::try_from("")` now fail, so electing an empty version ID is rejected by the type rather than by the invariant, with a different message (`must not be empty` instead of `Elected an empty version ID.`).
* Several invariant panic messages changed (`Using a version that isn't elected.` -> `Using a replica version that isn't elected: [...]`). Anyone matching on those strings, e.g. in tests or tooling, is affected.

The governance-checklist bot review was dismissed, so I just want to confirm that was a considered "no canister behavior changes" rather than a drive-by. Happy to be told the changed panic strings don't count.

rs/registry/canister/src/invariants/hostos_version.rs
bjoernek: Should we bring this test back rather than close the thread? It looks like it was resolved by *removing* `panic_with_inner_version_mismatch` in the walk-back rather than by changing it, and that happened on both sides. The replica version side still has an equivalent check in production code (`assert_eq!(replica_version_id, version)`), but is there anything left covering the hostos side? Master enforced it indirectly through the `get_hostos_version_record` lookup, so as far as I can tell this is a net loss of coverage for hostos only. Or is the intent to restore both together in the follow-up PR?

Suggested assert in https://github.com/dfinity/ic/pull/11136#discussion_r3977413272.

rs/registry/canister/src/invariants/hostos_version.rs
daniel-wong-dfinity-org-twin: It's fine if you want to explain why. I didn't mean you MUST remove it. I just meant that in my estimation, it's "extra" information that the reader doesn't really "need" at this very moment to follow what is happening.

If you want to state the motivation, I'd suggest something like this:

> Convert to BTreeSet<&String> from BTreeSet<String>. This allows us to call is_superset a little later.

My main point was to swap out the description of the literal implementation ("Re-collect") with abstract effect ("Convert to ..."), and to more carefully point out the subtle difference between the input and output types. The fact that `.collect` is used to convert is (to me) not the lede, but rather, an implementation detail. `.collect` can be gleaned very literally from the code. Meanwhile, the POINT of `.collect`ing (i.e. to convert to `&`) is subtle, particularly since `&` is not mentioned in the code itself.

rs/nns/integration_tests/src/upgrades_handler.rs
daniel-wong-dfinity-org-twin: asdf

rs/registry/canister/src/invariants/replica_version.rs
daniel-wong-dfinity-org-twin: But in the happy case, collect does just as much work as is_superset.

is_superset is acceptable.

The more important thing is to not spam the error message with compliant data, and both ways satisfy this.

rs/registry/canister/src/invariants/replica_version.rs
```diff
@@ -160,12 +166,21 @@ pub(crate) fn has_launch_measurements(
...
 ) -> bool {
-    get_value_from_snapshot::<ReplicaVersionRecord>(
+    match get_value_from_snapshot::<ReplicaVersionRecord>(
         snapshot,
-        make_replica_version_key(replica_version_id),
+        &make_replica_version_key(replica_version_id),
     )
-    .and_then(|replica_version_record| replica_version_record.guest_launch_measurements)
-    .is_some()
+    .unwrap()
```
daniel-wong-dfinity-org-twin, line 173: I guess the original code exploded in this case, but from within get_value_from_snapshot.

rs/registry/canister/src/invariants/common.rs
```diff
@@ -280,6 +244,6 @@ mod tests {
             vec![0],                  // incorrect value, not an encoded ApiBoundaryNodeRecord
         );
         // this call should panic when decoding the ApiBoundaryNodeRecord
-        get_value_from_snapshot::<ApiBoundaryNodeRecord>(&snapshot, key);
+        get_value_from_snapshot::<ApiBoundaryNodeRecord>(&snapshot, &key).unwrap();
```
daniel-wong-dfinity-org-twin, line 247: Instead of should_panic, you could assert. I think should_panic is generally only used for asserts that occur from within code under test, not from the test itself.

rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -21,34 +24,32 @@ pub(crate) fn check_hostos_version_invariants(
...
             &record.release_package_sha256_hex,
             false,
         );
-
-        // Enforce that the stored version always matches the key
-        assert_eq!(
-            record.hostos_version_id, key,
-            "The registry key and internal `hostos_version_id` must be consistent."
-        );
     }
```
daniel-wong-dfinity-org-twin, line 52: Why delete this? Is this going to be added later?

rs/types/types/src/replica_version.rs
```diff
@@ -66,9 +40,9 @@ impl AsRef<str> for ReplicaVersion {
     }
 }
 
-/// Checks if a valid replica version is allowed to contain specified char.
-fn is_valid_version_symbol(c: char) -> bool {
-    matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '.' | '_' | '-')
+/// Checks if a given char is invalid for use in a replica version.
```
daniel-wong-dfinity-org-twin, line 43: I think the name already conveys this. What doc comment should maybe instead do is, break down abstract "invalid" into concrete. I.e.

> /// Only alphanumeric, dot, underscore, and dash characters are allowed in replica version IDs. In practice, replica version IDs are generally git commit IDs (from the dfinity/ic repo), which are lower case hexadecimal strings of length 40.

rs/registry/canister/src/invariants/hostos_version.rs
daniel-wong-dfinity-org-twin: I was kind of wondering the same. I guess the scope of this PR changed to just refactor, and that will be added in a future PR?

rs/registry/canister/src/invariants/common.rs
daniel-wong-dfinity-org-twin: Are you saying this should have a doc comment that says callers should `.unwrap()` the result? I wouldn't add such a doc comment, because they can call this correctly without `.unwrap()`ing the result (e.g. they can `?`).

rs/registry/canister/src/invariants/common.rs
daniel-wong-dfinity-org-twin: So, are you saying to comment that the `.unwrap()` is unlikely to occur in practice, because we are very good about keeping our keys valid? SGTM.

daniel-wong-dfinity-org-twin:
> One process question: there's no `rs/registry/canister/unreleased_changelog.md` entry. The PR is labelled `refactor`, but a couple of things are observable from the outside:
> 
>     * `ReplicaVersion`/`HostosVersion::try_from("")` now fail, so electing an empty version ID is rejected by the type rather than by the invariant, with a different message (`must not be empty` instead of `Elected an empty version ID.`).
> 
>     * Several invariant panic messages changed (`Using a version that isn't elected.` -> `Using a replica version that isn't elected: [...]`). Anyone matching on those strings, e.g. in tests or tooling, is affected.
> 
> 
> The governance-checklist bot review was dismissed, so I just want to confirm that was a considered "no canister behavior changes" rather than a drive-by. Happy to be told the changed panic strings don't count.

I think we should consider such wording changes to be "non-substantive", and therefore, need not be documented.

daniel-wong-dfinity-org-twin (approved):

PR #11139, opened 2026-08-14
test(ledger): cover a subaccount-qualified ICRC-2 self-spend

mbjorkqvist:
[DEFI-2978](https://dfinity.atlassian.net/browse/DEFI-2978)

## Why

A `transfer_from` needs no allowance only when the spender *is* the account it spends from, and the ledger decides that on the whole account, not the owner alone. The state machine suite covered this for the default subaccount only (`test_transfer_from_self`, whose args hardcode `spender_subaccount: None`), so nothing pinned either half of the subaccount case: that naming the account's own subaccount succeeds, and that naming none fails for want of an allowance even though the owner matches.

Callers depend on the rule — the ckETH minter burns from its own fee subaccount — so a regression in it should surface in the ledger's own suite rather than in a consumer's integration test.

## What

One generic test in the shared suite, wired into the ICP and ICRC ledgers the same way `test_transfer_from_self` is, so both get it. The successful spend is asserted to land at block index 1, which also shows the rejected attempt wrote no block.

Verified against `ledger_test`, `ledger_test_u256` and the ICP `ledger_canister_test`.

[DEFI-2978]: https://dfinity.atlassian.net/browse/DEFI-2978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rs/ledger_suite/icp/ledger/tests/tests.rs
+++ b/rs/ledger_suite/icp/ledger/tests/tests.rs
@@ -35,6 +35,7 @@ use icrc_ledger_types::icrc1::{
 };
 use icrc_ledger_types::icrc2::allowance::{Allowance, AllowanceArgs};
 use icrc_ledger_types::icrc2::approve::{ApproveArgs, ApproveError};
+use icrc_ledger_types::icrc2::transfer_from::TransferFromError;
 use icrc_ledger_types::icrc21::errors::{ErrorInfo, Icrc21Error};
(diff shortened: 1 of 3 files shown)


gregorydemay (approved):
Thanks @mbjorkqvist !

PR #11141, opened 2026-08-14
fix: tolerate a cycles balance increase on DTS resume

mraszyk:
Every kind of paused execution re-creates its helper from the current clean canister state when it is resumed and thus compares the cycles balance of that state with the balance at the start of the DTS execution. The comparison used to reject any change of the balance; it now only rejects a decrease, which the recorded steps replayed on the clean canister state might no longer be able to cover.

An increase is safe: all cycles changes of a DTS execution are applied relative to the balance of the clean canister state (the Wasm execution reports a `CyclesBalanceChange` delta and the prepaid execution cycles are refunded relative as well), so the additional cycles are preserved.

The new tests `dts_resume_succeeds_after_cycles_increase` and `dts_install_code_resume_succeeds_after_cycles_increase` are the counterparts of `dts_resume_fails_due_to_cycles_decrease` and `dts_install_code_resume_fails_due_to_cycles_decrease`: they add cycles to the canister while its execution is paused and assert that the execution completes and that the added cycles are not lost. The former runs every scenario twice, once without adding cycles, and asserts that the two final balances differ by exactly the added cycles. The setup of all the scenarios is now shared by the tests that decrease and increase the cycles balance.

### DIFF
--- a/rs/execution_environment/src/execution/call_or_task.rs
+++ b/rs/execution_environment/src/execution/call_or_task.rs
@@ -443,8 +443,10 @@ impl CallOrTaskHelper {
     }
 
     /// Replays the previous update call steps on the given clean canister.
-    /// Returns an error if any step fails. Otherwise, it returns an instance of
-    /// the helper that can be used to continue the update call execution.
+    /// Returns an error if the cycles balance of the clean canister dropped
+    /// below the cycles balance at the start of the DTS execution or if any step
+    /// fails. Otherwise, it returns an instance of the helper that can be used
+    /// to continue the update call execution.
     fn resume(
         clean_canister: &CanisterState,
         original: &OriginalContext,
@@ -453,7 +455,13 @@ impl CallOrTaskHelper {
     ) -> Result<Self, UserError> {
         let mut helper = Self::new(clean_canister, original, deallocation_sender)?;
         helper.executed_wasm_instructions = paused.executed_wasm_instructions;
-        if helper.initial_cycles_balance != paused.initial_cycles_balance {
+        // The cycles balance of the clean canister must not decrease during the
+        // DTS execution: the recorded steps are replayed on the clean canister
+        // state and a lower balance might no longer be able to cover them.
+        // An increase is safe: all cycles changes of the DTS execution are
+        // applied relative to the balance of the clean canister state and hence
+        // the additional cycles are preserved.
+        if helper.initial_cycles_balance < paused.initial_cycles_balance {
             let msg = match original.call_or_task {
                 CanisterCallOrTask::Update(_) => {
                     "Mismatch in cycles balance when resuming an update call".to_string()
--- a/rs/execution_environment/src/execution_environment/tests.rs
+++ b/rs/execution_environment/src/execution_environment/tests.rs
@@ -6137,9 +6137,9 @@ struct LongRunningExecution {
     ingress_id: Option<MessageId>,
 }
 
-/// Sets up a long-running update call (`method` is `"update"`) or replicated
-/// query (`method` is `"query"`).
-fn long_running_call(method: &str) -> LongRunningExecution {
+/// Sets up a long-running update call, or a long-running replicated query if
+/// `replicated_query` is set.
+fn long_running_call(replicated_query: bool) -> LongRunningExecution {
     let mut test = ExecutionTestBuilder::new()
         .with_instruction_limit(1_000_000)
         .with_slice_instruction_limit(200_000)
@@ -6154,6 +6154,7 @@ fn long_running_call(method: &str) -> LongRunningExecution {
         .append_and_reply()
         .build();
 
+    let method = if replicated_query { "query" } else { "update" };
     let (ingress_id, _) = test.ingress_raw(a_id, method, a);
 
     LongRunningExecution {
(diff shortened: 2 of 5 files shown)


rs/execution_environment/src/execution/call_or_task.rs
```diff
@@ -453,7 +455,13 @@ impl CallOrTaskHelper {
...
         let mut helper = Self::new(clean_canister, original, deallocation_sender)?;
         helper.executed_wasm_instructions = paused.executed_wasm_instructions;
-        if helper.initial_cycles_balance != paused.initial_cycles_balance {
+        // The cycles balance of the clean canister must not decrease during the
+        // DTS execution: the recorded steps are replayed on the clean canister
+        // state and a lower balance might no longer be able to cover them.
+        // An increase is safe: all cycles changes of the DTS execution are
+        // applied relative to the balance of the clean canister state and hence
+        // the additional cycles are preserved.
+        if helper.initial_cycles_balance < paused.initial_cycles_balance {
```
alin-at-dfinity, line 464: Random observation: I suppose we could also fix this from the other direction, i.e. in the same way in which we avoid charging canisters with paused executions, we can avoid crediting them by temporarily pooling (and combining) any incoming refunds within the subnet's refund pool.

It would also make this check more reliable: if a canister gets a big refund and gets charged a smaller amount while paused, we would gloss over the latter.

Anyway, just an idea, not requesting you reimplement this.

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -6122,35 +6123,150 @@ fn paused_execution_fails_to_resume_after_cycles_decrease(
...
+    /// The canister whose long-running execution pauses.
+    canister_id: CanisterId,
+    /// The ingress message whose status reflects the outcome of the long-running
+    /// execution; `None` for canister tasks, which have no ingress status.
+    ingress_id: Option<MessageId>,
+}
+
+/// Sets up a long-running update call (`method` is `"update"`) or replicated
+/// query (`method` is `"query"`).
+fn long_running_call(method: &str) -> LongRunningExecution {
```
alin-at-dfinity: Nit: Why not make it a boolean flag then?

alin-at-dfinity (approved):

rs/execution_environment/src/execution_environment/tests.rs
mraszyk: https://github.com/dfinity/ic/pull/11141/commits/f80f1adf0ddab04fa14760dc4377e2094aeabfe0

rs/execution_environment/src/execution/call_or_task.rs
mraszyk: That sounds quite a bit more complex since the refund pool might be on a different subnet and thus it can't easily check if the receiver is paused when deciding whether to route the refund into the stream. Moreover, anonymous refunds are not the only source of cycles credits to a paused canister. Hence, pooling would need to happen in a new field on the receiver canister state, analogously to the existing `ingress_induction_cycles_debit`, and cover all the following cases (with the risk of new cases being added in the future and not handled properly):
  1. Inducted refund messages — StreamHandler::induct_refund → ReplicatedState::credit_refund → add_cycles (rs/messaging/src/routing/stream_handler.rs:1135, replicated_state.rs:1112). Refunds are pooled whenever a cycle-bearing
  best-effort message is dropped without a reject response carrying the cycles (queues.rs:1526 on_message_dropped); notably, an inbound best-effort response shed under memory pressure refunds to response.originator, i.e. this very
  canister (queues.rs, on_inbound_message_dropped), and it comes back through the loopback stream a round later.
  2. Responses dropped at induction time — SystemState::push_input_impl credits response.refund straight to the balance for a best-effort response with no matching callback, a duplicate best-effort response, or a stopped canister
  (system_state.rs:1408, :1441, :1454). The duplicate case is easily reachable: time_out_callbacks at the start of a round enqueues a DeadlineExpired reject for an earlier in-flight best-effort call (system_state.rs:1898), and the
  real response arriving afterwards is dropped as a duplicate and refunded.
  3. Self-induction — the same drop-and-credit paths in induct_messages_to_self (system_state.rs:1791, :1812), reachable from induct_messages_on_same_subnet (scheduler.rs:956) if the canister still has output messages queued.
  4. HTTP outcall accounting — deliver_canister_http_spent and refund_timed_out_canister_http_contexts run after the execution phase every round and call add_cycles on the sender for unspent/unresponsive-replica allowances
  (rs/messaging/src/canister_http_spent.rs:315, invoked at state_machine.rs:249,255).

alin-at-dfinity: I was thinking (without having a clear picture of the implementation in mind anymore) that we can pool them in the local subnet's refund pool: on induction, whenever we cannot credit the refund directly, add it to the pool instead.

But as said, just an idea. And not sure how practical.

mraszyk: > we can pool them in the local subnet's refund pool

That's a good point and it would indeed address the issue of the remote subnet not knowing if the local canister is paused, but addressing the fact that refunds are produced at various places would require quite some refactoring nevertheless. So I'd proceed with relaxing the balance check as implemented in this PR.

PR #11143, opened 2026-08-14
fix: dynamic HTTP outcalls per-replica cycles limit

eichhorl:
## Proposed Changes
Previously, the pay-as-you-go tracker would always return the same constant default limits (transform instructions, downloaded bytes), as the legacy tracker currently does. This is however incorrect, since the usage is only charged after the corresponding bytes were already downloaded/transformed. Any usage exceeding the available allowance (but within the constant default limit) is therefore uncharged. With this PR, we dynamically determine the usage limits based on the available allowance. Specifically, this is the inverse of the function charging for the usage, after it happened.

Auxiliary changes:
1. Also charge for any networking usage (request time), even if the adapter returns an error
2. Change most cycles parameters/constants that were using the `u128` type to `Cycles` which uses saturating arithmetic by default.


rs/https_outcalls/pricing/src/payg.rs
```diff
@@ -82,9 +94,17 @@ impl PayAsYouGoTracker {
...
     fn get_adapter_limits(&self) -> AdapterLimits {
-        // TODO: Adjust limits based on remaining budget.
+        let max_response_size = if self.is_free {
+            // A free cost schedule never charges, so the budget never constrains
+            // the download; the caller's own limit is all that applies.
+            self.max_response_size
+        } else {
+            // Never above what the caller asked for, even when it could afford more.
+            self.max_response_size
+                .min(max_downloaded_bytes(self.remaining()))
```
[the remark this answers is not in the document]
eichhorl: Good point, switched to constant `max_response_bytes` to limit the post-transform size

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -92,50 +92,67 @@ fn base_fee_amount(
...
-    (instructions.get() as u128) / TRANSFORM_INSTRUCTION_DIVISOR
+pub(crate) fn transform_usage_fee(instructions: NumInstructions) -> Cycles {
+    Cycles::from((instructions.get() as u128) / TRANSFORM_INSTRUCTION_DIVISOR)
+}
+
+/// The most instructions [`transform_usage_fee`] prices within `budget`, i.e. its inverse.
+pub(crate) fn max_transform_instructions(budget: Cycles) -> NumInstructions {
+    NumInstructions::from(
+        u64::try_from((budget * TRANSFORM_INSTRUCTION_DIVISOR).get()).unwrap_or(u64::MAX),
+    )
```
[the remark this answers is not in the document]
eichhorl: Done

rs/https_outcalls/client/src/client.rs
```diff
@@ -395,31 +400,45 @@ async fn execute_http_request(
...
-    budget
-        .subtract_network_usage(NetworkUsage {
-            response_size: downloaded_bytes,
-            response_time: elapsed,
-        })
-        .map_err(|PricingError::InsufficientCycles| CanisterHttpReject {
-            reject_code: RejectCode::CanisterReject,
-            message: "Insufficient cycles".to_string(),
-        })?;
+    charge_network_usage(budget, downloaded_bytes, elapsed)?;
```
pierugo-dfinity: Not a hard thought, but to keep it all the `charge_network_usage`s together, these lines could be moved inside the `Ok(Ok(adapter_response))` arm above and return just `result` out of the match

rs/https_outcalls/client/src/client.rs
```diff
@@ -644,17 +685,25 @@ mod tests {
...
         adapter_response: Result<HttpsOutcallResult, (Code, String)>,
     ) -> Channel {
+        setup_mock_adapter(SingleResponseAdapter::new(adapter_response)).await
+    }
+
+    async fn setup_hanging_adapter_mock() -> Channel {
+        setup_mock_adapter(SingleResponseAdapter::hanging()).await
+    }
+
+    async fn setup_mock_adapter(mock_adapter: SingleResponseAdapter) -> Channel {
```
pierugo-dfinity: Rant: `setup_adapter_mock` and `setup_mock_adapter` sound a bit conflicting...

rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -268,7 +292,7 @@ pub(crate) fn max_consensus_fee(
                 ),
                 subnet_size,
             ) + flexible_extra_response_fee(
-                responses.saturating_sub(min_responses as u128),
+                max_responses.saturating_sub(min_responses),
```
pierugo-dfinity, line 294:
```suggestion
                responses.saturating_sub(min_responses),
```
I would expect this to be consistent with the `responses` 5 lines above no?

pierugo-dfinity (approved):

rs/https_outcalls/client/src/client.rs
eichhorl: Good idea, done 👍 This also let's us inline `charge_network_usage` again.

rs/https_outcalls/client/src/client.rs
eichhorl: True, changed the second one to `serve_mock_adapter`

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Using `responses` directly doesn't type check anymore unfortunately. I think at this point it is easier to just remove the alias.

rs/https_outcalls/client/src/client.rs
pierugo-dfinity: Looks nicer!

rs/https_outcalls/pricing/src/fees.rs
pierugo-dfinity: I see, yeah I agree

PR #11144, opened 2026-08-14
feat(cketh): give the sweeper address its own transaction pipeline

gregorydemay:
## Why

The minter's dedicated sweeper address needs to send Ethereum transactions, and it must not share the main address' nonce sequence: a sweep stuck behind a fee-starved transaction would head-of-line-block every user withdrawal.

## What

A second instance of the transaction pipeline, for the sweeper address, on a nonce sequence of its own. A sweep burns no ckETH, so it is keyed by a plain counter rather than a ledger burn index, and is never reimbursed; it pays gas from the sweeper's prepaid balance, so it has no transaction fee it can fail to cover. Five audit events record the pipeline's transitions, with reconstruction, Candid mirrors and `.did`.

The sweeper's start nonce can be set from both lifecycle arguments. It is optional on install where the main address' equivalent is required, because install arguments are replayed from the event log and a required field would be missing from every event already written.

## Scope

The pipeline only. No timer drives it and nothing enqueues a sweep, so the new state stays empty: the sending task is [DEFI-2926](https://dfinity.atlassian.net/browse/DEFI-2926) in #11237, and EIP-7702 first-time delegation is #11250, stacked above. The sweep-queue source and prepaid-gas gating are still to come.

Sweeper *funding* — burning ckETH from the minter's fee subaccount to prepay that gas — is a separate stack (DEFI-2933) that meets this one only in the withdrawal pipeline, where its request variant has already landed. Hence the audit events here are numbered from `n28`, above the funding event master now owns.

## Candid compatibility

`CI_OVERRIDE_DIDC_CHECK` is set. The check flags the five cases this PR adds to the `Event` payload variant, since a variant returned to callers may not grow under Candid subtyping. It is the additive shape every new audit event has taken: existing cases keep their names and fields, and callers matching exhaustively on the old set see the new ones only for sweeper activity, which nothing enqueues yet. The new optional nonce field on the two argument types is compatible on its own.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/rs/ethereum/cketh/minter/BUILD.bazel
+++ b/rs/ethereum/cketh/minter/BUILD.bazel
@@ -343,7 +343,6 @@ rust_binary(
         "@crate_index//:ethers-core",
         "@crate_index//:ethnum",
         "@crate_index//:flate2",
-        "@crate_index//:hex",
         "@crate_index//:ic-stable-structures",
         "@crate_index//:num-traits",
         "@crate_index//:rlp",
--- a/rs/ethereum/cketh/minter/BUILD.bazel
+++ b/rs/ethereum/cketh/minter/BUILD.bazel
@@ -343,6 +343,7 @@ rust_binary(
         "@crate_index//:ethers-core",
         "@crate_index//:ethnum",
         "@crate_index//:flate2",
+        "@crate_index//:hex",
         "@crate_index//:ic-stable-structures",
         "@crate_index//:num-traits",
         "@crate_index//:rlp",
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -34,7 +34,7 @@ pub struct UpgradeArg {
     pub ethereum_sweeper_contract_address: Option<String>,
     /// Next transaction nonce of the dedicated sweeper address, on its own nonce sequence.
     /// Mirrors `next_transaction_nonce` for the main address.
-    #[cbor(n(12), with = "icrc_cbor::nat::option")]
+    #[cbor(n(11), with = "icrc_cbor::nat::option")]
     pub next_sweeper_transaction_nonce: Option<Nat>,
 }
 
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -32,6 +32,10 @@ pub struct UpgradeArg {
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
     #[n(10)]
     pub ethereum_sweeper_contract_address: Option<String>,
+    /// Next transaction nonce of the dedicated sweeper address, on its own nonce sequence.
+    /// Mirrors `next_transaction_nonce` for the main address.
+    #[cbor(n(12), with = "icrc_cbor::nat::option")]
+    pub next_sweeper_transaction_nonce: Option<Nat>,
 }
 
 pub fn post_upgrade(upgrade_args: Option<UpgradeArg>) {
--- a/rs/ethereum/cketh/minter/src/state/audit.rs
+++ b/rs/ethereum/cketh/minter/src/state/audit.rs
@@ -112,6 +112,44 @@ pub fn apply_state_transition(state: &mut State, payload: &EventType) {
         } => {
             state.record_finalized_transaction(withdrawal_id, transaction_receipt);
         }
+        EventType::AcceptedSweepRequest(request) => {
+            state.next_sweep_id = request.id.next();
+            state.sweeper_transactions.record_request(request.clone());
+        }
+        EventType::CreatedSweeperTransaction {
+            sweep_id,
+            transaction,
+        } => {
+            state
+                .sweeper_transactions
+                .record_created_transaction(*sweep_id, transaction.clone());
+        }
+        EventType::SignedSweeperTransaction {
+            sweep_id: _,
+            transaction,
+        } => {
+            state
+                .sweeper_transactions
+                .record_signed_transaction(transaction.clone());
+        }
+        EventType::ReplacedSweeperTransaction {
+            sweep_id: _,
+            transaction,
+        } => {
+            state
+                .sweeper_transactions
+                .record_resubmit_transaction(transaction.clone());
+        }
+        EventType::FinalizedSweeperTransaction {
+            sweep_id,
+            transaction_receipt,
+        } => {
+            // The sweeper pipeline is never reimbursed and holds no ckETH balance, so unlike the main
+            // pipeline there is no reimbursement tail or balance update — just the finalize mechanics.
+            let _ = state
+                .sweeper_transactions
+                .record_finalized_transaction(*sweep_id, transaction_receipt);
+        }
         EventType::ReimbursedEthWithdrawal(Reimbursed {
             burn_in_block: withdrawal_id,
             reimbursed_in_block,
--- a/rs/ethereum/cketh/minter/src/state/audit/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/audit/tests.rs
@@ -93,6 +93,7 @@ impl GetEventsFile {
         use crate::endpoints::events::{
             AccessListItem as CandidAccessListItem, EventSource as CandidEventSource,
             ReimbursementIndex as CandidReimbursementIndex,
+            TransactionReceipt as CandidTransactionReceipt,
             TransactionStatus as CandidTransactionStatus,
         };
         use crate::eth_logs::EventSource;
@@ -128,6 +129,20 @@ impl GetEventsFile {
                     ledger_id,
                     ckerc20_ledger_burn_index: map_nat(ckerc20_ledger_burn_index),
                 },
+            }
+        }
+
+        fn map_transaction_receipt(receipt: CandidTransactionReceipt) -> TransactionReceipt {
+            TransactionReceipt {
+                block_hash: receipt.block_hash.parse().unwrap(),
+                block_number: receipt.block_number.try_into().unwrap(),
+                effective_gas_price: receipt.effective_gas_price.try_into().unwrap(),
+                gas_used: receipt.gas_used.try_into().unwrap(),
+                status: match receipt.status {
+                    CandidTransactionStatus::Success => TransactionStatus::Success,
+                    CandidTransactionStatus::Failure => TransactionStatus::Failure,
+                },
+                transaction_hash: receipt.transaction_hash.parse().unwrap(),
             }
         }
 
@@ -357,20 +372,7 @@ impl GetEventsFile {
                     transaction_receipt,
                 } => ET::FinalizedTransaction {
                     withdrawal_id: map_nat(withdrawal_id),
-                    transaction_receipt: TransactionReceipt {
-                        block_hash: transaction_receipt.block_hash.parse().unwrap(),
-                        block_number: transaction_receipt.block_number.try_into().unwrap(),
-                        effective_gas_price: transaction_receipt
-                            .effective_gas_price
-                            .try_into()
-                            .unwrap(),
-                        gas_used: transaction_receipt.gas_used.try_into().unwrap(),
-                        status: match transaction_receipt.status {
-                            CandidTransactionStatus::Success => TransactionStatus::Success,
-                            CandidTransactionStatus::Failure => TransactionStatus::Failure,
-                        },
-                        transaction_hash: transaction_receipt.transaction_hash.parse().unwrap(),
-                    },
+                    transaction_receipt: map_transaction_receipt(transaction_receipt),
                 },
                 EventPayload::AcceptedSweepRequest {
                     sweep_id,
@@ -413,20 +415,7 @@ impl GetEventsFile {
                     transaction_receipt,
                 } => ET::FinalizedSweeperTransaction {
                     sweep_id: SweepId(sweep_id.0.to_u64().unwrap()),
-                    transaction_receipt: TransactionReceipt {
-                        block_hash: transaction_receipt.block_hash.parse().unwrap(),
-                        block_number: transaction_receipt.block_number.try_into().unwrap(),
-                        effective_gas_price: transaction_receipt
-                            .effective_gas_price
-                            .try_into()
-                            .unwrap(),
-                        gas_used: transaction_receipt.gas_used.try_into().unwrap(),
-                        status: match transaction_receipt.status {
-                            CandidTransactionStatus::Success => TransactionStatus::Success,
-                            CandidTransactionStatus::Failure => TransactionStatus::Failure,
-                        },
-                        transaction_hash: transaction_receipt.transaction_hash.parse().unwrap(),
-                    },
+                    transaction_receipt: map_transaction_receipt(transaction_receipt),
                 },
                 EventPayload::ReimbursedEthWithdrawal {
                     reimbursed_in_block,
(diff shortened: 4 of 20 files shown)


rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -753,7 +1087,7 @@ impl EthTransactions {
     pub fn sent_transactions_to_finalize(
         &self,
         finalized_transaction_count: &TransactionCount,
-    ) -> BTreeMap<Hash, LedgerBurnIndex> {
+    ) -> BTreeMap<Hash, R::Id> {
         let first_non_finalized_tx_nonce: TransactionNonce =
             finalized_transaction_count.change_units();
         let mut transactions = BTreeMap::new();
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed one PR down, in db25106 on #11178 — that is where the function became generic, so the message is now "for ids {prev_index} and {index}" for both lanes.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -809,89 +1147,15 @@ impl EthTransactions {
         if self
             .processed_withdrawal_requests
             .get(&ledger_burn_index)
-            .expect("BUG: missing processed withdrawal request")
+            .expect("BUG: missing processed request")
             .is_reimbursable()
         {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Already addressed: `record_finalized_transaction_mechanics` no longer exists on the current revision — the reimbursement split in #11177 removed it.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -471,16 +836,20 @@ pub enum CreateTransactionError {
...
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub enum ResubmitTransactionError {
+pub enum ResubmitTransactionError<Id = LedgerBurnIndex> {
     InsufficientTransactionFee {
-        ledger_burn_index: LedgerBurnIndex,
+        ledger_burn_index: Id,
         transaction_nonce: TransactionNonce,
         allowed_max_transaction_fee: Wei,
         max_transaction_fee: Wei,
     },
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Already addressed: the field is `id: Id` on the current revision, so `ResubmitTransactionError` no longer carries a withdrawal-specific name.

rs/ethereum/cketh/minter/src/state/audit.rs
```diff
@@ -115,6 +115,46 @@ pub fn apply_state_transition(state: &mut State, payload: &EventType) {
         } => {
             state.record_finalized_transaction(withdrawal_id, transaction_receipt);
         }
+        EventType::AcceptedSweepRequest(request) => {
+            state.next_sweep_id = SweepId(request.id.0.saturating_add(1));
+            state
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Already addressed: `next_sweep_id` is advanced via `SweepId::next`, which is `checked_add(1).expect("BUG: sweep id space exhausted")` — it traps rather than silently repeating an id.

rs/ethereum/cketh/minter/src/withdraw.rs
```diff
@@ -384,94 +396,108 @@ async fn send_transactions_batch(latest_transaction_count: Option<TransactionCou
...
-        Err(e) => {
-            log!(INFO, "Failed to get finalized transaction count: {e:?}");
         }
     }
+    let actual_finalized_ids: BTreeSet<Id> = receipts.keys().copied().collect();
+    assert_eq!(
+        expected_finalized_ids, actual_finalized_ids,
+        "ERROR: unexpected transaction receipts for some ids"
+    );
+    Some(receipts)
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Leaving this one as is. The assertion is not introduced here: it is the same `assert_eq!` that guards `process_finalized_transactions` on master today, and this PR only renames `withdrawal_id` to the generic `id` around it. Turning it into a retryable `None` changes withdrawal behaviour, so it does not belong in a refactor that is meant to be a no-op. Noting it here so the thread stays open as a pointer, rather than folding it into this stack.

rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
```diff
@@ -32,6 +32,10 @@ pub struct UpgradeArg {
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
     #[n(10)]
     pub ethereum_sweeper_contract_address: Option<String>,
+    /// Next transaction nonce of the dedicated sweeper address, on its own nonce sequence.
+    /// Mirrors `next_transaction_nonce` for the main address.
+    #[cbor(n(12), with = "icrc_cbor::nat::option")]
```
gregorydemay: Why is this number 12? Previous cbor field at 10.

rs/ethereum/cketh/minter/src/state/audit/tests.rs
```diff
@@ -371,6 +372,62 @@ impl GetEventsFile {
...
+                            .try_into()
+                            .unwrap(),
+                        gas_used: transaction_receipt.gas_used.try_into().unwrap(),
+                        status: match transaction_receipt.status {
+                            CandidTransactionStatus::Success => TransactionStatus::Success,
+                            CandidTransactionStatus::Failure => TransactionStatus::Failure,
+                        },
+                        transaction_hash: transaction_receipt.transaction_hash.parse().unwrap(),
+                    },
+                },
```
gregorydemay: duplicated code fragement

rs/ethereum/cketh/minter/BUILD.bazel
```diff
@@ -343,6 +343,7 @@ rust_binary(
         "@crate_index//:ethers-core",
         "@crate_index//:ethnum",
         "@crate_index//:flate2",
+        "@crate_index//:hex",
```
gregorydemay: why is this new dependency needed?

rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
gregorydemay: 🤖 A mistake, fixed in 99d2833. Index 11 is taken in `InitArg`, by its own `next_sweeper_transaction_nonce` — this field simply continued that count instead of restarting in `UpgradeArg`, whose index space is separate. It is now `n(11)`, following `ethereum_sweeper_contract_address` at 10.

Free to renumber only because no minter version has written an `Upgrade` event carrying the field yet; after this ships, the gap would have been permanent.

rs/ethereum/cketh/minter/src/state/audit/tests.rs
gregorydemay: 🤖 Fixed in 28c81f6. The fourteen-line Candid-to-minter receipt conversion was spelled out twice here (`FinalizedTransaction` and `FinalizedSweeperTransaction`) and twice again in `dump_stable_memory.rs`. It is now one `map_transaction_receipt` per file, sitting next to `map_nat` and `map_unsigned_transaction`, and both arms shrink to a call: -58/+36 across the two files.

rs/ethereum/cketh/minter/BUILD.bazel
gregorydemay: 🤖 `dump_stable_memory.rs:333` needs it: `AcceptedSweepRequest.data` is rendered as a hex string in Candid (`format!("0x{}", hex::encode(data))`), so mapping the event back to `EventType` calls `hex::decode`.

Which invites the better question. `UnsignedTransaction.data`, in the same `Event` variant, is a `blob`. If `AcceptedSweepRequest.data` were a `blob` too, the hex encode/decode pair and this dependency would both disappear, and the two calldata fields would agree. The one thing lost is readability in raw `get_events` output, where a blob prints as escaped bytes and a hex string does not.

I left it as is rather than change the interface unasked — say the word and I will switch it to `blob` and drop the dependency.

gregorydemay: 🤖 Done in de2eb00: `AcceptedSweepRequest.data` is a `blob` now, matching `UnsignedTransaction.data`, and `@crate_index//:hex` is gone from `cketh_minter_dump_stable_memory` — the tool builds without it.

One thing worth recording, because bazel caught what cargo could not. #11250 above also decoded hex, for the `r`/`s` of an EIP-7702 authorization, and `cargo check` was happy: `hex` is a crate-level dependency, so the per-target dependency list only exists in bazel. `bazel build` on the dump target failed there immediately. I applied the same reasoning one layer up — the signature components are `blob`s too now, next to `AccessListItem.storage_keys` — so the dependency stays dropped for the whole stack rather than coming back in the PR above.

rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -186,6 +186,41 @@ pub enum EventType {
...
     #[n(27)]
     AcceptedSweeperFundingRequest(#[n(0)] EthWithdrawalRequest),
+    /// The minter enqueued a sweep to be sent from its dedicated sweeper address.
+    #[n(28)]
+    AcceptedSweepRequest(#[n(0)] SweepRequest),
+    /// The minter created a sweep transaction.
+    #[n(29)]
+    CreatedSweeperTransaction {
+        #[n(1)]
+        sweep_id: SweepId,
```
mbjorkqvist: Would it make sense to start the numbering at `#[n(0)]` here, as well as for the fields in all the other newly-added event variants?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -194,6 +194,59 @@ pub struct Erc20WithdrawalRequest {
...
+                .expect("BUG: sweep id space exhausted"),
+        )
+    }
+}
+
+impl fmt::Display for SweepId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
```
mbjorkqvist: Commit https://github.com/dfinity/ic/pull/11178/commits/59aad5fa79936395493e510cc9fed8c7e9017632 in the parent PR dropped the `Display` bound from `PipelineRequest::Id`, so this shouldn't be needed?

rs/ethereum/cketh/minter/src/state/transactions/request.rs
```diff
@@ -186,3 +192,64 @@ impl PipelineRequest for WithdrawalRequest {
...
+            nonce,
+            max_priority_fee_per_gas: transaction_price.max_priority_fee_per_gas,
+            max_fee_per_gas: transaction_price.max_fee_per_gas,
+            gas_limit: transaction_price.gas_limit,
+            destination: self.destination,
+            amount: self.amount,
+            data: self.data.clone(),
+            access_list: Default::default(),
+        })
+    }
```
mbjorkqvist, line 256: Compared to the `create_transaction` for `impl PipelineRequest for WithdrawalRequest` above, there is no check on `request.max_transaction_fee` here. Would it make sense to add one, and also some corresponding tests?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -194,6 +194,59 @@ pub struct Erc20WithdrawalRequest {
...
+    /// sweeps every delegated deposit address named in `data`.
+    #[n(1)]
+    pub destination: Address,
+    /// ETH value moved by the transaction (zero for an ERC-20 sweep, which moves tokens via calldata).
+    #[n(2)]
+    pub amount: Wei,
+    /// Transaction call data: the sweeper contract's batch call, naming the deposit addresses to
+    /// sweep, the IC account each is credited to, and the tokens to move.
+    #[n(3)]
+    pub data: Vec<u8>,
```
mbjorkqvist, line 237: nit: Would it make sense to add a doc comment pointing to `MAX_DEPOSITS_PER_SWEEP` that bounds the size of `data`?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/state/event.rs
gregorydemay: 🤖 Done in 6de2dc82d5, for all four newly-added struct variants (`Created`/`Signed`/`Replaced`/`FinalizedSweeperTransaction`). It also matches what the existing variants do — `CreatedTransaction` numbers `withdrawal_id` as `n(0)` and `transaction` as `n(1)`. Free to change while nothing has written a sweeper event; it would not have been later.

*(Edited: SHA updated after this branch was rebased onto master.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Correct, it is dead: I removed the impl and `cargo check --all-targets` stays clean, so nothing formats a sweep id with `{}`. Dropped in fd513847bc.

*(Edited: SHA updated after this branch was rebased onto master.)*

rs/ethereum/cketh/minter/src/state/transactions/request.rs
gregorydemay: 🤖 Good catch — the asymmetry is worse than a missing check: creation ignored `max_transaction_fee` entirely while resubmission enforces it through `ResubmissionStrategy::GuaranteeEthAmount`, so a sweep created above its allowance was never refused, it just got stuck, unable to be fee-bumped.

Fixed in d4ce46e209 by following the ckERC20 path rather than adding an error: the whole allowance is allocated up front as `max_fee_per_gas`, with the priority fee capped at it. The sweep is then within its ceiling by construction, survives a rising base fee without a resubmission, and `type Error = Infallible` still holds — an error variant here would have cost that.

Four tests in `sweep_lane`: the created transaction spends exactly the allowance, a spiking estimate cannot push it past the allowance, a tiny allowance caps the priority fee rather than producing `max_priority_fee_per_gas > max_fee_per_gas`, and resubmission still refuses to exceed the allowance.

*(Edited: SHA updated after this branch was rebased onto master.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Added in 978f172f98, though as prose rather than a link: `MAX_DEPOSITS_PER_SWEEP` does not exist yet at this layer — it arrives four PRs up in #11258 as a private `const` in `crate::sweep`. Linking it from here would be an unresolved intra-doc link, and even from #11258 it would be public documentation pointing at a private item. So the field now says the size is bounded by the batch the enqueuing side builds, and the number stays where it is enforced.

*(Edited: SHA updated after this branch was rebased onto master.)*

rs/ethereum/cketh/minter/src/state/transactions/request.rs
gregorydemay: 🤖 Reconsidered, and you were right to push on this: clamping was the wrong half of the fix. Once `base_fee + tip` passes the allowance the clamped transaction cannot be mined, and it then holds the sweeper's nonce while resubmission refuses to bump it past that same allowance — stuck permanently, head-of-line blocking every later sweep on the lane, silently.

So 2d3a6bee77 makes the creation fallible with a `CreateSweepTransactionError`, keyed by `SweepId` rather than reusing the burn-index-shaped `CreateTransactionError`. The condition is the ckERC20 one — `gas_fee_estimate.min_max_fee_per_gas() > allowance_per_gas` — which also covers the case where the tip alone fits but `base_fee + tip` does not; the tip then needs no clamping, since `base + tip <= allowance` implies `tip <= allowance`. `Error = Infallible` is gone: a prepaid allowance falling behind the market is an ordinary condition, not an unrepresentable one.

The sweep stays queued for a cheaper block instead of burning a nonce. Tests: the allowance is still allocated in full when it covers the fee, and creation is refused both when the estimate spikes and when the tip alone exceeds the allowance, with the resubmission refusal test unchanged. The caller in #11237 handles the error by logging and rescheduling, as `create_transactions_batch` does for withdrawals.

PR #11145, opened 2026-08-14
feat(cketh): configure and expose the sweeper contract address

gregorydemay:
Adds an optional `ethereum_sweeper_contract_address` field to the ckETH minter's init and upgrade arguments, holding the address of the sweeper smart contract deployed on Ethereum (DEFI-2925). It can be set at install time and changed on upgrade, exactly like the existing deposit helper contract addresses. Nothing consumes it yet — this is the configuration plumbing the sweeper will build on.

Along the way, the per-contract `InvalidStateError` variants collapse into a single `InvalidContractAddress`, and both address rules now live in `validate_config`: no configured contract address may be zero, and no two contracts may share an address. Since the zero check no longer sits in the log scraping state, setting a contract address became infallible and `LogScrapingStateError` could go away.

Finally, the address is surfaced in `get_minter_info` and in the Metadata section of the minter dashboard, so it can be inspected like the minter address and the helper contracts. The dashboard row is omitted when no sweeper contract is configured.

Note: the distinctness rule is a new constraint on existing configurations. Mainnet's three helper contracts are distinct, so nothing currently deployed trips it.

### DIFF
--- a/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
+++ b/rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
@@ -30,6 +30,8 @@ pub struct UpgradeArg {
     pub deposit_with_subaccount_helper_contract_address: Option<String>,
     #[cbor(n(9), with = "icrc_cbor::nat::option")]
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
+    #[n(10)]
+    pub ethereum_sweeper_contract_address: Option<String>,
 }
 
 pub fn post_upgrade(upgrade_args: Option<UpgradeArg>) {
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -25,6 +25,7 @@ use icrc_ledger_types::icrc1::account::Account;
 use std::cell::RefCell;
 use std::collections::{BTreeMap, BTreeSet, HashSet, btree_map};
 use std::fmt::{Display, Formatter};
+use std::iter::once;
 use strum_macros::EnumIter;
 use transactions::EthTransactions;
 
@@ -113,6 +114,10 @@ pub struct State {
     /// ckERC20 deposit addresses registered via `deposit_erc20`, individually
     /// derived for each user and watched for incoming deposits.
     pub automatic_deposits: AutomaticDeposits,
+
+    /// Address of the sweeper smart contract on Ethereum, which the minter
+    /// delegates to when sweeping funded deposit addresses.
+    pub sweeper_contract_address: Option<Address>,
 }
 
 #[derive(Eq, PartialEq, Debug)]
@@ -122,6 +127,7 @@ pub enum InvalidStateError {
     InvalidLedgerId(String),
     InvalidEthereumContractAddress(String),
     InvalidErc20HelperContractAddress(String),
+    InvalidSweeperContractAddress(String),
     InvalidMinimumWithdrawalAmount(String),
     InvalidLastScrapedBlockNumber(String),
     InvalidLastErc20ScrapedBlockNumber(String),
@@ -474,6 +480,7 @@ impl State {
             evm_rpc_id,
             deposit_with_subaccount_helper_contract_address,
             last_deposit_with_subaccount_scraped_block_number,
+            ethereum_sweeper_contract_address,
         } = upgrade_args;
         if let Some(nonce) = next_transaction_nonce {
             let nonce = TransactionNonce::try_from(nonce)
@@ -547,6 +554,12 @@ impl State {
         if let Some(evm_id) = evm_rpc_id {
             self.evm_rpc_id = evm_id;
         }
+        if let Some(address) = ethereum_sweeper_contract_address {
+            let address = Address::from_str(&address).map_err(|e| {
+                InvalidStateError::InvalidSweeperContractAddress(format!("ERROR: {e}"))
+            })?;
+            self.sweeper_contract_address = Some(address);
+        }
         self.validate_config()
     }
 
@@ -583,6 +596,10 @@ impl State {
         );
         ensure_eq!(self.ckerc20_tokens, other.ckerc20_tokens);
         ensure_eq!(self.automatic_deposits, other.automatic_deposits);
+        ensure_eq!(
+            self.sweeper_contract_address,
+            other.sweeper_contract_address
+        );
 
         self.eth_transactions
             .is_equivalent_to(&other.eth_transactions)
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -25,7 +25,6 @@ use icrc_ledger_types::icrc1::account::Account;
 use std::cell::RefCell;
 use std::collections::{BTreeMap, BTreeSet, HashSet, btree_map};
 use std::fmt::{Display, Formatter};
-use std::iter::once;
 use strum_macros::EnumIter;
 use transactions::EthTransactions;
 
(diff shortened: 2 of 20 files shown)


rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -547,6 +556,11 @@ impl State {
         if let Some(evm_id) = evm_rpc_id {
             self.evm_rpc_id = evm_id;
         }
+        if let Some(address) = ethereum_sweeper_contract_address {
+            let address = Address::from_str(&address)
+                .map_err(|e| InvalidStateError::InvalidContractAddress(format!("ERROR: {e}")))?;
+            self.sweeper_contract_address = Some(address);
+        }
         self.validate_config()
```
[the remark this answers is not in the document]
gregorydemay: trapping rolls back state.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -182,6 +185,25 @@ impl State {
                     .to_string(),
             ));
         }
+
+        let mut seen_contract_addresses = BTreeSet::new();
```
mbjorkqvist: This rule is re-checked against every **historical** upgrade, not just the current configuration.

`post_upgrade` → `replay_events()` → for each `EventType::Upgrade`, `state.upgrade(arg).expect("applying upgrade event should succeed")` (`state/audit.rs:20-24`). So it has to hold for every configuration the minter has ever passed through, and a violation is a panic in `post_upgrade`: the minter can't be upgraded to this wasm at all, recoverable only by shipping a wasm without the check. The PR description's claim is present tense ("Mainnet's three helper contracts *are* distinct"), which isn't the invariant replay needs.

I checked the in-repo deployment docs and all four environments look clean:

| env | eth helper | erc20 helper | subaccount helper |
|---|---|---|---|
| mainnet | `0x7574eB42…` (init) | `0x6abDA043…` (2024-05-10) | `0x18901044…` (2024-11-22) |
| staging | `0x256905f9…` | `0x6815595c…` | — |
| testnet (Sepolia) | `0xb44B5e75…` | `0x674Cdbe6…` | — |
| playground | `0x3a1b77f3…` | `0x17146788…` | — |

The one incident that looks like it could bite — `minter_upgrade_2024_11_30.md`, "`eth_helper_contract_address` and `erc20_helper_contract_address` were wrongly reused to point to the new helper smart contract" — was a `get_minter_info` reporting bug fixed with empty upgrade args (`didc encode '()'`), not a state duplicate, so it doesn't trip the rule either.

So this looks safe. But those docs aren't necessarily an exhaustive record of every upgrade ever submitted, and the failure mode is hard. Would it make sense to manually confirm against the real audit log before merge? `cketh_minter_dump_stable_memory` already exists in this crate for exactly that. And I'd restate the note in the description as "no historical configuration" rather than "the current one".

Separately, on the messages: neither error says *which* contract. The zero case says only "contract address must not be zero", and the duplicate case names the address but not the two contracts sharing it — even though the loop has the `LogScrapingId` right here and discards it as `_id`. Iterating labelled pairs instead of bare `&Address` would make both actionable at no cost.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -113,15 +113,18 @@ pub struct State {
...
 }
 
 #[derive(Eq, PartialEq, Debug)]
 pub enum InvalidStateError {
     InvalidTransactionNonce(String),
     InvalidEcdsaKeyName(String),
     InvalidLedgerId(String),
-    InvalidEthereumContractAddress(String),
-    InvalidErc20HelperContractAddress(String),
+    InvalidContractAddress(String),
```
mbjorkqvist, line 127: Collapsing the variants is reasonable, but as it stands the failure loses which field was rejected.

All four contract addresses now yield `InvalidContractAddress(format!("ERROR: {e}"))` (here, `state.rs:512`/`520`/`536`/`560` and `init.rs:65-69`), and `Address::from_str`'s error doesn't name the field. These surface as a trap containing the `Debug` of this error — `main.rs:120`, `.expect("BUG: failed to initialize minter")` — so an operator who typos one of four addresses in a proposal gets `InvalidContractAddress("ERROR: …")` with no way to tell which one. `InvalidErc20HelperContractAddress` vs `InvalidEthereumContractAddress` at least narrowed it down before.

Keeping the single variant and putting the field name in the message would recover it: `InvalidContractAddress(format!("erc20_helper_contract_address: {e}"))`.

rs/ethereum/cketh/minter/src/lifecycle/upgrade.rs
```diff
@@ -30,6 +30,8 @@ pub struct UpgradeArg {
     pub deposit_with_subaccount_helper_contract_address: Option<String>,
     #[cbor(n(9), with = "icrc_cbor::nat::option")]
     pub last_deposit_with_subaccount_scraped_block_number: Option<Nat>,
+    #[n(10)]
```
mbjorkqvist: nit, but a hard-to-undo one: this arg is `ethereum_sweeper_contract_address`, while the `State` field, `MinterInfo`, the dashboard row and the template all use `sweeper_contract_address`. The existing divergence (`ethereum_contract_address` → `eth_helper_contract_address`) is a legacy accident; this is a brand-new field where both names are still free, and the Candid field name is the hardest one to change later. Worth picking one.

rs/ethereum/cketh/minter/src/state.rs
mbjorkqvist: nit: `upgrade` mutates field-by-field and only validates at the end, so a rejected upgrade leaves a partially-mutated `State`. That's harmless in-canister (the trap discards everything) and pre-existing, but the new cross-field distinctness rule makes it much easier to reach — and the new test at `state/tests.rs:612-630` depends on exactly that half-mutated state to set up its assertion. A comment noting that `upgrade` is not atomic and relies on the caller discarding `self` on error would be worth more than a restructure.

Also: there's no way to clear the sweeper address once set. That's consistent with the helper contracts, so presumably deliberate — just flagging it while the field has no consumers and the shape is still cheap to change.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Agreed — you're right that this newly-added rule now gates *every* historical upgrade during replay, and the failure mode (an un-upgradeable minter, recoverable only by shipping a wasm without the check) is severe enough to warrant confirming against the real history rather than the in-repo docs.

I'll take the verify-and-keep path, but in a **separate follow-up PR** to keep this one focused:
- confirm via `cketh_minter_dump_stable_memory` against the actual mainnet/staging/testnet audit logs that no historical configuration ever had a zero or duplicate contract address;
- reword the description from "the current config is distinct" to "no historical configuration is";
- make the zero/duplicate errors name the contract(s) by using the `LogScrapingId` the loop currently discards as `_id`.

Leaving this thread open to track that follow-up.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Done in 2539040 — kept the single `InvalidContractAddress` variant but each parse site now names its field in the message (`ethereum_contract_address` / `erc20_helper_contract_address` / `deposit_with_subaccount_helper_contract_address` / `ethereum_sweeper_contract_address`), at both init and upgrade, so a rejected address is attributable.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Update: I've pulled the error-labelling part into this PR after all — 651a7df. The loop no longer discards the `LogScrapingId`: the zero-address error now names the offending contract, and the duplicate error names both sides of the collision (`the ERC-20 and sweeper contract addresses must be distinct (both are 0x…)`), using the id's `Display` plus "sweeper" for the sweeper address.

The remaining two items stay for the separate follow-up PR — confirming against the historical mainnet/staging/testnet audit logs that no past configuration violated this new invariant, and rewording the description accordingly. Leaving this thread open to track those.

PR #11148, opened 2026-08-14
chore: Manually implement serde for `Cycles` and CBOR

eichhorl:
`serde_cbor` cannot represent a plain `u128` (that is greater than `u64::MAX`) and will fail when attempting to serialize one.

For the new pay-as-you-go pricing, cycles will be part of canister HTTP shares, which are cbor serialized in order to compute the signature. However, these values will always be far below `u64::MAX`.

Nevertheless, for consistency, this PR implements the `Serialize` and `Deserialize` traits manually as part of a new module, which falls back to a pair of `u64`.

### DIFF
--- a/rs/types/types/src/canister_http.rs
+++ b/rs/types/types/src/canister_http.rs
@@ -1117,6 +1117,11 @@ pub struct CanisterHttpPaymentReceipt {
     /// `per_replica_allowance - spent`. On free subnets it may exceed the (zero)
     /// allowance, since it is only used for cost accounting, but it may never
     /// exceed [`MAX_HTTP_OUTCALL_SPEND_FREE_SUBNET`].
+    ///
+    /// Encoded as a pair of `u64`s: this struct is part of the CBOR signed bytes
+    /// of [`CanisterHttpResponseReceipt`], and CBOR cannot represent a bare
+    /// `u128` above `u64::MAX`.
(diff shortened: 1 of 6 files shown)


rs/types/types/src/canister_http.rs
```diff
@@ -1409,6 +1369,59 @@ mod tests {
...
+
+        let amounts = [
+            Cycles::zero(),
+            Cycles::new(1),
+            Cycles::from(u64::MAX),
+            Cycles::from(u64::MAX) + Cycles::new(1),
+            Cycles::new(u128::MAX),
+        ];
+        let encodings: BTreeSet<_> = amounts.iter().map(|spent| signed_bytes(*spent)).collect();
+        assert_eq!(encodings.len(), amounts.len());
```
alin-at-dfinity, line 1426: Just curious, why no roundtrip encode-decode test?

alin-at-dfinity (approved):

rs/types/types/src/canister_http.rs
eichhorl: Good point, I assume it was because technically, these canister HTTP types are never deserialized with cbor. The only purpose is to generate a byte vector that can be signed.

But on the other hand, if that byte vector doesn't actually represent the struct (i.e. if a round trip failed), then that wouldn't be a good representation to be used in a signature.

I extended the test to do the roundtrip now, like you suggested.

PR #11151, opened 2026-08-16
fix(sns-wasm): actually overwrite existing SNS-specific upgrade path entries

Rachit2323:
## Problem

`insert_sns_specific_upgrade_path_entry` is used to set a one-off emergency upgrade step for a specific SNS (e.g. "if this SNS is stuck at version A, send it to version B instead"). If you try to change that step later (e.g. because B turned out to be wrong, so it should now go to C), the function only logs that it is "being overwritten" but never actually writes the new value. The old entry stays in place, while the caller is told the request succeeded.

## Fix

Actually write the new value when overwriting an existing entry, the same way the sibling function `insert_upgrade_path_entry` already does a few lines below it.

## Testing

Added `test_insert_sns_specific_upgrade_path_entry_overwrites_existing_entry`, which inserts an entry, overwrites it with a different target, and checks that the second value is the one actually returned. This test fails on the old code and passes with the fix.

### DIFF
--- a/rs/nns/sns-wasm/src/sns_wasm.rs
+++ b/rs/nns/sns-wasm/src/sns_wasm.rs
@@ -1940,7 +1940,7 @@ impl UpgradePath {
             .or_default()
             .entry(from)
         {
-            Entry::Occupied(occupied) => {
+            Entry::Occupied(mut occupied) => {
                 println!(
                     "Special Entry for {}  from {:?} to {:?} is being overwritten with new value {:?}",
                     sns_governance_canister_id,
@@ -1948,6 +1948,7 @@ impl UpgradePath {
                     occupied.get(),
                     to
                 );
+                occupied.insert(to);
             }
             Entry::Vacant(vacant) => {
                 vacant.insert(to);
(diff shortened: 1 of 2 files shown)


Rachit2323:
1. Changelog: Done. Added a note about this fix.

2. Breaking changes: No. This just fixes a function to actually do what it was already supposed to do. Nothing about how you call it changed.

3. Data migration: No. This only affects what happens the next time someone calls this function - it doesn't touch any existing data.

4. Security review: Not needed. This is a small internal logic fix, no new inputs or permissions involved. The automated security scan (zeropath-ai) already came back clean.

daniel-wong-dfinity-org-twin:
Hello, Rachit2323. Thank you for this. This is nearing the head of my queue. Sorry for the delay.

daniel-wong-dfinity-org-twin (approved):
Ah. So it only didn't work when `from` ALREADY had an SNS-specific successor, yes? And I guess we never tried to override in production, explaining this oversight was never noticed before, right? Nice catch.

# Rant

insert_sns_specific_upgrade_path_entry makes me sad, because what it's really doing is upsert, and upsert is kind of evil, because it clobbers silently, and in general, deletion should be INTENTIONAL.

Anyway, I'm not saying there should be a follow up here. This is just a general warning about the dangers of upsert, esp when it is misleadingly labeled "insert".

PR #11155, opened 2026-08-17
fix(canister_logging): reject a log_memory_limit below one page

mraszyk:
A non-zero `log_memory_limit` below 4096 was silently rounded up to one OS page (`DATA_CAPACITY_MIN`) by the log memory store, so a canister asking for e.g. 100 bytes was given, and charged for, a 4 KiB data region. Reject such a limit instead of quietly changing it: `update_settings` and `create_canister` now fail with `CanisterLogMemoryLimitIsTooLow`. A zero limit still disables canister logging, and the existing upper bound is unchanged.

The new `MIN_AGGREGATE_LOG_MEMORY_LIMIT` sits next to the existing aggregate log limits, with a compile-time assertion tying it to the ring buffer's `DATA_CAPACITY_MIN` so the accepted minimum and the storage granularity cannot drift apart.

The clamping in the store is kept as a release-build safety net, but is now funnelled through a single `data_capacity_for_limit` helper guarded by a `debug_assert!`, so a sub-minimum limit reaching the store is caught in tests rather than silently rounded. Store tests that exercised the clamp are updated to the minimum, and a new test pins the assertion itself.

### DIFF
--- a/rs/execution_environment/src/canister_manager.rs
+++ b/rs/execution_environment/src/canister_manager.rs
@@ -59,8 +59,8 @@ use ic_types::messages::{
 };
 use ic_types::{
     CanisterId, CanisterTimer, DEFAULT_AGGREGATE_LOG_MEMORY_LIMIT, MAX_AGGREGATE_LOG_MEMORY_LIMIT,
-    MAX_STABLE_MEMORY_IN_BYTES, MAX_WASM_MEMORY_IN_BYTES, MAX_WASM64_MEMORY_IN_BYTES, NumBytes,
-    NumInstructions, PrincipalId, SnapshotId, Time,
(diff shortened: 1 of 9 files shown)


alin-at-dfinity (approved):
The code change LGTM, I'm just wondering how useful this is. I mean yes, the user is technically asking for 100 bytes worth of logs and we're charging them for 4 KB, but in terms of cost this is entirely negligible (after all, the canister Wasm might want to allocate only 100 bytes of memory and get a full page instead).

And I guess it's also not an issue of correctness: yes, we won't be truncating their logs to 100 bytes, but they can do it themselves if they somehow really only want the last 100 bytes, not the full 4 KB.

Anyway, not strongly opposed, just thinking out loud.

mraszyk:
> I'm just wondering how useful this is

I believe it is more transparent to reject than silently round up. Allowing small values instead of rejecting could open the code up for weird bugs and I'd agree that setting small values is not useful which is why I went for rejecting.

PR #11158, opened 2026-08-17
feat: Delete canister directory via `UnflushedCheckpointOp::DeleteCanister`

mraszyk:
Introduce a new `UnflushedCheckpointOp::DeleteCanister` recorded when a canister is deleted during execution. The tip thread applies it in `flush_unflushed_checkpoint_ops`, removing the canister's directory from tip at the next flush (i.e., every round) instead of only at the next checkpoint, and in order relative to the other checkpoint operations.

`FilterTipCanisters` is retained as is: canisters dropped without a corresponding operation, i.e. during subnet splitting, are still removed from tip when creating a checkpoint.

### DIFF
--- a/rs/execution_environment/src/canister_manager.rs
+++ b/rs/execution_environment/src/canister_manager.rs
@@ -1362,13 +1362,9 @@ impl CanisterManager {
         // - its state is permanently deleted, and
         // - its cycles are discarded.
 
-        // Remove the canister from `ReplicatedState`; and record the removal, so that
-        // the canister's directory is also deleted from the tip.
+        // Remove the canister from `ReplicatedState`. This also records the removal, so
+        // that the canister's directory is deleted from the tip.
         let canister_to_delete = state.remove_canister(&canister_id_to_delete).unwrap();
-        state
-            .metadata
-            .unflushed_checkpoint_ops
-            .delete_canister(canister_id_to_delete);
         let canister_memory_allocated_bytes = canister_to_delete.memory_allocated_bytes();
 
         round_limits.subnet_available_memory.increment(
--- a/rs/execution_environment/src/canister_manager.rs
+++ b/rs/execution_environment/src/canister_manager.rs
@@ -1362,8 +1362,13 @@ impl CanisterManager {
         // - its state is permanently deleted, and
         // - its cycles are discarded.
 
-        // Remove the canister from `ReplicatedState`.
+        // Remove the canister from `ReplicatedState`; and record the removal, so that
+        // the canister's directory is also deleted from the tip.
         let canister_to_delete = state.remove_canister(&canister_id_to_delete).unwrap();
+        state
+            .metadata
+            .unflushed_checkpoint_ops
+            .delete_canister(canister_id_to_delete);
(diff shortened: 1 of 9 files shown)


rs/execution_environment/src/canister_manager.rs
```diff
@@ -1362,8 +1362,13 @@ impl CanisterManager {
         // - its state is permanently deleted, and
         // - its cycles are discarded.
 
-        // Remove the canister from `ReplicatedState`.
+        // Remove the canister from `ReplicatedState`; and record the removal, so that
+        // the canister's directory is also deleted from the tip.
         let canister_to_delete = state.remove_canister(&canister_id_to_delete).unwrap();
```
schneiderstefan, line 1367: Could this be inside of `remove_canister` to make it more robust when things get moved around?

mraszyk: Good point: https://github.com/dfinity/ic/pull/11158/commits/dbe9eacb2a421a02ba95ee4f715b77870156e005

rs/replicated_state/tests/replicated_state.rs
```diff
@@ -1255,6 +1255,13 @@ fn split() {
     assert_eq!(expected, state_b);
 }
 
+/// Drops a canister from the state the same way `online_split()` does, i.e. without
```
schneiderstefan, line 1258: Why can't online_split just do it properly by calling remove_canister?

mraszyk: Pasting Claude's reply: It could, but it wouldn't buy us anything and it's awkward where the drop happens:

  1. online_split() drops canisters in bulk, without a &mut self. It consumes and destructures self (deliberately, so a new field forces an explicit decision) and drops the non-hosted canisters via CanisterStates::retain()
  (rs/replicated_state/src/replicated_state.rs:1644). remove_canister() needs &mut ReplicatedState and works one canister at a time, so we'd need an extra pass over all canister IDs (a second routing-table lookup per canister)
  before the destructuring, or to hoist the drop out of online_split() altogether.
  2. FilterTipCanisters is needed for the split regardless. Snapshots live in a sibling directory (snapshots/<canister_id>/<snapshot_id>), not inside canister_states/<canister_id>, so UnflushedCheckpointOp::DeleteCanister does not
  remove the snapshots of the canisters that migrated away — filter_tip_snapshots() has to. Since we send that request anyway, recording ~half a subnet's worth of DeleteCanister ops would just duplicate the canister half of the
  filtering (a per-canister remove_dir_all either way).
  3. The scheduling-priority part is already covered. SystemMetadata::online_split() drops the priorities of non-hosted canisters via subnet_schedule.split(), which is the other thing remove_canister() does.

  If you'd rather have the strict invariant "canisters only ever leave canister_states through remove_canister()", the consistent version would be to record the delete ops in the split and drop filter_tip_canisters() entirely,
  keeping only filter_tip_snapshots(). Happy to do that as a follow-up, though I'd slightly prefer keeping the tip filtering as a safety net for the split.

schneiderstefan (approved):

PR #11168, opened 2026-08-17
fix: validate the Wasm memory in create_execution_state

mraszyk:
`Hypervisor::create_execution_state` is the single place that assembles the memories of a new execution state. Validate there that the Wasm memory it installs is compatible with the Wasm module that it is installed with:
- the Wasm memory does not exceed the maximum Wasm memory for the Wasm execution mode of the canister module;
- the Wasm memory is empty if the canister module declares no Wasm memory;
- the Wasm memory is at least as large as the minimum Wasm memory declared by the canister module.

To this end, the embedder reports whether a canister module declares a Wasm memory (derived like `is_wasm64` from the memory export of the instrumented module) via `SerializedModule`, and
`WasmExecutor::create_execution_state` returns the new `CreatedExecutionState` struct instead of a 3-tuple. A failed validation is reported as the new `HypervisorError::InvalidWasmMemory` and charged the same as a successful creation of the execution state.


rs/execution_environment/benches/lib/src/common.rs
```diff
@@ -53,6 +55,12 @@ pub enum Wasm64 {
     Disabled,
 }
 
+/// The number of Wasm pages of the heap declared by the Wasm64 benchmark modules.
+///
+/// This is the largest heap a Wasm64 canister module may declare: a module
+/// declaring more is rejected when its execution state is created.
+pub const WASM64_HEAP_NUM_PAGES: u64 = MAX_WASM64_MEMORY_IN_BYTES / WASM_PAGE_SIZE_IN_BYTES as u64;
```
michael-weigelt: Shouldn't `MAX` be in the name?

michael-weigelt (approved):

rs/execution_environment/benches/lib/src/common.rs
mraszyk: https://github.com/dfinity/ic/pull/11168/commits/9967d758102b34d453446cb04e3776b7af836fe7

PR #11171, opened 2026-08-17
fix: correctly map canister error calls

nmattia:
This makes all the ic-cdk error mapping helpers exhaustive and ensures canister errors are reported correctly; for instance, a "insuffient cycles balance" error should _not_ be treated as a transient error.

See code for exact behavior.


mbjorkqvist (approved):

mraszyk:
> a "insuffient cycles balance" error should _not_ be treated as a transient error.

I'm not sure about this: the corresponding error code is mapped to [SysTransient](https://github.com/dfinity/ic/blob/2cc18ad24c6429f7f57ac5e7f8fac24ab0c6530e/packages/ic-error-types/src/lib.rs#L80) in the protocol for asynchronous errors and the synchronous error is reported as the return value of 2 (SysTransient) of the corresponding system API; so I wouldn't remap the reject code in the helper.

packages/pocket-ic/test_canister/src/canister.rs
```diff
@@ -52,12 +52,18 @@ impl RejectionCode {
...
+        // The call was rejected and the system assigned a reject code; report it.
         CallError::CallRejected(rejected) => (
             RejectionCode::from_raw(rejected.raw_reject_code()),
             rejected.reject_message().to_string(),
         ),
-        // Nothing reached the callee, so there is no reject code to report.
-        other => (RejectionCode::Unknown, other.to_string()),
+        // The callee replied, but the response could not be decoded into the
+        // expected type, so the callee did not honor its interface: a canister-side error.
+        CallError::CandidDecodeFailed(e) => (RejectionCode::CanisterError, e.to_string()),
```
mraszyk: I'm not sure about this reject code: the following comment applies here, too - Neither of these produced a response from the callee, so there is no callee-assigned reject code to report; surface them as `Unknown`.

mraszyk (commented):
The link https://internetcomputer.org/docs/references/ic-interface-spec#reject-codes does not work for me: use https://docs.internetcomputer.org/references/ic-interface-spec/https-interface/#reject-codes

There's also a reject code SYS_UNKNOWN (6) produced by the protocol which is different from `Unknown` meaning that the reject code could not be parsed.

daniel-wong-dfinity-org-twin (approved):
Sorry if I'm beating a dead horse, but I would like to again advocate for minimizing copy & paste.

basvandijk (approved):

packages/pocket-ic/test_canister/src/canister.rs
```diff
@@ -43,21 +46,29 @@ impl RejectionCode {
...
             RejectionCode::from_raw(rejected.raw_reject_code()),
             rejected.reject_message().to_string(),
         ),
-        // Nothing reached the callee, so there is no reject code to report.
-        other => (RejectionCode::Unknown, other.to_string()),
+        // None of these carry a system-assigned reject code, so there is no
+        // faithful code to report; surface them as `Unknown`.
+        err @ (CallError::CandidDecodeFailed(_)
+        | CallError::InsufficientLiquidCycleBalance(_)
+        | CallError::CallPerformFailed(_)) => (RejectionCode::Unknown, err.to_string()),
```
mraszyk, line 71: an extra `RejectionCode::None` could be more precise here (in particular, for `CandidDecodeFailed`) since the IC did not produce any reject code explicitly

rs/tests/networking/canister_http_correctness_test.rs
```diff
@@ -2693,7 +2693,11 @@ where
                         RejectionCode::DestinationInvalid => RejectCode::DestinationInvalid,
                         RejectionCode::CanisterReject => RejectCode::CanisterReject,
                         RejectionCode::CanisterError => RejectCode::CanisterError,
-                        RejectionCode::NoError | RejectionCode::Unknown => {
+                        // `ic_agent::agent::RejectCode` has no counterpart for these; the
```
mraszyk, line 2696: IC agent is not aware of `SysUnknown`? are we using an outdated agent?

packages/icrc-ledger-client-cdk/src/lib.rs
```diff
@@ -8,16 +8,30 @@ pub use icrc_ledger_client::{ICRC1Client, Runtime};
...
+        // The system (or the callee) already rejected the call and assigned a reject
+        // code; surface it unchanged.
         Error::CallRejected(rejected) => (
             rejected.raw_reject_code() as i32,
             rejected.reject_message().to_string(),
         ),
-        // A response that cannot be decoded means the callee misbehaved.
+        // The callee replied, but its response did not decode into the expected type,
+        // so it did not honor its interface: treat it as a canister-side error.
         Error::CandidDecodeFailed(err) => (RejectCode::CanisterError as i32, err.to_string()),
```
mraszyk, line 25: this is now incosistent with some other such mappings in this PR

mraszyk (approved):

PR #11172, opened 2026-08-17
feat(registry): Allow blanking an existing Cloud Engine's replica_version_id.

pietrodimarco-dfinity:
## Motivation

A blank `SubnetRecord.replica_version_id` is the normal, documented state for a Cloud Engine: when blank, the engine's replica version comes from `StandardEngineReplicaVersionRecord` instead of being pinned (see the field docs in `subnet.proto`). `create_subnet` creates engines that way.

But an engine that *does* pin a version could not be put back on the standard version. `deploy_guestos_to_all_subnet_nodes` is the only endpoint that writes `replica_version_id`, and it required that version to be elected — a blank version never is, so the call traps with `Replica version '' is NOT elected`. That applies to every caller, Governance included, so there was no route at all (and `update_subnet` deliberately does not carry `replica_version_id`).

## Changes

Make the elected check conditional:

```rust
if payload.replica_version_id.is_empty() {
    self.check_blank_replica_version_id_is_allowed(subnet_id);
} else {
    check_replica_version_is_elected(self, &payload.replica_version_id);
}
```

`check_blank_replica_version_id_is_allowed` asserts the same three conditions under which the `SubnetRecord` invariant tolerates a blank id (`invariants::replica_version`), so the mutation cannot produce state the invariant would then reject:

1. `IS_BLANK_REPLICA_VERSION_ID_FOR_CLOUD_ENGINES_ENABLED` is on,
2. the subnet is a `CloudEngine`,
3. a `StandardEngineReplicaVersionRecord` exists to determine the version from.

Non-blank versions still have to be elected. The rule is caller-agnostic — Governance and the engine controller canister get the same treatment, and the existing engine-controller-only `CloudEngine` guard is untouched.

Incidental: `get_standard_engine_replica_version_record` becomes `pub(crate)` (it was private to its own module).

## Tests

Six unit tests: the engine controller and Governance can each blank a Cloud Engine; it panics when the feature is disabled, when there is no `StandardEngineReplicaVersionRecord`, and for a non-`CloudEngine` subnet; plus a non-blank unelected version still panics, so the new branch cannot silently weaken the existing check.

Verification beyond `rustfmt` is left to CI.


rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -38,7 +39,18 @@ impl Registry {
...
+        // StandardEngineReplicaVersionRecord instead (see
+        // SubnetRecord.replica_version_id in subnet.proto). Nothing blank can be
+        // elected, so it is validated against the conditions that let the
+        // SubnetRecord invariant tolerate it (see invariants::replica_version)
+        // rather than against the elected versions.
+        if payload.replica_version_id.is_empty() {
+            self.check_blank_replica_version_id_is_allowed(subnet_id);
+        } else {
+            check_replica_version_is_elected(self, &payload.replica_version_id);
+        }
```
daniel-wong-dfinity-org-twin: Can you spin this out to a method named something like `validate_payload`?

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -38,7 +39,18 @@ impl Registry {
...
         }
 
-        check_replica_version_is_elected(self, &payload.replica_version_id);
+        // A blank replica_version_id names no version at all: it takes a Cloud
+        // Engine off a pinned version and defers to
+        // StandardEngineReplicaVersionRecord instead (see
+        // SubnetRecord.replica_version_id in subnet.proto). Nothing blank can be
+        // elected, so it is validated against the conditions that let the
+        // SubnetRecord invariant tolerate it (see invariants::replica_version)
+        // rather than against the elected versions.
```
daniel-wong-dfinity-org-twin: The code seems pretty self-explanatory to me -> I would just forego this comment entirely.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -65,6 +77,35 @@ impl Registry {
         // Check invariants before applying mutations
         self.maybe_apply_mutation_internal(mutations)
     }
+
+    /// Panics unless `subnet_id` is allowed to have a blank
+    /// `replica_version_id` — i.e. it is a Cloud Engine, the feature is
+    /// enabled, and there is a StandardEngineReplicaVersionRecord to resolve
+    /// the blank against. Without the last one, blanking would leave the subnet
+    /// with no replica version at all (and trip the SubnetRecord invariant).
+    fn check_blank_replica_version_id_is_allowed(&self, subnet_id: SubnetId) {
```
daniel-wong-dfinity-org-twin: Can we instead name this `check_engine_can_have_blank_replica_version_id` or something like this?

daniel-wong-dfinity-org-twin (commented):
Still need to review tests. Will leave feedback for that soon.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -74,13 +115,30 @@ mod tests {
...
+    };
     use ic_registry_subnet_type::SubnetType;
+    use ic_registry_transport::insert;
     use ic_test_utilities_types::ids::subnet_test_id;
     use ic_types::ReplicaVersion;
     use maplit::btreemap;
 
+    /// A second elected version, so that a
+    /// StandardEngineReplicaVersionRecord's two version IDs can differ (as its
+    /// invariant requires). Replica version IDs are git commit IDs.
```
daniel-wong-dfinity-org-twin:
```suggestion
    /// A second elected version for
    /// StandardEngineReplicaVersionRecord.
```

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -122,6 +180,45 @@ mod tests {
         }
     }
 
+    /// A payload that takes the subnet off a pinned version, so that it follows
+    /// the StandardEngineReplicaVersionRecord instead.
+    fn blank_deploy_payload(subnet_id: SubnetId) -> DeployGuestosToAllSubnetNodesPayload {
```
daniel-wong-dfinity-org-twin: This seems misleading to me. It's not the deploy that's blank, just the replica_version_id. How about `new_blank_replica_version_id_payload` ?

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -122,6 +180,45 @@ mod tests {
         }
     }
 
+    /// A payload that takes the subnet off a pinned version, so that it follows
+    /// the StandardEngineReplicaVersionRecord instead.
```
daniel-wong-dfinity-org-twin:
```suggestion
    /// Creates a payload that makes the subnet (presumably, an engine)
    /// follow StandardEngineReplicaVersionRecord instead.
```

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -122,6 +180,45 @@ mod tests {
...
+        }
+    }
+
+    /// Installs the record a blank `replica_version_id` resolves against. Its
+    /// two version IDs must differ and both must be elected, hence the extra
+    /// election.
+    fn add_standard_engine_replica_version_record(registry: &mut Registry) {
+        registry.maybe_apply_mutation_internal(vec![insert(
+            make_replica_version_key(OTHER_REPLICA_VERSION_ID).as_bytes(),
+            ReplicaVersionRecord::default().encode_to_vec(),
```
daniel-wong-dfinity-org-twin: This is about to go away in another PR by David. How about, instead of just OTHER_REPLICA_VERSION_ID, we have OLD_REPLICA_VERSION_ID and NEW_REPLICA_VERSION_ID. And let's give them both realistic value. E.g. just pick any two real commit IDs in this repo.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -65,6 +66,47 @@ impl Registry {
         // Check invariants before applying mutations
         self.maybe_apply_mutation_internal(mutations)
     }
+
+    fn validate_deploy_guestos_payload(
+        &self,
+        subnet_id: SubnetId,
```
daniel-wong-dfinity-org-twin: This is already in payload, is it not??

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -134,10 +135,11 @@ mod tests {
     use ic_types::ReplicaVersion;
     use maplit::btreemap;
 
-    /// A second elected version, so that a
-    /// StandardEngineReplicaVersionRecord's two version IDs can differ (as its
-    /// invariant requires). Replica version IDs are git commit IDs.
-    const OTHER_REPLICA_VERSION_ID: &str = "63d086714a1e2bc6b0615008d5582f527d554cd3";
+    // Two elected replica versions, used as the old and new versions of a
+    // StandardEngineReplicaVersionRecord (whose invariant requires that they
+    // differ). Replica version IDs are git commit IDs.
```
daniel-wong-dfinity-org-twin:
```suggestion
    // StandardEngineReplicaVersionRecord.
```

I think we need NOT belabor these facts.

rs/registry/canister/src/mutations/do_deploy_guestos_to_all_subnet_nodes.rs
```diff
@@ -65,6 +66,47 @@ impl Registry {
...
+    /// the blank against. Without the last one, blanking would leave the subnet
+    /// with no replica version at all (and trip the SubnetRecord invariant).
+    fn check_engine_can_have_blank_replica_version_id(&self, subnet_id: SubnetId) {
+        assert!(
+            is_blank_replica_version_id_for_cloud_engines_enabled(),
+            "{LOG_PREFIX}do_deploy_guestos_to_all_subnet_nodes: a blank replica_version_id is \
+             not enabled yet.",
+        );
+
+        let subnet_record = self.get_subnet_or_panic(subnet_id);
```
daniel-wong-dfinity-org-twin, line 91: Maybe I'm suggesting a premature optimization, but this is the second fetch of the SubnetRecord within deploy_guestos, isn't it?

daniel-wong-dfinity-org-twin: Yeah, this suggestion is premature optimization. Let's not do anything about this.

# guest_launch_measurements

Do not add new issues to threads. For new issues, start a new thread. Anyway, I think that's probably resolved, so do not do that in this case. I'll start a thread on that if there are problems.

daniel-wong-dfinity-org-twin (approved):

PR #11173, opened 2026-08-17
feat(nns): support reserved_cycles_limit in canister settings proposals

Rachit2323:
## The problem

Every canister has a setting called "reserved cycles limit" - basically a cap on how many cycles it's allowed to keep in reserve. You can already change almost every other canister setting through an NNS vote (who controls it, memory limits, etc), but not this one. The code had it hardcoded to "off," with a comment saying it wasn't supported yet.

## What I did

Hooked this setting up the same way the existing settings already work, following the same pattern used for a similar setting called wasm_memory_threshold:

- Added the new field everywhere it needs to exist (the data format, the public interface, the code that converts between them)
- Made the two places that update canister settings actually use it instead of ignoring it

## Testing

- Updated the existing tests so they check this new setting too
- Added a new test for the second place that uses it
- Ran everything - all tests pass


rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs
daniel-wong-dfinity-org-twin, line 1: I am very confused by all the changes in this file. This is supposed to be derived from governance.proto, which only 1 new line. Therefore, I do not understand why there are a bunch of comment formatting changes here. Please, explain.

rs/nns/governance/src/proposals/create_canister_and_install_code.rs
```diff
@@ -287,3 +288,26 @@ impl From<CreateCanisterAndInstallCode> for SelfDescribingValue {
             .build()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_reserved_cycles_limit_passes_through_to_root_settings() {
```
daniel-wong-dfinity-org-twin: Wow. This test is pretty niche! I mean, it's not harmful, but I think our pattern would be to test the ENTIRE try_from method, not just one field that it's supposed to forward. Then again, I guess many of our converters have no tests, and THAT pattern should NOT be followed...

rs/nns/governance/src/proposals/create_canister_and_install_code.rs
```diff
@@ -287,3 +288,26 @@ impl From<CreateCanisterAndInstallCode> for SelfDescribingValue {
             .build()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
daniel-wong-dfinity-org-twin: Most production .rs files in this dir have a corresponding *_tests.rs file. For example, [here]. Do likewise.

[here]: https://github.com/dfinity/ic/blob/master/rs/nns/governance/src/proposals/add_or_remove_node_provider.rs#L235

daniel-wong-dfinity-org-twin (commented):
Thank you!

rs/nns/governance/src/gen/ic_nns_governance.pb.v1.rs
Rachit2323: This file gets auto-created from governance.proto - I don't write it by hand. When I added the one new field, I used a slightly different way to regenerate it than the repo normally uses. Same tool, but it ended up rewriting some old comments too (just formatting - bullet points, spacing, that kind of thing). Not something I did on purpose.

I'll regenerate it the correct way so the diff only shows the one real line I added.

Rachit2323:
Closing — #11389 already shipped this exact feature (reserved_cycles_limit on UpdateCanisterSettings and CreateCanisterAndInstallCode) and is merged to master.

PR #11174, opened 2026-08-17
fix(CI): verify that commit-sha is the head of pr-number in ci-kickoff-manual

basvandijk:
The `ci-kickoff-manual` workflow takes a `commit-sha` and a `pr-number` input, with the intention that the commit belongs to that PR. Nothing checked this.

A mismatch would run CI on unrelated code while reporting commit statuses and PR comments against the PR, and — because the auto-fix jobs in `ci-pr-only.yml` do `git push origin HEAD:'${{ inputs.head-ref }}'` — would let did-check / buf-format / cargo-lock fixups derived from a foreign tree be pushed onto the PR's head branch.

## Changes

- `get-pr-refs` already fetches the PR, so it is extended to verify that `commit-sha` is the **head commit** of the PR, and renamed to `verify-inputs-and-get-pr-refs`.
- It also rejects a `pr-number` that isn't a positive decimal integer or a `commit-sha` that isn't a full 40-hex SHA, and refuses non-open PRs.
- Neither `github-script` step interpolates the inputs into its JS body any more — both the verification job and `post-comment-with-ci-link` read them from `env:`, removing a script-injection surface. The remaining `${{ inputs.* }}` uses are in `with:` / `ref:` positions, where the gate above guarantees the values are a 40-hex SHA and a decimal integer.
- The `confirm` checkbox is now enforced: for boolean `workflow_dispatch` inputs `required: true` only means the field is present, not that it is ticked, so the acknowledgement that the PR was reviewed at this commit was optional in practice.
- Every job now depends on the verification. Previously only `ci-pr-only` did, so a verification failure did not stop `ci-main`, the PR comment, or the Slack notification — a failed job does not cancel independent jobs.

## Don't push autofixes to a branch named after a fork's branch

Follow-up on a review comment, in the same area: the pushing jobs in `ci-pr-only.yml` (`autofix`, `lock-generate`, `generate-config-fixtures`) do `git push origin HEAD:<head-ref>`. `actions/checkout` defaults to `repository: github.repository`, so `origin` is always this repository, while `head-ref` is `pr.head.ref` — an unqualified branch name.

On the `pull_request` path that is fine: `ci-kickoff.yml` only calls `ci-pr-only` when `head.repo.full_name == github.repository`. On the `ci-kickoff-manual` path — which exists for external contributions — `head-ref` is the *fork's* branch name, so the push creates a branch of that name **here** instead of updating the PR. It uses the `PR_CREATION_BOT` app token, so the job's `contents: read` does not restrain it. Two notable cases: a fork branch called `main`/`master` is rejected by the branch ruleset (so the job just fails), and a fork branch called `dev-gh-*` would land on a name that is a `push` trigger for Kickoff, running CI Main with secrets on unreviewed code.

The fix gates the push rather than the jobs:

- `ci-pr-only.yml` takes a new `head-repo` input, defaulting to `github.event.pull_request.head.repo.full_name` and supplied on the manual path from the `pulls.get` response the verification job already has (empty if the fork was deleted, which fails closed).
- The three jobs commit and push only when it equals `github.repository`; otherwise they print the staged diff and still exit non-zero. The commit is created only on the pushing path, so the printed diff is the patch the contributor actually has to apply (a `git diff --cached` after committing would print nothing).

Guarding the jobs with `if: github.event_name == 'pull_request' && …` instead would work, but `autofix` is the only place the `//pre-commit:*` checks run (`shfmt`, `ruff`, `buf`, `buf-breaking`, `do-not-merge`, `buildifier`, `gazelle`, `rustfmt`), so external PRs would get a green `ci-pr-only-status` with no lint gate at all. Gating the push keeps those checks gating exactly the PRs that need them most, and leaves the contributor an actionable diff in the log.

`head-ref` and `head-repo` are also passed to the scripts through the environment now instead of being interpolated: git allows single quotes in branch names, so `HEAD:'${{ inputs.head-ref }}'` was itself a fork-controlled shell-injection surface.

### DIFF
--- a/.github/workflows/ci-kickoff-manual.yml
+++ b/.github/workflows/ci-kickoff-manual.yml
@@ -17,8 +17,8 @@ on:
         type: boolean
 
 jobs:
-  get-pr-refs:
-    name: Get PR Refs
+  verify-inputs-and-get-pr-refs:
+    name: Verify Inputs and Get PR Refs
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-kickoff-manual.yml
```diff
@@ -27,27 +27,72 @@ jobs:
...
               repo: context.repo.repo,
               pull_number: prNumber
             });
+            if (pr.state !== 'open') {
+              core.setFailed(`PR #${prNumber} is ${pr.state}, refusing to run CI on it`);
+              return;
+            }
+
+            // GitHub returns at most 250 commits here; for larger PRs only the
+            // head commit can be verified this way.
```
nmattia: wouldn't it make sense to confirm the run is for the tip of the branch anyway?

nmattia (approved):

.github/workflows/ci-kickoff-manual.yml
basvandijk: Yeah we could have that more strict restriction. I'll push something...

basvandijk: Done in f2c870ce6c — `commit-sha` must now equal the PR head SHA, otherwise the job fails. That also let the `pulls.listCommits` pagination (and its 250-commit caveat) go, since head equality only needs the `pulls.get` response.

.github/workflows/ci-kickoff-manual.yml
```diff
@@ -27,27 +27,53 @@ jobs:
...
             const { data: pr } = await github.rest.pulls.get({
               owner: context.repo.owner,
               repo: context.repo.repo,
               pull_number: prNumber
             });
+            if (pr.state !== 'open') {
+              core.setFailed(`PR #${prNumber} is ${pr.state}, refusing to run CI on it`);
+              return;
+            }
+
```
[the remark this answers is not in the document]
basvandijk: Good catch, fixed in cd432c110e — but by disabling the push rather than by failing the run, since the manual path exists precisely for fork PRs.

Confirmed the mechanism: `actions/checkout` defaults to `repository: github.repository`, so `origin` is always `dfinity/ic`, while `head-ref` is `pr.head.ref` — an unqualified branch name. The push also uses the `PR_CREATION_BOT` app token, so the job's `contents: read` doesn't restrain it. Two refinements:

- It's *create*, not overwrite: the push isn't forced, so a collision with an existing branch here is rejected as non-fast-forward unless the fork commit descends from it. `main`/`master` are covered by the `branch-ruleset-level-2` ruleset, so those fail loudly.
- Worse than a stray branch: a fork branch named `dev-gh-*` would land on a name that is a `push` trigger for Kickoff, running CI Main with secrets on unreviewed code.
- It isn't only `autofix` — `lock-generate` and `generate-config-fixtures` do the identical push.

So `ci-pr-only.yml` now takes a `head-repo` input (from the event on the `pull_request` path, from this job's `pulls.get` response on the manual path, empty if the fork was deleted → fails closed), and all three jobs push only when it equals `github.repository`; otherwise they print the staged diff and still exit non-zero.

I deliberately did not gate the jobs themselves: `autofix` is the only place the `//pre-commit:*` checks run (`shfmt`, `ruff`, `buf`, `buf-breaking`, `do-not-merge`, `buildifier`, `gazelle`, `rustfmt`), so skipping it for fork PRs would give external contributions a green `ci-pr-only-status` with no lint gate at all.

While there: `head-ref` and `head-repo` now reach the scripts through the environment instead of `${{ }}` interpolation — git permits single quotes in branch names, so `HEAD:'${{ inputs.head-ref }}'` was a fork-controlled shell-injection surface too.

basvandijk:
Handled the suppressed comments from the last two Copilot reviews in 7a22de0da9 — both were right.

**`git diff --cached` printed after the commit** (suppressed ×3, one per pushing job, on review of `cd432c110e`): it would have printed nothing, since `git commit` had already cleared the index. Rather than switching to `git show`, the commit is now created only on the branch that also pushes it — so no pointless commit on the fork path, and the printed diff is exactly the patch the contributor has to apply.

**`confirm` was never checked** (suppressed on review of `f2c870ce6c`): correct, and the reasoning is the important part — for boolean `workflow_dispatch` inputs `required: true` only means the field is present in the dispatch, not that the box is ticked, so the "I reviewed the PR at this commit" acknowledgement was optional in practice. The verification job now fails unless `CONFIRM` is `true`, checked before the API call. It reads the value from `env:` like the other two inputs.

nmattia (approved):

PR #11175, opened 2026-08-17
fix: deflake //rs/consensus:integration_test

basvandijk:
## Root cause

`//rs/consensus:integration_test` has `timeout = "long"` (900 s), but **successful** CI runs in the last week took **7:01 – 13:36 (421 – 816 s)** — up to 91 % of the budget. Under ordinary runner load the target crosses the deadline and bazel kills it; the retry passes, so the target reports FLAKY. All 4 flakes of the last week are identical: attempt 1 ends in `-- Test timed out --`, 13 of the 14 long-running tests report `ok`, and `slow_checkpointing_at_upgrade_boundary` is the only one that never finishes. 18 flakes since 2026-05-01, the first on 2026-06-23 — two weeks after that test landed in #10347. Before it landed the target ran in 0:57 – 2:40.

That one test is ~87 % of the target's runtime. Measured locally on an idle 32-core box:

| measurement | result |
| --- | --- |
| the test alone, 10 runs | 195.9 – 200.8 s, avg 197.7, dev 1.8 |
| the other 17 tests (`--skip`) | 29.7 s |
| whole target, 4 concurrent runs | 232.0 – 234.6 s, dev 0.9 |

It is not a race or a deadlock: 14 local runs, ~1 % spread, and the config is fixed (`random_seed: 0`, `GlobalMessage`/`Sequential`, 4 nodes). The test is simply long — `ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP = 70` forces ~219 heights of 4-node consensus (upgrade at `2 × (74 + 1) = 150`, then blocks to `149 + 70 = 219`) — and the CI machines are 2 – 3.5× slower under load, which is enough to cross a wall this close. Raising the timeout is what #9115 already did in March (`moderate` → `long`); this PR goes after the runtime instead.

## Changes

* **Its own target.** `slow_checkpointing_at_upgrade_boundary` moves to `tests/slow_checkpointing.rs` / `//rs/consensus:slow_checkpointing_test`, so it owns its timeout budget instead of sharing one with 17 other tests. This required moving `TestRunner` out of `tests/integration.rs` into `tests/framework/test_runner.rs` so both crate roots can use it (the BUILD file already shares `glob(["tests/**"])` across crate roots).
* **No chain keys for this test.** Generating chain key transcripts, dealings and pre-signatures with real crypto is more than half of the test's runtime and has nothing to do with what it asserts. `setup_subnet` now takes the chain key ids, and `TestRunner::without_chain_keys()` selects none. The default is unchanged, so every other test keeps all four keys; `run_n_rounds_and_check_pubkeys` now asserts that keys are configured, so it can't pass vacuously.
* **A real ticket reference.** The `ICSUP-XXX` placeholder in the test's doc comment is replaced with the incident (subnet `3hhby`, 2026-05-22) and a link to #10347, the fix this test guards.

## Result

`bazel test --test_output=errors --runs_per_test=3 --jobs=3` on the affected targets — all pass:

| target | before | after | share of its 900 s timeout |
| --- | --- | --- | --- |
| `integration_test` | 234 s | **48 s** (43.4 – 47.7) | 5 % |
| `slow_checkpointing_test` | — | **104 s** (103.0 – 104.1) | 12 % |

Both targets keep `timeout = "long"`, so neither changes which CI events run it (`--test_timeout_filters=short,moderate` excludes `long` in merge groups either way). Lowering them now that they are fast is a possible follow-up.

## Notes for the reviewer

One hazard I noticed but did **not** address: `IDkgImpl::on_state_change` drives `purge_inactive_transcripts` off `Instant::now()` on a 60 s *wall-clock* interval inside an otherwise virtual-time simulation. Dropping the chain keys removes it from this test as a side effect, but it still applies to every other test in the suite.


basvandijk:
Addressed in 553a55e36a.

**1. `Ordering::Equal`-only release of the override** (and the same point on the `Ordering::Greater` arm) — agreed, and worse than "can intermittently panic": a node that skips the release keeps its state height capped forever and never creates a CUP share, so with two such nodes out of four no CUP can be aggregated and the test hangs until bazel kills it. Two independent ways to get there: a single `ConsensusDriver::step()` can validate the notarizations and finalizations of several heights at once, and `is_checkpointing` was one flag shared by all four nodes, so once the first node released and made progress the others could arrive at the bound with it already `false`.

The predicate now releases the override **per node, as soon as that node reaches the bound**, which removes the ordering dependency entirely. The assertions that still hold in that formulation are kept: blocks up to the bound carry the frozen certified height, consensus gets at most one block past the bound (`<=` rather than `==`, per the suppressed comment), and that block also still carries the frozen certified height. The now-vacuous `assert!(!is_checkpointing)` is gone — a lagging node legitimately learns about heights past the bound by validating other nodes' artifacts, so "still frozen implies at most the stall height" was never a valid per-node invariant.

**2. `ICSUP-XXX` placeholder** — replaced with the incident (subnet `3hhby`, 2026-05-22) and a link to #10347, which is the fix this test guards.

Re-verified with `bazel test --test_output=errors --runs_per_test=3 --jobs=3`: `slow_checkpointing_test` 103.3 – 104.8 s, `integration_test` 44.6 – 45.9 s, all green. Note the multi-height skip is a latent path — it has never reproduced locally (17+ runs), so this is a robustness change rather than a fix I can demonstrate failing beforehand.

pierugo-dfinity (commented):
Thanks for proposing a fix. I am happy to approve the first two changes (own target + no chain keys), though I am not sure about the third one (concerning the stop predicate).
I agree that one node could skip the stall height, but I do not think that 2 nodes out of 4 could (and thus stall the test): If this was true, then it would mean that there was a finalization for a height larger than the stall height. But a finalization requires `2f+1 == 3` nodes to have a full notarization for that height. And to have a full notarization at that height, we need `2f+1 == 3` notarization shares for it, which *need* to have released the state override otherwise they hit `ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP`.
Conclusion: the driver needs to uncap the certified height for at least 3 nodes before even being able to observe a finalization for a height higher than the stall height. Those 3 nodes can then create a CUP together.

Except if you have a concrete example of a flake caused by this, I would keep the current implementation which, for me, describes the behaviour more clearly by distinguishing the 3 steps (lower, equal, greater).

basvandijk:
You're right, and thanks for working through it — reverted in af5991c696.

My "two nodes out of four" claim didn't account for the fact that observing a finalization above the stall height is itself gated on the override being released. A still-capped node hits `ACCEPTABLE_NOTARIZATION_CERTIFICATION_GAP` and creates no notarization share for that height, so 2f+1 == 3 nodes must have released before any finalization above the stall height can exist — which leaves at most one node able to skip the observation, and the three that released are exactly the threshold committee needed for the CUP. No stall.

The `Ordering::Greater` assertions hold for the skipping node too, for the same reason: with the certified height sitting at the upgrade height while halting, the gap bound stops notarization at the stall height plus one, so `== stall + 1` is not just safe but the stronger check. I have no repro, so the three-way match stays as it was.

Kept from that commit: only the `ICSUP-XXX` placeholder replacement (now pointing at subnet `3hhby`, 2026-05-22 and #10347). The PR description is updated accordingly.

Re-verified after the revert with `bazel test --test_output=errors --runs_per_test=3 --jobs=3`: `slow_checkpointing_test` 106.7 – 107.8 s, `integration_test` 45.3 – 48.5 s, all green.

pierugo-dfinity (approved):
Thanks 🙏  Do you have an idea why the other tests went from 29.7s to 48s?

basvandijk:
@pierugo-dfinity 

> Do you have an idea why the other tests went from 29.7s to 48s?

Yes, for the 29.7s measurement Claude ran the `integration_test` on its own:

```
bazel test //rs/consensus:integration_test --test_arg=--skip --test_arg=slow_checkpointing_at_upgrade_boundary --cache_test_results=no --test_output=summary 2>&1 | tail -6

OUT
INFO: 2 processes: 2104 action cache hit, 1 internal, 2 linux-sandbox.
INFO: Build completed successfully, 2 total actions
//rs/consensus:integration_test                                          PASSED in 29.7s
```

While for the 48s measurement Claude ran both test at the same time: 
```
bazel test --test_output=errors --runs_per_test=3 --jobs=3 --cache_test_results=no //rs/consensus:slow_checkpointing_test //rs/consensus:integration_test

INFO: Build completed successfully, 78 total actions
//rs/consensus:integration_test                                          PASSED in 48.5s
  Stats over 3 runs: max = 48.5s, min = 45.3s, avg = 46.4s, dev = 1.5s
//rs/consensus:slow_checkpointing_test                                   PASSED in 107.8s
  Stats over 3 runs: max = 107.8s, min = 106.7s, avg = 107.3s, dev = 0.5s
```

PR #11178, opened 2026-08-18
refactor(cketh): make the transaction pipeline generic over its request

gregorydemay:
## Why

The transaction pipeline served exactly one sender, the minter's main address: one nonce sequence, every map keyed by a ckETH ledger burn index, the request type baked in. A second sender address cannot reuse any of it, and giving the sweeper its own address is the point of the stack.

## What

Introduces a trait for what a request must answer in order to travel the pipeline, and makes the pipeline generic over it. Withdrawals are the only implementation here, wrapped so that every existing call site and behaviour is unchanged; the sweeper's implementation arrives in #11144.

The pipeline's vocabulary follows the type: it speaks of requests and ids rather than withdrawals, since another pipeline's requests are not withdrawals and its ids are not ledger burn indices. Reimbursement and withdrawal status keep their names — they belong to the wrapper and really are withdrawal concepts.

No behaviour change.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/tests.rs
@@ -923,19 +923,6 @@ proptest! {
     }
 }
 
-fn accepted_withdrawal_request_event(
-    request: crate::state::transactions::WithdrawalRequest,
-) -> EventType {
-    use crate::state::transactions::WithdrawalRequest;
-    match request {
-        WithdrawalRequest::CkEth(request) => EventType::AcceptedEthWithdrawalRequest(request),
-        WithdrawalRequest::CkErc20(request) => EventType::AcceptedErc20WithdrawalRequest(request),
-        WithdrawalRequest::SweeperFunding(request) => {
-            EventType::AcceptedSweeperFundingRequest(request)
-        }
-    }
-}
-
 #[test]
 fn state_equivalence() {
     use crate::EVM_RPC_ID_PRODUCTION;
@@ -1120,8 +1107,8 @@ fn state_equivalence() {
         }),
     };
     let builder = WithdrawalTransactionsBuilder::default()
-        .with_pending_withdrawal_requests(pending_requests)
-        .with_processed_withdrawal_requests(processed_requests)
+        .with_pending_requests(pending_requests)
+        .with_processed_requests(processed_requests)
         .with_created_tx(created_tx)
         .with_sent_tx(sent_tx)
         .with_finalized_tx(finalized_tx)
@@ -1346,7 +1333,7 @@ fn state_equivalence() {
         state.is_equivalent_to(&State {
             withdrawal_transactions: builder
                 .clone()
-                .with_pending_withdrawal_requests(
+                .with_pending_requests(
                     vec![
                         withdrawal_request2.clone().into(),
                         withdrawal_request1.clone().into()
@@ -1365,9 +1352,7 @@ fn state_equivalence() {
         state.is_equivalent_to(&State {
             withdrawal_transactions: builder
                 .clone()
-                .with_pending_withdrawal_requests(
-                    vec![withdrawal_request1.into()].into_iter().collect()
-                )
+                .with_pending_requests(vec![withdrawal_request1.into()].into_iter().collect())
                 .build(),
             ..state.clone()
         }),
@@ -2082,6 +2067,18 @@ mod eth_balance {
         let mut state = initial_state();
         add_erc20_token(&mut state);
         state
+    }
+
+    fn accepted_withdrawal_request_event(request: WithdrawalRequest) -> EventType {
+        match request {
+            WithdrawalRequest::CkEth(request) => EventType::AcceptedEthWithdrawalRequest(request),
+            WithdrawalRequest::CkErc20(request) => {
+                EventType::AcceptedErc20WithdrawalRequest(request)
+            }
+            WithdrawalRequest::SweeperFunding(request) => {
+                EventType::AcceptedSweeperFundingRequest(request)
+            }
+        }
     }
 
     fn add_erc20_token(state: &mut State) {
--- a/rs/ethereum/cketh/minter/src/state/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/tests.rs
@@ -923,6 +923,19 @@ proptest! {
     }
 }
 
+fn accepted_withdrawal_request_event(
+    request: crate::state::transactions::WithdrawalRequest,
+) -> EventType {
+    use crate::state::transactions::WithdrawalRequest;
+    match request {
+        WithdrawalRequest::CkEth(request) => EventType::AcceptedEthWithdrawalRequest(request),
+        WithdrawalRequest::CkErc20(request) => EventType::AcceptedErc20WithdrawalRequest(request),
+        WithdrawalRequest::SweeperFunding(request) => {
+            EventType::AcceptedSweeperFundingRequest(request)
+        }
+    }
+}
+
 #[test]
 fn state_equivalence() {
     use crate::EVM_RPC_ID_PRODUCTION;
@@ -978,13 +991,13 @@ fn state_equivalence() {
         ledger_burn_index: LedgerBurnIndex::new(20),
         ..withdrawal_request1.clone()
     };
-    let pending_withdrawal_requests: VecDeque<WithdrawalRequest> = vec![
+    let pending_requests: VecDeque<WithdrawalRequest> = vec![
         withdrawal_request1.clone().into(),
         withdrawal_request2.clone().into(),
     ]
     .into_iter()
     .collect();
-    let processed_withdrawal_requests = btreemap! {
+    let processed_requests = btreemap! {
         LedgerBurnIndex::new(4) => EthWithdrawalRequest {
             withdrawal_amount: Wei::new(1_000_000_000_000),
             ledger_burn_index: LedgerBurnIndex::new(4),
@@ -1107,8 +1120,8 @@ fn state_equivalence() {
         }),
     };
     let builder = WithdrawalTransactionsBuilder::default()
-        .with_pending_withdrawal_requests(pending_withdrawal_requests)
-        .with_processed_withdrawal_requests(processed_withdrawal_requests)
+        .with_pending_withdrawal_requests(pending_requests)
+        .with_processed_withdrawal_requests(processed_requests)
         .with_created_tx(created_tx)
         .with_sent_tx(sent_tx)
         .with_finalized_tx(finalized_tx)
@@ -1511,7 +1524,7 @@ mod sweeper_funding {
 
         let request = state
             .withdrawal_transactions
-            .withdrawal_requests_iter()
+            .requests_iter()
             .next()
             .expect("BUG: the funding request was not recorded");
         assert_matches!(request, WithdrawalRequest::SweeperFunding(_));
@@ -1532,7 +1545,7 @@ mod eth_balance {
     use crate::state::audit::{EventType, apply_state_transition};
     use crate::state::tests::checked_sub;
     use crate::state::tests::{initial_state, received_eth_event};
-    use crate::state::transactions::{EthWithdrawalRequest, WithdrawalRequest, create_transaction};
+    use crate::state::transactions::{EthWithdrawalRequest, PipelineRequest, WithdrawalRequest};
     use crate::state::{EthBalance, State};
     use crate::test_fixtures::sweeper_funding_request;
     use crate::tx::{SignedEip1559TransactionRequest, TransactionSignature};
@@ -2008,20 +2021,19 @@ mod eth_balance {
         }
 
         fn apply(self, state: &mut State) -> TransactionReceipt {
-            let accepted_withdrawal_request_event = self
-                .withdrawal_request
-                .clone()
-                .into_accepted_withdrawal_request_event();
+            let accepted_withdrawal_request_event =
+                accepted_withdrawal_request_event(self.withdrawal_request.clone());
             apply_state_transition(state, &accepted_withdrawal_request_event);
 
-            let transaction = create_transaction(
-                &self.withdrawal_request,
-                self.nonce,
-                self.tx_fee,
-                self.gas_limit,
-                EthereumNetwork::Sepolia,
-            )
-            .expect("BUG: failed to create transaction");
+            let transaction = self
+                .withdrawal_request
+                .to_transaction(
+                    self.nonce,
+                    self.tx_fee,
+                    self.gas_limit,
+                    EthereumNetwork::Sepolia,
+                )
+                .expect("BUG: failed to create transaction");
             apply_state_transition(
                 state,
                 &EventType::CreatedTransaction {
--- a/rs/ethereum/cketh/minter/src/state/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/tests.rs
@@ -2027,7 +2027,7 @@ mod eth_balance {
 
             let transaction = self
                 .withdrawal_request
-                .to_transaction(
+                .create_transaction(
                     self.nonce,
                     self.tx_fee,
                     self.gas_limit,
--- a/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
@@ -863,6 +863,7 @@ impl WithdrawalTransactions {
             reimbursed: Default::default(),
         }
     }
+
     /// Record a created transaction, and remember that the request may still need paying back.
     pub fn record_created_transaction(
         &mut self,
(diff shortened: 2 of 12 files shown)


rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1066,16 +1208,18 @@ impl TransactionLane {
         transactions
     }
 
-    /// Move the transaction matching `receipt` into the finalized map and clean up its
-    /// superseded resubmissions, returning the finalized transaction.
+    /// Finalize the sent transaction matching `receipt` for `ledger_burn_index`: move it from the
+    /// sent to the finalized set, drop any never-mined resubmissions for that nonce, and (on a
+    /// reimbursable lane) clear the maybe-reimburse marker. Reimbursement recording itself is
+    /// lane-specific and lives on [`TransactionLane<WithdrawalRequest>`].
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in db25106. The doc now says `id`, and the maybe-reimburse clause is gone — you are right that the generic `finalize_transaction` only moves sent→finalized and cleans up resubmissions. While checking, the pointer at the end was also wrong: reimbursement lives on `EthTransactions` (which owns the withdrawal lane), not on `TransactionLane<WithdrawalRequest>`, so the link now goes there.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1274,90 +1401,12 @@ impl TransactionLane {
     }
 }
 
-/// Creates an EIP-1559 transaction for the given withdrawal request.
+/// Creates an EIP-1559 transaction for the given lane request.
 /// The transaction fees are paid by the beneficiary,
 /// meaning that the fees will be deducted from the withdrawal amount.
 ///
 /// # Errors
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in db25106. The orphaned comment is deleted and what it documented — beneficiary-paid fees and the `InsufficientTransactionFee` error — now sits on `LaneRequest::to_transaction`, where the transaction is actually built. The selector const keeps its own `// First 4 bytes of keccak256(...)` line.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -811,20 +985,24 @@ pub enum CreateTransactionError {
...
         transaction_nonce: TransactionNonce,
         allowed_max_transaction_fee: Wei,
         max_transaction_fee: Wei,
     },
 }
 
-impl TransactionLane {
+/// One outcome of [`TransactionLane::create_resubmit_transactions`]: the fee-bumped transaction to
+/// re-sign (paired with its lane id), or why it could not be bumped.
+type ResubmitResult<Id> = Result<(Id, Eip1559TransactionRequest), ResubmitTransactionError<Id>>;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The compile claim is not right, and CI on this exact commit (db25106) disproves it: `Cargo Build Linux`, `Cargo Lint Linux` and `Bazel Test All` are all green, and locally `cargo check --all-targets --all-features -p ic-cketh-minter` plus clippy with `--deny warnings` are clean.

E0446 is about private *types* — structs and enums — appearing in a public signature. A type alias is transparent: it is only a name, and the type it expands to here (`Result`, `Eip1559TransactionRequest`, `ResubmitTransactionError`) is entirely public, so nothing private leaks. The alias merely is not nameable from outside this module, and rustdoc renders the expanded type.

There is a smaller, real point buried in this though: the sibling alias `ReimbursedResult` right above is `pub type`, so making `ResubmitResult` public too would be more consistent and would let the rendered docs show the alias name. Leaving that to @gregorydemay as a judgement call rather than restacking the whole stack for a visibility keyword.

rs/ethereum/cketh/minter/src/state/tests.rs
```diff
@@ -1950,6 +1947,18 @@ mod eth_balance {
...
+            WithdrawalRequest::CkEth(request) => EventType::AcceptedEthWithdrawalRequest(request),
+            WithdrawalRequest::CkErc20(request) => {
+                EventType::AcceptedErc20WithdrawalRequest(request)
+            }
+            WithdrawalRequest::SweeperFunding(request) => {
+                EventType::AcceptedSweeperFundingRequest(request)
+            }
+        }
+    }
+
```
gregorydemay, line 2083: There are 2 instances of this method, let's keep only one and move it to test_fixtures.rs

gregorydemay: 🤖 This one can't go in `test_fixtures.rs` as things stand, and it's worth writing down why.

The two copies sit either side of a **crate boundary**:

- `state/tests.rs` is part of the **lib** (`ic_cketh_minter`).
- `dashboard/tests.rs` is part of the **bin**: `dashboard` is declared in `main.rs`, not `lib.rs`, and `BUILD.bazel` explicitly excludes `src/dashboard.rs` and `src/dashboard/tests.rs` from the `rust_library`. That's why it reaches everything through `ic_cketh_minter::` rather than `crate::`.

`test_fixtures` is `#[cfg(test)]`-gated in the lib, so when the bin's test target is built the lib is compiled *without* `cfg(test)` and the module isn't there. Putting the helper in `test_fixtures.rs` and importing it from `dashboard/tests.rs` gives:

```
error[E0432]: unresolved import `ic_cketh_minter::test_fixtures`
   |                      ^^^^^^^^^^^^^ could not find `test_fixtures` in `ic_cketh_minter`
note: found an item that was configured out
```

The duplication is the price of removing `WithdrawalRequest::into_accepted_withdrawal_request_event` from productive code, which had only test callers. Options, roughly by cost:

1. **Move `dashboard` into the lib.** Then `dashboard/tests.rs` is a lib test module and reaches `test_fixtures` natively — i.e. exactly what's asked here. Costs ~27 import rewrites (11 in `dashboard.rs`, 14 in `dashboard/tests.rs`, 2 in `main.rs`) plus dropping the two `exclude` entries in `BUILD.bazel`. It also removes the oddity of a chunk of the minter importing its own lib as an external crate. Worth doing, but as its own PR.
2. **Move it to `ic-cketh-test-utils`.** One copy both targets can reach, and `lib_tests` already depends on it — but `unit_tests` currently depends only on `:minter`, and that crate brings 25 direct dependencies including `pocket-ic` and `ic-icrc1-ledger`. A lot of build graph for an 8-line function.
3. **Feature-gate `test_fixtures`** (`#[cfg(any(test, feature = "test-utils"))]`). Cheaper-looking than it is: `crate_features` is per-Bazel-target and the `features` list is shared, so without a separate test-only lib target this compiles the fixtures into the production wasm.

Leaving the two copies for now and keeping this PR on the pipeline change. Happy to do (1) as a follow-up if you want the single copy.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -377,31 +367,38 @@ impl fmt::Debug for Erc20WithdrawalRequest {
...
 ///    [`WithdrawalTransactions`].
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub(in crate::state) struct TransactionPipeline {
-    pending_withdrawal_requests: VecDeque<WithdrawalRequest>,
-    // Processed withdrawal requests (transaction created, sent, or finalized).
-    processed_withdrawal_requests: BTreeMap<LedgerBurnIndex, WithdrawalRequest>,
-    created_tx: MultiKeyMap<TransactionNonce, LedgerBurnIndex, TransactionRequest>,
-    sent_tx: MultiKeyMap<TransactionNonce, LedgerBurnIndex, Vec<SignedTransactionRequest>>,
-    finalized_tx: MultiKeyMap<TransactionNonce, LedgerBurnIndex, FinalizedEip1559Transaction>,
+pub struct TransactionPipeline<R: PipelineRequest> {
```
mbjorkqvist: nit: I think this could still stay `pub(in crate::state) struct TransactionPipeline`, as long as also `pub type MinterTransactionPipeline` is made `pub(in crate::state)`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -377,31 +367,38 @@ impl fmt::Debug for Erc20WithdrawalRequest {
...
+    // Processed requests (transaction created, sent, or finalized).
+    processed_requests: BTreeMap<R::Id, R>,
+    created_tx: MultiKeyMap<TransactionNonce, R::Id, TransactionRequest>,
+    sent_tx: MultiKeyMap<TransactionNonce, R::Id, Vec<SignedTransactionRequest>>,
+    finalized_tx: MultiKeyMap<TransactionNonce, R::Id, FinalizedEip1559Transaction>,
     next_nonce: TransactionNonce,
 }
 
+/// The pipeline sending from the minter's main address, on which user withdrawals travel.
+pub type MinterTransactionPipeline = TransactionPipeline<WithdrawalRequest>;
```
mbjorkqvist: nit: This could be `pub(in crate::state)`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -424,18 +421,22 @@ pub enum ResubmitTransactionError {
 /// How far a transaction has got through the pipeline. Carries the transaction itself, since
 /// every caller that asks the stage also wants the transaction at it.
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub(in crate::state) enum TransactionStage<'a> {
+pub enum TransactionStage<'a> {
```
mbjorkqvist: nit: I think this could also still stay `pub(in crate::state) struct TransactionStage`, as long as also `pub type MinterTransactionPipeline` is made `pub(in crate::state)` (same as for `TransactionPipeline`).

rs/ethereum/cketh/minter/src/state/transactions/request.rs
```diff
@@ -0,0 +1,188 @@
...
+    type Id = LedgerBurnIndex;
+    type Error = CreateTransactionError;
+
+    fn id(&self) -> LedgerBurnIndex {
+        self.cketh_ledger_burn_index()
+    }
+
+    fn created_at(&self) -> Option<u64> {
+        WithdrawalRequest::created_at(self)
+    }
```
mbjorkqvist: nit: This is not called anywhere. In `WithdrawalTransaction::oldest_incomplete_request_timestamp` we iterate over the requests and call `req.created_at()`, but this goes to the `impl WithdrawalTransaction` of `fn created_at`, rather than this one in `impl PipelineRequest for WithdrawalRequest`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -451,107 +452,75 @@ impl TransactionPipeline {
...
-        self.pending_withdrawal_requests.push_back(request);
+        self.pending_requests.push_back(request);
     }
 
-    /// Move an existing withdrawal request to the back of the queue.
-    pub fn reschedule_withdrawal_request<R: Into<WithdrawalRequest>>(&mut self, request: R) {
-        let request = request.into();
+    /// Move an existing request to the back of the queue.
+    pub fn reschedule_request<Req: Into<R>>(&mut self, request: Req) {
+        let id = request.into().id();
```
mbjorkqvist: This method was changed internally from removing a request by equality (`self.remove_withdrawal_request(&request)`) to removing by position. There seem to be a couple of consequences:

1. We still pass in `request: Req`, but only consume it to obtain the `id. The caller is already aware of the `id`, so it would be cheaper to pass that instead.
2. The caller could pass a modified `request` (e.g., a request with the same `id`, but a different `created_at` time), the the request that gets rescheduled is the existing one at the same `id`, rather than the request passed in by the caller.

The scenario in 2. is perhaps not that can occur in practice, but there is also no test for it. Changing the method signature (as per 1.) would remove the need for a test IIUC, so choosing between the two I'd go for that.

rs/ethereum/cketh/minter/src/state/transactions/request.rs
```diff
@@ -0,0 +1,188 @@
+//! What it takes for a request to travel a [`TransactionPipeline`], and the minter's own
```
mbjorkqvist: nit: When building, I get a "unresolved link to `TransactionPipeline`" warning here, and also for the occurrence on line 10.

mbjorkqvist (approved):
Thanks @gregorydemay, just a few, mainly minor comments.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 You are right that nothing forced the widening: no module outside `state::transactions` names `TransactionPipeline`, `MinterTransactionPipeline` or `TransactionStage`, so narrowing all three back to `pub(in crate::state)` compiles as-is and restores the visibility this had on `master`. Done — squashed into 948f633779 when this PR merged.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Done, together with `TransactionPipeline` and `TransactionStage` — squashed into 948f633779 when this PR merged.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Done — squashed into 948f633779 when this PR merged. The intent of "keep `TransactionStage` as visible as its pipeline" holds either way: the pipeline went back to `pub(in crate::state)`, so the stage followed it there.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/request.rs
gregorydemay: 🤖 Good catch, and the compiler agrees: while experimenting with a `pub(crate)` trait, dead-code analysis reported `method 'created_at' is never used`. As you say, the iterator in `oldest_incomplete_request_timestamp` yields a concrete `&WithdrawalRequest`, so the inherent method shadows the trait one at the only call site. Removed from the trait — squashed into 948f633779 when this PR merged; it can come back when a caller needs it generically.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Agreed on both counts, and I went for option 1: `reschedule_request` now takes the id (squashed into 948f633779 when this PR merged). The single production caller already binds `ledger_burn_index` from the error it is handling, so nothing had to be threaded through, and scenario 2 stops being representable rather than needing a test.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/request.rs
gregorydemay: 🤖 Fixed — squashed into 948f633779 when this PR merged. Worth noting the interaction with the visibility nits: giving the trait doc an explicit `(super::TransactionPipeline)` link target only trades this warning for `public documentation for 'PipelineRequest' links to private item` once the pipeline goes back to `pub(in crate::state)`, and `PipelineRequest` itself has to stay `pub` (`dashboard/tests.rs` uses it). So the module doc on line 1 gets the explicit path, while the trait doc on line 10 keeps plain backticks and no link. `cargo doc` now reports no warning from this crate other than the pre-existing `watch_deposit` -> `MAX_TOKENS_PER_ACCOUNT` one in `automatic_deposits`.

*(Edited: the commit I first linked was squashed away by the merge.)*

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 One follow-up, since it is the interesting half of this: #11144 has to widen `TransactionPipeline` (and `TransactionStage` with it) back to `pub`, because the sweeper's pipeline is constructed in `lifecycle::init` and held as a field of `State`, both outside `crate::state`. Routing `init.rs` through the public `SweeperTransactionPipeline` alias does not help — rustc rejects the associated-function call on a private type and additionally warns that the alias and the `State` field are more public than the type behind them.

So the widening now happens in the PR that introduces the second lane rather than in this one, which is where your nit points: dfinity/ic@b02e9d5422 on `greg/sweeper-send-lane`.

*(Edited: the commit I first linked was rewritten when that branch was rebased onto master.)*

PR #11180, opened 2026-08-18
chore: Remove feature flag for DMT

michael-weigelt:


rs/embedders/src/wasmtime_embedder.rs
```diff
@@ -846,7 +840,7 @@ fn sigsegv_memory_tracker<S>(
                     log.clone(),
                     dirty_page_tracking,
                     page_map,
-                    maybe_missing_page_handler_kind,
+                    Some(MissingPageHandlerKind::Deterministic),
```
michael-weigelt, line 843: Can be removed when we remove the other memory tracker implementation.

schneiderstefan:
I see this removes the flag, but can we go a step further and remove the old memory tracker?

michael-weigelt:
@schneiderstefan yes, that's being done as we speak as a follow-up. Plus some renamings and refactors.

schneiderstefan (approved):

PR #11182, opened 2026-08-18
chore: remove systemd from build container

nmattia:
This replaces the `sd_journal_print` calls in `infogetty` with `syslog` from `syslog.h` which is standard Unix. This prints to `/dev/log` which gets picked up by systemd and is independent of stdout/stderr.


gregorydemay (approved):

basvandijk (approved):
Nice. It seems to work: 
https://farm.dfinity.systems/group/basic-health-test--1787051111937029/vm/ouyhm-aba6e-fem56-6r2zq-4lxjj-qs4sh-crxjt-4rgqh-pi2eb-x64xg-cae/console/

blind-oracle (approved):

PR #11183, opened 2026-08-18
fix: make `ci/container/container-run.sh` work when the checkout is a git worktree

mbjorkqvist:
Running `./ci/container/container-run.sh` from a linked git worktree gave a container in which git did not work at all, and the ci scripts that locate the repository with git did not notice.

**The container could not use git.** In a linked worktree `.git` is a file pointing at an admin directory under the main checkout's `.git` — an absolute path outside the checkout. Only the checkout was mounted, so that pointer dangled and every `git` call inside the container failed. `container-run.sh` now also mounts that directory at the same absolute path, which fixes every git-dependent script at once rather than one script at a time. Plain clones are unaffected. A worktree whose pointer is *relative* (git's `worktree.useRelativePaths`) cannot be made to resolve this way, so that case is reported and refused instead of silently handed over.

That mount makes all worktrees look prunable inside the container, since their back-pointers name host paths that are not mounted. Reading is unaffected, but `git gc` prunes worktrees it considers stale as part of its job, and would delete the host's admin directories — including those of worktrees you are not even using here. Worktree pruning is therefore switched off for the container, rather than gc as a whole: the repository still gets its object maintenance, and an explicit `git gc` is safe too.

**The failures were silent.** `cd ""` returns 0 in bash and does not move the shell, and a command substitution inside `${X:-...}` does not trip errexit, so a script whose `git rev-parse --show-toplevel` failed carried on against whatever directory it happened to be in and still exited 0; `export X="$(...)"` swallows the substitution's exit status the same way. The affected scripts now assign first and then `cd`/`export`, so `set -e` fires. A sweep found no other instances.

Only local developer tooling changes — CI sets `CI_PROJECT_DIR` and never hit this. Note that the VS Code dev container (`.devcontainer/devcontainer.json`) is still affected: its mounts are static, so it cannot pick up the main checkout's git directory the way this script does.

### DIFF
--- a/ci/container/build-ic.sh
+++ b/ci/container/build-ic.sh
@@ -10,7 +10,11 @@ if [ "${BUILD_IC_NESTED:-}" == 1 ]; then
 fi
 export BUILD_IC_NESTED=1
 
-export ROOT_DIR="$(git rev-parse --show-toplevel)"
+# Assign, then export: `export X="$(...)"` swallows the command
+# substitution's exit status, so a failing git would leave X empty and
+# `set -e` would not notice.
+ROOT_DIR="$(git rev-parse --show-toplevel)"
+export ROOT_DIR
 
 # This script needs to be run inside the ic-build container
 # If it isn't, we drop into the correct container.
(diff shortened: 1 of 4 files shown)


basvandijk:
I just noticed this PR. Note that the recently merged https://github.com/dfinity/ic/pull/11441 fixes the same problem. I'm also working on a simplified alternative in https://github.com/dfinity/ic/pull/11460.

mbjorkqvist:
> I just noticed this PR. Note that the recently merged #11441 fixes the same problem. I'm also working on a simplified alternative in #11460.

Thanks, I'll close this PR then!

PR #11184, opened 2026-08-18
chore: Bump current certification version to V29

mraszyk:
All mainnet subnets run replica versions supporting up to `V29`, so the current certification version can be bumped to `V29`.

This makes the fixed `consumed_cycles_total` (which no longer double counts the cycles consumed by deleted canisters and additionally includes the cycles consumed by all non-deleted canisters) visible in `/subnet/<subnet_id>/metrics`.

`SystemMetadata` encodes the current certification version, hence the expected hashes in `rs/state_manager/src/manifest/split/tests.rs` had to be updated.

## Tests

The cycles consumed by the canisters that still exist are not persisted: they are derived by `ReplicatedState::new_from_checkpoint` on load and refreshed by `StateManagerImpl::commit_and_certify` on every committed state. Now that `/subnet/<subnet_id>/metrics` certifies them, three `StateMachine` tests pin that, as promised in #11297.

- `subnet_metrics_are_unchanged_across_a_restart` writes a checkpoint and restarts the node, asserting the same `SubnetMetrics` and the same certified `/subnet/<subnet_id>/metrics` leaf, read via `read_certified_state`.
- `consumed_cycles_in_subnet_metrics_are_refreshed_every_round` installs a heartbeat burning cycles every round, and after each tick asserts the aggregate against one recomputed from the state, and that it grew by at least the burned amount.
- `consumed_cycles_in_subnet_metrics_are_refreshed_when_a_canister_is_deleted` deletes the only canister on the subnet, so that nothing else consumes cycles meanwhile and the aggregate can be pinned to exactly the balance burned with it -- the canister's consumption having merely moved from the canisters' part into `consumed_cycles_by_deleted_canisters`.


rs/canonical_state/certification_version/src/lib.rs
```diff
@@ -61,7 +61,7 @@ impl std::convert::TryFrom<u32> for CertificationVersion {
 
 /// The Canonical State certification version that should be used for newly
 /// computed states.
-pub const CURRENT_CERTIFICATION_VERSION: CertificationVersion = CertificationVersion::V28;
+pub const CURRENT_CERTIFICATION_VERSION: CertificationVersion = CertificationVersion::V29;
```
mraszyk, line 64: A new test should be added that asserts the value of SubnetMetrics (in particular, the consumed cycles which are not persisted) does not change across replica restarts.

mraszyk: Another new test should be added that asserts the value of consumed cycles in SubnetMetrics is properly refreshed every round.

mraszyk: Added `subnet_metrics_are_unchanged_across_a_restart` in `rs/execution_environment/tests/execution_test.rs`.

It builds up both parts of the aggregate -- a persisted one (a canister that is stopped and deleted, so `consumed_cycles_by_deleted_canisters` and the per-use-case map are non-zero) and a transient one (a live universal canister that burned cycles) -- writes a checkpoint and then `restart_node()`s. It asserts that the restarted replica resumes at the same height, holds the same `SubnetMetrics` (struct equality, so the non-persisted `consumed_cycles_total_including_canisters` is included), and certifies the same `/subnet/<subnet_id>/metrics` leaf under the same canonical state tree root hash, read via `read_certified_state`.

Two assertions guard against vacuity: the subnet-level total is non-zero, and the full aggregate is strictly above it, i.e. the transient part really is a positive share of what is being compared.

The test fails if the derivation in `ReplicatedState::new_from_checkpoint` is dropped -- and, with the `SubnetMetrics` equality assertion removed, the certified-leaf assertion alone still catches it.

mraszyk: Added `consumed_cycles_in_subnet_metrics_are_refreshed_every_round`, alongside the one above.

It installs a heartbeat that burns a fixed amount of cycles, so that every single round consumes cycles, and only on that one canister. After each of five ticks it checks the stored `consumed_cycles_total_including_canisters` against one recomputed from the state -- the cycles consumed by deleted canisters, which no test here has any other subnet-level consumption to add to, plus the sum of `CanisterMetrics::consumed_cycles()` over `canisters_iter()` -- and asserts it grew by at least the burned amount; a round whose commit skipped the refresh would fall behind by that much.

A third test, `consumed_cycles_in_subnet_metrics_are_refreshed_when_a_canister_is_deleted`, covers the subnet-level part moving in lockstep. It deletes the only canister on the subnet, so nothing else consumes cycles while the deletion executes and the aggregate can be pinned to an equality: it grows by exactly the balance burned with the canister, whose consumption merely moved into `consumed_cycles_by_deleted_canisters`. A bound rather than an equality would not have caught counting the canister on both sides.

Besides failing when the `refresh_consumed_cycles()` call in `commit_and_certify` is removed outright, the per-round test also fails when the refresh is merely made to skip rounds (guarding it by `height.get().is_multiple_of(2)`), so the "every round" part is genuinely pinned; the deletion test fails when `consumed_cycles_total()` is made to double count deleted canisters.

schneiderstefan (approved):

PR #11186, opened 2026-08-18
feat: declare mainnet binaries as bazel dependencies

basvandijk:
Note this PR is at the bottom of a stack of https://github.com/dfinity/ic/pull/11187 and https://github.com/dfinity/ic/pull/11188.

Several system-tests need to run a binary built at a version currently deployed on mainnet, and download it from `download.dfinity.systems` **while the test is running**. That pins them to the Farm backend — the `_local` backend runs in a network-isolated sandbox — and it means the binary is used without verifying its sha256 (`file_sync_helper::download_binary` passes `None` as the expected hash).

This PR adds the mechanism; the tests are converted in the PRs stacked on top.

### Where the hashes live

In `mainnet-icos-revisions.json`, as a `binaries` map next to the `version` they belong to, rather than in a new file. The binaries have to match the *exact* GuestOS revision a test boots — `ic-backup` looks up `binaries/<version the node reports>/<name>` — so a separate file updated by a separate bot PR could drift for a cron cycle and the backup tests would silently fall back to a download. One record keeps the update atomic. It also needs no CI-policy changes: `mainnet-*-revisions.json` is already covered by `BOT_APPROVED_FILES`, `CODEOWNERS` and `auto-approve.yml`.

### The pieces

* `mainnet_revisions.py` reads the hashes from the CDN's `binaries/x86_64-linux/SHA256SUMS` — the very directory the repository rule downloads from, so the recorded hash and the verified download can't refer to different copies. It raises when a binary is missing, so that fails the updater's own PR rather than `bazel build` for everyone.
* The "already up to date" guard now checks every field we record (`REQUIRED_RECORD_FIELDS`) instead of just one, so existing entries are backfilled on the next run instead of only picking the new field up on the next version bump. This is what generated the +36 lines in the JSON here.
* `bazel/mainnet-icos-binaries.bzl` downloads each recorded binary pinned by that sha256 and decompresses it into an executable file, exported as `@mainnet_{nns,app}_binaries//:<name>`.
* `rs/tests/common.bzl` turns those into per-purpose `runtime_deps` dicts.

### Notes for review

* Both repos download every binary in the map (~99 MB total), even though `@mainnet_app_binaries` only needs three of them. That keeps the JSON the single source of truth instead of duplicating the list in `MODULE.bazel`; the unused ones are only *downloaded*, never *built*, since their genrules never enter a test's `data`. If fetch size ever matters, the escape hatch is a `binaries` filter attribute — two lines.
* The genrule is `gunzip_{name}` with output `{name}` because a target may not share its name with its own output (same reason `//publish/binaries` uses `{name}_cleaned`). The output basename matters: `run_systest.sh` symlinks runtime deps as `<hash>-<basename>` and `ic-backup` looks its binaries up by name.
* Decompression is a genrule rather than `repository_ctx.execute(["gunzip", ...])` so that `repo_metadata(reproducible = True)` stays honest — repo content depends only on the watched JSON and the rule attributes.

### Verification

```
bazel build @mainnet_nns_binaries//... @mainnet_app_binaries//...   # 14 targets, all mode 0555
ic-replay --help / types-test --list                                 # run
python3 ci/src/mainnet_revisions/mainnet_revisions.py --dry-run icos  # 2nd run: all 4 records skip, empty diff
```

### DIFF
--- a/ci/src/mainnet_revisions/mainnet_revisions.py
+++ b/ci/src/mainnet_revisions/mainnet_revisions.py
@@ -21,13 +21,7 @@ SAVED_VERSIONS_CANISTERS_FILE = "mainnet-canister-revisions.json"
 # Release binaries (published on the CDN under `.../binaries/x86_64-linux/` as
 # `<name>.gz`) whose sha256 we record for every mainnet revision. They are exposed
 # to Bazel by //bazel:mainnet-icos-binaries.bzl and consumed by system-tests that
-# have to run a binary built at a mainnet version:
-#   * ic-replay, sandbox_launcher, canister_sandbox, compiler_sandbox
-#       -> //rs/tests/consensus/backup:backup_manager_{upgrade,downgrade}_test
-#   * types-test
-#       -> //rs/tests/consensus/orchestrator:cup_compatibility_test
-#   * replicated-state-test, state-layout-test
-#       -> //rs/tests/message_routing:queues_compatibility_test
+# have to run a binary built at a mainnet version.
 # Keep sorted; this determines the key order in the generated JSON.
 MAINNET_BINARIES = [
     "canister_sandbox",
@@ -38,12 +32,6 @@ MAINNET_BINARIES = [
     "state-layout-test",
     "types-test",
 ]
-
-# Fields every saved record must have. Listing them (rather than checking a single
-# field) means that when a new field is introduced, records that are already on the
-# current version are still rewritten once, so the new field gets backfilled instead
-# of only appearing on the next version bump.
-REQUIRED_RECORD_FIELDS = ("setupos_disk_img_hash", "binaries")
 
 
 class Command(Enum):
@@ -478,7 +466,7 @@ def get_binary_hashes(version: str) -> dict:
 
 def is_record_up_to_date(existing: dict, version: str) -> bool:
     """Whether `existing` is on `version` and already has every field we record."""
-    return existing.get("version", "") == version and all(f in existing for f in REQUIRED_RECORD_FIELDS)
+    return existing.get("version", "") == version and all(f in existing for f in ("setupos_disk_img_hash", "binaries"))
 
 
 def get_logger(level) -> logging.Logger:
--- a/ci/src/mainnet_revisions/mainnet_revisions.py
+++ b/ci/src/mainnet_revisions/mainnet_revisions.py
@@ -18,6 +18,33 @@ app_subnet_id = "io67a-2jmkw-zup3h-snbwi-g6a5n-rm5dn-b6png-lvdpl-nqnto-yih6l-gqe
 PUBLIC_DASHBOARD_API = "https://ic-api.internetcomputer.org"
 SAVED_VERSIONS_CANISTERS_FILE = "mainnet-canister-revisions.json"
 
+# Release binaries (published on the CDN under `.../binaries/x86_64-linux/` as
+# `<name>.gz`) whose sha256 we record for every mainnet revision. They are exposed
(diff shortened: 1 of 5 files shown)


ci/src/mainnet_revisions/mainnet_revisions.py
```diff
@@ -18,6 +18,33 @@
...
+# Release binaries (published on the CDN under `.../binaries/x86_64-linux/` as
+# `<name>.gz`) whose sha256 we record for every mainnet revision. They are exposed
+# to Bazel by //bazel:mainnet-icos-binaries.bzl and consumed by system-tests that
+# have to run a binary built at a mainnet version:
+#   * ic-replay, sandbox_launcher, canister_sandbox, compiler_sandbox
+#       -> //rs/tests/consensus/backup:backup_manager_{upgrade,downgrade}_test
+#   * types-test
+#       -> //rs/tests/consensus/orchestrator:cup_compatibility_test
+#   * replicated-state-test, state-layout-test
+#       -> //rs/tests/message_routing:queues_compatibility_test
```
nmattia: this is helpful but I'm worried it'll get out of date

ci/src/mainnet_revisions/mainnet_revisions.py
```diff
@@ -18,6 +18,33 @@
...
+    "sandbox_launcher",
+    "state-layout-test",
+    "types-test",
+]
+
+# Fields every saved record must have. Listing them (rather than checking a single
+# field) means that when a new field is introduced, records that are already on the
+# current version are still rewritten once, so the new field gets backfilled instead
+# of only appearing on the next version bump.
+REQUIRED_RECORD_FIELDS = ("setupos_disk_img_hash", "binaries")
```
nmattia: can you define this is `is_record_up_to_date`? here it's defined pretty far from where it's actually used which is confusing. I would even inline it where it's used

rs/tests/common.bzl
```diff
@@ -15,6 +15,66 @@ MAINNET_ENV = {
     "MAINNET_APP_GUESTOS_REVISION_ENV": MAINNET_APP["version"],
 }
 
+def mainnet_binaries_runtime_deps(repo, prefix, binaries):
+    """Runtime dependencies on binaries published for a mainnet revision.
+
+    These come from @mainnet_{nns,app}_binaries (see
+    //bazel:mainnet-icos-binaries.bzl) instead of being downloaded from the CDN by
+    the test itself, which is what lets the `_local` (network-isolated) variants of
+    the consuming tests run.
```
nmattia: nit: will get out of date. I think it's fine to not mention where it's used or how it's used (next paragraph). Same for "Used by ..." below.

nmattia (approved):

ci/src/mainnet_revisions/mainnet_revisions.py
basvandijk: Removed in https://github.com/dfinity/ic/pull/11186/changes/d07b48ebf93155956ebd584833e152f42ca220cd.

rs/tests/common.bzl
basvandijk: Removed in https://github.com/dfinity/ic/pull/11186/changes/d07b48ebf93155956ebd584833e152f42ca220cd.

ci/src/mainnet_revisions/mainnet_revisions.py
basvandijk: Inlined in https://github.com/dfinity/ic/pull/11186/changes/d07b48ebf93155956ebd584833e152f42ca220cd.

PR #11190, opened 2026-08-18
refactor(cketh): extract TransactionPipeline from WithdrawalTransactions

gregorydemay:
## Why

`WithdrawalTransactions` does two jobs. It drives transactions through create → sign → send → resubmit → finalize on the minter's nonce sequence, and it tracks which withdrawals still owe the user a refund. The two meet at only two moments — creating a transaction arms the refund, finalizing it disarms it — and everything else about sending a transaction is indifferent to whether anyone gets paid back.

The minter's sweeper address will need that send machinery on a nonce sequence of its own. It burns no ckETH, so it must be able to reuse the machinery without inheriting a refund path it can never use.

## What

`TransactionPipeline` becomes the send machinery alone — the request queues, the three transaction maps and the nonce. `WithdrawalTransactions` owns a pipeline plus the reimbursement bookkeeping, forwards the pipeline's API so callers are unchanged, and intercepts those two moments so the invariant lives in one place. Reimbursement records and withdrawal status move with it.

Two things follow from the split:

- `record_finalized_transaction` exists on both types: the pipeline's performs the finalize mechanics and hands back the finalized transaction; the wrapper's uses that to decide the refund. It keeps the same name on both, as `record_created_transaction` already did.
- `transaction_stage` reports where a transaction sits — created, sent or finalized — returning `Option`, so absence is the type's business rather than a variant. The wrapper maps that to `RetrieveEthStatus`, consulting reimbursement only once finalized. `RetrieveEthStatus` cannot move to the pipeline: its finalized payload is `Success | PendingReimbursement | Reimbursed`, which is reimbursement vocabulary throughout.

`state::tests` is untouched: it builds its fixture through a builder, so the new nesting is absorbed in `build()` and no assertion has to be restated.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
@@ -379,7 +379,7 @@ impl fmt::Debug for Erc20WithdrawalRequest {
 
 /// State machine holding Ethereum transactions issued by the minter.
 /// Overall the transaction lifecycle is as follows:
-/// 1. The user's withdrawal request is enqueued and processed in a FIFO order.
+/// 1. A withdrawal request is enqueued and processed in a FIFO order.
 /// 2. A transaction is created by either consuming a withdrawal request
 ///    (the first time a transaction is created for that nonce and burn index)
 ///    or re-submitting an already sent transaction for that nonce and burn index.
@@ -387,10 +387,9 @@ impl fmt::Debug for Erc20WithdrawalRequest {
 ///    previously created transaction or re-submitting an already sent transaction as is.
 /// 4. The transaction is sent to Ethereum. There may have been multiple
 ///    sent transactions for that nonce and burn index in case of resubmissions.
-/// 5. For a given nonce (and burn index), at most one sent transaction is finalized.
-///    The others sent transactions for that nonce were never mined and can be discarded.
-/// 6. For a given nonce, at most one sent transaction is finalized; a failed one is reported as
-///    such. Paying the requester back is not the pipeline's concern — see
+/// 5. For a given nonce (and burn index), at most one sent transaction is finalized; a failed
+///    one is reported as such. The other sent transactions for that nonce were never mined and
+///    can be discarded. Paying the requester back is not the pipeline's concern — see
 ///    [`WithdrawalTransactions`].
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub(in crate::state) struct TransactionPipeline {
@@ -425,7 +424,7 @@ pub enum ResubmitTransactionError {
 /// How far a transaction has got through the pipeline. Carries the transaction itself, since
 /// every caller that asks the stage also wants the transaction at it.
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub enum TransactionStage<'a> {
+pub(in crate::state) enum TransactionStage<'a> {
     Created(&'a Eip1559TransactionRequest),
     /// The most recently sent transaction, i.e. the one with the highest fee.
     Sent(&'a SignedTransactionRequest),
@@ -1012,11 +1011,14 @@ impl WithdrawalTransactions {
                     INFO,
                     "[record_finalized_transaction]: UNEXPECTED: sweeper funding {} of {} to {} \
                      FAILED (tx {}), which should be impossible for a transfer to an address the \
-                     minter controls; the burn is NOT reimbursed",
+                     minter controls; the burn is NOT reimbursed: no ETH reached the sweeper, the \
+                     failed transaction still paid {} of gas, and the rest of the burn now \
+                     over-backs ckETH",
                     ledger_burn_index,
                     request.withdrawal_amount,
                     request.destination,
                     receipt.transaction_hash,
+                    receipt.effective_transaction_fee(),
                 );
                 return;
             }
@@ -1252,16 +1254,9 @@ impl WithdrawalTransactions {
     }
 
     pub fn maybe_reimburse_requests_iter(&self) -> impl Iterator<Item = &WithdrawalRequest> {
-        self.pipeline
-            .processed_withdrawal_requests
+        self.maybe_reimburse
             .iter()
-            .filter_map(|(index, request)| {
-                if self.maybe_reimburse.contains(index) {
-                    Some(request)
-                } else {
-                    None
-                }
-            })
+            .filter_map(|index| self.pipeline.get_processed_withdrawal_request(index))
     }
 
     pub fn oldest_incomplete_withdrawal_timestamp(&self) -> Option<u64> {
@@ -1350,6 +1345,13 @@ impl WithdrawalTransactions {
             );
         }
         if tx.transaction_status() == &TransactionStatus::Failure {
+            // Unreachable for a funding: the destination is derived from the minter's own
+            // key, so a bare transfer there has no code to revert in. Were it reached, the
+            // status would be wrong, since nothing reimburses a funding — tolerable only
+            // because it cannot happen, and not worth a status of its own, which would mean
+            // adding a variant to `retrieve_eth_status`' return type and breaking existing
+            // clients. Revisit if funding ever goes through a contract, where a revert becomes
+            // possible.
             return (
                 RetrieveEthStatus::TxFinalized(TxFinalizedStatus::PendingReimbursement(
                     EthTransaction {
--- a/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
@@ -392,7 +392,7 @@ impl fmt::Debug for Erc20WithdrawalRequest {
 /// 6. If a given transaction fails the minter will reimburse the user who requested the
 ///    withdrawal with the corresponding amount minus fees.
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub struct WithdrawalTransactions {
+pub(in crate::state) struct TransactionPipeline {
     pending_withdrawal_requests: VecDeque<WithdrawalRequest>,
     // Processed withdrawal requests (transaction created, sent, or finalized).
     processed_withdrawal_requests: BTreeMap<LedgerBurnIndex, WithdrawalRequest>,
@@ -400,10 +400,6 @@ pub struct WithdrawalTransactions {
     sent_tx: MultiKeyMap<TransactionNonce, LedgerBurnIndex, Vec<SignedTransactionRequest>>,
     finalized_tx: MultiKeyMap<TransactionNonce, LedgerBurnIndex, FinalizedEip1559Transaction>,
     next_nonce: TransactionNonce,
-
-    maybe_reimburse: BTreeSet<LedgerBurnIndex>,
-    reimbursement_requests: BTreeMap<ReimbursementIndex, ReimbursementRequest>,
-    reimbursed: BTreeMap<ReimbursementIndex, ReimbursedResult>,
 }
 
 #[derive(Clone, Eq, PartialEq, Debug)]
@@ -425,7 +421,7 @@ pub enum ResubmitTransactionError {
     },
 }
 
-impl WithdrawalTransactions {
+impl TransactionPipeline {
     pub fn new(next_nonce: TransactionNonce) -> Self {
         Self {
             pending_withdrawal_requests: VecDeque::new(),
@@ -434,9 +430,6 @@ impl WithdrawalTransactions {
             sent_tx: MultiKeyMap::default(),
             finalized_tx: MultiKeyMap::default(),
             next_nonce,
-            maybe_reimburse: Default::default(),
-            reimbursement_requests: Default::default(),
-            reimbursed: Default::default(),
         }
     }
 
@@ -446,38 +439,6 @@ impl WithdrawalTransactions {
 
     pub fn update_next_transaction_nonce(&mut self, new_nonce: TransactionNonce) {
         self.next_nonce = new_nonce;
-    }
-
-    pub fn reimbursement_requests_iter(
-        &self,
-    ) -> impl Iterator<Item = (&ReimbursementIndex, &ReimbursementRequest)> {
-        self.reimbursement_requests.iter()
-    }
-
-    pub fn reimbursed_transactions_iter(
-        &self,
-    ) -> impl Iterator<Item = (&ReimbursementIndex, &ReimbursedResult)> {
-        self.reimbursed.iter()
-    }
-
-    fn find_reimbursed_transaction_by_cketh_ledger_burn_index(
-        &self,
-        searched_burn_index: &LedgerBurnIndex,
-    ) -> Option<&ReimbursedResult> {
-        self.reimbursed
-            .iter()
-            .find_map(|(index, value)| match index {
-                ReimbursementIndex::CkEth { ledger_burn_index }
-                    if ledger_burn_index == searched_burn_index =>
-                {
-                    Some(value)
-                }
-                ReimbursementIndex::CkErc20 {
-                    cketh_ledger_burn_index,
-                    ..
-                } if cketh_ledger_burn_index == searched_burn_index => Some(value),
-                _ => None,
-            })
     }
 
     pub fn record_withdrawal_request<R: Into<WithdrawalRequest>>(&mut self, request: R) {
@@ -576,15 +537,11 @@ impl WithdrawalTransactions {
             ),
             Ok(())
         );
-        let is_reimbursable = withdrawal_request.is_reimbursable();
         assert_eq!(
             self.processed_withdrawal_requests
                 .insert(withdrawal_id, withdrawal_request),
             None
         );
-        if is_reimbursable {
-            assert!(self.maybe_reimburse.insert(withdrawal_id));
-        }
     }
 
     pub fn record_signed_transaction(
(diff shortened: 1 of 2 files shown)


rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -358,18 +358,14 @@ impl fmt::Debug for Erc20WithdrawalRequest {
 /// 6. If a given transaction fails the minter will reimburse the user who requested the
 ///    withdrawal with the corresponding amount minus fees.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch — fixed in 8572a71. Step 6 of the lifecycle still promised to reimburse the user, which is exactly the responsibility this PR moves into `WithdrawalTransactions`. It now describes finalization only and points at the wrapper for the refund.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1082,18 +837,443 @@ impl WithdrawalTransactions {
...
+    ) -> Vec<(
+        &WithdrawalRequest,
+        WithdrawalStatus,
+        Option<&Eip1559TransactionRequest>,
+    )> {
+        // Pending requests matching the given search parameter
+        let pending = self
+            .pipeline
+            .pending_withdrawal_requests
+            .iter()
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed, fixed in 8572a71 — and applied to the four sibling sites the comment did not flag: `processed_withdrawal_requests`, `created_tx`, `sent_tx` and `finalized_tx` were being read directly too. `withdrawal_requests_iter()` and `get_finalized_transaction()` already existed; the other three lookups now have accessors (`processed_withdrawal_requests_iter`, `get_created_transaction`, `get_last_sent_transaction`). No field access from the wrapper remains.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -414,38 +408,6 @@ impl WithdrawalTransactions {
...
-                    Some(value)
-                }
-                ReimbursementIndex::CkErc20 {
-                    cketh_ledger_burn_index,
-                    ..
-                } if cketh_ledger_burn_index == searched_burn_index => Some(value),
-                _ => None,
-            })
-    }
-
```
gregorydemay, line 482: FYI: move to impl `WithdrawalTransactions` block

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -545,7 +507,6 @@ impl WithdrawalTransactions {
                 .insert(withdrawal_id, withdrawal_request),
             None
         );
-        assert!(self.maybe_reimburse.insert(withdrawal_id));
```
gregorydemay: FYI: move to `WithdrawalTransactions::record_created_transaction`

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -701,202 +664,7 @@ impl WithdrawalTransactions {
...
-                            reimbursed_amount: request.withdrawal_amount.change_units(),
-                            to: request.from,
-                            to_subaccount: request.from_subaccount.clone(),
-                            transaction_hash: Some(receipt.transaction_hash),
-                        },
-                    );
-                }
-            }
-        }
-    }
```
gregorydemay, line 818: FYI: move to `WithdrawalTransactions::record_finalized_transaction`

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -701,202 +664,7 @@ impl WithdrawalTransactions {
...
-                Ok(Reimbursed {
-                    burn_in_block,
-                    reimbursed_in_block,
-                    reimbursed_amount: reimbursement_request.reimbursed_amount,
-                    transaction_hash: reimbursement_request.transaction_hash,
-                }),
-            ),
-            None
-        );
-    }
```
gregorydemay, line 873: FYI: move to `WithdrawalTransactions`

gregorydemay:
🤖 **Reading guide** — the diff in `state/transactions/mod.rs` is one large block because a single 38-method `impl WithdrawalTransactions` becomes two impls (29 + 36 methods). A line-based diff cannot express "this method is now two methods on two types", so most of what looks like churn is code that changed `impl` block without changing.

I tried reordering the file to make it align method-by-method — both putting the wrapper first and reordering its methods into master's order. Neither helps (11 hunks → 10, the two large hunks remain), so the file keeps the order that reads best rather than one contorted for the diff.

Instead, here is what the 41 methods actually are:

| role | count | what to do |
|---|---|---|
| pure delegate | 20 | skim — the body is `self.pipeline.<same name>(args)` |
| moved verbatim into `TransactionPipeline` | 23 | skip — byte-identical to master |
| wrapper only (reimbursement + status) | 12 | skip — unchanged from master |
| pipeline only | 5 | 3 verbatim helpers, 2 new: `transaction_stage`, `processed_withdrawal_requests_iter` |
| **genuinely different** | **4** | **read carefully** |

**The four worth your attention**, which are exactly what the split is about:

- `new` — constructs a pipeline plus the reimbursement bookkeeping around it
- `is_equivalent_to` — compares the reimbursement state, then delegates the rest
- `record_created_transaction` — delegates, then *arms* the refund
- `record_finalized_transaction` — delegates the finalize mechanics, *disarms* the refund, then records the reimbursement

Those last two are the only places where the pipeline and the refund bookkeeping must stay in step, which is why they intercept rather than delegate. Both are now guarded by `is_reimbursable`, since a `SweeperFunding` request is never paid back — that variant arrived on master while this branch was in review.

To collapse the relocated code while reading:

```
git diff --color-moved=zebra --color-moved-ws=allow-indentation-change origin/master...greg/cketh-extract-pipeline
```

Pink and blue are removals that reappear elsewhere, cyan and yellow their arrivals; skipping all four leaves only what genuinely changed.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -389,21 +389,18 @@ impl fmt::Debug for Erc20WithdrawalRequest {
 ///    sent transactions for that nonce and burn index in case of resubmissions.
 /// 5. For a given nonce (and burn index), at most one sent transaction is finalized.
 ///    The others sent transactions for that nonce were never mined and can be discarded.
-/// 6. If a given transaction fails the minter will reimburse the user who requested the
-///    withdrawal with the corresponding amount minus fees.
+/// 6. For a given nonce, at most one sent transaction is finalized; a failed one is reported as
+///    such. Paying the requester back is not the pipeline's concern — see
+///    [`WithdrawalTransactions`].
```
mbjorkqvist: In bullet points 5 & 6, the "For a given nonce, at most one sent transaction is finalized" part is repeated. IIUC, the transaction lifecycle is now only 5 steps, so the two last bullets should be consolidated?

Also, since the transaction lifecycle will cover both user withdrawals and sweeper funding, maybe the "The user's withdrawal" wording in bullet point 1 should be amended?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1163,18 +902,471 @@ impl WithdrawalTransactions {
...
+                log!(
+                    INFO,
+                    "[record_finalized_transaction]: UNEXPECTED: sweeper funding {} of {} to {} \
+                     FAILED (tx {}), which should be impossible for a transfer to an address the \
+                     minter controls; the burn is NOT reimbursed",
+                    ledger_burn_index,
+                    request.withdrawal_amount,
+                    request.destination,
+                    receipt.transaction_hash,
+                );
```
mbjorkqvist, line 1022: nit: Would it make sense to still keep the `receipt.effective_transaction_fee()` in the log message (even though this arm should be unreachable)? It may have been dropped accidentally as part of the merge conflict. (The original log message merged in #11072 below.)

```
                    log!(
                        INFO,
                        "[record_finalized_transaction]: UNEXPECTED: sweeper funding {} of {} to \
                         {} FAILED (tx {}), which should be impossible for a transfer to an \
                         address the minter controls; the burn is NOT reimbursed: no ETH reached \
                         the sweeper, the failed transaction still paid {} of gas, and the rest \
                         of the burn now over-backs ckETH",
                        ledger_burn_index,
                        request.withdrawal_amount,
                        request.destination,
                        receipt.transaction_hash,
                        receipt.effective_transaction_fee(),
                    );
```

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1163,18 +902,471 @@ impl WithdrawalTransactions {
...
+        self.pipeline
+            .processed_withdrawal_requests
+            .iter()
+            .filter_map(|(index, request)| {
+                if self.maybe_reimburse.contains(index) {
+                    Some(request)
+                } else {
+                    None
+                }
+            })
```
mbjorkqvist: IIUC, this is still reading the `self.pipeline.processed_withdrawal_requests` private field, rather than going through the pipeline's API. Would it make sense to change it as follows?
```suggestion
        self.maybe_reimburse
          .iter()
          .filter_map(|index| self.pipeline.get_processed_withdrawal_request(index))
```

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -1163,18 +902,471 @@ impl WithdrawalTransactions {
...
+        }
+        if tx.transaction_status() == &TransactionStatus::Failure {
+            return (
+                RetrieveEthStatus::TxFinalized(TxFinalizedStatus::PendingReimbursement(
+                    EthTransaction {
+                        transaction_hash: tx.transaction_hash().to_string(),
+                    },
+                )),
+                Some(tx.as_ref()),
+            );
```
mbjorkqvist, line 1362: Would it make sense to keep (part of) the original comment added in #11072 here, to explain that this should be unreachable for funding transactions (and mentioning that this would need to be adjusted in case that changes)?

```
// Unreachable for a funding: the destination is derived from the minter's own
// key, so a bare transfer there has no code to revert in. Were it reached, the
// status would be wrong, since nothing reimburses a funding — tolerable only
// because it cannot happen, and not worth a status of its own, which would mean
// adding a variant to `retrieve_eth_status`' return type and breaking existing
// clients. Revisit if funding ever goes through a contract, where a revert becomes
// possible.
```

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -425,7 +422,17 @@ pub enum ResubmitTransactionError {
     },
 }
 
-impl WithdrawalTransactions {
+/// How far a transaction has got through the pipeline. Carries the transaction itself, since
+/// every caller that asks the stage also wants the transaction at it.
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum TransactionStage<'a> {
```
mbjorkqvist: nit: Should this be `pub(in crate::state)` as well?

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -496,7 +497,10 @@ mod withdrawal_transactions {
                         access_list: Default::default(),
                     }
                 );
-                assert_eq!(transactions.next_nonce, TransactionNonce::from(i + 1));
+                assert_eq!(
+                    transactions.pipeline.next_nonce,
```
mbjorkqvist: Would it make sense to use the new accessor here (same for the test below)?

```suggestion
                    transactions.next_transaction_nonce(),
```

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -1789,6 +1805,7 @@ mod withdrawal_transactions {
             transactions.record_resubmit_transaction(created_tx.clone());
             assert!(
                 transactions
+                    .pipeline
                     .created_tx
                     .contains_alt(&cketh_ledger_burn_index)
             );
```
mbjorkqvist: Would it make sense to use the new accessor here also?

```suggestion
            assert_eq!(
                transactions.transaction_status(&cketh_ledger_burn_index),
                crate::state::transactions::RetrieveEthStatus::TxCreated
            );
```

mbjorkqvist (approved):
Thanks @gregorydemay! Just a few minor comments.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Both fixed in 0cd82ae. Bullets 5 and 6 are one step now — you were right that the lifecycle is five steps: comparing against master shows how it happened, since master's bullet 6 was the reimbursement rule and the rewrite that replaced it kept bullet 5's opening clause. Bullet 1 now reads "A withdrawal request", since `SweeperFunding` is a `WithdrawalRequest` variant on this very branch, so a request is already not necessarily a user's.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed in 0cd82ae, and your reading of the cause looks right — master (5425f421, lines 804-814) has the fee plus the tail explaining it, and this branch had truncated the message at "the burn is NOT reimbursed". Restored `receipt.effective_transaction_fee()` along with "no ETH reached the sweeper, the failed transaction still paid {} of gas, and the rest of the burn now over-backs ckETH", which is the part that makes the log actionable.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed in 0cd82ae, applied as suggested. Worth noting how precise this catch was: of the 21 `self.pipeline.*` uses in `mod.rs`, this was the only one reading a field rather than calling a method — the single remaining hole in the encapsulation the PR is for. Iteration order is unchanged, since `maybe_reimburse` is a `BTreeSet` and `processed_withdrawal_requests` a `BTreeMap`, both keyed by `LedgerBurnIndex`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Restored in 0cd82ae, in full rather than in part — master carries all seven lines at 954-960 and this branch had none of them. The reasoning it records (nothing reimburses a funding; a status of its own would mean a new `retrieve_eth_status` variant and would break existing clients; revisit if funding ever goes through a contract) is exactly the kind that gets re-litigated once it is gone.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Done in 0cd82ae. Checked it was safe first: `TransactionStage` is referenced only inside this module, and it is only ever reachable through `TransactionPipeline`, which is already `pub(in crate::state)` — so its effective visibility was capped there regardless and the change is a no-op to the compiler. It just stops the `pub` from suggesting a wider surface than exists.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
gregorydemay: 🤖 Done in 0cd82ae, both sites. Also swept the rest of the file while in here: the tests reached into private pipeline fields in eight places — `sent_tx` ×3, `next_nonce` ×2, `created_tx` ×2, `finalized_tx` ×1 — and all eight now go through the API. There is no accessor returning every sent transaction for one burn index, so `sent_transactions` and `first_sent_transaction` are test-local helpers over `sent_transactions_iter()`, rather than new production API that only tests would use.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
gregorydemay: 🤖 Done in 0cd82ae, applied as suggested. The `finalized_tx` comparison just below it had the same problem, so that one now goes through `finalized_transactions_iter()` — which also let `use crate::map::MultiKeyMap` go, as that assertion was its only user in the module.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Closing: this annotates lines being deleted, and the reimbursement lookup it points at now lives in the `WithdrawalTransactions` impl as `find_reimbursed_transaction_by_cketh_ledger_burn_index`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Closing: the move happened — `assert!(self.maybe_reimburse.insert(withdrawal_id))` is now in `WithdrawalTransactions::record_created_transaction`, behind the `is_reimbursable` check.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Closing: the move happened — `record_reimbursement_request` and the reimbursement tail of `record_finalized_transaction` are both in the `WithdrawalTransactions` impl.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Closing: the move happened — `record_finalized_reimbursement` is in the `WithdrawalTransactions` impl.

PR #11193, opened 2026-08-18
feat: Do not route canister messages from a cooling down subnet

mraszyk:
Messages (requests and responses) in canister output queues are no longer routed into streams while the source subnet is cooling down. They are retained in their output queues until the subnet stops cooling down, rather than being rejected or dropped. This includes the loopback stream, i.e. a cooling down subnet does not route its canisters' messages to its loopback stream either.

The check is merged into the existing destination side check, so a queue is skipped (and counted in `mr_cooling_down_skipped_queues`) exactly once per round, whether it is the source subnet, the destination subnet, or both that are cooling down. As before, messages from the subnet's own output queues are exempt.

### DIFF
--- a/rs/messaging/src/routing/stream_builder.rs
+++ b/rs/messaging/src/routing/stream_builder.rs
@@ -480,6 +480,9 @@ impl StreamBuilderImpl {
                 // Destination subnet found.
                 Some(dst_subnet_id) => {
                     let is_loopback_stream = self.subnet_id == dst_subnet_id;
+                    let dst_subnet_topology = network_topology.subnets().get(&dst_subnet_id);
+                    let dst_subnet_type = dst_subnet_topology
+                        .map_or(SubnetType::Application, |topology| topology.subnet_type);
 
                     // No messages from canister output queues are routed while either
                     // this subnet (the source) or the destination subnet is cooling down;
@@ -488,7 +491,7 @@ impl StreamBuilderImpl {
                     // cooling down anymore, rather than rejecting or dropping it.
                     if !is_from_subnet_queues
                         && (own_subnet_is_cooling_down
-                            || network_topology.is_cooling_down(&dst_subnet_id))
+                            || dst_subnet_topology.is_some_and(|topology| topology.cooling_down))
                     {
                         self.metrics.cooling_down_skipped_queues.inc();
                         output_iter.exclude_queue();
@@ -496,15 +499,7 @@ impl StreamBuilderImpl {
                     }
 
                     let dst_stream_entry = streams.entry(dst_subnet_id);
-                    if !is_loopback_stream
-                        && is_at_limit(
-                            &dst_stream_entry,
-                            network_topology
-                                .subnets()
-                                .get(&dst_subnet_id)
-                                .map_or(SubnetType::Application, |topology| topology.subnet_type),
-                        )
-                    {
+                    if !is_loopback_stream && is_at_limit(&dst_stream_entry, dst_subnet_type) {
                         // Stream full, skip all other messages to this destination.
                         output_iter.exclude_queue();
                         continue;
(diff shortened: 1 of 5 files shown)


rs/messaging/src/routing/stream_builder.rs
```diff
@@ -480,11 +481,15 @@ impl StreamBuilderImpl {
...
-                    // cooling down, rather than rejecting or dropping it.
-                    if !is_from_subnet_queues && network_topology.is_cooling_down(&dst_subnet_id) {
+                    // No messages from canister output queues are routed while either
+                    // this subnet (the source) or the destination subnet is cooling down;
+                    // not even into the loopback stream. Retain the message (along with
+                    // everything behind it in the same queue) until neither subnet is
+                    // cooling down anymore, rather than rejecting or dropping it.
+                    if !is_from_subnet_queues
+                        && (own_subnet_is_cooling_down
+                            || network_topology.is_cooling_down(&dst_subnet_id))
```
alin-at-dfinity: At this point, we look up the destination subnet's `SubnetTopology` at least 3 times in succession (always behind a flag or other, but in most cases at least twice). Let's look it up once and use that.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -480,11 +481,15 @@ impl StreamBuilderImpl {
...
-                    // from canister output queues. Retain the message (along with
-                    // everything behind it in the same queue) until the destination stops
-                    // cooling down, rather than rejecting or dropping it.
-                    if !is_from_subnet_queues && network_topology.is_cooling_down(&dst_subnet_id) {
+                    // No messages from canister output queues are routed while either
+                    // this subnet (the source) or the destination subnet is cooling down;
+                    // not even into the loopback stream. Retain the message (along with
+                    // everything behind it in the same queue) until neither subnet is
+                    // cooling down anymore, rather than rejecting or dropping it.
+                    if !is_from_subnet_queues
```
alin-at-dfinity, line 491: Rant: I again had to look this up and wrap my head around "we haven't implemented the logic for subnet queues, but we're arbitrarily treating them differently anyway".

Can we please address this next? And rename the field to `is_subnet_response`, so it better documents the intent?

rs/replicated_state/src/metadata_state.rs
```diff
@@ -435,7 +435,10 @@ pub struct SubnetTopology {
     ///  * (on all subnets) no messages are routed to a cooling down subnet --
     ///    including into the cooling down subnet's own loopback stream -- but
     ///    retained in their respective output queues, so that the respective
-    ///    streams can be emptied.
+    ///    streams can be emptied;
+    ///  * it routes no messages from its canisters' output queues into any of
+    ///    its streams -- including its own loopback stream -- but retains them
+    ///    in those output queues, so that its streams can be emptied.
```
```suggestion
    ///    retained in their respective output queues, so that streams to the
    ///    cooling down subnet can be emptied;
    ///  * it routes no messages from its canisters' output queues into streams
    ///    -- including its own loopback stream -- but retains them in output
    ///    queues, so streams from the cooling subnet can be emptied.
```

alin-at-dfinity (approved):

rs/messaging/src/routing/stream_builder.rs
mraszyk: Absolutely, subnet output queues are next on the list.

rs/replicated_state/src/metadata_state.rs
mraszyk: https://github.com/dfinity/ic/pull/11193/commits/11db3c4bc83a06edeb4215da0eff66661d37e021

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11193/commits/c82b68f5a9b9b52ca927794517be855368c199ce and then https://github.com/dfinity/ic/pull/11193/commits/867cf0da839b407507e2ea6d6f7ce18f937752a7

PR #11195, opened 2026-08-18
feat: Record canisters dropped by an online subnet split as deleted

mraszyk:
`ReplicatedState::online_split()` retained only the canisters hosted by this subnet, dropping the rest without recording an
`UnflushedCheckpointOp::DeleteCanister`. Their directories were therefore removed from tip only when the next checkpoint was created, by `FilterTipCanisters`.

Record the removals, so that the directories are deleted from tip at the next flush, and in order relative to the other checkpoint operations. `FilterTipCanisters` is retained as a safety net for canisters that disappear from the state without a corresponding operation.

The `drop_canister` test helpers, which emulated dropping a canister without recording the operation, are replaced by `ReplicatedState::remove_canister()` again, which now models an online split exactly.

### DIFF
--- a/rs/replicated_state/src/metadata_state.rs
+++ b/rs/replicated_state/src/metadata_state.rs
@@ -1335,7 +1335,7 @@ impl SystemMetadata {
             // retaining everything else on subnet A'.
             blockmaker_metrics_time_series,
             // Just updated by `ReplicatedState::online_split()`, adding delete operations
-            // for the snapshots of no longer hosted canisters.
+            // for the canisters dropped by the split.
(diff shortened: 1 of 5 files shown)


alin-at-dfinity (approved):
Why is this necessary / useful? And shouldn't `FilterTipCanisters` now record a critical error or log an error or something?

mraszyk:
> Why is this necessary / useful?

Discussed offline.

> And shouldn't `FilterTipCanisters` now record a critical error or log an error or something?

This will be done in a follow-up PR.

PR #11198, opened 2026-08-18
chore: Even better error messages for invalid keys

schneiderstefan:
This is a follow-up to commit 42f67fe giving fine-grained error messages for some more cases of invalid pubkeys.

### DIFF
--- a/rs/crypto/internal/crypto_lib/basic_sig/der_utils/src/lib.rs
+++ b/rs/crypto/internal/crypto_lib/basic_sig/der_utils/src/lib.rs
@@ -103,6 +103,12 @@ pub fn algo_id_and_public_key_bytes_from_der(
 /// DER-encoded key.
 pub const MAX_DER_NESTING_DEPTH: usize = 4;
 
+const SEQUENCE_TAG_NUMBER: u8 = 0x10;
+const SET_TAG_NUMBER: u8 = 0x11;
+
+const INCOMPLETE_ELEMENT: &str = "DER ends in the middle of an element";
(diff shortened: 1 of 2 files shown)


rs/crypto/internal/crypto_lib/basic_sig/der_utils/src/lib.rs
```diff
@@ -255,31 +260,47 @@ impl KeyDerParser {
...
             // Identifier octets: skip the high-tag-number form if present.
             let first_identifier_octet = der[index];
+            let universal_class = first_identifier_octet & 0xc0 == 0;
             let constructed = first_identifier_octet & 0x20 != 0;
+            let tag_number = first_identifier_octet & 0x1f;
+            let high_tag_number_form = tag_number == 0x1f;
             index += 1;
-            if first_identifier_octet & 0x1f == 0x1f {
+            if high_tag_number_form {
                 while index < der.len() && der[index] & 0x80 != 0 {
```
pierugo-dfinity: Feel free to ignore: maybe
```suggestion
                while let Some(length) = der.get(index) && length & 0x80 != 0 {
```
to be more "Rusty" and follow a similar style as below

pierugo-dfinity (approved):

rs/crypto/internal/crypto_lib/basic_sig/der_utils/src/lib.rs
schneiderstefan: Thank you for the suggestion. I called the variable octet though, because length is not quite correct.

PR #11200, opened 2026-08-18
fix: Reserve enough HTTP outcall cycles for any deliverable result to fund itself

eichhorl:
## Background
Under pay-as-you-go pricing, the canister making an HTTP outcall can compute an estimate of the expected cycles cost (and thus the recommended amount of cycles to attach), by calling the `ic0_cost_http_request_v2` function.

The attached cycles are then split evenly among all participating replicas, into the "per-replica allowances". Each replica consumes some of this allowance while making the HTTP request, and transforming and gossiping the response shares.

Aggregating the shares into the final response, which will be included and delivered as part of a block, costs a separate consensus fee. This consensus fee must be covered by the "unspent allowances" of replicas participating in the response.

## Problem
If the recommended amount of cycles returned by `ic0_cost_http_request_v2` tightly bounds the real cost of the outcall, this represents a liveness risk for the request.

Because the attached cycles are evenly split between participating replicas, this means that also the cycles intended to cover the consensus fee are split evenly between them. In the worst case, this means that aggregating a response into a block requires the unspent allowance of all participating replicas in order to cover the fee. For example:
- A fully replicated outcall on a 13 node subnet, means that the attached cycles are split evenly between all 13 nodes.
- All 13 nodes will make the HTTP request, and gossip their share
- In order to reach consensus on the response, only 9 (= n - f) responses need to agree.
- However, in order to cover the consensus cost, responses from all n replicas are required, leaving a fault tolerance of 0.

The same problem exists for flexible outcalls under most configurations. Notably, this problem does not exist for non-replicated outcalls, and more generally flexible outcalls where `min == max == total`, since in those cases all nodes making the call need to participate in the response regardless.

## Proposed Changes
With this PR we therefore increase the recommended amount of cycles to attach to the call. 

Specifically, for fully-replicated outcalls, we increase the cycles reserved for the consensus cost such that (after it is split into the allowances) shares from any set of `threshold` many nodes are enough to cover it.

For flexible outcalls, we increase the reserved cycles, such that each replica obtains enough allowance to cover for its own response to be included into the block.


rs/https_outcalls/pricing/src/fees.rs
```diff
@@ -1415,43 +1517,69 @@ mod tests {
...
-            // A committee of 4, whose worst case is not a multiple of 4.
+            // A committee of 11, i.e. fewer allowances than the subnet has replicas, and
+            // one whose worst case is not a whole number of them before the rounding.
             Replication::Flexible {
-                committee: (0..4).map(node).collect(),
+                committee: (0..11).map(node).collect(),
                 min_responses: 2,
-                max_responses: 3,
+                max_responses: 7,
             },
```
pierugo-dfinity: Maybe we could add more flexible edge-cases here (`total_requests == n`, `total_requests == 1`, `min == max == total_requests`, etc...)

pierugo-dfinity (approved):

rs/https_outcalls/pricing/src/fees.rs
eichhorl: Good idea, done: https://github.com/dfinity/ic/pull/11200/changes/ce7e8631e8429e8a5ce9fa5fcdabb00dcd1ac298

PR #11202, opened 2026-08-18
refactor: read HTTP/ECDSA outcall cycles from the per-use-case map

mraszyk:
The cycles consumed by HTTP and ECDSA outcalls are tracked both in the legacy scalar fields consumed_cycles_http_outcalls / consumed_cycles_ecdsa_outcalls of SubnetMetrics and in the consumed_cycles_by_use_case map, which a migration introduced in #10844 keeps in lockstep with those fields. #11219 fixed that migration to run once per round from the scheduler's `finish_round`, i.e. regardless of subnet activity, so the map entries can no longer go stale on a subnet that performs no outcalls.
I could now successfully confirm that the migration completed on all subnets with a DFINITY-owned node.

This switches all readers over to the map:

- get_consumed_cycles_http_outcalls / get_consumed_cycles_ecdsa_outcalls now read the HTTPOutcalls / ECDSAOutcalls entries of consumed_cycles_by_use_case, via a new helper that also backs the existing Schnorr / VetKd / dropped messages getters.
- consumed_cycles_total and consumed_cycles_total_v28 add those entries instead of the scalar fields.

The scalar fields are still observed and serialized, so that downgrading to an earlier replica version keeps observing the correct totals; only their reads are gone.

The fixtures that populate only the scalar fields (`canonical_encoding_subnet_metrics`, the `state_manager` tree hash fixture and `consumed_cycles_total_is_the_same_across_a_restart`) now run the migration first, the way the scheduler does once per round. The expected canonical encodings and state hashes are unchanged.


rs/replicated_state/src/metadata_state.rs
```diff
@@ -635,9 +633,16 @@ impl SubnetMetrics {
...
+                // covered by the scalar metric above and must be added to the
+                // total. The `ECDSAOutcalls` and `HTTPOutcalls` entries also
+                // cover the history held by the legacy
+                // `consumed_cycles_ecdsa_outcalls` /
+                // `consumed_cycles_http_outcalls` fields, which
+                // `migrate_outcalls_cycles_to_use_cases` folds into the map
+                // (hence the fields themselves are not read here).
+                CyclesUseCase::ECDSAOutcalls
+                | CyclesUseCase::HTTPOutcalls
+                | CyclesUseCase::SchnorrOutcalls
```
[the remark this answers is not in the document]
mraszyk: Good catch, confirmed: `canonical_encoding_subnet_metrics` failed with 100B instead of the expected 200B. Fixed in 04eb48386b by calling the (now public, since #11219) `migrate_outcalls_cycles_to_use_cases` in both fixtures before `refresh_consumed_cycles`, which is what the scheduler does once per round; the legacy scalar fields are left populated for downgrade coverage. Did the same in `consumed_cycles_total_is_the_same_across_a_restart`, which claimed to cover the scalar outcall metrics but no longer did.

rs/replicated_state/src/metadata_state.rs
```diff
@@ -481,14 +481,20 @@ impl SubnetMetrics {
         self.consumed_cycles_http_outcalls += cycles;
     }
 
+    /// Cycles consumed by HTTP outcalls. See `get_consumed_cycles_subnet_use_case`.
```
alin-at-dfinity: Nit: Rather than the redundant "See `get_consumed_cycles_subnet_use_case`" (it's an implementation detail and eminently visible in the implementation), I'd rather mention the `CyclesUseCase` variant, if anything.

```suggestion
    /// Cycles consumed by HTTP outcalls (`CyclesUseCase::HTTPOutcalls`).
```

Same everywhere below.

rs/replicated_state/src/metadata_state.rs
```diff
@@ -544,33 +550,33 @@ impl SubnetMetrics {
...
-    /// Cycles lost due to dropped messages. See `get_consumed_cycles_schnorr_outcalls`.
+    /// Cycles lost due to dropped messages. See `get_consumed_cycles_subnet_use_case`.
     pub fn get_consumed_cycles_dropped_messages(&self) -> NominalCycles {
+        self.get_consumed_cycles_subnet_use_case(CyclesUseCase::DroppedMessages)
+    }
+
+    /// Cycles consumed by a subnet-level use case, i.e. one that is never
+    /// charged to a canister's balance and can thus never originate from a
+    /// deleted canister: the `consumed_cycles_by_use_case` entry is exactly the
+    /// subnet-level consumption.
```
```suggestion
    /// Cycles consumed by a subnet-level use case, i.e. one that is never
    /// charged to a canister's balance and can thus never originate from a
    /// deleted canister.
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -544,33 +550,33 @@ impl SubnetMetrics {
...
+    ///
+    /// For `HTTPOutcalls` and `ECDSAOutcalls` the entry also covers the history
+    /// tracked in the legacy scalar fields, which
+    /// `migrate_outcalls_cycles_to_use_cases` folds into the map.
```
alin-at-dfinity: This is soon to be irrelevant and likely to be left behind. I'd rather drop it now.

```suggestion
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -635,9 +633,16 @@ impl SubnetMetrics {
...
-                // covered by any of the scalar metrics above and must be added to
-                // the total.
-                CyclesUseCase::SchnorrOutcalls
+                // covered by the scalar metric above and must be added to the
+                // total. The `ECDSAOutcalls` and `HTTPOutcalls` entries also
+                // cover the history held by the legacy
+                // `consumed_cycles_ecdsa_outcalls` /
+                // `consumed_cycles_http_outcalls` fields, which
+                // `migrate_outcalls_cycles_to_use_cases` folds into the map
+                // (hence the fields themselves are not read here).
```
alin-at-dfinity: Same here:

```suggestion
                // total.
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -683,23 +688,26 @@ impl SubnetMetrics {
     /// to and including `V28`; from `V29` on the consumer reports
     /// [`Self::consumed_cycles_total_including_canisters`], which does not
     /// double count.
+    ///
+    /// The HTTP / ECDSA outcall contribution used to be read from the legacy
+    /// scalar fields; it is now read from the corresponding
+    /// `consumed_cycles_by_use_case` entries, which
+    /// `migrate_outcalls_cycles_to_use_cases` keeps equal to those fields.
```
alin-at-dfinity: And here. After this change, no one will care where the values were taken from before.

```suggestion
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -683,23 +688,26 @@ impl SubnetMetrics {
...
 
         for (use_case, cycles) in self.consumed_cycles_by_use_case.iter() {
             match use_case {
-                // For ecdsa outcalls, http outcalls and deleted canisters, skip
-                // updating the total using the use case specific metric as the
-                // update above should be sufficient (the old metric is a superset).
+                // For deleted canisters, skip updating the total using the use
+                // case specific metric as the update above should be sufficient
+                // (the old metric is a superset).
+                CyclesUseCase::DeletedCanisters => {}
```
alin-at-dfinity, line 698: Clueless question: why didn't we also migrate away from `consumed_cycles_by_deleted_canisters`?

alin-at-dfinity (approved):

rs/replicated_state/src/metadata_state.rs
mraszyk: because that scalar field [combines](https://github.com/dfinity/ic/pull/11202/commits/4c5e4ca0a87ac20fa9558bdf703c637c78e7e818) the deleted canister's left-over balance (the dedicated `DeletedCanisters` use case) and its consumption across all canister-level use cases and thus cannot be migrated into the map; on the other hand, it [predates](https://github.com/dfinity/ic/pull/11202/commits/1e39ec3d47928879d20bc2d987492337a703da59) the map and thus contains more historical data than the map itself.

rs/replicated_state/src/metadata_state.rs
mraszyk: Done

rs/replicated_state/src/metadata_state.rs
mraszyk: Done

rs/replicated_state/src/metadata_state.rs
mraszyk: Done

rs/replicated_state/src/metadata_state.rs
mraszyk: Done

rs/replicated_state/src/metadata_state.rs
mraszyk: Done

rs/replicated_state/src/metadata_state.rs
mraszyk: But note that this particular function is buggy (called `consumed_cycles_total_v28` to indicate that it is only relevant for certification versions up to v28).

PR #11205, opened 2026-08-18
feat(subnet-splitting): preemptively load the post-split DKG transcripts in the key manager

pierugo-dfinity:
When a subnet split is scheduled, the DKG interval of the `Scheduled` summary is skipped. Its next transcripts will thus not be used and current transcripts were already loaded by the previous summary's next transcripts. 

This PR proposes to make the DKG key manager load the transcripts of the soon-to-be-created `PostSplit` summary instead of the `Scheduled` summary as soon as it observes it. This is needed to have the necessary key material needed to sign shares for the post-split CUP. It also drives the first interval following the split.

We also make sure to retain the transcripts in the concurrent key removal so the keys it just loaded don't get dropped again.

### DIFF
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -252,7 +252,7 @@ impl DkgKeyManager {
                         warn!(
                             every_n_seconds => 5,
                             self.logger,
-                            "Couldn't get the next DKG summary for the new subnet after the split: {err:?}"
+                            "Couldn't compute the post-split DKG summary for the new subnet after the split: {err}"
                         );
                         return;
                     }
@@ -308,7 +308,7 @@ impl DkgKeyManager {
             scheduled,
         )
         .map_err(|err| {
-            format!("Couldn't determine the post-split subnet assignment after the split: {err:?}")
+            format!("Couldn't determine the post-split subnet assignment after the split: {err}")
         })?
         .new_subnet_id;
 
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -640,7 +640,7 @@ mod tests {
     use ic_registry_keys::make_catch_up_package_contents_key;
     use ic_test_utilities_logger::with_test_replica_logger;
     use ic_test_utilities_registry::SubnetRecordBuilder;
-    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id};
+    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id, test_replica_version};
     use ic_types::{
         NodeId, RegistryVersion, SubnetId,
         consensus::{
@@ -828,6 +828,7 @@ mod tests {
                     node_id: local_node_id,
                     // The local node always start in the source subnet.
                     subnet_id: source_subnet_id,
+                    replica_version: test_replica_version(),
                 })
                 .build();
 
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -297,7 +297,7 @@ impl DkgKeyManager {
     }
 
     fn get_post_split_summary(
-        &mut self,
+        &self,
         summary_block: &Block,
         scheduled: SplittingArgs,
     ) -> Result<DkgSummary, String> {
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -643,7 +643,6 @@ mod tests {
     use ic_test_utilities_types::ids::{node_test_id, subnet_test_id};
     use ic_types::{
         NodeId, RegistryVersion, SubnetId,
-        backwards_compatibility::BackwardsCompatible,
         consensus::{
             BlockPayload, HashedBlock, Payload,
             dkg::{SplittingArgs, SubnetSplittingStatus},
@@ -908,12 +907,10 @@ mod tests {
                     .with_current_transcripts(current_transcripts);
 
                 splitting_summary.dkg.subnet_splitting_status =
-                    BackwardsCompatible::new_for_test_only(Some(SubnetSplittingStatus::Scheduled(
-                        SplittingArgs {
-                            source_subnet_id,
-                            destination_subnet_id,
-                        },
-                    )));
+                    SubnetSplittingStatus::Scheduled(SplittingArgs {
+                        source_subnet_id,
+                        destination_subnet_id,
+                    });
                 splitting_block.payload = Payload::new(
                     crypto_hash,
                     BlockPayload::Summary(splitting_summary.clone()),
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -854,7 +854,7 @@ mod tests {
                 )
                 .with_replica_config(ReplicaConfig {
                     node_id: local_node_id,
-                    // The local node always start in the source subnet.
+                    // The local node always starts in the source subnet.
                     subnet_id: source_subnet_id,
                     replica_version: test_replica_version(),
                 })
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -234,7 +234,7 @@ impl DkgKeyManager {
             let summary = if let Some(scheduled) =
                 subnet_splitting::is_split_scheduled(&summary_block)
             {
-                // If a subnet split is in progress, we should skip the `Scheduled` summary's
+                // If a subnet split is in progress, we can skip the `Scheduled` summary's
                 // transcripts (as the interval is skipped) and instead load the post-split one's.
                 match self.get_post_split_summary(&summary_block, scheduled) {
                     Ok(post_split_summary) => {
--- a/rs/consensus/dkg/src/dkg_key_manager.rs
+++ b/rs/consensus/dkg/src/dkg_key_manager.rs
@@ -23,7 +23,6 @@ use ic_types::{
 use prometheus::{HistogramVec, IntCounterVec, IntGauge, IntGaugeVec};
 use std::{
     collections::{HashMap, HashSet},
-    hash::Hash,
     sync::{
         Arc,
         mpsc::{Receiver, sync_channel},
@@ -524,28 +523,15 @@ impl DkgKeyManager {
 
         while let Some(summary_block) = next_summary_block {
             let summary = &summary_block.payload.as_ref().as_summary().dkg;
-            let next_summary_height = summary.get_next_start_height();
-            for transcript in summary
-                .current_transcripts()
-                .values()
-                .chain(summary.next_transcripts().values())
-            {
-                transcripts_to_retain.insert_by_cloning_if_not_present(transcript);
-            }
+            insert_transcripts_of(&mut transcripts_to_retain, summary);
 
-            next_summary_block = pool_reader.get_finalized_block(next_summary_height);
+            next_summary_block = pool_reader.get_finalized_block(summary.get_next_start_height());
         }
 
         // The removal below runs concurrently with the loading of the `summaries_to_load`'s
         // transcripts, so they must be part of the retain set.
         for summary in summaries_to_load {
-            for transcript in summary
-                .current_transcripts()
-                .values()
-                .chain(summary.next_transcripts().values())
-            {
-                transcripts_to_retain.insert_by_cloning_if_not_present(transcript);
-            }
+            insert_transcripts_of(&mut transcripts_to_retain, summary);
         }
 
         let crypto = self.crypto.clone();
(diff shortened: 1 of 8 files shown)


rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -217,8 +231,34 @@ impl DkgKeyManager {
...
+                            "Couldn't get the next DKG summary for the new subnet after the split: {err:?}"
+                        );
+                        return;
+                    }
+                }
+            } else {
+                Cow::Borrowed(&summary_block.payload.as_ref().as_summary().dkg)
+            };
+
+            self.update_dkg_metrics(&summary);
```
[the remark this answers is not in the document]
pierugo-dfinity: This is already the case for recovery CUPs. I think we can keep that way (I also find it useful to receive such an expected alert during an emergency scenario to "confirm" everything works).

rs/replay/src/validator.rs
```diff
@@ -174,6 +174,8 @@ impl ReplayValidator {
             self.consensus_crypto.clone(),
             self.log.clone(),
             pool_reader,
+            self.registry.clone(),
+            self.replica_cfg.clone(),
```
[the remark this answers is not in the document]
pierugo-dfinity: Subnet splitting anyways is not yet properly supported in `ic-replay`.

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -217,8 +231,34 @@ impl DkgKeyManager {
...
+                        error!(
+                            self.logger,
+                            "Couldn't get the next DKG summary for the new subnet after the split: {err:?}"
+                        );
+                        return;
+                    }
+                }
+            } else {
+                Cow::Borrowed(&summary_block.payload.as_ref().as_summary().dkg)
+            };
```
eichhorl: Since we still create configs to reshare the next transcripts of this summary, the DKG client will probably still attempt to create dealings for these configs, and most likely fail. I guess this is fine, but it could be noisy too. More generally, would it simplify the code if we just loaded everything?

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -230,12 +270,37 @@ impl DkgKeyManager {
...
-            self.last_dkg_summary_height = Some(summary_block.height);
+            //
+            // Note that the removal we trigger here still runs in parallel with the loads
+            // below, so it is only harmless as long as the set of transcripts it retains is a
+            // superset of the ones we are about to load. This is why we hand `summary` to
+            // `delete_inactive_keys`: for a regular summary it is redundant, but the post-split
+            // summary is computed locally and thus invisible to the removal otherwise.
+            self.delete_inactive_keys(pool_reader, &summary);
+            self.load_transcripts_from_summary(&summary);
+            self.last_dkg_summary_height = Some(summary.height);
```
eichhorl: Would it make sense to continue using `summary_block.height` here, which is the value we compared against above? Then this code block would execute again when we actually have the post split CUP. Otherwise it wouldn't be called again on the source subnet until the first real CUP is created, but I guess the only consequence should be that some keys are deleted a bit later.
```suggestion
            self.last_dkg_summary_height = Some(summary_block.height);
```

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -409,8 +474,11 @@ impl DkgKeyManager {
     }
 
     /// Ask the CSP to drop DKG key material related to transcripts that are no
-    /// longer relevant
-    fn delete_inactive_keys(&mut self, pool_reader: &PoolReader<'_>) {
+    /// longer relevant.
+    ///
+    /// The transcripts of `summary_to_load`, i.e. the summary whose transcripts the caller is
+    /// about to load, are always retained. See the comment at the call site.
+    fn delete_inactive_keys(&mut self, pool_reader: &PoolReader<'_>, summary_to_load: &DkgSummary) {
```
eichhorl: We could also make `summary_to_load` optional

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: > the DKG client will still attempt to create dealings and fail

Ah I was only thinking about the payload builder, and thought no dealings were created because blocks were empty. But you're right that the DKG client will still attempt to create them. I think this isn't too too bad.

> More generally, would it simplify the code if we just loaded everything?

This is what I used to do but ended up going with the current implementation. The biggest challenge for me was error-handling:
- If we load the scheduled & post-split summaries and update `last_dkg_summary_height` only on success of loading the post-split summary: if loading the post-split summary fails, we would try again and enter `delete_inactive_keys` which will block the entire thread until the previous deletion is finished. This is on the hot path of Consensus and will affect Consensus quite badly. We will also retry loading the scheduled summary (not a big problem, but not optimal).
- If we decide instead to update `last_dkg_summary_height` in any case (ignoring errors of loading the post-split summary): if loading the post-split summary fails, we will not retry (potentially stalling).
- Another alternative is to retry only the post-split summary. But this would require keeping track of it separately, with its own retry/done state (a second field, a second guard, etc) which quickly becomes complex.

Additionally, the current implementation also acts as documentation: readers understand the scheduled summary is skipped as we do not attempt to load its transcripts.

Finally, it is also more interesting performance-wise (though maybe negligible). Loading the post-split transcripts is on the hot-path of subnet-splitting: we cannot create the post-split CUP without loading its transcripts, and this is possible only as soon as the scheduled summary is observed.
Additionally loading the scheduled summary's transcripts will be extra unnecessary work for the crypto vault.

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: > which is the value we compared against above

I like this argument. Also the fact that we'd be deleting obsolete keys more quickly.

Though I'm still unsure, because I think using `summary.height` is more truthful: the last loaded DKG summary height truly was the post-split height, and our job is done, we shouldn't try to enter the loop a second time in the future.
Under normal operation before my changes, we can figure out that the state machine will first enter the "summary code block" when observing the summary, then enter the "CUP code block" when observing the CUP and that's it, then the job is done for the remainder of the interval. With your suggestion, this flow would not be true anymore, and we'd execute the "summary code block" a second time, which could maybe have some unintentional behaviour.

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: We could. Though I don't think the current style hurts too much and I even think it's more defensive. In case some bug is introduced in the way we're counting the transcripts to retain, we are at least 100% sure that the ones we are about to load are protected.

Though if you want I can update the docs to abstract away that it's precisely for subnet-splitting. I.e. I can write something about this new argument being a first-class citizen that should be considered "not just for subnet-splitting".

eichhorl: Fine with me, I just tried to find a way to avoid the extra clones, but shouldn't be a big deal

rs/consensus/dkg/src/dkg_key_manager.rs
eichhorl: One disadvantage of using `summary.height` is that we would likely `consensus_dkg_intervals_failed` a second time, so ideally we would also fix that at the same time

rs/consensus/dkg/src/dkg_key_manager.rs
eichhorl: Couldn't we do something like this:
```rust
let mut summaries_to_load = vec![summary];
if let Some(post_split_summary) = &post_split_summary {
    summaries_to_load.push(post_split_summary);
}
self.delete_inactive_keys(pool_reader, &summaries_to_load);
for summary in &summaries_to_load {
    self.load_transcripts_from_summary(summary);
}
```
Then the order of operations shouldn't change regardless of errors.

Not loading the summary is also fine with me, but then we should at least do something about the logs, otherwise we will get 1 error log per node every 200ms.

Another consequence is that `is_transcript_loaded` will lie about the skipped summary, since it only compares the watermarks.

pierugo-dfinity: Ok I like your solution. It extends from having to load a single summary per height to multiple ones. And if we there's an error in determining one, we retry next tick without having deleted or loaded anything.

So this solves the DKG client problem.

Also concerning my argument on performance, I forgot we also need to wait for the manifest, which could take multiple seconds on a large subnet (which is the whole point of splitting), so not sure we're on the hot path anymore.

> Another consequence is that is_transcript_loaded will lie about the skipped summary

And this is one more argument in your favour, I hadn't seen that, thanks.
~Now I guess it does still lie, but more conservatively, i.e. it can return `false` even though the transcript was actually loaded (for the post-split summary), but it can't return `true` even though the transcript was not loaded. And it is definitely better to be more conservative.~
Actually it doesn't even lie because the post-split transcripts will be `Remote`.

Implemented here. https://github.com/dfinity/ic/pull/11205/commits/983524952eec1a51d8f33355da82247f7dbcd91b

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: Yeah I think in the end the code will be easier to follow with `summary_block.height`. Implemented as part of the other comment's changes.

> `summary.height`

You mean `summary_block.height`? With the proposed implementation, we still report the scheduled summary (which wouldn't fire the alert). And it would only fire once when the post-split CUP is created.

I guess we could add a check not to fire it if `SubnetSplittingStatus` is `PostSplit`, but I think it's great to have some "confirmation" that it works as intended, like for recovery. WDYT?

(Well *actually*, it will fire once before the replicas restart and once more after, so indeed twice. But not three times!)

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: Maybe this is better? https://github.com/dfinity/ic/pull/11205/commits/f0149e36e547b0f6c4e21c8704ece3f9c6dadac6

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -217,8 +231,41 @@ impl DkgKeyManager {
...
+                            post_split_summary.height
+                        );
+
+                        Some(post_split_summary)
+                    }
+                    Err(err) => {
+                        warn!(
+                            every_n_seconds => 5,
+                            self.logger,
+                            "Couldn't get the next DKG summary for the new subnet after the split: {err:?}"
```
eichhorl:
```suggestion
                            "Couldn't compute the post-split DKG summary for the new subnet after the split: {err}"
```

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -230,12 +277,44 @@ impl DkgKeyManager {
...
+
+            // Even though we might have actually loaded a summary with higher height (the
+            // post-split summary), we still drive the state machine consistently with the if-guard
+            // above.
             self.last_dkg_summary_height = Some(summary_block.height);
         }
     }
 
+    fn get_post_split_summary(
+        &mut self,
```
eichhorl:
```suggestion
        &self,
```

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -230,12 +277,44 @@ impl DkgKeyManager {
...
+        scheduled: SplittingArgs,
+    ) -> Result<DkgSummary, String> {
+        let new_subnet_id = subnet_splitting::get_post_split_subnet_assignment(
+            self.replica_config.node_id,
+            summary_block,
+            self.registry.as_ref(),
+            scheduled,
+        )
+        .map_err(|err| {
+            format!("Couldn't determine the post-split subnet assignment after the split: {err:?}")
```
eichhorl:
```suggestion
            format!("Couldn't determine the post-split subnet assignment after the split: {err}")
```

rs/consensus/dkg/src/dkg_key_manager.rs
```diff
@@ -444,12 +530,24 @@ impl DkgKeyManager {
...
+        // The removal below runs concurrently with the loading of the `summaries_to_load`'s
+        // transcripts, so they must be part of the retain set.
+        for summary in summaries_to_load {
+            for transcript in summary
+                .current_transcripts()
+                .values()
+                .chain(summary.next_transcripts().values())
+            {
+                transcripts_to_retain.insert_by_cloning_if_not_present(transcript);
+            }
```
eichhorl: Alternatively, this for loop could be a function `insert_transcripts_of` that we call here and above. Then we could inline the new trait implementation there.

eichhorl (approved):

rs/consensus/dkg/src/dkg_key_manager.rs
eichhorl: LGTM 👍

rs/consensus/dkg/src/dkg_key_manager.rs
eichhorl: Works for me, posted an alternative below 👍

rs/consensus/dkg/src/dkg_key_manager.rs
eichhorl: Yes, sorry, I meant `summary_block.height`. Fine with me! 👍

rs/consensus/dkg/src/dkg_key_manager.rs
pierugo-dfinity: Good idea, simpler

PR #11207, opened 2026-08-18
chore: Cleanup Type 4.0 deployment path

Bownairo:
4.0 is and was not used anywhere.

Type 4.0 originally was used to experiment with overcommitting vCPUs, and this later swapped with and replaced type 4.1.


blind-oracle:
4.0 isn't/wasn't used anywhere?

Bownairo:
> 4.0 isn't/wasn't used anywhere?

I just added a little blurb to the description. Maybe @r-birkner can confirm for sure that nobody is using this?

blind-oracle (approved):

frankdavid (approved):

r-birkner:
Yes, type4.0 was never used. We only use type4 and then type4.1 to type4.5

PR #10939, opened 2026-07-28
test(engines): System test for standard engine replica version.

daniel-wong-dfinity-org-twin:
This also relaxes what values can be used as Replica version IDs in UpdateStandardEngineReplicaVersion NNS proposals to match ReplicaVersion.

# Future Work

I'd like to make ReplicaVersion validation be stricter. E.g. it currently allows blank! I think the ideal rule would be that Replica version ID strings must be of the form `${GIT_COMMIT_ID}(-test)?`.

# References

[👈 Previous PR][prev]

[prev]: https://github.com/dfinity/ic/pull/10885

PR #11208, opened 2026-08-18
feat(system-tests): support API boundary node playnets on the local backend

basvandijk:
`with_api_boundary_nodes_playnet` gives API boundary nodes a domain name and a certificate for it that the nodes trust. Both halves came from Farm, so every system-test using it — and therefore every system-test with a cloud engine subnet, which requires it — was pinned to `backend = "farm"`.

This stands up an equivalent locally.

### DNS in the local backend

The group's `dnsmasq` becomes its DNS server. It already ran on the group bridge as an RA/DHCPv4 daemon with `--port=0`; dropping that and adding `--no-resolv --no-hosts` makes it a hermetic resolver — with no upstream left to forward to, anything it cannot answer is `REFUSED`. It answers from two sources: an `--addn-hosts` file, written by the new `LocalBackend::add_dns_record` and re-read on `SIGHUP`, and a `--synth-domain` that mirrors the public `nip.io` wildcard service (so `canister_http_socks_test` needs no DNS-specific code and stays identical on both backends).

GuestOS has no name-server knob and boots with `IPv6AcceptRA=no`, so rather than reconfiguring the guests, `create_group` assigns the four addresses GuestOS is hard-coded to query to the bridge. Inside the backend's own network namespace those addresses are free and no query can escape, so every node gets a working resolver without touching IC-OS. Those addresses are now a single `IPV6_NAME_SERVERS` constant, kept in sync with the networkd contents by a unit test.

### A local playnet

`InternetComputer::setup_api_bn_local_playnet` issues an ephemeral CA plus a leaf covering the API boundary nodes' domains, registers those domains with the group's `dnsmasq`, and stores the material in a new `LocalApiBoundaryNodesPlaynet` attribute — deliberately separate from `Playnet`, which the IC gateway VM also reads and writes. `bootstrap` serves the leaf from `ic-boundary` through the existing `ic_boundary_tls_cert` mechanism, so that pipeline is unchanged.

### The one production change

`nns_delegation_manager` built its root store from the compiled-in `webpki_roots` only, with no config, env var, feature or `cfg` hook — no test-issued certificate could ever satisfy it. It now also honours `extra_api_boundary_node_trust_anchors_pem`, a new dev-only `GuestOSDevSettings` field (config version 1.15.0 → 1.16.0). Production nodes get `None` and keep the public roots as their only trust anchors.

**This is the part that needs a careful look from the delegation-manager owners.**

### Tests enabled

`backend = "farm"` dropped from `canister_http_socks_test`, `cloud_engine_canister_sig_test`, `xnet_cloud_engine_isolation_test`, `nns_delegation_branch_nns_version_test` and `delete_subnet_test`. The `cpus` of `canister_http_socks_test` was understated — its comment omitted the four cloud engine nodes.

Two tests stay on Farm for unrelated reasons, now recorded accurately in place of the old comment:

* `nns_delegation_mainnet_nns_version_test` boots the *mainnet* GuestOS, whose replica predates the new field. It can be enabled once that has rolled out.
* `firewall_correctness_test` asserts that port 8080 is closed between particular pairs of nodes, but on the local backend `ic-prep` always adds 8080 to the `fd00::/8` rule the driver needs to reach the nodes at all — and the nodes themselves live in `fd00::/8`. Enabling it means narrowing that rule to the driver's own addresses, which changes shared local-backend behaviour for every local test and is left as a follow-up.

### Verification

Run on the local backend, all passing: `cloud_engine_canister_sig_test_local`, `xnet_cloud_engine_isolation_test_local`, `nns_delegation_branch_nns_version_test_local`, `delete_subnet_test_local`. Their logs show a cloud engine node resolving `apibn-0.ic.net` through the group's `dnsmasq`, completing a TLS handshake against the ephemeral CA, and fetching its NNS delegation.

`canister_http_socks_test_local` is enabled but was not run: at 11 node VMs plus a UVM it needs ~48 GiB, more than the machine I verified on had.

### DIFF
--- a/rs/ic_os/config/types/compatibility_tests/fixtures/guestos_v1.16.0.json
+++ b/rs/ic_os/config/types/compatibility_tests/fixtures/guestos_v1.16.0.json
@@ -1,0 +1,53 @@
+{
+  "config_version": "1.16.0",
+  "network_settings": {
+    "ipv6_config": {
+      "Fixed": {
+        "address": "2a00:fb01:400:200::1/64",
+        "gateway": "2a00:fb01:400:200::1"
+      }
+    },
+    "ipv4_config": {
+      "address": "192.168.1.1",
+      "gateway": "192.168.1.254",
+      "prefix_length": 24
+    },
+    "domain_name": "ic.test"
+  },
+  "icos_settings": {
+    "node_reward_type": "type3.1",
+    "mgmt_mac": "00:00:00:00:00:01",
+    "deployment_environment": "mainnet",
+    "nns_urls": [
+      "https://icp-api.io,https//icp0.io,https://ic0.app"
+    ],
+    "node_operator_private_key": null,
+    "enable_trusted_execution_environment": false,
+    "use_ssh_authorized_keys": false,
+    "icos_dev_settings": {}
+  },
+  "guestos_settings": {
+    "guestos_dev_settings": {
+      "backup_spool": null,
+      "malicious_behavior": null,
+      "query_stats_epoch_length": null,
+      "bitcoind_addr": null,
+      "dogecoind_addr": null,
+      "jaeger_addr": null,
+      "socks_proxy": null,
+      "hostname": null,
+      "generate_ic_boundary_tls_cert": null,
+      "ic_boundary_tls_cert": null,
+      "extra_api_boundary_node_trust_anchors_pem": null,
+      "nns_pub_key_override": null
+    }
+  },
+  "guest_vm_type": "default",
+  "upgrade_config": {
+    "peer_guest_vm_address": "2a00:fb01:400:200:6801:95ff:fed7:d475"
+  },
+  "trusted_execution_environment_config": {
+    "sev_cert_chain_pem": "-----BEGIN CERTIFICATE----------END CERTIFICATE-----"
+  },
+  "recovery_config": null
+}
--- a/rs/ic_os/config/types/compatibility_tests/fixtures/guestos_v1.16.0.json
+++ b/rs/ic_os/config/types/compatibility_tests/fixtures/guestos_v1.16.0.json
@@ -19,7 +19,9 @@
     "mgmt_mac": "00:00:00:00:00:01",
     "deployment_environment": "mainnet",
     "nns_urls": [
-      "https://icp-api.io,https//icp0.io,https://ic0.app"
+      "https://icp-api.io/",
+      "https://icp0.io/",
+      "https://ic0.app/"
     ],
     "node_operator_private_key": null,
     "enable_trusted_execution_environment": false,
--- a/rs/ic_os/config/types/compatibility_tests/fixtures/hostos_v1.16.0.json
+++ b/rs/ic_os/config/types/compatibility_tests/fixtures/hostos_v1.16.0.json
@@ -1,0 +1,53 @@
+{
+  "config_version": "1.16.0",
+  "network_settings": {
+    "ipv6_config": {
+      "Fixed": {
+        "address": "2a00:fb01:400:200::1/64",
+        "gateway": "2a00:fb01:400:200::1"
+      }
+    },
+    "ipv4_config": {
+      "address": "192.168.1.1",
+      "gateway": "192.168.1.254",
+      "prefix_length": 24
+    },
+    "domain_name": "ic.test"
+  },
+  "icos_settings": {
+    "node_reward_type": "type3.1",
+    "mgmt_mac": "00:00:00:00:00:01",
+    "deployment_environment": "mainnet",
+    "nns_urls": [
+      "https://icp-api.io,https//icp0.io,https://ic0.app"
+    ],
+    "node_operator_private_key": null,
+    "enable_trusted_execution_environment": false,
+    "use_ssh_authorized_keys": false,
+    "icos_dev_settings": {}
+  },
+  "hostos_settings": {
+    "hostos_dev_settings": {
+      "vm_memory": 16,
+      "vm_cpu": "kvm",
+      "vm_nr_of_vcpus": 64
+    },
+    "verbose": false
+  },
+  "guestos_settings": {
+    "guestos_dev_settings": {
+      "backup_spool": null,
+      "malicious_behavior": null,
+      "query_stats_epoch_length": null,
+      "bitcoind_addr": null,
+      "dogecoind_addr": null,
+      "jaeger_addr": null,
+      "socks_proxy": null,
+      "hostname": null,
+      "generate_ic_boundary_tls_cert": null,
+      "ic_boundary_tls_cert": null,
+      "extra_api_boundary_node_trust_anchors_pem": null,
+      "nns_pub_key_override": null
+    }
+  }
+}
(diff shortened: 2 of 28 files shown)


rs/ic_os/config/types/compatibility_tests/fixtures/hostos_v1.16.0.json
```diff
@@ -0,0 +1,53 @@
...
+      "prefix_length": 24
+    },
+    "domain_name": "ic.test"
+  },
+  "icos_settings": {
+    "node_reward_type": "type3.1",
+    "mgmt_mac": "00:00:00:00:00:01",
+    "deployment_environment": "mainnet",
+    "nns_urls": [
+      "https://icp-api.io,https//icp0.io,https://ic0.app"
```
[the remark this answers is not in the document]
basvandijk: Good catch, and it went deeper than a typo: the fixture generator passed the whole comma-separated list to `Url::parse`, which produces a single `Url` with host `icp-api.io,https` and path `//icp0.io,https://ic0.app`. Re-serialising that gives the malformed string you spotted. It has been in every fixture since v1.0.0.

Fixed the generator to emit one `Url` per entry and regenerated the two v1.16.0 fixtures (3877a4e4b9), which this branch introduces, so they now read:

```json
"nns_urls": [
  "https://icp-api.io/",
  "https://icp0.io/",
  "https://ic0.app/"
]
```

The older fixtures deliberately keep the malformed value — they are historical records of what each released config version looked like, and the README requires they stay byte-for-byte as generated. All 34 still deserialize.

rs/ic_os/config/types/compatibility_tests/fixtures/guestos_v1.16.0.json
```diff
@@ -0,0 +1,53 @@
...
+      "prefix_length": 24
+    },
+    "domain_name": "ic.test"
+  },
+  "icos_settings": {
+    "node_reward_type": "type3.1",
+    "mgmt_mac": "00:00:00:00:00:01",
+    "deployment_environment": "mainnet",
+    "nns_urls": [
+      "https://icp-api.io,https//icp0.io,https://ic0.app"
```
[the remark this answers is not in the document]
basvandijk: Good catch, and it went deeper than a typo: the fixture generator passed the whole comma-separated list to `Url::parse`, which produces a single `Url` with host `icp-api.io,https` and path `//icp0.io,https://ic0.app`. Re-serialising that gives the malformed string you spotted. It has been in every fixture since v1.0.0.

Fixed the generator to emit one `Url` per entry and regenerated the two v1.16.0 fixtures (3877a4e4b9), which this branch introduces, so they now read:

```json
"nns_urls": [
  "https://icp-api.io/",
  "https://icp0.io/",
  "https://ic0.app/"
]
```

The older fixtures deliberately keep the malformed value — they are historical records of what each released config version looked like, and the README requires they stay byte-for-byte as generated. All 34 still deserialize.

basvandijk:
Picking up the suppressed comment from the last Copilot review (`canister_http_socks_test.rs:162`): correct, that cross-reference pointed at `LocalBackend::start_ra_daemon`, which this PR renames to `start_dnsmasq`.

Fixed in 6c287ca019, along with three comments in `local_backend.rs` that still called it "the RA daemon" — it now serves RA, DHCPv4 *and* DNS, which is the whole reason for the rename. No remaining references to the old names.

daniel-wong-dfinity-org-twin (approved):

rs/ic_os/config/types/compatibility_tests/src/fixture.rs
```diff
@@ -66,8 +66,14 @@ fn generate_hostos_config() -> HostOSConfig {
...
         enable_trusted_execution_environment: false,
+        // One `Url` per entry. Passing the whole comma-separated list to
+        // `Url::parse` yields a single `Url` whose host is `icp-api.io,https`
+        // and whose path is the rest, which re-serialises to the malformed
+        // `https://icp-api.io,https//icp0.io,https://ic0.app`.
         nns_urls: vec![
-            url::Url::parse("https://icp-api.io,https://icp0.io,https://ic0.app").unwrap(),
+            url::Url::parse("https://icp-api.io").unwrap(),
+            url::Url::parse("https://icp0.io").unwrap(),
+            url::Url::parse("https://ic0.app").unwrap(),
```
pierugo-dfinity, line 76: I think it used to work because the registry replicator [expects](https://github.com/dfinity/ic/blob/b074ca2612105998f34bdb81d778b5d955d3b0d0/rs/orchestrator/registry_replicator/src/args.rs#L33) one comma-separated string. Splitting it into multiple entries should work because we then [join](https://github.com/dfinity/ic/blob/b074ca2612105998f34bdb81d778b5d955d3b0d0/rs/ic_os/config/tool/src/guestos/generate_ic_config.rs#L129-L135) them with a comma, and I agree it looks cleaner, but just wanted to leave a comment to point it out in case anything else is silently depending on the previous behaviour.

rs/tests/networking/firewall/firewall_correctness_test.rs
```diff
@@ -180,13 +185,19 @@ async fn add_necessary_ports_registry_rule(nns_node: &IcNodeSnapshot, log: &Logg
         user: None,
         direction: Some(FirewallRuleDirection::Inbound as i32),
     };
+    // The global scope does not necessarily start out empty — the local backend
+    // seeds a rule that lets the test driver reach the nodes — and the proposal
+    // has to carry a hash of the rule set it applies to.
+    let previous_rules = topology_snapshot
+        .firewall_rules(&FirewallRulesScope::Global)
+        .expect("Could not read the global firewall rules");
```
pierugo-dfinity: Since the local backend still is not supported for this test, do we actually need to keep the previous rules here?

pierugo-dfinity (approved):

rs/tests/networking/firewall/firewall_correctness_test.rs
basvandijk: Good point. This was a remnant of trying to support the `firewall_correctness_test` on the local backend. I'm now working on this in a follow up / stacked PR but then I agree we should move this code there.

basvandijk: Done in 3d1e1e0c62 — the `previous_rules` read and the `TopologySnapshot::firewall_rules` accessor it needed are both gone, and the two files are byte-identical to master again. They will come back in the follow-up that enables the test locally.

One thing worth flagging from digging into that follow-up: narrowing the seeded rule at prep time does *not* work. Every node gets `nns_urls = http://[nns_node]:8080`, and a non-cloud-engine node never whitelists a `Type4` node on 8080, so that broad `fd00::/8` rule is what lets cloud-engine nodes sync the registry on the local backend at all. On Farm the config-file `default_rules` play that role until the test's own proposal lands, so the local equivalent has to tighten the rules at *test* time rather than in the initial registry. The `backend = "farm"` comment in BUILD.bazel understates that; I will correct it in the follow-up.

blind-oracle (approved):

rs/tests/networking/firewall/firewall_correctness_test.rs
basvandijk: Follow-up stacked on this: https://github.com/dfinity/ic/pull/11217

daniel-wong-dfinity-org-twin:
**UPDATE**: [This PR of mine that adds a new system test][new] has merged into master.

<strike>I think a [PR of mine that adds a system test][new] is going to go in first. If so, </strike>

[here] is another TODO for this PR, I think?

[new]: https://github.com/dfinity/ic/pull/10939

[here]: https://github.com/dfinity/ic/pull/10939/changes/3104786fac51eee6e61b617d3bcc9e247399fe27

rzakrzyk (approved):

PR #11209, opened 2026-08-19
fix: Do not charge pay-as-you-go base fee on system subnets

eichhorl:
System subnets have a normal cost schedule, but HTTP outcalls are still free.

This was already correctly enforced for the per-replica and consensus fees under the new pay-as-you-go pricing.

With this PR we also enforce a free base fee on system subnets.

Additionally, similar to (non-)replicated outcalls, we disable flexible outcalls on subnets where the `http_requests` feature is disabled.

### DIFF
--- a/rs/embedders/src/wasmtime_embedder/system_api.rs
+++ b/rs/embedders/src/wasmtime_embedder/system_api.rs
@@ -28,7 +28,7 @@ use ic_types::{
     messages::{CallContextId, MAX_INTER_CANISTER_PAYLOAD_IN_BYTES, RejectContext, SenderInfo},
     methods::{SystemMethod, WasmClosure},
 };
-use ic_types_cycles::Cycles;
+use ic_types_cycles::{CanisterCyclesCostSchedule, Cycles};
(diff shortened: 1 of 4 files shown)


pierugo-dfinity (approved):
LGTM. I wonder if it would help to add a helper (not sure where though, maybe as a method of `CanisterCyclesCostSchedule`?) that knowing a `SubnetType` and a `CanisterCyclesCostSchedule` would return the cost_schedule, but replaced by `Free` if on a system subnet?

That would avoid the logic duplication. Still not sure if there's a clean way to avoid the manual override inside `try_add_http_context_to_replicated_state` though.

eichhorl:
> I wonder if it would help to add a helper that knowing a `SubnetType` and a `CanisterCyclesCostSchedule` would return the cost_schedule, but replaced by `Free` if on a system subnet?

I sympathize with this because I also dislike the current behavior. However, if the implication "System subnet => Free cost schedule" were always true, then system subnets should just be on a free cost schedule. But since this isn't the case, I would asume that the implication does not always hold, and instead keep the decision local to where it does (HTTP outcalls).

PR #11210, opened 2026-08-19
refactor: resolve the DSM-11 canister log TODOs

mraszyk:
Two follow-ups left behind by the canister log memory store work:

- Move the hardcoded `LOG_RESIZE_COST_PER_BYTE` out of `canister_manager.rs` into `SchedulerConfig` as `canister_log_resize_instructions_per_byte` (default `DEFAULT_CANISTER_LOG_RESIZE_INSTRUCTIONS_PER_BYTE` = 32), threaded through `CanisterMgrConfig`. The TODO suggested `CyclesAccountManagerConfig`, but that config holds only `Cycles` denominated fees while this is an instruction cost per byte, so it belongs next to `upload_wasm_chunk_instructions` and the `canister_snapshot_*_instructions` knobs. The integration test no longer hardcodes the value either.

- Drop the aggregate flavour of `CanisterLog`, which the migration to `LogMemoryStore` (completed in #10821) made dead: `new_aggregate`, `default_aggregate`, `CanisterLog::append_delta_log`, `clear`, `retention` (superseded by `LogMemoryStore::retention`, already used by the metric) and the `Records` helpers that only served them. `Records::from` collapses into `Records::new(byte_capacity)` as every remaining constructor starts out empty. `CanisterLog` is now purely the per-message delta buffer, as its doc comment now states.

  The aggregate append tests are dropped, as the surviving behaviour is covered by the `LogMemoryStore::append_delta_log` tests; the rest are rewritten against the delta constructors, with new coverage for capacity capping and `take()`.


rs/execution_environment/src/canister_manager.rs
```diff
@@ -79,17 +79,6 @@ pub(crate) mod types;
...
 const MAX_SLICE_SIZE_BYTES: u64 = 2_000_000;
 
-/// Instructions charged per byte of stored log data during log memory resize.
-///
-/// When the log memory limit changes, all existing records must be read from
-/// the old ring buffer into heap memory, re-encoded, and written into a newly
-/// allocated ring buffer. The cost is proportional to the bytes currently
-/// stored (not the allocated capacity).
-///
-/// TODO(DSM-11): Consider moving this constant into `CyclesAccountManagerConfig`
```
mraszyk, line 89: `CyclesAccountManagerConfig` stores cycles fees, not instruction amounts so it is not a suitable place after all...

rs/config/src/subnet_config.rs
```diff
@@ -170,6 +170,16 @@ pub const DEFAULT_CANISTERS_SNAPSHOT_BASELINE_INSTRUCTIONS: NumInstructions =
 pub const DEFAULT_CANISTERS_SNAPSHOT_DATA_BASELINE_INSTRUCTIONS: NumInstructions =
     NumInstructions::new(5_000_000);
 
+/// Instructions charged per byte of stored log data when resizing the canister
+/// log memory.
+///
+/// When the log memory limit changes, all existing records must be read from
+/// the old ring buffer into heap memory, re-encoded, and written into a newly
+/// allocated ring buffer. The cost is proportional to the bytes currently
+/// stored (not the allocated capacity).
```
alin-at-dfinity, line 179: Idle question: Is it actually proportional to bytes stored or to bytes migrated? I.e. if I resize from 1 MB of actual log content to 4 KB, should I have to pay for the full 1 MB?

alin-at-dfinity (approved):

rs/config/src/subnet_config.rs
mraszyk: It must be proportional to the bytes stored because resizing is not optimized to only read the log records that will be migrated to the new (resized) log buffer. Instead, the current resizing implementation reads every log record and inserts it into the new (resized) log buffer.

PR #11214, opened 2026-08-19
fix: Update time of HTTP outcalls contexts when moving them to `delivered`

eichhorl:
HTTP outcalls contexts are moved to a different `delivered` collection, as soon as the response is delivered. There, they wait for further asynchronous refunds to be delivered. Previously, the context would be removed from this collection after a timeout of 2 minutes, since the creation of the _original_ context.

With this PR, we instead update the context's time when moving it to `delivered`, and remove it after a timeout of 1 minutes, since the insertion into `delivered` contexts. This reduces the amount of time that the context needs to stay in replicated state.

Additionally, we add metrics for the amount of cycles being refunded. We observe the total amount, and its fraction against the per-replica allowances.

### DIFF
--- a/rs/execution_environment/src/execution_environment.rs
+++ b/rs/execution_environment/src/execution_environment.rs
@@ -714,10 +714,12 @@ impl ExecutionEnvironment {
 
         let mut msg = match msg {
             SubnetMessage::Response(response) => {
-                let context = state
-                    .metadata
-                    .subnet_call_context_manager
-                    .retrieve_context(response.originator_reply_callback, &self.log);
+                let time = state.time();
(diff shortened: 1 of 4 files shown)


pierugo-dfinity (approved):
Unrelated: should we consider `delivered_canister_http_request_contexts` in `fn get_oldest_state_registry_version`?

eichhorl:
> Unrelated: should we consider `delivered_canister_http_request_contexts` in `fn get_oldest_state_registry_version`?

I think it's fine that leaving nodes will just refund in full if they haven't produced a response before leaving the subnet. The problem is more that this would also lock in all the nodes that did already respond to the context.

PR #11215, opened 2026-08-19
feat(migration): always export the oldest-request-age metric

mraszyk:
`migration_canister_oldest_request_in_flight_age_seconds` used to be omitted from the metrics output if no request was in flight, which makes the metric hard to use in queries and alerts. Export it unconditionally instead, reporting 0 if there is no pending request.

### DIFF
--- a/rs/migration_canister/src/canister_state.rs
+++ b/rs/migration_canister/src/canister_state.rs
@@ -133,10 +133,12 @@ pub mod requests {
     }
 
     /// Returns the age in nanos of the request that has been in flight the longest,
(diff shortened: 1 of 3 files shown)


alin-at-dfinity (approved):
LGTM

If you had a proper Prometheus metrics library, you could also use a histogram, where you could observe the ages of all requests. And a "no (oldest) request" scenario would manifest as no changes to the histogram. (A gauge has the issue that if it (essentially) flips on and off quickly, it may never be observed. That's not the case here, but a histogram would still be nicer, if it was available.)

Alas...

PR #11216, opened 2026-08-19
chore: Drop `NetworkTopology::full_topology`

alin-at-dfinity:
`NetworkTopology::full_topology` (only populated for the NNS subnet) holds the same subnets as `NetworkTopology::subnets` and the same routing table as `NetworkTopology::routing_table`. Drop it altogether, as it is unused by all subnets but the NNS; and the NNS behaves identically with or without it (including in the case of a replica downgrade to a version that still populates it).


.vscode/settings.json
basvandijk, line 1: Were the changes in this file meant to be added to this PR?

basvandijk (approved):

schneiderstefan (approved):
LGTM, but could you please do the RUN_ALL_BAZEL_TESTS thing, so that all upgrade downgrade tests have a chance to run?

.vscode/settings.json
alin-at-dfinity: I added some words to the ignore list of a spell checking VS Code extension. They're unrelated to the PR, but they're benign.

PR #11220, opened 2026-08-19
chore(dkg): remove summary's transcripts for remote subnets (1/4)

pierugo-dfinity:
Since the work performed to speedup the latency of remote DKG requests, their results are delivered as part of data blocks instead of summaries. The `transcripts_for_remote_subnets` field in summaries is thus now unused. This PR begins the migration to remove it entirely. Once it is, the batch delivery can be rewritten more cleanly as we would know that summary heights never contain any consensus responses.

We cannot simply replace the field with a `BackwardsCompatible<Vec<_>, _>` (i.e. replace the `Vec<_>` with an `Option<Vec<_>>`) because the field is a `repeated`, and an empty vector is not differentiable from an absent one. In contrast, `Some(vec![])` and `None` have different hashes. For that reason, we need to add a new sentinel field to the protobuf definition that will be set to `true` to explicitely indicate that the field is absent. The entire rollout can be reviewed by looking at the stacked PRs and is described below (together with justifying why upgrades & downgrades are safe).

- V0 (current) <-> V1 (after this PR):
    - The goal of V1 is to keep the exact same representation (including hash) of V0's summaries while still understanding the sentinel and compute the same hash as V2 for V2's summaries.
    - Upgrade V0 -> V1: A V0's summary doesn't set the sentinel so the field deserializes as `Some(vec![])` and the hash of an empty vector is appropriately computed.
    - Rollback V1 -> V0: a V1's summary doesn't set the sentinel just yet. V0 does not read it anyways and always includes an empty vector in the hash, same as V1.
- V1 <-> V2 ([#11221](https://github.com/dfinity/ic/pull/11221)):
    - The goal of V2 is to actually stop hashing an empty vector, and instead just ignore it.
    - Upgrade V1 -> V2: V2 still understands `Some(vec![])` so will compute V1's summary's hash identically. For its own summaries, it starts to fill the sentinel (because the transcripts are `None`).
    - Rollback V2 -> V1: V1 understands the sentinel and thus ignores the transcripts while computing the hash.
- V2 <-> V3 ([#11223](https://github.com/dfinity/ic/pull/11223)):
    - The goal of V3 is to remove the field entirely.
    - Upgrade V2 -> V3: V3 doesn't know about the transcripts anymore (the field is removed). V2 created a summary with `None` transcripts meaning their hash was not included, good.
    - Rollback V3 -> V2: V3 unconditionally sets the sentinel to `true`, meaning V2 will interpret the transcripts as `None` and not try to hash them.
 - V3 <-> V4 ([#11224](https://github.com/dfinity/ic/pull/11224)):
     - The goal of V4 is to remove the introduced sentinel field.
     - Upgrade V3 -> V4: V4 will ignore the sentinel field set by V3.
     - Rollback V4 -> V3: V3 will deserialize a summary with an unset sentinel, but it does not read it anyways.


rs/types/types/src/consensus/dkg.rs
```diff
@@ -268,7 +268,7 @@ pub struct DkgSummary {
     #[serde_as(as = "Vec<(_, _)>")]
     next_transcripts: BTreeMap<NiDkgTag, NiDkgTranscript>,
     /// Transcripts that are computed for remote subnets.
-    pub transcripts_for_remote_subnets: Vec<RemoteTranscriptResult>,
+    pub transcripts_for_remote_subnets: BackwardsCompatible<Vec<RemoteTranscriptResult>, true>,
     /// The length of the current interval in rounds (following the start
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/11095 merged since then, so this is obsolete

rs/consensus/src/consensus/batch_delivery.rs
```diff
@@ -331,10 +331,10 @@ fn generate_responses_to_subnet_calls(
...
                 summary_payload.dkg.configs.keys().collect::<Vec<_>>()
             );
-            consensus_responses.append(&mut generate_responses_to_remote_dkgs(
-                &summary_payload.dkg.transcripts_for_remote_subnets,
-                log,
-            ));
+            if let Some(transcripts) = summary_payload.dkg.transcripts_for_remote_subnets.as_ref() {
+                consensus_responses
+                    .append(&mut generate_responses_to_remote_dkgs(transcripts, log));
+            }
```
eichhorl, line 337: Can't we just remove this?

rs/protobuf/def/types/v1/dkg.proto
```diff
@@ -56,6 +56,14 @@ message Summary {
...
     PostSplitArgs post_split = 15;
   }
+  // Set by replica versions that no longer maintain `transcripts_for_remote_subnets` (field 10).
+  //
+  // When set, field 10 must be ignored entirely, including when hashing the summary. This is what
+  // allows the field to be removed without changing the hash of a summary: replica versions that
+  // still maintain the field and versions that have dropped it derive the same hash from the same
+  // wire bytes, because both read this marker. `repeated` fields cannot express the difference
+  // between "absent" and "empty" on the wire, hence the separate marker.
+  optional bool transcripts_for_remote_subnets_removed = 16;
```
eichhorl: Why does it need to be optional? It should be `false` by default
```suggestion
  bool transcripts_for_remote_subnets_removed = 16;
```

eichhorl (approved):

rs/protobuf/def/types/v1/dkg.proto
pierugo-dfinity: It doesn't need to, good catch thanks https://github.com/dfinity/ic/pull/11220/commits/4e41f85b52594759715b01a45a1f5230811b742c

rs/consensus/src/consensus/batch_delivery.rs
pierugo-dfinity: Just for correctness, I'd prefer to leave it like so (technically the protocol still understands that field despite now being optional). I'd remove it when the field truly isn't present anymore (https://github.com/dfinity/ic/pull/11223)

pierugo-dfinity: Hmm then following that argument I should also continue to return them in the `extract_remote_dkgs_from_highest_block` test method: https://github.com/dfinity/ic/pull/11220/commits/d2d2363b08ac371ebae2da377850f276598c120c

PR #11230, opened 2026-08-20
feat: Do not route subnet messages to a cooling down subnet

mraszyk:
Messages in the subnet's own output queues (only ever responses, as the
management canister makes no calls of its own) are no longer routed into
streams to a cooling down subnet, as long as this subnet is not cooling down
itself. They are retained in their output queue until the destination subnet
stops cooling down, rather than being rejected or dropped; and the skipped
queue is counted in `mr_cooling_down_skipped_queues`, exactly like a canister
output queue.

The subnet's own output queues stay exempt from the source side of the check,
so that a cooling down subnet can still respond to the calls it has already
accepted: while this subnet is cooling down, its own output responses are
routed regardless of whether the destination subnet is cooling down, the
loopback stream included.

Also renames `is_from_subnet_queues` to `is_subnet_output_response`.

### DIFF
--- a/rs/messaging/src/routing/stream_builder.rs
+++ b/rs/messaging/src/routing/stream_builder.rs
@@ -491,13 +491,14 @@ impl StreamBuilderImpl {
                     // subnet (the source) or the destination subnet is cooling down; not
                     // even into the loopback stream.
                     //
-                    // The subnet's own output responses are exempt from the source side of
-                    // the check, so that a cooling down subnet can still respond to the
-                    // calls it has already accepted. But they are held back just the same
-                    // while the destination subnet is cooling down.
+                    // The subnet's own output responses are exempt for as long as this
+                    // subnet is cooling down, so that it can still respond to the calls it
+                    // has already accepted; whether or not the destination subnet is cooling
+                    // down, the loopback stream included. They are only held back if this
+                    // subnet is not cooling down but the destination subnet is.
                     //
                     // Retain the message (along with everything behind it in the same queue)
-                    // until the subnet that is cooling down stops doing so, rather than
+                    // until the subnet that holds it back stops cooling down, rather than
                     // rejecting or dropping it.
                     let skip_while_cooling_down = if is_subnet_output_response {
                         !own_subnet_is_cooling_down && dst_subnet_is_cooling_down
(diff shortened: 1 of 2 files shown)


rs/messaging/src/routing/stream_builder.rs
```diff
@@ -474,7 +476,7 @@ impl StreamBuilderImpl {
             // Cheap to clone, `RequestOrResponse` wraps `Arcs`.
             let msg = msg.clone();
 
-            let is_from_subnet_queues = msg.sender() == own_subnet_as_canister_id;
+            let is_subnet_output_response = msg.sender() == own_subnet_as_canister_id;
```
alin-at-dfinity: Ultimate nitpicking: You could also say `msg.sender().get() == self.subnet_id.get()`. Then you wouldn't need an explicit `own_subnet_as_canister_id`.

And regardless, you should probably move `own_subnet_as_canister_id`'s comment here, as it's more relevant here.

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -483,15 +485,28 @@ impl StreamBuilderImpl {
...
+
+                    // No messages from canister output queues are routed while either this
+                    // subnet (the source) or the destination subnet is cooling down; not
+                    // even into the loopback stream.
+                    //
+                    // The subnet's own output responses are exempt for as long as this
+                    // subnet is cooling down, so that it can still respond to the calls it
+                    // has already accepted; whether or not the destination subnet is cooling
+                    // down, the loopback stream included. They are only held back if this
+                    // subnet is not cooling down but the destination subnet is.
```
alin-at-dfinity:
```suggestion
                    // Subnet output queues of cooling down subnets (only holding responses) are
                    // exempt, so they can deliver responses to all the calls they have already
                    // accepted.
```

You can optionally add "... deliver responses (before the subnet is deleted) to all ...".

alin-at-dfinity (approved):

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11230/commits/184f7c1ced44becb061913346c70d6ab5fc37a4e

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11230/commits/c5d1b5c4b2b9828e47fe6f816b51ffa9c2fe868b

PR #11233, opened 2026-08-20
chore: Add some HTTP outcalls metrics

eichhorl:
Some metrics and logs for HTTPS outcalls:
- Add a label to the metric counting delivered flexible errors
- Add error and informational metrics to the pool manager
- Add a detailed log whenever a timeout is included in a block
- Track whenever the payload reaches the responses per payload limit (500)

### DIFF
--- a/rs/https_outcalls/consensus/src/metrics.rs
+++ b/rs/https_outcalls/consensus/src/metrics.rs
@@ -21,7 +21,7 @@ pub struct CanisterHttpPoolManagerMetrics {
     /// A count of the total number of shares marked invalid.
     pub shares_marked_invalid: IntCounter,
     /// Notable, but expected events observed by the pool manager, by kind.
-    pool_manager_metrics: IntCounterVec,
+    pool_manager_events: IntCounterVec,
     /// Operations the pool manager failed to perform, by kind.
     pool_manager_errors: IntCounterVec,
 }
@@ -51,9 +51,9 @@ impl CanisterHttpPoolManagerMetrics {
             shares_marked_invalid: metrics_registry.int_counter(
                 "canister_http_shares_marked_invalid", "A count of the total number of shares marked invalid."
             ),
-            pool_manager_metrics: metrics_registry.int_counter_vec(
-                "canister_http_pool_manager_metrics",
-                "Canister http pool manager related metrics, by kind.",
+            pool_manager_events: metrics_registry.int_counter_vec(
+                "canister_http_pool_manager_events",
+                "Notable, but expected events observed by the pool manager, by kind.",
                 &[LABEL_TYPE],
             ),
             pool_manager_errors: metrics_registry.int_counter_vec(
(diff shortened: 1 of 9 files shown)


pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/metrics.rs
```diff
@@ -43,9 +50,29 @@ impl CanisterHttpPoolManagerMetrics {
             ),
             shares_marked_invalid: metrics_registry.int_counter(
                 "canister_http_shares_marked_invalid", "A count of the total number of shares marked invalid."
-            )
+            ),
+            pool_manager_metrics: metrics_registry.int_counter_vec(
+                "canister_http_pool_manager_metrics",
+                "Canister http pool manager related metrics, by kind.",
+                &[LABEL_TYPE],
+            ),
```
alin-at-dfinity: Nit: The doc comment on the field ("Notable, but expected events observed by the pool manager, by kind") actually suggests a better name:


```suggestion
            pool_manager_events: metrics_registry.int_counter_vec(
                "canister_http_pool_manager_events",
                "Notable, but expected events observed by the pool manager, by kind.",
                &[LABEL_TYPE],
            ),
```

As is, "metrics" is exceedingly vague. Everything is a metric, around here and in Prometheus.

rs/https_outcalls/consensus/src/metrics.rs
```diff
@@ -43,9 +50,29 @@ impl CanisterHttpPoolManagerMetrics {
...
+            pool_manager_errors: metrics_registry.int_counter_vec(
+                "canister_http_pool_manager_errors",
+                "Canister http pool manager related errors, by kind.",
+                &[LABEL_TYPE],
+            ),
         }
     }
+
+    /// Records a notable, but expected event of the given kind.
+    pub(crate) fn pool_manager_metrics_inc(&self, label: &str) {
```
alin-at-dfinity: `observe_pool_manager_event`?

rs/https_outcalls/consensus/src/metrics.rs
```diff
@@ -43,9 +50,29 @@ impl CanisterHttpPoolManagerMetrics {
...
+        self.pool_manager_metrics.with_label_values(&[label]).inc();
+    }
+
+    /// Records a failed operation of the given kind.
+    pub(crate) fn pool_manager_errors_inc(&self, label: &str) {
```
alin-at-dfinity: `observe_pool_manager_error`?

You could also merge the two `IntCounterVecs` into one by adding an `error="true|false"` label. Not sure whether that makes sense in this specific context, but it's an option.

alin-at-dfinity (approved):

rs/https_outcalls/consensus/src/metrics.rs
eichhorl: Good point, done: https://github.com/dfinity/ic/pull/11233/commits/abcccb7d7f0b36bc3aa5e0c61d182584dafda1ec

rs/https_outcalls/consensus/src/metrics.rs
eichhorl: Done 👍

rs/https_outcalls/consensus/src/metrics.rs
eichhorl: Done: https://github.com/dfinity/ic/pull/11233/commits/cdc1354a31711f1f169ec5fbd636ece032a1f89b, I kept both metric separate, which is more consistent with what other protocols do

PR #11237, opened 2026-08-20
feat(cketh): drive the sweeper transaction pipeline from its own timer task

gregorydemay:
## Why

- The sweeper address has its own transaction pipeline, but nothing moves requests through it.
- This is the task that does, on the sweeper's own nonce sequence.

## What

- `sweep::process_sweeper_transactions` runs the sweeper pipeline through the same create → sign → send → resubmit → finalize cycle as withdrawals.
- It signs with the sweeper derivation path `[3]` and reads the sweeper address' own transaction count.
- It takes its own `TaskType::SweeperSend` timer guard, so a slow sweep round neither blocks the withdrawal task nor is blocked by it — the point of a separate nonce sequence.
- A sweep whose prepaid gas cannot cover the current fee is logged and moved back to the end of the queue, as the withdrawal task does, rather than priced at something that could not be mined.
- The transaction type stays the request's business: a sweep that must still install delegations is sent as EIP-7702, any other as EIP-1559.

## Prerequisite commit

- Four RPC helpers in `withdraw` were implicitly about the main address, calling `minter_address()` themselves or keying receipts by `LedgerBurnIndex`.
- Reading a transaction count, broadcasting signed transactions and fetching finalized receipts are specific to neither a sender nor what a request is keyed by, so the sender becomes an argument and the receipt fetch becomes generic over the pipeline id.
- Behaviour is unchanged: the receipt loop moves into `fetch_finalized_receipts`, which returns `None` where the loop used to return early.

## Scope

- Nothing enqueues a `SweepRequest` yet, so the task early-returns on an empty pipeline and is inert in production.
- Still to come: the sweep-queue→`SweepRequest` source, and gating on prepaid sweep gas.

## Note for reviewers

- The sweep driver and the withdrawal driver are now near-identical five-stage loops over different request types, down to a private trio of batch-size constants that are `5` in both files.
- Deduplicating them is the obvious follow-up now that the pipeline is generic, but it would bury this change, so it is left for its own PR.

### DIFF
--- a/rs/ethereum/cketh/minter/src/main.rs
+++ b/rs/ethereum/cketh/minter/src/main.rs
@@ -47,7 +47,8 @@ use ic_cketh_minter::withdraw::{
 };
 use ic_cketh_minter::{
     BALANCE_SCAN_INTERVAL, PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, PROCESS_REIMBURSEMENT,
-    REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL, SCRAPING_ETH_LOGS_INTERVAL, state, storage,
+    PROCESS_SWEEPER_TRANSACTIONS_INTERVAL, REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL,
+    SCRAPING_ETH_LOGS_INTERVAL, state, storage,
 };
 use ic_cketh_minter::{endpoints, erc20};
 use ic_ethereum_types::Address;
@@ -100,9 +101,7 @@ fn setup_timers() {
     ic_cdk_timers::set_timer_interval(PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, async || {
         process_retrieve_eth_requests().await;
     });
-    // Drains the sweeper send-pipeline on the same cadence as withdrawals; early-returns cheaply while
-    // no sweep is enqueued (nothing enqueues one in production yet — see `crate::sweep`).
-    ic_cdk_timers::set_timer_interval(PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, async || {
+    ic_cdk_timers::set_timer_interval(PROCESS_SWEEPER_TRANSACTIONS_INTERVAL, async || {
         process_sweeper_transactions().await;
     });
     ic_cdk_timers::set_timer_interval(PROCESS_REIMBURSEMENT, async || {
--- a/rs/ethereum/cketh/minter/src/main.rs
+++ b/rs/ethereum/cketh/minter/src/main.rs
@@ -38,6 +38,7 @@ use ic_cketh_minter::state::transactions::{
 use ic_cketh_minter::state::{
     STATE, State, lazy_call_ecdsa_public_key, mutate_state, read_state, transactions,
 };
+use ic_cketh_minter::sweep::process_sweeper_transactions;
 use ic_cketh_minter::timed_sized_map::Timestamp;
 use ic_cketh_minter::tx::lazy_refresh_gas_fee_estimate;
 use ic_cketh_minter::withdraw::{
@@ -98,6 +99,11 @@ fn setup_timers() {
     });
     ic_cdk_timers::set_timer_interval(PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, async || {
         process_retrieve_eth_requests().await;
+    });
+    // Drains the sweeper send-pipeline on the same cadence as withdrawals; early-returns cheaply while
+    // no sweep is enqueued (nothing enqueues one in production yet — see `crate::sweep`).
+    ic_cdk_timers::set_timer_interval(PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, async || {
+        process_sweeper_transactions().await;
     });
     ic_cdk_timers::set_timer_interval(PROCESS_REIMBURSEMENT, async || {
         process_reimbursement().await;
--- a/rs/ethereum/cketh/minter/src/sweep.rs
+++ b/rs/ethereum/cketh/minter/src/sweep.rs
@@ -1,0 +1,257 @@
+//! Sending transactions **from the minter's dedicated sweeper address** on the sweeper's own nonce
+//! sequence, so a stuck sweep can never head-of-line-block a withdrawal on the main address'
+//! pipeline.
+//!
+//! This is the preparatory send-pipeline: it drives the sweeper [`TransactionPipeline`] through the same
+//! create → sign → send → resubmit → finalize state machine as user withdrawals
+//! ([`crate::withdraw`]), reusing that module's sender-agnostic RPC helpers
+//! (`latest_transaction_count`, `finalized_transaction_count`, `send_signed_transactions`,
+//! `fetch_finalized_receipts`), but signing with the sweeper derivation path (`[3]`) and reading
+//! the sweeper address' own transaction count.
+//!
+//! Deliberately out of scope here (follow-ups): EIP-7702 (`0x04`) first-time delegation, building
+//! the delegate sweep call data from the balance-sweep queue, and gating the pipeline on prepaid sweep
+//! gas. Nothing enqueues a [`SweepRequest`] in production yet, so this task early-returns on an
+//! empty pipeline.
+
+use crate::{
+    deposit_address::sweeper_derivation_path,
+    guard::TimerGuard,
+    logs::{DEBUG, INFO},
+    numeric::{GasAmount, TransactionCount},
+    state::{
+        State, TaskType,
+        audit::{EventType, process_event},
+        mutate_state, read_state,
+        transactions::{CreateSweepTransactionError, PipelineRequest},
+    },
+    tx::{GasFeeEstimate, lazy_refresh_gas_fee_estimate},
+    withdraw::{
+        fetch_finalized_receipts, finalized_transaction_count, latest_transaction_count,
+        send_signed_transactions,
+    },
+};
+use futures::future::join_all;
+use ic_canister_log::log;
+use ic_ethereum_types::Address;
+use ic_management_canister_types_private::DerivationPath;
+
+/// Gas limit of a sweep transaction. A fixed, conservative bound for the preparatory type-`0x02`
+/// pipeline; a per-request estimate arrives with the real delegate sweep call.
+pub(crate) const SWEEP_TRANSACTION_GAS_LIMIT: GasAmount = GasAmount::new(100_000);
+
+const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
+const SWEEP_TRANSACTIONS_TO_SIGN_BATCH_SIZE: usize = 5;
+const SWEEP_TRANSACTIONS_TO_SEND_BATCH_SIZE: usize = 5;
+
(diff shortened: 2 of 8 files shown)


rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -98,6 +99,11 @@ fn setup_timers() {
     ic_cdk_timers::set_timer_interval(PROCESS_ETH_RETRIEVE_TRANSACTIONS_INTERVAL, async || {
         process_retrieve_eth_requests().await;
     });
+    // Drains the sweeper send-pipeline on the same cadence as withdrawals; early-returns cheaply while
+    // no sweep is enqueued (nothing enqueues one in production yet — see `crate::sweep`).
```
gregorydemay: why should it use the same cadence as withdrawals? I would define another constant, even if it had the same value

rs/ethereum/cketh/minter/src/withdraw.rs
```diff
@@ -191,9 +194,11 @@ pub async fn process_retrieve_eth_requests() {
     }
 }
 
-async fn latest_transaction_count() -> Option<TransactionCount> {
+/// The latest (unconfirmed) transaction count of `sender` on chain, used to gate resubmission and
+/// sending. Reused by both the main-address pipeline and the sweeper-address pipeline (`sender` differs).
```
gregorydemay:
```suggestion
/// The latest (unconfirmed) transaction count of `sender` on chain, used to gate resubmission and
/// sending.
```

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -0,0 +1,257 @@
...
+    send_transactions_batch(latest_transaction_count).await;
+    finalize_transactions_batch(sender).await;
+
+    if read_state(|s| s.sweeper_transactions.has_pending_requests()) {
+        ic_cdk_timers::set_timer(
+            crate::PROCESS_ETH_RETRIEVE_TRANSACTIONS_RETRY_INTERVAL,
+            async { process_sweeper_transactions().await },
+        );
+    }
+}
```
gregorydemay, line 92: That looks a lot like `process_retrieve_eth_requests`, can't we reuse it by making it more generic or with parameters?

Similar comment for the other methods in this file.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -0,0 +1,257 @@
...
+//! create → sign → send → resubmit → finalize state machine as user withdrawals
+//! ([`crate::withdraw`]), reusing that module's sender-agnostic RPC helpers
+//! (`latest_transaction_count`, `finalized_transaction_count`, `send_signed_transactions`,
+//! `fetch_finalized_receipts`), but signing with the sweeper derivation path (`[3]`) and reading
+//! the sweeper address' own transaction count.
+//!
+//! Deliberately out of scope here (follow-ups): EIP-7702 (`0x04`) first-time delegation, building
+//! the delegate sweep call data from the balance-sweep queue, and gating the pipeline on prepaid sweep
+//! gas. Nothing enqueues a [`SweepRequest`] in production yet, so this task early-returns on an
+//! empty pipeline.
```
gregorydemay: This belongs to the PR description, not Rust doc
```suggestion
```

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Agreed, done in 67330555d6: `PROCESS_SWEEPER_TRANSACTIONS_INTERVAL` and `PROCESS_SWEEPER_TRANSACTIONS_RETRY_INTERVAL`, both equal to the withdrawal values for now.

The retry interval had the same problem one level down — `sweep.rs` was reaching for `PROCESS_ETH_RETRIEVE_TRANSACTIONS_RETRY_INTERVAL` — so it moved to the sweeper constant too. The comment claiming the cadences are shared is gone with them.

rs/ethereum/cketh/minter/src/withdraw.rs
gregorydemay: 🤖 Applied in 70b0b0b1be. The `sender` parameter already says the helper serves both pipelines, so the sentence only restated the signature.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Removed in 70b0b0b1be — it is in the PR description, which is where it belongs.

I also dropped "preparatory" from the sentence above it: it was true when this PR sat below the EIP-7702 one, and reads wrongly now that the order is reversed and a sweep can already be type `0x04`.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 You are right that they are near-identical: guard, `has_pending_requests`, gas estimate, five stages, retry timer, differing only in the guard type, which pipeline field is read, how the sender is resolved, the derivation path and the gas limit. Those differences factor cleanly into a per-lane descriptor.

What does not factor as cleanly is the event vocabulary — `CreatedTransaction` against `CreatedSweeperTransaction`, and the same for signed, replaced and finalized — plus the reimbursement work that only the withdrawal lane does when finalizing. Sharing the driver means a trait whose associated functions name each lane's five events, which is a change of its own size.

So: doing it, but in its own refactoring PR rather than here. Beyond keeping this PR readable, nine PRs currently sit on this branch, and refactoring at this layer forces a restack of all of them; on top of the stack it costs nothing.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -0,0 +1,244 @@
...
+    let gas_fee_estimate = match lazy_refresh_gas_fee_estimate().await {
+        Some(gas_fee_estimate) => gas_fee_estimate,
+        None => {
+            log!(
+                INFO,
+                "[process_sweeper_transactions]: failed retrieving gas fee estimate",
+            );
+            return;
+        }
+    };
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed and fixed in 05ef67e93c. Tracing it: `lazy_refresh_gas_fee_estimate` serves a cached estimate only while it is under 60s old, and both send tasks run on a 6-minute cadence, so the estimate is always stale when they fire. The withdrawal task is registered first, enters `do_refresh` and holds the `RefreshGasFeeEstimate` guard across the fee-history call; this task then finds the estimate still stale and the guard held, gets `None`, and returned before reaching the retry timer at the end. Next interval, same order — and only while withdrawals are pending, since the withdrawal task early-returns otherwise, exactly as you describe.

The `None` path now schedules the retry, which also resolves the collision sooner than the next interval: by the time it fires, the refresh it lost to has cached an estimate this task can simply reuse.

No production impact in this PR, since nothing enqueues a sweep yet, but it would have bitten as soon as something did.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Gas limit of a sweep transaction. A fixed, conservative bound for the preparatory type-`0x02`
+/// pipeline; a per-request estimate arrives with the real delegate sweep call.
+pub(crate) const SWEEP_TRANSACTION_GAS_LIMIT: GasAmount = GasAmount::new(100_000);
+
+const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
+const SWEEP_TRANSACTIONS_TO_SIGN_BATCH_SIZE: usize = 5;
+const SWEEP_TRANSACTIONS_TO_SEND_BATCH_SIZE: usize = 5;
+
+pub async fn process_sweeper_transactions() {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 True of this PR, and deliberate rather than an oversight.

Nothing can enqueue a `SweepRequest` at this layer, so a test here would have to inject one by mutating production state from a test-only hook, which this repo's conventions rule out.

The coverage you describe arrives with the end-to-end test four PRs up (#11261): it drives `process_sweeper_transactions` against a real chain in anvil and asserts the sweeper address paid the gas, the transaction type and its on-chain success, the delegation designators installed, and the created/signed/finalized sweeper audit events — including the interaction between the two timer tasks, since both run throughout.

blind-oracle (approved):

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -0,0 +1,260 @@
...
+        send_signed_transactions,
+    },
+};
+use futures::future::join_all;
+use ic_canister_log::log;
+use ic_ethereum_types::Address;
+use ic_management_canister_types_private::DerivationPath;
+
+/// Gas limit of a sweep transaction. A fixed, conservative bound for the preparatory type-`0x02`
+/// pipeline; a per-request estimate arrives with the real delegate sweep call.
```
mbjorkqvist: An [earlier commit](https://github.com/dfinity/ic/pull/11237/commits/9447f277bf0c1b59a30ecc77a85470890824a284) dropped "preparatory" from the module doc - should it be dropped here also?

rs/ethereum/cketh/minter/src/withdraw.rs
```diff
@@ -359,7 +364,15 @@ async fn send_transactions_batch(latest_transaction_count: Option<TransactionCou
         s.withdrawal_transactions
             .transactions_to_send_batch(latest_transaction_count, TRANSACTIONS_TO_SEND_BATCH_SIZE)
     });
+    send_signed_transactions(&transactions_to_send).await;
+}
 
+/// Broadcast already-signed transactions via the EVM RPC canister. Sender- and
+/// transaction-type-agnostic, so both the main-address withdrawal pipeline (type `0x02`) and the
+/// sweeper-address pipeline (type `0x02` or `0x04`) reuse it.
```
mbjorkqvist, line 378: The log message here, and for `latest_transaction_cost`, is the same for both pipelines (main-address withdrawal and sweeper-address) - would it make sense to add a lane label or the sender to the message to be able to differentiate them?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Good catch, I missed it — dropped in 3f8f452277. It was stale twice over: the pipeline is not preparatory any more, and since the EIP-7702 PR now sits below this one, a sweep can already be type `0x04`, so the "type-`0x02` pipeline" framing was wrong too. The forward reference to a per-request estimate stays, since that arrives with the enqueue PR.

rs/ethereum/cketh/minter/src/withdraw.rs
gregorydemay: 🤖 Agreed, and I went with the sender rather than a lane label: the sender *is* the lane, so it needs no new vocabulary and it says which address the RPC was about. In 3f8f452277.

`latest_transaction_count` already had it, so its message just interpolates it. `send_signed_transactions` did not — a signed EIP-1559 transaction carries no `from` — so it now takes the sender, threaded from the two `send_transactions_batch` callers that already hold it, and all three of its messages name it.

I left each lane's own logs alone: the sweeper task's carry a `[process_sweeper_transactions]` prefix already, so the ambiguity was confined to the shared helpers.

PR #11238, opened 2026-08-20
feat(subnet-splitting): make destination orchestrators restart the replica

pierugo-dfinity:
This PR implements changes to the orchestrator to support subnet splitting.

Today, the orchestrator tries to determine the latest CUP by looking at its local CUP's subnet ID. It contacts peers determined by this subnet's membership at the latest registry version and looks for a potential recovery (or genesis for a new subnet) CUP for this subnet, also at the latest registry version.

The proposed change are the following:
- The orchestrator still looks at its local CUP subnet ID.
- Before contacting peers or looking at the registry, first look inside the registry for a potential split that was scheduled *between the local CUP's registry version and the latest version*. If there is one, manually overwrite the subnet ID with the new expected one (source or destination).
- If the subnet ID of our local CUP is different from the subnet ID of the latest CUP, restart the replica.

Note that, unlike @kpop-dfinity's [original design](https://github.com/dfinity/ic/compare/pierugo/kpop/final_ss_untouched), the proposed implementation does not change the existing behaviour during membership changes that are independent of a subnet split. In his approach, orchestrators would fetch & verify CUPs from the subnet ID they were part of *at the latest registry version*. Though simpler, this could negatively affect situations where nodes get moved from one subnet to the other without a split, since they would leave the subnet before being able to finish their work. It would also make recoveries harder.

## Happy path
- All orchestrators will detect the split as soon as the registry changes are in, meaning that already before the split actually triggers, source orchestrators will start fetching CUPs from the source subset of nodes and destination orchestrators from the complement.
- Each of them also verify the received CUP according to the post-split membership's public key. This means that until the split is actually triggered, all destination orchestrators will **reject** received CUPs, since those served CUPs are still the original subnet's signed by the original public key. This can be a problem if there was an upgrade scheduled shortly before (see *An upgrade was scheduled shortly before a split but we're unlucky*).
- When the split actually triggers
    - Source orchestrators fetch the post-split CUP from source peers, validate it with the new source's public key and do not do much more. Their subnet ID doesn't change and the replicas continue to run.
    - Destination orchestrators fetch the post-spilt CUP from destination peers, validate it with the destination's public key and detect that their subnet ID changed. They restart their replicas.

Note: if an upgrade was scheduled shortly before or after a split, it should still "just work" (except if we are unlucky, see later below). If it was scheduled after, the subnet anyways does not bump the registry version past the scheduled split, so the upgrade will simply be triggered on a following summary. If it was scheduled shortly before in such a way that its registry version ends up in the `Scheduled` (and thus also `PostSplit`) summary, the orchestrators will execute the upgrade right at the same time as the split. It shouldn't lead to misbehaviour.

### An upgrade was scheduled shortly before a split but we're unlucky
As explained previously, destination orchestrators will not pick up CUPs created between the moment the registry schedules the split and the split actually triggers:
- If the split is scheduled when the subnet is at height `H=498`, replicas create a summary at height `H=500` *without* `Scheduled` status because the registry version is not stable, then create a CUP. Then, destination orchestrators **will not** adopt this CUP because the signature doesn't verify. If this CUP indicates an upgrade (scheduled during the previous interval), half of the subnet (intended for source) will upgrade while the other half (intended for destination) will stay stuck. We'd have to split the state manually and recover both halves individually, just like in the next scenario.

### Subnet stalls between the registry scheduled the split but before the subnet actually triggers it
Any recovery CUP would pin the latest registry version, which contains the post-split topology. This mean that if such a scenario were to happen, we would have to split the state manually and upload them to the respective halves. All orchestrators are looking for potential recovery CUPs according to the post-split membership already as soon as the registry changes are observed, so the two subnets can be recovered individually.
- If the orchestrators only looked at the subnet ID of their local CUP, the original subnet would have to be recovered "as a whole" but a registry CUP pins the latest registry version. Thus destination orchestrators would adopt it while also determining that they are not part of the subnet anymore and would delete their state. They would then join the destination subnet with an empty state with height 0, just like a newly created subnet. Though they would be responsible for a bunch of live canisters of the original subnet. It would be a giant mess.

### One half manages to create a post-split CUP and make progress while the other half stays stuck
Be it source or destination, orchestrators will always fetch peer/registry CUPs according to their post-split membership and will not be confused by the other half's "normal" CUPs that will be created past the split. For the same reason, it is also possible to recover this stuck half without affecting the other one.

### DIFF
--- a/rs/consensus/cup_utils/src/lib.rs
+++ b/rs/consensus/cup_utils/src/lib.rs
@@ -7,7 +7,9 @@ use ic_consensus_idkg::{
 };
 use ic_interfaces_registry::RegistryClient;
 use ic_logger::{ReplicaLogger, warn};
-use ic_protobuf::registry::subnet::v1::CatchUpPackageContents;
+use ic_protobuf::registry::subnet::v1::{
+    CatchUpPackageContents, catch_up_package_contents::CupType,
+};
 use ic_registry_client_helpers::subnet::SubnetRegistry;
 use ic_types::{
     Height, RegistryVersion, SubnetId, Time,
@@ -24,8 +26,10 @@ use ic_types::{
 };
 use phantom_newtype::Id;
 
-/// Constructs a genesis/recovery CUP from the CUP contents associated with the
-/// given subnet from the provided CUP contents
+/// Constructs a genesis/recovery CUP from the CUP contents associated with the given subnet from
+/// the provided CUP contents.
+/// Registry CUPs intended for subnet splitting are explicitely excluded here as they are used for a
+/// different purpose, directly by Consensus
 pub fn make_registry_cup_from_cup_contents(
     registry: &dyn RegistryClient,
     subnet_id: SubnetId,
@@ -33,6 +37,15 @@ pub fn make_registry_cup_from_cup_contents(
     registry_version: RegistryVersion,
     logger: &ReplicaLogger,
 ) -> Option<CatchUpPackage> {
+    // If the CUP we are about to build is a subnet splitting CUP, return early. It makes no sense
+    // to build a registry CUP out of subnet splitting CUP contents because the transcripts here are
+    // used directly by consensus to build the CUP themselves, i.e. nodes threshold-sign it, instead
+    // of blindly taking it from the registry here.
+    match cup_contents.cup_type {
+        Some(CupType::SubnetSplitting(..)) => return None,
+        Some(CupType::Genesis(..)) | Some(CupType::Recovery(..)) | None => {}
+    }
+
     let replica_version = match registry.get_replica_version(subnet_id, registry_version) {
         Ok(Some(replica_version)) => replica_version,
         err => {
--- a/rs/consensus/cup_utils/src/lib.rs
+++ b/rs/consensus/cup_utils/src/lib.rs
@@ -407,7 +407,7 @@ mod tests {
     }
 
     #[test]
-    fn test_make_registry_cup_with_subnet_splitting_cup_contents() {
+    fn test_make_registry_cup_with_subnet_splitting_cup_type() {
         let registry_client = setup_registry_with_cup_type(
             /*registry_store_uri=*/ None,
             CupType::SubnetSplitting(Default::default()),
--- a/rs/consensus/cup_utils/src/lib.rs
+++ b/rs/consensus/cup_utils/src/lib.rs
@@ -28,7 +28,7 @@ use phantom_newtype::Id;
 
 /// Constructs a genesis/recovery CUP from the CUP contents associated with the given subnet from
 /// the provided CUP contents.
-/// Registry CUPs intended for subnet splitting are explicitely excluded here as they are used for a
+/// Registry CUPs intended for subnet splitting are explicitly excluded here as they are used for a
 /// different purpose, directly by Consensus
 pub fn make_registry_cup_from_cup_contents(
     registry: &dyn RegistryClient,
--- a/rs/consensus/cup_utils/src/lib.rs
+++ b/rs/consensus/cup_utils/src/lib.rs
@@ -278,8 +278,11 @@ mod tests {
     const LATEST_REGISTRY_VERSION: RegistryVersion = RegistryVersion::new(12345);
 
     /// Builds a registry client serving the CUP contents and the subnet record which
-    /// [`make_registry_cup`] needs, with the given `registry_store_uri` in the CUP contents.
-    fn setup_registry(registry_store_uri: Option<RegistryStoreUri>) -> impl RegistryClient {
+    /// [`make_registry_cup`] needs, with the given `registry_store_uri` and `cup_type`.
+    fn setup_registry_with_cup_type(
+        registry_store_uri: Option<RegistryStoreUri>,
+        cup_type: CupType,
+    ) -> impl RegistryClient {
         MockRegistryClient::new(LATEST_REGISTRY_VERSION, move |key, _| {
             use prost::Message;
             if key.starts_with("catch_up_package_contents_") {
@@ -299,11 +302,7 @@ mod tests {
                         registry_store_uri: registry_store_uri.clone(),
                         ecdsa_initializations: vec![],
                         chain_key_initializations: vec![],
-                        cup_type: Some(CupType::Recovery(RecoveryArgs {
-                            height: 54321,
-                            time: 1,
-                            state_hash: vec![1, 2, 3, 4, 5],
-                        })),
+                        cup_type: Some(cup_type.clone()),
                     };
 
                 // Encode the cup to protobuf
@@ -328,6 +327,17 @@ mod tests {
                 None
             }
         })
+    }
+
+    fn setup_registry(registry_store_uri: Option<RegistryStoreUri>) -> impl RegistryClient {
+        setup_registry_with_cup_type(
+            registry_store_uri,
+            CupType::Recovery(RecoveryArgs {
+                height: 54321,
+                time: 1,
+                state_hash: vec![1, 2, 3, 4, 5],
+            }),
+        )
     }
 
     #[test]
@@ -394,6 +404,26 @@ mod tests {
             );
             assert_eq!(cup.content.registry_version(), expected_registry_version);
         }
+    }
+
+    #[test]
+    fn test_make_registry_cup_with_subnet_splitting_cup_contents() {
+        let registry_client = setup_registry_with_cup_type(
+            /*registry_store_uri=*/ None,
+            CupType::SubnetSplitting(Default::default()),
+        );
+
+        let result = make_registry_cup(
+            &registry_client,
+            subnet_test_id(0),
+            registry_client.get_latest_version(),
+            &no_op_logger(),
+        );
+
+        assert!(
+            result.is_none(),
+            "Expected None for subnet splitting CUP contents"
+        );
     }
 
     /// `RegistryClient` implementation that allows to provide a custom function
--- a/rs/orchestrator/src/catch_up_package_provider.rs
+++ b/rs/orchestrator/src/catch_up_package_provider.rs
@@ -459,7 +459,10 @@ impl CatchUpPackageProvider {
                 self.node_id,
                 split_in_progress_reg_ver
             );
-            // Default to rejecting the CUP and try again later
+            // If we cannot read the registry or we are actually unassigned (which should not
+            // happen), default to rejecting the CUP and try again later
+            // If the subnet is healthy, it will at some point in the future create a `NotScheduled`
+            // CUP and make us truly aware of what we are assigned to.
             return true;
         };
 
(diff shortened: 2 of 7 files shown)


rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -463,13 +549,20 @@ impl CatchUpPackageProvider {
...
+            .get_peer_cup(subnet_id, latest_registry_version, local_cup.as_ref())
             .await;
 
         let registry_cup = self
             .registry
-            .get_registry_cup(registry_version, subnet_id)
+            .get_registry_cup(latest_registry_version, subnet_id)
+            .inspect_err(|err| warn!(self.logger, "Failed to create a registry cup: {err}"))
             .map(pb::CatchUpPackage::from)
             .ok();
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/11238/commits/0142882362286892f28dba99d68b6bf7afec6d29

rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -390,6 +393,89 @@ impl CatchUpPackageProvider {
...
+        let Some(local_cup) = local_cup else {
+            return Ok(());
+        };
+        let Ok(local_cup) = CatchUpPackage::try_from(local_cup) else {
+            return Ok(());
+        };
+
+        let from_version = latest_registry_version.get();
+        let to_version = local_cup.content.registry_version().get() + 1;
+        for registry_version in (to_version..=from_version).rev() {
```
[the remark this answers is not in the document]
pierugo-dfinity: Sure https://github.com/dfinity/ic/pull/11238/commits/6f3759e724202471ec5cdaab6c7524fe04f69e10

rs/orchestrator/src/upgrade.rs
```diff
@@ -267,13 +277,40 @@ impl Upgrade {
...
+        // In the vast majority of cases, they will be identical. In case of a subnet split, this is
+        // not necessarily the case, and we want the rest of this function to consider the new
+        // subnet ID.
+        subnet_id = get_subnet_id(&self.registry, &latest_cup).map_err(|err| {
+            OrchestratorError::UpgradeError(format!(
+                "Couldn't determine the subnet id of the latest CUP: {err:?}"
+            ))
+        })?;
+
+        if subnet_id != old_subnet_id {
```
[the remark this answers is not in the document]
pierugo-dfinity: This is true, and this also applies when observing a new recovery CUP, whose flow already exists today.
The fallible errors are for the moment only registry-related and very unlikely to trigger. I would accept the risk for now and consider this issue in a follow-up: maybe we could move that responsibility to the process manager.

rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -390,6 +393,96 @@ impl CatchUpPackageProvider {
...
+                // Any other assignment than source or destination is unexpected. This is very
+                // similar to the previous condition about being unassigned. This could happen if we
+                // were moved to a different subnet prior to the split, independently of the latter.
+                // But there could also very well be a previous split of the same subnet: we would
+                // not want to remove the state by mistake, continue scanning.
+                continue;
+            }
+
+            *subnet_id = new_subnet_id;
+            return Ok(());
```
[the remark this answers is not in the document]
pierugo-dfinity: Your suggestion is much more complex while the current implementation should still work. By contacting B's peers/registry for B, we have two cases:
- B hasn't yet split into B & C. We are thus still part B's membership in its latest CUP and we join B, as expected.
- B has already split into B & C. We are thus **not** part of B's membership in its latest CUP. We delete our state and later join C the same way an unassigned node joins a subnet. This is suboptimal but not a drastic issue because it is not possible for all nodes to be in this situation by hypothesis that "B has already split into B & C".

rs/orchestrator/src/upgrade.rs
```diff
@@ -2864,6 +2961,738 @@ mod tests {
...
+            // which will be caught exactly at the same CUP as the post-split one.
+            // (Upgrading the subnet at the same version as the split would not be realistic, but
+            // doing it beforehand is totally valid)
+            add_subnet_record_to_provider(
+                &data_provider,
+                SPLIT_REGISTRY_VERSION - RegistryVersion::from(1),
+                new_subnet_id,
+                SubnetType::Application,
+                [node_id, other_node_id],
+                replica_version,
```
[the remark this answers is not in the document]
pierugo-dfinity: Correct, thanks https://github.com/dfinity/ic/pull/11238/commits/4365003c7ac42be3c8b1bf2a54e1eb669713de1a

rs/orchestrator/src/upgrade.rs
```diff
@@ -2864,6 +2961,738 @@ mod tests {
...
+    #[case::moves_to_destination(DESTINATION_SUBNET_ID)]
+    #[tokio::test]
+    async fn test_subnet_split_scenarios(#[case] new_subnet_id: SubnetId) {
+        let node_id = NODE_1;
+        let other_node_id = NODE_2;
+        let current_replica_version = ReplicaVersion::try_from("replica_version_0.1").unwrap();
+        let local_cup_height = Height::from(100);
+        let post_split_cup_height = Height::from(200);
+        // Only the destination subnet's replicas need to be restarted, as only their subnet public
+        // key changed.
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/11238/commits/c084bd25f9709483f6d13230c4bd503b4181a486

rs/consensus/cup_utils/src/lib.rs
```diff
@@ -24,15 +26,26 @@ use ic_types::{
...
     registry_version: RegistryVersion,
     logger: &ReplicaLogger,
 ) -> Option<CatchUpPackage> {
+    // If the CUP we are about to build is a subnet splitting CUP, return early. It makes no sense
+    // to build a registry CUP out of subnet splitting CUP contents because the transcripts here are
+    // used directly by consensus to build the CUP themselves, i.e. nodes threshold-sign it, instead
+    // of blindly taking it from the registry here.
+    match cup_contents.cup_type {
+        Some(CupType::SubnetSplitting(..)) => return None,
+        Some(CupType::Genesis(..)) | Some(CupType::Recovery(..)) | None => {}
```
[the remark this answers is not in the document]
pierugo-dfinity: https://github.com/dfinity/ic/pull/11238/commits/91bfa701ff213d8dcbb94a0731a88f3fad4bb50c

rs/orchestrator/src/upgrade.rs
```diff
@@ -267,13 +277,40 @@ impl Upgrade {
...
             .await?;
+        // Replace the subnet ID of the local CUP with the subnet ID of the latest CUP.
+        // In the vast majority of cases, they will be identical. In case of a subnet split, this is
+        // not necessarily the case, and we want the rest of this function to consider the new
+        // subnet ID.
+        subnet_id = get_subnet_id(&self.registry, &latest_cup).map_err(|err| {
+            OrchestratorError::UpgradeError(format!(
+                "Couldn't determine the subnet id of the latest CUP: {err:?}"
+            ))
+        })?;
```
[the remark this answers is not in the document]
pierugo-dfinity: Looks unrelated

rs/orchestrator/src/catch_up_package_provider.rs
eichhorl, line 239: Could it happen that our local registry is a bit behind, such that we still fetch the post-split CUP from the source subnet (even though we should have fetched it from the destination subnet according to the latest registry version), and then validate it? Then later we would notice that we aren't actually assigned to A and delete the state.

rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -383,6 +386,96 @@ impl CatchUpPackageProvider {
...
+            };
+
+            if new_subnet_id != *subnet_id && new_subnet_id != destination_subnet_id {
+                // Any other assignment than source or destination is unexpected. This is very
+                // similar to the previous condition about being unassigned. This could happen if we
+                // were moved to a different subnet prior to the split, independently of the latter.
+                // But there could also very well be a previous split of the same subnet: we would
+                // not want to remove the state by mistake, continue scanning.
+                continue;
+            }
```
eichhorl, line 494: Not sure I understand why we are handling these cases differently

rs/orchestrator/src/catch_up_package_provider.rs
pierugo-dfinity: Yes, something like this could indeed happen, I missed that, thanks. We would have to be particularly unlucky, since it would require the `RegistryClientImpl` to poll precisely between the moment we read the latest registry version and the moment we verify the CUP (but considering we are doing some network I/O between them, this actually isn't so unreasonable), otherwise crypto would fail with `VersionNotAvailable`.

Also, this is not a problem for correctness because in that situation, there is a post-split CUP, meaning that 2f+1 replicas have seen the registry change prior to observing the CUP and thus their orchestrators too. What I mean is that this issue cannot happen on more than f nodes, and thus the overall subnet state is still preserved.

Will think about how we could avoid this without complexifying too much if possible.

rs/orchestrator/src/catch_up_package_provider.rs
pierugo-dfinity: My main concern was that, if A splits into A & B, and later into A & C, I wouldn't want nodes intended to go to B to be shadowed by the second split. This justifies the `continue;`.
The `return Ok(());` if we're unassigned was for me more of an optimization. It wouldn't hurt to continue scanning for a potential shadowed split, but we anyways know that we would get unassigned later so might as well return only.

Ok, writing this and taking a step back, I think it actually makes sense to replace it with `continue;`. If there *was* a previous split, it would be best to contact the subnet we were supposed to be part of to learn *from that subnet* that we're getting unassigned. It's also more consistent with the second `if`.

pierugo-dfinity: >  learn from that subnet that we're getting unassigned

It's actually necessary: the subnet could still need us for some work and we should turn unassigned only when we get a CUP *from that subnet* telling us we're free to go.

pierugo-dfinity: Replaced `return Ok(());` with `continue;` [here](https://github.com/dfinity/ic/pull/11238/commits/49dd97e07478a0cab2e8034b8f009a60f6959ff3)

rs/orchestrator/src/catch_up_package_provider.rs
pierugo-dfinity: > without complexifying too much

Looks like it can be mitigated with a 1-line change. WDYT? https://github.com/dfinity/ic/pull/11238/commits/4518415a5d2ae499f4ebc49393dd8fd0255c5643

rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -383,6 +396,99 @@ impl CatchUpPackageProvider {
...
+            let versioned_record = self
+                .registry
+                .get_registry_client()
+                .get_cup_contents(*subnet_id, lookup_version)
+                .map_err(OrchestratorError::RegistryClientError)?;
+
+            // The record was last written at or before the local CUP's registry version: any
+            // recorded split here or below is a past one.
+            if versioned_record.version <= local_cup_registry_version {
+                break;
```
eichhorl: Might be more readable to use either `break` or `return` consistently in the loop
```suggestion
                return Ok(());
```

rs/orchestrator/src/catch_up_package_provider.rs
eichhorl: LGTM, should we add any logs or metrics for these cases?

eichhorl (approved):

rs/orchestrator/src/catch_up_package_provider.rs
pierugo-dfinity: Yep https://github.com/dfinity/ic/pull/11238/commits/f0f42ad9d78203c3dda34ddcc7fc427d565be5e1

rs/orchestrator/src/catch_up_package_provider.rs
pierugo-dfinity: I agree, good idea https://github.com/dfinity/ic/pull/11238/commits/975659bb2e80a1c42d96dbbe3b206eb0239793c9

PR #11240, opened 2026-08-20
feat(cketh): expose the sweeper address in MinterInfo

gregorydemay:
The minter derives a dedicated sweeper address from its master threshold-ECDSA public key, and that address' ETH balance is the prepaid gas available for sweeping funded deposit addresses. Until now that address was not observable from outside the canister, so there was no way to check how much sweep gas was left or to top it up.

`get_minter_info` now returns it as `sweeper_address`, alongside the existing `minter_address`. Like `minter_address`, it is `null` until the minter has fetched its public key.


mbjorkqvist (approved):
LGTM, thanks @gregorydemay! Would it make sense to expose the sweeper address in the minter dashboard also (not necessarily as part of this PR)?

gregorydemay:
> LGTM, thanks @gregorydemay! Would it make sense to expose the sweeper address in the minter dashboard also (not necessarily as part of this PR)?

Good idea, I added it as a bullet to DEFI-2922.

PR #11242, opened 2026-08-20
feat(registry): ensure no version deployment is in progress when splitting a cloud engine + that the record was not changed

pierugo-dfinity:
Even though splitting cloud engines are not supported for now, this PR makes the registry additionally check that the `StandardEngineReplicaVersionRecord` hasn't changed during the time taken to generate fresh key material for the two new subnets during a split.
This is to avoid weird interactions where an upgrade would trigger at the same time as a subnet split.

It also adds unit tests for the other records that were previously checked.

Also, ensure that no version is in deployment when splitting a cloud engine, as the newly created engine could otherwise end up in a different replica version as the original engine.


rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -384,6 +385,12 @@ impl Registry {
             return Err("Subnet changed");
         }
 
+        // The replica version that the newly created subnet is going to run may be derived from
+        // this record, so it must not change under our feet either.
+        if record_changed_across_versions(make_standard_engine_replica_version_record_key()) {
```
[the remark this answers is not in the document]
pierugo-dfinity: Cloud engines are indeed not supported for the moment, but they could be in the future. This PR preemptively adds this check in case we forget to do so when enabling the feature for cloud engines.

It is true that it could abort a subnet split that was intended for regular subnets even though this record only affects cloud engines, but I would keep it as is for simplicity since this has quite a low probability of happening anyways (the async call should take about 10 seconds).

Something else came to mind though: https://github.com/dfinity/ic/pull/11242/commits/531cc436f247db03e1774d5ab5313764ccde8424. Cloud engines shouldn't be split if the deployment of a new replica version is in progress as the new subnet could end up in a new version.

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -275,6 +277,27 @@ impl Registry {
...
+            // We want to reject a split if the source subnet is a cloud engine and a deployment of
+            // a new replica version is in progress. This is because the new destination subnet will
+            // have a new subnet ID which we do not know yet and could end up upgrading to a
+            // different version. We would like to avoid weird situations where both a split and an
+            // upgrade are scheduled at the same time, so we enforce that the cloud engine
+            // deployment is complete before allowing a split. In that case, it is guaranteed that
+            // both subnets will be on the same replica version after the split.
+            if 0.0 < standard_engine_replica_version_record.deployment_progress
+                && standard_engine_replica_version_record.deployment_progress < 1.0
+            {
```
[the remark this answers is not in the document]
pierugo-dfinity: The probability of hitting a priority precisely equal to 0.0 is negligible (`1/2^64`), and I could imagine that a priority 0.0 could genuinely happen in practice to tell all cloud engines to stay on `old_replica_version_id`. So I think this is fine, but I leave the final decision to the code owners.

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -275,6 +277,27 @@ impl Registry {
...
+
+            // We want to reject a split if the source subnet is a cloud engine and a deployment of
+            // a new replica version is in progress. This is because the new destination subnet will
+            // have a new subnet ID which we do not know yet and could end up upgrading to a
+            // different version. We would like to avoid weird situations where both a split and an
+            // upgrade are scheduled at the same time, so we enforce that the cloud engine
+            // deployment is complete before allowing a split. In that case, it is guaranteed that
+            // both subnets will be on the same replica version after the split.
+            if 0.0 < standard_engine_replica_version_record.deployment_progress
+                && standard_engine_replica_version_record.deployment_progress < 1.0
```
daniel-wong-dfinity-org-twin: I would say that 0.0 indicates that a deployment is in progress. Only when it reaches 1.0 would I say it's complete. But even then, changing from 1.0 to < 1.0 is (currently) allowed... But maybe, we can disallow roll back once it reaches 1.0? That would be fine, because you can still roll back by doing

```
{
  new: previous
  old: current
  progress: ...
}
```

if current is already at 100%.

Anyway, this is moot if we go with my other suggestion.

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -275,6 +277,27 @@ impl Registry {
...
+                self.get_standard_engine_replica_version_record()
+                .expect("StandardEngineReplicaVersionRecord should exist if the subnet has no replica version");
+
+            // We want to reject a split if the source subnet is a cloud engine and a deployment of
+            // a new replica version is in progress. This is because the new destination subnet will
+            // have a new subnet ID which we do not know yet and could end up upgrading to a
+            // different version. We would like to avoid weird situations where both a split and an
+            // upgrade are scheduled at the same time, so we enforce that the cloud engine
+            // deployment is complete before allowing a split. In that case, it is guaranteed that
+            // both subnets will be on the same replica version after the split.
```
daniel-wong-dfinity-org-twin, line 294: This seems verbose.

Basically, we want upgrading a subnet and splitting it to be mutually exclusive, right?

In that case, what I'd suggest is that we just disallow splitting engines that have blank replica_version_id. Seems simpler, right?

What I like about my suggestion is that splitting one engine does NOT block upgrading all the others. That seems like a feature we really want to preserve.

Furthermore, what we can do here (to more conveniently fulfill "engines with blank replica_version_id cannot be split") is fill in the replica_version_id field, fix the field to whatever StandardEngineReplicaVersion is currently assigning the engine to. But this is just a convenience; you could instead force people to do a deploy_guestos to fix the replica_version_id manually.

rs/registry/canister/src/mutations/do_split_subnet.rs
daniel-wong-dfinity-org-twin, line 539: I have not reviewed changes past this line. Let me resolve this once I've done that.

rs/registry/canister/unreleased_changelog.md
```diff
@@ -12,6 +12,13 @@ on the process that this file is part of, see
 * Invariant requiring that SEV-enabled subnets may only run a GuestOS version that has
   `guest_launch_measurements`.
 
+* A subnet-split request will now fail if a concurrent call modified the `StandardEngineReplicaVersionRecord`
+  while the fresh key material was being generated for the splitting subnet.
```
daniel-wong-dfinity-org-twin, line 18: Does this only apply to engines that have blank replica_version_id?

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: Changing from 1.0 to < 1.0 is not a problem because these would end up in different registry versions. We ensure in Consensus that while the subnet is getting split, it only reads at the registry version at which the registry is applying those mutations, not before, not later.

The problem is that even for a fixed registry version, two engines reading the same record can derive a different replica version. Only 0.0 or 1.0 ensure that both engines will derive the same version, though I have to say that since this is a continuous spectrum, only 1.0 is strictly valid since one of the engines could have a priority of exactly 0.0. I just wanted to relax the check 0.0 is expected to be used somewhat frequently, but I'm happy to change it to 1.0 only.

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: > Seems simpler, right?

Simpler in code but definitely not simpler operationally, since you'd first need to set `replica_version_id` to the engine, then unset the field on both created engines -> 3 additional proposals + potential human error in the loop.

> What I like about my suggestion is that splitting one engine does NOT block upgrading all the others.

The current implementation satisfies that too. It's the other way around: upgrading the others will block splitting an engine (which is acceptable).

> fix the field to whatever StandardEngineReplicaVersion is currently assigning the engine to

Hmm, yes that could work, but you'd still need 2 proposals to unset them. That would make it asymmetric + it isn't natural to make an operation responsible for splitting a subnet also mutate that subnet's replica version.

rs/registry/canister/unreleased_changelog.md
pierugo-dfinity: No this applies to all subnets including engines setting their `replica_version_id` and regular subnets. This is suboptimal, but I kept it as is for simplicity.
See discussion [here](https://github.com/dfinity/ic/pull/11242#discussion_r3828438980) and feel free to suggest relaxing that check if you think it is worth the added complexity.

rs/registry/canister/src/mutations/do_split_subnet.rs
daniel-wong-dfinity-org-twin: > two engines reading the same record can derive a different replica version

ah. I think this is the key.

> one of the engines could have a priority of exactly 0.0

Ah. this is an interesting secondary point. Maybe we should treat 0 progress specially: even an engine with 0 upgrade priority should get the old version. That would be an easy fix. I think for this PR, let's assume either that no engine ever lands on exactly 0.0, or we this rule ☝️ is in place. In that case 0% and 100% should both be treated as safe, because in both cases, the new "daughter" engine WILL match the original "mother" engine of the split. In that case, I think no action is needed here. If you agree, feel free to close.

rs/registry/canister/src/mutations/do_split_subnet.rs
daniel-wong-dfinity-org-twin: > 3 additional proposals

No, I was not suggesting such mutations be done via proposal. Isn't there something that orchestrates splitting? That thing could order those mutations without proposals.

> The current implementation satisfies that too.

Ah. I didn't realize you do not mind changing the standard engine Replica version during a split. What you do not want is to split when not all engines (that follow the standard Replica version) are on the same version. In that case, I drop this suggestion.

rs/registry/canister/src/mutations/do_split_subnet.rs
daniel-wong-dfinity-org-twin: You might want to do something like

```
let fleet_is_split = ...
if fleet_is_split {
```

I think [multi-line if expressions should be avoided][s&c]. I mean, in this case, you are just trying to say some value is between two values (or equivalently, not 0 or 1), but Rust (and many other languages) force you to repeat yourself.

[s&c]: https://github.com/dfinity/ic/blob/master/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L225

daniel-wong-dfinity-org-twin: Ah. Claude suggested something to me:

```
let fleet_is_unified = deployment_progress.fract() == 0.0;
```

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -803,6 +840,69 @@ mod tests {
         assert_eq!(validation_result, expected_result);
     }
 
+    /// Every record that `check_if_registry_changed_across_versions` guards, and the error that
+    /// it is expected to produce when it changes while the `setup_initial_dkg` calls are in
+    /// flight.
+    #[rstest]
```
daniel-wong-dfinity-org-twin, line 847: Make no change on account of this thread, because I see that rstest was already used locally. This is just so that my opinion is in the record.

I don't like this. `#[test]` is used in the OVERWHELMING majority of cases. Also, attributes are a bit magical, so the bar for "earns its keep" should be much higher. Also, the way it is being used here is not really special. Basically, you have a list of input output pairs. A for loop can do that:

```
#[test]
fn test_...() {
    for (changed_record_key, expected_error) in [
        (a, b),
        ...
    ] {
        ...
    }
}
```

This does have one small downside: more indentation. That's less bad than using a not well known library that makes heavy use of attributes.

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -803,6 +840,69 @@ mod tests {
...
+
+        assert_eq!(
+            registry.check_if_registry_changed_across_versions(
+                source_subnet_id,
+                pre_call_registry_version,
+                registry.latest_version(),
+            ),
+            Err(expected_error)
+        );
+    }
```
daniel-wong-dfinity-org-twin, line 908: It took me a while to grok what this test is trying to verify. IIUC, what we want to see is that check_if_registry_changed_across_versions returns an Err when the subnet is changed. Yes?

There are a couple things that could be done to make this clearer.

1. Mention the name of the thing under test in the name of the test. Currently, the name just says that SOME thing is supposed to fail. Maybe, let's mention it.

2. [Delineate] the three standard test phases with comments:
    1. Prepare the world: lines 866 through 894
    2. Run the code under test: line 897
    3. Verify results: lines 896 through 903, excluding 897

[Delineate]: https://github.com/dfinity/ic/blob/5a0c6ee3850557aa380f7887c9d2ac113f16f68e/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L365

Do as you see fit (including nothing at all).

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/src/mutations/do_split_subnet.rs
```diff
@@ -33,6 +34,7 @@ enum PayloadValidationError {
         pre_split_source_subnet_size: usize,
     },
     DisallowedSourceSubnetType(SubnetType),
+    CloudEngineDeploymentInProgress,
```
daniel-wong-dfinity-org-twin:
```suggestion
    CloudEngineFleetIsSplit,
```

Why: To me, 0 indicates that a deployment is in progress, but you consider this an acceptable value. Furthermore, 1 does not mean that a deployment is really complete, because, we can still roll back.

What you really care about is whether some engines (the follow the standard Replica version) are on one side of the split while others are on the other.

You can't look at a snapshot and KNOW whether something is in motion, which is what the current name is talking about. But you do not actually care about that. All you care about is the current POSITION: are we in the middle of a river crossing, or are we on one of the two banks? That CAN be determined by looking at just one snapshot.

pierugo-dfinity: Yep that makes sense 👍 https://github.com/dfinity/ic/pull/11242/commits/5705ee5c783394dd1364853c803623a21aceae14

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/11242/commits/898ba3e6c7e5979d0f60910d778b891b426ff749

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: (I find `rstest` tests much easier to read)

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: > Isn't there something that orchestrates splitting? That thing could order those mutations without proposals.

There is a single proposal `do_split_subnet` which is the one we are calling here, and there is a single mutation. The rest is driven by the subnet's Consensus on its own.

rs/registry/canister/src/mutations/do_split_subnet.rs
pierugo-dfinity: > I think no action is needed here. If you agree, feel free to close.

Yep, good

> let fleet_is_unified = deployment_progress.fract() == 0.0;

This felt a bit cryptic, so I kept the original condition but avoided the multi-line `if`: https://github.com/dfinity/ic/pull/11242/commits/00feae550e2ab95c257839fa52b2b67fe559795f

PR #11244, opened 2026-08-20
fix(registry-helper): get_replica_version_record supports blank replica_version_id.

daniel-wong-dfinity-org-twin:
Blank replica_version_id only allowed when the subject of the request is an engine, not general subnet.

This is per StandardEngineReplicaVersionRecord.

This should have been done earlier, when we modified get_replica_version in #10840. Björn commented in a later PR that [we forgot to do this][fulfills].

[fulfills]: https://github.com/dfinity/ic/pull/10888#issuecomment-5094374132

### DIFF
--- a/rs/registry/helpers/src/subnet.rs
+++ b/rs/registry/helpers/src/subnet.rs
@@ -515,17 +515,10 @@ impl<T: RegistryClient + ?Sized> SubnetRegistry for T {
         subnet_id: SubnetId,
         version: RegistryVersion,
     ) -> RegistryClientResult<ReplicaVersionRecord> {
-        let bytes = self.get_value(&make_subnet_record_key(subnet_id), version);
-        Ok(match deserialize_registry_value::<SubnetRecord>(bytes)? {
-            Some(record) => {
-                let bytes = self.get_value(
-                    &make_replica_version_key(record.replica_version_id),
-                    version,
-                );
-                deserialize_registry_value::<ReplicaVersionRecord>(bytes)?
-            }
-            None => None,
-        })
+        let Some(replica_version) = self.get_replica_version(subnet_id, version)? else {
+            return Ok(None);
+        };
+        self.get_replica_version_record_from_version_id(&replica_version, version)
     }
 
     fn get_replica_version_record_from_version_id(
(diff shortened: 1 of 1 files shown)


rs/registry/helpers/src/subnet.rs
pierugo-dfinity, line 625: Unrelated: why is this a free function and not a method of `RegistryClient`?

pierugo-dfinity (approved):

rs/registry/helpers/src/subnet.rs
daniel-wong-dfinity-org-twin: Hmm. Yeah. Seems weirdly inconsistent. I cannot remember there being a rationale for this. It could just be that Claude generated this, and I didn't notice, and therefore, simply didn't object.

claude-of-daniel-wong-dfinity-org: It's a private, module-internal helper — not part of the public `SubnetRegistry` trait. It's used exactly once, inside `get_replica_version`'s own blank-resolution logic (line 484), so it's really just an implementation detail of that method rather than something external callers need directly. Making it a trait method would add public surface area for no benefit.

daniel-wong-dfinity-org-twin: Pierugo, if you hate this, we can change it in another PR.

bjoernek:
Late follow-up from re-reading this (nothing blocking, it's already merged): the doc comment on the trait method is now stale.

`rs/registry/helpers/src/subnet.rs:193`:

```rust
    /// Return the [ReplicaVersionRecord] as recorded in the subnet record
    /// at the given height.
    fn get_replica_version_record(
```

"as recorded in the subnet record" is precisely the behavior this PR removes: for a CloudEngine subnet with a blank `replica_version_id`, the record now comes from `StandardEngineReplicaVersionRecord`. The doc also doesn't mention that the method can now return `Err(DecodeError)` in the three cases enumerated on `get_replica_version`, which it could not before.

Suggested replacement:

```rust
    /// Returns the [ReplicaVersionRecord] that the subnet is running at the
    /// given registry version.
    ///
    /// The replica version ID is resolved via
    /// [SubnetRegistry::get_replica_version], so a CloudEngine subnet with a
    /// blank `replica_version_id` resolves through
    /// [StandardEngineReplicaVersionRecord] rather than being read out of the
    /// [SubnetRecord] directly. See [SubnetRegistry::get_replica_version] for
    /// the full resolution rules, and for the cases in which DecodeError is
    /// returned.
    ///
    /// Ok(None) is returned when the subnet has no [SubnetRecord], or when the
    /// resolved replica version has no [ReplicaVersionRecord].
```

(This also drops the "at the given height" wording, which was already inaccurate: the parameter is a `RegistryVersion`, not a height.)

Not worth a PR on its own, happy for it to ride along with whatever touches this file next.

rs/registry/helpers/src/subnet.rs
pierugo-dfinity: I think it will probably come up useful in the future, so the risk is that someone would add such a method to `RegistryClient` without realizing there already was this free function, and we'd have duplicate implementation.

I'd be in favour of turning it into a method now, but we can also wait until someone actually needs it.

@daniel-wong-dfinity-org-twin

PR #11245, opened 2026-08-20
feat: CON-1700 Support flexible outcalls and pay-as-you-go pricing in `pocket-ic`

eichhorl:
## Replica Changes
The option to choose pay-as-you-go pricing for non- and fully-replicated outcalls is now gated behind the same feature flag that is already gating flexible outcalls (which use pay-as-you-go pricing by default).

### Added
- Add a function and the type  to mock the responses of the committee nodes of a pending *flexible* canister HTTP outcall.
- Add the `replication` field on `CanisterHttpRequest` describing how a pending canister HTTP outcall is replicated across the nodes of its subnet: `FullyReplicated`, `NonReplicated`, or `Flexible` with the outcall's `total_requests`, `min_responses`, and `max_responses`.
- Add the `pricing_version` field  on `CanisterHttpRequest`, reporting whether a pending canister HTTP outcall is priced with the `Legacy` or the `PayAsYouGo` pricing model.
- Enabling beta features (`IcpConfig::beta_features` passed to `PocketIcBuilder::with_icp_config`) now also enables the pay-as-you-go pricing model, which covers both the `flexible_http_request` management canister endpoint (whose outcalls are always priced that way) and the `pricing_version` field of `http_request` (through which a fully replicated or non-replicated outcall can select it).

### Changed
- Mocked canister HTTP responses report the cycles their node actually spent on the outcall, instead of reporting no spend at all.
- Mocking a canister HTTP response whose reject message exceeds 1 KiB now fails. Such a response is not one any node could have reported.
- A node that cannot pay for gossiping its mocked reject reports an out-of-cycles reject instead of the mocked one, matching what a node of a real subnet does under pay-as-you-go pricing.


packages/pocket-ic/HOWTO.md
```diff
@@ -378,6 +378,106 @@ e.g., 13 for a regular application subnet.
...
+*Note.* A mocked reject message must fit into the 1 KiB a node truncates its reject messages to; mocking a
+longer one fails, since no node could have reported it. This applies to both `PocketIc::mock_canister_http_response`
+and `PocketIc::mock_flexible_canister_http_response`.
+
+Note that a flexible outcall never rejects the calling canister's call: every outcome above, including the
+errors, is delivered as a `flexible_http_request_result` reply. Only a synchronous failure (invalid arguments,
+insufficient cycles attached, or the endpoint not being enabled on the subnet) rejects the call.
+
+*Warning.* The cycles a node reports having spent on a mocked outcall include a term for how long the
+response took to arrive, which PocketIC derives from how long it took to run the mocked outcall in process.
```
mraszyk: > how long it took to run the mocked outcall in process

could this be made deterministic, or even better let user mock the spent cycles (also see below)

packages/pocket-ic/test_canister/canister.did
```diff
@@ -123,6 +127,8 @@ service : {
   node_metrics_history_proxy: (NodeMetricsHistoryArgs) -> (vec NodeMetricsHistoryResponse);
   canister_http : (text) -> (HttpResponseResult);
   canister_http_with_transform : (text) -> (HttpResponse);
+  canister_http_raw : (blob, nat) -> (RawHttpResponseResult);
+  flexible_canister_http : (blob, nat) -> (RawHttpResponseResult);
```
mraszyk: let's generalize to proxying an arbitrary call to an arbitrary callee then

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1174,6 +1174,47 @@ pub struct CanisterHttpHeader {
     pub value: String,
 }
 
+/// How a canister HTTP outcall is replicated across the nodes of its subnet.
+#[derive(
+    Clone, Serialize, Deserialize, Debug, Hash, Eq, PartialEq, Ord, PartialOrd, JsonSchema,
+)]
+pub enum CanisterHttpReplication {
+    /// Every node of the subnet performs the outcall and a response is delivered
+    /// once `n - f` of them agree on it. Too many differing responses make the
```
mraszyk, line 1185: > once `n - f` of them agree on it

add a doc comment to RawMockCanisterHttpResponse relating this to `response` and `additional_responses`

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1174,6 +1174,47 @@ pub struct CanisterHttpHeader {
...
+    /// outcall fail with a "no consensus could be reached" rejection instead.
+    /// Mock such an outcall with `PocketIc::mock_canister_http_response`.
+    FullyReplicated,
+    /// A single node performs the outcall (`is_replicated = false`) and its
+    /// response is the one delivered.
```
mraszyk: what if somebody provides `additional_responses` in this case; or uses `MockFlexible...`?

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1299,6 +1349,61 @@ impl From<MockCanisterHttpResponse> for RawMockCanisterHttpResponse {
...
+    pub subnet_id: RawSubnetId,
+    pub request_id: u64,
+    pub responses: Vec<CanisterHttpResponse>,
+}
+
+/// Mocked responses to a pending *flexible* canister HTTP outcall, i.e. one made
+/// through the `flexible_http_request` management canister endpoint.
+///
+/// Unlike a fully replicated outcall, which delivers the one response its nodes
+/// agree on, a flexible outcall is performed by a committee of `total_requests`
```
mraszyk: the corresponding mock type still supports multiple responses to trigger the "divergence" error

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1299,6 +1349,61 @@ impl From<MockCanisterHttpResponse> for RawMockCanisterHttpResponse {
...
+pub struct MockFlexibleCanisterHttpResponse {
+    pub subnet_id: Principal,
+    pub request_id: u64,
+    /// One response per committee node that responded. Each response is attributed
+    /// to a different committee node.
+    ///
+    /// There may be at most `total_requests` responses; providing fewer models a
+    /// committee whose remaining nodes never respond. With at least `min_responses`
+    /// successful ones among them the outcall still succeeds; with fewer it stays
+    /// pending until the time is advanced past its 60 second timeout.
```
mraszyk: we should also spell out the case of many rejects that results in a response (reject), not a timeout

rs/execution_environment/src/execution_environment/tests.rs
```diff
@@ -3936,6 +3936,111 @@ fn execute_canister_http_request_caps_allowance_at_worst_case_cost() {
...
+            "unexpected pricing version with the feature flag \
+             {flexible_http_requests_enabled}"
+        );
+    }
+}
+
+/// An unknown pricing version falls back to the default one, whether or not the
+/// pay-as-you-go pricing model is enabled.
+#[test]
+fn execute_canister_http_request_unknown_pricing_version_falls_back() {
```
mraszyk: I guess we could merge this test with the previous one.

rs/https_outcalls/client/src/client.rs
```diff
@@ -165,30 +164,6 @@ impl NonBlockingChannel<CanisterHttpRequest> for CanisterHttpAdapterClientImpl {
             let max_response_size_bytes = request_max_response_bytes
                 .map_or(MAX_CANISTER_HTTP_RESPONSE_BYTES, |bytes| bytes.get());
 
-            if request_pricing_version == ic_types::canister_http::PricingVersion::PayAsYouGo {
```
mraszyk, line 168: how could any (system) test work if pay-as-you-go was disabled here?

rs/pocket_ic_server/CHANGELOG.md
```diff
@@ -11,6 +11,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+- The endpoint `/instances/<instance_id>/update/mock_flexible_canister_http` to mock the responses of the committee nodes
+  of a pending *flexible* canister HTTP outcall, i.e. one made through the `flexible_http_request` management canister endpoint.
+- The endpoint `/instances/<instance_id>/read/get_canister_http` reports two additional fields for every pending canister HTTP
+  outcall: `replication`, describing how the outcall is replicated across the nodes of its subnet (`FullyReplicated`,
+  `NonReplicated`, or `Flexible` with the outcall's `total_requests`, `min_responses`, and `max_responses`), and
+  `pricing_version`, reporting whether it is priced with the `Legacy` or the `PayAsYouGo` pricing model.
+- Enabling beta features now also enables the pay-as-you-go pricing model, which covers both the `flexible_http_request` management
+  canister endpoint (whose outcalls are always priced that way) and the `pricing_version` field of `http_request` (through which a fully replicated or non-replicated outcall can select it).
```
mraszyk: the flexible outcalls themselves are also enabled by beta features, right? otherwise it sounds like only the pricing model changes for flexible outcalls due to beta features

rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -3981,22 +4030,125 @@ fn process_mock_canister_https_response(
...
+            // out-of-cycles reject instead of the one it produced, just like the
+            // HTTPS outcalls client does.
+            let reject =
+                match budget.subtract_gossip_usage(NumBytes::from(reject.count_bytes() as u64)) {
+                    Ok(()) => reject,
+                    Err(PricingError::InsufficientCycles) => CanisterHttpReject {
+                        reject_code: ic_error_types::RejectCode::CanisterReject,
+                        message: "Insufficient cycles".to_string(),
+                    },
+                };
```
mraszyk, line 4077: wouldn't it make more sense to let the user specify how much was spent by the node by including a corresponding field in every "mocked" response (also in the reply case above)?

rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -4009,13 +4161,75 @@ fn process_mock_canister_https_response(
...
-        return OpOut::Error(PocketIcError::InvalidMockCanisterHttpResponses((
-            contents.len(),
-            subnet.nodes.len(),
+    // Every node of the subnet answers, which is the contract of this
+    // (non-flexible) mock. That is exactly what a fully replicated outcall needs,
+    // since its committee is the whole node set. For a non-replicated or flexible
+    // outcall the payload builder ignores the shares of the nodes outside the
+    // designated node / the committee, so the surplus responses are harmless (see
+    // `mock_flexible_canister_http_response` for mocking a flexible outcall
+    // committee by committee).
```
mraszyk: let's remark this also in the public docs - at a high level, namely that the legacy "mocking" can be used for non-replicated calls, too

rs/pocket_ic_server/src/pocket_ic.rs
```diff
@@ -4009,13 +4161,75 @@ fn process_mock_canister_https_response(
         vec![content.clone(); subnet.nodes.len() - 1]
     };
     contents.push(content);
-    if contents.len() != subnet.nodes.len() {
```
mraszyk, line 4012: why drop this?, at least we should do this: if num_responses > committee.len() {
        return OpOut::Error(PocketIcError::TooManyMockCanisterHttpResponses

rs/state_machine_tests/src/lib.rs
```diff
@@ -2765,13 +2765,47 @@ impl StateMachine {
             .push(msg, self.get_time(), self.nodes[0].node_id);
     }
 
+    /// Injects one response share per node of the subnet, all reporting that the
+    /// node spent nothing on the outcall.
+    ///
+    /// For an outcall performed by only a subset of the nodes, or one whose nodes
+    /// report having spent cycles, see [`Self::mock_canister_http_response_for_nodes`].
     pub fn mock_canister_http_response(
```
mraszyk: is this function still used?

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        )
+        .unwrap();
+
+    // We need a pair of ticks for the test canister method to make the http outcall
+    // and for the management canister to start processing the http outcall.
+    pic.tick();
+    pic.tick();
+    let mut canister_http_requests = pic.get_canister_http();
+    assert_eq!(canister_http_requests.len(), 1);
+    (call_id, canister_http_requests.pop().unwrap())
```
mraszyk: this could be deduplicated with the code above

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+    Decode!(&bytes, FlexibleHttpRequestResult).unwrap()
+}
+
+/// Submits a non-flexible HTTP outcall through the test canister that asks for the
+/// given `pricing_version`, returning the message ID of the pending update call and
+/// the pending outcall.
+fn submit_pay_as_you_go_outcall(
+    pic: &PocketIc,
+    canister_id: Principal,
+    pricing_version: u32,
```
mraszyk: why pass pricing_Version if this is called "pay-as-you-go"?

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        responses: vec![reply(b"aa"), reply(b"bb")],
+    });
+
+    assert!(pic.get_canister_http().is_empty());
+
+    let reply = pic.await_call(call_id).unwrap();
+    let FlexibleHttpRequestResult::Ok(payloads) = decode_flexible_result(&reply) else {
+        panic!("expected a successful flexible outcall");
+    };
+    // The payloads are delivered smallest first; both have the same size here, so
```
mraszyk: if we claim that smallest come first, then let's also test this

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        total_requests: 4,
+        min_responses: 4,
+        max_responses: 4,
+    };
+    let (call_id, request) =
+        submit_flexible_outcall(&pic, canister_id, flexible_args(Some(replication)));
+    // While the outcall is in flight, its base fee and the per-replica allowances
+    // are withheld from the canister.
+    let balance_in_flight = pic.cycle_balance(canister_id);
+    assert!(balance_in_flight < balance_before);
```
mraszyk: we should assert that diff exceeds the constant amount of cycles transferred if we want to support the above claim

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+    // putting the responses into a block are not refunded.
+    let balance_after = pic.cycle_balance(canister_id);
+    assert!(
+        balance_after > balance_in_flight,
+        "expected a refund: {balance_in_flight} -> {balance_after}"
+    );
+    assert!(
+        balance_after < balance_before,
+        "expected the outcall to cost something: {balance_after} >= {balance_before}"
+    );
```
mraszyk: these assertions are pretty weak, they'd result already from regular message execution costs (WASM), independently of canister http outcalls

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        max_responses: 13,
+    };
+    let (call_id, request) = submit_flexible_outcall_with_cycles(
+        &pic,
+        canister_id,
+        flexible_args(Some(replication)),
+        CYCLES,
+    );
+
+    let body = vec![b'x'; 1_000];
```
mraszyk: maybe use a larger body to make it even more expensive and thus future-proof?

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+    // The unspent part of the per-replica allowances is refunded.
+    let balance_after = pic.cycle_balance(canister_id);
+    assert!(
+        balance_after > balance_in_flight,
+        "expected a refund: {balance_in_flight} -> {balance_after}"
+    );
+    assert!(
+        balance_after < balance_before,
+        "expected the outcall to cost something: {balance_after} >= {balance_before}"
+    );
```
mraszyk: as before, these assertions are trivial

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+}
+
+/// Invalid replication counts are rejected synchronously.
+#[test]
+fn test_flexible_canister_http_invalid_replication_counts() {
+    let pic = flexible_outcalls_pic();
+    let canister_id = deploy_test_canister(&pic);
+
+    for (replication, expected) in [
```
mraszyk: do we need a full matrix here? it seems to belong to the feature unit tests.

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        );
+        assert!(
+            message.contains(expected),
+            "rejection message {message:?} does not contain {expected:?}"
+        );
+    }
+}
+
+/// Without the beta features enabled, flexible outcalls are still available on a
+/// subnet where HTTP outcalls are free, falling back to the legacy pricing model.
```
mraszyk: assert the legacy pricing in the test (by reading out the field of the request) to support this claim

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+        max_responses: 2,
+    };
+    let (_call_id, request) =
+        submit_flexible_outcall(&pic, canister_id, flexible_args(Some(replication)));
+
+    pic.mock_flexible_canister_http_response(MockFlexibleCanisterHttpResponse {
+        subnet_id: request.subnet_id,
+        request_id: request.request_id,
+        responses: vec![reject(&"x".repeat(1025))],
+    });
```
mraszyk: could be a matrix test combined with the above one

packages/pocket-ic/tests/tests.rs
```diff
@@ -1717,6 +1725,919 @@ fn test_canister_http_timeout() {
...
+
+    let balance_after = pic.cycle_balance(canister_id);
+    assert!(
+        balance_after > balance_in_flight,
+        "expected a refund: {balance_in_flight} -> {balance_after}"
+    );
+    assert!(
+        balance_after < balance_before,
+        "expected the outcall to cost something: {balance_after} >= {balance_before}"
+    );
```
mraszyk: trivial assertions

rs/pocket_ic_server/src/pocket_ic.rs
mraszyk: + add test coverage for that

mraszyk (commented):
Tests could be made more systematic (i.e., less code) and have better coverage if we turn them into matrix tests over flexible/legacy and pay-as-you-go/legacy and ... (the scenario under test). Also prompting for "gaps in test coverage" revealed some so maybe worthwhile iterating, but ideally together with reducing code duplication via matrix tests (otherwise, more tests would make reviewing difficult...).

packages/pocket-ic/HOWTO.md
eichhorl: Changed let users mock the spent cycles now, which is a breaking change I guess

packages/pocket-ic/test_canister/canister.did
eichhorl: changed to `proxy_call : (principal, text, blob, nat) -> (RawCallResult);`

packages/pocket-ic/src/common/rest.rs
eichhorl: Added a comment to `MockCanisterHttpResponse`

packages/pocket-ic/src/common/rest.rs
eichhorl: Extended the docs to state that adding additional responses will make the returned response unpredictable and using `MockFlexible...` is not allowed.

packages/pocket-ic/src/common/rest.rs
eichhorl: true, removed the subtle implication that the non-flexible mock takes only one response

packages/pocket-ic/src/common/rest.rs
eichhorl: Added documentation for all possible results

rs/execution_environment/src/execution_environment/tests.rs
eichhorl: done

rs/https_outcalls/client/src/client.rs
eichhorl: The only system tests (and production use) that we have are for free subnets, where flexible outcalls fall back to legacy pricing, i.e. charging everything up front.

rs/pocket_ic_server/CHANGELOG.md
eichhorl: Technically, flexible outcalls are already enabled without beta features on free subnets, where they fall back to legacy pricing. Updated the changelogs

rs/pocket_ic_server/src/pocket_ic.rs
eichhorl: The spent cycles can now optionally be mocked

rs/pocket_ic_server/src/pocket_ic.rs
eichhorl: Added non-replicated documentation and matrix tests

rs/pocket_ic_server/src/pocket_ic.rs
eichhorl: It wasn't dropped, it just moved before converting the responses into the contents. The `num_responses > committee.len()` check is in `process_mock_flexible_canister_https_response`. Added tests for both (`test_canister_http_wrong_number_of_responses`, `test_flexible_canister_http_too_many_responses`).

rs/state_machine_tests/src/lib.rs
eichhorl: Nope, removed

packages/pocket-ic/tests/tests.rs
eichhorl: merged into generic `submit_outcall`

packages/pocket-ic/tests/tests.rs
eichhorl: There is now a generic `submit_outcall` with `Outcall` enum describing parameters

packages/pocket-ic/tests/tests.rs
eichhorl: done

packages/pocket-ic/tests/tests.rs
eichhorl: good idea, done

packages/pocket-ic/tests/tests.rs
eichhorl: reduced to one case

packages/pocket-ic/tests/tests.rs
eichhorl: done as part of the matrix tests

mbjorkqvist (approved):

rs/pocket_ic_server/CHANGELOG.md
```diff
@@ -18,13 +18,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `flexible_http_request` management canister endpoint on subnets that charge for HTTP outcalls; without beta features that
+  endpoint is only available on subnets where HTTP outcalls are free (system subnets and subnets with a free cycles cost
+  schedule). Beta features also enable the pay-as-you-go pricing model, which flexible outcalls are always priced with and
+  which an `http_request` can select through its `pricing_version` field.
 
 ### Changed
 - Mocked canister HTTP responses report the cycles their node actually spent on the outcall, instead of reporting no spend at
   all. This applies to `/instances/<instance_id>/update/mock_canister_http` and
 `/instances/<instance_id>/update/mock_flexible_canister_http`.
+- **Breaking:** every mocked response of `/instances/<instance_id>/update/mock_canister_http` and
```
mraszyk: check

packages/pocket-ic/src/common/rest.rs
```diff
@@ -1307,20 +1313,96 @@ pub enum CanisterHttpResponse {
...
+    Clone, Serialize, Deserialize, Debug, Hash, Eq, PartialEq, Ord, PartialOrd, JsonSchema,
+)]
+pub struct MockCanisterHttpNodeResponse {
+    pub response: CanisterHttpResponse,
+    /// The cycles the node reports having spent on the outcall.
+    ///
+    /// `None` lets PocketIC derive the spend with the same pricing machinery a real
+    /// node uses. That derivation includes a term for how long the outcall took,
+    /// which PocketIC measures in process, so set this explicitly to make a test's
+    /// cycles accounting fully deterministic.
```
mraszyk: If it's just the time that makes the thing non-deterministic, we could also make that time a user-supplied parameter, defaulting to zero (instantaneous outcall) instead of using any wall clock time measurement in PocketIC.

packages/pocket-ic/HOWTO.md
mraszyk: alternatively, we could add an optional field to both `CanisterHttpReply` and `CanisterHttpReply`; it'd also be a breaking change, but we'd avoid one type-level indirection

packages/pocket-ic/src/common/rest.rs
eichhorl: Defaulting to zero would have to be threaded into the HTTP client's budget tracker, which can be done of course, or recompute everything without the time term as part of pocket-ic.

In that case, would we still need the ability to mock the spent cycles?

packages/pocket-ic/HOWTO.md
eichhorl: I feel like those types are intended to mirror the replica's types, which do not carry the spent. A difference in cycles spent would also make two replies incomparable even if they are otherwise equal,

packages/pocket-ic/src/common/rest.rs
mraszyk: > Defaulting to zero would have to be threaded into the HTTP client's budget tracker

I'd go for this unless it's too complicated/fragile (since this is also the production path)

> recompute everything without the time term as part of pocket-ic.

that code duplication doesn't sound great

> would we still need the ability to mock the spent cycles?

no, I don't think so, we should faithfully reuse as much logic (formulas) as possible

mraszyk:
I also extracted the following gaps in test coverage when chatting with Claude about it (I admit to not have reviewed the new tests manually):

- The timeout test dropped the drain assertion. Master asserted `get_canister_http()` is empty after advancing past the timeout.
- `process_mock_canister_https_response` deliberately does not reject a flexible outcall. That's a documented non-failure, and it's the inverse of `test_flexible_mock_of_*_outcall` — but no test confirms the claim holds, so if it ever became an error, nothing would notice.
- `FlexibleHttpRequestErr::message` is never asserted. None of the four global-error tests looks at it.
- node_id and report are never asserted anywhere. And the entire `HttpRequestResourceReport` goes unchecked; notably nothing ever observes a `ResourceUsage::Exceeded`, which is how a node that blew a budget is meant to be reported.

packages/pocket-ic/src/common/rest.rs
eichhorl: Done: response time is now zero for pocket-ic, and the spent cycles mocking is removed again

eichhorl:
> * The timeout test dropped the drain assertion. Master asserted `get_canister_http()` is empty after advancing past the timeout.
> * `process_mock_canister_https_response` deliberately does not reject a flexible outcall. That's a documented non-failure, and it's the inverse of `test_flexible_mock_of_*_outcall` — but no test confirms the claim holds, so if it ever became an error, nothing would notice.
> * `FlexibleHttpRequestErr::message` is never asserted. None of the four global-error tests looks at it.
> * node_id and report are never asserted anywhere. And the entire `HttpRequestResourceReport` goes unchecked; notably nothing ever observes a `ResourceUsage::Exceeded`, which is how a node that blew a budget is meant to be reported.

Thanks, prompted to add those tests. Note that the `HttpRequestResourceReport` are still unpopulated even by the replica (this will be done eventually as a follow up). Until then the tests now assert that they are indeed unpopulated.

mraszyk (approved):

PR #11249, opened 2026-08-21
feat(registry): ensure all elected versions can be parsed

pierugo-dfinity:
This PR adds a registry invariant that ensures all elected GuestOS and HostOS versions can be parsed by the replicas that will consume them.

Note the
> This wouldn't happen in practice (because of validation that happens elsewhere)

[here](https://github.com/dfinity/ic/blob/b495c4345cc5a59d756e1da93381447db47aeac0/rs/registry/helpers/src/subnet.rs#L449-L451), which referred to validation when updating the `StandardEngineReplicaVersion` in *governance*, but the registry itself did not validate them. In particular, versions were not validated neither at election nor at deployment to regular subnets.


rs/registry/canister/src/invariants/hostos_version.rs
```diff
@@ -60,3 +68,47 @@ fn get_all_hostos_versions_of_nodes(snapshot: &RegistrySnapshot) -> Vec<String>
...
+    // to build HostOS).
+    const HOSTOS_VERSION_ID: &str = "eb3ab997954f2a91db8a42f84132cf37078d481c";
+
+    fn register_version_mutation(hostos_version_id: &str) -> Vec<RegistryMutation> {
+        vec![insert(
+            make_hostos_version_key(hostos_version_id).as_bytes(),
+            HostosVersionRecord {
+                release_package_urls: vec![],
+                release_package_sha256_hex: "".to_string(),
+                hostos_version_id: hostos_version_id.to_string(),
```
daniel-wong-dfinity-org-twin: This should be first, and set apart using a blank line, since the other stuff is basically noise. [Lead with the lede][lede].

[lede]: https://github.com/dfinity/ic/blob/master/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L49

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/src/invariants/hostos_version.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/11249/commits/b73b5f4b164dd3f60fdad3c192adf2d07b2cd419

PR #11250, opened 2026-08-21
feat(cketh): install a deposit address' delegation with its first sweep

gregorydemay:
## Why

- A deposit address holds ERC-20 tokens but no code, so it cannot sweep itself.
- The minter delegates it to the sweeper contract with an EIP-7702 authorization.
- The cheapest carrier for that authorization is the sweep that needs it: the first sweep touching an address installs the delegation on the way.
- The delegation persists, so every later sweep of that address is a plain EIP-1559 transaction.
- `deposit_from_cex_demo` measures 94'932 gas for the first sweep of an address against 63'252 for the next.

## What

- The sweeper lane sends either transaction type, and a sweep carries the signed authorizations it must install.
- Two no-op refactorings come first: finalizing and fee-bumping stop being EIP-1559-only.
- That is also what gives EIP-7702 fee-bumping for free, with no machinery of its own.
- Authorizations are held in the request rather than derived while the transaction is built, so no replay and no fee bump ever re-signs one.
- An authorization covers the chain, the delegate and the authority's nonce — nothing of the outer transaction.
- A single place decides which type a sweep becomes, from whether anything is left to install.

## Scope

- Nothing enqueues a sweep yet, so nothing builds an authorization in production.
- Choosing which addresses still need delegating belongs to the sweep-queue source, along with reading each address' delegation from the chain.
- Hence the request carries signed authorizations rather than a "needs delegation" flag.
- The batch-dependent gas limit a delegating sweep needs is deferred with it: #11237's flat 100'000 does not cover one.

## Candid compatibility

- Needs the `CI_OVERRIDE_DIDC_CHECK` label.
- Two sweeper event cases change the type of their `transaction` field, and a third gains a required field.
- Neither is a Candid subtype on a returned variant.
- Those cases only reached the interface with #11144 and have never been released, so no deployed canister has emitted them and no client can be reading them.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/audit/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/audit/tests.rs
@@ -1,5 +1,8 @@
 use crate::checked_amount::CheckedAmountOf;
-use crate::endpoints::events::{Event as CandidEvent, EventPayload, UnsignedTransaction};
+use crate::endpoints::events::{
+    Event as CandidEvent, EventPayload, SignedAuthorization as CandidSignedAuthorization,
+    UnsignedSweeperTransaction, UnsignedTransaction,
+};
 use crate::erc20::CkErc20Token;
 use crate::eth_logs::{LedgerSubaccount, ReceivedErc20Event, ReceivedEthEvent};
 use crate::eth_rpc_client::responses::TransactionReceipt;
@@ -12,8 +15,9 @@ use crate::state::transactions::{
 };
 use crate::timed_sized_map::Timestamp;
 use crate::tx::{
-    AccessList, AccessListItem, Eip1559TransactionRequest, SignedEip1559TransactionRequest,
-    StorageKey,
+    AccessList, AccessListItem, Eip1559TransactionRequest, SignedAuthorization,
+    SignedEip1559TransactionRequest, SignedSweepTransaction, StorageKey, SweepTransaction,
+    TransactionSignature,
 };
 use candid::Principal;
 use ic_agent::identity::AnonymousIdentity;
@@ -180,7 +184,9 @@ impl GetEventsFile {
             }
         }
 
-        fn map_signed_transaction(raw_transaction: &str) -> SignedEip1559TransactionRequest {
+        fn decode_signed_transaction(
+            raw_transaction: &str,
+        ) -> (Eip1559TransactionRequest, TransactionSignature) {
             use crate::tx::TransactionSignature;
             use ethers_core::types::transaction::eip2718::TypedTransaction;
             use ethnum::u256;
@@ -252,7 +258,38 @@ impl GetEventsFile {
                 s: map_ethers_u256(decoded_sig.s),
             };
 
-            SignedEip1559TransactionRequest::from((request, signature))
+            (request, signature)
+        }
+        fn map_signed_sweep_transaction(raw_transaction: &str) -> SignedSweepTransaction {
+            let (transaction, signature) = decode_signed_transaction(raw_transaction);
+            SignedSweepTransaction::from((SweepTransaction::Eip1559(transaction), signature))
+        }
+
+        fn map_unsigned_sweeper_transaction(tx: UnsignedSweeperTransaction) -> SweepTransaction {
+            SweepTransaction::new(
+                map_unsigned_transaction(tx.transaction),
+                map_authorizations(tx.authorization_list),
+            )
+        }
+
+        fn map_authorizations(
+            authorizations: Vec<CandidSignedAuthorization>,
+        ) -> Vec<SignedAuthorization> {
+            fn map_signature_component(bytes: &[u8]) -> ethnum::u256 {
+                ethnum::u256::from_be_bytes(<[u8; 32]>::try_from(bytes).unwrap())
+            }
+
+            authorizations
+                .into_iter()
+                .map(|authorization| SignedAuthorization {
+                    chain_id: authorization.chain_id.0.to_u64().unwrap(),
+                    delegate: authorization.delegate.parse().unwrap(),
+                    nonce: authorization.nonce.try_into().unwrap(),
+                    y_parity: authorization.y_parity,
+                    r: map_signature_component(&authorization.r),
+                    s: map_signature_component(&authorization.s),
+                })
+                .collect()
         }
 
         Event {
@@ -358,7 +395,9 @@ impl GetEventsFile {
                     raw_transaction,
                 } => ET::SignedTransaction {
                     withdrawal_id: map_nat(withdrawal_id),
-                    transaction: map_signed_transaction(&raw_transaction),
+                    transaction: SignedEip1559TransactionRequest::from(decode_signed_transaction(
+                        &raw_transaction,
+                    )),
                 },
                 EventPayload::ReplacedTransaction {
                     withdrawal_id,
@@ -381,6 +420,7 @@ impl GetEventsFile {
                     data,
                     max_transaction_fee,
                     created_at,
+                    authorizations,
                 } => ET::AcceptedSweepRequest(SweepRequest {
                     id: SweepId(sweep_id.0.to_u64().unwrap()),
                     destination: destination.parse().unwrap(),
@@ -388,27 +428,28 @@ impl GetEventsFile {
                     data: data.into_vec(),
                     max_transaction_fee: max_transaction_fee.try_into().unwrap(),
                     created_at,
+                    authorizations: map_authorizations(authorizations),
                 }),
                 EventPayload::CreatedSweeperTransaction {
                     sweep_id,
                     transaction,
                 } => ET::CreatedSweeperTransaction {
                     sweep_id: SweepId(sweep_id.0.to_u64().unwrap()),
-                    transaction: map_unsigned_transaction(transaction),
+                    transaction: map_unsigned_sweeper_transaction(transaction),
                 },
                 EventPayload::SignedSweeperTransaction {
                     sweep_id,
                     raw_transaction,
                 } => ET::SignedSweeperTransaction {
                     sweep_id: SweepId(sweep_id.0.to_u64().unwrap()),
-                    transaction: map_signed_transaction(&raw_transaction),
+                    transaction: map_signed_sweep_transaction(&raw_transaction),
                 },
                 EventPayload::ReplacedSweeperTransaction {
                     sweep_id,
                     transaction,
                 } => ET::ReplacedSweeperTransaction {
                     sweep_id: SweepId(sweep_id.0.to_u64().unwrap()),
-                    transaction: map_unsigned_transaction(transaction),
+                    transaction: map_unsigned_sweeper_transaction(transaction),
                 },
                 EventPayload::FinalizedSweeperTransaction {
                     sweep_id,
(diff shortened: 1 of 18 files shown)


rs/ethereum/cketh/minter/tests/dump_stable_memory.rs
```diff
@@ -181,7 +186,38 @@ fn map_signed_transaction(raw_transaction: &str) -> SignedEip1559TransactionRequ
         s: map_ethers_u256(decoded_sig.s),
     };
 
-    SignedEip1559TransactionRequest::from((request, signature))
+    (request, signature)
+}
+fn map_signed_sweep_transaction(raw_transaction: &str) -> SignedSweepTransaction {
+    let (transaction, signature) = decode_signed_transaction(raw_transaction);
+    SignedSweepTransaction::from((SweepTransaction::Eip1559(transaction), signature))
+}
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Already handled on the branch, in ebdb22b58c ("fix(cketh): replay a signed EIP-7702 sweep as the type it was sent as") — this comment predates that push.

`map_signed_sweep_transaction` now reads the leading type byte and, for `0x04`, decodes through `SignedEip7702TransactionRequest::decode` into `SweepTransaction::Eip7702(signed.transaction().clone())`, which carries the `authorization_list`, so the delegations survive the round trip. Only a non-`0x04` transaction reaches the EIP-1559 path. The `decode` it uses is new in `tx/eip_7702.rs` and rejects both a wrong type byte and empty input, covered by `should_decode_the_raw_bytes_it_encoded`, `should_refuse_to_decode_a_transaction_of_another_type` and `should_refuse_to_decode_an_empty_transaction` in `tx/tests.rs`.

rs/ethereum/cketh/minter/src/state/audit/tests.rs
```diff
@@ -237,7 +243,40 @@ impl GetEventsFile {
                 s: map_ethers_u256(decoded_sig.s),
             };
 
-            SignedEip1559TransactionRequest::from((request, signature))
+            (request, signature)
+        }
+        fn map_signed_sweep_transaction(raw_transaction: &str) -> SignedSweepTransaction {
+            let (transaction, signature) = decode_signed_transaction(raw_transaction);
+            SignedSweepTransaction::from((SweepTransaction::Eip1559(transaction), signature))
+        }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Same as the sibling comment on `tests/dump_stable_memory.rs`: fixed in ebdb22b58c ("fix(cketh): replay a signed EIP-7702 sweep as the type it was sent as"), which landed after this review.

The replay mapper no longer forces the EIP-1559 variant — it branches on the leading type byte and reconstructs `SweepTransaction::Eip7702` (authorization list included) for a type-`0x04` sweep, so the mainnet/Sepolia replay tests cannot hit the `decode_signed_transaction` rejection you describe.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -188,4 +192,147 @@ impl SignableTransaction for Eip7702TransactionRequest {
...
+                    })
+                })
+                .collect::<Result<_, String>>()?,
+        };
+        let signature = TransactionSignature {
+            signature_y_parity: decode_val(&rlp, 10)?,
+            r: decode_u256(&rlp, 11)?,
+            s: decode_u256(&rlp, 12)?,
+        };
+        Ok(Self::from((transaction, signature)))
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed and fixed in 2ca527dd60.

The path is exactly as described: `decode` ended with `Self::from(...)`, which hashes eagerly in `Signed::new`, so the RLP encoding ran and hit `rlp_inner`'s `assert!(!self.authorization_list.is_empty())` — a fallible decoder panicking on the malformed input whose error is the easiest to describe. I mutation-checked it: with the guard removed, the new test panics at `eip_7702.rs:160`, which is that assertion.

The list is now decoded before the transaction is built and an empty one returns `Err`, so `rlp_inner`'s assertion stays what it is meant to be — an internal invariant check that untrusted input can no longer reach. Covered by `should_refuse_to_decode_a_transaction_without_authorizations`, which re-encodes a valid sample transaction with field 9 replaced by an empty list, alongside the existing wrong-type and empty-input cases.

For the record on severity: every caller of this decoder is a test or test tool (`tests/dump_stable_memory.rs`, `state/audit/tests.rs`, `tx/tests.rs`), so nothing in the canister could reach the panic — but the function documents `# Errors` and is `Result`-returning, so the contract was wrong regardless.

rs/ethereum/cketh/minter/src/tx/sweep.rs
```diff
@@ -0,0 +1,135 @@
...
+/// A deposit address is delegated once and stays delegated, so a sweep needs type `0x04` only as
+/// long as it still touches an address no earlier sweep delegated. The
+/// [`SweepTransaction::Eip7702`] variant therefore always carries a non-empty authorization list:
+/// [`SweepTransaction::new`] is what decides the variant, and it decides on exactly that.
+#[derive(Clone, Eq, PartialEq, Debug, Decode, Encode)]
+pub enum SweepTransaction {
+    #[n(0)]
+    Eip1559(#[n(0)] Eip1559TransactionRequest),
+    #[n(1)]
+    Eip7702(#[n(0)] Eip7702TransactionRequest),
```
mbjorkqvist: `SweepTransaction::new` is the only thing upholding "`Eip7702` implies a non-empty authorization list", but the variant is public, so the state it forbids stays representable — and the check that catches it is `rlp_inner`'s `assert!`, reached eagerly through `Signed::new`'s hash computation.

That is the same mechanism as the decoder panic fixed in 2ca527dd, at a second entry point: `minicbor::Decode for Signed<T>` also calls `Signed::new`, so a persisted `SignedSweeperTransaction` holding an empty-list `Eip7702` would hit the assertion during `post_upgrade` event replay. The event log is append-only in stable memory, so that is an unupgradeable canister rather than a failed operation.

Nothing can produce such an event today: `new` never picks `Eip7702` for an empty list, and it is the only production constructor. But the bypass exists and is already in use — `tests/dump_stable_memory.rs` and `state/audit/tests.rs` both build `SweepTransaction::Eip7702(signed.transaction().clone())` directly.

Both this and the decoder bug are one gap: the type permits a state the encoder rejects, and each entry point gets its own guard. Closing it structurally needs a thin newtype (`Eip7702(DelegatingSweep)` with a private inner and a fallible constructor), since Rust cannot make a variant's field private — more ceremony than a third guard, so worth weighing against the blast radius of the CBOR path rather than an automatic change.

rs/ethereum/cketh/minter/src/state/transactions/request.rs
```diff
@@ -242,16 +260,19 @@ impl PipelineRequest for SweepRequest {
...
-            nonce,
-            max_priority_fee_per_gas: gas_fee_estimate.max_priority_fee_per_gas,
-            max_fee_per_gas,
-            gas_limit,
-            destination: self.destination,
-            amount: self.amount,
-            data: self.data.clone(),
-            access_list: Default::default(),
-        })
+        Ok(SweepTransaction::new(
```
mbjorkqvist: Nothing checks that each authorization's `chain_id` matches the chain the sweep is built for. `SweepTransaction::new` copies `transaction.chain_id` into the EIP-7702 transaction and takes the authorizations as given.

EIP-7702 *skips* an authorization whose chain id is neither 0 nor the current chain; it does not reject the transaction. So a mismatch fails silently on-chain: the sweep still succeeds, still pays the per-authorization gas, the delegation is never installed, and the sweep call then hits an EOA with no code. The pipeline records it as a finalized success.

`Authorization::sign` already refuses `chain_id == 0`, so the hazard is clearly in view — this is the other half of it. `ethereum_network` is in scope right here, and `assert!(gas_limit > GasAmount::ZERO, "BUG: ...")` a few lines up is the established style; an equivalent assert over `self.authorizations` would cost nothing and fail loudly at construction rather than silently on-chain.

The authority-nonce and derivation-path-vs-authority checks genuinely need chain state and belong with the sweep-queue source. This one does not.

rs/ethereum/cketh/minter/src/tx/sweep.rs
```diff
@@ -0,0 +1,135 @@
...
+impl SweepTransaction {
+    /// The sweep `transaction`, installing the delegations `authorizations` attests to: type
+    /// `0x04` if there are any to install, and type `0x02` if there are none.
+    pub fn new(
+        transaction: Eip1559TransactionRequest,
+        authorizations: Vec<SignedAuthorization>,
+    ) -> Self {
+        if authorizations.is_empty() {
+            return Self::Eip1559(transaction);
+        }
```
mbjorkqvist, line 79: EIP-7702 skips an authorization whose chain id is neither 0 nor the current chain; it does not reject the transaction. So a mismatch fails silently on-chain: the sweep still succeeds, still pays the per-authorization gas, the delegation is never installed, and the sweep call then hits an EOA with no code. The pipeline records it as a finalized success.

Authorization::sign already refuses chain_id == 0, but would it make sense to add a check here to make sure that the chain ID of the authorizations match the current chain, as well as a corresponding test?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/tx/sweep.rs
gregorydemay: 🤖 Applied in 7448f33b14, taking the newtype rather than a third guard — the CBOR path decided it. `minicbor::Decode for Signed<T>` maps through `Signed::new`, which hashes eagerly, so an event carrying an empty-list `Eip7702` would trap in `post_upgrade` replay: an unupgradeable canister against an append-only log, not a failed operation.

`Eip7702` now holds a `DelegatingSweep` with a private field, constructed only through `DelegatingSweep::new` (returning `Option`), and its `Decode` goes through that same constructor rather than the derive, so the guard cannot be skipped by the deserialization path. The encoding stays `#[cbor(transparent)]`, so the wire format and the CBOR golden test are unchanged.

Both bypasses you named — `state/audit/tests.rs` and `tests/dump_stable_memory.rs` — now go through the constructor, as does the trap test in `state/transactions/tests.rs`. `rlp_inner`'s assertion stays as a belt-and-braces check on an invariant the type now carries, and the decoder guard from 2ca527dd stays too, since `SignedEip7702TransactionRequest::decode` returns that type rather than a `SweepTransaction`.

Two tests: constructing with an empty list yields `None`, and decoding CBOR that carries an empty list fails with a message rather than trapping, while a well-formed one round-trips.

rs/ethereum/cketh/minter/src/tx/sweep.rs
gregorydemay: 🤖 Good catch — the silent part is what makes it worth guarding. Added in 78bb2dcb2c: `SweepTransaction::new` now asserts every authorization's chain id is either zero (valid on any chain, per the EIP) or the transaction's own.

An assertion rather than a `Result`, because the minter signs these authorizations itself with its own chain id: a mismatch is a bug in the minter, not a condition the sweep should carry. That matches how the file already treats a request and its transaction disagreeing.

Two tests: a mismatched chain id traps, and a zero chain id is accepted and kept in the list.

PR #11251, opened 2026-08-21
feat(cketh): expose per-token minimum deposit amounts in MinterInfo

gregorydemay:
Balance-scan filter 1 only treats a deposit address as a scan candidate once its balance reaches a per-token minimum, so a smaller deposit is silently never detected and never credited. For the CEX deposit flow that threshold was invisible: a user could send a small amount to their deposit address, see nothing happen, and have no way to find out why.

`get_minter_info` now reports `minimum_deposit_amounts` — the minimum for each supported ckERC20 token, in that token's own units — so users and frontends can show the threshold alongside the deposit address. It is `null` while the ckERC20 feature is inactive, matching the existing per-token fields.

The thresholds are still the hard-coded rate snapshot; they will start tracking live prices once DEFI-2961's daily exchange-rate refresh lands, and this endpoint will reflect that automatically.

### DIFF
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -250,6 +250,10 @@ type MinterInfo = record {
     // This might be less that the actual amount available on the `minter_address()`.
     erc20_balances : opt vec record { erc20_contract_address: text; balance: nat};
 
+    // Minimum balance, in each token's own units, that a deposit address must hold for the
+    // balance scan to treat it as a deposit candidate. A smaller balance is never detected.
+    minimum_deposit_amounts : opt vec record { erc20_contract_address: text; minimum_deposit_amount: nat};
+
     // Last scraped block number for logs of the ETH helper contract.
     last_eth_scraped_block_number: opt nat;
 
--- a/rs/ethereum/cketh/minter/src/main.rs
+++ b/rs/ethereum/cketh/minter/src/main.rs
@@ -4,7 +4,7 @@ use dashboard::DashboardTemplate;
 use ic_canister_log::log;
 use ic_cdk::{init, post_upgrade, pre_upgrade, query, update};
 use ic_cketh_minter::address::{AddressValidationError, validate_address_as_destination};
-use ic_cketh_minter::balance_scan::balance_scan;
+use ic_cketh_minter::balance_scan::{balance_scan, min_deposit};
 use ic_cketh_minter::deposit::{refresh_latest_block_height, scrape_logs};
 use ic_cketh_minter::endpoints::ckerc20::{
     RetrieveErc20Request, WithdrawErc20Arg, WithdrawErc20Error,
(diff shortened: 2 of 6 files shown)


rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -240,6 +240,10 @@ type MinterInfo = record {
     // This might be less that the actual amount available on the `minter_address()`.
     erc20_balances : opt vec record { erc20_contract_address: text; balance: nat};
 
+    // Minimum balance, in each token's own units, that a deposit address must hold for the
+    // balance scan to treat it as a deposit candidate. A smaller balance is never detected.
+    minimum_deposit_amounts : opt vec record { erc20_contract_address: text; minimum_deposit_amount: nat};
```
mbjorkqvist, line 255: Would it make sense to change `minimum_deposit_amount` to `opt nat`, to handle cases where a token is supported, but no minimum deposit amount has been set? Currently, `Erc20Value::MAX` is returned in such cases, which can be confusing.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -284,26 +285,34 @@ async fn eip_1559_transaction_price(
...
+                        )
+                    })
+                    .unzip();
+                let minimum_deposit_amounts = s
+                    .supported_ck_erc20_tokens()
+                    .map(|token| Erc20MinimumDeposit {
+                        erc20_contract_address: token.erc20_contract_address.to_string(),
+                        minimum_deposit_amount: min_deposit(&token.erc20_contract_address).into(),
+                    })
+                    .collect();
```
mbjorkqvist, line 311: The way support for new tokens is added uses two separate mechanisms: A token is added to `supported_ck_erc20_tokens()` through a proposal, but `min_deposit()` for a new token is added through a compile-time addition to `MIN_DEPOSITS`. It is currently possible to add a new supported token using a proposal, even though no minimum deposit has been configured for that token. As noted in the comment on the `minimum_deposit_amounts` in the `MinterInfo` struct in `cketh_minter.did`, this results in an `Erc20Value::MAX` value being returned for such a token. Would it make sense to address this in one of the following ways?

- Make the `minimum_deposit_amount` value be `opt nat` in the Candid response (as suggested in the other comment); or
- Add a check when trying to add support for a new token, and fail the request in case no minimum deposit amount has been configured; or
- Take the (initial) minimum deposit amount as input when adding support for a new token.

The last options seems the most robust, but would involve more (breaking) changes to the API, and seems to be (partially) conflicting with DEFI-2961 that proposes to refresh the minimum deposit amount daily using the XRC.

mbjorkqvist (commented):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: I think it makes sense to first discuss where we want to go/where we are now:

- I think we need what DEFI-2961 describes, where the exchange rates are fetched from the XRC and refreshed, say daily and could be fetched when a new token is initially added.
- Assuming this is what we want to do (don't hesitate to challenge), the current implementation is definitely a (temporary) crutch but I would push back on changes externally visible (API)
    - why should `minimum_deposit_amount` be optional? Every supported token has actually a minimum deposit amount, maybe initially set to a very high unrealistic value.
    - The minter should decide what's the minimum deposit amount (because it depends on the amount of work it will need to do and the minter knows that best) so I wouldn't ask the caller to initially choose and I would not take the first deposit for it.

WDYT?

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: See my [response](https://github.com/dfinity/ic/pull/11251#discussion_r3843778579) below.

rs/ethereum/cketh/minter/src/main.rs
mbjorkqvist: I do think DEFI-2961 makes sense, but it's probably also not the highest priority at the moment. Maybe an optional value is a bad idea, but a comment would at least help? I believe this would now affect ckBAT, since [this proposal](https://dashboard.internetcomputer.org/proposal/143606) passed, so if we were to go live with deposits from CEXes now without updating `MIN_DEPOSITS`, BAT deposits would not arrive as ckBAT.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: Discussed offline, left as is. Added #11296 to add that information also when the user is notifying the minter of a deposit.

PR #11254, opened 2026-08-21
fix(cketh): recover a signature against the key that made it

gregorydemay:
## Why

`compute_recovery_id` recovers a signature's parity bit against `lazy_call_ecdsa_public_key()` — the minter's *master* key — while `sign` and `Authorization::sign` sign under a caller-supplied derivation path. Recovery only ever succeeds against the key that produced the signature, so signing under anything but the empty main path finds no parity at all: `try_recovery_from_digest` returns an error and the `unwrap_or_else` traps. The `debug_assert!` above it fails first in debug builds.

Nothing calls it with a non-empty path yet, so this is latent today. It stops being latent with the first EIP-7702 authorization, which is signed by a deposit address' own key by definition — and a tuple that does not `ecrecover` to that address is skipped silently by the protocol, leaving the batch's inner call to hit a code-less address and revert the whole sweep.

## What

Recovery now derives the master key along the same path before recovering. The derivation is shared with address derivation rather than duplicated: `deposit_address::derive_public_key` becomes the single place a subkey is computed, and address derivation is that key's address. The key a signature is recovered against is therefore, by construction, the key whose address it belongs to.

The signing functions now take the path as `Vec<ByteBuf>` — the form `MAIN_DERIVATION_PATH`, `sweeper_derivation_path()` and `deposit_derivation_path()` already produce — and build the management-canister `DerivationPath` themselves. That is the structural half of the fix: one owner for both uses of the path, so the signing key and the recovery key cannot drift apart again. Call sites lose a wrapper.

An empty path derives to the master key, so the main address and every withdrawal are unaffected.

## Tests

Three pure unit tests, no canister involved. The one that pins the bug signs a digest with a key derived along a deposit path and asserts recovery succeeds against the derived key and fails against the master key, so it demonstrates the trap rather than asserting the new code agrees with itself.

### DIFF
--- a/rs/ethereum/cketh/minter/src/deposit_address/tests.rs
+++ b/rs/ethereum/cketh/minter/src/deposit_address/tests.rs
@@ -141,3 +141,93 @@ fn master_private_key() -> (PrivateKey, [u8; 32]) {
 fn account(owner: Principal, subaccount: Option<Subaccount>) -> Account {
     Account { owner, subaccount }
 }
+
+mod derive_public_key {
+    use crate::deposit_address::{
+        DepositAddressSchema, deposit_address, derive_public_key, sweeper_derivation_path,
+    };
+    use crate::{MAIN_DERIVATION_PATH, address::ecdsa_public_key_to_address};
+    use candid::Principal;
+    use ic_secp256k1::{DerivationIndex, DerivationPath, PrivateKey};
+    use icrc_ledger_types::icrc1::account::Account;
+    use serde_bytes::ByteBuf;
+
+    const CHAIN_CODE: [u8; 32] = [7_u8; 32];
+
+    #[test]
+    fn should_derive_the_key_whose_address_is_the_deposit_address() {
+        let master = master_private_key().public_key();
+        let account = account();
+
+        for schema in [DepositAddressSchema::CkErc20, DepositAddressSchema::CkEth] {
+            let path = super::super::deposit_derivation_path(schema, &account);
+            let derived = derive_public_key(&master, &CHAIN_CODE, &path);
+
+            assert_eq!(
+                &ecdsa_public_key_to_address(&derived),
+                deposit_address(&master, &CHAIN_CODE, schema, &account).as_address()
+            );
+        }
+    }
+
+    #[test]
+    fn should_derive_the_master_key_for_the_main_address() {
+        let master = master_private_key().public_key();
+
+        assert_eq!(
+            derive_public_key(&master, &CHAIN_CODE, &MAIN_DERIVATION_PATH),
+            master
+        );
+    }
+
+    /// A signature made under a derivation path only recovers against the key derived along that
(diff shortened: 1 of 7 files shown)


rs/ethereum/cketh/minter/src/tx/mod.rs
```diff
@@ -177,8 +179,18 @@ pub fn encode_u256<T: Into<u256>>(stream: &mut RlpStream, value: T) {
     stream.append(&value.to_be_bytes()[leading_empty_bytes..].as_ref());
 }
 
-async fn compute_recovery_id(digest: &Hash, signature: &[u8]) -> RecoveryId {
-    let ecdsa_public_key = lazy_call_ecdsa_public_key().await;
+/// The parity bit of a signature made under `derivation_path`.
+///
+/// Recovery only succeeds against the public key that produced the signature, so the master key is
+/// derived along the same path first: signing under a deposit address' path and recovering against
+/// the master key finds no parity at all.
```
gregorydemay:
```suggestion
/// derived along the same path first.
```

rs/ethereum/cketh/minter/src/deposit_address/tests.rs
```diff
@@ -141,3 +141,93 @@ fn master_private_key() -> (PrivateKey, [u8; 32]) {
...
+                    .is_ok()
+            );
+            assert!(
+                master
+                    .public_key()
+                    .try_recovery_from_digest(&digest, &signature)
+                    .is_err()
+            );
+        }
+    }
```
mbjorkqvist, line 213: In addition to this, would it make sense to add a PocketIC test that signs a digest under a derived path, and checks `try_recovery_from_digest` against `derive_public_key`?

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/tx/mod.rs
gregorydemay: 🤖 Applied in 8a2235f4d0.

rs/ethereum/cketh/minter/src/deposit_address/tests.rs
gregorydemay: 🤖 Agreed, and worth recording why, since the idea is sound in itself.

What a PocketIC test would add over the unit test here is agreement between *our* derivation and the *platform's*: both sides of this test come from the same crypto library, so it would still pass if `derive_public_key` disagreed with what the management canister actually signs under. But testing that is testing the threshold-ECDSA implementation, and at this layer nothing in production signs under a derived path — the main-address flow signs under the empty one — so it would need test-only surface on the minter.

That agreement is checked end-to-end further up the stack instead: the minter signs an attestation under a deposit address' path via threshold ECDSA, and the delegate's `ecrecover` has to return the deposit address computed from `derive_public_key`. If the two derivations disagreed, the sweep would revert on chain.

PR #11256, opened 2026-08-21
feat(cketh): attest a deposit address' account so anyone can sweep it

gregorydemay:
## Why

- A deposit address' funds belong to an IC account, and the sweeper contract has to know which one before it moves them.
- The minter is the only party that knows, since it derived the address from that account.
- Attesting it on chain means anyone can pay for a sweep: the minter's signature, not the caller, is what says where the funds go.

## What

- `attestation_digest` commits to the account with `keccak256("ck-deposit-owner" || chain_id || helper || principal || subaccount)`.
- The domain separator and chain id stop an attestation being replayed as anything else, and the helper address ties it to one deposit-address scheme.
- `sign_attestation` signs that digest with the deposit address' own threshold-ECDSA key, so `ecrecover` in the delegate yields the address being swept.
- An absent subaccount attests as the default one, and a principal is encoded the way the helper contract carries it.

## Tests

- The digest is pinned against a golden value computed independently in Python, not by the code under test.
- Every field is shown to change the digest, so nothing is left out of the commitment.


rs/ethereum/cketh/minter/src/eth_logs/tests.rs
```diff
@@ -522,3 +522,34 @@ mod subaccount {
...
+            Principal::from_slice(&[1, 2, 3, 4]),
+            Principal::from_slice(&[0xff; 29]),
+            candid::Principal::from_text("2chl6-4hpzw-vqaaa-aaaaa-c").unwrap(),
+        ] {
+            assert_eq!(
+                parse_principal_from_slice(&encode_principal(&principal)),
+                Ok(principal)
+            );
+        }
+    }
```
gregorydemay: Let's turn this into a proptest, see arb_principal

rs/ethereum/cketh/minter/src/attestation.rs
```diff
@@ -0,0 +1,71 @@
+//! The one-time attestation that binds a deposit address to the IC account its funds are credited
```
gregorydemay: nit: let's have this file as attestation/mod.rs

rs/ethereum/cketh/minter/src/attestation.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub v: u8,
+}
+
+/// The digest a deposit address signs to attest that it credits `account`, mirroring
+/// `CkSweeperAttested._attestationDigest`:
+/// `keccak256("ck-deposit-owner" || chain_id || helper || principal || subaccount)`.
+///
+/// Every field is fixed-length, so no two field assignments share a preimage. `chain_id` prevents
+/// replay onto another chain and `helper` binds the attestation to one helper deployment.
+pub fn attestation_digest(chain_id: u64, helper: &Address, account: &Account) -> Hash {
```
gregorydemay: Let's be more explicit about the `helper`. Is this the sweeper batch smart contract or is it the deposit helper smart contract. Those are not the same

rs/ethereum/cketh/minter/src/attestation/tests.rs
```diff
@@ -0,0 +1,88 @@
+use crate::attestation::attestation_digest;
```
gregorydemay: Let's add a proptest that the attestation preimage (for abitrary chain_id, Ethereum address, ledger account) is always constant length

rs/ethereum/cketh/minter/src/attestation.rs
```diff
@@ -0,0 +1,71 @@
...
+const DOMAIN_SEPARATOR: &[u8] = b"ck-deposit-owner";
+
+/// A deposit address' attestation of the account it credits, as the `(r, s, v)` components the
+/// delegate passes to `ecrecover`.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub struct Attestation {
+    pub r: [u8; 32],
+    pub s: [u8; 32],
+    pub v: u8,
+}
```
gregorydemay: Why not reuse what we have?

We could have the unsigned statement as
```rust
pub struct AttestationRequest {
    pub chain_id: u64,
    pub helper: Address,
    pub account: Account,
}
```

and then the signed attestation is simply something like InnerSigned<AttestationRequest>

rs/ethereum/cketh/minter/src/attestation.rs
gregorydemay: 🤖 Done in 5fceac8805 — `attestation.rs` is now `attestation/mod.rs`, beside its `tests.rs`.

rs/ethereum/cketh/minter/src/attestation.rs
gregorydemay: 🤖 You are right that it reads either way, and the contract settles it: `CkSweeperAttested` takes `HELPER` in its constructor and its digest is `keccak256("ck-deposit-owner", block.chainid, HELPER, principal, subaccount)`. That is the **deposit helper** — the subaccount-aware contract whose events name the account an address credits. The sweeper delegate is deliberately absent from the preimage, because inside the delegate call its own address is `address(this)`, which is what `ecrecover` compares against.

So the field is `deposit_helper` now, in 5fceac8805, and its doc says which contract and why getting it wrong is expensive: attesting under the sweeper address yields a signature the delegate rejects, and nothing catches that until the sweep runs on chain.

rs/ethereum/cketh/minter/src/attestation.rs
gregorydemay: 🤖 Took the first half, and a reuse you did not name; declined the `InnerSigned` part. 5fceac8805.

`AttestationRequest { chain_id, deposit_helper, account }` is in, with `preimage()` and `digest()` on it. It replaces two three-argument functions, says which helper is meant, and gives the constant-length property something to assert on — so it answers three of your comments at once.

The signature reuse goes further than the struct: `sign_digest` already returns a `TransactionSignature`, and the bespoke `Attestation { r, s, v }` was that same data with `v = 27 + parity`. That is an ABI-encoding convention, so `Attestation` is gone and the conversion now lives in `sweeper_contract` where the call data is built. The golden call-data test passes unchanged, i.e. the bytes are identical.

Against `InnerSigned<AttestationRequest>`: `InnerSigned` and `Signed` are transaction machinery. The field is named `transaction`, every impl on them is gated on `T: SignableTransaction` — a type byte, RLP encoding, chain id, nonce, gas, destination, amount, access list — and `Signed::new` memoises a *transaction* hash from `type_byte || rlp(...)`. An attestation is a keccak over a fixed 132-byte preimage, signed by a deposit address' key and consumed by `ecrecover`; it has no type byte, no nonce and no RLP form. Fitting it in would mean either implementing `SignableTransaction` for it, which is meaningless, or loosening that bound across the transaction code — paying in the transaction layer for a saving in the attestation one.

rs/ethereum/cketh/minter/src/attestation/tests.rs
gregorydemay: 🤖 Added in e902721580, over arbitrary chain id, address, principal and subaccount: the preimage is always 16 + 32 + 20 + 32 + 32 bytes.

No filtering needed — `encode_principal` pads to 32 bytes whatever the principal's length, which is the point the property demonstrates. It needed `preimage()` to exist, which the `AttestationRequest` refactor provides.

rs/ethereum/cketh/minter/src/eth_logs/tests.rs
gregorydemay: 🤖 Done in e433e30161, using `arb_principal` from `test_fixtures`.

One wrinkle worth naming: `arb_principal` draws 0..=29 random bytes, so it generates both principals the helper refuses — the empty slice (management canister) and the single byte `4` (anonymous). The proptest filters those two, and `should_reject_what_the_helper_never_carries` keeps pinning them as the negative case, so the pair now says "everything round-trips except exactly these two".

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
```diff
@@ -256,6 +256,22 @@ pub enum EventSourceError {
...
+        bytes.len() <= 29,
+        "BUG: a principal is at most 29 bytes, got {}",
+        bytes.len()
+    );
+    let mut encoded = [0_u8; 32];
+    encoded[0] = bytes.len() as u8;
+    encoded[1..=bytes.len()].copy_from_slice(bytes);
+    encoded
+}
+
```
mbjorkqvist, line 262: It seems this new `encode_principal` function was added in the middle of the documentation for `parse_principal_from_slice`.

rs/ethereum/cketh/minter/src/attestation/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Sign `request` with the deposit address' own derived key. The delegate wants the signature as
+/// `(r, s, v)`, which is where [`crate::sweeper_contract`] encodes it.
+///
+/// # Errors
+/// * a description of why the threshold-ECDSA signature could not be produced.
+pub async fn sign_attestation(
+    request: &AttestationRequest,
+    schema: DepositAddressSchema,
+) -> Result<TransactionSignature, String> {
```
mbjorkqvist, line 96: The parameters here (`request.chain_id`, `request.deposit_helper`, `request.account`, and `schema`) are not tied together or verified here (or at construction of `AttestationRequest`). If incorrect values are passed (e.g., another `chain_id`, or an unknown `deposit_helper`), we would return a well-formed attestation for the wrong address. The minter already knows about, and could check 3/4 (the chain ID, deposit helper address, and the schema).

A check could be added at construction (make the fields private in `AttestationRequest`, including `schema`), together with a constructor, or; a runtime check could be added for whatever is passed in.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/eth_logs/mod.rs
gregorydemay: 🤖 Fixed in c6dcf78c90 — the encoder and its own doc moved above that block, so the decoding rules document `parse_principal_from_slice` again.

It was worse than a misplacement: the whole block, including the list of principals the helper refuses, had become the encoder's documentation, and `parse_principal_from_slice` was left with only "This method MUST never panic".

rs/ethereum/cketh/minter/src/attestation/mod.rs
gregorydemay: 🤖 Agreed, and done in cee4b88a12 — the silent part is what makes it worth enforcing rather than documenting: the signature is well formed either way, and only the delegate's `ecrecover` rejects it, on chain, far from the caller that passed the wrong value.

Rather than checking the values, they are no longer the caller's to pass. `AttestationRequest`'s fields are private, `schema` travels with them so the signing path cannot disagree with the digest, and `State::attestation_request(account)` builds one from the chain id and the subaccount-aware helper the minter actually runs against — returning `None` while that helper is unknown. `sign_attestation` takes only the request, and the enqueue path no longer names a chain or a helper at all, so the variable that could have been wrong is gone from it.

`new` stays `pub(crate)`, documented as "prefer `State::attestation_request`". Worth noting why that is the ceiling rather than a compromise: a constructor in `crate::attestation` cannot be scoped `pub(in crate::state)`, since Rust only allows restricting visibility to an ancestor module. The digest tests keep building requests directly, being a child module of `attestation` and therefore able to see its private fields.

PR #11257, opened 2026-08-21
feat(cketh): encode the delegate's batch sweep call

gregorydemay:
## Why

- One sweep transaction moves the tokens of many deposit addresses, so the minter has to build the delegate's batch call itself.

## What

- `encode_sweep_erc20_batch` encodes the call: selector `3a7ce054`, then the deposit addresses with the account each is credited to and its attestation, then the tokens to move.
- The tokens are a separate array so a batch can sweep several of them without repeating a token per item.

## Tests

- The encoded call is checked against an independent ABI encoder (`ethers-core`).
- The selector is checked against the keccak of the function signature.
- Offsets are checked where ABI encoding usually goes wrong: the tokens array starts right after the items, and an empty batch still encodes both heads.


rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
```diff
@@ -0,0 +1,76 @@
+//! The sweeper delegate contract's interface, as the minter calls it.
```
gregorydemay, line 1: Let's move this to sweeper_contract/mod.rs

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::deposit_address::DepositAddress;
+use crate::eth_logs::encode_principal;
+use crate::tx::TransactionSignature;
+use ic_ethereum_types::Address;
+use icrc_ledger_types::icrc1::account::Account;
+
+/// First 4 bytes of
+/// keccak256("sweepErc20Batch((address,bytes32,bytes32,bytes32,bytes32,uint8)[],address[])").
+const SWEEP_ERC20_BATCH_SELECTOR: [u8; 4] = hex_literal::hex!("3a7ce054");
```
gregorydemay, line 18: We should add a unit test that this indeed matches `keccak256("sweepErc20Batch((address,bytes32,bytes32,bytes32,bytes32,uint8)[],address[])")`

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub attestation: TransactionSignature,
+}
+
+/// Encode `sweepErc20Batch(SweepItem[] items, address[] tokens)` on the deployed delegate, which
+/// sweeps every `token` balance held by every `item`'s deposit address to the minter's main address
+/// through the deposit helper.
+///
+/// See the [Contract ABI Specification](https://docs.soliditylang.org/en/develop/abi-spec.html#contract-abi-specification):
+/// both arguments are dynamic, so the head holds one offset each and the two blocks follow.
+pub fn encode_sweep_erc20_batch(items: &[SweepItem], tokens: &[Address]) -> Vec<u8> {
```
gregorydemay, line 45: This seems to assume that we will use the contract rs/ethereum/cketh/minter/tests/deposit_from_cex_demo/CkSweeperAttested.sol.

This is correct, but then let's move the contract to the same dir as the other ones in rs/ethereum/cketh/minter/

rs/ethereum/cketh/minter/src/sweeper_contract/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::tx::TransactionSignature;
+use candid::Principal;
+use ethnum::u256;
+use ic_ethereum_types::Address;
+use icrc_ledger_types::icrc1::account::Account;
+
+// Encoded independently of this crate, from the ABI specification, for the two items and two tokens
+// `sweep()` builds below: selector, then the two head offsets (0x40 and 0x1e0), the inline
+// `SweepItem[]` and the `address[]`.
+const GOLDEN: &str = "3a7ce054\
```
gregorydemay: instead of hard-coding this, can't we use the existing dev dependency ethers to verify this?

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+/// and that account's attestation, signed by the address' own key.
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct SweepItem {
+    pub deposit: DepositAddress,
+    pub account: Account,
+    pub attestation: TransactionSignature,
+}
+
+/// Encode `sweepErc20Batch(SweepItem[] items, address[] tokens)` on the deployed delegate, which
+/// sweeps every `token` balance held by every `item`'s deposit address to the minter's main address
```
gregorydemay, line 36: How would  a batch be build if for example:

1. Address 1 only has USDT
2. Address 2 has USDT and USDC

Would Address 1 still be swept for USDC?

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
gregorydemay: 🤖 Done in 0e4ee151c7 — `sweeper_contract/mod.rs`, beside its `tests.rs`.

*(Edited: split into separate commits rather than amended, so the SHA changed.)*

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
gregorydemay: 🤖 Added in 7a88e8c7dc: `should_call_the_function_the_delegate_exposes` hashes the signature string and compares the first four bytes of the encoded call data.

I checked the string against the contract rather than trusting it: `sweepErc20Batch(SweepItem[] calldata items, address[] calldata tokens)` with `SweepItem { address deposit; bytes32 principal; bytes32 subaccount; bytes32 r; bytes32 s; uint8 v; }`, so the tuple matches field for field, `uint8 v` last.

It hashes with `ic_sha3` rather than an ABI library on purpose, so it stays independent of the encoder check below.

*(Edited: split into separate commits rather than amended, so the SHA changed.)*

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
gregorydemay: 🤖 Moved in 0e4ee151c7: `CkSweeperAttested.sol` now sits next to `DepositHelperWithSubaccount.sol`, `ERC20DepositHelper.sol` and `EthDepositHelper.sol`, with the `srcs` entry and the `CKSWEEPER_ATTESTED_SOL` rootpath in `BUILD.bazel` updated.

`CkSweeperViaHelper.sol` and `MockUSDT.sol` stay under `tests/deposit_from_cex_demo/` — nothing in production encodes against either, so they are still demo fixtures.

*(Edited: split into separate commits rather than amended, so the SHA changed.)*

rs/ethereum/cketh/minter/src/sweeper_contract/tests.rs
gregorydemay: 🤖 Done in 7a88e8c7dc, with one thing worth flagging: your DEFI-2970 chain deletes this dependency. #11302 checks the EIP-7702 encoding against alloy, #11303 is "encode the ABI calls the tests check with alloy", and #11304 drops `ethers-core` altogether.

So I wrote it with `ethers-core`, which is the dev-dependency that exists at this layer, and #11303 will convert it along with the rest — rather than adding `alloy` here, which would mean touching `Cargo.toml`, `bazel/rust.MODULE.bazel` and re-pinning in a PR about call-data encoding, and colliding with that work.

`should_encode_a_batch_sweep` now builds the arguments with `ethers_core::abi::encode` from `Token`s and compares against everything after the selector, so the hand-computed golden is gone. The selector keeps its own test, so the two checks stay independent: one library-derived, one a constant.

*(Edited: split into separate commits rather than amended, so the SHA changed.)*

rs/ethereum/cketh/minter/src/sweeper_contract/mod.rs
gregorydemay: 🤖 Address 1 would be swept for USDC too, and you are right that it should be called out — TODO(DEFI-2980) added on the encoder in 6b081174c3.

Concretely: `tokens` is one array applied to every item. `sweepErc20Batch` loops the items and passes the same `tokens` to each address' `sweepErc20`, which loops the tokens and reads `balanceOf` per token. So a batch is a cross-product. Nothing incorrect happens for address 1 and USDC — the `balance > 0` guard skips the transfer — but it pays a `balanceOf` call for a pair that holds nothing, about 4'600 gas in the numbers I measured.

DEFI-2980 removes it by grouping the queue by token, so each batch carries exactly one token and address 1 only appears in the USDT batch. The call data shape does not change: a one-element `tokens` array encodes the same way, which is why that fix touches only the enqueue side.

*(Edited: split into separate commits rather than amended, so the SHA changed.)*

rs/ethereum/cketh/minter/src/sweeper_contract/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+    ]);
+
+    assert_eq!(&data[4..], expected_arguments.as_slice());
+}
+
+#[test]
+fn should_place_the_tokens_right_after_the_items() {
+    for count in 0..4_usize {
+        let (items, tokens) = sweep();
+        let items = &items[..count.min(items.len())];
```
mbjorkqvist: Understanding question: We iterate `for count in 0..4_usize`, but `sweep()` only returns `Vec`s of length 2, so the fourth iteration is a repeat of the third. Is this intentional (to potentially handle future changes in what `sweep()` returns), is it a leftover from when `sweep()` returned more items, or something else? In particular, I'm wondering why we wouldn't do `for count in 0..=items.len()`?

mbjorkqvist (approved):
Thanks @gregorydemay! One minor comment regarding the PR description - it says "The encoding is pinned against a golden byte string computed independently in Python.", but this was replaced in https://github.com/dfinity/ic/pull/11257/commits/8a2a2191c5d3a964fe45a5bf2b6ce1aaf4cde41b .

rs/ethereum/cketh/minter/src/sweeper_contract/tests.rs
gregorydemay: 🤖 Good catch, and no — nothing intentional behind it. The hardcoded `4` and the `min()` were two clamps papering over each other, so the last iteration was a duplicate of the one before and bought nothing, not even robustness to `sweep()` growing.

Your form in 7f3960ecd2:

```rust
    let (items, tokens) = sweep();
    for count in 0..=items.len() {
        let items = &items[..count];
```

`sweep()` also moved out of the loop, since it returns the same batch every time.

gregorydemay:
🤖 Thanks — you are right, and I had missed it: my sweep of the PR looked at review threads only, so a review body with no inline comment went unseen.

The description now says what the tests do:

- The encoded call is checked against an independent ABI encoder (`ethers-core`).
- The selector is checked against the keccak of the function signature.

I checked the other claim in the same breath: the selector `3a7ce054` in the **What** section is still correct for `sweepErc20Batch((address,bytes32,bytes32,bytes32,bytes32,uint8)[],address[])`.

PR #11258, opened 2026-08-21
feat(cketh): enqueue a batched sweep from the sweep queue

gregorydemay:
## Why

- Detected deposits sat in the sweep queue with nothing turning them into a transaction.

## What

- A minute task folds the sweep queue into one batch per token, capped at 20 deposits each.
- One sweep per token, never a mixed batch: the delegate's batch entry point runs its whole token list against every address it touches, so a mixed batch would pay for the full cross product of balance checks.
- Each batch signs one attestation and one EIP-7702 authorization per deposit address; stored attestations are reused instead of re-signed.
- Every authorization is signed for nonce 0, whatever the address holds: it either installs the delegation or is skipped because one is already installed, so no delegation state needs tracking or chain reads.
- The signed batch becomes a sweep request that the existing sweeper pipeline prices, signs, sends, and finalizes.
- A deposit whose attestation or authorization cannot be signed is left out of the batch and offered again on the next tick.

### DIFF
--- a/publish/canisters/BUILD.bazel
+++ b/publish/canisters/BUILD.bazel
@@ -89,7 +89,8 @@ CANISTERS_MAX_SIZE_COMPRESSED_E5_BYTES = {
     # Size is currently at 923_414 bytes.
     "ic-ckbtc-minter.wasm.gz": 10,
     # Size when constraint addded: 1_230_630 bytes
-    "ic-cketh-minter.wasm.gz": 13,
+    # Size is currently at 1_309_035 bytes.
+    "ic-cketh-minter.wasm.gz": 14,
     # The orchestrator needs to embed 3 wasms at compile time
     # (ICRC1 index, ICRC1 ledger, and ICRC1 archive) and size is
     # therefore strictly controlled.
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -28,7 +28,7 @@ use crate::{
     numeric::{TransactionCount, TransactionNonce, Wei},
     state::{
         State, TaskType,
-        audit::{EventType, process_event},
+        audit::{EventType, process_event, process_event_at},
         automatic_deposits::SweepTarget,
         mutate_state, read_state,
         transactions::{CreateSweepTransactionError, PipelineRequest, SweepRequest, SweptDeposit},
@@ -346,9 +346,25 @@ pub async fn enqueue_batched_sweep() {
         return;
     };
 
+    let context = SweepContext {
+        sweeper_contract,
+        gas_fee_estimate,
+        created_at: ic_cdk::api::time(),
+    };
     for batch in batches {
-        enqueue_token_sweep(batch, sweeper_contract, &gas_fee_estimate).await;
+        enqueue_token_sweep(batch, &context).await;
     }
+}
+
+/// What every sweep enqueued in one tick shares: the delegate they authorize and send to, what their
+/// gas costs, and the time they are decided at.
+///
+/// The time is carried rather than read because reading it needs a canister, and every sweep of one
+/// tick is decided at the same instant anyway.
+struct SweepContext {
+    sweeper_contract: Address,
+    gas_fee_estimate: GasFeeEstimate,
+    created_at: u64,
 }
 
 /// The sweep queue folded into one batch per token, each holding that token's queued deposits in
@@ -384,8 +400,7 @@ fn sweep_batches_by_token(state: &State) -> BTreeMap<Address, Vec<SweepTarget>> 
 /// for, or when the sweep would prepay more than [`max_sweep_transaction_fee`].
 async fn enqueue_token_sweep(
     (token, targets): (Address, Vec<SweepTarget>),
-    sweeper_contract: Address,
-    gas_fee_estimate: &GasFeeEstimate,
+    context: &SweepContext,
 ) {
     let Some(requests) = attestation_requests(&targets) else {
         log!(
@@ -395,7 +410,7 @@ async fn enqueue_token_sweep(
         return;
     };
 
-    let prepared = prepare_deposits(&targets, &requests, sweeper_contract).await;
+    let prepared = prepare_deposits(&targets, &requests, context).await;
     if prepared.is_empty() {
         log!(
             INFO,
@@ -407,15 +422,16 @@ async fn enqueue_token_sweep(
     let batch = SweepBatch::from(prepared);
     let mut request = SweepRequest {
         id: read_state(|s| s.next_sweep_id),
-        destination: sweeper_contract,
+        destination: context.sweeper_contract,
         amount: Wei::ZERO,
         data: encode_sweep_erc20_batch(&batch.items, &batch.tokens),
         max_transaction_fee: Wei::ZERO,
-        created_at: ic_cdk::api::time(),
+        created_at: context.created_at,
         authorizations: batch.authorizations,
         deposits: batch.deposits,
     };
-    request.max_transaction_fee = gas_fee_estimate
+    request.max_transaction_fee = context
+        .gas_fee_estimate
         .clone()
         .to_price(request.gas_limit())
         .max_transaction_fee();
@@ -431,11 +447,18 @@ async fn enqueue_token_sweep(
     }
     log!(
         INFO,
-        "[enqueue_batched_sweep]: sweeping {} deposits of {token} from {} addresses, via {sweeper_contract}",
+        "[enqueue_batched_sweep]: sweeping {} deposits of {token} from {} addresses, via {}",
         request.deposits.len(),
-        request.authorizations.len()
+        request.authorizations.len(),
+        context.sweeper_contract
     );
-    mutate_state(|s| process_event(s, EventType::AcceptedSweepRequest(request)));
+    mutate_state(|s| {
+        process_event_at(
+            context.created_at,
+            s,
+            EventType::AcceptedSweepRequest(request),
+        )
+    });
 }
 
 /// The most a single sweep may prepay: the balance the sweeper address is kept above, so that a
@@ -548,13 +571,13 @@ struct PreparedDeposit {
 async fn prepare_deposits(
     targets: &[SweepTarget],
     requests: &BTreeMap<Account, AttestationRequest>,
-    sweeper_contract: Address,
+    context: &SweepContext,
 ) -> Vec<PreparedDeposit> {
-    let attestations = sign_attestations_batch(requests).await;
+    let attestations = sign_attestations_batch(requests, context).await;
     // Only the attested accounts: a deposit that could not be attested drops out of the sweep
     // anyway, so authorizing it would spend a signature on nothing.
     let addresses = attested_addresses(targets, &attestations);
-    let authorizations = sign_authorizations_batch(&addresses, sweeper_contract).await;
+    let authorizations = sign_authorizations_batch(&addresses, context).await;
     prepared_deposits(targets, &attestations, &authorizations)
 }
 
@@ -588,6 +611,7 @@ fn prepared_deposits(
 /// have yet cost a signature, and each is recorded the moment it exists.
 async fn sign_attestations_batch(
     requests: &BTreeMap<Account, AttestationRequest>,
+    context: &SweepContext,
 ) -> BTreeMap<Account, TransactionSignature> {
     let mut attestations: BTreeMap<Account, TransactionSignature> = read_state(|s| {
         requests
@@ -612,7 +636,8 @@ async fn sign_attestations_batch(
         match result {
             Ok(attestation) => {
                 mutate_state(|s| {
-                    process_event(
+                    process_event_at(
+                        context.created_at,
                         s,
                         EventType::AttestedDepositAddress {
                             request: request.clone(),
@@ -652,12 +677,12 @@ fn attested_addresses(
 /// fixed at 0, so nothing a sweep does can invalidate it — see [`delegation_authorization`].
 async fn sign_authorizations_batch(
     addresses: &BTreeMap<Account, DepositAddress>,
-    sweeper_contract: Address,
+    context: &SweepContext,
 ) -> BTreeMap<Account, SignedAuthorization> {
     let chain_id = read_state(State::ethereum_network).chain_id();
     let signed = join_all(addresses.keys().map(|account| {
         let account = *account;
-        let authorization = delegation_authorization(chain_id, sweeper_contract);
+        let authorization = delegation_authorization(chain_id, context.sweeper_contract);
         async move {
             (
                 account,
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -35,8 +35,8 @@ use crate::{
     },
     sweeper_contract::{SweepItem, encode_sweep_erc20_batch},
     tx::{
-        Authorization, GasFeeEstimate, SignedAuthorization, TransactionSignature,
-        lazy_refresh_gas_fee_estimate,
+        Authorization, EcdsaSigner, GasFeeEstimate, IcEcdsaSigner, SignedAuthorization,
+        TransactionSignature, lazy_refresh_gas_fee_estimate,
     },
     withdraw::{
         fetch_finalized_receipts, finalized_transaction_count, latest_transaction_count,
@@ -347,6 +347,7 @@ pub async fn enqueue_batched_sweep() {
     };
 
     let context = SweepContext {
+        signer: IcEcdsaSigner,
         sweeper_contract,
         gas_fee_estimate,
         created_at: ic_cdk::api::time(),
@@ -356,12 +357,14 @@ pub async fn enqueue_batched_sweep() {
     }
 }
 
-/// What every sweep enqueued in one tick shares: the delegate they authorize and send to, what their
-/// gas costs, and the time they are decided at.
+/// What every sweep enqueued in one tick shares: the delegate they authorize and send to, the key
+/// that signs for them, what their gas costs, and the time they are decided at.
 ///
-/// The time is carried rather than read because reading it needs a canister, and every sweep of one
-/// tick is decided at the same instant anyway.
-struct SweepContext {
+/// Generic over the signer, and carrying the time rather than reading it, because those are the two
+/// things the enqueue path needs a canister for; everything else it does is arithmetic over state,
+/// so a test that supplies both drives the real code.
+struct SweepContext<S> {
+    signer: S,
     sweeper_contract: Address,
     gas_fee_estimate: GasFeeEstimate,
     created_at: u64,
@@ -398,9 +401,9 @@ fn sweep_batches_by_token(state: &State) -> BTreeMap<Address, Vec<SweepTarget>> 
 ///
 /// Skips the token — leaving its deposits queued for the next tick — when nothing could be signed
 /// for, or when the sweep would prepay more than [`max_sweep_transaction_fee`].
-async fn enqueue_token_sweep(
+async fn enqueue_token_sweep<S: EcdsaSigner>(
     (token, targets): (Address, Vec<SweepTarget>),
-    context: &SweepContext,
+    context: &SweepContext<S>,
 ) {
     let Some(requests) = attestation_requests(&targets) else {
         log!(
@@ -568,10 +571,10 @@ struct PreparedDeposit {
 /// Everything the queued `targets` need signed, in target order. Each threshold-ECDSA signature is
 /// an outcall, so the attestations are signed in one batch and the authorizations in another, one of
 /// each per deposit address rather than per queued token.
-async fn prepare_deposits(
+async fn prepare_deposits<S: EcdsaSigner>(
     targets: &[SweepTarget],
     requests: &BTreeMap<Account, AttestationRequest>,
-    context: &SweepContext,
+    context: &SweepContext<S>,
 ) -> Vec<PreparedDeposit> {
     let attestations = sign_attestations_batch(requests, context).await;
     // Only the attested accounts: a deposit that could not be attested drops out of the sweep
@@ -609,9 +612,9 @@ fn prepared_deposits(
 /// signed. An attestation is signed once per address and reused by every later sweep — it names the
 /// account and the helper, neither of which a sweep changes — so only the ones the minter does not
 /// have yet cost a signature, and each is recorded the moment it exists.
-async fn sign_attestations_batch(
+async fn sign_attestations_batch<S: EcdsaSigner>(
     requests: &BTreeMap<Account, AttestationRequest>,
-    context: &SweepContext,
+    context: &SweepContext<S>,
 ) -> BTreeMap<Account, TransactionSignature> {
     let mut attestations: BTreeMap<Account, TransactionSignature> = read_state(|s| {
         requests
@@ -627,7 +630,11 @@ async fn sign_attestations_batch(
             .iter()
             .filter(|(account, _request)| !attestations.contains_key(*account))
             .map(|(account, request)| async move {
-                (*account, request, sign_attestation(request).await)
+                (
+                    *account,
+                    request,
+                    sign_attestation(&context.signer, request).await,
+                )
             }),
     )
     .await;
@@ -675,9 +682,9 @@ fn attested_addresses(
 ///
 /// Every address is signed afresh: the tuple names the chain and the delegate, and its nonce is
 /// fixed at 0, so nothing a sweep does can invalidate it — see [`delegation_authorization`].
-async fn sign_authorizations_batch(
+async fn sign_authorizations_batch<S: EcdsaSigner>(
     addresses: &BTreeMap<Account, DepositAddress>,
-    context: &SweepContext,
+    context: &SweepContext<S>,
 ) -> BTreeMap<Account, SignedAuthorization> {
     let chain_id = read_state(State::ethereum_network).chain_id();
     let signed = join_all(addresses.keys().map(|account| {
@@ -687,10 +694,10 @@ async fn sign_authorizations_batch(
             (
                 account,
                 authorization
-                    .sign(deposit_derivation_path(
-                        DepositAddressSchema::CkErc20,
-                        &account,
-                    ))
+                    .sign(
+                        &context.signer,
+                        deposit_derivation_path(DepositAddressSchema::CkErc20, &account),
+                    )
                     .await,
             )
         }
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -63,8 +63,8 @@ use std::collections::{BTreeMap, BTreeSet};
 /// batch of twenty: the second doubling saves ~2%, one transaction's intrinsic gas spread over the
 /// extra deposits. A wider batch does not pay for what it costs — `sweepErc20Batch` has no
 /// per-item error handling, so it doubles how many deposits one revert drops (see
-/// [`next_batch_of`]), and it doubles what a single sweep prepays against the sweeper's low-water
-/// mark (see [`max_sweep_transaction_fee`]).
+/// [`sweep_batches_by_token`]), and it doubles what a single sweep prepays against the sweeper's
+/// low-water mark (see [`max_sweep_transaction_fee`]).
 const MAX_DEPOSITS_PER_SWEEP: usize = 10;
 
 const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
@@ -351,8 +351,8 @@ pub async fn enqueue_batched_sweep() {
     }
 }
 
-/// The sweep queue folded into one batch per token, each holding at most
-/// [`MAX_DEPOSITS_PER_SWEEP`] of that token's queued deposits.
+/// The sweep queue folded into one batch per token, each holding that token's queued deposits in
+/// queue order, up to [`MAX_DEPOSITS_PER_SWEEP`] of them.
 ///
 /// One sweep per token, never a batch spanning several. The delegate's batch entry point runs its
 /// whole token list against every deposit address it touches, so a mixed batch pays a `balanceOf` at
@@ -362,26 +362,20 @@ pub async fn enqueue_batched_sweep() {
 /// [`MAX_DEPOSITS_PER_SWEEP`] bound a sweep's gas at all. It costs one transaction's 21'000
 /// intrinsic gas per token, which a multi-token sweep pays anyway in the cold storage write each
 /// token's balance at the minter needs.
-fn sweep_batches_by_token(state: &State) -> BTreeMap<Address, Vec<SweepTarget>> {
-    let mut queued: BTreeMap<Address, Vec<SweepTarget>> = BTreeMap::new();
-    for target in state.automatic_deposits.sweep_targets_iter() {
-        queued.entry(target.token()).or_default().push(target);
-    }
-    queued
-        .into_iter()
-        .map(|(token, queued)| (token, next_batch_of(queued)))
-        .collect()
-}
-
-/// Which of one token's queued deposits go into the next sweep of it: its deposits in queue order,
-/// up to [`MAX_DEPOSITS_PER_SWEEP`].
 ///
 /// Every deposit gets one sweep and no more. `sweepErc20Batch` has no per-item error handling, so
 /// whatever makes a sweep revert — one address blacklisted for the token, say — reverts every batch
 /// that deposit is in; a failed sweep therefore drops its whole batch from the queue rather than
 /// leaving anything behind to retry (DEFI-2981).
-fn next_batch_of(queued: Vec<SweepTarget>) -> Vec<SweepTarget> {
-    queued.into_iter().take(MAX_DEPOSITS_PER_SWEEP).collect()
+fn sweep_batches_by_token(state: &State) -> BTreeMap<Address, Vec<SweepTarget>> {
+    let mut batches: BTreeMap<Address, Vec<SweepTarget>> = BTreeMap::new();
+    for target in state.automatic_deposits.sweep_targets_iter() {
+        let batch = batches.entry(target.token()).or_default();
+        if batch.len() < MAX_DEPOSITS_PER_SWEEP {
+            batch.push(target);
+        }
+    }
+    batches
 }
 
 /// Enqueue one sweep of `token`, moving as many of `targets` as could be signed for.
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -405,6 +405,9 @@ async fn enqueue_token_sweep<S: EcdsaSigner>(
     (token, targets): (Address, Vec<SweepTarget>),
     context: &SweepContext<S>,
 ) {
+    if targets.is_empty() {
+        return;
+    }
     let Some(requests) = attestation_requests(&targets) else {
         log!(
             DEBUG,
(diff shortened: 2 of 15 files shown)


rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +267,157 @@ async fn finalize_transactions_batch(sender: Address) {
...
+                }
+                Err(e) => {
+                    log!(
+                        INFO,
+                        "[enqueue_batched_sweep]: failed attesting {account:?}, skipping the whole sweep: {e}"
+                    );
+                    return;
+                }
+            },
+        };
```
gregorydemay: we should sign an attestation batch in parallel. Skip items that couldn't be signed but proceed with the rest. See `sign_transactions_batch`

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +267,157 @@ async fn finalize_transactions_batch(sender: Address) {
...
+        data: encode_sweep_erc20_batch(&items, &tokens),
+        max_transaction_fee: Wei::ZERO,
+        created_at: ic_cdk::api::time(),
+        authorizations,
+        deposits,
+    };
+    // What this sweep may spend on gas, and the ceiling its transaction is priced at. Taken from
+    // the estimate now rather than left open: an unbounded allowance would price the transaction
+    // above anything the sweeper address could pay, and the pipeline refuses to send a sweep whose
+    // allowance the fee has outgrown, so it waits for a cheaper block instead.
```
gregorydemay:
```suggestion
```

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +267,157 @@ async fn finalize_transactions_batch(sender: Address) {
...
+                Ok(authorization) => authorizations.push(authorization),
+                Err(e) => {
+                    log!(
+                        INFO,
+                        "[enqueue_batched_sweep]: failed authorizing {account:?}, skipping the whole sweep: {e}"
+                    );
+                    return;
+                }
+            }
+        }
```
gregorydemay: same here, let's process the authorizations in parallel, log errors, and proceed with the non-error cases.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Done in 1a55e6951f, following `sign_transactions_batch`.

`sign_attestations_batch` reads the stored attestations first, then signs only the missing ones with `join_all`, recording each new one as it comes back. It returns them in request order, `None` where signing failed, and logs the failures together:

```rust
async fn sign_attestations_batch(requests: &[AttestationRequest]) -> Vec<Option<TransactionSignature>>
```

A deposit left without an attestation drops out of the sweep instead of returning; the others do not depend on it. Worth flagging what this cost before: a sweep of twenty deposits made up to forty sequential outcalls, and one failure discarded the nineteen deposits that had nothing to do with it.

Verified with `//rs/ethereum/cketh/minter:deposit_from_cex` locally — the happy path still credits both CEX deposits. The skip paths are not unit-testable as written, since signing goes through the management canister.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Same treatment in 1a55e6951f: `sign_authorizations_batch` signs one authorization per address the sweep still has to delegate, all with `join_all`, and returns them keyed by index so a failure is simply absent.

One thing to note about "proceed with the non-error cases" here: an address that has no authorization yet cannot be swept at all, so a failed authorization drops that deposit from the batch rather than sweeping it undelegated — the delegated call would revert. `prepare_deposits` is where the two phases meet: it keeps a deposit only if it has an attestation and, when the address still needs one, an authorization.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Removed in 8aa4cac6f8, its own commit. The `TODO(DEFI-2933)` right below it stays.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -299,6 +306,53 @@ impl AutomaticDeposits {
...
+                .get_mut(&request)
+                .unwrap_or_else(|| panic!("BUG: {request:?} is not queued for sweeping"));
+            assert_eq!(
+                entry.swept_by, None,
+                "BUG: {request:?} was already taken by sweep {:?}",
+                entry.swept_by
+            );
+            entry.swept_by = Some(sweep_id);
+            if deposit.delegating {
+                self.delegated.insert(deposit.address);
```
gregorydemay: 🤖 **Delegation is recorded at enqueue, but sweeps can be reordered — a later sweep can then no-op silently.**

The doc justifies recording here with "an EIP-7702 authorization is applied before the call runs", which holds only if the carrying transaction actually lands first. It need not:

1. Account `A` funds USDC → sweep #1 enqueued with `A`'s authorization, `delegated.insert(A)` runs here.
2. `A` funds USDT → sweep #2 enqueued, `is_delegated(A)` is already `true`, so it carries **no** authorization.
3. Sweep #1 is rescheduled by `create_transactions_batch` (`sweep.rs:176-186`) because its frozen `max_transaction_fee` is below the current fee. `reschedule_request` (`state/transactions/mod.rs:600`) moves it to the *back* of the queue, and `next_transaction_nonce()` is read fresh per creation — so sweep #2 takes the lower nonce and lands first.

On chain `A` still has no code. `CkSweeperAttested.sol:74` does `CkSweeperAttested(item.deposit).sweepErc20(...)`, which returns no value under `pragma ^0.8.20`, so Solidity emits no `extcodesize` guard: the CALL to a codeless address succeeds with empty returndata and the loop continues. The sweep is mined "successfully", moves zero tokens, and `(A, USDT)` stays marked `swept_by` forever.

Recording the delegation on finalization, or having sweep #2 carry an authorization whenever the installing sweep has not yet finalized, would both close this.

gregorydemay (commented):
🤖 ## Adversarial clean-code review

Two independent passes over the 5 commits. **12 findings**, ordered by severity in the inline comments.

**Three correctness findings I'd resolve before merge:**

1. **Delegation is recorded at enqueue, but sweeps can be reordered** — a later sweep that skipped its authorization can land first against a still-codeless EOA. Because `sweepErc20` returns no value under `pragma ^0.8.20`, Solidity emits no `extcodesize` guard, so the call succeeds as a *silent no-op*: the sweep mines fine, moves nothing, and the deposit stays marked swept forever.
2. **`gas_limit()` is linear in `deposits.len()` while the contract does `items × tokens` work** — `sweepErc20Batch` hands the whole token array to every item. A mixed-token batch can run out of gas, and marking `swept_by` at enqueue means it's never retried.
3. **`swept_by` is never cleared anywhere in the tree**, and `max_transaction_fee` is frozen at enqueue — so a sweep that never lands strands its deposits with no recovery path.

**One root cause behind two more findings:** `items`, `authorizations` and the attestation requests are all keyed per `(account, token)`, but a deposit address, its attestation and its EIP-7702 authorization are derived per **account** (`deposit_derivation_path` takes no token). A multi-token account therefore pays N threshold-ECDSA signatures where 1 suffices and puts N duplicate entries in the call data — which undoes the attestation reuse that `45209e5688` in this same stack was written to add.

Notably the state layer handles this case *correctly* and has a fixture documented as "One account with two funded tokens at the same deposit address"; it's the consumer that doesn't, and nothing tests the join. That's the theme: `sweep.rs` gained ~275 lines and has no test module at all, and every correctness finding above lives in that untested seam.

**Deliberately not flagged**, having checked the rest of the stack: `enqueue_batched_sweep` having no caller, and `SweepTarget::scanned_balance` looking dead — both are wired up downstream and are staging artifacts, not defects. The shared task guard *is* flagged, but only because the PR description claims it's already fixed here.

The docs throughout are unusually good — the `# Panics` sections and the explicit note on `apply_state_transition` blast radius made this reviewable. Most of what's above is about invariants that hold at enqueue time but not at execution time.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -246,6 +256,47 @@ pub struct SweepRequest {
...
+        }
+    }
+}
+
+impl SweepRequest {
+    /// Gas limit for this sweep's transaction, scaled to the number of deposits it moves.
+    pub fn gas_limit(&self) -> GasAmount {
+        GasAmount::from(
+            SWEEP_BASE_GAS
+                .saturating_add(SWEEP_GAS_PER_DEPOSIT.saturating_mul(self.deposits.len() as u64)),
```
gregorydemay: 🤖 **`gas_limit()` is linear in `deposits.len()`, but the on-chain work is `items × tokens`.**

`CkSweeperAttested.sol:74` passes the *whole* `tokens` array to every item, and `sweepErc20` (line 60) loops `balanceOf` over each token — one external call per `(item, token)` pair. `tokens` is the deduplicated union of the batch (`sweep.rs:363`) and `sweep_targets_iter` applies no token grouping, so a batch routinely spans several ERC-20s.

A 20-deposit batch over 5 distinct tokens performs ~100 balance lookups plus per-pair overhead against a flat `60_000 + 20 × 80_000 = 1.66M` budget. The constant is documented as "deliberately above the ~42'000 measured for a batch of 20", but that measurement does not say how many distinct tokens it covered — and a linear model cannot bound a quadratic cost.

Since the same doc notes "an underestimate wastes the whole transaction", and the deposits are marked `swept_by` at enqueue, an out-of-gas revert is never retried. Either bound the token count per sweep, or scale on `deposits.len() × distinct_tokens` — and count `authorizations.len()` separately, since only delegating deposits pay per-auth gas.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -299,6 +306,53 @@ impl AutomaticDeposits {
...
+            let entry = self
+                .sweep
+                .get_mut(&request)
+                .unwrap_or_else(|| panic!("BUG: {request:?} is not queued for sweeping"));
+            assert_eq!(
+                entry.swept_by, None,
+                "BUG: {request:?} was already taken by sweep {:?}",
+                entry.swept_by
+            );
+            entry.swept_by = Some(sweep_id);
```
gregorydemay, line 348: 🤖 **`swept_by` is set here and never cleared anywhere in the tree — a sweep that never lands strands its deposits permanently.**

`grep swept_by` finds only this write and the read at line 313: no reset, no entry removal, on this branch or any descendant. Meanwhile `max_transaction_fee` is frozen at enqueue from `to_price(gas_limit()).max_transaction_fee()`, and both `create_transaction` and the `GuaranteeEthAmount` resubmission strategy enforce that same cap.

So a sustained base-fee rise above ~2× leaves the request rescheduling indefinitely (`sweep.rs:186`) while its deposits stay marked taken; an out-of-gas revert lands in the same state. Either way the funds become invisible to the sweeper with no operator recovery path.

A terminally-failed sweep needs to release its deposits back to the queue.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
...
+        attestation,
+        authorization,
+    } in prepared
+    {
+        let account = target.account();
+        let delegating = authorization.is_some();
+        if let Some(authorization) = authorization {
+            authorizations.push(authorization);
+        }
+        items.push(SweepItem {
```
gregorydemay: 🤖 **A multi-token account contributes duplicate `SweepItem`s and duplicate nonce-0 authorizations to the same transaction.**

`items` and `authorizations` are built one per `(account, token)` target, but `deposit_derivation_path` takes only `(schema, account)` — no token. So all of an account's tokens share one deposit address.

With USDC and USDT queued for account `A` (supported: `MAX_TOKENS_PER_ACCOUNT = 5`; the fixture at `state/automatic_deposits/tests.rs:826` is documented as exactly "One account with two funded tokens at the same deposit address"), `is_delegated(A)` is false for both and `delegating` is computed per index without being updated inside the batch. Result:

- Two threshold-ECDSA signatures spent on the identical `Authorization { nonce: 0 }`, both pushed into `SweepRequest::authorizations`. On chain the first bumps the authority nonce to 1 and the second is skipped as a nonce mismatch — while still costing its per-auth intrinsic gas.
- Two identical `SweepItem`s in the call data, so `sweepErc20Batch` walks the same address twice over the whole token list.

Dedupe by account when building `items` and `authorizations`, keeping `deposits` per `(account, token)` for the queue bookkeeping.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
...
+            })
+        })
+        .collect()
+}
+
+/// The attestation of each request, in request order, `None` where it could not be signed. An
+/// attestation is signed once per address and reused by every later sweep — it names the account
+/// and the helper, neither of which a sweep changes — so only the ones the minter does not have
+/// yet cost a signature, and each is recorded the moment it exists.
+async fn sign_attestations_batch(
```
gregorydemay: 🤖 **The same attestation digest is signed once per queued token, defeating the reuse the previous commit in this stack adds.**

The reuse cache is seeded for every request *before* any signing happens, so duplicate requests are all cache misses.

`attestation_request(account)` takes no token, so an account with 5 queued tokens yields 5 *identical* `AttestationRequest`s. All 5 read `None` from `s.attestation(...)` in the initial `read_state`, so `join_all` fires 5 concurrent `sign_attestation` calls on one digest and emits 5 `AttestedDepositAddress` events, each `record_attestation` insert overwriting the last.

That is 5 threshold-ECDSA signatures where 1 suffices — directly undoing "reuse a deposit address' attestation instead of re-signing it" within a single batch. Deduplicate the request list before signing, then fan the result back out to the targets that share it.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
...
+        destination: sweeper_contract,
+        amount: Wei::ZERO,
+        data: encode_sweep_erc20_batch(&items, &tokens),
+        max_transaction_fee: Wei::ZERO,
+        created_at: ic_cdk::api::time(),
+        authorizations,
+        deposits,
+    };
+    // TODO(DEFI-2933): gate this on the gas the sweeper address has actually been prepaid.
+    let Some(gas_fee_estimate) = lazy_refresh_gas_fee_estimate().await else {
```
gregorydemay: 🤖 **The gas estimate is fetched *after* every signature has been spent.**

With 20 queued deposits, `prepare_deposits` spends up to 20 attestation + 20 authorization threshold-ECDSA signatures, and only then does this line run. On `None` (RPC failure, or the `RefreshGasFeeEstimate` guard being held) the whole sweep is dropped: the attestations survive in state, but every authorization is discarded and must be re-signed next tick.

Nothing about the estimate depends on the signing — only the final `to_price(request.gas_limit())` needs the deposit count. Hoisting this `await` above `prepare_deposits` (line 326) makes the failure free.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
...
+/// Enqueue a single sweep of everything the balance scan has queued, up to
+/// [`MAX_DEPOSITS_PER_SWEEP`] deposits.
+///
+/// Each deposit contributes an ownership attestation, and each address that is not delegated yet
+/// also contributes an EIP-7702 authorization — so the first sweep of an address rides a type-`0x04`
+/// transaction and later ones a plain type-`0x02`. An authorization pins the address' nonce, which
+/// is `0` for a deposit EOA: incoming ERC-20 transfers never touch it, and only one of its own
+/// authorizations can advance it.
+pub async fn enqueue_batched_sweep() {
+    let _guard = match TimerGuard::new(TaskType::SweeperSend) {
```
gregorydemay: 🤖 **This takes `TaskType::SweeperSend` — the same guard `process_sweeper_transactions` holds (`sweep.rs:53`), contradicting the PR description.**

The description says "Enqueuing takes its own task guard, so it cannot deadlock against the sending task it feeds", but `TaskType` (`state.rs:978`) has no enqueue variant in this PR. There is no deadlock, but the two tasks mutually exclude each other: `process_sweeper_transactions` holds the guard across many RPC round trips and reschedules itself while any request is pending, so a backlogged pipeline can persistently drop enqueue ticks — and conversely enqueue holds it across two ECDSA batches plus a fee refresh.

`TaskType::SweeperEnqueue` already exists on the descendant branch `sweep-without-waiting`, so the fix is written — it just is not in this PR while the PR text claims it is. If the mutual exclusion is actually wanted (it does protect the `next_sweep_id` read at line 370 across the later await), say so in the doc instead.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -246,6 +256,47 @@ pub struct SweepRequest {
...
+    /// The deposits this sweep moves, in the order `data` names them. Recording them makes the
+    /// request self-describing: the sweep queue and the set of delegated addresses are both
+    /// reconstructible from the log, without decoding call data.
+    #[n(7)]
+    pub deposits: Vec<SweptDeposit>,
+}
+
+/// One `(account, token)` deposit a sweep moves, and the address it moves it from.
+#[derive(Clone, Eq, PartialEq, Debug, Decode, Encode)]
+pub struct SweptDeposit {
```
gregorydemay, line 300: 🤖 **This reimplements `crate::cbor::account`.**

The owner + optional-32-byte-subaccount split and the `account()` accessor duplicate the codec the base branch added in `cbor/account.rs`, which `AttestationRequest` already uses. Two representations of the same knowledge drift independently — a future change to how the minter encodes an `Account` updates `cbor/account.rs` and `AttestationRequest` but silently misses this one.

```rust
#[cbor(n(0), with = "crate::cbor::account")]
pub account: Account,
```

also removes the accessor. This is free to do now: `AcceptedSweepRequest::deposits` is introduced in this PR and is not yet deployed, so changing the CBOR layout breaks no replay.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
...
+/// deposit left without either signature drops out of the sweep rather than sinking it, since
+/// nothing the others contribute depends on it.
+async fn prepare_deposits(
+    targets: &[SweepTarget],
+    requests: &[AttestationRequest],
+    chain_id: u64,
+    sweeper_contract: Address,
+) -> Vec<PreparedDeposit> {
+    let attestations = sign_attestations_batch(requests).await;
+    let delegating: Vec<usize> = targets
```
gregorydemay: 🤖 **Three position-aligned collections plus an index-keyed map, on an invariant the types don't enforce.**

`prepare_deposits` threads `targets`, `requests` and `attestations` by position, plus a `BTreeMap<usize, _>` and a `Vec<usize>` used as a set. `attestations[*index]` / `attestations[index]` panic-index on a length invariant that holds only because the *caller* built `requests` by mapping over `targets` — any future reordering or filtering inside `sign_attestations_batch` turns that into an out-of-bounds trap rather than a compile error. `delegating.contains(&index)` (line 434) is also a linear scan of a `Vec` used as a set.

Carrying one struct per deposit (target + request + optional attestation) through the stages removes all the indexing, both four-argument signatures, and the `match` on a bool at line 434 — and it is the natural place to fix the two duplication findings above, since dedup becomes a `BTreeMap<Account, _>` rather than index bookkeeping.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -299,6 +306,53 @@ impl AutomaticDeposits {
         }
     }
 
+    /// The queued deposits no sweep has taken yet, oldest `(account, token)` key first.
```
gregorydemay: 🤖 **"oldest ... first" describes key order, not age.**

This iterates a `BTreeMap` in key order — lexicographic by principal then token address — which is unrelated to when the deposit was queued. `SweepEntry` carries no queue timestamp, so age ordering isn't available.

Worth being precise here because the ordering also feeds which deposits a `take(MAX_DEPOSITS_PER_SWEEP)` batch picks. Either say "in `(account, token)` key order", or record an enqueue time if fairness is actually wanted.

rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -260,3 +271,275 @@ async fn finalize_transactions_batch(sender: Address) {
         }
     }
 }
+
+/// Enqueue a single sweep of everything the balance scan has queued, up to
+/// [`MAX_DEPOSITS_PER_SWEEP`] deposits.
+///
+/// Each deposit contributes an ownership attestation, and each address that is not delegated yet
+/// also contributes an EIP-7702 authorization — so the first sweep of an address rides a type-`0x04`
+/// transaction and later ones a plain type-`0x02`. An authorization pins the address' nonce, which
```
gregorydemay: 🤖 **This paragraph is duplicated verbatim on `sign_authorizations_batch` (~line 498).**

"An authorization pins the address' nonce, which is `0` for a deposit EOA: incoming ERC-20 transfers never touch it, and only one of its own authorizations can advance it."

Two copies of a safety-critical invariant drift when the reasoning is revised, leaving the reader no way to tell which is current. Keep it on `sign_authorizations_batch`, where the nonce is actually set, and refer to it from here.

Note the last clause — "only one of its own authorizations can advance it" — is precisely the case the duplicate-authorization comment above is about.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -212,6 +213,15 @@ impl SweepId {
...
 }
 
+/// Gas a sweep transaction needs beyond its per-deposit cost: the 21'000 intrinsic cost, the batch
+/// call's own overhead and the first, most expensive write to the minter's token balance.
+const SWEEP_BASE_GAS: u64 = 60_000;
+
+/// Gas each additional deposit in a batch needs: its authorization, the delegated call, the
+/// approval and the transfer. Deliberately above the ~42'000 measured for a batch of 20, since
+/// unspent gas is refunded whereas an underestimate wastes the whole transaction.
+const SWEEP_GAS_PER_DEPOSIT: u64 = 80_000;
```
gregorydemay: 🤖 **Nit: these are bare `u64` where the constant they replace was typed.**

`SWEEP_TRANSACTION_GAS_LIMIT: GasAmount = GasAmount::new(100_000)`, deleted in this PR, carried its unit at the constant. These two need a `GasAmount::from` plus a `self.deposits.len() as u64` cast (line 297), and bare gas `u64`s can be mixed with unrelated quantities — wei, counts — without a compile error, which is what the `GasAmount` newtype exists to prevent.

Declaring both as `GasAmount` consts, or at least converting the length with `u64::try_from`, keeps the units checked and drops the `as`.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+        created_at: ic_cdk::api::time(),
+        authorizations: batch.authorizations,
+        deposits: batch.deposits,
+    };
+    request.max_transaction_fee = gas_fee_estimate
+        .clone()
+        .to_price(request.gas_limit())
+        .max_transaction_fee();
+    let ceiling = read_state(max_sweep_transaction_fee);
+    if request.max_transaction_fee > ceiling {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Deferred to the prepaid-gas gate tracked in DEFI-2933 (the `TODO` on `enqueue_batched_sweep`). With the provision accounting landing on top of this stack, the right check becomes: refuse a sweep whose `amount + max_transaction_fee` exceeds `sweeper_balance_lower_bound()`, which nets out prior provisions. Until then the failure mode is a transient wait, not a wedge — an unfundable transaction stays queued, earlier sweeps refund their unspent ceilings at finalization, and the funding task tops the sweeper back up. Gating also needs the e2e harness to seed the bound through a funding rather than `anvil.set_balance`, which is why it is its own change.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+/// the tuple's intrinsic gas — 12'500 for a skipped one, which [`SweepRequest::gas_limit`] budgets
+/// for every address.
+///
+/// The one thing this forecloses is **re-delegation**: an address delegated to an earlier sweeper
+/// contract holds nonce >= 1, so a nonce-0 tuple naming the new one is skipped and its code keeps
+/// pointing at the old contract. Its sweeps still execute, through that old delegate, which is
+/// harmless while the helper it calls is unchanged and reverts if it is not. Supporting
+/// re-delegation would mean reading the authority's real nonce for exactly those addresses — the
+/// ones whose stored authorization names a different delegate, which is itself the signal that the
+/// address is delegated to something else.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Accepted as a documented limitation — the trade-off spelled out on `delegation_authorization`: nonce-0 tuples buy order-independence and zero delegation state, at the price of foreclosing re-delegation. Replacing the helper or delegate is an operator-driven configuration change, and the sketched fix (read the real nonce only for the addresses whose stored authorization names a different delegate) is non-breaking to add when needed. No funds are lost meanwhile: the reverting sweeps drop their deposits and the balances stay at the addresses.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+                        );
+                    }
+                    mutate_state(|s| {
+                        process_event(
+                            s,
+                            EventType::FinalizedSweeperTransaction {
+                                sweep_id,
+                                transaction_receipt,
+                            },
+                        );
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed by #11342, now the top of this stack: an accepted sweep provisions `amount + max_transaction_fee` against the balance bound and gets back what it did not need at finalization — the unpaid fee, plus the value if it failed. Provisioning at acceptance keeps the bound erring low while sweeps are in flight; the counters stay out of `cumulative_spent` (R14) and the bound floors at zero. That PR is #11329 integrated per the hand-off on DEFI-2933.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed by the restructure: delegation tracking is gone entirely. Every sweep authorizes every address it touches at nonce 0, which is correct in any ordering of the carrying transactions — the reasoning lives on `delegation_authorization`.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed: `gas_limit()` now budgets distinct addresses × distinct tokens for both balance checks and transfers, plus one authorization per address — and the enqueue side sends one token per sweep, so the product degenerates to the address count in practice. Pinned by a test where more pairs from fewer deposits must cost more.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: `SweepBatch` dedups items and authorizations by account (`should_name_a_multi_token_account_once_in_the_call_data`, `should_authorize_every_address_it_sweeps_exactly_once`). An address swept twice in one round — once per token — deliberately carries a tuple in each transaction; the skipped one costs only its intrinsic gas, which the limit budgets.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: `attestation_requests` is keyed by account, and a batch holds a single token, so a digest is signed at most once per batch; stored attestations cover every later one.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: the estimate is fetched before anything is signed, with a comment saying exactly why.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: `TaskType::SweeperEnqueue` is introduced in this PR and the enqueue takes it.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed: `SweptDeposit` now carries `account: Account` encoded with the ledger type's own codec; the owner/subaccount split and the accessor are gone.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: `PreparedDeposit { target, attestation, authorization }` carries each deposit through the stages — no index bookkeeping, no positional invariants.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Fixed: the doc now says "(account, token) key order" and states outright that nothing records when a deposit was queued.

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Fixed: the nonce reasoning lives once, on `delegation_authorization`; `sign_authorizations_batch` refers to it.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed: typed `GasAmount` constants with checked arithmetic, and the length converted with `u64::try_from`.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Accepted as wait-and-retry, the same discipline the withdrawal pipeline applies. Failed and successful sweeps now release their deposits in the `FinalizedSweeperTransaction` transition; a sweep priced out by a fee spike re-prices every tick under its frozen ceiling (~2× headroom at enqueue) and proceeds once the spike subsides — holding its deposits meanwhile is deliberate, since they must not be offered to a second sweep while one owns them. The wait is visible as `in_flight` in the sweep-queue gauge. A permanent fee-regime shift beyond the ceiling strands withdrawals just the same, and is an operator situation in either pipeline.

publish/canisters/BUILD.bazel
```diff
@@ -89,7 +89,8 @@ CANISTERS_MAX_SIZE_COMPRESSED_E5_BYTES = {
     # Size is currently at 923_414 bytes.
     "ic-ckbtc-minter.wasm.gz": 10,
     # Size when constraint addded: 1_230_630 bytes
-    "ic-cketh-minter.wasm.gz": 13,
+    # Size is currently at 1_309_035 bytes.
+    "ic-cketh-minter.wasm.gz": 14,
```
gregorydemay, line 93: why is this change needed? This PR doesn't change any library

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -29,10 +46,18 @@ use crate::{
...
-pub(crate) const SWEEP_TRANSACTION_GAS_LIMIT: GasAmount = GasAmount::new(100_000);
+/// Deposits swept in one transaction, which with one token per sweep is also the addresses it
+/// touches.
+///
+/// This is what bounds a sweep's gas, but only because the batch carries a single token: the
+/// delegate applies its whole token list to every address, so a one-token batch pays one balance
+/// check and at most one transfer per address and [`SweepRequest::gas_limit`] stays linear in the
+/// batch — ~3.4M gas at twenty deposits. A batch spanning several tokens would grow as their
+/// product instead, which is why [`sweep_batches_by_token`] never builds one.
+const MAX_DEPOSITS_PER_SWEEP: usize = 20;
```
gregorydemay: Doesn't the test in `deposit_from_cex_demo.rs` show that the sweet spot is 10?

gregorydemay: 🤖 It does — done in 8311a24d2d, the cap is now ten.

The measured attested scenarios put a deposit at 98'000 gas alone, 60'943 in a batch of ten and 59'645 in a batch of twenty: the second doubling saves ~2%, which is one transaction's intrinsic 21'000 spread over the extra deposits. Against that, twenty doubles how many deposits a single revert drops — `sweepErc20Batch` has no per-item error handling, so a failed sweep discards its whole batch until DEFI-2981 — and doubles what one sweep prepays against the sweeper's low-water mark, which is what makes `max_sweep_transaction_fee` bind for a minter configured with a smaller minimum withdrawal amount.

The constant now carries that reasoning instead of just the shape argument. It also makes the e2e a *full*-batch test: it already deposits ten per token, so each token's sweep is now exactly at the cap.

publish/canisters/BUILD.bazel
gregorydemay: 🤖 This PR adds library code: the enqueue implementation in `rs/ethereum/cketh/minter/src/sweep/mod.rs` (~490 productive lines — `enqueue_batched_sweep`, `sweep_batches_by_token`, `enqueue_token_sweep`, `SweepBatch`, `prepare_deposits`, the two signing batches), plus the `TaskType` variant and the timer in `main.rs`, all compiled into the canister. What it does *not* change is any third-party dependency — no `Cargo.toml`, `Cargo.lock` or `crate.spec` diff anywhere in the stack — so the growth is our own code and the async state machines it monomorphizes.

Measured on `//publish/canisters:ic-cketh-minter.wasm.gz`:

| tree | compressed bytes |
| --- | --- |
| #11336, this PR's base | 1_241_743 |
| this PR | 1_309_035 |
| cap before | 1_300_000 (13) |
| cap after | 1_400_000 (14) |

So the layer adds ~67 KB and is where the 1.3 MB tripwire trips; without the bump this PR and everything above it are red, which is what CI showed before the commit. The new cap leaves ~6% headroom at the stack top, in line with the neighbouring entries.

Happy to split it into a one-line `build(cketh)` PR between #11336 and this one if you would rather keep this PR to feature code — I kept it here because a cap bump with no accompanying growth reads as unmotivated, and the layer that pays the cost is the one that can record the measured size.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -260,3 +304,424 @@ async fn finalize_transactions_batch(sender: Address) {
...
+/// several tokens are supported — and its gas grows as addresses × tokens rather than with the
+/// deposits it moves. Grouping keeps that product equal to the addresses, which is what lets
+/// [`MAX_DEPOSITS_PER_SWEEP`] bound a sweep's gas at all. It costs one transaction's 21'000
+/// intrinsic gas per token, which a multi-token sweep pays anyway in the cold storage write each
+/// token's balance at the minter needs.
+///
+/// Every deposit gets one sweep and no more. `sweepErc20Batch` has no per-item error handling, so
+/// whatever makes a sweep revert — one address blacklisted for the token, say — reverts every batch
+/// that deposit is in; a failed sweep therefore drops its whole batch from the queue rather than
+/// leaving anything behind to retry (DEFI-2981).
```
gregorydemay, line 388: No Rust docs on private method, remove

gregorydemay:

PR #11259, opened 2026-08-21
test(cketh): teach the live harness to deploy and drive the sweep contracts

gregorydemay:
## Why

- The live harness could scan balances on anvil, but nothing test-side could run a sweep against a real chain: no sweep contracts on the node, and no way to read what a sweep did.

## What

- `LiveSetup::new_sweep` deploys the real deposit helper and the attested sweeper delegate on the harness' anvil before installing the minter, so the minter starts knowing both, and pre-funds the sweeper address with ETH (a shortcut for the funding pipeline, which has its own end-to-end test).
- The anvil client learns to read what a sweep did: an address' code (for delegation designators), the transactions an address sent with their type, status, and gas, and to set an ETH balance.
- anvil now mines on an interval, so block-based scan backoffs and the `finalized` tag advance on their own instead of freezing whenever no test transaction lands.

### DIFF
--- a/rs/ethereum/cketh/test_utils/src/anvil.rs
+++ b/rs/ethereum/cketh/test_utils/src/anvil.rs
@@ -19,6 +19,14 @@ use std::time::{Duration, Instant};
 /// `eth_sendTransaction` without any local signing.
 pub const DEV_ACCOUNT: &str = "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266";
 
+/// The chain the harness pretends to be, matching the `EthereumNetwork::Mainnet` the fixture
+/// installs the minter with.
+pub const CHAIN_ID: u64 = 1;
+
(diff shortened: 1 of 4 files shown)


rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -281,7 +281,7 @@ async fn finalize_transactions_batch(sender: Address) {
 /// is `0` for a deposit EOA: incoming ERC-20 transfers never touch it, and only one of its own
 /// authorizations can advance it.
 pub async fn enqueue_batched_sweep() {
-    let _guard = match TimerGuard::new(TaskType::SweeperSend) {
+    let _guard = match TimerGuard::new(TaskType::SweeperEnqueue) {
```
gregorydemay: why does this change? Why is SweeperEnqueue added?

gregorydemay:

PR #11261, opened 2026-08-21
test(cketh): credit twenty CEX deposits through one EIP-7702 sweep per token

gregorydemay:
## Why

- Every piece of the attested sweep flow was tested on its own. Nothing had shown the whole path working against a real chain.

## What

- An end-to-end test on anvil: twenty users — each with its own principal and subaccount, ten holding USDC and ten USDT — are funded by plain CEX-style transfers to their deposit addresses, and all twenty are credited on the ledgers with no further user action.
- The minter sweeps them in exactly two EIP-7702 transactions, one per token, paying gas from the pre-funded sweeper address.
- Assertions read the chain, not just the ledgers: transaction type and success, the delegation designators installed on every deposit address, the addresses swept empty, the minter's main address holding all the funds, and the sweeper paying out of its own balance.
- Failures name the sweep pipeline stage the run stopped at, counted off the minter's durable audit events rather than its rolling canister log.

[DEFI-2926]: https://dfinity.atlassian.net/browse/DEFI-2926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11264, opened 2026-08-21
perf(cketh): sweep one token per transaction instead of the whole batch

gregorydemay:
## Why

- A batch sweeping several tokens paid for the cross-product: every address in the batch was tried against every token, and most pairs held nothing.

## What

- The enqueue path groups the queue by token, so one sweep transaction moves one token.
- Each batch is still capped at 20 deposits.

## Measurements

- 20 deposits of two tokens now cost 623'360 and 623'687 gas in two transactions, about 62'350 per deposit.
- The demo contract's 10-deposit figure is 609'431, so the minter is within 2.3% of it, the difference being non-zero subaccount bytes in the call data.

[DEFI-2980]: https://dfinity.atlassian.net/browse/DEFI-2980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gregorydemay:
🤖 Absorbed into #11258 and closing.

An adversarial review of #11258 established that per-token sweeping is not a performance improvement but a correctness prerequisite. Once the gas limit has to budget a transfer for every `(address, token)` pair the delegate may move — `sweepErc20` transfers any balance it finds, including residue the batch never queued — a mixed batch is unbudgetable: 20 deposits over 20 tokens is 400 pairs, ~51M gas, beyond a block's limit and far beyond what the sweeper address can prepay. Grouping by token keeps the product equal to the addresses, which is what lets `MAX_DEPOSITS_PER_SWEEP` bound a sweep's gas at all.

So the grouping now lands in #11258, where batches are first built:

- The production change and its rationale moved there, restated as a correctness argument.
- The end-to-end assertions and the `Anvil::transactions_of` helper moved to #11261, where the sweep e2e test lives, as `test(cketh): check that a mixed queue is swept one token per transaction`.

Nothing is lost; the stack is one PR shorter.

PR #11265, opened 2026-08-21
feat(system-tests): boot nested VMs on the local backend

basvandijk:
> Stacked on #11270, which gives the local backend's ic-gateway a resolvable domain and a certificate that nodes trust. Review that one first.

## Problem

Nested system-tests (`rs/tests/nested`) drive the whole node deployment chain: SetupOS installs HostOS onto a blank disk, HostOS boots a nested GuestOS VM, and that GuestOS registers with the NNS. All five non-bare-metal targets were pinned to Farm because `setup_and_start_nested_vms` was not supported on the local backend.

## What was missing

Three things in the driver:

- the install target was built as a `DiskImage::Url` whose `_URL` variable `run_systest.sh` never sets under the local backend;
- `setup_and_start_nested_vms` needed its `Local` arm;
- `NESTED_CONFIG_IMAGE_PATH` claimed `.zst` for a file nothing compresses, which the local backend would have run `unzstd` over.

The fourth — the nested node's NNS URL — is what #11270 is for. A node that is not in the registry yet cannot pass the replicas' firewall, so it registers through the ic-gateway; that needed the gateway to be reachable by name and trusted, which it now is on both backends. `nested_nns_url` is consequently backend-agnostic.

Also: `pad_to_request_alignment` now refuses a GPT image rather than silently displacing its backup header, and the nested VM drops from 32 GiB to 16 GiB — the floor, since the driver takes 8 GiB off and HostOS then takes 4 more, leaving the GuestOS the same 4 GiB every other node VM gets. That reduction applies on Farm too; all 20 Farm nested targets pass with it.

## The blocker: SetupOS halts because chrony can never synchronise

`check-ntp.sh` is the one SetupOS check *not* gated on `ic.setupos.run_checks`, so `setupos-disable-checks` does not skip it. It waits 60s for `NTPSynchronized` and then halts the installation *forever* — `log_and_halt_installation_on_error` ends in `sleep infinity`.

Serving NTP to the group ourselves does not fix it, and the first two commits here do exactly that before it is reverted. chrony's `authselectmode` defaults to `mix`, and `sources.c` then marks the *authenticated* sources `require`d as soon as both authenticated and unauthenticated sources are configured — precisely what `chrony.conf` does in pairing fourteen `nts` servers with two plain pools. Sources configured from names that never resolve still count. So the clock can only ever be synchronised from one of those fourteen NTS servers; the driver's SNTP server was polled every 64s for a whole run and never selected, with chronyd saying so in the node journals:

```
Can't synchronise: no required source in selectable sources
```

Reproduced against chronyd 4.8 directly: given that config shape it polls a reachable plain server for 100s without selecting it, and selects it in 4s once `authselectmode ignore` is added. Satisfying the check for real would mean serving NTS under one of those fourteen hostnames with a certificate the guest trusts, which no test harness can arrange.

So the check is gated instead, like the four that also need connectivity, and the reasoning is recorded on `check_ntp`. Nothing is lost: every test VM takes its clock from the hypervisor via `-rtc base=utc`.

## Scope

`registration` and both `*_upgrade_smoke_test` targets are un-pinned, along with the `_head_nns` variants `system_test_nns` generates from them — `backend` is per target definition. All six were run locally before dropping the pins.

The two `*_from_latest_release_to_current` targets stay pinned: they install from a prebuilt *mainnet* SetupOS image, of which `setupos-disable-checks` can only rewrite the grub boot args, so their `check-ntp.sh` predates this gate.

One thing for reviewers to weigh: this takes the non-manual nested `_local` targets from two to six, and `system_tests.bzl:342` reserves CPU only — `ceil(cpus / cpus_oversubscription_factor)` = 9 of the 27 these declare — with no accounting for memory or disk. Three can therefore run concurrently on a 32-core runner, for ~84 GiB of nominal VM memory and ~60 GiB of qcow2 overlays. `cpus_oversubscription_factor = 1` would serialise them if that turns out to matter; no target in the repo sets it today, so this is left at the default.


rs/tests/driver/src/driver/bootstrap.rs
```diff
@@ -458,52 +458,70 @@ pub fn setup_and_start_nested_vms(
...
+                        "usb-storage",
+                        vec![setupos_image, config_image_spec],
+                    )?;
+                    t_farm.start_vm(&t_group_name, &vm_name)?;
+                }
+                SystemTestBackend::Local => {
+                    let backend = LocalBackend::from_test_env(&t_env)?;
+                    let DiskImage::Local {
+                        path: setupos_image,
+                    } = get_setupos_disk_image(&t_env)
```
basvandijk: `get_setupos_disk_image` is only called here. Consider inlining it and special casing it for the `SystemTestBackend::Local` since we're already in that branch.

rs/tests/nested/src/lib.rs
```diff
@@ -21,6 +23,23 @@ pub mod util;
...
+/// which is what fixes the floor at 16 GiB:
+///
+/// * the driver subtracts `HOSTOS_MEMORY_RESERVED_GIB` (8) for HostOS itself when
+///   it writes `dev_vm_resources.memory` into the SetupOS `deployment.json`, and
+/// * HostOS then subtracts `UPGRADE_VM_MEMORY_GIB` (4) for the upgrade VM
+///   (`rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs`).
+///
+/// So 16 GiB leaves the nested GuestOS 4 GiB, the same as every other node VM.
+/// At 12 GiB it would be sized 0 GiB and libvirt would refuse the domain.
+const LOCAL_BACKEND_NESTED_VM_MEMORY: AmountOfMemoryKiB = AmountOfMemoryKiB::new(16 * 1024 * 1024);
```
basvandijk: Why not have the same 16 GiB default for Farm as well?

rs/tests/driver/src/driver/bootstrap.rs
basvandijk: Done in a05a7cd802 — inlined and the getter deleted.

You're right that the indirection wasn't paying for itself: it built a `DiskImage` so it could have a Farm arm that this call site can never reach, and then made the caller destructure it with a `let ... else { bail!(...) }` for a case that can't happen. Now it just reads `ENV_DEPS__SETUPOS_DISK_IMG_PATH` in the branch, which is what the existing `Local` arms in `resource.rs` and `universal_vm.rs` already do.

I kept `get_empty_disk_image` — that one is genuinely called for both backends, since `get_resource_request_for_nested_nodes` needs a `DiskImage` either way.

rs/tests/nested/src/lib.rs
basvandijk: Agreed — done in a05a7cd802. `HOSTOS_MEMORY_KIB_PER_VM` is now 16 GiB for both backends and the override is gone, so that's a deletion rather than an addition.

I went looking for a reason to keep 32 GiB on Farm and couldn't find one. It gave the nested GuestOS 20 GiB (32 minus the driver's 8 for HostOS, minus HostOS's 4 for the upgrade VM), where every *other* node VM in a test gets `DEFAULT_MEMORY_KIB_PER_VM` = 4 GiB — on Farm too, since that constant isn't backend-specific. A nested node runs the same replica as its unnested peers, so there was nothing for the extra 16 GiB to do; the figure looks inherited from production sizing, where a real node has 512 GiB.

Blast radius is small: the only other consumer of the default is `//rs/tests/node:launch_single_host`. `rs/tests/nested/nns_recovery` sets both vcpus and memory explicitly (40 / 48 GiB via `NNS_RECOVERY_VM_RESOURCE_OVERRIDES`), so it is unaffected.

And it is verified rather than assumed: because this PR carries `CI_ALL_BAZEL_TARGETS`, the previous run exercised *all* the Farm nested targets at 32 GiB — `registration`, both upgrade smoke tests and both `from_latest_release` ones, all green — so this push re-runs the same set at 4 GiB of GuestOS.

One drive-by: two `Default` impls in `rs/ic_os/config` documented themselves as `(HOSTOS_VCPUS_PER_VM / 2, HOSTOS_MEMORY_KIB_PER_VM / 2)`. That was already wrong for vcpus (8 / 2 = 4, not the 16 they set), and halving the memory constant would have made it wronger, so I replaced the formula with what those defaults actually are.

rs/tests/driver/src/driver/nested.rs
```diff
@@ -31,7 +31,11 @@ use url::Url;
 
 pub const NESTED_VMS_DIR: &str = "nested_vms";
 pub const NESTED_VM_PATH: &str = "vm.json";
-pub const NESTED_CONFIG_IMAGE_PATH: &str = "config.img.zst";
+// `build-setupos-config-image.sh` writes a raw FAT image here -- despite what an
+// earlier `.zst` in this name suggested, nothing compresses it. The Local backend
+// dispatches on the extension when it attaches the image
+// (`LocalBackend::extract_image`), so the name has to be honest.
```
basvandijk: Don't mention the previous situation. Just remove the whole comment.
```suggestion
```

rs/tests/driver/src/driver/resource.rs
```diff
@@ -33,7 +32,20 @@ const DEFAULT_VM_RESOURCES: VmResources = VmResources {
...
+/// [`HOSTOS_MEMORY_RESERVED_GIB`] (8) off for HostOS when it writes
+/// `dev_vm_resources.memory` into the SetupOS `deployment.json`, and HostOS then
+/// takes another 4 off for the upgrade VM (`UPGRADE_VM_MEMORY_GIB` in
+/// `rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs`).
+///
+/// So 16 GiB leaves the GuestOS [`DEFAULT_MEMORY_KIB_PER_VM`] (4 GiB), the same
+/// as every other node VM in a test: a nested node runs the same replica, and
+/// sizing it larger than its unnested peers only made nested tests harder to fit
+/// on one host. 12 GiB would be too little -- the GuestOS would come out at 0 GiB
+/// and libvirt would refuse the domain. Tests that need more say so with
```
basvandijk: Why mention the 12 GiB? Just remove that sentence:
```suggestion
/// on one host. Tests that need more say so with
```

rs/tests/driver/src/driver/bootstrap.rs
```diff
@@ -531,6 +550,48 @@ pub fn setup_and_start_nested_vms(
...
+        Url::parse("http://localhost:8080").unwrap()
+    };
+    match SystemTestBackend::read_attribute(env) {
+        SystemTestBackend::Farm => env
+            .get_deployed_ic_gateway(IC_GATEWAY_VM_NAME)
+            .map(|gateway| gateway.get_public_url())
+            .unwrap_or_else(|_| dummy()),
+        SystemTestBackend::Local => env
+            .topology_snapshot()
+            .root_subnet()
```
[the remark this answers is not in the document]
basvandijk: Good catch, and it was worse than reported — there are two panic sites, not one.
`env.topology_snapshot()` panics with "Could not save topology snapshot because No
no-name Internet Computer" before `root_subnet()` is even reached, and only then
does `root_subnet_id()` panic with "root subnet id is not set". The
`unwrap_or_else(dummy)` I had covered just `nodes().next() == None`, i.e. the one
case that never happens. The Farm arm has no such hole because
`get_deployed_ic_gateway` returns an `Err`, which is exactly why the asymmetry was
easy to miss.

Fixed by adding `TopologySnapshot::try_root_subnet`, which returns `None` rather
than panicking — shaped like the neighbouring `unassigned_nodes_config` — with
`root_subnet` delegating to it so the two cannot drift. `nested_nns_url` then
threads `safe_topology_snapshot().ok()` through and falls back to `dummy()`.

Verified by actually running the target, since being `manual` it gets no CI
signal and that is how this slipped through in the first place:

```
Task setup  PASSED in 15.06s
[setup:bootstrap.rs:574] No NNS node or gateway found, using dummy URL
```

SetupOS then installs and the VM reboots into HostOS on its own
(`BdsDxe: loading Boot0004 "IC-OS" ... Booting linux from system A`), where before
it panicked in setup within seconds.

One thing I am deliberately *not* fixing, so it is on the record rather than
silently shipped: `launch_single_host_local` still fails after that, for a reason
that predates this PR. `await_can_login_as_admin_via_ssh` takes no timeout
parameter and uses the shared 500s `SSH_RETRY_TIMEOUT`. Here the setup phase is
only 15s — there is no IC to build — so the entire ~5.5 min SetupOS install has to
fit inside that same budget, and HostOS finishes booting with about 1.5 min left,
which is not enough for sshd to come up. Giving it a longer wait needs either a
new trait method or a `retry_with_msg!` around the call, in a target this PR does
not otherwise touch and whose 500s default is shared with ten other call sites.
Happy to do it here if you would rather it went in together.

ic-os/components/setupos/check-ntp.sh
```diff
@@ -39,7 +48,11 @@ function set_hwclock_utc() {
 
 main() {
     log_start "$(basename $0)"
-    check_ntp
+    if check_cmdline_var ic.setupos.run_checks; then
+        check_ntp
+    else
+        echo "* NTP synchronization check skipped by request via kernel command line"
```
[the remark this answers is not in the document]
basvandijk: You are right about the mismatch, but it is stale: the description had already been
rewritten before this review landed. The review was requested at 15:49:17Z and
submitted at 15:53:55Z, and the body was replaced in between — so this was reviewed
against the pre-rewrite text.

Checked the live body against each claim listed here:

| claim | live body |
| --- | --- |
| backend starts an SNTP server | gone |
| seeds the pool DNS names | gone |
| supports prebuilt SetupOS images | gone |
| adds SNTP reply-builder tests | gone |

`serve_ntp_task` now appears only where the body explains it was tried and
reverted, and the description leads with the reason it cannot work: chrony's
`authselectmode` defaults to `mix`, which marks the *authenticated* sources
`require`d as soon as both authenticated and unauthenticated sources are
configured, so the clock can only ever be synchronised from one of the fourteen
NTS servers in `chrony.conf`. Names that never resolve still count. Reproduced
against chronyd 4.8: with that config shape it polls a reachable plain server for
100s without selecting it, and selects it in 4s once `authselectmode ignore` is
added.

So the implemented approach is exactly what you describe — skip the check on
images whose boot args we rewrite — and that is what the description and the
`check_ntp` comment now say, including why the prebuilt-image targets stay pinned.
No change needed here.

dfinityianblenke (approved):

rs/tests/driver/src/driver/bootstrap.rs
```diff
@@ -458,52 +458,71 @@ pub fn setup_and_start_nested_vms(
     let logger = env.logger();
     info!(logger, "Setting up nested VM(s) ...");
 
-    let ic_gateway_url = env
-        .get_deployed_ic_gateway(IC_GATEWAY_VM_NAME)
-        .map(|v| v.get_public_url())
-        .unwrap_or_else(|_| {
-            info!(logger, "No gateway found, using dummy URL");
-            url::Url::parse("http://localhost:8080").unwrap()
-        });
```
basvandijk, line 467: Considering that the `nested_nns_url` function does the same as the removed code except for some additional logging. I suggest to keep the original code to keep the diff smaller.

basvandijk: Done in 766a4fa29e — the function is deleted and the original code restored verbatim, `ic_gateway_url` name and all. The only remaining `bootstrap.rs` diff in `setup_and_start_nested_vms` is now the Farm/Local match block.

ic-os/components/setupos/check-ntp.sh
```diff
@@ -8,6 +8,15 @@ PATH="/sbin:/bin:/usr/sbin:/usr/bin"
 
 source /opt/ic/bin/functions.sh
 
+# Needs a reachable NTP source, and specifically one of the NTS servers:
```
frankdavid: nit: This comment feels very sloppy, I don't think we really need all this detail here, maybe just add a comment in the else branch below saying that NTP not available in tests.

rs/ic_os/config/tool/src/setupos/deployment_json.rs
```diff
@@ -42,8 +42,9 @@ pub struct VmResources {
 }
 
 impl Default for VmResources {
-    /// These currently match the defaults for nested tests on Farm:
-    /// (`HOSTOS_VCPUS_PER_VM / 2`, `HOSTOS_MEMORY_KIB_PER_VM / 2`)
+    /// Fallback for a `deployment.json` that omits the field. Nested system-tests
```
frankdavid: Slop, the original comments were more useful.

rs/ic_os/config/types/src/lib.rs
```diff
@@ -194,8 +194,9 @@ pub struct HostOSDevSettings {
 }
 
 impl Default for HostOSDevSettings {
-    /// These currently match the defaults for nested tests on Farm:
-    /// (`HOSTOS_VCPUS_PER_VM / 2`, `HOSTOS_MEMORY_KIB_PER_VM / 2`)
+    /// Fallback for a `deployment.json` that omits the field. Nested system-tests
```
frankdavid: Same

rs/tests/driver/src/driver/bootstrap.rs
```diff
@@ -485,25 +485,50 @@ pub fn setup_and_start_nested_vms(
...
-                    "usb-storage",
-                    vec![setupos_image, config_image_spec],
-                )?;
-                t_farm.start_vm(&t_group_name, &vm_name)?;
-                Ok(())
+                return setup_baremetal_instance(&t_env, &node, &config_image)
+                    .context("Setting up baremetal instance failed");
             }
+
+            // Attach the SetupOS image and the config image, then boot. The VM's
```
frankdavid: Not too useful comment

rs/tests/driver/src/driver/ic_images.rs
```diff
@@ -301,6 +301,39 @@ pub fn get_setupos_img_sha256() -> String {
     std::env::var(env).unwrap_or_else(|_| panic!("Failed to read '{env}'"))
 }
 
+/// Get the all-zero disk image a nested node uses as its primary disk from the
+/// environment.
+///
+/// This is the *install target*: the nested VM boots SetupOS from an attached
```
frankdavid: This second part of the comment is bloat, consider just adding a 1-liner in the match mentioning that farm uses a hosted image, whereas local uses a local path (but also this is kinda obvious from the code).

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -1828,28 +1828,71 @@ fn extract_image(src: &Path, dst: &Path, logger: &Logger) -> Result<()> {
 /// (`rs/tests/driver/assets/create-universal-vm-config-image.sh`), and prebuilt
 /// UVM config images need not be aligned either, so pad them here. They carry a
 /// FAT filesystem that records its own length and ignores trailing bytes, so the
-/// padding is inert. Only these extra (config) disks are padded; boot disks are
-/// already block aligned and may use a GPT backup header at the last sector,
-/// which padding would displace.
+/// padding is inert.
+///
+/// A GPT-partitioned image is a different matter: its backup header lives in the
```
frankdavid: Just say unpadded GPT images will return an error, the error message below already explains the reason.

frankdavid: Btw, we could just move the GPT backup header if it's necessary.

rs/tests/driver/src/driver/local_backend.rs
```diff
@@ -1828,28 +1828,71 @@ fn extract_image(src: &Path, dst: &Path, logger: &Logger) -> Result<()> {
...
+    use std::io::{Read, Seek, SeekFrom};
+
+    /// LBA size assumed by the GPT signature's location. Every image the backend
+    /// attaches uses 512-byte logical sectors.
+    const LBA_SIZE: u64 = 512;
+    const GPT_SIGNATURE: &[u8; 8] = b"EFI PART";
+
+    let mut file = std::fs::File::open(path)
+        .with_context(|| format!("opening {} to look for a GPT", path.display()))?;
+    if file.seek(SeekFrom::Start(LBA_SIZE)).is_err() {
```
frankdavid: optional nit: [read_exact_at](https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#method.read_exact_at)

rs/tests/driver/src/driver/resource.rs
```diff
@@ -33,7 +32,19 @@ const DEFAULT_VM_RESOURCES: VmResources = VmResources {
 };
 
 pub const HOSTOS_VCPUS_PER_VM: NrOfVCPUs = NrOfVCPUs::new(8);
-pub const HOSTOS_MEMORY_KIB_PER_VM: AmountOfMemoryKiB = AmountOfMemoryKiB::new(33554432); // 32GiB
+/// Memory for a nested VM, out of which the nested GuestOS gets what is left
```
frankdavid: Consider just writing

```
/// Allocates 16 GiB so the nested GuestOS retains the standard 4 GiB [DEFAULT_MEMORY_KIB_PER_VM].
/// This accounts for two 4 GiB host deductions: [HOSTOS_MEMORY_RESERVED_GIB] and [UPGRADE_VM_MEMORY_GIB].
```

Or maybe even better if we actually want to express this relationship by writing:
`const HOSTOS_MEMORY_KIB_PER_VM = DEFAULT_MEMORY_KIB_PER_VM + HOSTOS_MEMORY_RESERVED_GIB + UPGRADE_VM_MEMORY_GIB`

But it may require some unwanted cross-dependencies to get these constants which may not be worth it.

rs/tests/driver/src/driver/resource.rs
```diff
@@ -235,15 +246,10 @@ pub fn get_resource_request_for_nested_nodes(
...
-    let empty_disk_img_url = get_empty_disk_img_url()?;
-    let empty_disk_img_sha256 = get_empty_disk_img_sha256()?;
-
-    // Add a VM request for each node.
-    let mut res_req = ResourceRequest::new(DiskImage::Url {
-        ic_os_image: true,
-        url: empty_disk_img_url,
-        sha256: empty_disk_img_sha256,
-    });
+    // The primary disk is the install target: the nested VM boots SetupOS from an
```
frankdavid: Unnecessary comment

rs/tests/nested/BUILD.bazel
```diff
@@ -133,13 +115,14 @@ system_test_nns(
...
     name = "guestos_upgrade_from_latest_release_to_current",
     # TODO: support this test on the local backend (drop `backend = "farm"`).
-    # This is a nested test (SetupOS -> HostOS -> GuestOS). The local backend
-    # does not yet implement nested VMs: `setup_and_start_nested_vms`
-    # (rs/tests/driver/src/driver/bootstrap.rs) only has a Farm code path and
-    # calls into Farm directly to upload the config image, attach the SetupOS
-    # disk image and start the nested VM. Supporting this requires adding a
-    # `SystemTestBackend::Local` branch that boots the nested VM via the local
-    # QEMU-based backend.
+    # The local backend does boot nested VMs now -- see the `registration` target,
```
frankdavid: I don't really get what this comment means. Does it mean we use "farm" backend because the local backend does not support nested VMs? Why is guestos_upgrade_from_latest_release_to_current different from registration? And is the comment about NTP relevant here?

rs/tests/nested/BUILD.bazel
```diff
@@ -218,13 +192,14 @@ system_test_nns(
...
     name = "hostos_upgrade_from_latest_release_to_current",
     # TODO: support this test on the local backend (drop `backend = "farm"`).
-    # This is a nested test (SetupOS -> HostOS -> GuestOS). The local backend does
-    # not yet implement nested VMs: `setup_and_start_nested_vms`
-    # (rs/tests/driver/src/driver/bootstrap.rs) only has a Farm code path and calls
-    # into Farm directly to upload the config image, attach the SetupOS disk image
-    # and start the nested VM. Supporting this requires adding a
-    # `SystemTestBackend::Local` branch that boots the nested VM via the local
-    # QEMU-based backend.
+    # The local backend does boot nested VMs now -- see the `registration` target,
```
frankdavid: Same

frankdavid (changes requested):

ic-os/components/setupos/check-ntp.sh
basvandijk: Done in e5bc593417 — the block comment is gone; a two-line note in the else branch now says NTP isn't available in test environments and test VMs take their clock from the hypervisor.

rs/ic_os/config/tool/src/setupos/deployment_json.rs
basvandijk: Trimmed to one line in e5bc593417. I didn't restore the original text because it had gone stale: with today's constants its formula gives (4, 8 GiB), not the (16, 16) set here, and nested tests always write explicit values anyway.

rs/ic_os/config/types/src/lib.rs
basvandijk: Same, done in e5bc593417.

rs/tests/driver/src/driver/bootstrap.rs
basvandijk: Removed in e5bc593417.

rs/tests/driver/src/driver/ic_images.rs
basvandijk: Done in e5bc593417 — dropped the second paragraph, kept a single sentence noting SetupOS installs HostOS onto this disk.

rs/tests/driver/src/driver/local_backend.rs
basvandijk: Done in e5bc593417 — now a one-liner; the error message carries the reason. On moving the backup header: doable (rewrite both headers and their CRCs), but the only GPT image we attach is 1 MiB-aligned, so the bail guards a case that shouldn't occur. Happy to add relocation if you'd prefer.

rs/tests/driver/src/driver/local_backend.rs
basvandijk: Done in e5bc593417 — this also drops the seek-failed branch, since read_exact_at reports a too-short file as UnexpectedEof.

rs/tests/driver/src/driver/resource.rs
basvandijk: Took your suggested comment in e5bc593417, with the numbers fixed: the deductions are 8 GiB (HOSTOS_MEMORY_RESERVED_GIB) + 4 GiB (UPGRADE_VM_MEMORY_GIB), not two 4s. Skipped the const-arithmetic form since UPGRADE_VM_MEMORY_GIB lives in guest_vm_runner, which the driver doesn't depend on.

rs/tests/driver/src/driver/resource.rs
basvandijk: Removed in e5bc593417.

rs/tests/nested/BUILD.bazel
basvandijk: Rewritten in e5bc593417. To answer directly: the local backend does support nested VMs now; this target differs from `registration` in installing from a prebuilt *mainnet* SetupOS image, whose check-ntp.sh still halts the install when NTP is unreachable — which on the local backend it always is. That is the whole reason for the pin, and the new five-line comment says exactly that.

rs/tests/nested/BUILD.bazel
basvandijk: Same rewrite in e5bc593417.

frankdavid (approved):

lwshang (approved):

PR #11266, opened 2026-08-21
test(nns): Allow using production versions of NNS canisters in standard_engine_replica_version_test

daniel-wong-dfinity-org-twin:
Reverts dfinity/ic#11248

PR 11248 was needed, because relaxed validation rules had not yet reached production. The relaxed rules were in regards to replica versions used by StandardEngineReplicaVersion. Because the production code was more strict, the test would fail with

```
Proposal invalid because of new_replica_version_id is not a 40-character hexadecimal string (it was \"c818497b5cd2bbc60af597f7aadc036640ed0463-test\")
```

This will be merged after the relaxed rules go into production. In the meantime, the `_head_nns` version of this test still works, and is valuable, because it tells us if a version that we might try to put into production might be broken.


pierugo-dfinity:
We should be able to open this now

pierugo-dfinity (approved):
I guess the next step is to drop the feature flag entirely?

daniel-wong-dfinity-org-twin:
> I guess the next step is to drop the feature flag entirely?

Yes, I am fine with removing it (or them), because the has been successfully used. I assume you mean the flag in Registry that allows blank replica_version_id?

Also, it would not be super straightforward to roll back right now, because there are now SubnetRecords with blank replica_version_id...

pierugo-dfinity:
> I assume you mean the flag in Registry that allows blank replica_version_id

Yes, [this guy](https://github.com/dfinity/ic/blob/b8bd732c637a7f5e6f313fb0e8ff13587c054ed0/rs/registry/canister/src/flags.rs#L17)

PR #11269, opened 2026-08-22
fix(cmc): retry ICP burns that fail after cycles were already delivered

Rachit2323:
## The problem

When you send ICP to get cycles, CMC does two things: give you the cycles, then burn your ICP. If the burn fails, it just logs it and forgets. You keep the cycles, but your ICP never actually gets burned.
No retry, ever.

## What I did

If the burn fails, save it instead of forgetting it. Try again every heartbeat until it works. Never repeat the "give you cycles" part again - only retry the burn.

## Testing

Added tests for the new save/retry logic. All existing tests still pass.


rs/nns/cmc/src/main.rs
```diff
@@ -295,6 +304,158 @@ impl StateV2 {
...
+            recent_icp_xdr_rates,
+            cycles_per_xdr,
+            base_cycles_limit,
+            subnet_rental_cycles_limit,
+            base_limiter,
+            subnet_rental_canister_limiter,
+            total_cycles_minted,
+            blocks_notified,
+            last_purged_notification,
+            pending_burns: BTreeMap::new(),
```
daniel-wong-dfinity-org-twin: Put this at the end and separate using a blank line.

[Do not bury the lede][lede]. The lede here is "pending_burns field was added".

[lede]: https://github.com/dfinity/ic/blob/1614219ff6f7abcbd6ab78df77d9158c5f95fb9e/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L49

rs/nns/cmc/src/main.rs
```diff
@@ -295,6 +304,158 @@ impl StateV2 {
...
+    /// during canister creation, a subnet without a special type will be picked
+    /// at random as no special requirements were provided.
+    ///
+    /// Each subnet can be assigned to at most one type and cannot be a default
+    /// or an authorized subnet.
+    pub subnet_types_to_subnets: Option<BTreeMap<String, BTreeSet<SubnetId>>>,
+
+    /// This is used to ensure that only one exchange rate update is being performed at a time from heartbeat.
+    pub update_exchange_rate_canister_state: Option<UpdateExchangeRateState>,
+}
```
daniel-wong-dfinity-org-twin, line 412: Just for the record: I have verified that the only change vs. StateV2 is the addition of pending_burns.

rs/nns/cmc/src/main.rs
```diff
@@ -295,6 +304,158 @@ impl StateV2 {
     }
 }
 
+#[derive(Clone, Eq, PartialEq, Debug, CandidType, Deserialize, Serialize)]
+pub struct StateV3 {
```
daniel-wong-dfinity-org-twin, line 314: **RANT**

I know that the following issue long predates this PR, but I still have to say it: this is a massive amount of code to achieve something super simple, and super common: add one field.

rs/nns/cmc/src/main.rs
```diff
@@ -295,6 +304,158 @@ impl StateV2 {
...
+    // status, set it to Processing. Then, we can proceed with calling the other
+    // canister (i.e. ledger). Once that comes back, we update the block's
+    // status. This avoids using the same ICP to perform multiple operations.
+    pub blocks_notified: BTreeMap<BlockIndex, NotificationStatus>,
+    // The status of blocks not new than this is ambiguous. This is because we
+    // must bound how much memory we use; in particular, blocks_notified must
+    // not grow without bound.
+    pub last_purged_notification: BlockIndex,
+
+    pub pending_burns: BTreeMap<BlockIndex, PendingBurn>,
```
daniel-wong-dfinity-org-twin: Please, explain. E.g.

> When an ICP burn fails, it gets enqueued here for later retry. Retries are performed by retry_pending_burns, which gets called during heartbeat. When a retry is launched, the entry remains, and is not modified. Therefore, there is no way to tell from this collection whether a pending burn is currently in flight, or waiting to be retried. This is keyed on BlockIndex so that earlier operations are retried first.

If you take my status suggestion, some of this comment needs to be omitted.

daniel-wong-dfinity-org-twin:
```suggestion
    pub block_index_to_pending_burn: BTreeMap<BlockIndex, PendingBurn>,
```

`widgets` is for a non-associative collection of `Widget`s. But a map is not merely a collection of its values. A map also has KEYS! And what makes a map really valuable is that it ASSOCIATES a key with a value. It is better to think of it as a collection of KEY VALUE PAIRS (where each key is unique). If the name does not describe the relationship encoded in these pairs, I think a VITAL part of the story has been left out.

rs/nns/cmc/src/main.rs
```diff
@@ -314,14 +475,16 @@ impl State {
                      Please upgrade to a hotfix instead."
                 ));
             }
+            Ordering::Less if stored_state_version == StateV2::state_version() => {
```
daniel-wong-dfinity-org-twin: Do not use `if` in match arms. Compose `match` and `if`. That is, use `if` within `match` arms.

rs/nns/cmc/src/main.rs
```diff
@@ -356,7 +519,7 @@ impl State {
     }
 }
 
-impl Default for State {
+impl Default for StateV2 {
```
daniel-wong-dfinity-org-twin, line 537: I know this issue (long) predates this PR, but I have to ask...

Are these impls just for testing, or what? I really cannot imagine how it would NOT be a mistake to use these in production.

rs/nns/cmc/src/main.rs
daniel-wong-dfinity-org-twin: Is block index just a breadcrumb? Like, it's not actually strictly necessary to complete the burn, right? This is maybe one of the facts you can mention in doc comment.

rs/nns/cmc/src/main.rs
```diff
@@ -1213,7 +1382,8 @@ async fn notify_top_up(
     match maybe_early_result {
         Some(result) => result,
         None => {
-            let result = process_top_up(canister_id, from, amount, limiter_to_use).await;
+            let result =
+                process_top_up(block_index, canister_id, from, amount, limiter_to_use).await;
```
daniel-wong-dfinity-org-twin: The number of arguments is starting to get out of hand. This is how code "organically growed" to become a giant mess, which are difficult to debug, fix, add features, and just generally further develop.

I'm not really sure what alternative to suggest though. Maybe, some of these arguments could be grouped? E.g. block_index, from, and amount are part of the ICP transaction that funds this operation, right?

One thing is, block_index is just a breadcrumb, not essential to the operation. Therefore, it should go towards the end of the list. At the beginning, it tries to steal the lime light.

The name of this function (which I know, predates this PR) really doesn't help. First of all, "process" is literally a most generic word you can use in the name of a function. ALL functions do SOME kind of "processing".

Furthermore, it's really not clear how the behavior of process_top_up differs from notify_top_up itself. I mean, doesn't notify_top_up ALSO "process top ups"?? In fact, it does!

Ditto elsewhere.

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
     }
 }
 
+fn record_pending_burn(
+    source_block_index: BlockIndex,
+    from_subaccount: Subaccount,
+    amount: Tokens,
+) {
+    with_state_mut(|state| {
+        state.pending_burns.insert(
```
daniel-wong-dfinity-org-twin: What if there is already an entry? Might make sense to log that case.

Note that insert clobbers. Therefore, it needs to be used with care. When it clobbers, it returns the removed value.

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+            source_block_index,
+            PendingBurn {
+                from_subaccount,
+                amount,
+            },
+        );
+    });
+}
+
+fn clear_pending_burn(source_block_index: BlockIndex) {
```
daniel-wong-dfinity-org-twin: Do not use the word `clear` in this case. That sounds like you are deleting many entries. Furthermore, `*Map` already has a word for this operation: `remove`. There is no reason to "get creative" with the name. Crib from `*Map`. Cribbing is boring. [Boring code is good][bor].

[bor]: https://github.com/dfinity/ic/blob/a1919f7f96dd4fd50ded1403510325b24f461819/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L73

> Good artists copy. Great artists steal.
> --Pablo Picasso

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+    });
+}
+
+async fn burn_or_record_pending(
+    source_block_index: BlockIndex,
+    from_subaccount: Subaccount,
+    amount: Tokens,
+) {
+    match burn_and_log(from_subaccount, amount).await {
+        Ok(()) => clear_pending_burn(source_block_index),
```
daniel-wong-dfinity-org-twin: Is removing necessary here? If so, pending burns is ambiguous. The possible meanings are

1. Burning is in flight.
2. We tried to burn, but it didn't go through. Therefore, we'll try again later.

Seems like a super important distinction.

If removing is not necessary, but rather, just "clean up, just in case", then we really should [not just let "unexpectedly present" pass quietly][rug]. If burning is idempotent (I believe this is possible, but not the default), then "unexpectedly present" is not that terrible.

[rug]: https://github.com/dfinity/ic/blob/5a0c6ee3850557aa380f7887c9d2ac113f16f68e/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L76

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+        Err(err) => {
+            record_pending_burn(source_block_index, from_subaccount, amount);
```
daniel-wong-dfinity-org-twin: See, this line is telling me that `pending_burns` records that we need to try again later. So, which is it? Is it in flight, or do we need to try again later?

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+            print(format!(
```
daniel-wong-dfinity-org-twin: Why not println??

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+                "[cycles] Burn for notification block {source_block_index} will be retried: {err}"
+            ));
+        }
+    }
+}
+
+struct PendingBurnsGuard;
```
daniel-wong-dfinity-org-twin: Do not re-invent the wheel: https://github.com/dfinity/ic/blob/1e3614fb537f7cccf97abde0e5975a5d134d2fd5/rs/nervous_system/lock/src/tests.rs#L8

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+
+impl PendingBurnsGuard {
+    fn try_acquire() -> Option<Self> {
+        if PENDING_BURNS_IN_PROGRESS.with(|in_progress| in_progress.replace(true)) {
```
daniel-wong-dfinity-org-twin: It is [not good that this is thousands of lines away][locality]. However, taking my other feedback would make this moot.

[locality]: https://github.com/dfinity/ic/blob/d9fd141cb4cbcc72b81d09a5603a128897448849/rs/nervous_system/STANDARDS_AND_CONVENTIONS.md?plain=1#L187

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+}
+
+async fn retry_pending_burns() {
+    let Some(_guard) = PendingBurnsGuard::try_acquire() else {
+        return;
+    };
+
+    let pending_burns = with_state(|state| {
+        state
+            .pending_burns
```
daniel-wong-dfinity-org-twin: There are two things here with the same name, but they are pretty different:

1. batch
2. ALL burns that still need to be done.

I mean, the distinction is not SUPER important, but making the distinction is also pretty easy. Just name this new thing `pending_burn_batch` or something.

rs/nns/cmc/src/main.rs
daniel-wong-dfinity-org-twin: The name should mention ICP (e.g. `pending_icp_burns`), because in this ecosystem, there are many resources that can be burned, mainly ICP and cycles, but also ck tokens, and SNS tokens.

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+    for (source_block_index, pending_burn) in pending_burns {
+        match burn_and_log(pending_burn.from_subaccount, pending_burn.amount).await {
+            Ok(()) => {
+                clear_pending_burn(source_block_index);
+                print(format!(
+                    "[cycles] Pending burn for notification block {source_block_index} succeeded."
+                ));
+            }
+            Err(err) => print(format!(
+                "[cycles] Pending burn for notification block {source_block_index} failed: {err}"
```
daniel-wong-dfinity-org-twin: We should have a way to detect when the same burn keeps failing. Maybe that can be done in another PR, and here, just leave a TODO. Here's what we might do later:

normally, `pending_burns` would be empty, right? Let's have a Prometheus metric with a name like `cycles_minting_pending_burns` (whose value is simply `pending_burns.len()`).

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+        state
+            .pending_burns
+            .iter()
+            .take(MAX_PENDING_BURNS_PER_HEARTBEAT)
+            .map(|(block_index, pending_burn)| (*block_index, pending_burn.clone()))
+            .collect::<Vec<_>>()
+    });
+
+    for (source_block_index, pending_burn) in pending_burns {
+        match burn_and_log(pending_burn.from_subaccount, pending_burn.amount).await {
```
daniel-wong-dfinity-org-twin: If you are going to serially .await each burn, I'm a bit confused why batch? Please, comment. I guess it's just to limit how long this function takes?

daniel-wong-dfinity-org-twin: How about

```suggestion
        match pending_burn.retry().await {
```

?

daniel-wong-dfinity-org-twin: Aha! I was right: block index is not actually essential. It is just breadcrumbs. Breadcrumbs are good, ofc. It is just a little bit surprising that pending_burns is keyed on block index. It's not bad; just a little surprising.

rs/nns/cmc/src/main.rs
```diff
@@ -2015,22 +2206,107 @@ async fn process_top_up(
...
+    }
+}
+
+impl Drop for PendingBurnsGuard {
+    fn drop(&mut self) {
+        PENDING_BURNS_IN_PROGRESS.with(|in_progress| in_progress.set(false));
+    }
+}
+
+async fn retry_pending_burns() {
```
daniel-wong-dfinity-org-twin: probably would be good if doc comment said that this returns right away if another call is in progress.

rs/nns/cmc/src/main.rs
```diff
@@ -157,6 +160,12 @@ pub struct NotMeaningfulMemo {
     refund_block_index: Option<BlockIndex>,
 }
 
+#[derive(Clone, Eq, PartialEq, Debug, CandidType, Deserialize, Serialize)]
+pub struct PendingBurn {
```
daniel-wong-dfinity-org-twin: I think what you could do is give each pending burn a status. Something like

```
enum PendingBurnStatus {
    WaitingToBeRetried,
    InFlight,
}
```

This is maybe YAGNI though. I thought of this as an alternative to locking. Ofc, this is not mutually exclusive vs. locking.

rs/nns/cmc/src/main.rs
daniel-wong-dfinity-org-twin: I don't hate this, but it since canisters are single-threaded, locking is highly unusual. Therefore, when locks are used in canisters, there should probably be a special reason, and that reason should be documented. I guess the special thing here is that retry_pending_burns is part of an async heartbeat, and so, it is possible that while one retry_pending_burns is running, it gets called again, and for the sake of simplicity, we'd rather have the second call return immediately, i.e. without looking at pending_burns at all.

rs/nns/cmc/src/main.rs
```diff
@@ -2443,6 +2728,11 @@ fn encode_metrics(w: &mut ic_metrics_encoder::MetricsEncoder<Vec<u8>>) -> std::i
             state.blocks_notified.len() as f64,
             "Number of notifications stored in the cache.",
         )?;
+        w.encode_gauge(
+            "cmc_pending_burn_count",
```
daniel-wong-dfinity-org-twin: Claude is telling me that cmc_pending_burns would be more in line with what the rest of the world does.

rs/nns/cmc/src/main.rs
```diff
@@ -2612,6 +2909,79 @@ mod tests {
...
+    fn test_record_and_clear_pending_burn() {
+        STATE.with(|state| state.replace(Some(State::default())));
+
+        let block_index = 42;
+        let from_subaccount = Subaccount([1; 32]);
+        let amount = Tokens::from_e8s(123_456);
+        record_pending_burn(block_index, from_subaccount, amount);
+
+        assert_eq!(
+            with_state(|state| state.pending_burns.get(&block_index).cloned()),
```
daniel-wong-dfinity-org-twin: better to get the whole collection, not just one entry.

rs/nns/cmc/src/main.rs
```diff
@@ -2612,6 +2909,79 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_record_and_clear_pending_burn() {
+        STATE.with(|state| state.replace(Some(State::default())));
+
+        let block_index = 42;
+        let from_subaccount = Subaccount([1; 32]);
+        let amount = Tokens::from_e8s(123_456);
+        record_pending_burn(block_index, from_subaccount, amount);
```
daniel-wong-dfinity-org-twin: How about we insert two things? That way, we can see that removing one thing leaves other(s) in place?

rs/nns/cmc/src/main.rs
```diff
@@ -2612,6 +2909,79 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_burn_or_record_pending_records_burn_when_burn_fails() {
+        STATE.with(|state| state.replace(Some(State::default())));
+
+        let block_index = 42;
+        let from_subaccount = Subaccount([1; 32]);
+        let amount = Tokens::from_e8s(123_456);
+        futures::executor::block_on(burn_or_record_pending(block_index, from_subaccount, amount));
```
daniel-wong-dfinity-org-twin: I guess the burn fails due to state being blank? Seems like a fact worth commenting.

rs/nns/cmc/src/main.rs
```diff
@@ -2612,6 +2909,79 @@ mod tests {
...
+
+        assert_eq!(
+            with_state(|state| state.pending_burns.get(&block_index).cloned()),
+            Some(PendingBurn {
+                from_subaccount,
+                amount,
+            }),
+        );
+    }
+
```
daniel-wong-dfinity-org-twin, line 3024: Perhaps, in another PR, can we have a test that is more "end to end"? Like, suppose Alice wants some cycles, so she sends some ICP to the CMC. She gets her cycles, but CMC initially fails to burn the ICP for whatever reason, but it later retries, and the retry succeeds, and the pending burn count goes back down to 0. This would, probably require running a fake ledger or something so that we can force the first burn attempt to fail in order to see that a retry is attempted.

I mean, one of the more interesting parts of this PR is retry, but that doesn't seem to be tested, right??

rs/nns/cmc/src/main.rs
daniel-wong-dfinity-org-twin: "Pending" is ok, but it does not convey that it got to this state because the first attempt failed. Here are some other ideas:

* hung
* stalled
* unresolved
* unsettled

These more indicate that something bad happened, but we'll try to recover (by retrying). I kind of feel that recovery from some earlier issue is an interesting aspect of this, and would like the name to reflect.

daniel-wong-dfinity-org-twin: The name should mention ICP (e.g. `PendingIcpBurns`), because in this ecosystem, there are various resources that can be burned, mainly ICP and cycles, but also ck tokens, and SNS tokens.

rs/nns/cmc/src/main.rs
Rachit2323: lol yeah, agreed. can look at cleaning this up in a separate PR if useful

rs/nns/cmc/src/main.rs
Rachit2323: added

rs/nns/cmc/src/main.rs
Rachit2323: it's used in init as well, not test-only. that one's fine .

rs/nns/cmc/src/main.rs
Rachit2323: yeah that's right, added a comment

rs/nns/cmc/src/main.rs
Rachit2323: renamed to StalledIcpBurn

Rachit2323: good idea, maybe later. locking's enough for now

Rachit2323: went with stalled, renamed it throughout

rs/nns/cmc/src/main.rs
Rachit2323: done — went with block_index_to_stalled_icp_burn after the naming discussion below

Rachit2323: yep, added that to the doc comment

rs/nns/cmc/src/main.rs
Rachit2323: fixed, composed match + if instead of the guard

rs/nns/cmc/src/main.rs
Rachit2323: moved block_index to the end since it's just a breadcrumb. agree with the rest but don't have a clean fix either — let's do the bigger cleanup in a separate PR, like the other one

rs/nns/cmc/src/main.rs
Rachit2323: added a warning log for that case

rs/nns/cmc/src/main.rs
Rachit2323: switched to ic_nervous_system_lock

Rachit2323: added a doc comment on retry_stalled_icp_burns calling that out

rs/nns/cmc/src/main.rs
Rachit2323: moot now that it's using the lock crate

rs/nns/cmc/src/main.rs
Rachit2323: added a retry() method on the struct

rs/nns/cmc/src/main.rs
Rachit2323: added a second entry so removal is actually proven

rs/nns/cmc/src/main.rs
Rachit2323: updated to assert on the whole map

rs/nns/cmc/src/main.rs
Rachit2323: let's track that as a follow-up

rs/nns/cmc/src/main.rs
Rachit2323: added

daniel-wong-dfinity-org-twin:
Hi Rachit2323,

FYI: I will be on vacation until Sep 21. I still have a few more hours in the office, but I assume I won't be around when you are ready for another round of review. Seems like we can pick this up when I come back. In the meantime, if you need more support, consider reaching out to @bjoernek . Thank you again for this contribution.

Daniel

PR #11274, opened 2026-08-24
fix: lower the consumed cycles gauge on a full refund

mraszyk:
`SystemState::observe_consumed_cycles_with_use_case()` maintains two kinds of metrics: the gauges (`consumed_cycles` and `consumed_cycles_by_use_cases`), which are raised by the prepayment and lowered by the refund; and the monotonic counters (`consumed_cycles_by_use_cases_as_counters`), which for the refundable use cases (`Instructions` and `RequestAndResponseTransmission`) are only bumped at refund time, by `prepayment - refund`.

The early return conflated the two: for a refund equal to its prepayment it skipped the update altogether, so the gauges kept the prepayment even though nothing was consumed in the end. The inflated value also outlives the canister, as `CanisterManager::delete_canister()` folds `consumed_cycles_by_use_cases` into the subnet level metrics.

Skip only if there is nothing to record at all, i.e. if both the prepayment and the refund are zero. Non-refundable use cases are unaffected, as a zero prepayment implies a zero refund for them, and `NominalCycles` saturates on subtraction, so a refund against a freshly created entry cannot underflow.

Tests: the new `full_refund_resets_consumed_cycles` refunds a prepayment in full and checks that the balance and both gauges are back to where they started and that the counter did not move. It covers both refundable use cases under both cost schedules. The gauges are kept in nominal cycles, which the cost schedule does not affect, so a full refund under `Free` hits the same skipped update; it is in fact the more interesting case, as `Free` waives the real amount and hence leaves the metrics as the only observable of the prepayment. The assertion on the balance after the prepayment (which drops by the real amount under `Normal` and stays put under `Free`) pins down that difference. Each of the four combinations fails without the fix.

### DIFF
--- a/rs/replicated_state/src/canister_state/tests.rs
+++ b/rs/replicated_state/src/canister_state/tests.rs
@@ -921,6 +921,10 @@ fn full_refund_resets_consumed_cycles() {
         let mut system_state = CanisterStateFixture::new().canister_state.system_state;
         let use_case = T::cycles_use_case();
         let ctx = format!("{use_case:?} with {cost_schedule:?} cost schedule");
+        // An earlier prepayment for the same use case that is never refunded, so the
+        // gauges must fall back to it (rather than all the way down to zero) once the
+        // prepayment below is refunded in full.
+        let outstanding_cycles = CompoundCycles::<T>::new(Cycles::from(500_u128), cost_schedule);
         let cycles_to_consume = Cycles::from(1000_u128);
         let prepaid_cycles = CompoundCycles::<T>::new(cycles_to_consume, cost_schedule);
         // The cost schedule only waives the real amount charged to the balance;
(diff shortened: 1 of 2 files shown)


rs/replicated_state/src/canister_state/tests.rs
```diff
@@ -912,6 +912,76 @@ fn update_balance_and_consumed_cycles_by_use_case_correctly() {
...
+        let prepaid_cycles = CompoundCycles::<T>::new(cycles_to_consume, cost_schedule);
+        // The cost schedule only waives the real amount charged to the balance;
+        // the nominal amount recorded in the metrics is the same either way.
+        assert_eq!(
+            prepaid_cycles.nominal(),
+            NominalCycles::new(cycles_to_consume.get()),
+            "{ctx}"
+        );
+
+        system_state.consume_cycles(prepaid_cycles);
```
eichhorl, line 939: we could add one prepayment that isn't refunded

eichhorl (approved):

rs/replicated_state/src/canister_state/tests.rs
mraszyk: https://github.com/dfinity/ic/pull/11274/commits/594a0042e9965889e75d9cce1803669ee38fe692

PR #11277, opened 2026-08-24
feat: Do not route refunds to or from a cooling down subnet

mraszyk:
Refunds in the refund pool are only routed into a stream if neither the source subnet nor the destination subnet is cooling down; not even into the loopback stream. They are retained in the refund pool until neither subnet is cooling down anymore, rather than being dropped (which would lose their cycles).

Unlike the subnet's own output queues (which are exempt from the source side of the check, so that a cooling down subnet can still respond to the calls it has already accepted), refunds have no such exemption.

Skipped refunds are counted in the new `mr_cooling_down_skipped_refunds` counter, once per refund per round, mirroring
`mr_cooling_down_skipped_queues`.

### DIFF
--- a/rs/messaging/src/routing/stream_builder.rs
+++ b/rs/messaging/src/routing/stream_builder.rs
@@ -835,23 +835,22 @@ impl StreamBuilderImpl {
             match network_topology.route(refund.recipient().get()) {
                 Some(dst_subnet_id) => {
                     let is_loopback_stream = dst_subnet_id == self.subnet_id;
+                    let dst_subnet_topology = network_topology.subnets().get(&dst_subnet_id);
+                    let dst_subnet_type = dst_subnet_topology.map(|topology| topology.subnet_type);
 
                     // No refunds are routed while either this subnet (the source) or the
                     // destination subnet is cooling down; not even into the loopback stream.
                     // Retain them in the refund pool until neither subnet is cooling down
                     // anymore, rather than dropping them (which would lose their cycles).
-                    if own_subnet_is_cooling_down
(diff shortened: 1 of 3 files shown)


rs/messaging/src/routing/stream_builder.rs
eichhorl: Alternatively we could get the subnet topology once, and then derive both the subnet type and the cooling down flag from it

rs/messaging/src/routing/stream_builder.rs
```diff
@@ -806,8 +817,9 @@ impl StreamBuilderImpl {
 
     /// Routes up to `refund_limit` refunds per stream from `state` into `streams`.
     ///
-    /// Refunds that could not be routed due to reaching the per stream limit are
-    /// retained in `state`.
+    /// Refunds that could not be routed due to reaching the per stream limit, or
+    /// because this subnet or their destination subnet is cooling down, are retained
+    /// in `state`.
```
eichhorl, line 822: We could also record this in the docs of the `cooling_down` field

eichhorl (approved):

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11277/commits/16f92299cbaa9e444de98f5123c444b7ac3a83f4

rs/messaging/src/routing/stream_builder.rs
mraszyk: https://github.com/dfinity/ic/pull/11277/commits/84b69a07eed67113e622f1953bb40899121045b2

PR #11282, opened 2026-08-24
fix: Verify that a DKG dealing's config exists before checking the cache

eichhorl:
Otherwise we might accept blocks with dealings for DKG instances that have already completed.

### DIFF
--- a/rs/consensus/dkg/src/payload_validator.rs
+++ b/rs/consensus/dkg/src/payload_validator.rs
@@ -204,15 +204,16 @@ fn validate_dealings_payload(
     for message in &dealings.messages {
         metrics.with_label_values(&["total"]).inc();
 
-        // Skip the rest if already present in DKG pool
+        let Some(config) = configs.get(&message.content.dkg_id) else {
+            return Err(InvalidDkgPayloadReason::MissingDkgConfigForDealing.into());
+        };
+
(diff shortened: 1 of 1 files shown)


rs/consensus/dkg/src/payload_validator.rs
```diff
@@ -415,6 +416,47 @@ mod tests {
...
     }
 
+    #[test]
+    fn validate_dealings_payload_when_wrong_dkg_id_and_in_dkg_pool_fails_test() {
+        // Validation should fail if the block contains a dealing without a coresponding config,
+        // even if the dealing is present in the validated pool.
+        let dealing: ic_types::crypto::Signed<
+            DealingContent,
+            ic_types::signature::BasicSignature<DealingContent>,
+        > = fake_dkg_message(SUBNET_2, NODE_1);
```
pierugo-dfinity:
```suggestion
        let dealing = fake_dkg_message(SUBNET_2, NODE_1);
```
Is the hint necessary?

pierugo-dfinity (approved):

rs/consensus/dkg/src/payload_validator.rs
eichhorl: It is not 👍

PR #11283, opened 2026-08-24
feat: Respect query walltime limit

schneiderstefan:
The commit respects `maximum_query_call_seconds` in the replica. This is a follow-up on commit f84327316, that did the same for the query instructions limit.

This commit also clarifies some comments to no longer wrongly imply that the limit applies to non-composite queries.


rs/execution_environment/src/query_handler.rs
```diff
@@ -246,12 +246,16 @@ impl InternalHttpQueryHandler {
...
+        // The subnet's registry-configured `ResourceLimits` override the query limits:
+        // `maximum_query_instructions` bounds both the per-query and the composite-query-graph
+        // instruction limits, and `maximum_query_walltime_seconds` bounds the composite-query
+        // call-graph wall-clock time (the only query wall-clock limit; a single query is bounded
+        // by instructions). Each falls back to its own replica default when unset.
         let resource_limits = state.get_ref().resource_limits();
         let max_query_call_graph_instructions = resource_limits
             .maximum_query_instructions_or(self.config.max_query_call_graph_instructions);
+        let max_query_call_walltime =
+            resource_limits.maximum_query_walltime_seconds_or(self.config.max_query_call_walltime);
```
eichhorl, line 258: Should we add some tests?

rs/execution_environment/src/query_handler.rs
```diff
@@ -246,12 +246,16 @@ impl InternalHttpQueryHandler {
...
             },
         );
-        // The subnet's registry-configured `maximum_query_instructions` (`ResourceLimits`) overrides
-        // both the per-query and the composite-query-graph instruction limits; each falls back to
-        // its own replica default when unset.
+        // The subnet's registry-configured `ResourceLimits` override the query limits:
+        // `maximum_query_instructions` bounds both the per-query and the composite-query-graph
+        // instruction limits, and `maximum_query_walltime_seconds` bounds the composite-query
+        // call-graph wall-clock time (the only query wall-clock limit; a single query is bounded
+        // by instructions). Each falls back to its own replica default when unset.
```
eichhorl, line 253: Then could there be a single query that never replies?

eichhorl (approved):

rs/execution_environment/src/query_handler.rs
schneiderstefan: I thought about it, but ended up not including them. Unit tests that the value gets correctly set ended up just a bunch of non-statements that don't test anything. The other kind of tests that the walltime is respected is tricky, because of course they are not enforced very tightly. We just check occasionally if enough time has passed.

rs/execution_environment/src/query_handler.rs
schneiderstefan: In principle, though you'd simply run into the instructions budget first. I am not sure why a wallclock parameter was introduced for composite queries specifically, considering the same argument that you accrue too many instructions should also hold there.

That being said, it's somewhat easy to check against a wallclock for composite queries, simply check between messages. Timing out a single message while it's running in a sandbox is a harder thing to do.

daniel-wong-dfinity-org-twin (approved):

PR #11286, opened 2026-08-24
fix: cap the consumed cycles metrics at the cycles actually charged

mraszyk:
`SystemState::consume_cycles()` subtracts the real amount from the balances with saturating arithmetic, but reported the full nominal amount as consumed. Whenever the balances cannot cover the request, the difference is never removed from any balance, yet it still showed up in the canister's (and, transitively, the subnet's) consumed cycles metrics.

Compute the part that the balances cannot cover before draining them and report only the charged remainder, via the new `CompoundCycles::minus_uncharged()`, which reduces both the real and the nominal part. The two parts coincide under the normal cost schedule and some use cases under the free cost schedule; otherwise the real part is zero, so nothing can be left uncharged and this is a no-op.

`apply_ingress_induction_cycles_debit()` is the caller that legitimately runs into this: it may be called with a cycles balance that is smaller than the pending debit, after a cleanup callback, which is allowed to burn the balance below the debit and must always be able to succeed. As documented, the part of the debit that the balance cannot cover is then dropped, making some of the postponed ingress induction charges free -- and the dropped part was reported as consumed. It now passes only the part of the debit that the balance covers.

Regression tests cover a `consume_cycles()` request exceeding the balance, one exceeding the balance and the reserved balance together, and an ingress induction debit exceeding the balance after a cleanup callback.


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1042,8 +1042,14 @@ impl SystemState {
                 // some of the postponed charges free.
             }
         }
+        // Charge only the part of the debit that the balance can cover. The remaining
+        // debit is dropped, so it must not be reported as consumed either. This is the
+        // one caller that legitimately charges less than it requests, so it caps the
+        // amount here instead of leaving that to the defense in depth in
+        // `consume_cycles()`.
```
eichhorl: This comment would have to be updated if another caller is added, which seems unrealistic
```suggestion
        // Charge only the part of the debit that the balance can cover. The remaining
        // debit is dropped, so it must not be reported as consumed either.
```

rs/replicated_state/src/canister_state/tests.rs
```diff
@@ -993,6 +1049,61 @@ fn full_refund_resets_consumed_cycles() {
     }
 }
 
+#[test]
+fn consume_cycles_exceeding_balance_reports_only_the_charged_amount() {
+    let mut system_state = CanisterStateFixture::new().canister_state.system_state;
+    let cost_schedule = CanisterCyclesCostSchedule::Normal;
```
eichhorl: We could also add a test for the `Free` cost schedule

eichhorl (approved):

rs/replicated_state/src/canister_state/tests.rs
mraszyk: https://github.com/dfinity/ic/pull/11286/commits/2a14dc874af06d2dc1f61e75ccdd008a8cefadb8

PR #11288, opened 2026-08-24
fix: CON-1739 Validate consistency when receiving HTTP outcalls shares with a response body

eichhorl:
Validate that the response body is for the same callback ID as the surrounding share

Additionally, remove the unused canister ID field of the response struct.

Additionally, when finding a flexible result, derive the number of _unseen_ shares based on the number of _seen shares_ (not _seen signers_). This is more similar to what the validation side does, and prevents both sides from diverging if in the future, flexible outcalls might produce shares without a response attached.

### DIFF
--- a/rs/https_outcalls/consensus/src/pool_manager.rs
+++ b/rs/https_outcalls/consensus/src/pool_manager.rs
@@ -1888,7 +1888,6 @@ pub mod test {
                         reject_code: RejectCode::SysTransient,
                         message: "A transient error occurred.".to_string(),
                     }),
-                    ..empty_canister_http_response(0)
                 };
 
                 let response_metadata = CanisterHttpResponseReceipt {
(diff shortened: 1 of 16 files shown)


rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -561,6 +561,27 @@ impl CanisterHttpPoolManagerImpl {
...
+                                share.clone(),
+                                format!(
+                                    "Response is for request ID {} rather than the share's {}",
+                                    response.id,
+                                    share.content.id(),
+                                ),
+                            ));
+                        }
+
+                        if response.canister_id != context.request.sender {
```
pierugo-dfinity: I struggle to understand what the `canister_id` is used for (generally). I see it written to, but never really read?

pierugo-dfinity (approved):

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Nice! We already removed a few of these fields, seems like this is another one: https://github.com/dfinity/ic/pull/11288/changes/6b90f7bdd470a499fb82fba8d842b348d0806328

PR #11289, opened 2026-08-24
chore: Rename config

michael-weigelt:
The `dirty_page_overhead` is now applied for written and accessed pages, so this PR renames it to `page_overhead`.


alin-at-dfinity (approved):

schneiderstefan (approved):
LGTM. Could you add a description maybe? I assume this description will end up as the commit message like for regular PRs?

PR #11294, opened 2026-08-24
fix: Only observe consumed ECDSA/Schnorr/VetKD cycles after all checks succeeded 

eichhorl:
If the request queue is full or the key is disabled, the caller is refunded.


rs/execution_environment/src/execution_environment.rs
```diff
@@ -4011,6 +3995,26 @@ impl ExecutionEnvironment {
             ));
         }
 
+        if let Some(signature_fee) = signature_fee {
```
mraszyk, line 3998: Is all the above work cheap enough to justify moving the payment down here?

eichhorl: I would say yes, the map of the first lookup is very small, and the second one is a linear scan over at most 400 entries. But note that the payment didn't move, this function is called with a _clone_ of the request, which is mutated and then inserted into the context, unless the function returns an error (in which case everything is refunded).

mraszyk (approved):

PR #11295, opened 2026-08-24
chore: Remove prefetching memory tracker

michael-weigelt:
The prefetching memory tracker is no longer used anywhere.


rs/canister_sandbox/src/replica_controller/sandboxed_execution_controller.rs
```diff
@@ -357,20 +355,6 @@ impl SandboxedExecutionMetrics {
...
-                exponential_buckets(1.0, 2.0, 22),
-                &["api_type", "memory_type"],
-            ),
-            direct_write_count: metrics_registry.histogram_vec(
-                "sandboxed_execution_direct_write_count",
-                "Number of write accesses handled where the page had not yet been read \
-                    by type of memory (wasm, stable) and api type.",
-                exponential_buckets(1.0, 2.0, 22),
-                &["api_type", "memory_type"],
-            ),
```
alin-at-dfinity, line 373: Both of these histograms are used by the [Subnet Load Tester dashboard](https://github.com/dfinity-ops/k8s/blob/9468cdefa4f964fc2aab35938889da463cf3720b/bases/apps/ic-dashboards/execution/subnet-load-tester.json). Maybe you should replace them with their new equivalent (or rather, replace `rate(old_metric[1m])` with `rate(new_metric[1m]) or rate(old_metric[1m])`, so it's backward compatible; maybe you also need some gymnastics around the labels, if they're different across the two metrics; I'm happy to help).

alin-at-dfinity (commented):
Just one comment about the deleted metrics.

rs/canister_sandbox/src/replica_controller/sandboxed_execution_controller.rs
michael-weigelt: The DMT's equivalent metrics are nondeterministic, so I am not sure how to handle them. Is it ok if I move this out of the scope of this PR via a jira ticket? Missing the metrics for a while shouldn't be too bad.

alin-at-dfinity: Sure, you can also do it later. The Subnet Load Tester dashboard is not exactly critical.

Regarding nondeterministic metrics, many of our metrics are. And even the ones that are deterministic will be updated and scraped at different times; and reset to zero when a replica restarts; so they will never be exactly the same.

But particularly with counters and histograms (which are collections of counters), where you use the rate or increase, they will be similar across replicas so we just take the median: `quantile without(ic_node, instance) (0.5, rate(...))`.

alin-at-dfinity (approved):

PR #11296, opened 2026-08-24
feat(cketh): report the token's minimum deposit amount from deposit_erc20

gregorydemay:
A deposit below the balance-scan minimum is never detected and never credited, so a caller that has just registered a deposit address needs to know the threshold that applies to the token it armed.

- `deposit_erc20` now returns `minimum_deposit_amount` next to the deposit address and its status, in the requested token's own units.
- It is the same threshold `get_minter_info` reports and the balance scan enforces, so a frontend can show the minimum right where it shows the deposit address, without a second call.
- Once DEFI-2961's daily exchange-rate refresh lands, this endpoint tracks the live value automatically.

### DIFF
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -436,6 +436,11 @@ type DepositErc20Response = record {
     // balance scan to detect it. The scan reads the address' whole balance for the token, so
     // several smaller transfers count together; the funds stay undetected only while their
     // total is below this.
+    //
+    // A supported token with no configured minimum reports 2^256 - 1, which no real balance can
+    // reach: a deposit of that token would never be detected. Frontends should treat such a value
+    // as "deposits unavailable for this token" and warn the user, rather than display it as an
+    // amount to send.
     minimum_deposit_amount : nat;
     // Where the deposit stands in the detect-and-sweep pipeline.
     status : DepositStatus;
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -432,6 +432,9 @@ type DepositMode = variant {
 type DepositErc20Response = record {
     // The Ethereum deposit address derived for the caller.
     address : text;
+    // Minimum balance, in the token's own units, that the deposit address must hold for the
+    // balance scan to detect it. A smaller deposit is never detected and never credited.
+    minimum_deposit_amount : nat;
     // Where the deposit stands in the detect-and-sweep pipeline.
     status : DepositStatus;
 };
@@ -873,7 +876,8 @@ service : (MinterArg) -> {
     // Derive and register the ckERC20 deposit address for the caller.
     // The account is derived from the caller's principal and the subaccount in the argument.
     // Returns the EIP-55 checksummed deposit address together with `valid_until`, the timestamp
-    // (nanoseconds since the Unix epoch) until which a deposit to it is guaranteed to be noticed.
+    // (nanoseconds since the Unix epoch) until which a deposit to it is guaranteed to be noticed,
+    // and the minimum amount of the requested token that a deposit must reach to be noticed.
     deposit_erc20 : (DepositErc20Arg) -> (variant { Ok : DepositErc20Response; Err : DepositErc20Error });
 
     // Retrieve the status of a Eth withdrawal request.
--- a/rs/ethereum/cketh/minter/cketh_minter.did
+++ b/rs/ethereum/cketh/minter/cketh_minter.did
@@ -433,7 +433,9 @@ type DepositErc20Response = record {
     // The Ethereum deposit address derived for the caller.
     address : text;
     // Minimum balance, in the token's own units, that the deposit address must hold for the
-    // balance scan to detect it. A smaller deposit is never detected and never credited.
+    // balance scan to detect it. The scan reads the address' whole balance for the token, so
+    // several smaller transfers count together; the funds stay undetected only while their
+    // total is below this.
     minimum_deposit_amount : nat;
     // Where the deposit stands in the detect-and-sweep pipeline.
     status : DepositStatus;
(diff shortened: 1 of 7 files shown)


rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -63,9 +63,16 @@ pub struct Erc20Balance {
...
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug, CandidType, Deserialize)]
+pub struct Erc20MinimumDeposit {
+    pub erc20_contract_address: String,
+    pub minimum_deposit_amount: Nat,
+}
+
 #[derive(Clone, Eq, PartialEq, Debug, CandidType, Deserialize)]
 pub struct MinterInfo {
     pub minter_address: Option<String>,
+    pub sweeper_address: Option<String>,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Good catch — that was a stale base, not an intentional API change. GitHub rebased the parent branch onto master after #11240 squash-merged, while this branch still pointed at the pre-rebase parent, so #11240's commits reappeared in the diff. Rebased onto the current base; the diff is now limited to the six files this change actually touches, and `sweeper_address` is gone from it.

rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -241,6 +249,9 @@ pub enum DepositMode {
 pub struct DepositErc20Response {
     /// The Ethereum deposit address derived for the caller.
     pub address: String,
+    /// Minimum balance, in the token's own units, that the deposit address must hold for the
+    /// balance scan to detect it. A smaller deposit is never detected and never credited.
+    pub minimum_deposit_amount: Nat,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct, and worth fixing — `scan_outcome` compares the address' aggregate `balanceOf` for the token against the minimum, so several individually smaller transfers do count together once their total reaches it. Reworded to describe it as a balance threshold and to say the funds stay undetected only while their total is below it.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -414,6 +422,9 @@ type DepositMode = variant {
 type DepositErc20Response = record {
     // The Ethereum deposit address derived for the caller.
     address : text;
+    // Minimum balance, in the token's own units, that the deposit address must hold for the
+    // balance scan to detect it. A smaller deposit is never detected and never credited.
+    minimum_deposit_amount : nat;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Same fix applied here and on the service-level comment (the line 844 occurrence): both now describe the minimum as a balance the address must hold, and note that smaller transfers accumulate rather than being lost.

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -331,6 +331,36 @@ mod deposit_erc20 {
...
+            .minimum_deposit_amount;
+
+        let (_ckerc20, response) = ckerc20
+            .call_minter_deposit_erc20(caller, Some(DEFAULT_USER_SUBACCOUNT), token)
+            .expect_deposit_response();
+
+        assert_eq!(
+            response.minimum_deposit_amount,
+            Nat::from(CKUSDC_MINIMUM_DEPOSIT)
+        );
```
mbjorkqvist: `add_supported_erc20_tokens()` adds USDT and USDC, but both have a minimum deposit of `10_000_000`. The implementation could return the first minimum deposit (in general), or swap the values for USDT and USDC, and this test would still pass. Would it make sense to verify the minimum deposit amount for all supported tokens, and/or ideally also add another supported token that has a different minimum deposit amount?

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -432,6 +432,11 @@ type DepositMode = variant {
 type DepositErc20Response = record {
     // The Ethereum deposit address derived for the caller.
     address : text;
+    // Minimum balance, in the token's own units, that the deposit address must hold for the
+    // balance scan to detect it. The scan reads the address' whole balance for the token, so
+    // several smaller transfers count together; the funds stay undetected only while their
+    // total is below this.
+    minimum_deposit_amount : nat;
```
mbjorkqvist, line 444: Would it make sense to add a note here for frontend developers (or somewhere else where they may be more likely to look), telling them that if no `MIN_DEPOSITS` has been configured for a token, `Erc20Value::MAX` will be returned for `minimum_deposit_amount`, and they could use that information to e.g., add a warning banner for the user, saying that deposits won't arrive (if the user doesn't do the math themselves, or thinks that the huge number is just some frontend display error)?

mbjorkqvist (approved):
Thanks @gregorydemay, just a couple of minor comments!

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Agreed — with ckUSDC and ckUSDT both at `10_000_000`, nothing here could tell a per-token lookup apart from one that always returned the first minimum. (Swapping the two values would also have been invisible, since they are equal.)

Done in c6c94133e5: the test now registers ckWBTC as well, whose minimum is `15_000`, and asserts every supported token reports its own value, each cross-checked against `get_minter_info`.

To register a third token I split the per-token body out of `add_supported_erc20_tokens` into `add_supported_erc20_token(token)`, so the extra token reuses the existing registration instead of duplicating it. That helper is on the shared setup path (36 call sites), so I ran `ckerc20_test`, `cketh_test`, `lib_tests`, `unit_tests` and `deposit_from_cex` (the one caller via `live_scan.rs`) — all green. The only semantic change is assign-then-register becoming register-then-push; no setup calls it twice, and a second call would already have failed at `expect_new_ledger_and_index_canisters()` before.

Note the token→minimum mapping itself was already covered in `balance_scan/tests.rs` (unknown token → `MAX`, plus every deployed supported token having a `MIN_DEPOSITS` entry); the gap was specifically that nothing proved `deposit_erc20` looks up the token it was asked about.

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Good idea — that value is genuinely misleading if shown as an amount. Documented in 55822e8794, in both the Candid record and the Rust doc comment: a supported token with no configured minimum reports `2^256 - 1`, which no real balance can reach, so it means "deposits unavailable for this token" and should be surfaced as a warning rather than displayed as an amount to send.

PR #11297, opened 2026-08-24
feat: Store the canisters' consumed cycles in SubnetMetrics

mraszyk:
`SubnetMetrics::consumed_cycles_total()` reports the cycles removed from
circulation by the subnet itself and by canisters that have since been deleted:
a sum over three scalar fields plus the subnet-level entries of
`consumed_cycles_by_use_case`. The cycles consumed by the canisters that
currently exist are not part of it; every consumer of the full total folded them
out of `CanisterStates` and added them on. That fold is `O(|hot canisters|)`
(the cold pool keeps a precomputed aggregate), it was spelled out once per
consumer, and it forced the `/subnet` subtree of the canonical state to depend
on `CanisterStates` for a single leaf.

Hold the canisters' part in a new `SubnetMetrics::consumed_cycles_by_canisters`,
refreshed once per round, and add `consumed_cycles_total_including_canisters()`
as the single definition of the full total -- constant time, being
`consumed_cycles_total()` plus one field. The canonical state tree at
`/subnet/<subnet_id>/metrics` reads that starting from the certification version
`V29`, which is not yet enabled (`CURRENT_CERTIFICATION_VERSION` is still
`V28`), so `encode_subnet_metrics` no longer takes the canisters' contribution
as a parameter and the `/subnet` subtree does not need `CanisterStates` at all.

The `replicated_state_consumed_cycles_since_replica_started` gauge reports the
same definition: `ReplicatedStateMetrics::observe` no longer accumulates the
canisters' consumed cycles while iterating them, nor adds the subnet-level use
cases one by one afterwards, but sets the gauge from
`consumed_cycles_total_including_canisters()`. The reported value is unchanged,
as `consumed_cycles_total()` covers exactly the same use cases that the inline
sum did, and the gauge can no longer drift from the certified leaf. Unlike that
leaf, this path is live today; it reads a refreshed value because
`commit_and_certify` enqueues the observation after the refresh described below.
The per-use-case breakdowns alongside the gauge still fold over the canisters,
as no aggregate holds them.

The refresh lives in `StateManagerImpl::commit_and_certify`, beside
`repartition_canister_states()`. That is the one choke point every committed
state passes through, with nothing able to mutate a canister between the refresh
and the hash; refreshing earlier (in message routing or the scheduler) would go
stale, because canisters are still charged afterwards -- HTTP spend refunds,
stream-builder rejects, message shedding -- and because drivers such as
`StateMachine` tests sometimes do not run message routing at all. Like the
repartitioning, it must therefore stay unconditional.

The field is transient: it is not persisted, and it is marked
`#[validate_eq(Ignore)]` because `validate_eq_checkpoint` compares the state
against a `SystemMetadata` loaded from the proto alone, where the field is zero.
`SubnetMetrics` derives `ValidateEq` and `SystemMetadata::subnet_metrics` is
compared with `CompareWithValidateEq` so that the exemption covers this field
alone and the rest of the struct stays validated field by field. This follows
`RefundPool::{amounts, total}`, which are likewise transient and re-derived on
load, and `SystemMetadata::subnet_ids_at_last_reject_generation`, which is
transient but merely reset. The field is `pub`, rather than hidden behind an
observer like its accumulating siblings, because the refresh overwrites it
wholesale.

`ReplicatedState::new_from_checkpoint` derives the field from the canisters it
loads. That is what makes a replica restarting from a checkpoint agree with one
that kept running, and hence certify the same `/subnet/<subnet_id>/metrics`
leaf; a test for that will be added when bumping the certification version. The certified
encoding is unchanged, as the untouched expected bytes in
`encoding/tests/compatibility.rs` show, and so is the state hash.


rs/canonical_state/src/traversal.rs
```diff
@@ -1263,25 +1276,19 @@ mod tests {
...
 
-            // The tree must pass the sum of the non-deleted canisters' consumed
-            // cycles to `encode_subnet_metrics`; it is only reflected in the
-            // encoding starting with V29.
-            let expected_blob = encode_subnet_metrics(
-                &state.metadata.subnet_metrics,
-                consumed_by_canisters,
-                certification_version,
-            );
+            // The tree encodes the `SubnetMetrics` as-is.
```
mraszyk:
```suggestion
```

rs/replicated_state/src/metadata_state.rs
```diff
@@ -681,6 +695,19 @@ impl SubnetMetrics {
         total
     }
 
+    /// All cycles removed from circulation on the subnet, by both deleted and
+    /// still-existing canisters: the subnet-level aggregate
+    /// ([`Self::consumed_cycles_total`]) plus [`Self::consumed_cycles_by_canisters`],
+    /// the end-of-round fold over the canisters that currently exist.
```
mraszyk:
```suggestion
    /// the fold over the canisters that currently exist.
```

rs/canonical_state/src/encoding/tests/compatibility.rs
eichhorl, line 268: Do we need to add the new field here?

rs/replicated_state/src/replicated_state.rs
```diff
@@ -683,6 +692,26 @@ impl ReplicatedState {
...
 
+    /// Refreshes [`crate::metadata_state::SubnetMetrics::consumed_cycles_by_canisters`]
+    /// from the current canister states.
+    ///
+    /// **The caller in `commit_and_certify` must not be made conditional.** The
+    /// field is derived, not persisted, so `Self::new_from_checkpoint` re-derives it
+    /// from the canisters in the checkpoint. Refreshing immediately before the state
+    /// is hashed — with nothing able to mutate a canister in between — is what makes
+    /// the committed value equal the one derived at load, and hence makes a replica
+    /// that keeps running agree with one that restarts from the checkpoint.
```
eichhorl: Do we have a test for this?

rs/replicated_state/src/metadata_state.rs
```diff
@@ -483,6 +484,19 @@ pub struct SubnetMetrics {
...
+    ///
+    /// A *derived* aggregate over `ReplicatedState::canister_states`, refreshed by
+    /// `ReplicatedState::refresh_consumed_cycles_by_canisters` as part of committing
+    /// a state, just before it is hashed. Computing it is `O(|hot canisters|)`; see
+    /// [`crate::CanisterStates::total_consumed_cycles`].
+    ///
+    /// Transient: not persisted, and re-derived from the canisters by
+    /// `ReplicatedState::new_from_checkpoint` on checkpoint load.
+    #[validate_eq(Ignore)]
+    pub consumed_cycles_by_canisters: NominalCycles,
```
eichhorl: Why not persist it?

rs/state_manager/tests/state_manager.rs
```diff
@@ -552,6 +552,79 @@ fn rejoining_node_doesnt_accumulate_states() {
...
+            "consumed_cycles_by_canisters was not re-derived at checkpoint load"
+        );
+        assert_eq!(
+            state.metadata.subnet_metrics.consumed_cycles_by_canisters,
+            state.canister_states().total_consumed_cycles(),
+            "stored aggregate disagrees with a fresh fold over the canisters"
+        );
+
+        // Same value in, same certified state hash out.
+        assert_eq!(hash_before, hash_after);
```
eichhorl: Since the field isn't persisted, should we compare the certified state hash instead (or in addition)?

rs/state_manager/tests/state_manager.rs
```diff
@@ -552,6 +552,79 @@ fn rejoining_node_doesnt_accumulate_states() {
...
+        // so the test does not set it up. Only canister 100 was charged, and the
+        // `Normal` cost schedule values a cycle at par.
+        let expected = state.canister_states().total_consumed_cycles();
+        assert_eq!(expected, NominalCycles::new(123_456));
+        assert_eq!(
+            state.metadata.subnet_metrics.consumed_cycles_by_canisters,
+            NominalCycles::zero()
+        );
+
+        state_manager.commit_and_certify(state, CertificationScope::Full, None);
```
eichhorl: would it make sense to also test with `CertificationScope::Metadata`?

eichhorl (approved):

rs/replicated_state/src/metadata_state.rs
mraszyk: That would require a staged roll-out since the field would be deserialized as 0 after an upgrade and thus could not be used immediately.

rs/replicated_state/src/metadata_state/proto.rs
```diff
@@ -386,6 +386,10 @@ impl TryFrom<pb_metadata::SubnetMetrics> for SubnetMetrics {
             threshold_signature_agreements,
             consumed_cycles_by_use_case,
             consumed_cycles_by_use_case_as_counters,
+            // Transient, with no corresponding proto field:
+            // `ReplicatedState::new_from_checkpoint` derives it from the canisters
+            // it loads.
+            consumed_cycles_by_canisters: NominalCycles::zero(),
```
alin-at-dfinity, line 392: The usual approach we take with transient fields is to initialize them as early as possible by precomputing the value and passing it into the `TryFrom` implementation along the proto: `impl TryFrom<(pb_foo::Foo, NominalCycles)> for Foo`. That way you wouldn't e.g. need the `#[validate_eq(Ignore)]` and rely on checkpoint validation as an additional consistency check.

rs/canonical_state/src/encoding/tests/compatibility.rs
mraszyk: No, this is a different type (the type exported in the certified state tree) and this type (or its values) did not change (that would require a new certification version). What this PR changes in the (internal) type used in the ReplicatedState itself, not the (external) type exported in the state tree.

rs/replicated_state/src/metadata_state.rs
mraszyk: If you think it's better that way and the subnet metrics feature is not urgent, we could take that path.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -683,6 +692,26 @@ impl ReplicatedState {
...
+    ///
+    /// Order relative to [`Self::repartition_canister_states`] does not matter:
+    /// repartitioning moves a canister's contribution between the hot fold and the
+    /// cold aggregate, leaving the total unchanged.
+    ///
+    /// `O(|hot canisters|)`.
```
alin-at-dfinity, line 699: Nit: This (and a lot of other comments and doc comments) are exceedingly verbose. You could drop the two middle paragraphs (maybe replaced by a sentence in the first) without any loss of information. It's not obvious to me why Claude insists on documenting what the caller does with the method and how and when.

All of this "must not be made conditional" and "computed immediately before the state is hashed so it's equal to the value derived at load" should be inline comments in `commit_and_certify()`. And, without checking, I'm pretty sure they are, so this is just redundant. As for the third paragraph, it's a vaguely interesting observation, but totally irrelevant here.

rs/state_manager/tests/state_manager.rs
mraszyk: Good point! But as of the current certification version (before this [PR](https://github.com/dfinity/ic/pull/11184) is merged), the certified state tree does not depend on the new field, so comparing the hashes is trivial (as is the current comparison since the field is not persisted) => dropping the test and adding a [comment](https://github.com/dfinity/ic/pull/11184/files#r3854114100) to the other PR to add the test instead.

rs/state_manager/src/lib.rs
```diff
@@ -3563,6 +3563,12 @@ impl StateManager for StateManagerImpl {
             .observe(state.canister_states().hot_len() as f64);
         state.repartition_canister_states();
 
+        // Like the repartitioning above, this must stay unconditional; see
+        // `ReplicatedState::refresh_consumed_cycles_by_canisters`. Pinned by
+        // `consumed_cycles_by_canisters_is_rederived_at_restart` in
+        // `tests/state_manager.rs`.
+        state.refresh_consumed_cycles_by_canisters();
```
alin-at-dfinity: Well, for once this is under-documented. But only because it says "see XYZ" (in a different crate describing behavior specific to this crate).

But it still managed to add a superfluous reference to a unit test in production code.

rs/replicated_state/src/metadata_state.rs
```diff
@@ -483,6 +484,19 @@ pub struct SubnetMetrics {
     ///
     /// Transactions here refer to all messages processed in replicated mode.
     pub update_transactions_total: u64,
+    /// The total cycles consumed by the canisters that currently exist on this
```
alin-at-dfinity, line 488:
```suggestion
    pub update_transactions_total: u64,

    /// The total cycles consumed by the canisters that currently exist on this
```

rs/state_manager/tests/state_manager.rs
mraszyk: With the same justification as [here](https://github.com/dfinity/ic/pull/11297/files#r3854118418), I added a [comment](https://github.com/dfinity/ic/pull/11184/files#r3854134909) on another PR to add such a test.

alin-at-dfinity (approved):
Only a couple of nitpicks and ranting about Claude's exceedingly verbose comments.

rs/replicated_state/src/metadata_state.rs
alin-at-dfinity: I don't think it's necessary. Nor is it likely a good idea: what should we do if the persisted value differs from the sum of the canisters' consumed cycles?

eichhorl: I'm fine with the current approach, just for my own understanding 👍

rs/canonical_state/src/encoding/tests/compatibility.rs
eichhorl: Ah, makes sense, thanks!

rs/replicated_state/src/metadata_state/proto.rs
mraszyk: That approach would have two downsides:
- `ReplicatedState::new_from_checkpoint` does not derive the field anymore so some tests using that would get the default zero value;
- the `validate_eq_checkpoint` would need to derive the value "manually" by changing `validate_eq_canister_states` from `Result<(), String>` to `Result<NominalCycles, String>)` and computing it inside.

rs/replicated_state/src/replicated_state.rs
mraszyk: I let Claude fix this: https://github.com/dfinity/ic/pull/11297/commits/34ad262ccd796b78fd66a629f4d8c768028f806e

rs/state_manager/src/lib.rs
mraszyk: Fixed in the same commit as the above comment.

rs/replicated_state/src/metadata_state/proto.rs
alin-at-dfinity: `ReplicatedState::new_from_checkpoint` would derive the field: it first loads the canisters, computes their consumed cycles, then passes the amount to the metadata state loader.

Otherwise you wouldn't be able to construct a `MetadataState`. And that's precisely the point.

I believe the same applies to `validate_eq_checkpoint`: every `MetadataState` would have to have an accurate (at least as of last round) `consumed_cycles_by_canisters`; there's never any need to guard against "what if it wasn't set?".

alin-at-dfinity: Else, I suppose you could move `consumed_cycles_by_canisters` into `CanisterStates`, so it's initialized as soon as canister states are loaded. But that may break users that only have the `MetadataState`, not `CanisterStates`; or whenever you temporarily remove a canister from `CanisterStates` to mutate it.

mraszyk: > ReplicatedState::new_from_checkpoint would derive the field: it first loads the canisters, computes their consumed cycles, then passes the amount to the metadata state loader.

`ReplicatedState::new_from_checkpoint` receives the components as input arguments:
```
    pub fn new_from_checkpoint(
        canister_states: BTreeMap<CanisterId, Arc<CanisterState>>,
        mut metadata: SystemMetadata,
        subnet_queues: CanisterQueues,
        refunds: RefundPool,
        epoch_query_stats: RawQueryStats,
    ) -> Self {
```

So I'll keep the merged version.

PR #11300, opened 2026-08-25
refactor(cketh): mock the EVM RPC responses with alloy's RPC types

gregorydemay:
## Why

- `ethers-rs` is deprecated, and its successor `alloy` is where EIP-7702 support lives.
- `ethers-core` is test-only here, but its job is to be an independent oracle for the minter's hand-rolled encoding, so it cannot be replaced by in-repo types without gutting the tests.
- This layer moves the part that decides whether the migration is viable at all: the fixtures the mocked EVM RPC serves.

## What

- Logs, blocks, transaction receipts, fee histories and transaction hashes come from `alloy_rpc_types_eth`.
- Only the crates actually needed are added, not the `alloy` umbrella: it drags in a provider and an HTTP transport for types we never touch.

## Wire format

- The fixtures are serialized straight into the HTTP body the EVM RPC canister parses, and `max_response_bytes` is compared against that body's length, so the bytes are load-bearing.
- Six of the seven fixtures serialize byte-for-byte as before.
- The block loses `sealFields`, which ethers emitted and no node does, and its unset roots now carry the canonical empty-trie hashes instead of zeros.

## Scope

- `ethers-core` stays for now: the transaction and ABI oracles still use it, and they move in the PRs above this one.

[DEFI-2970]: https://dfinity.atlassian.net/browse/DEFI-2970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/ethereum/cketh/test_utils/src/response.rs
```diff
@@ -29,57 +28,46 @@ pub fn balance_scan_response(balances: &[u128]) -> String {
...
-        nonce: Some(H64::zero()),
-        logs_bloom: Some(log_blooms.into()),
-        total_difficulty: Some(0xc70d815d562d3cfa955_u128.into()),
-        mix_hash: Some(H256::zero()),
-        author: Some(miner.into()),
-        size: Some(0x19eea_u64.into()),
+pub fn block_response(block_number: u64) -> alloy_rpc_types_eth::Block {
+    use alloy_primitives::U256;
+
+    alloy_rpc_types_eth::Block {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed, and fixed: `should_serialize_the_mocked_responses_as_the_evm_rpc_expects` now pins the serialized JSON of every fixture the mock serves — block, receipt, fee history, log entry, transaction hash and the empty-log list.

I had captured those bodies before and after the migration to check the change was intentional, but you are right that nothing kept them pinned afterwards. Flipping a single expected field makes the new test fail, so it is checking what it claims to.

For the record, the diff that test now locks in: the block loses `sealFields` and its unset roots go from zeros to the canonical empty-trie hashes; the other six fixtures serialize byte-for-byte as before.

gregorydemay:
🤖 CI note for the whole stack (#11300–#11304): the red `Bazel Test All` and `Bazel Test All on RBE` checks are all the same unrelated target.

`//rs/ic_os/sev/host:host_test` times out 3/3 at 300s, and every run reports `959 tests pass and 1 fails`. The hanging case is `tests::test_load_from_amd_key_server`, which waits on the AMD key distribution server with no timeout of its own, so Bazel's 300s limit kills it; the other five tests in that target pass.

It is not this stack:

- It reproduces on two different CI runners (`zh1-dll11`, `zh1-spm31`), on Namespace RBE remote execution, and on my own machine on an unrelated worktree — the target depends only on `pem`, `rcgen`, `tokio` and the `sev/host` crate.
- Every ckETH target passes in CI, including both PocketIC integration suites, the anvil-backed `deposit_from_cex_demo_test`, `deposit_from_cex`, the canbench test and the doc tests.
- The reason this stack reaches that target at all is that it changes the crate index (`bazel/rust.MODULE.bazel` and the `Cargo.Bazel.*.lock` files), which invalidates every Rust target and so runs the full 960-test suite, whereas most PRs only run their inferred subset.

Re-running would not help while the AMD endpoint is unreachable. Flagging it for whoever owns `rs/ic_os/sev` — the test looks like it wants its own network timeout, or a tag that keeps it out of PR runs.

basvandijk (approved):

Cargo.toml
```diff
@@ -600,6 +600,11 @@ actix-rt = "2.10.0"
 actix-web = "4.9.0"
 actix-web-prom = "0.10.0"
 aide = { version = "^0.14.2", features = ["axum", "axum-json"] }
+alloy-consensus = { version = "1.8.3", features = ["k256", "serde"] }
+alloy-eips = { version = "1.8.3", features = ["k256", "serde"] }
+alloy-primitives = { version = "1.6.0", features = ["k256", "serde"] }
+alloy-rpc-types-eth = "1.8.3"
+alloy-sol-types = "1.6.0"
```
mbjorkqvist, line 607: nit: IIUC, `alloy-eips` and `allow-sol-types` aren't used anywhere in this PR, but since they seem to be used in the follow-up(s), adding them already here seems fine.

rs/ethereum/cketh/test_utils/src/flow.rs
```diff
@@ -1091,28 +1091,45 @@ pub fn encode_principal(principal: Principal) -> String {
...
+    receipt
+        .inner
+        .as_receipt_with_bloom_mut()
+        .expect("BUG: transaction receipt without an EIP-658 status")
+        .receipt
+        .status = succeeded.into();
+}
+
+pub fn increment_max_priority_fee_per_gas(fee_history: &mut alloy_rpc_types_eth::FeeHistory) {
+    for rewards in fee_history.reward.iter_mut().flatten() {
```
mbjorkqvist: This now silently no-ops when `reward` is `None` (it used to be a plain `Vec`, now it's an `Option`). Would it make sense to `expect` here?

mbjorkqvist (approved):
Thanks @gregorydemay, just a couple of nits!

Cargo.toml
gregorydemay: 🤖 Correct — at this layer `test_utils` only pulls `alloy-consensus`, `alloy-primitives` and `alloy-rpc-types-eth`; `alloy-eips` arrives with the EIP-2718 codec in #11301 and `alloy-sol-types` with the ABI oracle in #11303.

They are declared here on purpose: `[workspace.dependencies]`, the `crate.spec(...)` entries in `bazel/rust.MODULE.bazel` and `./bin/bazel-pin.sh` are one unit, and regenerating `Cargo.Bazel.*.lock` three times would put a large, unreviewable lockfile diff in three separate layers instead of one. Each crate is still only wired into a `Cargo.toml` / `BUILD.bazel` target in the layer that actually uses it.

rs/ethereum/cketh/test_utils/src/flow.rs
gregorydemay: 🤖 You are right about the shape change: ethers had `reward: Vec<Vec<U256>>`, alloy has `reward: Option<Vec<Vec<u128>>>`, and `.iter_mut().flatten()` turns a `None` into a silent no-op instead of a loud one.

No test is silently skipped today — `fee_history()` builds the fixture from a JSON body that always carries `reward`, so it is `Some` at every call site. The trap is that a future fixture change could drop it and the mutator would quietly stop mutating, leaving tests that think they raised the priority fee asserting on unchanged data.

So yes, an `expect` here is the right call — flagging it rather than folding it into this pass.

gregorydemay: 🤖 Added in `2324ed2b1c`: the helper now takes `fee_history.reward.as_mut()` and `expect`s it, so a fee history without rewards fails loudly instead of mutating nothing.

Both integration suites still pass, which confirms no call site relies on the `None` case.

PR #11306, opened 2026-08-25
chore(BOUN-1977): Bump ic-bn-lib & ic-gateway

blind-oracle:
* Enable `acme` feature of `ic-gateway`
* Use tags instead of git rev
* Bump `pebble` binary versions to align with `ic-bn-lib` (CLI arg changed)
* Some formatter kicked in on TOML

@basvandijk @nmattia this caused `pem` crate to diverge into `3.x` and `4.x` (since we switched to 4 already), not sure if that's a problem, probably others will catch up soon. In the worst case I can downgrade.

### DIFF
--- a/Cargo.Bazel.json.lock
+++ b/Cargo.Bazel.json.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "ecab59e54cd3dfcb3d049bf925c4d1c1e1ab338ef237c40f066ef568f79e0321",
+  "checksum": "9cb8107bbffe0c12d79395ab33e0d176425c3e09b084c8c79e8cf0ebf9bcf237",
   "crates": {
     "actix-codec 0.5.1": {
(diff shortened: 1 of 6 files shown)


basvandijk (approved):
> @basvandijk @nmattia this caused pem crate to diverge into 3.x and 4.x (since we switched to 4 already), not sure if that's a problem, probably others will catch up soon. In the worst case I can downgrade.

What's holding `pem` at 3.x?

blind-oracle:
> > @basvandijk @nmattia this caused pem crate to diverge into 3.x and 4.x (since we switched to 4 already), not sure if that's a problem, probably others will catch up soon. In the worst case I can downgrade.
> 
> What's holding `pem` at 3.x?

There are some transient crates that depend on 3.x, so I didn't bump it in the monorepo since that wouldn't fix the divergence.

PR #11309, opened 2026-08-25
fix(bazel): validate the mainnet-revision JSON fields that reach download URLs

basvandijk:
Addresses the **first recommended fix of security finding F-051**: URL components taken from an unreviewed, externally-derived JSON file were interpolated into download URLs without validation.

## Why

`mainnet-canister-revisions.json` and `mainnet-icos-revisions.json` are written by the mainnet-revisions bot, whose PRs are auto-approved and auto-merged, from data that is not independently authenticated (the public dashboard REST API, CDN-served `SHA256SUMS` files, and the hash of whatever the CDN currently serves). The repository rules that read them treated every field as trusted, interpolating it into:

- download URLs — a `tag` such as `../../../org/repo/releases/download/x` escapes the hard-coded GitHub repository under standard RFC 3986 dot-segment normalization; a `rev`/`version` selects any other artifact on the CDN,
- the `output` path of `repository_ctx.download`, which `..` can walk out of,
- the text of a generated `BUILD.bazel` file (the ICOS binary names), where quotes and newlines inject Starlark.

Two things this also closes, found while implementing:

- **An empty hash silently disables verification.** `repository_ctx.download(sha256 = "")` skips the integrity check, and `""` passed the existing `== None` guard.
- **`binaries` keys are interpolated raw into generated Starlark.** Unlike canister keys (which must match a reviewed `filenames` entry in `MODULE.bazel`), they are constrained by nothing.

## What

New `bazel/mainnet-artifact-refs.bzl` holds the validators; every choke point uses them:

| File | Checks added |
| --- | --- |
| `mainnet-canisters.bzl` | URL construction extracted into `canister_download_url` (validates rev/tag/repository/filename); impl validates the canister key and sha256 |
| `mainnet-icos-binaries.bzl` | `binary_download_url` validates; whole record validated before any download and before the BUILD text is generated |
| `mainnet-icos-images.bzl` | both URL builders validate, `variant` gated on a fixed set, record validated before the multi-GB fetches |
| `mainnet-icos-versions.bzl` | every record validated before `defs.bzl` is generated |
| `mainnet_revisions.py` | `VersionInfo.__post_init__` mirrors the checks so a poisoned upstream value fails when the JSON is written, not only when it is fetched |

Because the URL builders themselves validate, every call site is covered by construction — including the `ENV_DEPS__*_IMG_URL` env vars built by `rs/tests/configure_icos.bzl`. Validating in `mainnet-icos-versions.bzl` covers the matching `*_IMG_HASH` values, which the Farm backend uses to fetch and verify images itself, never through `repository_ctx.download`.

The validators return an error message rather than calling `fail()` (uncatchable in Starlark) — that is what makes the rejecting path unit-testable. `//bazel/mainnet_artifact_refs_tests` is the first Starlark unit test in the repo (bazel_skylib `unittest`, already a `bazel_dep`).

## Verification

- **The F-051 criterion, end to end**: with the exploit tag planted in the JSON, `bazel fetch --repo=@mainnet_canisters` fails with `canister "internet_identity_backend": tag: a tag must not contain '..'`.
- `bazel test //bazel/mainnet_artifact_refs_tests/...` — 9/9 pass; I also confirmed the suite fails on a deliberately bad assertion, so the passes are not vacuous.
- `bazel build //... --nobuild` clean; `bazel fetch` of `@mainnet_canisters` and `@mainnet_nns_binaries` succeed; all four committed ICOS records are accepted by both the Starlark and the Python validators. Accept-path tests pin the exact URLs fetched today, so validation cannot silently move a fetch.

## Not in this PR

- F-051's **second** recommended fix (anchoring hashes to an independent trust source; removing auto-approve for values the updater did not cross-verify on-chain).
- Mirroring the checks in `rs/nervous_system/tools/sync-with-released-nervous-system-wasms`, which writes the canister-side `rev`/`tag`/`sha256`. Same few lines, easy follow-up.

### DIFF
--- a/bazel/mainnet_artifact_refs_tests/mainnet_artifact_refs_test.bzl
+++ b/bazel/mainnet_artifact_refs_tests/mainnet_artifact_refs_test.bzl
@@ -1,0 +1,433 @@
+"""Unit tests for //bazel:mainnet-artifact-refs.bzl and the download URL builders.
+
+The values that mainnet-canister-revisions.json and mainnet-icos-revisions.json
+contribute to a download URL are not reviewed by a human before they reach master, so
+the repository rules validate them (F-051). These tests pin down what is accepted and
+what is rejected, including the fields of poisoned copies of the real JSON records.
+
+The rejecting path of the URL builders themselves cannot be asserted -- `fail()` is
+uncatchable in Starlark -- which is exactly why the validators return an error message
+instead of failing: the builders are the only path from JSON to URL and they call
+these very functions.
+"""
+
+load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load(
+    "//bazel:mainnet-artifact-refs.bzl",
+    "artifact_name_error",
+    "commit_id_error",
+    "github_repository_error",
+    "icos_record_error",
+    "sha256_error",
+    "tag_error",
+    "variant_error",
+)
+load("//bazel:mainnet-canisters.bzl", "canister_download_url")
+load("//bazel:mainnet-icos-binaries.bzl", "binary_download_url")
+load("//bazel:mainnet-icos-images.bzl", "icos_dev_image_download_url", "icos_image_download_url")
+
+# Values taken from the committed JSON files: these must never start failing.
+_REAL_COMMIT_IDS = [
+    "3ae3649a2366aaca83404b692fc58e4c6e604a25",
+    "512cf412f33d430b79f42330518166d14fc6884e",
+    "59c42492cd6ee3ef5e60cd35d7b766e7bfd085da",
+    "79c01052b5f7f49d3cf53d04d696cb2893294cd3",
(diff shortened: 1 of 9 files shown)


bazel/mainnet_artifact_refs_tests/mainnet_artifact_refs_test.bzl
```diff
@@ -0,0 +1,433 @@
+"""Unit tests for //bazel:mainnet-artifact-refs.bzl and the download URL builders.
+
+The values that mainnet-canister-revisions.json and mainnet-icos-revisions.json
+contribute to a download URL are not reviewed by a human before they reach master, so
+the repository rules validate them (F-051). These tests pin down what is accepted and
```
basvandijk:
```suggestion
the repository rules validate them. These tests pin down what is accepted and
```

dfinityianblenke (approved):

PR #11313, opened 2026-08-25
docs(cketh): rework deposit-from-CEX diagrams around minter addresses

gregorydemay:
## Why

- The deposit-from-CEX design doc lumped all on-chain actors into a single "Ethereum" participant. The addresses controlled by the minter (main, sweeper, deposit EOAs) and the contracts involved (sweeper delegate, deposit helper) were indistinguishable in the diagrams.
- Diagrams and tables for discarded alternatives were interleaved with the decided design.
- The doc had drifted from the implementation: stale gas figures, and an ETH deposit flow built on a separate per-asset address that #11449 superseded.

## What

- Add an architecture diagram separating the minter-key EOAs from the deployed contracts, with the role of each address spelled out.
- Rework the end-to-end sequence diagrams with the sweeper address, deposit EOA, main address, delegate, and helper as distinct participants.
- Diagram the sweeper-funding flow and, in the EIP-7702 primer, the two sweep-transaction shapes between the deposit EOA and the sweeper contract.
- Describe the Phase 2 ETH deposit flow in its own section, following the #11449 mainnet validation: ETH reuses the same delegated deposit address and the same batch sweeper contract, and the per-asset address split becomes the fallback.
- Align the text with the implemented sweeping: permissionless, guarded by a one-time attestation per address (recorded and reused), per-token batches, immediate enqueue on detection, dedicated sweeper lane with prepaid-gas accounting.
- Address the Copilot review: measured gas figures, signature table restricted to the paths that actually sign, EVM-RPC statements scoped to the minter, and an explicit note that R3 sender screening is not yet enforced for automatically detected deposits.
- Move everything about discarded alternatives to a section at the bottom.

https://claude.ai/code/session_01Vv3wMNq4fxGRws91DSoMru

### DIFF
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -317,7 +317,7 @@ sequenceDiagram
     Note over Minter: balance detected: queue (deposit address, USDT) for sweeping.<br/>Scheduling hint only, no finality needed —<br/>a reorged deposit just wastes the sweep's gas
     Note over Minter,D: first sweep of this address only — sign with the deposit address' key (tECDSA):<br/>the one-time attestation of the account (principal, subaccount) and an EIP-7702<br/>authorization (delegate = SweeperContract), both recorded and reused by later sweeps
     Note over Minter,Sw: sweeper address already funded:<br/>ckETH burned from the fee account at funding time (R14)
-    Sw->>S: sweep tx on the sweeper's own nonce lane (R17), signed with path [3]:<br/>type 0x04 (0x02 once delegated), one token per tx, up to 20 deposits:<br/>sweepErc20Batch([(deposit address, principal, subaccount, attestation)], [USDT])
+    Sw->>S: sweep tx on the sweeper's own nonce lane (R17), signed with path [3]:<br/>type 0x04 (0x02 once delegated), one token per tx, up to 10 deposits:<br/>sweepErc20Batch([(deposit address, principal, subaccount, attestation)], [USDT])
     S->>D: sweepErc20 — delegated: SweeperContract's code runs as the deposit address
     Note right of D: ecrecover(attestation digest) must equal address(this):<br/>only the account attested by the deposit address' own key can be credited
     D->>T: approve(Helper, 250)
@@ -788,7 +788,7 @@ Sweeping is two background tasks feeding one transaction pipeline:
   token, because the delegate's batch entry point runs its whole token list
   against every deposit address it touches, so a mixed batch would pay a
   `balanceOf` on every `(address, token)` cross-product pair — most of them
-  holding nothing. Up to 20 deposits ride one transaction
+  holding nothing. Up to 10 deposits ride one transaction
   (`MAX_DEPOSITS_PER_SWEEP`, gas-bound). Per deposit the task signs the account
   attestation and the EIP-7702 authorization if it holds no recorded one yet —
   both are signed once per address, recorded, and reused by every later sweep.
@@ -799,9 +799,12 @@ Sweeping is two background tasks feeding one transaction pipeline:
 * The **send task** drives the sweeper pipeline through the same
   create → sign → send → resubmit → finalize state machine as user withdrawals,
   but signs with the sweeper derivation path `[3u8]` and tracks the sweeper
-  address' own transaction count (`R17`). A freshly enqueued sweep kicks it
-  immediately rather than waiting for its next tick: the mint follows the sweep
-  (step 4), so every interval spent waiting is crediting latency a user sees.
+  address' own transaction count (`R17`). The two tasks run on independent
+  timers — enqueue every minute, the pipeline every six
+  (`SWEEP_ENQUEUE_INTERVAL`, `PROCESS_SWEEPER_TRANSACTIONS_INTERVAL`) — so an
+  accepted sweep waits for the pipeline's next tick: since the mint follows the
+  sweep (step 4), those intervals bound the crediting latency a user sees on
+  top of detection.
   Confirmation is via transaction receipt, like withdrawals, emitting audit
   events (`R5`, `R8`).
 
@@ -1133,7 +1136,8 @@ tuple skipped by the protocol (e.g. stale nonce) makes the corresponding inner
 call hit a code-less address, which reverts the *whole* batch (atomic, funds
 safe, gas wasted — retry); mixed batches are fine (tuples only for
 not-yet-delegated addresses, already-delegated ones ride along without tuples);
-batch size is bounded only by gas (`N ≈ 20` per the sweeping policy in step 5).
+batch size is bounded by gas, and the sweeping policy caps it at `N = 10`
+(`MAX_DEPOSITS_PER_SWEEP`, step 5), comfortably under that bound.
 
 Delegation cannot be scoped to a single transaction: the original EIP-7702 draft
 was ephemeral but the final spec is persistent, tuples are all applied *before*
@@ -1329,7 +1333,7 @@ Notes:
   `balanceOf(address(this))` equals the amount `depositErc20` transfers. This is the
   same assumption as today's helper-contract deposit flow.
 * The integration harness deploys exactly this contract,
-  [`CkSweeperAttested.sol`](../minter/tests/deposit_from_cex_demo/CkSweeperAttested.sol).
+  [`CkSweeperAttested.sol`](../minter/CkSweeperAttested.sol).
 
 ### Test plan
 
@@ -1337,7 +1341,9 @@ A runnable end-to-end demonstration of the sweep mechanism (unfunded deposit EOA
 plain USDT-style transfers, single and batched type-`0x04` sweep transactions with gas
 paid by the minter, gas assertions) against a local dev node (any post-Pectra
 version) is available in
-[`deposit_from_cex_demo/`](../minter/tests/deposit_from_cex_demo/).
+[`deposit_from_cex_demo.rs`](../minter/tests/deposit_from_cex_demo.rs)
+(its Solidity support contracts sit in
+[`deposit_from_cex_demo/`](../minter/tests/deposit_from_cex_demo/)).
 It exercises the attested delegate against the *real*
 `DepositHelperWithSubaccount.sol` bytecode, asserting the emitted
 `ReceivedEthOrErc20` events carry the right principals, re-delegation of already
@@ -1410,7 +1416,8 @@ and HTTPS outcalls (both paid in cycles on the IC) and Ethereum gas (paid in ETH
 the sweeper address, backed by the fee account per `R14`). The scenarios below size a
 **single ERC-20 deposit to a fresh address, first sweep, under the decided
 helper-based sweep with one-time self-attestation**, for the two batch extremes
-and two latencies.
+— solo, and a full batch of 10 (`MAX_DEPOSITS_PER_SWEEP`, the sweeping policy's
+cap) — and two latencies.
 
 **Unit costs** — from the
 [IC cycle-cost reference](https://docs.internetcomputer.org/references/cycle-costs/),
@@ -1430,7 +1437,8 @@ logical × 3, each charged fully); reserved `max_response_bytes` ≈ 2 KB for sm
 at $2'500 and the attested harness' measured sweep gas (`ATTESTED_SCENARIOS`
 in the demo, all through the batch entry point): 98'075 gas for a single
 first-sweep deposit in a batch of one, 60'975 each at ten deposits of one
-token (609'750 total), 59'674 each at twenty (1'193'491).
+token (609'750 total), 59'674 each at twenty (1'193'491; the harness measures
+up to twenty, the production policy caps a batch at 10).
 
 **Detection schedule** — per armed `(address, token)` pair (each pair is one
 `balanceOf` sub-call), backing off over the 24h scanning window of
@@ -1463,8 +1471,8 @@ scan, and of addresses sharing its sweep):
 |---|---|---|---|---|---|
 | **B=1, swept ≤5 min** | 3 → $0.107 | ≈ $0.011 | **$0.118** | $0.25 | **$0.36** |
 | **B=1, swept ≤24h** | 3 → $0.107 | ≈ $0.037 | **$0.144** | $0.25 | **$0.39** |
-| **B=20, swept ≤5 min** | 2.05 → $0.073 | ≈ $0.0016 | **$0.075** | $0.15 | **$0.22** |
-| **B=20, swept ≤24h** | 2.05 → $0.073 | ≈ $0.0042 | **$0.077** | $0.15 | **$0.23** |
+| **B=10, swept ≤5 min** | 2.1 → $0.075 | ≈ $0.002 | **$0.077** | $0.15 | **$0.23** |
+| **B=10, swept ≤24h** | 2.1 → $0.075 | ≈ $0.006 | **$0.081** | $0.15 | **$0.23** |
 
 Signatures: one EIP-7702 authorization and one account attestation (both by the
 deposit key) plus one outer sweep-transaction signature (the sweeper key, shared
@@ -1472,7 +1480,7 @@ deposit key) plus one outer sweep-transaction signature (the sweeper key, shared
 window before detection; the sweep itself is enqueued as soon as the balance is
 seen (step 5), so latency only moves outcall cost. Ethereum gas is unchanged by
 latency and scales linearly with the gas price (B=1: $0.025 / $0.25 / $2.45 at
-0.1 / 1 / 10 gwei; B=20 ≈ 40% less).
+0.1 / 1 / 10 gwei; B=10 ≈ 38% less).
 
 **Takeaways:**
 
@@ -1481,8 +1489,8 @@ latency and scales linearly with the gas price (B=1: $0.025 / $0.25 / $2.45 at
   sub-cent even for a full-day solo scan.
 * **Ethereum gas is the swing factor**, with a crossover near **≈ 0.5 gwei**: below it
   the signatures set the floor, above it gas dominates.
-* **Batching trims gas by ≈ 40% and shares the outer signature**; the big win is
-  outcalls, collapsed ≈ 7×.
+* **Batching trims gas by ≈ 38% and shares the outer signature**; the big win is
+  outcalls, collapsed ≈ 5–6×.
 * **Fee floor (`R7`):** to break even excluding gas, `deposit_fee` must cover ≈ $0.08
   (batched) to ≈ $0.14 (solo, full-window scan) of IC resources plus the prevailing
   sweep gas — so at low gas the signatures, not the gas, set the floor. The estimate is
@@ -1593,7 +1601,7 @@ on.
 | Variant | Pros | Cons |
 |---|---|---|
 | **Single shared address** with permanent delegation (chosen): the delegate carries an empty payable `receive()` and attested `sweepEth`/`sweepEthBatch` entry points | One address, one authorization, one attestation per account — across assets; ETH sweeps reuse the whole ERC-20 sweeping machinery (`R14` gas, `R17` lane) | The delegate must accept ETH (guarded against sends to the implementation itself); a sender hard-coding the 21'000 gas limit fails against a delegated address — safely, at the sender (`R12`), and not observed at Binance/Kraken |
-| **Per-asset addresses** (previously chosen): ERC-20 address (schema 1, delegated once, permanently) + ETH address (schema 2, never delegated) | ETH address never has code → fixed-21'000-gas CEX withdrawals always work (`R12`), no failure window; ETH sweeps need no EIP-7702 at all (deposit pays its own gas, 21'000 gas, cheapest possible) | Two addresses per account to register/scan; user must use the right address per asset; ETH sweeps need their own send path (key-signed, fee-capped against the deposit) instead of reusing the sweeper lane. Kept as the fallback should an exchange's ETH withdrawal path prove incompatible with a delegated address |
+| **Per-asset addresses** (previously chosen): ERC-20 address (schema 1, delegated once, permanently) + ETH address (schema 2, never delegated) | ETH address never has code → fixed-21'000-gas CEX withdrawals always work (`R12`), no failure window; ETH sweeps need no EIP-7702 at all (a key-signed `depositEth` helper call, gas paid from the deposit) | Two addresses per account to register/scan; user must use the right address per asset; ETH sweeps need their own send path (key-signed, fee-capped against the deposit) instead of reusing the sweeper lane. Kept as the fallback should an exchange's ETH withdrawal path prove incompatible with a delegated address |
 | **Single shared address** with *set-and-clear* delegation (install delegate, sweep, re-delegate to `address(0)`) | One address per account | Two tECDSA signatures + ≈ 2 × 12'500–25'000 gas per sweep cycle; short window in which fixed-gas ETH transfers fail at the sender; more complex delegation lifecycle |
 
 ### What triggers detection (step 3)
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -964,6 +964,29 @@ tuple single-use (its application consumes the pinned nonce) and revocable
 (consuming the nonce any other way invalidates an outstanding signed tuple). The
 signature is not a transaction — nothing on chain changes.
 
+The two sweep transactions that follow, side by side — the only difference
+between an EOA's first sweep (Transaction 2) and every later one
+(Transaction 3) is the authorization list:
+
+```mermaid
+sequenceDiagram
+    participant Sw as SweeperAddress
+    participant S as SweeperContract (deployed instance)
+    participant D as Deposit (EOA)
+
+    Note over Sw,D: Transaction 2 — first sweep, type 0x04,<br/>authorization_list = [tuple signed by Deposit's key]
+    Note over D: tuple processing, before any code runs:<br/>Deposit's code := 0xef0100‖SweeperContract, nonce 0 → 1
+    Sw->>S: sweepErc20Batch([(Deposit, p, s, attestation)], [Token])
+    Note right of S: plain-contract hat: address(this) = SweeperContract,<br/>msg.sender = SweeperAddress
+    S->>D: sweepErc20([Token], p, s, attestation)
+    Note right of D: delegate hat: the designator loads SweeperContract's code,<br/>which runs AS Deposit — address(this) = Deposit,<br/>msg.sender = SweeperContract.<br/>Attestation verified, then approve + Helper.depositErc20:<br/>the whole Token balance lands at MainAddress (R6)
+
+    Note over Sw,D: Transaction 3 — any later sweep, plain type 0x02 (EIP-1559):<br/>no authorization list
+    Sw->>S: sweepErc20Batch([(Deposit, p, s, attestation)], [Token])
+    S->>D: sweepErc20([Token], p, s, attestation)
+    Note right of D: same delegated execution — the designator persists,<br/>and Deposit's nonce is unchanged since<br/>(only its own authorizations advance it)
+```
+
 **Transaction 2 — first sweep (type `0x04`, sent and paid by `SweeperAddress`,
 `to = SweeperContract`).** This is the transaction the minter actually sends:
 the sweep is *never* addressed to the deposit EOA — every sweep, including a
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -379,18 +379,54 @@ with an `icrc1_balance_of` of `1_762_128_000_000_000_000` wei ≈ 1.76 ckETH as 
    the optional fee arguments (step 1) transfers the caller-specified ckETH amount
    into the fee account (`icrc2_transfer_from`, see the variants below) — plus
    treasury top-ups and converted `deposit_fee` revenue (see Non-goals).
-1. **Daily funding task**: compare the sweeper address' prepaid gas against its
-   low-water mark. The minter needs no chain read for this: it tracks a *lower
-   bound* on that balance from its own recorded events, and if the bound is below
-   the low-water mark it **withdraws ckETH from the fee account to the sweeper
-   address** — an ordinary ckETH withdrawal through the existing pipeline (burn
-   from the fee account, then send the ETH on the main address' nonce sequence).
-   `R14` holds by construction, with no new burn path to audit; this pipeline is
-   the *only* way ETH is spent on sweeps. Unlike sweeps (`R17`), this funding
-   transaction rides the main address' nonce sequence — the withdrawal lane — so a
-   stuck funding transaction can head-of-line-block withdrawals; acceptable because
-   funding is infrequent (batched to cover many sweeps) and uses the same
-   resubmission machinery as withdrawals.
+1. **Daily funding task**: once a day (and once at install, right after the
+   ECDSA key is cached), decide from the minter's **own records** whether the
+   sweeper address needs a top-up. No chain read: the balance used is a *lower
+   bound* — what finalized fundings delivered, less what accepted sweeps have
+   provisioned, plus what finalized sweeps handed back — so erring low can
+   only over-provision gas, never spend against gas that is not there. If the
+   bound is below the **low-water mark**
+   (half the target; the target is 10 × the ckETH minimum withdrawal amount,
+   ≈ 0.3 ETH against mainnet's 0.03, provisional until calibrated on Sepolia),
+   **withdraw ckETH from the fee account to the sweeper address**, topping the
+   bound up to the target — an ordinary ckETH withdrawal through the existing
+   pipeline (burn from the fee account first, then send the ETH on the main
+   address' nonce sequence). One funding at a time: none is started while a
+   previous one is still in flight, or when the minter's deposit-backed ETH
+   does not cover the amount. `R14` holds by construction, with no new burn
+   path to audit; this pipeline is the *only* way ETH is spent on sweeps.
+   Unlike sweeps (`R17`), the funding transaction rides the main address'
+   nonce sequence — the withdrawal lane — so a stuck funding transaction can
+   head-of-line-block withdrawals (and, one at a time, later fundings);
+   acceptable because funding is infrequent (batched to cover many sweeps) and
+   uses the same resubmission machinery as withdrawals.
+
+The funding flow end-to-end — note that the decision reads nothing but the
+minter's own accounting, and that the burn strictly precedes any ETH movement:
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant Minter as Minter canister
+    participant Ledger as ckETH ledger
+    box Minter-key EOAs
+        participant Main as main address — path []
+        participant Sw as sweeper address — path [3]
+    end
+
+    Note over Minter: daily timer, plus once at install.<br/>No chain read: the sweeper balance used is the minter's own lower bound —<br/>what finalized fundings delivered, less what accepted sweeps have provisioned,<br/>plus what finalized sweeps handed back
+    Minter->>Minter: plan_funding: bound below the low-water mark (target/2)?<br/>no funding still in flight? deposit-backed ETH covers it?
+    alt not due, or one already in flight, or insufficient backing
+        Note over Minter: skip — nothing burned, nothing moves
+    else funding due: amount = target − bound
+        Minter->>Ledger: burn amount of ckETH from the minter fee account 0x…fee<br/>(memo: convert to the sweeper address)
+        Note over Minter,Ledger: burn FIRST (R14): a failed burn (e.g. empty fee account)<br/>means no ETH moves. From here on,<br/>cumulative burned ≥ cumulative spent
+        Minter->>Main: enqueue an ordinary withdrawal request,<br/>destination = sweeper address, keyed by the ledger burn index
+        Main->>Sw: EIP-1559 transfer of amount − fee, on the MAIN address'<br/>nonce lane — the withdrawal lane, unlike sweeps (R17)
+        Note over Minter,Sw: the receipt finalizes the funding: transferred + fee are recorded.<br/>A failed transfer is NOT reimbursed —<br/>the burn minus its gas stays as backing
+        Note over Minter,Sw: sweeper-balance lower bound += transferred:<br/>the prepaid gas future sweeps are priced against
+    end
+```
 
 * The sweeper address' balance is the `prepaid_sweep_gas` counter, and the minter
   tracks a lower bound on it from its own events — what finalized fundings
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -12,6 +12,7 @@ tags: [cketh, ckerc20, minter, deposit, eip-7702]
 - [Design](#design)
   - [Addresses and contracts](#addresses-and-contracts)
   - [End-to-end flows](#end-to-end-flows)
+  - [The ETH deposit flow (Phase 2)](#the-eth-deposit-flow-phase-2)
   - [Step 0: Fund the transaction fees](#step-0-fund-the-transaction-fees-without-touching-the-cketh-backing)
   - [Step 1: Retrieve the deposit address](#step-1-retrieve-the-deposit-address)
   - [Step 2: Withdraw from the CEX](#step-2-withdraw-from-the-cex-to-the-deposit-address)
@@ -70,9 +71,10 @@ The design is delivered in two phases:
 * **Phase 1 — ckERC20 only** (ckUSDC, ckUSDT, …): deposits are ERC-20 `Transfer`s,
   which always emit logs and never execute recipient code, making detection and
   crediting straightforward.
-* **Phase 2 — ckETH**: native ETH transfers emit no logs and interact badly with
-  EIP-7702 delegated code under fixed 21'000 gas limits; this phase has additional
-  design constraints, described per step.
+* **Phase 2 — ckETH**: native ETH transfers emit no logs, and a send with a
+  fixed 21'000 gas limit cannot reach EIP-7702 delegated code; this phase has
+  additional design constraints, described in
+  [The ETH deposit flow](#the-eth-deposit-flow-phase-2) and per step.
 
 ## Requirements
 
@@ -161,8 +163,10 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
   account, exactly once per balance observation (monotone accounting: total credited
   never exceeds total received).
 * `R12`: A plain ETH transfer sent with a fixed 21'000 gas limit to a deposit address
-  MUST NOT be permanently locked: it either succeeds (address has no code at transfer
-  time) or fails on the sender side (funds never leave the exchange).
+  MUST NOT be permanently locked: it either succeeds (a code-less address takes
+  21'000 gas; the delegate's empty `receive()` takes 21'055, so it lands given any
+  higher or estimated limit) or fails on the sender side (funds never leave the
+  exchange).
 
 ## Non-goals
 
@@ -195,10 +199,11 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
   * A CEX that batches ETH withdrawals through a contract (internal transactions)
     provides no sender information without trace APIs; Phase 2 compliance screening is
     therefore weaker for ETH than for ERC-20 (see step 3).
-  * An account's ETH deposit address (Phase 2) differs from its ERC-20 one, matching
-    the per-asset deposit-address UX of exchanges; sending the wrong asset class to
-    an address is not credited automatically, but funds always remain recoverable
-    (key-controlled addresses, `R12` guarantees no loss on the sender side).
+  * An account's ETH deposits (Phase 2) reuse its ERC-20 deposit address, so there
+    is no wrong-address case between the two asset classes; an *unsupported* ERC-20
+    token sent there is not credited automatically, but funds always remain
+    recoverable (key-controlled addresses, `R12` guarantees no loss on the sender
+    side).
 
 ## Design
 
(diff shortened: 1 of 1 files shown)


rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -204,116 +205,158 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
+  never sends sweeps.
+* **Sweeper address** (path `[3]`): holds nothing but prepaid gas (funded by
+  ckETH withdrawals from the fee account, step 0, `R14`) and sends every sweep
+  transaction on its own nonce lane, so a stuck sweep can never
+  head-of-line-block a withdrawal and vice versa (`R17`). Swept funds never
+  pass through it.
+* **Deposit EOAs** (paths `[1, p, s]`, Phase 2 `[2, p, s]`): one per IC
+  account, the only addresses a user ever sees. They hold deposits between the
+  CEX transfer and the sweep and never need ETH for gas; because the minter
+  holds their keys, funds there stay recoverable even without any contract.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Addressed by the restructure in this PR (#11313): the walkthrough now names six actors with no separate Phase-2 ETH EOA — ETH deposits ride the shared delegated address and the sweeper address pays the gas (see "The ETH deposit flow").

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -204,116 +205,158 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
+
+### End-to-end flows
+
+Deposit of USDT from a CEX. Every arrow into an Ethereum participant is
+submitted (and every log read) through the EVM-RPC canister with multi-provider
+consensus; the boxes separate the minter-key EOAs from the deployed contracts:
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Addressed in this PR (#11313): the flow intro now reads "Every transaction the minter submits (and every log or balance it reads) goes through the EVM-RPC canister … — the CEX submits its own transfer directly".

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -574,9 +616,9 @@ changes through `transfer`/`transferFrom`/mint, all emitting a `Transfer` log, s
 every balance increase has a corresponding log in the batched query above, carrying
 the sender — including transfers initiated by contracts (internal transactions). No
 crediting, screening decision, or sweep is based on a balance observation alone.
-**Under variant B the log query is mandatory before scheduling any sweep** (the
-helper event's `owner` is the deposit EOA, not the real sender, so screening cannot
-be left to the pipeline; the owner↔principal cross-check remains as
+**The log query is mandatory before scheduling any sweep** (the sweep's helper
+event carries the deposit EOA as `owner`, not the real sender, so screening
+cannot be left to the pipeline; the owner↔principal cross-check remains as
 belt-and-braces).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Addressed by disclosure in this PR (#11313): the balance-scan section now explicitly flags this — "Not yet enforced: the implemented enqueue task turns a detected balance straight into a sweep request without running this log query first, so R3 sender screening for automatically detected deposits is an open gap … to be closed before launch by gating sweep scheduling on completed sender screening."

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -204,116 +205,158 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
+  already-deployed helper (`DepositHelperWithSubaccount.sol`), so every sweep
+  emits the canonical `ReceivedEthOrErc20` event and the **existing**
+  scrape→parse→dedup→mint pipeline credits the deposit unchanged (step 4). The
+  discarded alternative — a direct sweep with its own detection→mint path,
+  "variant A" — is at the bottom.
+* **Sweeping is permissionless, guarded by a one-time self-attestation**: any
+  caller may submit a sweep and only ever donates gas — a signature by the
+  deposit address' own key, not the identity of the caller, fixes which IC
+  account is credited, and funds only ever move through the helper to the
+  minter's main address (`R6`, step 5).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The implementation caches attestations now: sign_attestations_batch signs only requests with no recorded signature and records them in automatic_deposits.attestations, where every later sweep reuses them — so "once per address" matches the shipped code (see also commit 432e24bcf0 in this PR).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -923,13 +989,14 @@ the four families disjoint (and disjoint from the empty main path). Nesting the
...
-sibling paths, alongside the unchanged withdrawal path:
+**Who signs what.** A first sweep uses three independent signatures on two
+unrelated sibling paths, alongside the unchanged withdrawal path:
 
 | Signature | Signed with derivation path | Purpose |
 |---|---|---|
-| EIP-7702 authorization tuple `(chain_id, S, nonce)` over `keccak256(0x05 ‖ rlp(...))` | the deposit EOA, `[1\|2, p, s]` | `ecrecover` must yield `addr(p, s)`, so the delegation designator installs on the deposit EOA |
-| type-`0x04` sweep transaction (sender / gas payer) | the sweeper address, `[3]` | `R17`: sweeps are issued from the dedicated sweeper address |
+| EIP-7702 authorization tuple `(chain_id, S, nonce)` over `keccak256(0x05 ‖ rlp(...))` | the deposit EOA, `[1\|2, p, s]` | `ecrecover` must yield `addr(p, s)`, so the delegation designator installs on the deposit EOA (first sweep only) |
+| account attestation over `keccak256("ck-deposit-owner" ‖ ...)` (step 5) | the deposit EOA, `[1\|2, p, s]` | `ecrecover` in the delegate must yield `addr(p, s)`, so only the attested IC account can be credited |
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Moot after the shared-delegated-address decision in this PR (#11313): the schema-2 path is gone from the signing rows; the cost table counts one authorization and one attestation by the deposit key, both on the shared address.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1137,9 +1225,10 @@ logical × 3, each charged fully); reserved `max_response_bytes` ≈ 2 KB for sm
 ≈ 1 KB for a 20-address balance batch (the deployless batcher returns 32 bytes/result,
 ~5× less than a Multicall3 `aggregate3`), ≈ 16 KB for `eth_getLogs` — so a small logical call
 ≈ $0.00084 and an `eth_getLogs` ≈ $0.0027 (the base dominates until responses grow). ETH
-at $2'500 and the sweep-gas figures from the
-[demo](deposit_from_cex_demo/README.md) (variant B: 82'207 gas for a single sweep,
-≈ 42'000 gas per address in a batch of 20).
+at $2'500 and the measured sweep-gas figures (≈ 85'500 gas for a single attested
+first sweep — 82'207 before the ≈ 3k attestation check — and ≈ 62'350 gas per
+first-sweep deposit in a per-token batch: 10 deposits of one token measured at
+623'360 gas).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Addressed in this PR (#11313): the cost inputs now use the attested harness' batch-entry figures (98'075 gas for a batch of one, 60'975/deposit at ten, 59'674/deposit at twenty) and the dollar figures follow from those.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -204,116 +205,162 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
+    CEX->>T: USDT.transfer(deposit address, 250)
+    Note over D,T: the "250 USDT at the deposit address" is a storage slot inside the<br/>token contract — the EOA itself stays untouched (no ETH, no nonce, no code)
+    loop while the (deposit address, USDT) pair is armed
+        Minter->>T: balance scan of all registered pairs<br/>(deployless batcher eth_call, latest block)
+    end
+    Note over Minter: balance detected: queue (deposit address, USDT) for sweeping.<br/>Scheduling hint only, no finality needed —<br/>a reorged deposit just wastes the sweep's gas
+    Note over Minter,D: sign with the deposit address' key (tECDSA): the attestation of the<br/>account (principal, subaccount) for every sweep, plus an EIP-7702 authorization<br/>(delegate = SweeperContract) on the first sweep of this address only
+    Note over Minter,Sw: sweeper address already funded:<br/>ckETH burned from the fee account at funding time (R14)
+    Sw->>S: sweep tx on the sweeper's own nonce lane (R17), signed with path [3]:<br/>type 0x04 (0x02 once delegated), one token per tx, up to 20 deposits:<br/>sweepErc20Batch([(deposit address, principal, subaccount, attestation)], [USDT])
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed: the shipped cap is MAX_DEPOSITS_PER_SWEEP = 10 while the doc says 20 in these spots (the ≈ 20 figure is the gas bound, not the policy). Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -645,57 +731,81 @@ is deducted: the full amount goes to the beneficiary. Fees are flat and
...
+storage-less sweeper delegate (`CkSweeperAttested`,
+[below](#sweeper-delegate-contract)). Sweep
+transactions are sent from the **dedicated sweeper address** (tECDSA-derived
+with derivation path `[3u8]` — its own schema tag, no account components),
+whose nonce sequence is independent of the main address' so that a stuck sweep
+can never delay a withdrawal (`R17`); the sweeper address holds only gas money
+(funded by ckETH withdrawals from the fee account, step 0) while swept funds
+always land at the main address (`R6`). No deposit address is ever pre-funded
+for gas: an ERC-20 deposit address never holds ETH at all, and a Phase 2 ETH
+address pays its sweep's gas out of the deposited balance itself (below).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed — the enforceable invariant is that the minter never pre-funds it and sweeps need no ETH there. Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -645,57 +731,81 @@ is deducted: the full amount goes to the beneficiary. Fees are flat and
...
+  always the deployed delegate instance `SweeperContract`, whose batch entry
+  point fans out to every deposit address in the batch.
+* The **send task** drives the sweeper pipeline through the same
+  create → sign → send → resubmit → finalize state machine as user withdrawals,
+  but signs with the sweeper derivation path `[3u8]` and tracks the sweeper
+  address' own transaction count (`R17`). A freshly enqueued sweep kicks it
+  immediately rather than waiting for its next tick: the mint follows the sweep
+  (step 4), so every interval spent waiting is crediting latency a user sees.
+  Confirmation is via transaction receipt, like withdrawals, emitting audit
+  events (`R5`, `R8`).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed: enqueue ends at AcceptedSweepRequest and the send task runs on its own 6-minute timer (PROCESS_SWEEPER_TRANSACTIONS_INTERVAL), so the immediate kick is design intent, not shipped behavior. Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1054,19 +1257,21 @@ Notes:
 * Supported tokens are assumed standard: non-fee-on-transfer and non-rebasing, so
   `balanceOf(address(this))` equals the amount `depositErc20` transfers. This is the
   same assumption as today's helper-contract deposit flow.
-* The demo's [`CkSweeperViaHelper.sol`](deposit_from_cex_demo/contracts/CkSweeperViaHelper.sol)
-  is this contract with the sweeper address approximated by a single minter EOA.
+* The integration harness deploys exactly this contract,
+  [`CkSweeperAttested.sol`](../minter/tests/deposit_from_cex_demo/CkSweeperAttested.sol).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed dead (the contract lives at minter/CkSweeperAttested.sol). Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1054,19 +1257,21 @@ Notes:
...
 
 ### Test plan
 
 A runnable end-to-end demonstration of the sweep mechanism (unfunded deposit EOAs,
 plain USDT-style transfers, single and batched type-`0x04` sweep transactions with gas
 paid by the minter, gas assertions) against a local dev node (any post-Pectra
-version) is available in [`deposit_from_cex_demo/`](deposit_from_cex_demo/README.md).
-It exercises both sweep variants — including variant B against the *real*
+version) is available in
+[`deposit_from_cex_demo/`](../minter/tests/deposit_from_cex_demo/).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed — the runnable harness is the sibling deposit_from_cex_demo.rs. Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1182,35 +1389,150 @@ scan, and of addresses sharing its sweep):
...
+[`CkSweeperViaHelper.sol`](../minter/tests/deposit_from_cex_demo/CkSweeperViaHelper.sol)
+implements this caller-gated variant and is kept for comparison.
+
+### Address layout across asset classes (step 1)
+
+Decided: per-asset addresses, introduced with Phase 2.
+
+| Variant | Pros | Cons |
+|---|---|---|
+| **Per-asset addresses** (chosen): ERC-20 address (schema 1, delegated once, permanently) + ETH address (schema 2, never delegated) | ETH address never has code → fixed-21'000-gas CEX withdrawals always work (`R12`), no failure window; ETH sweeps need no EIP-7702 at all (deposit pays its own gas, 21'000 gas, cheapest possible); ERC-20 delegation stays one-signature-ever | Two addresses per account to register/scan; user must use the right address per asset (matches CEX per-asset UX; wrong-asset deposits recoverable, see Non-goals) |
```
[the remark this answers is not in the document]
gregorydemay: 🤖 In the current text the 21'000-gas sweep claim survives only in the per-asset fallback row of the variants table, where it is correct (a key-signed plain transfer); the chosen design's ETH sweep goes through sweepEthBatch/depositEth and is costed as such. Any remaining ambiguity will be addressed with #11313, which aims at improving the docs.

gregorydemay:
🤖 Collected doc fixes deferred to this PR from the review threads here and on #11449:

**Doc ↔ shipped-code mismatches**
1. **Sweep batch size**: the shipped cap is `MAX_DEPOSITS_PER_SWEEP = 10` (`minter/src/sweep/mod.rs`), the doc says 20 in four spots — the enqueue-task paragraph ("Up to 20 deposits ride one transaction"), the sweep sequence-diagram note ("up to 20 deposits"), the EIP-7702 primer ("`N ≈ 20` per the sweeping policy in step 5" — ≈20 is the gas bound, not the policy), and the cost table's `B=20` scenario rows (as `B=10`: 2.1 sigs → \$0.075, outcalls ≈ \$0.0032 / \$0.0084, IC subtotal \$0.078 / \$0.083, totals \$0.23 / \$0.24; footnote "B=20 ≈ 40% less" → B=10).
2. **Immediate send-task kick**: enqueue ends at `AcceptedSweepRequest`; the send task runs on its own timer (`SWEEP_ENQUEUE_INTERVAL` = 60 s, `PROCESS_SWEEPER_TRANSACTIONS_INTERVAL` = 6 min). Flag "a freshly enqueued sweep kicks it immediately" as design intent / not yet implemented, mirroring the R3-gap disclosure style.

**Fallout from the `receive()` guard (#11449)**
3. **"empty `receive()`" is stale**: the delegate's `receive()` carries the `SELF` guard since dfinity/ic@000660b461 — update the occurrences around the mainnet data point and the "an empty payable `receive()` — guarded to revert…" note, which reads as a contradiction.
4. **Gas figures**: the mainnet experiment ran against the deployed, guard-less bytecode (21'055 gas); the guarded contract measures 21'095 on anvil. Clarify next to the data point (harmless for the tested exchanges — 207'128 / 31'830 limits — but the numbers should be attributed to the right bytecode).
5. **Misplaced bullet**: the "delegate carries an empty payable `receive()` + attested `sweepEth`/`sweepEthBatch`" bullet sits in the notes under the `CkSweeperViaHelper` listing, which has neither — scope it to `CkSweeperAttested`, reconcile with the nearby "a plain ETH transfer to the now-delegated `D` reverts at the sender, since `S` has no `receive()`" line, and point the superseded "Breaks ETH deposits entirely" variants-table row at the paragraph that supersedes it.

**Wording / links**
6. **Sweeper-address phrasing**: "never holds anything beyond gas money" is not enforceable (anyone can send ETH there); state the enforceable invariant — the minter never funds it with more than gas money.
7. **Dead link**: `CkSweeperAttested.sol` lives at `minter/CkSweeperAttested.sol`, not under the demo directory.
8. **Test-plan link**: the runnable harness is `minter/tests/deposit_from_cex_demo.rs`; the linked directory holds only the support contracts.

**Open questions to record**
9. **Compliance scope**: `receive()` on the ERC-20 delegate extends the unscreened-ETH-funder weakening (the helper event's `from_address` is the deposit EOA — exactly the field the blocklist checks — and ETH has no `Transfer` log to recover the real sender) from Phase 2 to Phase 1 addresses. Make that Phase-1 scope explicit in the "to review with compliance before Phase 2 ships" paragraph so it cannot slip into a Phase 1 sign-off unnoticed.
10. **`rescueEth()` vs the `SELF` guard**: the alternative proposed on #11449 (zero hot-path gas, locked ETH becomes recoverable, needs an immutable recovery principal chosen at deployment) should be recorded as an open decision for the finalized Phase 2 delegate.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -744,46 +843,116 @@ digest = keccak256("ck-deposit-owner" ‖ chain_id ‖ helper_address
...
+  carries a minimal payable `receive()` (nothing but the guard rejecting sends
+  to the implementation itself), so the send lands whether or not the
+  address is already delegated: a code-less address takes the usual 21'000 gas,
+  a delegated one 21'095 on anvil (the mainnet experiment confirmed 21'055
+  with its guard-less bytecode), within
+  the 2'300-gas `transfer`/`send` stipend of contract-batched CEX withdrawals.
+  A send with a *fixed* 21'000 limit to an already-delegated address fails on
+  the sender side — funds never leave the exchange (`R12`).
+* **Detection is by balance delta, not logs.** A plain ETH send emits no
+  `Transfer` log, so crediting follows the finalized balance delta (`R11`) and
```
mbjorkqvist: This bullet credits the user the variant-A way, while the rest of the section credits the variant-B way.

Two sentences above, the section says everything carries over from ERC-20 including "the helper event and the mint pipeline", and the diagram below credits from the helper event:

```
Minter->>H: eth_getLogs(helper contract, up to finalized)
Minter->>Ledger: mint ckETH to the user's account (existing pipeline, unchanged)
```

Step 3 makes this bullet's claim too (line 629: "the finalized balance delta *is* the observation (`R11`) — there are no logs to confirm against"), two sentences after stating the opposite rule: "A balance is only ever a trigger, never a source of truth".

"No logs to confirm against" was true when ETH had its own never-delegated address swept by a plain transfer. It is not true of the design this PR documents: `sweepEth` forwards through `depositEth`, so the helper emits `ReceivedEthOrErc20` exactly as for ERC-20 — `CkSweeperAttested.sol`:

```solidity
ICkDeposit(HELPER).depositEth{value: balance}(principal, subaccount);
```

This is not only wording: the two paths credit different amounts and dedup differently. Variant A credits the balance observed, variant B credits what the sweep delivered — and since `sweepEth` forwards the *whole* balance, those diverge whenever more ETH arrives between the scan and the sweep. Variant A also needs dedup "per balance observation" (`R11`) rather than by `(tx hash, log index)` (`R2`), i.e. the second crediting path step 4 rejects as "the highest-risk part of the feature".

Suggest: detection by balance delta, crediting from the sweep's helper event like ERC-20 — and drop "there are no logs to confirm against" at line 629.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -160,8 +163,10 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
   account, exactly once per balance observation (monotone accounting: total credited
```
mbjorkqvist: `R11` is stranded by the shared-address decision. It mandates minting "the difference minus the deposit fee ... exactly once per balance observation", which is variant-A crediting; the chosen design mints ETH from the sweep's own finalized helper event through the unchanged pipeline (step 4, and the ETH-flow diagram), deduped by `(tx hash, log index)`.

As written it is a requirement no part of the chosen design satisfies. Worth restating around the helper event, with the balance delta as the detection trigger only — or marking it superseded.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -744,46 +843,116 @@ digest = keccak256("ck-deposit-owner" ‖ chain_id ‖ helper_address
...
+        participant H as Helper (deposit helper)
+    end
+    participant Ledger as ckETH ledger
+
+    User->>Minter: deposit_eth()
+    Minter-->>User: the same deposit address as for ERC-20 (schema 1)
+    User->>CEX: withdraw ETH to the deposit address
+    CEX->>D: plain ETH send — 21'000 gas while code-less, 21'095 into the<br/>delegate's minimal receive() once delegated (R12)
+    loop while the (address, ETH) pair is armed
+        Minter->>D: finalized ETH balance, read in the same<br/>deployless-batcher call as the ERC-20 scans
```
mbjorkqvist: One `eth_call` carries one block parameter, so this cannot be "the same ... call as the ERC-20 scans" *and* finalized — diagram 1 (line 316) labels that call "latest block".

The shipped scan is on latest, and pins one block number for the whole batch:

- `src/balance_scan/mod.rs:57` — `let latest_block = read_state(|s| s.latest_block_height)`
- `src/balance_scan/mod.rs:333` — `block: Some(evm_rpc_types::BlockTag::from(block))`, commented "Pinned to the refreshed latest block height so every provider reads the same block"
- `src/deposit.rs:213` — `latest_block_height` is refreshed from `BlockTag::Latest`

Latest is also the deliberate choice: step 4 names sweeping on `latest`-block observations as the latency mitigation, since "a reorged deposit only wastes the sweep's gas". Under variant B that reasoning covers ETH too — the finality that matters is the helper event's, which this diagram already scrapes "up to finalized" two lines down.

Suggest dropping "finalized" here and at line 629. Same root cause as the crediting comment above.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -826,105 +995,169 @@ offline for free, and *anyone* can embed it in a transaction and pay the gas —
...
+with three tuples in the authorization list — Transaction 2 with more items:
+all three designators are installed in phase 2, then the batch loop runs each
+inner sweep in its own `Depositᵢ` context. One batch never mixes tokens
+(step 5).
+
+Measured on the minter's per-token batches: 609'750 gas for ten first-sweep
+deposits of one token (60'975 each) and 1'193'491 for twenty (59'674 each),
+against 98'075 for a batch of one. Batching pays once for the 21'000 base, the cold accesses to
+`Token` and the first (zero→nonzero) write to `MainAddress`' token balance slot,
+while each extra address adds only its 25'000 authorization, a warm inner call
```
mbjorkqvist: This explanation came from the variant-A measurements, where it was right (118'876 gas for 3 addresses vs 3 x 66'854 standalone, so ~26k marginal). Against the attested figures quoted just above, the marginal cost is roughly double what the sentence claims:

| batch size | 1 | 10 | 20 |
|---|---|---|---|
| type 0x04 (first sweep) | 98'075 | 609'750 | 1'193'491 |
| type 0x02 (already delegated) | 66'395 | 429'750 | 833'491 |

- first sweep: (609'750 - 98'075)/9 = **~56'900**, (1'193'491 - 609'750)/10 = **~58'400**
- already delegated, i.e. no authorization at all: (429'750 - 66'395)/9 = **~40'400**

So the 25'000 authorization is the smaller half. The dominant marginal cost is the per-deposit sweep work itself — the inner call, `balanceOf`, `approve`, `depositErc20`'s `transferFrom`, the event, plus the attestation `ecrecover` and its 65 bytes of calldata. The doc quantifies that correctly elsewhere: "+15'353 gas per deposit for the approve + `depositErc20` + event", in the variants table.

Suggest either keeping the sentence with the variant-A numbers it describes, or rewriting it for these: batching amortizes the 21'000 base, the cold `Token` accesses and the first zero-to-nonzero write to the main address' balance slot, while each extra deposit pays ~40k of its own sweep work plus, on a first sweep, 25'000 for its authorization.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -512,7 +569,7 @@ wallet. The design must absorb:
 
 No variants: the CEX side is not under our control.
```
mbjorkqvist: Line 564, just above, still ends: "Native ETH sends execute recipient code — hence the per-asset address layout of step 1."

Step 1 no longer has a per-asset layout — this PR flips it to "one shared, permanently delegated address across assets", with the split demoted to a fallback in Discussed Alternatives. The sentence is verbatim from master, but this PR is what made it stale, and it points the reader at a decision that no longer exists.

Suggest ending the clause on the mechanism that replaced it: the delegate's payable `receive()` is now what makes a shared, delegated address safe for ETH.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -744,46 +843,116 @@ digest = keccak256("ck-deposit-owner" ‖ chain_id ‖ helper_address
...
+lane and `R14`-prepaid gas, the helper event and the mint pipeline. Only three
+things differ:
+
+* **The deposit is a plain ETH send, not a token transfer.** The delegate
+  carries a minimal payable `receive()` (nothing but the guard rejecting sends
+  to the implementation itself), so the send lands whether or not the
+  address is already delegated: a code-less address takes the usual 21'000 gas,
+  a delegated one 21'095 on anvil (the mainnet experiment confirmed 21'055
+  with its guard-less bytecode), within
+  the 2'300-gas `transfer`/`send` stipend of contract-batched CEX withdrawals.
```
mbjorkqvist: Read literally, this says 21'095 <= 2'300. They are different budgets:

- **21'095** is a whole top-level transaction: 21'000 intrinsic plus ~95 gas actually executing the delegate's `receive()`.
- **2'300** is what `.transfer()` / `.send()` forwards to the callee for execution only, in an internal call from a contract. The intrinsic 21'000 is not part of that budget — the outer transaction already paid it.

The intended claim is true (the ~95 gas of execution fits in 2'300), but the number that has to fit is never stated. Suggest splitting the two.

Two related points:

1. Step 1 (line 549) says contract-batched paths "fit the minimal `receive()` **by construction** but were not exercised on mainnet". Not on mainnet — but they *are* exercised on anvil, by a test written for exactly this: `a_stipend_transfer_from_a_contract_reaches_a_delegated_deposit_address` (`minter/tests/deposit_from_cex_demo.rs:717`), driving `StipendForwarder.sol` ("Forwards ETH with Solidity's `transfer`, i.e. a 2'300-gas stipend — the worst-case sender profile of a contract-batched CEX withdrawal"). Worth citing, because "by construction" is doing real work here and is not self-evident: the call also has to resolve the delegation designator and load the delegate's code, and whether that is charged as a cold account access (2'600, which would blow the stipend) is exactly the kind of thing to settle with a test rather than an argument.
2. 21'095 is quoted twice (here and in `R12`) but is not pinned by anything: the anvil test only prints it and asserts `> 21_000` (`deposit_from_cex_demo.rs:617-623`), unlike every other gas figure in the doc, which traces to an asserted constant. Either assert it or mark it indicative.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -678,57 +745,89 @@ is deducted: the full amount goes to the beneficiary. Fees are flat and
...
+
+Each ERC-20 deposit EOA signs (threshold ECDSA) **one** EIP-7702
+authorization — lazily, together with its first sweep, never at registration
+(`R13`) — delegating its code to `SweeperContract`, a single immutable,
+storage-less sweeper delegate (`CkSweeperAttested`,
+[below](#sweeper-delegate-contract)). Sweep
+transactions are sent from the **dedicated sweeper address** (tECDSA-derived
+with derivation path `[3u8]` — its own schema tag, no account components),
+whose nonce sequence is independent of the main address' so that a stuck sweep
+can never delay a withdrawal (`R17`); the sweeper address holds only gas money
```
mbjorkqvist: nit: "holds only gas money" was raised earlier, and was addressed in a couple of other locations in the doc, but not here (anyone could potentially transfer funds to the sweeper address).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -194,129 +199,140 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
     provides no sender information without trace APIs; Phase 2 compliance screening is
     therefore weaker for ETH than for ERC-20 (see step 3).
-  * An account's ETH deposit address (Phase 2) differs from its ERC-20 one, matching
-    the per-asset deposit-address UX of exchanges; sending the wrong asset class to
-    an address is not credited automatically, but funds always remain recoverable
-    (key-controlled addresses, `R12` guarantees no loss on the sender side).
+  * An account's ETH deposits (Phase 2) reuse its ERC-20 deposit address, so there
+    is no wrong-address case between the two asset classes; an *unsupported* ERC-20
+    token sent there is not credited automatically, but funds always remain
+    recoverable (key-controlled addresses, `R12` guarantees no loss on the sender
```
mbjorkqvist: `R12` does not support this. It is about a plain ETH transfer sent with a fixed 21'000 gas limit failing at the sender; it says nothing about an unsupported ERC-20 token landing at the address, which is what this bullet is about after the rewrite. `R4` is the requirement that covers "not credited, nothing lost".

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1074,38 +1359,50 @@ contract CkSweeperViaHelper {
...
-  would be locked — plus attested `sweepEth`/`sweepEthBatch` entry points that
-  forward the balance through the helper's `depositEth`
-  ([#11449](https://github.com/dfinity/ic/pull/11449)). This keeps the
-  single-shared-address variant open for Phase 2 (see the mainnet data point in
-  step 1); under the decided per-asset layout the schema-2 ETH addresses are
-  never delegated and sweep via a key-signed `depositEth` call (step 5), and a
-  plain ETH send to a delegated ERC-20 address is accepted as a recoverable
-  wrong-asset deposit (see Non-goals) rather than bounced.
+* The delegate's payable `receive()` holds nothing but the guard reverting on
+  the implementation address itself — no attestation can ever recover to the
```
mbjorkqvist: The real contract's comment carries a caveat this copy drops — `CkSweeperAttested.sol`:

> selfdestruct and coinbase payments can still deposit unsweepable ETH — the guard closes the only path a plain send takes

Worth keeping. As written, the doc says ETH at the implementation "would be locked forever" and the guard prevents it, which reads as airtight, whereas the guard closes only the plain-send path. The caveat also strengthens the `rescueEth()` open decision recorded two bullets down.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1074,38 +1359,50 @@ contract CkSweeperViaHelper {
...
   `balanceOf(address(this))` equals the amount `depositErc20` transfers. This is the
   same assumption as today's helper-contract deposit flow.
-* The demo's [`CkSweeperViaHelper.sol`](deposit_from_cex_demo/contracts/CkSweeperViaHelper.sol)
-  is this contract with the sweeper address approximated by a single minter EOA.
+* The integration harness deploys exactly this contract,
+  [`CkSweeperAttested.sol`](../minter/CkSweeperAttested.sol).
 
 ### Test plan
 
 A runnable end-to-end demonstration of the sweep mechanism (unfunded deposit EOAs,
```
mbjorkqvist: The harness summary does not mention the ETH coverage the new ETH-flow section rests on: `sweepEth` / `sweepEthBatch` end to end, the forged-attestation-for-ETH rejection, `the_sweeper_implementation_rejects_plain_eth`, and the stipend forwarder (`deposit_from_cex_demo.rs:717`). Now that the ETH flow is a decided section rather than future work, worth listing.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -954,27 +1187,27 @@ the minter can `sign_with_ecdsa` for it under the same `path`. The paths are
 
 ```
 master key (+ root chain code)
-├── []          → minter main address        (MAIN_DERIVATION_PATH; withdrawals, R6 destination)
-├── [1, p, s]   → ckERC20 deposit EOA addr(p, s)
-├── [2, p, s]   → ckETH   deposit EOA addr(p, s)   (Phase 2)
-└── [3]         → dedicated sweeper address   (R17; its own schema tag, no account components)
+├── []                            → minter main address      (MAIN_DERIVATION_PATH; withdrawals, R6 destination)
+├── [1, principal, subaccount]    → deposit address, one per IC account (ERC-20 and, Phase 2, ETH)
+└── [3]                           → dedicated sweeper address (R17; its own schema tag, no account components)
```
mbjorkqvist, line 1196: nit: The tree now omits `[2]`, which is still defined in the code as [`CKETH_DEPOSIT_SCHEMA_TAG`](https://github.com/dfinity/ic/blob/c5230771874481c283fe5c23361a320fed07b66d/rs/ethereum/cketh/minter/src/deposit_address/mod.rs#L49). Would it make sense to keep a mention here, e.g., say it is reserved?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -744,46 +843,116 @@ digest = keccak256("ck-deposit-owner" ‖ chain_id ‖ helper_address
...
+  to the implementation itself), so the send lands whether or not the
+  address is already delegated: a code-less address takes the usual 21'000 gas,
+  a delegated one 21'095 on anvil (the mainnet experiment confirmed 21'055
+  with its guard-less bytecode), within
+  the 2'300-gas `transfer`/`send` stipend of contract-batched CEX withdrawals.
+  A send with a *fixed* 21'000 limit to an already-delegated address fails on
+  the sender side — funds never leave the exchange (`R12`).
+* **Detection is by balance delta, not logs.** A plain ETH send emits no
+  `Transfer` log, so crediting follows the finalized balance delta (`R11`) and
+  sender screening is weaker than for ERC-20 (step 3).
```
mbjorkqvist: IIUC, this is the "Variant A" approach, but what is now implemented is "Variant B"? The text in `R11` (not touched by this PR) also seems outdated, in particular:

> the minter mints the difference minus the deposit fee to the associated account, exactly once per balance observation

Should this be something like "once a balance over the minimum amount is detected, the entire balance is swept and credited"?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -744,46 +843,116 @@ digest = keccak256("ck-deposit-owner" ‖ chain_id ‖ helper_address
...
+    end
+    participant Ledger as ckETH ledger
+
+    User->>Minter: deposit_eth()
+    Minter-->>User: the same deposit address as for ERC-20 (schema 1)
+    User->>CEX: withdraw ETH to the deposit address
+    CEX->>D: plain ETH send — 21'000 gas while code-less, 21'095 into the<br/>delegate's minimal receive() once delegated (R12)
+    loop while the (address, ETH) pair is armed
+        Minter->>D: finalized ETH balance, read in the same<br/>deployless-batcher call as the ERC-20 scans
+    end
```
mbjorkqvist: Shouldn't this be "latest" rather than "finalized"? The `eth_call` calls are made with the `latest` block as parameter. The later mint event is based on the finalized balance. Same for the sentence on lines 628-630 (not touched in this PR).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -826,105 +995,169 @@ offline for free, and *anyone* can embed it in a transaction and pay the gas —
...
+all three designators are installed in phase 2, then the batch loop runs each
+inner sweep in its own `Depositᵢ` context. One batch never mixes tokens
+(step 5).
+
+Measured on the minter's per-token batches: 609'750 gas for ten first-sweep
+deposits of one token (60'975 each) and 1'193'491 for twenty (59'674 each),
+against 98'075 for a batch of one. Batching pays once for the 21'000 base, the cold accesses to
+`Token` and the first (zero→nonzero) write to `MainAddress`' token balance slot,
+while each extra address adds only its 25'000 authorization, a warm inner call
+and a transfer that earns the slot-clearing refund. Operational notes: a
```
mbjorkqvist: This explanation came from the variant-A measurements, where it was right (118'876 gas for 3 addresses vs 3 x 66'854 standalone, so ~26k marginal). Against the attested figures quoted just above, the marginal cost is roughly double what the sentence claims:

| batch size | 1 | 10 | 20 |
|---|---|---|---|
| type 0x04 (first sweep) | 98'075 | 609'750 | 1'193'491 |
| type 0x02 (already delegated) | 66'395 | 429'750 | 833'491 |

- first sweep: (609'750 - 98'075)/9 = **~56'900**, (1'193'491 - 609'750)/10 = **~58'400**
- already delegated, i.e. no authorization tuple at all: (429'750 - 66'395)/9 = **~40'400**

The second row settles it on its own: with zero authorization cost, each extra deposit still costs ~40k, so the 25'000 authorization is the smaller half. What dominates is the sweep work the batch runs once *per deposit address*: the outer loop's external call plus resolving that address' delegation designator to load the delegate's code; the attestation `ecrecover` (~3k) with its 65-byte signature and `SweepItem` calldata; a cold `balanceOf` slot read (the token contract is warm after the first deposit, but each address has its own balance slot); an `approve` SSTORE into a fresh zero-valued allowance slot; `transferFrom` clearing that address' balance slot (the refund this sentence does mention), writing the main address' slot and clearing the allowance; and the `ReceivedEthOrErc20` LOG. Of that, only "a warm inner call" appears in the current text.

Suggest either keeping the sentence with the variant-A numbers it describes, or rewriting it for these: batching amortizes the 21'000 base, the cold `Token` accesses and the first zero-to-nonzero write to the main address' balance slot, while each extra deposit pays ~40k of its own sweep work plus, on a first sweep, 25'000 for its authorization.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -194,129 +199,140 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
...
     therefore weaker for ETH than for ERC-20 (see step 3).
-  * An account's ETH deposit address (Phase 2) differs from its ERC-20 one, matching
-    the per-asset deposit-address UX of exchanges; sending the wrong asset class to
-    an address is not credited automatically, but funds always remain recoverable
-    (key-controlled addresses, `R12` guarantees no loss on the sender side).
+  * An account's ETH deposits (Phase 2) reuse its ERC-20 deposit address, so there
+    is no wrong-address case between the two asset classes; an *unsupported* ERC-20
+    token sent there is not credited automatically, but funds always remain
+    recoverable (key-controlled addresses, `R12` guarantees no loss on the sender
+    side).
```
mbjorkqvist: Shouldn't this guarantee be due to `R4` rather than `R12`?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -487,14 +543,15 @@ to the EOA standing in for the minter,
...
+delegation designator, so the feared "ETH deposits are impossible" failure of a
+permanently delegated address did not materialize. This is the evidence behind
+the shared-address decision above. Exchange behavior stays heterogeneous —
+contract-batched withdrawal paths, where only the 2'300-gas stipend is
+forwarded, fit the minimal `receive()` by construction but were not exercised on
+mainnet — hence the per-asset fallback kept in
+[Discussed Alternatives](#discussed-alternatives).
 
-### Step 2: Withdraw from the CEX to the deposit address
+#### Step 2: Withdraw from the CEX to the deposit address
```
mbjorkqvist, line 553: I believe the second bullet below is no longer accurate:

> * ERC-20 transfers always emit a `Transfer` log and never execute recipient code, so a delegated ERC-20 deposit address is harmless to the sender. Native ETH sends execute recipient code — hence the per-asset address layout of step 1.

Now there's just a single shared address across assets?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1074,38 +1359,50 @@ contract CkSweeperViaHelper {
...
+  Phase 2 accept plain ETH at the same delegated address and sweep it with the
+  same contract (see [The ETH deposit flow](#the-eth-deposit-flow)); it is kept
+  that minimal so it fits the 2'300-gas `transfer`/`send` stipend of
+  contract-batched CEX withdrawals, and detection is balance-based, so it emits
+  no event. **Open decision** for the finalized Phase 2 delegate (proposed on
+  [#11449](https://github.com/dfinity/ic/pull/11449)): replace the guard with
+  an unguarded `receive()` plus a `rescueEth()` on the implementation — zero
+  extra gas on the hot deposit path and ETH sent to the implementation becomes
+  recoverable instead of bounced, at the price of an immutable recovery
+  principal chosen at deployment.
```
mbjorkqvist: Do we still want to keep this as an "Open decision"? I think your response [here](https://github.com/dfinity/ic/pull/11449#discussion_r3934273811) is valid, and since the contract is deployed, it's not straight-forward to change this anymore?

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1074,38 +1359,50 @@ contract CkSweeperViaHelper {
...
 * Reentrancy is moot: no storage, and funds can only move through the helper
   toward the minter's main address.
 * Supported tokens are assumed standard: non-fee-on-transfer and non-rebasing, so
   `balanceOf(address(this))` equals the amount `depositErc20` transfers. This is the
   same assumption as today's helper-contract deposit flow.
-* The demo's [`CkSweeperViaHelper.sol`](deposit_from_cex_demo/contracts/CkSweeperViaHelper.sol)
-  is this contract with the sweeper address approximated by a single minter EOA.
+* The integration harness deploys exactly this contract,
+  [`CkSweeperAttested.sol`](../minter/CkSweeperAttested.sol).
```
mbjorkqvist, line 1395: Related to the "Test plan" below, should we explicitly mention something regarding the ETH flow?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fixed in 52d24ea954: step 5 now reads "the sweeper address is funded by the minter with only gas money" — the enforceable invariant, matching the earlier fixes.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fixed in 52d24ea954: the tree keeps a `[2, principal, subaccount]` line marked reserved (CKETH_DEPOSIT_SCHEMA_TAG — defined in code, never derived under the shared-address decision).

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Right on both counts — that sentence and R11 described variant-A mint-on-observation. Fixed in 52d24ea954: the bullet now says the latest-block balance delta only triggers the sweep while the mint follows the sweep's finalized helper event (step 4), and R11 is rewritten along your line: once a balance at or above the minimum is detected, the entire balance is swept and the swept amount minus the deposit fee credited exactly once per sweep (dedup by the helper event, monotone accounting kept).

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Confirmed — the scan pins to the latest block (balance_scan/mod.rs:331). Fixed in 52d24ea954 in both spots: the ETH diagram loop reads "latest-block ETH balance" and the step-3 sentence now says the latest-block delta is only the sweep trigger, with finality entering via the helper-event mint.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Agreed — R4 is exactly the not-credited-but-recoverable requirement, and "no loss on the sender side" was R12-flavored (an ERC-20 transfer does leave the sender). Fixed in 52d24ea954: the bullet now cites R4 and drops the sender-side phrase.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fixed in 52d24ea954: the bullet now reads "hence the delegate's minimal receive() and the fixed-21'000-gas caveat (R12)" instead of the stale per-asset rationale.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Closed in 52d24ea954: the note is now "Decided: keep the guard", recording rescueEth as the discarded alternative with the rationale from #11449 — no extra permissionless entry point to review, no recovery principal to govern, and the guard's ~40 gas per deposit is negligible.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Added in 52d24ea954: the test plan now lists the demo's ETH coverage — plain sends to code-less and delegated addresses (including the 2'300-gas stipend path via StipendForwarder.sol and the receive() gas assertion) and permissionless sweepEth/sweepEthBatch sweeps by a non-minter relayer.

PR #11319, opened 2026-08-25
feat(cketh): remember the attestations the minter has signed

gregorydemay:
## Why

- A deposit address' attestation binds it to the account it credits and to one deposit-helper deployment, and nothing else: no token, no amount, no nonce, no expiry.
- It is therefore signed once and valid forever, yet every sweep re-signed it: one threshold-ECDSA signature per deposit, per round, and per token once batches are grouped by token.
- The delegation is already treated as one-time work; the attestation was the artefact the minter kept paying for.

## What

- The minter records each attestation it signs and reuses it afterwards.
- Its own event, not a field on the sweep: signing costs cycles and can fail, and the enqueue can still bail after it succeeds, so the signature becomes durable the moment it exists rather than only if a sweep is enqueued.
- Replay keys them by the helper each attestation names, which is also what retires them — a new helper deployment simply misses, so no invalidation rule is needed.
- The lookup is on `State`, beside the chain and helper the attestation is derived from, so a caller cannot consult the wrong deployment.

## Candid compatibility

- Needs the `CI_OVERRIDE_DIDC_CHECK` label: the returned `Event` variant gains a case, which Candid subtyping forbids.
- The case is additive and nothing has emitted it, so existing clients see the same events they saw before.


rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -189,6 +189,27 @@ pub enum EventType {
     /// The minter burned ckETH from its fee subaccount to top up the sweeper address with gas.
     #[n(27)]
     AcceptedSweeperFundingRequest(#[n(0)] EthWithdrawalRequest),
+    /// A deposit address attested to the account it credits. Signing costs a threshold-ECDSA
+    /// signature and can fail, so it is recorded on its own rather than with the sweep that
+    /// needed it: the attestation outlives that sweep and every later one reuses it.
+    #[n(33)]
```
gregorydemay: That's confusing it's number 33, between 27 and 28. I would move it at the end of the enum (last added variant)

rs/ethereum/cketh/minter/src/state/event.rs
```diff
@@ -189,6 +189,27 @@ pub enum EventType {
...
+        #[n(0)]
+        chain_id: u64,
+        /// The deposit helper the attestation names. An attestation is only valid against this
+        /// deployment, which is why replay keys them by it.
+        #[n(1)]
+        deposit_helper: Address,
+        #[cbor(n(2), with = "icrc_cbor::principal")]
+        owner: Principal,
+        #[cbor(n(3), with = "minicbor::bytes")]
+        subaccount: Option<[u8; 32]>,
```
gregorydemay: This is exactly `AttestationRequest`, why not reuse it 1:1 in the event?

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -61,6 +62,11 @@ impl MintedEvent {
 #[derive(Clone, PartialEq, Debug)]
 pub struct State {
     pub ethereum_network: EthereumNetwork,
+    /// Attestations the minter has signed, keyed by the deposit helper each one names and then by
+    /// the account it credits. An attestation binds an account to one helper deployment and never
+    /// expires, so a later sweep of the same address reuses it instead of paying for another
+    /// threshold-ECDSA signature; a new helper deployment simply misses.
+    pub attestations: BTreeMap<Address, BTreeMap<Account, TransactionSignature>>,
```
gregorydemay: - Shouldn't the attestation store be a `BTreeMap<AttestationRequest, Signature>`, otherwise how a signature would be retrieved efficiently?Shouldn't the attestation store be a `BTreeMap<AttestationRequest, Signature>`, otherwise how a signature would be retrieved efficiently?
- I would move it to the `AutomaticDeposits` struct

rs/ethereum/cketh/minter/src/state/event.rs
gregorydemay: 🤖 Agreed — moved to the end of the enum in cbab125bf8, after `FinalizedSweeperTransaction` (32).

rs/ethereum/cketh/minter/src/state/event.rs
gregorydemay: 🤖 Done in 37765b4864 — the event now carries the `AttestationRequest` itself:

```rust
AttestedDepositAddress {
    #[n(0)]
    request: AttestationRequest,
    #[n(1)]
    signature: TransactionSignature,
},
```

- The request gained `Ord`/`PartialOrd` and minicbor `Encode`/`Decode`, plus a small `cbor_account` module that splits the `Account` into owner + optional subaccount, the way every other event encodes one.
- `DepositAddressSchema` gained the same derives, so the event now also records which derivation path signed — one more field the digest depends on that could previously drift.
- The Candid mirror stays flat (it is the audit log) and grew a `schema : variant { CkErc20; CkEth }`.
- One consequence worth your call: `AttestationRequest::new` had to go back to `pub`, because `tests/dump_stable_memory.rs` is an external crate and reconstructs events from a Candid dump. The doc now says what `DepositAddress::new` says in the same situation — `State::attestation_request` is the only caller that composes a request, everything else reconstructs one the minter already built.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Both applied.

- 37765b4864 flattens the store to `BTreeMap<AttestationRequest, TransactionSignature>`. One lookup instead of two, and the key is exactly what was signed, so an attestation cannot be reused under a chain or a schema it was not signed for — the nested map silently ignored both. A new test covers the chain case; the fixture is Sepolia, and the old test hardcoded `chain_id: 1`, which the nested store simply never looked at.
- 2a87152cf8 moves it into `AutomaticDeposits`. `State::attestation(account)` stays where the sweeper wants it, building the request from the configuration in force and delegating the lookup.

rs/ethereum/cketh/minter/src/attestation/tests.rs
```diff
@@ -122,3 +122,101 @@ fn account() -> Account {
...
+            .log_scrapings
+            .set_contract_address(LogScrapingId::EthOrErc20DepositWithSubaccount, helper());
+        state
+    }
+
+    fn in_force(state: &State) -> AttestationRequest {
+        state
+            .attestation_request(account())
+            .expect("BUG: the deposit helper is configured")
+    }
```
gregorydemay: `in_force` seem like a very bad name. the account for which the attestation is intended for should also be an explicit parameter, otherwise a reader does not know who it is intended for.

rs/ethereum/cketh/minter/src/attestation/tests.rs
```diff
@@ -122,3 +122,101 @@ fn account() -> Account {
...
+        }
+    }
+
+    fn helper() -> Address {
+        super::helper()
+    }
+
+    fn account() -> Account {
+        super::account()
+    }
```
gregorydemay: just use test_fixtures.rs for common helpers, move them there if needed (if they are not already similar methods)

rs/ethereum/cketh/minter/src/attestation/tests.rs
gregorydemay: 🤖 Agreed on both counts — `in_force` is gone in c6b4d51338. There was no need for a helper at all: each test now builds the request where it uses it, naming the account explicitly.

```rust
let mut state = state_with_deposit_helper(deposit_helper());
let request = state
    .attestation_request(account())
    .expect("BUG: the deposit helper is configured");
```

The two negative tests already named the account; the `chain_id(state)` shim went the same way, inlined as `state.ethereum_network.chain_id()`.

rs/ethereum/cketh/minter/src/attestation/tests.rs
gregorydemay: 🤖 Done in c6b4d51338. Nothing similar existed in `test_fixtures.rs`, so four helpers moved there — none of them are about attestations:

- `account()` — an account a deposit address is derived for.
- `deposit_helper()` — the helper whose events name the account an address credits.
- `transaction_signature()` — a well-formed stand-in signature, of which the crate had several ad-hoc copies.
- `state_with_deposit_helper(helper)` — `initial_state()` with that helper configured, which `eth_logs/tests.rs` was also doing by hand; it now calls the fixture.

The `mod recorded` shims that just forwarded to the outer module are gone, and the golden digest is unchanged, so the fixtures carry the same values the local helpers did.

`request(chain_id, deposit_helper, account)` stayed in `attestation/tests.rs`: it pins the schema to `CkErc20`, which is a fact about attestations rather than a shared fixture.

rs/ethereum/cketh/minter/src/attestation/mod.rs
```diff
@@ -100,3 +118,38 @@ pub async fn sign_attestation(
...
+        e: &mut Encoder<W>,
+        ctx: &mut Ctx,
+    ) -> Result<(), minicbor::encode::Error<W::Error>> {
+        CborAccount {
+            owner: v.owner,
+            subaccount: v.subaccount,
+        }
+        .encode(e, ctx)
+    }
+}
```
gregorydemay: - Let's move this to a new file in `rs/ethereum/cketh/minter/src/cbor/account.rs`
- Add a roundtrip proptest in `rs/ethereum/cketh/minter/src/cbor/tests.rs` (see the tests there, follow the same style)

gregorydemay: 🤖 Both done in f79f9ceb2b.

- The encoding moved to `src/cbor/account.rs`, next to `cbor/id.rs` and in the same shape (free `encode`/`decode` plus a private `CborAccount`). `AttestationRequest` now says `with = "crate::cbor::account"`. You are right that it was in the wrong place: nothing about it is specific to attestations, it is how the minter's events carry an account.
- `cbor/tests.rs` gained an `AccountContainer` and a roundtrip proptest alongside the existing two. It draws the owner from `arb_principal` (0..=29 bytes, since the length is what the encoding has to preserve) and the subaccount from `option::of(uniform32)`, so an all-zero subaccount is covered too — `arb_ledger_subaccount` folds that case into `None`, which would have left it untested.

rs/ethereum/cketh/minter/src/attestation/mod.rs
```diff
@@ -28,24 +29,29 @@ const DOMAIN_SEPARATOR: &[u8] = b"ck-deposit-owner";
...
 pub struct AttestationRequest {
     /// Prevents replaying the attestation onto another chain.
+    #[n(0)]
     chain_id: u64,
     /// The **deposit helper** contract, the one whose deposit events name the account an address
     /// credits — not the sweeper delegate, whose own address is `address(this)` in the delegate
     /// call and is deliberately absent from the preimage.
+    #[n(1)]
     deposit_helper: Address,
+    #[cbor(n(2), with = "crate::cbor::account")]
```
mbjorkqvist: Is this necessary (e.g., to insulate the minter's event log from potential future changes to the ledger type's derive), or could we just have `#[n(2)]` and use what is already in [`icrc-ledger-types`](https://github.com/dfinity/ic/blob/cc0e780430bf578e9824da2fd205fad35c9f4f5e/packages/icrc-ledger-types/src/icrc1/account.rs#L21-L29), since it produces the same encoding as what `crate::cbor::account` reimplements?

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -274,6 +275,18 @@ impl State {
         ))
     }
 
+    /// The attestation `account` has already signed for the configuration this minter runs
+    /// against, if any: signing another would cost a threshold-ECDSA signature for the same digest.
+    pub fn attestation(&self, account: Account) -> Option<&TransactionSignature> {
+        self.automatic_deposits
+            .attestation(&self.attestation_request(account)?)
```
mbjorkqvist, line 282: `attestation_request(account)` hard-codes `DepositAddressSchema::CkErc20` as the `schema`, so only attestations for this schema can be retrieved, so an attestation recorded under `ckEth` would be unreachable and re-signed every round. If only `ckERC20` schemas should be supported, it would make sense to make that clear; otherwise, taking the schema as a parameter would be another option.

rs/ethereum/cketh/minter/src/attestation/tests.rs
```diff
@@ -112,13 +116,78 @@ fn request(chain_id: u64, deposit_helper: Address, account: Account) -> Attestat
     )
 }
 
-fn helper() -> Address {
-    HELPER.parse().unwrap()
-}
+mod recorded {
```
mbjorkqvist, line 112: Would it make sense to add a test to verify that signing an attestation with the same account, but two different schemas, keeps two attestations? Or, if `ckEth` attestations are not supported, add a test for that?

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -661,6 +661,24 @@ type Event = record {
...
+            owner : principal;
+            subaccount : opt blob;
+            // Which family of per-account deposit addresses the attesting address
+            // belongs to, and so which derivation path signed.
+            schema : variant { CkErc20; CkEth };
+            y_parity : bool;
+            // 32-byte signature components.
+            r : blob;
+            s : blob;
+        };
```
mbjorkqvist: nit: This variant is added in the middle of the list here, whereas in `EventType` in `rs/ethereum/cketh/minter/src/state/event.rs`, `AttestedDepositAddress` is added to the end.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -55,9 +57,28 @@ pub struct AutomaticDeposits {
     /// Funded `(account, token)` pairs moved out of the watchlist, awaiting sweeping,
     /// keyed by the funded [`DepositRequest`]; each holds one [`SweepEntry`].
     sweep: BTreeMap<DepositRequest, SweepEntry>,
+    /// Attestations the minter has signed, keyed by exactly what each one signed. An attestation
+    /// binds an account to one chain and one helper deployment and never expires, so a later sweep
+    /// of the same address reuses it instead of paying for another threshold-ECDSA signature; a new
+    /// helper deployment yields a different key and simply misses.
+    attestations: BTreeMap<AttestationRequest, TransactionSignature>,
```
mbjorkqvist, line 68: The growth of `attestations` is currently unbounded. Does it make sense to have a metric and alert, a way to purge signatures (e.g., for an old helper if the helper deployment changes), or something else (and/or a backlog ticket to track it), even though this probably isn't something that will become an issue for a while?

mbjorkqvist (commented):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/attestation/mod.rs
gregorydemay: 🤖 You are right that it produces the same encoding, so the wrapper was buying nothing but a reimplementation to keep in step by hand. Dropped it in 9bec20a — the field is now plain `#[n(2)]` on the ledger type's own derive.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Good catch, and the conclusion is stronger than either option: deriving further deposit addresses is not acceptable for ckETH, so a `CkEth` attestation is not something the minter may ever produce. Carrying the schema in the request therefore only created the miss you describe.

a5f075b removes the field and names `CkErc20` where the derivation path is built. That also retires the Candid schema enum, its mapping, and the orderings and cbor derives the map key had needed.

rs/ethereum/cketh/minter/src/attestation/tests.rs
gregorydemay: 🤖 Moot after a5f075b — with the schema out of the request there are no two schemas to keep apart, and the ckERC20-only scope is now stated on `sign_attestation` and `State::attestation`.

rs/ethereum/cketh/minter/cketh_minter.did
gregorydemay: 🤖 Fixed in ea41e34 — moved to the end of the variant list, where `#[n(33)]` says it belongs.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Agreed on both counts: it grows with the number of accounts ever swept, and entries naming a retired helper stay behind forever.

ebe9eff adds a `cketh_minter_stored_attestations` gauge and says outright in the field doc that nothing prunes the map, so the growth is visible before it needs bounding. Holding off on a purge mechanism — that wants the helper-redeployment story settled first, which is beyond this PR.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -269,11 +270,23 @@ impl State {
...
             account,
         ))
     }
 
+    /// The attestation `account`'s ckERC20 deposit address has already signed for the configuration
+    /// this minter runs against, if any: signing another would cost a threshold-ECDSA signature for
+    /// the same digest.
+    pub fn attestation(&self, account: Account) -> Option<&TransactionSignature> {
+        self.automatic_deposits
+            .attestation(&self.attestation_request(account)?)
```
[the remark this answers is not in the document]
gregorydemay: come upstack

PR #11322, opened 2026-08-26
feat: Refactor VSOCK

Bownairo:
Restructure the VSOCK code to make it easier to test and build off of, for future use in the type 4.X nodes with more than one GuestOS.


rs/ic_os/vsock/src/bin/guest.rs
```diff
@@ -1,109 +1,77 @@
...
+use clap::{Parser, Subcommand};
 
 #[derive(Debug, Parser)]
-#[clap(
-    version = "1.0.0",
-    about = "A CLI for sending vsock commands",
-    author = "DFINITY Stiftung (c) 2023"
-)]
+#[command(version = "1.0.0")]
+/// A CLI for sending vsock commands
```
frankdavid: nit: https://doc.rust-lang.org/style-guide/#doc-comments

> Put doc comments before attributes.

rs/ic_os/vsock/src/bin/guest.rs
```diff
@@ -1,109 +1,77 @@
...
-            message,
-            count: cli.notify.count,
-        }))
-    } else {
-        Err("no command matched".into())
+    };
+
+    let response = LinuxVsockClient::with_port(cli.port)
+        .send_command(command)
+        .context("sending command")?;
```
frankdavid: Shall we add the command to the context? I don't think it ever contains PII.

rs/ic_os/vsock/src/bin/guest.rs
```diff
@@ -1,109 +1,77 @@
...
+
+    // Output the values directly
+    match response {
+        Ok(Payload::HostOSVsockVersion(version)) => println!("{version}"),
+        Ok(Payload::HostOSVersion(version)) => println!("{version}"),
+        Err(error) => bail!("Server responded with error: '{error}'"),
```
frankdavid: When printing anyhow errors. use {:?}, otherwise only the top error is printed, never the cause (in our case "sending command", nothing else)

rs/ic_os/vsock/src/bin/host.rs
```diff
@@ -1,6 +1,48 @@
...
+            join_handle.await
+        },
+        result = &mut join_handle => result,
+    }
+    .context("cleaning up task")?;
+
+    Ok(())
+}
+
+// Adapted from rs/http_endpoints/async_utils/src/lib.rs
```
frankdavid, line 31: nit: Is it useful as a comment? If the don't have to be kept in sync, maybe we don't really need to know where it originated from (especially because it's a pretty common pattern).

rs/ic_os/vsock/src/server/misc.rs
```diff
@@ -1,26 +1,36 @@
...
+    // Echo notify messages to the local GuestOS console so they are visible
+    // in cloud environments where the host console is not accessible.
+    for path in ["/dev/tty1", "/dev/ttyS0"] {
+        if let Ok(mut tty) = OpenOptions::new().write(true).open(path) {
+            let _ = writeln!(tty, "\n{}", notify_data.message);
+        }
+    }
+
+    // Skip logging to host if manual recovery TUI is running to avoid interfering with the display
     if procfs::process::all_processes().is_ok_and(|processes| {
```
frankdavid, line 33: Uh, this is a bit ugly 🫣 (but no need to fix it now)

rs/ic_os/vsock/src/server/misc.rs
```diff
@@ -34,32 +44,24 @@ pub(crate) fn notify(notify_data: &NotifyData) -> Response {
 
     let message_output_count = std::cmp::min(notify_data.count, 10);
     let message = notify_data.message.clone();
```
frankdavid: nit: I don't think clone is necessary here, since you clone on line 52

rs/ic_os/vsock/src/server/misc.rs
```diff
@@ -34,32 +44,24 @@ pub(crate) fn notify(notify_data: &NotifyData) -> Response {
...
+    let mut handles = Vec::new();
 
     for device_path in &["/dev/tty1", "/dev/ttyS0"] {
```
frankdavid: Why not move it inside the message count loop, then you don't need to spawn tokio tasks.

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
 mod upgrade;
 
-use crate::protocol::{Command, HostOSVsockVersion, Request, Response};
+use crate::protocol::{Command, HostOSVsockVersion, MAX_MESSAGE_SIZE, Request, Response};
 use hsm::{attach_hsm, detach_hsm};
 use misc::{get_hostos_version, get_hostos_vsock_version, notify};
 use upgrade::{start_upgrade_guest_vm, upgrade_hostos};
 
-use vsock::{VMADDR_CID_ANY, VsockAddr, VsockListener, VsockStream};
+use ic_http_utils::file_downloader::FileDownloadError;
```
frankdavid, line 13: nit: any reason for splitting the imports into groups?

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
 
-fn process_connection(stream: &mut VsockStream) -> io::Result<()> {
-    let mut buffer = [0; 4096];
-    let bytes_read = stream.read(&mut buffer)?;
-    let json_request: String = match std::str::from_utf8(&buffer[..bytes_read]) {
-        Ok(json_str_request) => json_str_request.to_string(),
-        Err(error) => {
-            println!("Error converting bytes to string: {error}");
-            return Err(io::Error::new(io::ErrorKind::InvalidData, error));
+        drop(self.listener);
```
frankdavid, line 123: Is it important to control the drop order? If so, could you add a comment why it's important to drop the listener first?

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
         Command::GetHostOSVersion => get_hostos_version(),
         Command::StartUpgradeGuestVM => start_upgrade_guest_vm(),
-    };
-
-    stream.write_all(serde_json::to_string::<Response>(&response)?.as_bytes())
-}
+    }
+    .map_err(|e| {
+        let error = e.to_string();
+        println!("{error}");
```
frankdavid: Nit: this can lead to double logging. Usually it's recommended to do exactly one of logging or returning an error.

rs/ic_os/vsock/src/server/upgrade.rs
```diff
@@ -1,58 +1,66 @@
...
+    let output = std::process::Command::new("systemctl")
         .arg("restart")
         .arg(GUESTOS_UPGRADER_SERVICE)
-        .output()
-    {
-        Ok(output) if output.status.success() => return Ok(Payload::NoPayload),
-        Ok(_) => {} // systemctl failed, fallthrough to error handling below
-        Err(err) => return Err(format!("Could not start {GUESTOS_UPGRADER_SERVICE}: {err}")),
-    };
+        .output()?;
```
frankdavid: Consider mapping this to an UpgraderService error, too (and perhaps add some context string, like `could not start upgrader service`)

rs/ic_os/vsock/src/server/hsm.rs
```diff
@@ -44,45 +46,33 @@ fn hsm_helper(command: &str) -> Response {
         .arg(DOMAIN_NAME)
         .arg("--file")
         .arg(hsm_xml_file.path())
-        .output();
-
-    handle_command_output(command_output)
+        .output()?;
```
blind-oracle: Error context?

rs/ic_os/vsock/src/server/hsm.rs
```diff
@@ -44,45 +46,33 @@ fn hsm_helper(command: &str) -> Response {
...
-    let hsm_info = get_hsm_info().map_err(|err| format!("Could not get hsm info: {err}"))?;
+fn create_hsm_xml_file() -> Result<NamedTempFile, VsockServerError> {
+    let hsm_info = get_hsm_info()?;
 
     println!("HSM found: {hsm_info}");
 
     let xml: String = get_hsm_xml_string(&hsm_info);
 
-    write_to_temp_file(&xml).map_err(|err| format!("Could not write to temp file: {err}"))
+    let mut file: NamedTempFile = NamedTempFile::with_prefix("hsm")?;
```
blind-oracle: is type annotation needed?

rs/ic_os/vsock/src/server/misc.rs
```diff
@@ -34,32 +44,24 @@ pub(crate) fn notify(notify_data: &NotifyData) -> Response {
...
-            Ok(())
-        };
 
-        std::thread::spawn(write_lambda);
+            Ok::<(), VsockServerError>(())
+        }));
+    }
+
+    for handle in handles {
+        handle.await??;
```
blind-oracle: Add error context here too?

rs/ic_os/vsock/src/server/mod.rs
blind-oracle: It needs to be dropped so that the socket is closed & new tasks are not added to the tracker. Otherwise it might hang forever.

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
-    {
+async fn process_connection(mut stream: VsockStream) -> Result<(), VsockServerError> {
+    let mut buffer = Vec::new();
+    timeout(
+        CONNECTION_TIMEOUT,
+        (&mut stream)
+            .take(MAX_MESSAGE_SIZE)
+            .read_to_end(&mut buffer),
+    )
+    .await??;
```
blind-oracle: Context

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
+    let request = match serde_json::from_slice::<Request>(&buffer) {
         Ok(request) => request,
-        Err(err) => {
-            stream
-                .write_all(serde_json::to_string::<Response>(&Err(err.to_string()))?.as_bytes())?;
-            return Err(err);
+        Err(source) => {
+            let error = VsockServerError::InvalidRequest {
+                request: String::from_utf8_lossy(&buffer).into_owned(),
```
blind-oracle, line 157: Does it need into_owned?

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
+) -> Result<(), VsockServerError> {
+    Ok(timeout(
+        CONNECTION_TIMEOUT,
+        stream.write_all(
+            // Make sure we put the right type on the wire
+            &serde_json::to_vec::<Response>(&response)
+                .map_err(|source| VsockServerError::InvalidResponse { response, source })?,
+        ),
+    )
+    .await??)
```
blind-oracle: Context?

rs/ic_os/vsock/src/server/mod.rs
```diff
@@ -1,192 +1,210 @@
...
-        assert_eq!(request.command, Command::GetHostOSVersion);
-
-        // Test malformed command
-        let json_str = r#"{"sender_cid": 123, "message": "attach-hsm"#; // Missing closing brace
-        let request = parse_request(json_str);
-        assert!(request.is_err());
+// As a sanity check, we request that the sender adds its own CID to the message, and that CID must match the CID in the stream peer address.
+// NOTE: The kernel vhost driver also enforces this. Any packet with a forged source is dropped.
+fn verify_sender_cid(stream: &mut VsockStream, guest_cid: u32) -> Result<(), VsockServerError> {
+    if stream.peer_addr()?.cid() != guest_cid {
```
blind-oracle: Context

rs/ic_os/vsock/src/server/upgrade.rs
```diff
@@ -1,58 +1,66 @@
...
-        .map_err(|e| e.to_string())
-    })?;
+        )
+        .await?;
 
     println!("Download completed, starting upgrade installation...");
     let command_output = std::process::Command::new(INSTALL_UPGRADE_FILE_PATH)
         .arg(UPGRADE_FILE_PATH)
-        .output();
+        .output()?;
```
blind-oracle: Context? :)

rs/ic_os/vsock/src/client.rs
```diff
@@ -12,103 +32,56 @@ pub use linux::*;
...
-            let guest_cid = vsock::get_local_cid().map_err(|e| e.to_string())?;
-
+    impl VsockClient for LinuxVsockClient {
+        fn send_command(&self, command: Command) -> Result<Response, VsockClientError> {
+            let guest_cid = vsock::get_local_cid()?;
             let request = Request { guest_cid, command };
 
-            let mut stream = VsockStream::connect_with_cid_port(VMADDR_CID_HOST, self.port)
-                .map_err(|e| e.to_string())?;
+            let mut stream = VsockStream::connect_with_cid_port(VMADDR_CID_HOST, self.port)?;
```
blind-oracle: Context :) and below too

rs/ic_os/vsock/src/protocol.rs
```diff
@@ -2,6 +2,10 @@ use std::fmt;
 
 use serde::{Deserialize, Serialize};
 
+// 64 KiB - generous for current messages (typically <1 KiB) while
```
blind-oracle: Doc comment?

rs/ic_os/vsock/Cargo.toml
```diff
@@ -11,16 +11,25 @@ clap = { workspace = true } # only needed by bin, can't express in direct Cargo.
...
 # neither build needs libusb-1.0-0-dev (Bazel additionally disables the
 # libudev probe; see the libusb1-sys annotation in bazel/rust.MODULE.bazel).
 rusb = { version = "0.9", features = ["vendored"] }
-sha2 = { workspace = true }
 tempfile = { workspace = true }
-vsock = "0.4"
+vsock = "0.5.4"
 ic-http-utils = { path = "../../http_utils" }
-tokio = { workspace = true, features = ["rt", "rt-multi-thread"] }
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "sync"] }
```
blind-oracle: Workspace already has `features = ["full"]` so this is redundant

blind-oracle (dismissed):

blind-oracle:
Overall, please add as much context to errors as possible - would ease debuggng in the future.

rs/ic_os/vsock/src/server/mod.rs
frankdavid: Since the above loop no longer runs, no more tasks are added to the tracker. The only thing the Drop implementation does is closing the vsock socket fd and thereby rejecting connection requests right away (instead of making the kernel buffer them until the end of the method is reached when the listener is dropped anyway). I don't mind having it but if we actually use it as an optimization, it's worth mentioning it in a comment.

rs/ic_os/vsock/src/bin/host.rs
Bownairo: It feels like it might be nice to re-use the lib at some point, I just didn't want to pull in the logger. How about I turn this into a TODO with that note instead?

rs/ic_os/vsock/src/server/misc.rs
Bownairo: Agreed 🙈.

rs/ic_os/vsock/src/server/misc.rs
Bownairo: We still want to spawn so that we don't need to keep the socket open the whole time (I accidentally broke this), but flipping the loop is much nicer either way.

rs/ic_os/vsock/src/server/mod.rs
Bownairo: I mostly follow the groups [here](https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=#StdExternalCrate%5C%3A), but maybe it's just a (bad) habit at this point 😬.

rs/ic_os/vsock/src/server/mod.rs
Bownairo: I started basing this off `ic-bn-lib` so I'm sure it's a little overkill. Still, I've left it with a comment because a few of the timeouts are pretty long.

rs/ic_os/vsock/src/server/mod.rs
Bownairo: Any errors from the command execution we don't return up the stack. Here we print for the server, and write the `Response` back to the client, so both sides can log. Is that fine?

rs/ic_os/vsock/src/server/mod.rs
Bownairo: Unfortunately `String::from_utf8_lossy` mysteriously returns a `Cow`, and the direct variant that actually gives a `String` is nightly. 🐮

rs/ic_os/vsock/src/bin/guest.rs
Bownairo: It's actually annoying to wire this up. Can I argue that this level is only transport errors, so the command itself matters less, and it's in the binary, so the command that was run should be more obvious?

basvandijk (approved):

pierugo-dfinity (approved):
Approving core-protocol-owned `rs/sys/src/utility_command.rs` changes

frankdavid (approved):

PR #11323, opened 2026-08-26
feat(ci): attest CDN artifact uploads with build provenance

basvandijk:
First producer-side step for **security finding 3618194** (MEDIUM, CWE-494): the release pipeline has no integrity anchor independent of the serving CDN — `tag-release.yml` and `ledger-suite-release.yml` publish checksums computed *from their own CDN downloads*, and `mainnet_revisions.py` records CDN-served hashes into `mainnet-icos-revisions.json`.

`ci/src/artifacts/upload.sh` already hashes every file it uploads (from the **local** bytes, never from anything served back by the CDN) and prints one manifest line per file — but the `upload-artifacts` action discarded that output into the step summary. This PR switches the output to sha256sum format (`<hex>  <path>`, path relative to the CDN root) and turns it into a durable, attested anchor.

### Changes

* **`.github/actions/upload-artifacts`**: converts the upload manifest to `sha256sum` format, validates its shape (fail closed on empty/malformed lines), exposes its own sha256 as an output, and uploads it as an Actions artifact (`manifest-name` input).
* **`ci-main.yml`**: both upload jobs (`bazel-test-all`, `upload-external-artifacts`) produce named manifests; a new unprivileged `upload-manifests` job re-checks each one against the producing job's output sha256, merges them into a single `upload-manifest` artifact, and anchors it in its own job output. That the uploads happen in two jobs stays an implementation detail of `ci-main.yml`: callers see only the `release-build` and `manifest-sha256` `workflow_call` outputs. The hand-assembled `binaries/{arm64-linux,arm64-darwin,x86_64-darwin}/` directories now also get the `SHA256SUMS` file every other CDN directory already gets from the `artifact_bundle` rule. **No permission changes** (the merge job runs no build or test code, and same-run artifact transfer uses the runtime token, so it drops the workflow-level `actions`/`id-token` grants).
* **`bazel-test-arm64` → `upload-external-artifacts` binding**: the arm64/macOS binaries transit Actions artifacts between the Namespace builders and the CDN-upload job; each `bazel-test-arm64` leg now records the `sha256sum` lines of every file it uploads in its job outputs — immediately after the builds, keeping the whole-codebase build and test steps outside the trust chain — and `upload-external-artifacts` re-checks each downloaded artifact against them (rejecting mismatches, empty outputs, and files beyond the recorded set) before bundling. Closes the same artifact-swap hole for the binaries that the manifest checks close for the manifests.
* **New `attest-uploads` job** in `ci-kickoff.yml` (master builds) and `release-testing.yml` (`rc--*`/`hotfix-*` builds): re-checks the merged manifest artifact against ci-main's `manifest-sha256` output (an Actions artifact can be replaced mid-run by anything holding an `actions: write` token; job outputs cannot — the same check anchors every hop of the merge) and runs `actions/attest` (pinned, v4.2.2; the `attest-build-provenance` wrapper's README directs new implementations to it — same SLSA provenance predicate) with `subject-checksums` — one attestation covering every uploaded file **including each directory's `SHA256SUMS`**.

### Why the attest job lives in the callers, not in ci-main.yml

GitHub validates a called workflow's permission requests against the caller's grant **statically, even for jobs whose `if:` would skip them** ([community discussion #155062](https://github.com/orgs/community/discussions/155062)). An attest job inside `ci-main.yml` would therefore force `attestations: write` into `ci-kickoff-manual.yml`, which runs CI on fork PRs. With this layout no job that executes build or test code ever holds `attestations: write`, and `ci-kickoff-manual.yml` is untouched.

### How consumers will verify (follow-up PRs)

```
gh attestation verify SHA256SUMS --repo dfinity/ic \
  --signer-workflow dfinity/ic/.github/workflows/<ci-kickoff|release-testing>.yml \
  --source-digest <commit>
```
then `sha256sum --check` each downloaded artifact against the verified sums. The `--source-digest` binding prevents replaying another commit's legitimately-attested objects. Follow-ups: `ledger-suite-release.yml`, `tag-release.yml`, `mainnet_revisions.py`.

### Validation

Tested **end-to-end pre-merge** via a `workflow_dispatch` of `ci-kickoff.yml` on this branch with `release-build=true` (the `upload-artifacts` environment temporarily allowed this branch via a custom deployment-branch policy, removed after the test): [run 33004648043](https://github.com/dfinity/ic/actions/runs/33004648043) on `fa328bb033` — all jobs green, including both `bazel-test-arm64` legs (checksum recording), `upload-external-artifacts` (download verification + extras guard), `upload-manifests`, and `Attest Uploaded Artifacts` (one attestation, 180 subjects, `predicateType: https://slsa.dev/provenance/v1`, see: https://github.com/dfinity/ic/attestations/43234804).

Workstation verification against that run's artifacts:

* Positive: `curl -fsSLO https://download.dfinity.systems/ic/fa328bb033…/binaries/x86_64-linux/SHA256SUMS && gh attestation verify SHA256SUMS --repo dfinity/ic --signer-workflow dfinity/ic/.github/workflows/ci-kickoff.yml --source-digest fa328bb033…` → exit 0. The follow-up PR's `ci/scripts/fetch-attested-sums.sh` also ran end-to-end against this attestation (exit 0, fetched file byte-identical), confirming the `--format json` subject structure (`.[].verificationResult.statement.subject[]`) and the per-directory subject-name binding on a live attestation.
* Negative tests (per F-012 house style), all correctly failing: one flipped hex char in the file → exit 1; `--source-digest` of a different commit → exit 1; `--signer-workflow …/release-testing.yml` (wrong pipeline) → exit 1.
* All changed workflow/action files parse as YAML; the manifest-shape check and the sha256 cross-checks fail closed by construction.

### Scope / residual risk

The attestation certifies *what this CI built and uploaded*, not that the build is reproducible — `build-determinism` and `repro-check` remain the complementary controls. A run that uploads but dies before attesting leaves that commit unattested until re-run (the re-upload is checked byte-for-byte by `rclone --immutable --checksum`). Builds in ic-private are not attested; their commits gain attestations when the branch is pushed to dfinity/ic and rebuilt there.

### DIFF
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -6,7 +6,7 @@ description: |
 
   The attestation is the integrity anchor that lets release workflows and the
   mainnet-revisions updater verify CDN downloads against something independent
-  of the CDN (security finding 3618194): they fetch a directory's SHA256SUMS
+  of the CDN: they fetch a directory's SHA256SUMS
   from the CDN, check it with
   `gh attestation verify --repo dfinity/ic --signer-workflow <this pipeline>
   --source-digest <commit>` and then verify each artifact against the now
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -1,13 +1,12 @@
 name: Attest Uploaded Artifacts
 description: |
   Generate a build-provenance attestation covering every file that this run's
-  ci-main.yml uploaded to the CDN (see ci-main's merge-upload-manifests job,
+  ci-main.yml uploaded to the CDN (see ci-main's upload-manifests job,
   which produces the upload manifest consumed here).
 
   The attestation is the integrity anchor that lets release workflows and the
   mainnet-revisions updater verify CDN downloads against something independent
-  of the CDN: they fetch a directory's SHA256SUMS
-  from the CDN, check it with
+  of the CDN: they fetch a directory's SHA256SUMS from the CDN, check it with
   `gh attestation verify --repo dfinity/ic --signer-workflow <this pipeline>
   --source-digest <commit>` and then verify each artifact against the now
   trusted sums.
@@ -16,7 +15,7 @@ description: |
   `manifest-sha256` output: an Actions artifact can be replaced mid-run by
   anything holding an `actions: write` token, but job outputs cannot be
   tampered with by other jobs. The same check anchors the merge that produced
-  the artifact: ci-main's merge-upload-manifests job verified each upload
+  the artifact: ci-main's upload-manifests job verified each upload
   job's manifest against that job's output before merging.
 
   The calling job needs `attestations: write` and `id-token: write` and must
@@ -49,11 +48,8 @@ runs:
         # empty output also fails here: sha256sum rejects the malformed line.
         echo "$MANIFEST_SHA256  upload-manifest.sha256sums" | sha256sum --check
 
-    # No continue-on-error: an unattested release upload is exactly the gap
-    # this anchor exists to close, so a failure here must fail the run.
     # With no predicate input, actions/attest emits a SLSA build-provenance
-    # attestation — the predicate type `gh attestation verify` expects by
-    # default.
+    # attestation — the predicate type `gh attestation verify` expects by default.
     - name: Attest
       uses: actions/attest@1e69f48acb82d1966a394da916b4c1698aa569d6 # v4.2.2
       with:
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -1,8 +1,8 @@
 name: Attest Uploaded Artifacts
 description: |
   Generate a build-provenance attestation covering every file that this run's
-  ci-main.yml uploaded to the CDN (see the upload-artifacts action, which
-  produces the upload manifests consumed here).
+  ci-main.yml uploaded to the CDN (see ci-main's merge-upload-manifests job,
+  which produces the upload manifest consumed here).
 
   The attestation is the integrity anchor that lets release workflows and the
   mainnet-revisions updater verify CDN downloads against something independent
@@ -12,60 +12,46 @@ description: |
   --source-digest <commit>` and then verify each artifact against the now
   trusted sums.
 
-  The manifest artifacts are re-checked against the sha256s carried by
-  ci-main's job outputs: an Actions artifact can be replaced mid-run by
+  The manifest artifact is re-checked against the sha256 carried by ci-main's
+  `manifest-sha256` output: an Actions artifact can be replaced mid-run by
   anything holding an `actions: write` token, but job outputs cannot be
-  tampered with by other jobs.
+  tampered with by other jobs. The same check anchors the merge that produced
+  the artifact: ci-main's merge-upload-manifests job verified each upload
+  job's manifest against that job's output before merging.
 
   The calling job needs `attestations: write` and `id-token: write` and must
-  therefore never execute build or test code — it may consume only these
-  manifests and 64-hex job outputs.
+  therefore never execute build or test code — it may consume only this
+  manifest and 64-hex job outputs.
 inputs:
-  manifest-sha256-main:
+  manifest-sha256:
     required: true
-    description: sha256 of the upload-manifest-main artifact, from ci-main's outputs.
-  manifest-sha256-external:
-    required: true
-    description: sha256 of the upload-manifest-external artifact, from ci-main's outputs.
+    description: sha256 of the upload-manifest artifact, from ci-main's outputs.
 
 runs:
   using: "composite"
   steps:
-    - name: Download main upload manifest
+    - name: Download upload manifest
       uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
       with:
-        name: upload-manifest-main
-        path: ${{ runner.temp }}/upload-manifests/main
+        name: upload-manifest
+        path: ${{ runner.temp }}/upload-manifest
 
-    - name: Download external upload manifest
-      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-      with:
-        name: upload-manifest-external
-        path: ${{ runner.temp }}/upload-manifests/external
-
-    - name: Verify and merge manifests
+    - name: Verify manifest
       shell: bash
       env:
-        MANIFEST_SHA256_MAIN: ${{ inputs.manifest-sha256-main }}
-        MANIFEST_SHA256_EXTERNAL: ${{ inputs.manifest-sha256-external }}
+        MANIFEST_SHA256: ${{ inputs.manifest-sha256 }}
       run: |
         set -euo pipefail
-        cd "$RUNNER_TEMP/upload-manifests"
+        cd "$RUNNER_TEMP/upload-manifest"
 
-        # Verify the downloaded artifacts against the hashes recorded in the
-        # producing jobs' outputs before attesting anything. A missing or
+        # Verify the downloaded artifact against the hash recorded in
+        # ci-main's job outputs before attesting anything. A missing or
         # empty output also fails here: sha256sum rejects the malformed line.
-        echo "$MANIFEST_SHA256_MAIN  main/upload-manifest.sha256sums" | sha256sum --check
-        echo "$MANIFEST_SHA256_EXTERNAL  external/upload-manifest.sha256sums" | sha256sum --check
-
-        LC_ALL=C sort -k2 \
-          main/upload-manifest.sha256sums \
-          external/upload-manifest.sha256sums \
-          >merged.sha256sums
+        echo "$MANIFEST_SHA256  upload-manifest.sha256sums" | sha256sum --check
 
     # No continue-on-error: an unattested release upload is exactly the gap
     # this anchor exists to close, so a failure here must fail the run.
     - name: Attest
       uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
       with:
-        subject-checksums: ${{ runner.temp }}/upload-manifests/merged.sha256sums
+        subject-checksums: ${{ runner.temp }}/upload-manifest/upload-manifest.sha256sums
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -51,7 +51,10 @@ runs:
 
     # No continue-on-error: an unattested release upload is exactly the gap
     # this anchor exists to close, so a failure here must fail the run.
+    # With no predicate input, actions/attest emits a SLSA build-provenance
+    # attestation — the predicate type `gh attestation verify` expects by
+    # default.
     - name: Attest
-      uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
+      uses: actions/attest@1e69f48acb82d1966a394da916b4c1698aa569d6 # v4.2.2
       with:
         subject-checksums: ${{ runner.temp }}/upload-manifest/upload-manifest.sha256sums
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -1,35 +1,11 @@
 name: Attest Uploaded Artifacts
 description: |
-  Generate a build-provenance attestation covering every file that this run's
-  ci-main.yml uploaded to the CDN (see ci-main's upload-manifests job,
-  which produces the upload manifest consumed here).
+  Generate a build-provenance attestation from a `upload-manifest` artifact.
 
-  The attestation is the integrity anchor that lets release workflows and the
-  mainnet-revisions updater verify CDN downloads against something independent
-  of the CDN: they fetch a directory's SHA256SUMS from the CDN, check it with
-  `gh attestation verify --repo dfinity/ic --signer-workflow <this pipeline>
-  --source-digest <commit> --format json`, require the verified attestation to
-  record the file's digest under the exact subject name
-  ic/<commit>/<directory>/SHA256SUMS — gh authenticates the digest only, and
-  this attestation covers every uploaded directory, so without the name check
-  a compromised CDN could serve another same-commit directory's attested
-  SHA256SUMS here — and then verify each artifact against the now trusted
-  sums.
-
-  The manifest artifact is re-checked against the sha256 carried by ci-main's
-  `manifest-sha256` output: an Actions artifact can be replaced mid-run by
-  anything holding an `actions: write` token, but job outputs cannot be
-  tampered with by other jobs. The same check anchors the merge that produced
-  the artifact: ci-main's upload-manifests job verified each upload
-  job's manifest against that job's output before merging.
-
-  The calling job needs `attestations: write` and `id-token: write` and must
-  therefore never execute build or test code — it may consume only this
-  manifest and 64-hex job outputs.
 inputs:
   manifest-sha256:
     required: true
-    description: sha256 of the upload-manifest artifact, from ci-main's outputs.
+    description: sha256 of the upload-manifest artifact.
 
 runs:
   using: "composite"
@@ -47,14 +23,8 @@ runs:
       run: |
         set -euo pipefail
         cd "$RUNNER_TEMP/upload-manifest"
-
-        # Verify the downloaded artifact against the hash recorded in
-        # ci-main's job outputs before attesting anything. A missing or
-        # empty output also fails here: sha256sum rejects the malformed line.
         echo "$MANIFEST_SHA256  upload-manifest.sha256sums" | sha256sum --check
 
-    # With no predicate input, actions/attest emits a SLSA build-provenance
-    # attestation — the predicate type `gh attestation verify` expects by default.
     - name: Attest
       uses: actions/attest@1e69f48acb82d1966a394da916b4c1698aa569d6 # v4.2.2
       with:
--- a/.github/actions/attest-uploads/action.yaml
+++ b/.github/actions/attest-uploads/action.yaml
@@ -1,0 +1,71 @@
+name: Attest Uploaded Artifacts
+description: |
+  Generate a build-provenance attestation covering every file that this run's
+  ci-main.yml uploaded to the CDN (see the upload-artifacts action, which
+  produces the upload manifests consumed here).
+
+  The attestation is the integrity anchor that lets release workflows and the
+  mainnet-revisions updater verify CDN downloads against something independent
+  of the CDN (security finding 3618194): they fetch a directory's SHA256SUMS
+  from the CDN, check it with
+  `gh attestation verify --repo dfinity/ic --signer-workflow <this pipeline>
+  --source-digest <commit>` and then verify each artifact against the now
+  trusted sums.
+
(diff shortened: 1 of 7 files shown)


.github/actions/upload-artifacts/action.yaml
```diff
@@ -33,4 +48,43 @@ runs:
             echo '</details>';
           } >>"$GITHUB_STEP_SUMMARY"
 
+          # Persist the upload manifest in sha256sum format ("<hex>  <path>",
+          # path relative to https://download.dfinity.systems/). The hashes
+          # were computed by the uploader from the local files before upload,
+          # so this is a build-time record of what was published, not a
```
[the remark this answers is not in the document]
basvandijk: Good catch — fixed in 7916ae25c4. `upload.sh` indeed hashes after the rclone copies, so "before upload" was wrong. The security-relevant property is the hash *source*, not the timing: the uploader hashes its local copy of the file (which nothing mutates during the loop), never anything served back by the CDN. The comment and the PR description now say exactly that; changing the uploader to hash first would not add integrity, so left as-is.

.github/actions/attest-uploads/action.yaml
```diff
@@ -0,0 +1,71 @@
...
+  The calling job needs `attestations: write` and `id-token: write` and must
+  therefore never execute build or test code — it may consume only these
+  manifests and 64-hex job outputs.
+inputs:
+  manifest-sha256-main:
+    required: true
+    description: sha256 of the upload-manifest-main artifact, from ci-main's outputs.
+  manifest-sha256-external:
+    required: true
+    description: sha256 of the upload-manifest-external artifact, from ci-main's outputs.
```
basvandijk: The fact that `ci-main.yml` produces both a main and external manifest is an implementation detail of `ci-main.yml` which should not leak outside of it. So what about doing the merging of those two manifest in `ci-main.yml` itself such that it outputs a single manifest?

basvandijk: Agreed — done in b21acff573. ci-main.yml now has an unprivileged `merge-upload-manifests` job that re-checks each per-job manifest against the producing job's output sha256, merges them into a single `upload-manifest` artifact, and anchors that in its own job output; the workflow exposes just `release-build` + `manifest-sha256`, and this action shrank to download-one/check-one/attest. The trust chain is unchanged in kind (every artifact hop still re-verified against a tamper-proof job output), one verified hop longer. The merge deliberately lives in a dedicated job rather than in `upload-external-artifacts`: that would have coupled it to the long-pole `bazel-test-all` via `needs` and made the external-upload job trusted for the entire attested subject list instead of just its own files.

.github/workflows/ci-main.yml
```diff
@@ -470,6 +489,15 @@ jobs:
           mkdir -p "$bundledir/binaries/x86_64-darwin/"
           cp ~/.cache/pocket-ic-server-x86_64-darwin/pocket-ic-server-x86_64-darwin.gz "$bundledir/binaries/x86_64-darwin/pocket-ic.gz"
 
+          # Give these hand-assembled directories the same SHA256SUMS file
+          # that the bazel-built bundles get from the artifact_bundle rule
+          # (publish/defs.bzl), so every binaries/<platform>/ directory on the
+          # CDN carries a checksum manifest that release workflows can verify
+          # against its attestation.
+          for d in arm64-linux arm64-darwin x86_64-darwin; do
+            (cd "$bundledir/binaries/$d" && sha256sum *.gz | LC_ALL=C sort -k2 >SHA256SUMS)
```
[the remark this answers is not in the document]
basvandijk: Valid finding — this was listed as a residual risk in the plan, and it's now closed in 9b83a3cefe using the same job-output anchoring as the manifests: each `bazel-test-arm64` leg records the `sha256sum` lines of every file it uploads in leg-specific job outputs (recorded immediately after the builds, so the whole-codebase build and test steps stay outside the trust chain; `shasum -a 256` fallback for the macOS runner, mirroring `.github/actions/bazel`), and `upload-external-artifacts` re-checks every download against them before bundling — failing closed on mismatch or empty output, and also rejecting artifacts that carry files beyond the recorded set, so a future glob over the download dirs can't pick up unverified extras. We chose the binding over removing `actions: write`: artifact upload/overwrite rides the job's runtime token rather than `GITHUB_TOKEN`, so stripping the permission would not reliably stop same-run artifact replacement, while job outputs are tamper-proof regardless.

.github/workflows/ci-main.yml
```diff
@@ -451,6 +527,32 @@ jobs:
...
+          # an Actions artifact can be replaced mid-run by anything holding
+          # an `actions: write` token, but job outputs cannot be tampered
+          # with by other jobs. A missing or empty output also fails here:
+          # sha256sum rejects input without valid checksum lines.
+          check() { # <download-dir> <recorded sha256sum lines>
+            echo "$2" | (cd "$1" && sha256sum --check --strict)
+            # Also reject artifacts carrying files beyond the recorded ones,
+            # so a future edit that globs from these directories cannot pick
+            # up unrecorded extras that the check above never covered.
+            diff <(echo "$2" | awk '{print $2}' | LC_ALL=C sort) <(ls "$1" | LC_ALL=C sort)
```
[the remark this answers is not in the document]
basvandijk: Confirmed and fixed in 20350c6b65 — the old guard indeed accepted a hidden extra (reproduced locally). The directory listing now comes from `find . -mindepth 1`, which counts hidden entries, directories, and nested content alike, so any unrecorded object fails the diff. Re-tested all four cases: exact set passes; a hidden file, a hidden directory with nested content, and a plain subdirectory are each rejected. (Filenames with newlines or other hostile characters also can only break the comparison toward a mismatch, i.e. toward failure.)

.github/workflows/ci-kickoff.yml
basvandijk, line 102: This job should additionally be conditional on `attest-uploads` not having failed.

basvandijk: No it shouldn't since `attest-uploads` doesn't run on PRs.

.github/actions/attest-uploads/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+  Generate a build-provenance attestation covering every file that this run's
+  ci-main.yml uploaded to the CDN (see ci-main's upload-manifests job,
+  which produces the upload manifest consumed here).
+
+  The attestation is the integrity anchor that lets release workflows and the
+  mainnet-revisions updater verify CDN downloads against something independent
+  of the CDN: they fetch a directory's SHA256SUMS from the CDN, check it with
+  `gh attestation verify --repo dfinity/ic --signer-workflow <this pipeline>
+  --source-digest <commit>` and then verify each artifact against the now
+  trusted sums.
```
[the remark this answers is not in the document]
basvandijk: Fixed in 9fd15da6f1: the description now spells out the full consumer procedure — verify with `--format json`, then require the verified attestation to record the file's digest under the exact subject name `ic/<commit>/<directory>/SHA256SUMS` before trusting the sums — and explains why (gh authenticates the digest only, and this attestation covers every uploaded directory). That exact-subject check is what the shared verifier `ci/scripts/fetch-attested-sums.sh` (added in #11324, used by all consumers) implements: its jq filter requires the same attestation entry to carry both the expected subject name and the file's digest, with cross-directory substitution documented there as a negative test.

nmattia (approved):

.github/workflows/ci-main.yml
```diff
@@ -367,6 +382,54 @@ jobs:
...
+        if: ${{ needs.config.outputs.release-build == 'true' }}
+        shell: bash
+        env:
+          OS: ${{ matrix.os }}
+        run: |
+          set -euo pipefail
+          cd build
+
+          # macOS runners may only have perl's shasum, not GNU coreutils
+          # (same fallback as in .github/actions/bazel).
```
nmattia, line 396: I think we can just use `sha256sum` generally

nmattia (approved):

.github/workflows/ci-main.yml
basvandijk: `sha256sum` is not present on `namespace-profile-darwin`.

.github/actions/upload-artifacts/action.yaml
```diff
@@ -33,4 +39,29 @@ runs:
...
 
-          rm "$upload_output"
+    - name: Check manifest
+      id: manifest
+      shell: bash
+      run: |
+        set -euo pipefail
+        manifest="$RUNNER_TEMP/upload-manifest.sha256sums"
+
+        # Fail closed rather than attest garbage: the manifest must be
```
nmattia, line 49: What do you mean by "Failed closed"?

.github/workflows/ci-main.yml
nmattia: sorry I meant `shasum -a 256` of course, ie use the fallback everywhere (unless there's a good reason for not always using `shasum`)

ci/src/artifacts/upload.sh
```diff
@@ -102,6 +102,8 @@ for fullrelpath in $(find -L "$BUNDLE" -type f); do
 
     upload "$fullrelpath" "$bucket_path"
 
+    # One sha256sum-format line ("<hex>  <bucket path>") per uploaded file on stdout.
+    # The upload-artifacts action persists this output as the upload manifest that gets attested.
```
nmattia, line 106: Let's move this line out of here: the dependency goes the other way around, the action depends on this script and this script shouldn't have to know anything about or refer to the action. It's fine to say that "other tools" expect a specific format though.

.github/workflows/ci-main.yml
```diff
@@ -367,6 +382,54 @@ jobs:
             ./bazel-bin/publish/binaries/pocket-ic.gz \
             ./build/pocket-ic-server-x86_64-darwin.gz
 
+      - name: Record artifact checksums
```
nmattia, line 385: why do we record the checksums here just to verify them again on another machine? do we gain anything by doing this instead of  simply computing the checksums once when we gather all the artifacts in `upload-external-artifacts`?

.github/actions/upload-artifacts/action.yaml
```diff
@@ -6,6 +6,12 @@ inputs:
     required: true
   name:
     required: true
+  manifest-name:
+    required: true
```
nmattia, line 10: add a description (name of the artifact containing the manifest)

.github/actions/attest-uploads/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+    required: true
+    description: sha256 of the upload-manifest artifact.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Download upload manifest
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+      with:
+        name: upload-manifest
```
nmattia, line 16: guessing this has to match the value set in `Upload merged manifest`, correct? Also it looks like this assumes there was only one artifact with name `upload-manifest` that was uploaded, meaning this action here only has a single use site, correct? If that's the case, can we inline the content where it's used?

.github/workflows/ci-main.yml
```diff
@@ -470,6 +561,15 @@ jobs:
           mkdir -p "$bundledir/binaries/x86_64-darwin/"
           cp ~/.cache/pocket-ic-server-x86_64-darwin/pocket-ic-server-x86_64-darwin.gz "$bundledir/binaries/x86_64-darwin/pocket-ic.gz"
 
+          # Give these hand-assembled directories the same SHA256SUMS file
+          # that the bazel-built bundles get from the artifact_bundle rule
+          # so every binaries/<platform>/ directory on the CDN carries
+          # a checksum manifest that release workflows can verify
+          # against its attestation.
+          for d in arm64-linux arm64-darwin x86_64-darwin; do
```
nmattia, line 569: let's be consistent: everywhere else we avoid loops and explicitly handle each platform

.github/workflows/ci-main.yml
```diff
@@ -480,14 +580,73 @@ jobs:
...
+          name: upload-manifest-main
+          path: ${{ runner.temp }}/upload-manifests/main
+
+      - name: Download external upload manifest
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: upload-manifest-external
+          path: ${{ runner.temp }}/upload-manifests/external
+
+      - name: Verify and merge manifests
```
nmattia, line 617: what do we gain by having a separate step that merges the manifests? why not keep the "main" and "external" manifests separate and attest them separately? That's a lot of artifact upload & download

.github/workflows/ci-main.yml
```diff
@@ -480,14 +580,73 @@ jobs:
...
+          cd "$RUNNER_TEMP/upload-manifests"
+
+          # Verify the downloaded artifacts against the hashes recorded in
+          # the producing jobs' outputs before merging anything: an Actions
+          # artifact can be replaced mid-run by anything holding an
+          # `actions: write` token, but job outputs cannot be tampered with
+          # by other jobs. A missing or empty output also fails here:
+          # sha256sum rejects the malformed line.
+          echo "$MANIFEST_SHA256_MAIN  main/upload-manifest.sha256sums" | sha256sum --check
+          echo "$MANIFEST_SHA256_EXTERNAL  external/upload-manifest.sha256sums" | sha256sum --check
```
nmattia, line 634: I think it would be both simpler and more robust to check the resulting manifest

.github/actions/attest-uploads/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+        path: ${{ runner.temp }}/upload-manifest
+
+    - name: Verify manifest
+      shell: bash
+      env:
+        MANIFEST_SHA256: ${{ inputs.manifest-sha256 }}
+      run: |
+        set -euo pipefail
+        cd "$RUNNER_TEMP/upload-manifest"
+        echo "$MANIFEST_SHA256  upload-manifest.sha256sums" | sha256sum --check
```
nmattia, line 26: what sort of attack/issue are we preventing by verifying this checksum?

.github/workflows/ci-main.yml
basvandijk: Ah right. Oh yes, let me push a PR ...

.github/actions/attest-uploads/action.yaml
basvandijk: > guessing this has to match the value set in Upload merged manifest, correct? 

Correct.

> Also it looks like this assumes there was only one artifact with name upload-manifest that was uploaded, meaning this action here only has a single use site, correct?

Correct.

> If that's the case, can we inline the content where it's used?

That was the first design I had but that ran into issues with the fact that the inlined code then needs to have the `attestations: write` permission which is not the case on PRs and for forked PRs.

Another way of saying this: I only want ci-kickoff and release-testing to get the `attestations: write` permissions and only for release builds to ensure (forked) PRs never gain this permission.

.github/actions/attest-uploads/action.yaml
basvandijk: I guess I removed too much AI comments because this was [explained before](https://github.com/dfinity/ic/pull/11323/changes/eb091adb59edc40fc277764d418ff809efc9edf5). If we blindly trust the downloaded `upload-manifest` artifact we run the risk of a job accidentally or maliciously overriding the artifact. It's impossible to override job outputs hence why we check it against the `inputs.manifest-sha256` which will be the `ci-main.outputs.manifest-sha256` output.

.github/actions/upload-artifacts/action.yaml
basvandijk: I removed the description when removing the slop comments. I will add something back.

.github/actions/upload-artifacts/action.yaml
basvandijk: Ha, I actually had a bit of back and forth with Claude on this before because I had the same initial question of why not just say "Fail rather than attest garbage". Claude argued:

> "Fail closed" is a security idiom: when a check can't be satisfied, take the restrictive action (abort) rather than the permissive one (carry on). The opposite is "fail open" — degrade to letting things through.
> 
> "Fail" describes what this step does; "fail closed" describes the policy it's an instance of, and that's information the reader can't recover from the code.
> 
> The extra word does three things:
>
> * It names a pattern with a well-known opposite, so it implicitly says "fail open was considered and rejected here." Bare "fail" reads as a description; "fail closed" reads as a decision.
> * It's a directive to the next editor. Someone hitting a flaky manifest in CI is tempted to soften this into a warning-and-continue, or add || true. "Fail closed" says that's off the table by design, not just untried.
> * It scopes beyond the two checks present. If a third condition gets added later, the comment already says which way it should resolve.
>
> That said, your instinct isn't wrong that it's doing less work than usual here, because rather than attest garbage already supplies the contrast — the sentence names both branches on its own. Normally "fail closed" earns its keep by implying the alternative; in a sentence that states the alternative outright, it's mostly reinforcement. So it's mild redundancy rather than a wasted word.
>
> One genuine wrinkle: the term comes from access control and circuit design, where it means the mechanism denies when the mechanism itself can't function (auth server unreachable → deny). The malformed-line check isn't really that — it's ordinary input validation, and the check is working exactly as intended. The empty/missing-manifest check is closer to the real thing: the manifest being absent likely means the upload step didn't run at all, and we refuse rather than assume the best. So the term fits about half of what it's labelling.
>
> Happy to trim it to Fail rather than attest garbage if you'd rather the comment not lean on jargon — it loses the "don't soften this later" signal but says the same thing about today's code.

But since both you and me tripped over this I'll remove the "closed". There are also other instances in the follow ups IIRC.

.github/workflows/ci-main.yml
basvandijk: Yes, we should compute the checksums of the artifacts in the same job that produced the artifacts. If we compute the checksums in `upload-external-artifacts` we run the risk of some job either accidentally or maliciously overriding the GitHub artifacts that will be uploaded by this job below.

.github/workflows/ci-main.yml
basvandijk: Make sense. Will do.

.github/workflows/ci-main.yml
basvandijk: I wanted to have a single attestation per release to keep things simple. But this comes at the cost of a merge step somewhere. In my first design this merge step was in the `.github/actions/attest-uploads/action.yaml` which I found a leaky abstraction so I moved it here.

.github/workflows/ci-main.yml
basvandijk: Check the resulting manifest against what?

.github/actions/upload-artifacts/action.yaml
basvandijk: https://github.com/dfinity/ic/pull/11404

.github/actions/upload-artifacts/action.yaml
basvandijk: https://github.com/dfinity/ic/pull/11404

.github/workflows/ci-main.yml
basvandijk: https://github.com/dfinity/ic/pull/11404

.github/workflows/ci-main.yml
basvandijk: https://github.com/dfinity/ic/pull/11404

ci/src/artifacts/upload.sh
basvandijk: https://github.com/dfinity/ic/pull/11404

.github/workflows/ci-main.yml
nmattia: Excellent question, looks like something ate the rest of the comment.

PR #11325, opened 2026-08-26
fix(ci): verify tag-release artifacts against build attestations

basvandijk:
Third step for **security finding 3618194** (MEDIUM, CWE-494), building on the producer-side attestations of #11323 (moved into `ci-main.yml` by #11569) and the `fetch-verified-artifacts` action of #11324. **Based on #11324's branch** (it reuses the action); GitHub will retarget this PR to master when #11324 merges.

`tag-release.yml` downloaded 9 tool binaries, 2 OS images and 19 canister wasm/did pairs from the CDN with plain curl and published a `sha256sums.txt` computed **from the downloads themselves** — pure self-attestation, so anyone with CDN/bucket write could get their artifacts republished as an official GitHub release with matching checksums.

### Changes

* Every download now goes through `.github/actions/fetch-verified-artifacts`: the per-directory `SHA256SUMS` is verified with `gh attestation verify --repo dfinity/ic --signer-workflow dfinity/ic/.github/workflows/ci-main.yml --source-digest <commit>`, and each artifact is checked against the verified sums — **any CDN substitution fails the release**. Pinning `release-testing.yml` as the run's **top-level workflow** — the certificate's Build Config URI (OID 1.3.6.1.4.1.57264.1.18), checked in `fetch-attested-sums.sh`'s `jq` filter — additionally enforces that a `release-*` tag can only ship commits that went through the release-qualification pipeline.

  Since #11569 the attestation is minted inside the reusable `ci-main.yml`, so `ci-main.yml` is the signer for every pipeline and the pipeline distinction lives in the Build Config URI rather than in the certificate SAN. ⚠️ Attestations minted between #11323 and #11569 carry the old caller-signed identity and are rejected by design; `release-*` tags on commits from that window cannot be shipped through this workflow.
* `sha256sums.txt` now contains the verified **build-time** hashes (plus the hash of the locally assembled `canisters.tar`), covering exactly what ships.

### Bugs fixed in passing

* **OS images were silently never attached**: since `9bd0a407b7` (2024-08) the download step saved `*-img.tar.zst` while the `files:` list said `*-img.tar.gz` and softprops/action-gh-release v0.1.15 only warned. The image downloads are  now **dropped entirely** — this preserves the two-year de-facto release contents; the canonical image channel is the CDN + NNS election proposal hash, verified by `ci/scripts/repro-check`. `fail_on_unmatched_files: true` makes any repeat of this class of bug fail loudly.
* On the `workflow_dispatch` path the release was created against the dispatched ref, not the requested tag — `tag_name` is now passed explicitly.
* `softprops/action-gh-release` upgraded from the ancient v0.1.15 pin to the same v2.4.2 pin used by `ledger-suite-release.yml`.


.github/workflows/tag-release.yml
```diff
@@ -1,6 +1,9 @@
 name: Publish Release
 
 permissions:
+  # `gh attestation verify` reads the attestations API with this workflow's
+  # GITHUB_TOKEN; an explicit permissions map zeroes everything unlisted.
```
basvandijk: Let's just drop this comment.
```suggestion
```

.github/workflows/tag-release.yml
```diff
@@ -31,31 +31,46 @@ jobs:
         with:
           ref: ${{ github.event.inputs.tag-name || github.ref }}
 
-      - name: Collect Release Artifacts
+      - name: Determine release commit
```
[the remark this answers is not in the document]
basvandijk: Fixed in bb715a0e6e: the dispatched input is now qualified as `refs/tags/<tag-name>`, so a nonexistent tag fails the checkout (git fetch of a fully-qualified missing ref exits non-zero, before any verification) and a branch sharing the tag's name can no longer shadow it — checkout resolves the tag, matching the `tag_name` the release attaches to. Also added an explicit guard rejecting an empty `tag-name` on dispatch, since `required: true` is only enforced by the web UI and an API-triggered dispatch could omit it and fall through to the dispatched branch.

.github/workflows/tag-release.yml
basvandijk: Dropped in 34e7461fcc.

mbjorkqvist (approved):

rzakrzyk (approved):

PR #11330, opened 2026-08-26
feat(cketh): record the deposits a sweep moves and size its gas to the delegate's work

gregorydemay:
## Why

- A sweep request said what call data it carried but not which deposits it took, so the sweep queue could only be reconstructed by decoding the call data.
- Its gas limit was a flat base plus a per-deposit cost, which cannot bound what the delegate actually executes.

## What

- The request records the `(account, token)` deposits it moves, so the sweep queue and the set of delegated addresses are both reconstructible from the event log.
- The gas limit follows the delegate's work: `sweepErc20Batch` hands the whole token array to every address it sweeps, so balance checks and transfers both grow as distinct addresses times distinct tokens.
- It budgets a transfer for every such pair, not only the pairs the request names: `sweepErc20` moves any balance it finds, and a deposit address accumulates residue.
- Only the addresses a sweep delegates pay an authorization cost on top.

## Tests

- A table over batch shapes pins each term of the limit.
- One test asserts a batch of more pairs from fewer deposits costs more, so a limit that followed the deposit count would fail it.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
@@ -244,6 +244,11 @@ const SWEEP_GAS_PER_TRANSFER: GasAmount = GasAmount::new(110_000);
 
 /// Gas one EIP-7702 authorization costs: 12'500 (`PER_AUTH_BASE_COST`) plus the 25'000
 /// (`PER_EMPTY_ACCOUNT_COST`) a deposit EOA pays, its account holding no ETH and no code yet.
+///
+/// Budgeted for every address the sweep touches, since every one of them carries a tuple. A tuple
+/// the EVM skips — the address is already delegated, so the nonce it was signed for no longer
+/// matches — still pays `PER_AUTH_BASE_COST`, and pays no `PER_EMPTY_ACCOUNT_COST` because such an
+/// account exists, so this is the worst case either way.
 const SWEEP_GAS_PER_AUTHORIZATION: GasAmount = GasAmount::new(40_000);
 
 /// A sweep the minter issues **from its dedicated sweeper address**, on the sweeper's own nonce
@@ -275,16 +280,17 @@ pub struct SweepRequest {
     /// The IC time at which the sweep was decided.
     #[n(5)]
     pub created_at: u64,
-    /// Delegations to install on the way, one signed by each deposit address the sweep touches
-    /// that is not yet delegated to the sweeper contract. Empty once they all are, and a sweep
-    /// with none is a plain EIP-1559 transaction.
+    /// Delegations to install on the way, one signed by every deposit address the sweep touches,
+    /// whether or not that address is already delegated: the tuples are signed for nonce 0, which
+    /// installs a delegation that is missing and is skipped otherwise. A sweep that delegates
+    /// nothing at all carries none, and is then a plain EIP-1559 transaction.
     #[n(6)]
     pub authorizations: Vec<SignedAuthorization>,
     /// The deposits this sweep moves, one per `(account, token)` pair, in the order the sweep queue
     /// offered them. Not the shape of `data`, which names one item per deposit address: an account
     /// with two tokens queued is two deposits here and one item there. Recording them makes the
-    /// request self-describing — the sweep queue and the set of delegated addresses are both
-    /// reconstructible from the log, without decoding call data.
+    /// request self-describing — the sweep queue is reconstructible from the log, without decoding
+    /// call data.
     #[n(7)]
     pub deposits: Vec<SweptDeposit>,
 }
@@ -298,11 +304,6 @@ pub struct SweptDeposit {
     pub erc20_contract_address: Address,
     #[n(2)]
     pub address: DepositAddress,
-    /// Whether this sweep installs the delegation of `address`, i.e. whether that address'
-    /// authorization is in [`SweepRequest::authorizations`]. All of an account's deposits sit at the
-    /// one address, so they share one authorization and agree on this flag.
-    #[n(3)]
-    pub delegating: bool,
 }
 
 impl SweepRequest {
@@ -310,21 +311,19 @@ impl SweepRequest {
     /// the number of deposits: the delegate hands the whole token array to every address it sweeps,
     /// so both the balance checks and the transfers grow as distinct addresses × distinct tokens. It
     /// checks every pair, and moves any pair it finds a balance at — not only the pairs `deposits`
-    /// names. Only the addresses this sweep delegates pay an authorization on top.
+    /// names. Every address pays an authorization on top, since the sweep carries a tuple for each
+    /// of them.
     ///
     /// The enqueuing side sends one token per sweep, so the product is the addresses in practice;
     /// the general form is what keeps the limit safe for any request that reaches here.
     pub fn gas_limit(&self) -> GasAmount {
-        let pairs = self
-            .distinct_count(|deposit| deposit.address)
-            .saturating_mul(self.distinct_count(|deposit| deposit.erc20_contract_address));
+        let addresses = self.distinct_count(|deposit| deposit.address);
+        let pairs =
+            addresses.saturating_mul(self.distinct_count(|deposit| deposit.erc20_contract_address));
         [
             (SWEEP_GAS_PER_BALANCE_CHECK, pairs),
             (SWEEP_GAS_PER_TRANSFER, pairs),
-            (
-                SWEEP_GAS_PER_AUTHORIZATION,
-                as_u64(self.authorizations.len()),
-            ),
+            (SWEEP_GAS_PER_AUTHORIZATION, addresses),
         ]
         .into_iter()
         .fold(SWEEP_BASE_GAS, |total, (gas_per_unit, units)| {
--- a/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/transactions/mod.rs
@@ -5,6 +5,7 @@ pub(in crate::state) mod tests;
 
 pub use request::PipelineRequest;
 
+use crate::deposit_address::DepositAddress;
 use crate::endpoints::{EthTransaction, RetrieveEthStatus, TxFinalizedStatus, WithdrawalStatus};
 use crate::eth_logs::LedgerSubaccount;
 use crate::eth_rpc::Hash;
@@ -13,7 +14,7 @@ use crate::eth_rpc_client::responses::TransactionStatus;
 use crate::logs::INFO;
 use crate::map::MultiKeyMap;
 use crate::numeric::{
-    CkTokenAmount, Erc20Value, LedgerBurnIndex, LedgerMintIndex, TransactionCount,
+    CkTokenAmount, Erc20Value, GasAmount, LedgerBurnIndex, LedgerMintIndex, TransactionCount,
     TransactionNonce, Wei,
 };
 use crate::tx::{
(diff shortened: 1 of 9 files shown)


rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -241,11 +280,68 @@ pub struct SweepRequest {
...
+    /// checks every pair, and moves any pair it finds a balance at — not only the pairs `deposits`
+    /// names. Every address pays an authorization on top, since the sweep carries a tuple for each
+    /// of them.
+    ///
+    /// The enqueuing side sends one token per sweep, so the product is the addresses in practice;
+    /// the general form is what keeps the limit safe for any request that reaches here.
+    pub fn gas_limit(&self) -> GasAmount {
+        let addresses = self.distinct_count(|deposit| deposit.address);
+        let pairs =
+            addresses.saturating_mul(self.distinct_count(|deposit| deposit.erc20_contract_address));
```
gregorydemay, line 324: I don't get this part. A given address sweeps a certain number of tokens and this is not necessarily the same as for another address. Example

1. Adress1 USDC
2. Address2 USDC
3. Address2 USDT

Then we shouldn't count Address 1 USDT

gregorydemay: 🤖 Address1 USDT *is* counted on purpose, because the delegate really does check it. The call data carries **one** token array — the union of the tokens over all the deposits — and `sweepErc20Batch` hands that same array to every item:

```solidity
for (uint256 i = 0; i < items.length; ++i) {
    CkSweeperAttested(items[i].deposit).sweepErc20(tokens, ...);
}
```

So your example encodes as `items = [Address1, Address2]`, `tokens = [USDC, USDT]`, and on chain that is four `balanceOf` calls, not three: Address1 is walked over the full list, USDT included.

For the balance checks that makes `addresses × tokens` exactly the count, not an over-estimate. For the transfers it is a deliberate one: `sweepErc20` moves *any* non-zero balance it finds, so Address1/USDT can also transfer if that address holds USDT the queue never named — residue from a pair armed but not yet scanned, a watchlist window that closed before the funds landed, or a token the sender was never asked for. Budgeting three transfers and hitting four is a revert that burns the gas and consumes a sweep attempt; budgeting four and using three refunds the difference.

The cost of the slack is the prepaid `gas_limit × max_fee_per_gas`, which is why `enqueue_token_sweep` refuses a sweep whose prepayment exceeds the low-water mark rather than sending one it cannot pay for.

In practice the product never bites: `sweep_batches_by_token` gives each sweep a single token precisely because a mixed batch pays for pairs holding nothing, so `pairs == addresses` for every sweep the minter builds. The general form is what keeps the limit safe for a request that reaches `gas_limit()` some other way.

I have expanded the doc comment with your example, since the mechanism — the union array applied to every item — was the part it left implicit (9b3a27e56f).

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -3349,6 +3356,118 @@ mod sweep_lane {
         );
     }
 
+    #[test]
+    fn should_scale_the_gas_limit_with_the_work_the_delegate_does() {
+        struct Case {
+            scenario: &'static str,
+            deposits: Vec<(u8, u8)>,
```
gregorydemay: It's very unclear that the first u8 is meant as an address, while the second one is ERC-20 address. Better to have something like `Vec<DepositAddress, Address>` and some helper methods to easily instantiate that

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
```diff
@@ -3349,6 +3356,118 @@ mod sweep_lane {
...
+                "{}",
+                case.scenario
+            );
+        }
+    }
+
+    #[test]
+    fn should_budget_one_authorization_per_address() {
+        // Two batches of two pairs each, differing only in how the pairs are spread: one address
+        // holding two tokens carries a single tuple, two addresses holding one token each carry two.
```
gregorydemay: too many comments -> code smell. optimize the tests for the read, intent of test should be clear from the code, otherwise delete.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
gregorydemay: 🤖 Agreed — the pair now carries the real types. `sweep_of(&[(DepositAddress, Address)])` replaces `with_deposits(&[(u8, u8)])`, built from two seed helpers, so the cases read off their own addresses and tokens:

```rust
let (alice, bob) = (deposit_address(1), deposit_address(2));
let (usdc, usdt) = (erc20(1), erc20(2));

for (deposits, pairs, addresses) in [
    (vec![], 0, 0),
    (vec![(alice, usdc)], 1, 1),
    (vec![(alice, usdc), (alice, usdt)], 2, 1),
    (vec![(alice, usdc), (bob, usdc)], 2, 2),
    (vec![(alice, usdc), (bob, usdt)], 4, 2),
] {
```

That also let the `scenario` field go: the deposits are their own description, so the assertion message is just `"{deposits:?}"`. And the expectation is now `BASE + pairs * PER_PAIR + addresses * PER_ADDRESS`, which names what each factor is instead of leaving four bare literals to be matched up by position. 098ae8cfee.

rs/ethereum/cketh/minter/src/state/transactions/tests.rs
gregorydemay: 🤖 Every comment in that block is gone, and two of the four tests went with them — they were arithmetically implied by the table, which is why they needed a comment to justify their existence at all:

- `should_budget_one_authorization_per_address` — the table pins the exact total for both `[(alice, usdc), (alice, usdt)]` and `[(alice, usdc), (bob, usdc)]`, so their difference being one authorization is already determined.
- `should_charge_a_mixed_token_batch_more_than_a_single_token_one` — same, from the `[(alice, usdc), (bob, usdt)]` and `[(alice, usdc), (bob, usdc)]` rows: 4 pairs against 2.

What survives is the table plus `should_budget_more_for_more_pairs_than_for_more_deposits`, which asserts the one thing the table cannot: gas is not monotone in the deposit count. Three addresses in three tokens are nine pairs from three deposits; six addresses in one token are six pairs from six. Named locals carry that, so the assertion message says it in one line instead of a four-line comment.

Net −87/+48 on the block. 098ae8cfee.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -241,11 +280,70 @@ pub struct SweepRequest {
...
+    /// tokens over all of them, and `sweepErc20Batch` hands that one array to every address it
+    /// sweeps. A sweep of one address in USDC and a second in USDC and USDT therefore checks the
+    /// first address' USDT as well, and moves it if it finds a balance there — a deposit address
+    /// accumulates residue the queue never named. Both the balance checks and the transfers grow as
+    /// distinct addresses × distinct tokens, and every address pays an authorization on top, since
+    /// the sweep carries a tuple for each of them.
+    ///
+    /// The enqueuing side sends one token per sweep, so the product is the addresses in practice;
+    /// the general form is what keeps the limit safe for any request that reaches here.
+    pub fn gas_limit(&self) -> GasAmount {
```
mbjorkqvist, line 321: This `gas_limit()` is now used in tests, but in production, the `SWEEP_TRANSACTION_GAS_LIMIT` is [still used](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/minter/src/sweep.rs#L151), is that on purpose?

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -241,11 +280,70 @@ pub struct SweepRequest {
...
+        let addresses = self.distinct_count(|deposit| deposit.address);
+        let pairs =
+            addresses.saturating_mul(self.distinct_count(|deposit| deposit.erc20_contract_address));
+        [
+            (SWEEP_GAS_PER_BALANCE_CHECK, pairs),
+            (SWEEP_GAS_PER_TRANSFER, pairs),
+            (SWEEP_GAS_PER_AUTHORIZATION, addresses),
```
mbjorkqvist, line 328: Should this be:
```suggestion
            (SWEEP_GAS_PER_AUTHORIZATION, as_u64(self.authorizations.len())),
```
?

Otherwise, e.g., [`delegating_sweep_request()`](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/minter/src/state/transactions/tests.rs#L3007-L3013) has 2 tuples but `deposits: vec![]`, so `gas_limit() == 60_000`, while the intrinsic cost alone is `21_000 + 2 × 25_000 = 71_000`.

Alternatively, if they should always be the same length (although the [`deposit_from_cex.md`](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/docs/deposit_from_cex.md?plain=1#L878-L883) design doc says:

> mixed batches are fine (tuples only for not-yet-delegated addresses, already-delegated ones ride along without tuples)

), add enforcement of this.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -212,6 +213,44 @@ impl SweepId {
...
+///
+/// Budgeted for every pair the batch touches rather than only for the deposits the queue named:
+/// `sweepErc20` moves whatever balance it finds, and a deposit address accumulates residue — a pair
+/// armed but not yet scanned, a pair whose watchlist window closed before the funds arrived, a token
+/// the sender was never asked for. A pair therefore costs a balance check and, on top of it,
+/// possibly a transfer.
+const SWEEP_GAS_PER_TRANSFER: GasAmount = GasAmount::new(110_000);
+
+/// Gas one EIP-7702 authorization costs: 12'500 (`PER_AUTH_BASE_COST`) plus the 25'000
+/// (`PER_EMPTY_ACCOUNT_COST`) a deposit EOA pays, its account holding no ETH and no code yet.
```
mbjorkqvist, line 246: This doesn't seem to match what [the design doc](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/docs/deposit_from_cex.md?plain=1#L822-L824) says:

> *Upfront gas*, charged to `M`: 21'000 base + calldata + 25'000 per tuple (`PER_EMPTY_ACCOUNT_COST`; the 12'500 refund does not apply since `D`, holding only token balances, is not in the state trie).

The value of the constant itself seems fine.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -212,6 +213,44 @@ impl SweepId {
...
 }
 
+/// Gas a sweep transaction needs before any of the addresses it sweeps: the 21'000 intrinsic cost,
+/// the batch call's own dispatch and the two calldata arrays it copies into memory.
+///
+/// This and the constants below are derived from the EVM's own costs rather than from the ~42'000
+/// per deposit measured for a batch of 20, which does not record how many distinct tokens that
+/// batch covered and so cannot separate the work that grows with the pairs from the work that grows
+/// with the transfers. Each is deliberately generous: unspent gas is refunded, whereas an
+/// underestimate wastes the whole transaction.
```
mbjorkqvist, line 223: `create_transaction` derives `max_fee_per_gas = max_transaction_fee / gas_limit`, so overshooting here doesn't reserve more - it bids proportionally less, and creation is refused once `base_fee + priority` exceeds the quotient (resubmission can't recover it either: the whole allowance is already committed). Is the plan for the enqueuing side to size `max_transaction_fee` as `gas_limit() x ceiling`? If so, worth a line here - as written, "unspent gas is refunded" reads as though the slack were free, when it's what sets the base-fee headroom.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -241,11 +280,70 @@ pub struct SweepRequest {
...
-    /// that is not yet delegated to the sweeper contract. Empty once they all are, and a sweep
-    /// with none is a plain EIP-1559 transaction.
+    /// Delegations to install on the way, one signed by every deposit address the sweep touches,
+    /// whether or not that address is already delegated: the tuples are signed for nonce 0, which
+    /// installs a delegation that is missing and is skipped otherwise. A sweep that delegates
+    /// nothing at all carries none, and is then a plain EIP-1559 transaction.
     #[n(6)]
     pub authorizations: Vec<SignedAuthorization>,
+    /// The deposits this sweep moves, one per `(account, token)` pair, in the order the sweep queue
+    /// offered them. Not the shape of `data`, which names one item per deposit address: an account
```
mbjorkqvist, line 290: `gas_limit()` now derives the limit from `deposits`, but nothing ties `deposits` to `data`: [`assert_created_transaction`](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/minter/src/state/transactions/request.rs#L210-L231) checks destination, amount, data and authorizations, but not that the deposits describe the calldata. A request whose deposits under-describe its batch would silently get an undersized limit.

Is the plan to enforce the correspondence at the (still to come) construction site, or should there be an assertion here?

rs/ethereum/cketh/minter/src/tx/sweep.rs
```diff
@@ -12,16 +12,17 @@ use rlp::RlpStream;
...
 
-/// A transaction sent from the minter's dedicated sweeper address: a plain EIP-1559 transaction
-/// (`0x02`), or an EIP-7702 one (`0x04`) whose authorization list additionally installs the
-/// delegation to the sweeper contract of those deposit addresses it sweeps that are not delegated
-/// yet. The ones already delegated are swept without an authorization tuple, so the list can be
-/// shorter than the set of addresses swept.
+/// A transaction sent from the minter's dedicated sweeper address: an EIP-7702 transaction
+/// (`0x04`) whose authorization list carries one tuple per deposit address it sweeps, or a plain
+/// EIP-1559 one (`0x02`) for a sweep that delegates nothing — the native-ETH deposit addresses are
+/// deliberately never delegated, so their sweeps take that shape.
```
mbjorkqvist, line 18: Is this the right justification for the EIP-1559 variant? Per the [design doc](https://github.com/dfinity/ic/blob/098ae8cfee62431ec19bcd69a25fe782354a92db/rs/ethereum/cketh/docs/deposit_from_cex.md?plain=1#L456), a native-ETH sweep is paid by the deposit address itself:

> ETH sweeps need no EIP-7702 at all (deposit pays its own gas, 21'000 gas, cheapest possible)

so it is sent *from* that address rather than from the sweeper address on its own nonce sequence, which is what a `SweepTransaction` is. And `SweptDeposit` requires an `erc20_contract_address`, so a native-ETH sweep cannot be represented in `deposits` either.

Now that every touched address carries a tuple, is `SweepTransaction::Eip1559` reachable for any real sweep?

mbjorkqvist (commented):
Thanks @gregorydemay!

The PR description still describes the first commit. Since ab70606 dropped `SweptDeposit::delegating` and gave every touched address a tuple, two claims under "What" no longer hold:

  - "the sweep queue and the set of delegated addresses are both reconstructible from the event log" — only the sweep queue is; a skipped tuple leaves no trace of whether the address was already delegated.
  - "Only the addresses a sweep delegates pay an authorization cost on top" - every address does now.

gregorydemay:

PR #11331, opened 2026-08-26
feat(cketh): act on a sweep's outcome

gregorydemay:
## Why

- The sweep queue only knew how to hand deposits out: nothing marked them as taken, and nothing recorded what became of a sweep once its transaction was finalized.

## What

- A sweep now owns the deposits it takes: they are marked taken when its request is accepted, so no second sweep can move the same funds.
- A sweep whose transaction failed drops its deposits from the queue. No funds are lost — a reverted sweep moved nothing — but the minter stops trying: the same batch would revert again on the next tick and burn more of the sweeper's gas. What to retry, and how, is DEFI-2981.
- A sweep that moved its deposits releases them: each pair leaves the queue as it entered it, so it can be armed again for the next deposit to the same address.
- A gauge reports where the queue's entries stand (sweepable vs. in flight).

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
@@ -4,10 +4,13 @@ mod tests;
 use crate::attestation::AttestationRequest;
 use crate::deposit_address::DepositAddress;
 use crate::endpoints::{DepositErc20Error, DepositErc20Response, DepositStatus, DetectedDeposit};
+use crate::logs::INFO;
 use crate::numeric::{BlockNumber, Erc20Value};
 use crate::state::event::{AutomaticDeposit, DepositAddressRegistration, DepositAddressRegistry};
+use crate::state::transactions::{SweepId, SweptDeposit};
 use crate::timed_sized_map::{Entry, InsertError, TimedSizedMap, Timestamp};
 use crate::tx::TransactionSignature;
+use ic_canister_log::log;
 use ic_ethereum_types::Address;
 use icrc_ledger_types::icrc1::account::Account;
 use std::collections::BTreeMap;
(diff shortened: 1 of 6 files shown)


rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -303,6 +310,123 @@ impl AutomaticDeposits {
...
+        for deposit in deposits {
+            let request = DepositRequest::new(deposit.account, deposit.erc20_contract_address);
+            let entry = self
+                .sweep
+                .get_mut(&request)
+                .unwrap_or_else(|| panic!("BUG: {request:?} is not queued for sweeping"));
+            assert_eq!(
+                entry.swept_by, None,
+                "BUG: {request:?} was already taken by another sweep"
+            );
```
mbjorkqvist, line 347: IIUC, this could also occur if the `deposits` vec named the same pair twice. Adding some validation to `SweepRequest::deposits` (at construction and deserialization), which is currently missing, would help. Perhaps changing the type to `pub deposits: BTreeMap<(Account, Address), DepositAddress>` would be enough?

rs/ethereum/cketh/minter/src/state/automatic_deposits/tests.rs
```diff
@@ -751,6 +752,277 @@ fn entry(account: &Account, expires_at: Timestamp) -> Entry<ScanProgress> {
...
+    #[test]
+    fn should_let_a_pair_whose_sweep_failed_be_armed_again() {
+        let mut deposits = queued();
+        deposits.record_sweep_scheduled(SweepId(0), &[swept(usdc())]);
+        deposits.record_sweep_failed(SweepId(0), &[swept(usdc())]);
+
+        let armed = deposits.watch_deposit(ts(0), account(0), usdc(), deposit_address(&account(0)));
+
+        assert_matches!(armed, Ok(_));
+    }
```
mbjorkqvist, line 897: This currently happens rather silently, and a user may end up wondering what happened to their deposit. Would it make sense to make this visible in the minter dashboard somehow (probably not in this PR, and maybe it's already tracked somewhere)?

mbjorkqvist (commented):
Thanks @gregorydemay!

gregorydemay:

PR #11333, opened 2026-08-26
chore: Add the `RequestId` to `VetKdArgs`

eichhorl:
This PR adds the VetKD request's `RequestId` to the `VetKdArgs`, and consequently the encrypted key share content, which makes the shares identifiable against the context.

Additionally, we skip malformed shares during share combination. This will speed up response delivery based on the remaining shares.


rs/crypto/src/vetkd/mod.rs
```diff
@@ -344,26 +353,45 @@ fn combine_encrypted_key_shares_internal<C: ThresholdSignatureCspClient>(
...
-            let clib_share = EncryptedKeyShare::deserialize(&share.encrypted_key_share.0).map_err(
-                |e| match e {
-                    EncryptedKeyShareDeserializationError::InvalidEncryptedKeyShare => {
-                        VetKdKeyShareCombinationError::InvalidArgumentEncryptedKeyShare
-                    }
-                },
-            )?;
-            Ok((node_id, *node_index, clib_share))
+                );
+                return None;
```
pierugo-dfinity, line 368: I feel like this could/should also be checked when validating shares (currently not the case in `verify_encrypted_key_share_internal`)

rs/crypto/src/vetkd/mod.rs
```diff
@@ -344,26 +353,45 @@ fn combine_encrypted_key_shares_internal<C: ThresholdSignatureCspClient>(
...
+        .collect();
+    // `ensure_sufficient_shares_to_fail_fast` above already rejected the case of too
+    // few shares having been collected, so getting here means shares were skipped.
+    if clib_shares.len() < reconstruction_threshold {
+        return Err(VetKdKeyShareCombinationError::CombinationError(format!(
+            "only {} of {} encrypted key shares were usable, fewer than the \
+                reconstruction threshold of {reconstruction_threshold}",
+            clib_shares.len(),
+            shares.len()
+        )));
```
pierugo-dfinity, line 393: This looks more like a `::UnsatisfiedReconstructionThreshold`, but I understand if you wanted to differentiate it from a "regular" unsatisfied reconstruction threshold, since here it's a sign of malicious behaviour.

rs/crypto/tests/vetkd.rs
```diff
@@ -478,7 +518,99 @@ mod combine_shares {
...
+        let client = VetKDTestClient::new(&mut rng, &server);
+        let vetkd_args = client.create_args(&server.dkg_id);
+
+        let mut shares = server
+            .create_key_shares(&vetkd_args, &mut rng)
+            .expect("Share creation unexpectedly failed");
+
+        let to_corrupt = shares.len() - server.config.threshold().get().get() as usize;
+        modify_n_random_shares(to_corrupt, &mut shares, &mut rng, |share, _rng| {
+            share.encrypted_key_share.0.truncate(1);
```
pierugo-dfinity, line 533: Looks like an existing test used `corrupt_share_contents`

rs/crypto/tests/vetkd.rs
```diff
@@ -478,7 +518,99 @@ mod combine_shares {
...
+        let victims = shares.keys().copied().choose_multiple(&mut rng, 2);
+        let [malformed, invalid] = victims.as_slice() else {
+            panic!("Missing shares");
+        };
+        shares
+            .get_mut(malformed)
+            .expect("Missing share")
+            .encrypted_key_share
+            .0
+            .truncate(1);
```
pierugo-dfinity, line 581: ditto

rs/crypto/tests/vetkd.rs
```diff
@@ -614,7 +817,10 @@ mod verify_encrypted_key {
             .expect("Share combination failed")
             .1;
 
-        flip_random_bit(&mut vetkd_args.transport_public_key, &mut rng);
+        // Set the infinity bit which causes the point to become an invalid encoding.
+        // Flipping a random bit instead would not do: the y-sign bit merely negates the
+        // point, which still deserializes and thus fails with a VerificationError.
+        vetkd_args.transport_public_key[0] ^= 0x40;
```
pierugo-dfinity, line 823: How is this related to your changes?

rs/types/types/src/crypto/vetkd.rs
```diff
@@ -52,6 +55,31 @@ impl SignedBytesWithoutDomainSeparator for VetKdEncryptedKeyShareContent {
     }
 }
 
+/// The content that a node signs when it creates a vetKD encrypted key share.
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct VetKdEncryptedKeyShareSigningContent<'a> {
+    request_id: RequestId,
```
pierugo-dfinity, line 61: Wouldn't a `callback_id` be sufficient? `RequestId` is a consensus concept that we are leaking into a crypto type here.

rs/crypto/src/vetkd/mod.rs
pierugo-dfinity: (intended to comment on lines 361-368 but, GitHub...)

rs/crypto/src/vetkd/mod.rs
eichhorl: Yes, returning `CombinationError` instead will automatically cause a slack alert, which I think is good

rs/crypto/src/vetkd/mod.rs
eichhorl: Yes, `verify_encrypted_key_share_internal` intentionally only checks the basic signature. This is the "optimized  variant" in the paper where most of the verification is done by the single block maker.

rs/types/types/src/crypto/vetkd.rs
eichhorl: Technically `CallbackId` is similarly a DSM concept. I think it's fine to use `RequestId` which is more consistent with what the chain key client does

rs/crypto/tests/vetkd.rs
eichhorl: `corrupt_share_contents` only flips a random bit, which doesn't guarantee that we will hit the "malformed" branch which is asserted below

rs/crypto/tests/vetkd.rs
eichhorl: same argument

rs/crypto/tests/vetkd.rs
eichhorl: Claude discovered that this test is flaky when only flipping a random bit. If the bit only flips the sign of the point, then it still deserializes.

The same flakiness was fixed the same way previously for `should_err_if_transport_public_key_is_invalid.

rs/crypto/src/vetkd/mod.rs
pierugo-dfinity: Ok, that's cool!

rs/crypto/tests/vetkd.rs
pierugo-dfinity: Ok, I assume this means the test was flaky previously then

rs/crypto/tests/vetkd.rs
pierugo-dfinity: Ah I see, that makes sense

pierugo-dfinity (approved):

PR #11335, opened 2026-08-26
refactor: Hold the whole consumed-cycles total in SubnetMetrics

mraszyk:
`SubnetMetrics::consumed_cycles_by_canisters` holds only the canisters' part of
the total; every consumer adds the subnet-level aggregate on top, recomputing
that part each time it reads. That is fine for a consumer that only ever looks
at a committed state -- the certified tree and the gauge -- but it makes the
combined value ill-defined mid-round: the canisters' part is as of the last
commit, while the subnet-level accumulators move as the round proceeds. A
`delete_canister` adds the deleted canister's consumption to
`consumed_cycles_by_deleted_canisters` at once, so between that and the next
commit the two parts count the same canister twice.

Hold the whole total instead. `consumed_cycles_by_canisters` becomes
`consumed_cycles_total_including_canisters`, still transient, now covering the
subnet-level aggregate as well. The field is private, with a getter of the same
name and a single writer, `SubnetMetrics::refresh_consumed_cycles`, which takes
the canisters' part and adds `consumed_cycles_total()` on top: callers hand over
that part only, so the sum lives in one place and none of them can get it wrong.
The certified state tree at `/subnet/<subnet_id>/metrics` (from certification
version `V29`) and the `replicated_state_consumed_cycles_since_replica_started`
gauge both read the getter, so they cannot drift and any future consumer gets a
self-consistent value whenever it reads.

`ReplicatedState::refresh_consumed_cycles`, renamed from
`refresh_consumed_cycles_by_canisters` now that it publishes more than the
canisters' part, hands the canister states' sum over on every
`commit_and_certify`, and `new_from_checkpoint` does the same on load, so a
replica restarting from a checkpoint agrees with one that keeps running --
which `consumed_cycles_total_is_the_same_across_a_restart` pins, comparing a
live state against one reloaded from the same canisters and subnet metrics.

The certified encoding and the state hash are unchanged: the byte-exact
expectations in `encoding/tests/compatibility.rs` and the `V29` hash in
`state_manager/src/tree_hash.rs` still hold, with their fixtures now handing
their canisters' part to `refresh_consumed_cycles` rather than writing the
aggregate themselves. `test_traverse_subnet_metrics_includes_canister_consumed_cycles_at_v29`
now also consumes subnet-level cycles, so the total it pins covers both parts
rather than only the canisters' one. The gauge does now depend on the refresh
having run, which in production it has, as `commit_and_certify` enqueues the
observation after it; the scheduler metrics tests, which never commit a state,
refresh explicitly through the new `observe_state_metrics` helper.


rs/replicated_state/src/replicated_state.rs
```diff
@@ -488,12 +488,15 @@ impl ReplicatedState {
...
+        // the canisters just loaded. A running replica gets the same value from
+        // `Self::refresh_consumed_cycles`, so the canonical state tree at
         // `/subnet/<subnet_id>/metrics` hashes identically across a restart.
-        metadata.subnet_metrics.consumed_cycles_by_canisters =
-            canister_states.total_consumed_cycles();
+        metadata
+            .subnet_metrics
+            .consumed_cycles_total_including_canisters =
+            metadata.subnet_metrics.consumed_cycles_total()
+                + canister_states.total_consumed_cycles();
```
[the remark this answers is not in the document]
mraszyk: https://github.com/dfinity/ic/pull/11335/commits/4175eb125c08024a6552ee2bb3c1050ce97d17d4

rs/replicated_state/src/metadata_state.rs
```diff
@@ -647,18 +653,6 @@ impl SubnetMetrics {
...
-    /// still-existing canisters: the subnet-level aggregate
-    /// ([`Self::consumed_cycles_total`]) plus [`Self::consumed_cycles_by_canisters`].
-    ///
-    /// Both the certified state tree at `/subnet/<subnet_id>/metrics` (from
-    /// certification version `V29`) and the
-    /// `replicated_state_consumed_cycles_since_replica_started` gauge report this
-    /// same definition, so the two cannot drift apart.
-    pub fn consumed_cycles_total_including_canisters(&self) -> NominalCycles {
-        self.consumed_cycles_total() + self.consumed_cycles_by_canisters
-    }
```
eichhorl, line 660: Alternatively we could keep this function and let it return `consumed_cycles_total_including_canisters` (which could be private), and then add a new function, i.e. `refresh_consumed_cycles(&mut self, consumed_by_canisters: NominalCycles)`, which sets the field after adding `consumed_cycles_total()`.

Then the calculation (and the potential of doing it wrong) disappears from callers

rs/canonical_state/src/traversal.rs
eichhorl, line 1251: We could also consume some subnet level cycles

rs/execution_environment/src/scheduler/tests/metrics.rs
eichhorl: Maybe this could be its own helper

eichhorl: More generally, should we `refresh_consumed_cycles()` in `execute_round(_)`?

rs/replicated_state/src/metadata_state.rs
eichhorl: This comment about the caller seems to be no longer true, maybe there are more leftovers from the previous PR

rs/canonical_state/src/lazy_tree_conversion.rs
```diff
@@ -1145,9 +1145,10 @@ fn subnets_as_tree<'a>(
                         "metrics",
                         // Starting with `V29`, the reported total also includes
                         // the cycles consumed by all non-deleted canisters, read
-                        // from `SubnetMetrics::consumed_cycles_by_canisters`
+                        // from the stored
+                        // `SubnetMetrics::consumed_cycles_total_including_canisters`
                         // (refreshed by
-                        // `ReplicatedState::refresh_consumed_cycles_by_canisters`).
+                        // `ReplicatedState::refresh_consumed_cycles`).
```
eichhorl: Are these assumptions actually relevant here?

rs/replicated_state/src/metadata_state.rs
```diff
@@ -440,15 +440,21 @@ pub struct SubnetMetrics {
...
+    /// All cycles removed from circulation on this subnet, by both deleted and
+    /// still-existing canisters: [`Self::consumed_cycles_total`] plus the sum of
+    /// `CanisterMetrics::consumed_cycles()` over the canisters that currently
+    /// exist, as of the end of the last committed round.
     ///
     /// Derived, not persisted: refreshed by
-    /// `ReplicatedState::refresh_consumed_cycles_by_canisters` when a state is
-    /// committed, and re-derived by `ReplicatedState::new_from_checkpoint` on load.
+    /// `ReplicatedState::refresh_consumed_cycles` when a state is committed, and
+    /// re-derived by `ReplicatedState::new_from_checkpoint` on load.
```
eichhorl: Maybe remove this

eichhorl (approved):

rs/replicated_state/src/metadata_state.rs
mraszyk: why? isn't the lifecycle actually useful to remark in the doc comment of the field?

eichhorl: Sure, as long as it is kept in sync when the lifecycle changes 😄. I guess AI should be able to find these cross-references (evidently in this case it did)

rs/execution_environment/src/scheduler/tests/metrics.rs
mraszyk: > Maybe this could be its own helper

What do you refer to exactly?

> More generally, should we refresh_consumed_cycles() in execute_round(_)?

That would only be possible at the very end of `StateMachine::execute_round`, but the risk of this approach is that more operations could be added afterwards that modify the cycles balances.

eichhorl: Referring to new lines 1062-1070, which seem to be repeated 8 times in this file, but no strong opinion

PR #11337, opened 2026-08-26
perf(cketh): reuse the delegation authorization a deposit address signed

gregorydemay:
## Why

- Every sweep signed a fresh EIP-7702 authorization for each deposit address it touched, paying a threshold-ECDSA signature for a tuple that never changes: nonce 0, one chain, one delegate.

## What

- The minter remembers the authorization a deposit address signed and reuses it on every later sweep of that address; only addresses without a usable stored tuple cost a signature.
- A stored tuple is checked against the configuration before reuse — one naming a retired sweeper contract or another chain misses and is simply re-signed.
- The store lives on the heap alone: losing it costs one signature per address, never a sweep, so an upgrade may empty it rather than the event log carrying an entry per address forever.
- A gauge reports the store's size, so growth is visible before it needs bounding.

### DIFF
--- a/rs/ethereum/cketh/minter/src/main.rs
+++ b/rs/ethereum/cketh/minter/src/main.rs
@@ -1279,6 +1279,12 @@ fn http_request(req: HttpRequest) -> HttpResponse {
                     "Number of deposit address attestations the minter has signed and stored.",
                 )?;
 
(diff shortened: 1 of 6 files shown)


gregorydemay:

PR #11342, opened 2026-08-27
fix(cketh): provision sweep gas out of the sweeper balance bound

gregorydemay:
## Why

- The funding task decides whether to top the sweeper address up from a lower bound on its ETH balance, tracked from the minter's own events (#11086). Nothing debited that bound, which was true while nothing spent from the sweeper address — and the layers below connect the sweep queue to the pipeline that does.
- Without a debit the bound only ever grows: fundings top the sweeper up once, sweeps spend the gas, and every counter still reports a full sweeper — `amount_due` declines fundings the sweeper actually needs, and sweeping stalls while looking healthy.

## What

- An accepted sweep provisions the most it can cost the sweeper address: its fee ceiling, the same ceiling every resubmission is checked against. A sweep moves its ERC-20 through call data and sends no ETH — settlement reads the finalized transaction and its receipt, asserts exactly that, hands back the fee the sweep did not pay, and records the fee it did.
- A sweep is only *accepted* while the bound covers its ceiling, so every hold is backed by delivered gas. That lets the balance be checked like the main address' `EthBalance` instead of floored at zero: provisioning beyond deliveries is a bug to surface, not a state to tolerate (no deployed minter ever wrote a sweep event, so there is no legacy log to indulge).
- Provisioning at acceptance rather than debiting at spend keeps the bound erring *low* while sweeps are in flight: a sweep whose finalization is never observed brings a funding forward rather than letting the minter believe in gas that is gone. Acceptance also happens exactly once per sweep, where a signature happens once per resubmission.
- The holds stay out of `cumulative_spent` — that ETH was counted there when the funding delivered it, and counting it twice would trip the burn-first invariant (R14).
- The sweep e2e now funds the sweeper the way production does — ckETH burned from the fee account, delivered and finalized by the funding pipeline — since a minter that trusts only its own accounting no longer sweeps on gas credited behind its back.

This is #11329 (thanks @mbjorkqvist), integrated on top of the sweep stack per the hand-off on DEFI-2933: rebased onto the sweep-queue lifecycle, whose deposit release now lives in the same finalization transition the refund settles in.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -527,12 +527,12 @@ impl State {
     }
 
     /// Takes the whole cost an accepted sweep can put on the sweeper address out of the balance
-    /// bound: the ETH it will transfer plus its fee ceiling, which caps every resubmission the
-    /// pipeline makes for it. An ERC-20 sweep moves its tokens through call data and carries no
-    /// ETH value, so the fee is all it can cost.
+    /// bound: its fee ceiling, which caps every resubmission the pipeline makes for it. An ERC-20
+    /// sweep moves its tokens through call data and carries no ETH value, so the fee is all it
+    /// can cost.
     pub fn update_sweeper_balance_upon_accepted_sweep(&mut self, request: &SweepRequest) {
         self.sweeper_funding
-            .record_accepted_sweep(Wei::ZERO, request.max_transaction_fee);
+            .record_accepted_sweep(request.max_transaction_fee);
     }
 
     fn update_balance_upon_withdrawal(
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -497,10 +497,16 @@ impl State {
         sweep_id: &SweepId,
         receipt: &TransactionReceipt,
     ) {
-        let _ = self
+        let max_transaction_fee = self
+            .automatic_deposits
+            .processed_sweep_request(sweep_id)
+            .expect("BUG: missing sweep request")
+            .max_transaction_fee;
+        let finalized = self
             .automatic_deposits
             .record_finalized_sweep_transaction(*sweep_id, receipt);
-        self.update_sweeper_balance_upon_sweep(sweep_id, receipt);
+        self.sweeper_funding
+            .record_finalized_sweep(max_transaction_fee, &finalized);
     }
 
     pub fn next_request_id(&mut self) -> u64 {
@@ -520,36 +526,13 @@ impl State {
         };
     }
 
-    /// Takes the most an accepted sweep can cost the sweeper address out of the balance bound: its
-    /// fee ceiling, which caps every resubmission the pipeline makes for it. An ERC-20 sweep moves
-    /// its tokens through call data and carries no ETH value, so the fee is all it can cost.
+    /// Takes the whole cost an accepted sweep can put on the sweeper address out of the balance
+    /// bound: the ETH it will transfer plus its fee ceiling, which caps every resubmission the
+    /// pipeline makes for it. An ERC-20 sweep moves its tokens through call data and carries no
+    /// ETH value, so the fee is all it can cost.
     pub fn update_sweeper_balance_upon_accepted_sweep(&mut self, request: &SweepRequest) {
         self.sweeper_funding
-            .record_accepted_sweep(request.max_transaction_fee);
-    }
-
-    /// Hands back the part of a finalized sweep's provision it did not need: the fee it did not
-    /// pay, which is the whole of it either way, a reverted sweep being charged for its gas like
-    /// any other.
-    fn update_sweeper_balance_upon_sweep(
-        &mut self,
-        sweep_id: &SweepId,
-        receipt: &TransactionReceipt,
-    ) {
-        let fee_ceiling = self
-            .automatic_deposits
-            .processed_sweep_request(sweep_id)
-            .expect("BUG: missing sweep request")
-            .max_transaction_fee;
-        // Cannot underflow: a sweep transaction is created, and resubmitted, only while its fee
-        // stays within this ceiling, and the fee it actually pays is at most the fee it was signed
-        // for.
-        let effective_fee = receipt.effective_transaction_fee();
-        let unspent_fee = fee_ceiling
-            .checked_sub(effective_fee)
-            .expect("BUG: a sweep may not pay more than its fee ceiling");
-        self.sweeper_funding
-            .record_finalized_sweep(effective_fee, unspent_fee);
+            .record_accepted_sweep(Wei::ZERO, request.max_transaction_fee);
     }
 
     fn update_balance_upon_withdrawal(
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -16,7 +16,9 @@ use crate::runtime::CanisterRuntime;
 use crate::state::automatic_deposits::{AutomaticDeposits, ScanProgress};
 use crate::state::eth_logs_scraping::{LogScrapingId, LogScrapings};
 use crate::state::sweeper_funding::{SweeperFundingAccounting, SweeperFundingConfig};
-use crate::state::transactions::{Erc20WithdrawalRequest, TransactionCallData, WithdrawalRequest};
+use crate::state::transactions::{
+    Erc20WithdrawalRequest, SweepRequest, TransactionCallData, WithdrawalRequest,
+};
 use crate::timed_sized_map::{Entry, Timestamp};
 use crate::tx::AuthorizationRequest;
 use crate::tx::GasFeeEstimate;
@@ -487,6 +489,20 @@ impl State {
         self.update_balance_upon_withdrawal(withdrawal_id, receipt);
     }
 
+    /// Finalizes a sweep: releases the deposits it held and settles what it actually cost the
+    /// sweeper address. The sweeper pipeline is never reimbursed and holds no ckETH balance, so
+    /// unlike the main pipeline there is no reimbursement tail and no ckETH accounting here.
+    pub fn record_finalized_sweeper_transaction(
+        &mut self,
+        sweep_id: &SweepId,
+        receipt: &TransactionReceipt,
+    ) {
+        let _ = self
+            .automatic_deposits
+            .record_finalized_sweep_transaction(*sweep_id, receipt);
+        self.update_sweeper_balance_upon_sweep(sweep_id, receipt);
+    }
+
     pub fn next_request_id(&mut self) -> u64 {
         let current_request_id = self.http_request_counter;
         // overflow is not an issue here because we only use `next_request_id` to correlate
@@ -502,6 +518,36 @@ impl State {
                 .erc20_balances
                 .erc20_add(event.erc20_contract_address, event.value),
         };
+    }
+
+    /// Takes the most an accepted sweep can cost the sweeper address out of the balance bound: its
+    /// fee ceiling, which caps every resubmission the pipeline makes for it. An ERC-20 sweep moves
+    /// its tokens through call data and carries no ETH value, so the fee is all it can cost.
+    pub fn update_sweeper_balance_upon_accepted_sweep(&mut self, request: &SweepRequest) {
+        self.sweeper_funding
+            .record_accepted_sweep(request.max_transaction_fee);
+    }
+
+    /// Hands back the part of a finalized sweep's provision it did not need: the fee it did not
+    /// pay, which is the whole of it either way, a reverted sweep being charged for its gas like
+    /// any other.
+    fn update_sweeper_balance_upon_sweep(
+        &mut self,
+        sweep_id: &SweepId,
+        receipt: &TransactionReceipt,
+    ) {
+        let fee_ceiling = self
+            .automatic_deposits
+            .processed_sweep_request(sweep_id)
+            .expect("BUG: missing sweep request")
+            .max_transaction_fee;
+        // Cannot underflow: a sweep transaction is created, and resubmitted, only while its fee
+        // stays within this ceiling, and the fee it actually pays is at most the fee it was signed
+        // for.
+        let unspent_fee = fee_ceiling
+            .checked_sub(receipt.effective_transaction_fee())
+            .expect("BUG: a sweep may not pay more than its fee ceiling");
+        self.sweeper_funding.record_finalized_sweep(unspent_fee);
     }
 
     fn update_balance_upon_withdrawal(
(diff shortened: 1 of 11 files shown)


rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -73,12 +108,25 @@ impl SweeperFundingAccounting {
...
-    /// debits it yet; when sweeping lands it will subtract the gas submitted sweeps provisioned, so
-    /// the bound stays conservative. ETH sent to the address by anyone else only pushes the true
-    /// balance above it.
+    /// A lower bound on the sweeper address' ETH balance: what finalized fundings delivered, less
+    /// what accepted sweeps have provisioned, plus what finalized sweeps handed back.
+    ///
+    /// Provisioning at acceptance rather than at spend is what keeps this a bound while sweeps are in
+    /// flight: gas a committed sweep will pay stops counting as available immediately, and a sweep
+    /// whose finalization is never observed leaves the bound too *low* — which delays a funding —
+    /// rather than too high, which would let the minter believe in gas that is gone. ETH sent to the
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed against the code — the doc had the direction backwards. `SweeperFundingConfig::amount_due` returns `None` only while `sweeper_balance >= low_water_mark` (`state/sweeper_funding.rs:178-187`), so a bound that reads too low crosses the mark *sooner*, and the amount it then asks for is `target - sweeper_balance`, i.e. larger too. `plan_funding`'s own doc already said so ("Erring low can only make a funding look due sooner than it is, never hide one that is due", `sweeper.rs:149-150`), so the two were contradicting each other.

Fixed in ee264fb9f2: the bound now reads as bringing a funding forward and making it larger, with the pointer to `amount_due`. The safety argument is unchanged — erring low is still the safe direction, just for the reason the planner gives.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
```diff
@@ -62,6 +73,30 @@ impl SweeperFundingAccounting {
...
+    pub fn record_accepted_sweep(&mut self, provisioned: Wei) {
+        self.cumulative_sweep_provisioned = self
+            .cumulative_sweep_provisioned
+            .checked_add(provisioned)
+            .expect("BUG: overflow in cumulative sweep provisioned");
+    }
+
+    /// Records the part of a finalized sweep's provision it did not need, putting it back into the
+    /// balance bound.
+    pub fn record_finalized_sweep(&mut self, refunded: Wei) {
```
mbjorkqvist: `cumulative_sweep_provisioned` is only ever offset here, so a sweep that never finalizes keeps its provision permanently.

That's reachable. A sweep's ceiling is priced once at acceptance (`sweep/mod.rs:158-161`) and never re-priced. Once `base + priority` exceeds `floor(ceiling / gas_limit)` — roughly a doubling of the base fee — `create_transaction` fails and the request is rescheduled indefinitely (`sweep/mod.rs:394-402`); `reschedule_request` only moves it within `pending_requests`, and nothing else takes it out. Not a spike problem, since those mean-revert — the realistic case is a sweep accepted in a quiet trough at ~0.5 gwei, whose ~1.5 gwei ceiling can then sit unaffordable for weeks.

Small per event (~0.005–0.02 ETH at normal gas, against a 0.3 ETH target), but permanent, cumulative, and currently invisible — nothing in `sweeper_funding` is exported as a metric.

The signal worth alerting on is the **age of the oldest unfinalized sweep**, not the amount: age is independent of load and gas price, where any wei-denominated gauge moves with the base fee and can't separate "busy" from "leaking". The withdrawal pipeline already has exactly this — `oldest_incomplete_request_timestamp()` (`transactions/mod.rs:1409-1414`), exported as `cketh_oldest_incomplete_eth_withdrawal_request_age_seconds` (`main.rs:1313-1321`) — and the pieces for the sweep equivalent are already in place: `SweepRequest.created_at` is public, and the "incomplete" enumeration is the one `outstanding_sweeper_funding` already uses (`transactions/mod.rs:1223-1229`).

```rust
pub fn oldest_incomplete_sweep_timestamp(&self) -> Option<u64> {
    self.incomplete_requests_iter().map(|r| r.created_at).min()
}
```

A wei gauge is still worth having to size the impact on the bound, but under a steady stream it never returns to zero, so it wants a floor comparison (`min_over_time(...[24h])` ratcheting upward) rather than a threshold on the level:

```rust
/// Provision still held by sweeps that have not finalized.
pub fn outstanding_sweep_provision(&self) -> Wei {
    self.cumulative_sweep_provisioned
        .checked_sub(self.cumulative_sweep_refunded)
        .unwrap_or(Wei::ZERO)
}
```

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
```diff
@@ -207,3 +207,153 @@ mod config {
...
+        state.sweeper_funding.record_burn(Wei::new(DELIVERED));
+        state
+            .sweeper_funding
+            .record_finalized_funding(Wei::new(DELIVERED), Wei::ZERO);
+    }
+
+    /// Drives the already-accepted `request` through the sweeper pipeline to a receipt of `status`,
+    /// priced below its ceiling so that part of the provision comes back, and returns the fee that
+    /// receipt charged.
+    fn finalize(state: &mut State, request: &SweepRequest, status: TransactionStatus) -> Wei {
```
mbjorkqvist, line 450: Would it make sense to add a test that, after `SignedSweeperTransaction`, apply `ReplacedSweeperTransaction` at a higher price still under the ceiling, finalize on the replacement's receipt, and assert the bound moved by that receipt's fee rather than by two ceilings?

Two smaller gaps:

- No test with two accepted sweeps. `cumulative_sweep_provisioned` is a running total, and every test uses one sweep, so nothing would catch an overwrite instead of an add.
- `should_floor_the_bound_at_zero_rather_than_trap` never calls `deliver()`, so `cumulative_transferred == 0` and it only proves `0 - x` saturates. The case the doc describes - an upgrade that restarts the counters, or a sweeper funded before it was tracked - is a nonzero delivery smaller than the provision. Calling `deliver()` with an amount below `request.max_transaction_fee` exercises that path.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
gregorydemay: 🤖 Agreed on both the leak and the signal. Confirmed at head: `cumulative_sweep_provisioned` is only ever offset by `record_finalized_sweep`, and a sweep priced in a trough can sit rescheduling forever, keeping its provision. Two updates since this comment:

- e9dfee9104 adds a guard: a sweep is only *accepted* while the balance bound covers its fee ceiling, and since a leaked provision permanently lowers the bound, the funding task now tops the sweeper back up to target — so the leak's cost is earlier/larger fundings (real ETH parked at the sweeper), not blocked sweeps.
- The age metric you describe (`cketh_oldest_incomplete_sweep_age_seconds`, mirroring `cketh_oldest_incomplete_eth_withdrawal_request_age_seconds`) is the right observable and the pieces exist as you note. Not in this PR yet — happy to add it here or as a follow-up, whichever you prefer.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
gregorydemay: 🤖 Status on the three gaps at the current head (e9dfee9104):

- **Floor with nonzero delivery: now covered.** The sweep fixture funds the sweeper with 1 ETH (`PREPAID_SWEEP_GAS`) before acceptance — required since sweeps are now only accepted when the bound covers them — and `should_floor_the_bound_at_zero_rather_than_trap` records an extra provision of a further 1 ETH on top of the accepted sweep's ceiling, so it exercises exactly the shape you describe: nonzero delivery smaller than total provision.
- **Replacement-under-the-ceiling refund test: still missing.** Worth having — `update_sweeper_balance_upon_sweep` reads the ceiling off the *processed request* and the fee off the receipt it's given, so finalizing on a replacement's receipt should hand back ceiling minus that receipt's fee, once, and a test would pin that it isn't debited per signed transaction.
- **Two accepted sweeps: still missing.** Agreed that nothing currently distinguishes a running total from an overwrite.

Happy to add the two missing tests on this PR if you want them here.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -297,6 +300,36 @@ impl State {
...
+    /// What every deposit address in `accounts` authorizes to let the configured sweeper contract
+    /// sweep it: the tuple naming this minter's chain, that contract, and nonce zero. `None` while
+    /// no sweeper contract is configured.
+    ///
+    /// The nonce is always zero, whatever the address actually holds. A deposit address is at
+    /// nonce zero exactly while it has never been delegated — applying an authorization spends it
+    /// — so the tuple either installs the delegation or is skipped, and both are correct in any
+    /// order the sweeps carrying them land. That is what lets a sweep authorize every address it
+    /// touches without tracking which ones are already delegated, at the price of the intrinsic
+    /// gas a skipped tuple still costs.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed at the current head: `authorization_requests` pins nonce 0 (`state.rs:326`), and its doc explains why that is correct for *installing* the delegation — but you're right that it makes replacing the delegate ineffective for any address already swept: the applied authorization spent nonce 0, so the new delegate's tuple is skipped on-chain, and `should_sign_a_fresh_authorization_when_the_sweeper_contract_changes` only pins the cache key, not the on-chain outcome. A replaced delegate would also have its batch entry point calling addresses still running the old delegate's code.

Two honest mitigating facts: the old delegate's helper and payout address are immutable and attested, so stranded addresses keep sweeping into the minter's control — this is an upgrade-flexibility gap, not a fund-safety one; and no deployed minter has a sweeper contract configured yet, so nothing on mainnet is delegated.

Proposal: treat delegate replacement as unsupported for now — a follow-up ticket to either refuse the `UpgradeArg` once any address is delegated, or track authority nonces and sign replacements with the current value, with the change-of-contract test run against a real EVM as you suggest. This PR is about gas provisioning, so I'd rather not fold that in here.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -786,10 +804,16 @@ type Event = record {
             deposit_helper : text;
             owner : principal;
             subaccount : opt blob;
-            y_parity : bool;
-            // 32-byte signature components.
-            r : blob;
-            s : blob;
+            // The attestation signed by the deposit address itself.
+            attestation : TransactionSignature;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Two reasons this reshape doesn't strand any client. First, `get_events` is explicitly documented in the .did as "meant as a debugging tool and is not guaranteed to be backwards-compatible" (cketh_minter.did:969). Second, and decisively: no deployed minter has ever emitted an `AttestedDepositAddress` event — the whole attestation/sweep feature requires a sweeper contract configured by upgrade, which no released minter has — so there are no historical events in any audit log for a pre-reshape client to fail on. The old spelling existed only inside this unreleased stack; by the time a release ships these variants, the released .did and the emitted events carry the same (new) shape.

rs/ethereum/cketh/minter/cketh_minter.did
```diff
@@ -664,14 +684,12 @@ type Event = record {
         AcceptedSweepRequest : record {
             sweep_id : nat;
             destination : text;
-            amount : nat;
-            // Transaction call data (the delegate sweep call).
-            data : blob;
+            // The single ERC-20 contract this sweep moves.
+            token : text;
+            // The deposits the sweep moves, one per account.
+            items : vec AuthorizedSweepItem;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Same situation as the `AttestedDepositAddress` thread: `get_events` is documented as a debugging endpoint with no backwards-compatibility guarantee (cketh_minter.did:969), and no released minter can have emitted an `AcceptedSweepRequest` — the sweep pipeline only runs with a sweeper contract configured, which exists on no deployment. The pre-reshape fields (`amount`, `data`, `authorizations`, the old `SignedAuthorization`) never made it into any release, so there is no client-visible history to preserve; versioning the event API would be defending a wire format that was never shipped.

rs/ethereum/cketh/minter/src/state/sweeper_funding/tests.rs
gregorydemay: 🤖 Correcting my earlier status, which the design's evolution made stale. As of f5c3ff1125 the accounting changed shape twice: the sweeper's balance is now its own two-field record (`eth_balance` + `total_effective_tx_fees`) debited like the main address' `EthBalance` with checked arithmetic, and settlement reads the finalized transaction and its receipt, asserting a sweep sends no ETH.

- **The floor test is gone because the floor is gone**: since sweeps are only accepted while the bound covers their ceiling, an uncovered provision is a bug and panics — pinned by `should_panic_when_provisioning_more_than_was_delivered`, which provisions past a nonzero 1 ETH delivery, i.e. the exact shape you asked the old floor test to exercise, with the opposite (now intended) outcome.
- **Replacement-under-the-ceiling refund test: still missing**, and still worth having: settlement refunds ceiling minus the *finalized* receipt's fee, once, whichever resubmission won.
- **Two accepted sweeps: still missing**, and more relevant than before — the bound is now a running balance rather than recomputed cumulative counters, so two sequential holds are exactly the add-vs-overwrite shape nothing pins.

Happy to add both tests here if you want them on this PR.

gregorydemay: 🤖 Both remaining tests landed in 1835071c03:

- `should_settle_on_the_receipt_of_the_resubmission_that_finalized`: creates and signs at the fixture estimate, replaces via `ReplacedSweeperTransaction` at a different price, finalizes on the replacement's receipt, and asserts the balance moved by exactly that receipt's fee — one refund, measured against the request's ceiling. One honest note: your "higher price still under the ceiling" variant is unconstructible — `create_transaction` allocates the whole allowance up front (`max_fee_per_gas = ceiling/gas_limit`), so no replacement can outbid the original; the test re-prices the priority fee downward instead, which pins the same settlement property.
- `should_hold_every_accepted_sweep`: two accepted sweeps, distinct ceilings, balance down by their sum — the add-vs-overwrite shape, now on a running balance where it matters more.

With the floor superseded by the checked-balance panic test, all three gaps are covered; resolving.

rs/ethereum/cketh/minter/src/state/sweeper_funding.rs
gregorydemay: @mbjorkqvist The metric is a good idea but I see it as part of DEFI-2965, I added a bullet there.

mbjorkqvist (approved):
LGTM, thanks @gregorydemay!

PR #11344, opened 2026-08-27
chore: Cleanup

michael-weigelt:
Some renaming and comments.


rs/config/src/embedders.rs
eichhorl, line 227: Do I understand correctly that we are talking about two different kinds of pages here? And that a Wasm page may cover multiple OS pages?
```suggestion
    /// The number of instructions to charge for every OS page of heap or stable
    /// memory that a message touches: once when the Wasm page is first accessed and
    /// once more when it is first written to. The default value should be
```

rs/config/src/subnet_config.rs
eichhorl, line 281: Also here
```suggestion
    /// The number of instructions to charge for every OS page of heap or stable
    /// memory that a message touches: once when the Wasm page is first accessed and
    /// once more when it is first written to.
```

rs/embedders/src/wasmtime_embedder.rs
```diff
@@ -907,17 +905,13 @@ pub struct PageAccessResults {
     pub wasm_accessed_os_pages_count: usize,
     /// Non-deterministic number of accessed Wasm (64 KiB) pages (read + write).
     pub wasm_accessed_wasm_pages_count: usize,
```
eichhorl, line 907: Are these still non-deterministic with the deterministic memory tracker? Or is it unrelated?

rs/embedders/src/wasmtime_embedder.rs
```diff
@@ -1423,17 +1403,10 @@ fn accessed_os_and_wasm_pages(accessed_pages: &[PageIndex]) -> (usize, usize) {
     (accessed_pages.len(), wasm_pages.len())
 }
 
-fn dirty_os_and_wasm_pages(
-    speculatively_dirty_pages: &[PageIndex],
-    dirty_pages: &[PageIndex],
-) -> (usize, usize) {
-    let wasm_pages: HashSet<u64> = speculatively_dirty_pages
+fn dirty_os_and_wasm_pages(dirty_pages: &[PageIndex]) -> (usize, usize) {
```
eichhorl, line 1406: This seems to do the same thing as the function above

rs/interfaces/src/execution_environment.rs
```diff
@@ -53,14 +53,6 @@ pub struct InstanceStats {
     /// Non-deterministic number of dirty Wasm (64 KiB) pages (write).
     pub wasm_dirty_wasm_pages_count: usize,
```
eichhorl, line 54: Same determinism question

rs/memory_tracker/benches/traps.rs
```diff
@@ -1,77 +1,77 @@
...
+        PageMap::new_for_testing(),
+        MemoryLimits {
+            max_memory_size: NumBytes::new(WASM_PAGE_SIZE_IN_BYTES as u64),
+            max_dirty_pages: NumOsPages::new((WASM_PAGE_SIZE_IN_BYTES / PAGE_SIZE) as u64),
+        },
+        /* page_overhead */ 0,
+        Arc::new(SignalMutex::new(|_| {})),
+    )
+    .unwrap()
+}
```
eichhorl, line 55: Claude is telling me that it will not reset the whole region to `PROT_NONE` if the page map is empty (like `PageMap::new_for_testing()` is)

rs/memory_tracker/src/deterministic.rs
eichhorl, line 291: This seems to be the only caller now, so we could also pass the range only once

rs/execution_environment/benches/system_api/execute_update.rs
```diff
@@ -604,9 +604,7 @@ pub fn execute_update_bench(c: &mut Criterion) {
                 Wasm64::Disabled,
             ),
             // stable_read writes to Wasm heap (+32) and reads from stable memory (+16)
```
eichhorl, line 606: Are these instruction counts still correct?

rs/memory_tracker/src/tests.rs
eichhorl, line 29: Claude complains that there are some test gaps in this file, but that could also be addressed in a follow up.

rs/embedders/src/wasmtime_embedder/signal_handler.rs
```diff
@@ -4,14 +4,14 @@ use crate::wasmtime_embedder::host_memory::MemoryPageSize;
...
+    DeterministicMemoryTracker, signal_access_kind_and_address, signal_mutex::SignalMutex,
 };
 use std::convert::TryFrom;
 use std::sync::{Arc, atomic::Ordering};
 
 /// Helper function to create a memory tracking SIGSEGV handler function.
 pub(crate) fn sigsegv_memory_tracker_handler(
-    memories: Vec<(Arc<SignalMutex<SigsegvMemoryTracker>>, MemoryPageSize)>,
+    memories: Vec<(Arc<SignalMutex<DeterministicMemoryTracker>>, MemoryPageSize)>,
 ) -> impl Fn(i32, *const libc::siginfo_t, *const libc::c_void) -> bool + Send + Sync {
```
eichhorl, line 15: Is the function name and comment still up to date?

eichhorl (approved):
As far as I can tell the only difference compared to master is this commit: https://github.com/dfinity/ic/pull/11344/changes/4009c549b40a70d1816e96a2485ceada20a6e47d, which looks good to me. But claude also had some comments about the rest, so I'm posting those as well

rs/config/src/embedders.rs
michael-weigelt: Yes, a Wasm page is 64KiB, a Linux page is 4KiB and a MacOS Page is 16KiB. The comment here is only concerned with OS pages though, because that's what actually incurs a cost.

rs/embedders/src/wasmtime_embedder/signal_handler.rs
michael-weigelt: Yes, all our memory trackers are sigsegv trackers, and DMT is a specific implementation.

PR #11345, opened 2026-08-27
refactor(cketh): make event recording generic over a TimeProvider

gregorydemay:
Reading the current time was an ambient dependency on the IC system API buried deep in the ckETH minter's event-sourcing paths, which made anything that records an event impossible to unit test outside a canister.

This introduces a `TimeProvider` trait (new `time` module) with the production implementation delegating to `ic_cdk::api::time()`, and makes `record_event` — and every caller up to the canister entry points in `main.rs` — generic over it. Only `main.rs` names the concrete implementation.

A mockall-based mock of the trait is added to the crate's test fixtures, and the balance-scan tests now use it, including a new test asserting a detected deposit is recorded with the time the provider returns.

Follows the `CanisterRuntime` pattern already used in the ckBTC minter.

### DIFF
--- a/rs/ethereum/cketh/minter/src/sweep.rs
+++ b/rs/ethereum/cketh/minter/src/sweep.rs
@@ -39,7 +39,7 @@ const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SIGN_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SEND_BATCH_SIZE: usize = 5;
 
-pub async fn process_sweeper_transactions<T: TimeProvider + 'static>(time_provider: T) {
+pub async fn process_sweeper_transactions<T: TimeProvider>(time_provider: T) {
     let _guard = match TimerGuard::new(TaskType::SweeperSend) {
         Ok(guard) => guard,
         Err(e) => {
@@ -91,7 +91,7 @@ pub async fn process_sweeper_transactions<T: TimeProvider + 'static>(time_provid
     }
 }
 
-fn schedule_retry<T: TimeProvider + 'static>(time_provider: T) {
+fn schedule_retry<T: TimeProvider>(time_provider: T) {
     ic_cdk_timers::set_timer(
         crate::PROCESS_SWEEPER_TRANSACTIONS_RETRY_INTERVAL,
         async move { process_sweeper_transactions(time_provider).await },
--- a/rs/ethereum/cketh/minter/src/sweep.rs
+++ b/rs/ethereum/cketh/minter/src/sweep.rs
@@ -63,7 +63,7 @@ pub async fn process_sweeper_transactions<T: TimeProvider>(time_provider: T) {
         return;
     };
 
-    let gas_fee_estimate = match lazy_refresh_gas_fee_estimate().await {
+    let gas_fee_estimate = match lazy_refresh_gas_fee_estimate(&time_provider).await {
         Some(gas_fee_estimate) => gas_fee_estimate,
         None => {
             // The withdrawal task refreshes the same estimate under a shared guard and runs first
--- a/rs/ethereum/cketh/minter/src/sweep.rs
+++ b/rs/ethereum/cketh/minter/src/sweep.rs
@@ -20,6 +20,7 @@ use crate::{
         mutate_state, read_state,
         transactions::{CreateSweepTransactionError, PipelineRequest},
     },
+    time::TimeProvider,
     tx::{GasFeeEstimate, lazy_refresh_gas_fee_estimate},
     withdraw::{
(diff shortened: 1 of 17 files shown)


rs/ethereum/cketh/minter/src/sweep.rs
```diff
@@ -38,7 +39,7 @@ const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SIGN_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SEND_BATCH_SIZE: usize = 5;
 
-pub async fn process_sweeper_transactions() {
+pub async fn process_sweeper_transactions<T: TimeProvider + 'static>(time_provider: T) {
```
mbjorkqvist: `lazy_refresh_gas_fee_estimate()` that is called from `process_sweeper_transactions()` (and also from `process_retrieve_eth_requests()`) internally calls `ic_cdk::api::time()` - would it make sense to pass `TimeProvider` through there?

rs/ethereum/cketh/minter/src/time/mod.rs
```diff
@@ -0,0 +1,20 @@
+/// Source of the current time.
+///
+/// Abstracting the system API away lets code that timestamps its actions be exercised without a
+/// canister environment.
+pub trait TimeProvider {
```
mbjorkqvist: Would it make sense to make this `pub trait TimeProvider: Clone + 'static` to avoid having inconsistent bounds in the various places it is used? On a higher level, would it make sense to pass this as `&dyn TimeProvider` rather than using generics?

mbjorkqvist (approved):
Thanks @gregorydemay, just a couple of comments/questions!

rs/ethereum/cketh/minter/src/sweep.rs
gregorydemay: 🤖 Good catch — done in 175bd9b76b.

`lazy_refresh_gas_fee_estimate` reached for `ic_cdk::api::time()` twice: once for the cache freshness check, once to stamp `last_transaction_price_estimate`. Both sit on the sweep and withdrawal paths, so a mock clock did not actually isolate them from the real one. It now takes a `&T: TimeProvider`; the two pipeline callers pass theirs through and `estimate_erc20_transaction_fee` in `main.rs` passes `IC_TIME_PROVIDER`.

rs/ethereum/cketh/minter/src/time/mod.rs
gregorydemay: 🤖 Took the supertrait in 08787a39eb — `pub trait TimeProvider: Clone + 'static`. The bounds were indeed spelled three different ways (`TimeProvider`, `+ 'static` where a timer takes it by value, `+ Clone + 'static` where `deposit.rs` clones it into a spawned task); they all collapse to `T: TimeProvider` now, and both implementors already satisfied it — `IcTimeProvider` is a `Copy` ZST and `MockTimeProvider` has a `Clone` impl in its `mock!` block.

On `&dyn TimeProvider`: sticking with generics. Beyond preferring static dispatch here, the two suggestions are mutually exclusive — `Clone` is not object-safe, so the supertrait rules out `dyn TimeProvider`. And a borrowed `&dyn` would not have worked at the timer sites anyway (`schedule_retry`, `deposit.rs`) since the provider is moved into a `'static` future; that would have forced either `&'static dyn` — which needs the mocks leaked, silently disabling mockall's drop-time verification of the `.times(1)` expectation in `balance_scan/tests.rs` — or `Arc<dyn ...>`. Neither buys anything over a monomorphised `T`.

Worth noting for #11347: since `CanisterRuntime: TimeProvider`, the bound propagates there, and the same three-way spelling shows up on that branch too, so it collapses the same way. `IcCanisterRuntime` and `MockCanisterRuntime` both already derive/impl `Clone`. The cost is that `CanisterRuntime` is permanently non-object-safe — despite `#[async_trait]`, which is usually reached for precisely to enable `dyn` — and a future runtime implementor cannot borrow (`'static`) or hold non-`Clone` state. If you expect either, the alternative is to drop the supertrait and instead put `Clone + 'static` on the handful of call sites that genuinely need it, keeping the traits object-safe.

.bazel_gate.sh
basvandijk, line 1: Does this need to be checked in?

gregorydemay: definitely not, that's a mistake from claude.

gregorydemay: 🤖 No — that was committed by mistake, thanks for catching it. It is a local scratch script for running the bazel query/test gate, swept in by an overly broad `git add -A`. The branch has since been rewritten and the file is gone entirely: the current head (175bd9b76b) never contains it, so it is not in this PR's diff.

basvandijk (approved):

PR #11347, opened 2026-08-27
feat(cketh): task skeleton to create sweeper requests

gregorydemay:
## Why

- The sweep queue that the balance scan fills has nothing turning its deposits into sweeps, and the signing that will do so needs a canister — which is what makes it untestable today.

## What

- An async `CanisterRuntime` trait extending `TimeProvider` with tECDSA signing as its first capability, plus the implementation backed by the system API and a mock for tests.
- A skeleton `create_pending_sweeper_requests`, its task type, and the timer that runs it.
- The skeleton does nothing yet, and a unit test pins that down: a deposit sits in the sweep queue, the runtime carries no expectation, and the whole state comes back unchanged. Each later commit then shows up as a diff against a passing test.

### DIFF
--- a/rs/ethereum/cketh/minter/Cargo.toml
+++ b/rs/ethereum/cketh/minter/Cargo.toml
@@ -16,6 +16,7 @@ path = "bin/principal_to_hex.rs"
 
 [dependencies]
 askama = { workspace = true }
+async-trait = { workspace = true }
 canbench-rs = { workspace = true, optional = true }
 candid = { workspace = true }
 ethnum = { workspace = true }
--- a/rs/ethereum/cketh/minter/src/tx/mod.rs
+++ b/rs/ethereum/cketh/minter/src/tx/mod.rs
@@ -27,6 +27,7 @@ use crate::{
     guard::TimerGuard,
     logs::{DEBUG, INFO},
     numeric::{GasAmount, Wei, WeiPerGas},
+    runtime::CanisterRuntime,
     state::{TaskType, lazy_call_ecdsa_public_key_with_chain_code, mutate_state, read_state},
     time::TimeProvider,
 };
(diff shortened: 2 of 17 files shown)


rs/ethereum/cketh/minter/Cargo.toml
```diff
@@ -16,6 +16,7 @@ path = "bin/principal_to_hex.rs"
 
 [dependencies]
 askama = { workspace = true }
+async-trait = { workspace = true }
```
mbjorkqvist, line 19: nit: It seems we are using `mockall` version `0.13.1`, whereas if we were to upgrade to `0.15` (or even `0.14`), we could drop this production dependency on `async-trait`.

mbjorkqvist (approved):
Thanks @gregorydemay!

Typo nit for the PR title: *skeleton* instead of *skeletton*.

rs/ethereum/cketh/minter/Cargo.toml
gregorydemay: I will try let Claude in Slack handle that, let's see if this works

gregorydemay:
@mbjorkqvist asking for a review because 9be14d24f2917d248ca4906074f8b86b23d14bde was added after your approval 🙏

rs/ethereum/cketh/minter/src/tx/mod.rs
```diff
@@ -182,18 +183,20 @@ pub fn encode_u256<T: Into<u256>>(stream: &mut RlpStream, value: T) {
...
-    .await
-    .map_err(|e| format!("failed to sign digest: {e}"))?;
+    let signature = runtime
+        .sign_with_ecdsa(
+            key_name,
+            derivation_path.iter().map(|path| path.to_vec()).collect(),
+            digest.0,
+        )
+        .await
+        .map_err(|e| format!("failed to sign digest: {e}"))?;
```
mbjorkqvist, line 199: It seems it's not (yet) possible to test this, since `sign_digest` calls `compute_recovery_id` on the line below which is not behind a runtime (similar issue as with `lazy_call_ecdsa_public_key_with_chain_code` earlier).

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/tx/mod.rs
gregorydemay: 🤖 Good catch — fixed in 588e861980.

`CanisterRuntime` now also exposes `ecdsa_public_key`, and `lazy_call_ecdsa_public_key_with_chain_code` (plus `lazy_call_ecdsa_public_key` and `minter_address`) go through it, so `compute_recovery_id` is no longer ambient. `sign_digest` is now driven entirely off the mock — `tx::tests::sign_digest::should_sign_digest_through_the_runtime` mocks both the signature and the public key and checks the recovered parity bit, with the `debug_assert!` inside `compute_recovery_id` verifying the signature against the derived key along the way.

PR #11349, opened 2026-08-27
feat: all-in-one nodes for cloud engines

r-birkner:
Cloud engines nodes will be "all-in-one" nodes. This means they don't just run the replica, but also `ic-gateway` and potentially also `ic-boundary`. The orchestrator already supports running multiple processes, but we can't just start `ic-gateway`, we also need to provide it with the right, engine-specific config. That's what this PR is doing:

This PR adds an additional service to the orchestrator that fetches the HTTP Gateway config from the engine operator canister and uses it to run `ic-gateway`.

The config includes the domain names, DNS API URLs and API key (for the DNS-01 ACME challenge to obtain the certificates), and the ACME account. That's the bare minimum for `ic-gateway` to run. In the future, the config will be extended.

The config is stored in the engine operator canister, which resides on the engine itself. The orchestrator first has to discover this canister using the engine management canister. Then, it needs to use the node's identity to make the call as the config is protected and can only be accessed by the nodes making up the engine.


rs/config/src/cloud_engine.rs
```diff
@@ -0,0 +1,16 @@
...
+
+/// Configuration that only cloud engine nodes use.
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize, Serialize)]
+pub struct Config {
+    /// Principal of the engine management canister, which an all-in-one node
+    /// needs to discover the operator canister of its own engine. When unset,
+    /// the node cannot find its operator and therefore does not run `ic-gateway`.
+    #[serde(default)]
+    pub engine_management_canister_id: Option<CanisterId>,
+}
```
pierugo-dfinity, line 16: Did we consider putting that information to the registry instead of the config? Personally, having a hard-coded canister ID inside the config feels weird + the registry can be modified if we need to + I think it would be easier to mock in tests.

And the orchestrator would know whether to launch the `CloudEngineManager` or not based on its node reward type, the same way it tells it is a cloud-engine node for the registry replicator & firewall.

But this is not a strong opinion, just interested in what you think

rs/ic_os/config/tool/src/guestos/generate_ic_config.rs
```diff
@@ -200,6 +201,21 @@ fn get_config_vars(guestos_config: &GuestOSConfig) -> Result<IcConfigTemplate> {
...
+        .or(matches!(
+            guestos_config.icos_settings.deployment_environment,
+            DeploymentEnvironment::Mainnet
+        )
+        .then_some(ic_config::cloud_engine::MAINNET_ENGINE_MANAGEMENT_CANISTER_ID))
+    {
+        Some(id) => serde_json::to_string(id)
+            .context("Failed to encode the engine management canister id")?,
+        None => "null".to_string(),
+    };
```
pierugo-dfinity, line 217:
```suggestion
    let engine_management_canister_id = match guestos_config
        .guestos_settings
        .engine_management_canister_id
        .as_deref() {
            Some(id) => id,
            None if guestos_config.icos_settings.deployment_environment == DeploymentEnvironment::Mainnet => ic_config::cloud_engine::MAINNET_ENGINE_MANAGEMENT_CANISTER_ID,
            None => "null",
    }
    .to_string();
```
Maybe something like this would clarify the control flow. I'm not sure why use `serde_json::to_string` either

rs/orchestrator/src/error.rs
```diff
@@ -81,6 +81,9 @@ pub(crate) enum OrchestratorError {
 
     /// The given node is missing a domain name at the given registry version.
     DomainNameMissingError(NodeId, RegistryVersion),
+
+    /// A step of the cloud engine configuration lookup failed.
+    CloudEngineError(String),
```
pierugo-dfinity, line 86: There are a lot of different `Error` enums (`FetchError`, `OperatorError`, `ConfigError`), yet not complete (`AgentFactory` and `Discovery` return `OrchestratorError`s).
- Would it be feasible to flatten the existing errors into one big `CloudEngineManagementError` enum holding every possible variant? Maybe in its own `orchestrator/src/cloud_engine/error.rs`
- I don't think we need a new variant in `OrchestratorError`. Existing callers of `cloud_engine_error` can have their own enums (`AgentFactory`/`Discovery`) or use a different variant (in `registry_helper.rs` and in `fn env_overlay`.

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+mod config;
+mod discovery;
+mod operator;
+
+use crate::{
+    error::OrchestratorError, metrics::OrchestratorMetrics, registration::NodeRegistrationCrypto,
+    registry_helper::RegistryHelper,
+};
+use config::ConfigError;
+pub(crate) use config::GatewayConfig;
```
pierugo-dfinity, line 19: No need to re-export, I think callers can import the submodule directly

basvandijk (approved):
Approve for the files under `rs/tests/driver/src/driver/`.

A PR description would be nice for a >2k diff ;)

rs/config/src/cloud_engine.rs
```diff
@@ -0,0 +1,16 @@
...
+pub const MAINNET_ENGINE_MANAGEMENT_CANISTER_ID: &str = "q6cfj-fyaaa-aaaar-qb77q-cai";
+
+/// Configuration that only cloud engine nodes use.
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize, Serialize)]
+pub struct Config {
+    /// Principal of the engine management canister, which an all-in-one node
+    /// needs to discover the operator canister of its own engine. When unset,
+    /// the node cannot find its operator and therefore does not run `ic-gateway`.
+    #[serde(default)]
+    pub engine_management_canister_id: Option<CanisterId>,
```
pierugo-dfinity, line 15: Note that despite the very misleading documentation, `CanisterId`'s implementation of `Deserialize` does **not** check that it is a valid canister ID (it could be any principal) because `CanisterId::try_from` does not actually try, it just casts without validating.

https://github.com/dfinity/ic/blob/c2cebaff70d5331903f5f58c91d426acbf7eb807/rs/types/base_types/src/canister_id.rs#L351-L363

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
...
+    ) -> OrchestratorResult<Agent> {
+        if let Some(agent) = &self.operator_agent {
+            return Ok(agent.clone());
+        }
+
+        // Reading the public keys is an RPC to the CSP vault that blocks
+        // on `block_on` internally.
+        let crypto = Arc::clone(&self.crypto);
+        #[allow(clippy::disallowed_methods)]
+        let public_keys = tokio::task::block_in_place(|| crypto.current_node_public_keys());
```
blind-oracle, line 114: Use `spawn_blocking` instead of `block_in_place` - https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
...
+            .ok_or_else(|| {
+                OrchestratorError::cloud_engine_error("the node signing public key is missing")
+            })?;
+
+        let crypto = Arc::clone(&self.crypto);
+        let sign = move |message: &MessageId| {
+            // `sign_basic` blocks on an RPC to the crypto service, which panics
+            // when called from an async context without this.
+            #[allow(clippy::disallowed_methods)]
+            tokio::task::block_in_place(|| {
```
blind-oracle, line 131: Here too

rs/orchestrator/src/processes.rs
```diff
@@ -464,23 +564,19 @@ impl MultipleProcessesManager {
...
+                .registry
+                .is_cloud_engine_subnet(subnet_id, registry_version)?
+            {
+                let gateway_config = self.gateway_config.read().unwrap().clone();
+                self.ic_gateway_manager
+                    .ensure_running_and_restarted_on_config_change(replica_version, gateway_config)
+            } else {
+                self.ic_gateway_manager.stop()
+            },
+        );
```
pierugo-dfinity, line 579:
```suggestion
        if self
            .registry
            .is_cloud_engine_subnet(subnet_id, registry_version)?
        {
            let gateway_config = self.gateway_config.read().unwrap().clone();
            result = result.and(    
                self.ic_gateway_manager
                    .ensure_running_and_restarted_on_config_change(replica_version, gateway_config);
        } else {
            result = result.and(self.ic_gateway_manager.stop());
        }
```
nit: easier to read imo

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// The environment that overrides the shipped `ic-gateway.env`, which only
+    /// carries policy (which challenge, which DNS backend, which ports).
+    ///
+    /// The two credentials go into the environment rather than the argument
+    /// list: arguments are logged by the process runner and are world-readable
+    /// through `/proc/<pid>/cmdline`, the environment is neither.
+    pub(crate) fn env_overlay(
+        &self,
+        acme_cache_dir: &Path,
+    ) -> OrchestratorResult<Vec<(String, String)>> {
```
pierugo-dfinity, line 111: Probably it makes sense to return a `HashMap<OsString, OsString>`

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+use std::{fmt, path::Path};
+use url::Url;
+
+/// A complete, validated engine configuration.
+///
+/// `ic-gateway` terminates TLS for the engine, so it cannot run without all of
+/// these. An incomplete config is therefore not an error but simply nothing to
+/// apply, which [`TryFrom`] reports as [`Incomplete`](ConfigError::Incomplete).
+#[derive(Clone, PartialEq, Eq)]
+pub(crate) struct GatewayConfig {
```
pierugo-dfinity, line 16: This is really just a thought, but this doesn't look like it changes over time (it looks static). Maybe we could rename it to something that indicates it's dynamic (also to differentiate it from `IcGatewayProcessConfig`)

rs/orchestrator/src/processes.rs
```diff
@@ -367,6 +392,89 @@ impl IcBoundaryManager {
...
+        Self {
+            inner,
+            current_config: None,
+        }
+    }
+
+    /// Runs `ic-gateway` with `gateway_config`, restarting it when the config
+    /// changed. `None` means the engine configuration is not (yet) known, in
+    /// which case `ic-gateway` must not run at all: it terminates TLS for the
+    /// engine and has nothing to serve without it.
```
pierugo-dfinity, line 430: What about plain HTTP? `ic-gateway` still serves through port 80, do we absolutely not want to serve anything until we have a config, or wouldn't it be too bad to launch `ic-gateway` without TLS first, serve some clients and once we have a config in the future, restart it?

Maybe fetching the config could fail transiently for quite some time (on a subnet upgrade f.ex.)

rs/orchestrator/src/processes.rs
```diff
@@ -367,6 +392,89 @@ impl IcBoundaryManager {
...
+        let Some(gateway_config) = gateway_config else {
+            return self.stop();
+        };
+
+        // Restart only on a change we actually observed: with nothing applied
+        // yet there is nothing to compare against, and `ensure_running` is a
+        // no-op when the process is already up. The explicit stop is needed
+        // because the process runner only compares versions, so an
+        // environment-only change would otherwise not take effect.
+        if self.current_config.is_some() && self.current_config.as_ref() != Some(&gateway_config) {
```
pierugo-dfinity, line 445:
```suggestion
        if self.current_config.as_ref() != Some(&gateway_config) {
```
If `self.current_config` is `None`, it doesn't hurt to stop (keeps it consistent with `ic-boundary`)

rs/orchestrator/src/processes.rs
```diff
@@ -367,6 +392,89 @@ impl IcBoundaryManager {
...
+    ) -> OrchestratorResult<()> {
+        let Some(gateway_config) = gateway_config else {
+            return self.stop();
+        };
+
+        // Restart only on a change we actually observed: with nothing applied
+        // yet there is nothing to compare against, and `ensure_running` is a
+        // no-op when the process is already up. The explicit stop is needed
+        // because the process runner only compares versions, so an
+        // environment-only change would otherwise not take effect.
```
pierugo-dfinity, line 444:
```suggestion
        // no-op when the process is already up.
```
Implementation detail (we anyways shouldn't really hit the described branch since we never run different processes' versions within the same orchestrator)

rs/orchestrator/src/processes.rs
```diff
@@ -367,6 +392,89 @@ impl IcBoundaryManager {
...
+        if self.current_config.is_some() && self.current_config.as_ref() != Some(&gateway_config) {
+            self.inner.stop()?;
+            // `stop` only signals the process. While the old process is still
+            // draining, neither start nor record the new config: the next call
+            // still sees the old config, signals the process again (in case it
+            // ignored the first SIGTERM), and starts the new one once the old
+            // one is gone.
+            if self.inner.process_runner.is_running() {
+                return Ok(());
+            }
```
pierugo-dfinity, line 454: Hmmmmm that's a good finding. Maybe there's a cleaner way to handle that, I can think about it.

It also affects `ic-boundary`.

rs/orchestrator/src/processes.rs
```diff
@@ -164,21 +171,39 @@ impl Process for IcGatewayProcess {
...
+                        "unable to read ic-gateway environment variables".to_string(),
+                        e,
+                    ));
+                }
+            };
+
+        // `ic-gateway` writes its ACME account and certificates here.
+        std::fs::create_dir_all(&config.acme_cache_dir).map_err(|e| {
+            OrchestratorError::IoError(format!("Failed to create {:?}", config.acme_cache_dir), e)
+        })?;
```
pierugo-dfinity, line 197: This feels out of place. Maybe we could move it to `OrchestratorArgs::create_dirs`

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// has nothing to do.
+    pub(crate) fn new(
+        registry: Arc<RegistryHelper>,
+        crypto: Arc<dyn NodeRegistrationCrypto>,
+        engine_management_canister_id: Option<CanisterId>,
+        replica_listen_addr: SocketAddr,
+        current_config: Arc<RwLock<Option<GatewayConfig>>>,
+        metrics: Arc<OrchestratorMetrics>,
+        logger: ReplicaLogger,
+    ) -> Option<Self> {
```
pierugo-dfinity, line 70: Not a big fan of an optional constructor, I think we can leave that responsibility to the caller (check for the existence of the engine management canister, and pass the URL directly)

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            ConfigError::Invalid(_) => Self::Failed(err.to_string()),
+        }
+    }
+}
+
+fn unix_timestamp() -> i64 {
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .map_or(0, |since_epoch| since_epoch.as_secs() as i64)
+}
```
pierugo-dfinity, line 267: `ic_types::time::current_time`

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /// Refreshes the engine configuration, keeping the previous one on failure.
+    pub(crate) async fn check(&mut self) {
+        let version = self.registry.get_latest_version();
+        let subnet_id = match self.registry.get_subnet_id(version) {
+            Ok(subnet_id) => subnet_id,
+            // Unassigned nodes are not part of an engine.
+            Err(_) => return,
+        };
```
pierugo-dfinity, line 111: Hmmm subnet membership is a bit tricky, I'd reuse the shared `subnet_assignment` that the upgrade loop, SSH key manager, and dashboard already share.

rs/orchestrator/src/orchestrator.rs
```diff
@@ -357,6 +367,16 @@ impl Orchestrator {
...
+        let cloud_engine_manager = CloudEngineManager::new(
+            Arc::clone(&registry),
+            Arc::clone(&crypto) as _,
+            config.cloud_engine.engine_management_canister_id,
+            config.http_handler.listen_addr,
+            gateway_config,
+            Arc::clone(&metrics),
+            logger.clone(),
+        );
+
```
pierugo-dfinity, line 379: A possibility would be to run that task only if we are a cloud engine node (by checking one's own node reward type) to not run a useless task on regular nodes and we could get rid of the `self.registry.is_cloud_engine_subnet` check at the beginning of `check`

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+                warn!(
+                    every_n_seconds => 60,
+                    self.logger, "Could not determine the type of subnet {}: {}", subnet_id, err
+                );
+                return;
+            }
+        }
+
+        match self.fetch(subnet_id, version).await {
+            Ok(config) => {
```
pierugo-dfinity, line 132:
```suggestion
            Ok(new_config) => {
```

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            Ok(parts) => parts,
+            Err(err) => {
+                // A cold `isNode` cache is expected; anything else suggests we
+                // resolved the wrong canister, so look it up again next time.
+                if matches!(err, OperatorError::Failed(_)) {
+                    self.discovery.invalidate();
+                }
+                return Err(err.into());
+            }
+        };
```
pierugo-dfinity, line 225: This is hard to read. By trying to refactor the error-handling of `check` to also call `discovery.invalidate` in case this returns an `OperatorError::Failed` (which already does it for `FetchError::NotReady`), it should be possible to rewrite this as follows.
```suggestion
        let gateway_config = operator.http_gateway_config().await?;
        let acme_credentials = operator.acme_credentials().await?;
```
 I think this relates to my other comment that generally the huge number of error enums and their casts/transitions are a bit awkward and difficult to follow.

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+/// operator. A cold `isNode` cache resolves itself with the operator's next
+/// registry refetch, so a genuinely fresh operator recovers well within this
+/// budget; an operator that keeps not recognizing this node more likely is not
+/// (or no longer) our operator at all. At one check every 10 seconds this is
+/// about 5 minutes.
+const MAX_CONSECUTIVE_NOT_READY: u32 = 30;
+
+pub(crate) struct CloudEngineManager {
+    registry: Arc<RegistryHelper>,
+    agent_factory: agent::AgentFactory,
```
pierugo-dfinity, line 47:
```suggestion
    agent_factory: AgentFactory,
```

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+            })
+    }
+
+    /// Accepts `candidate` only if the local registry independently confirms it
+    /// is a canister on `own_subnet` and an admin of it. This is what keeps a
+    /// hostile node of the management subnet from pointing us at an arbitrary
+    /// canister.
+    fn validate(
+        &self,
+        candidate: PrincipalId,
```
pierugo-dfinity, line 152: This really just should be a `CanisterId` I think. And make `ask_management_canister` return a `CanisterId` too.

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+                    "the engine management canister does not know an operator for this subnet",
+                )
+            })
+    }
+
+    /// Accepts `candidate` only if the local registry independently confirms it
+    /// is a canister on `own_subnet` and an admin of it. This is what keeps a
+    /// hostile node of the management subnet from pointing us at an arbitrary
+    /// canister.
+    fn validate(
```
pierugo-dfinity, line 150: This function name is not very descriptive

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+
+    /// Forgets the resolved id, so the next [`Self::resolve`] asks the engine
+    /// management canister again.
+    pub(super) fn invalidate(&mut self) {
+        self.resolved = None;
+    }
+
+    async fn ask_management_canister(
```
pierugo-dfinity, line 103: It is not clear from the function name what we are "ask"ing without docs. Also, "management canister" clashes with the regular `ic_00`.

Having a proper error type will also help this function...

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+        if !self
+            .registry
+            .get_subnet_canister_ranges(own_subnet, version)?
+            .iter()
+            .any(|range| range.contains(&candidate))
+        {
+            return Err(OrchestratorError::cloud_engine_error(format!(
+                "operator candidate {candidate} is not hosted by subnet {own_subnet}"
+            )));
+        }
```
pierugo-dfinity, line 171: Do we check that in case of a canister migration/subnet split?

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+use url::Url;
+
+/// Value of the `outcome` label of `cloud_engine_config_fetches`.
+const OUTCOME_OK: &str = "ok";
+const OUTCOME_INCOMPLETE: &str = "incomplete";
+const OUTCOME_NOT_READY: &str = "not_ready";
+const OUTCOME_ERROR: &str = "error";
+
+/// How many consecutive `NotReady` answers to tolerate before re-resolving the
+/// operator. A cold `isNode` cache resolves itself with the operator's next
```
pierugo-dfinity, line 38: What is this `isNode` cache the docs mention three times?

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::sync::Arc;
+use url::Url;
+
+/// Builds the agents that [`super::CloudEngineManager`] talks to canisters with.
+///
+/// Query signature verification is enabled on every agent, so all of them need
+/// the NNS root key. Unlike [`crate::registration`], a missing root key is an
+/// error here rather than a fallback to the key hard-coded in `ic-agent`:
+/// falling back would make every call fail outside mainnet, which is harder to
+/// diagnose than refusing to build the agent.
```
pierugo-dfinity, line 25: I think it's better to stay consistent with what registration does, no?

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        Ok(agent)
+    }
+
+    fn build<I: Identity + 'static>(
+        &self,
+        url: Url,
+        identity: I,
+        version: RegistryVersion,
+        client: Option<reqwest::Client>,
```
pierugo-dfinity, line 152: Is this argument ever `Some`?

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
+//! Construction of the `ic-agent`s used to reach the engine management canister
```
pierugo-dfinity, line 1: I see the point of having an agent to reach the operator canister, because it needs crypto, replica_url, etc. But I don't think we need anything fancy to build an anonymous agent to an API BN except a `&dyn RegistryClient`, so I would personally move `anonymous_via_api_boundary_node` (and, transitively, `build`) out of `AgentFactory`.

Then if we do that, I do not see what big added value this factory brings us despite caching the agent. `CloudEngineManager` could hold an `Agent` directly that it could build in its constructor and call it a day.

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,173 @@
...
+                )
+            })
+    }
+
+    /// An agent aimed at this node's own replica and signing as this node. The
+    /// operator canister only serves the nodes of its own engine, so the calls
+    /// have to be authenticated with the node signing key.
+    ///
+    /// Built once and then reused: everything it depends on is stable.
+    pub(super) fn node_signed_to_local_replica(
```
pierugo-dfinity, line 102: This function looks very suspiciously similar (if not identical) to `registration.rs::try_to_register_key`. I think we can factor out common functionality to `signer.rs`

rs/orchestrator/src/cloud_engine/operator.rs
```diff
@@ -0,0 +1,109 @@
...
+//! interface can grow without a breaking change; candid drops wire fields that
+//! are absent from these types, so added fields cannot break here either.
+
+use candid::{CandidType, Decode, Encode, Principal};
+use ic_agent::Agent;
+use ic_types::CanisterId;
+use serde::Deserialize;
+
+#[derive(CandidType, Deserialize, Debug)]
+pub(super) enum OperatorApiError {
```
pierugo-dfinity, line 17:
```suggestion
enum OperatorApiError {
```

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+
+#[derive(Debug)]
+pub(crate) enum ConfigError {
+    /// At least one field is not configured yet. Nothing to apply, not a failure.
+    Incomplete(&'static str),
+    /// A configured field is unusable, e.g. a domain `ic-gateway` would reject.
+    Invalid(String),
+}
+
+impl TryFrom<(HttpGatewayConfig, AcmeCredentials)> for GatewayConfig {
```
pierugo-dfinity, line 40: This does way more work than a regular "cast" from one type to the other. I think it could be its own free function

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn try_from(
+        (gateway, acme): (HttpGatewayConfig, AcmeCredentials),
+    ) -> Result<Self, Self::Error> {
+        let base_domains = gateway
+            .base_domains
+            .filter(|domains| !domains.is_empty())
+            .ok_or(ConfigError::Incomplete("base_domains"))?;
+        // `ic-gateway` parses DOMAIN as a comma-separated list of FQDNs, so a
+        // value it would reject must not reach it: it would exit at startup.
+        let base_domains = base_domains
```
pierugo-dfinity, line 52:
```suggestion
            .ok_or(ConfigError::Incomplete("base_domains"))?
        // `ic-gateway` parses DOMAIN as a comma-separated list of FQDNs, so a
        // value it would reject must not reach it: it would exit at startup.
```
nit

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+}
+
+/// Redacts the credentials, so that a `GatewayConfig` is safe to log.
+impl fmt::Debug for GatewayConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("GatewayConfig")
+            .field("base_domains", &self.base_domains)
+            .field("dns_api_url", &self.dns_api_url.as_str())
```
pierugo-dfinity, line 160: Just to make sure: the DNS API URL doesn't contain anything sensitive (like query parameters or whatnot)?

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+    #[test]
+    fn empty_config_is_incomplete() {
+        assert_matches!(
+            parse(HttpGatewayConfig::default(), AcmeCredentials::default()),
+            Err(ConfigError::Incomplete("base_domains"))
+        );
+    }
+
+    #[test]
+    fn every_missing_field_is_incomplete() {
```
pierugo-dfinity, line 223: style nit: we could use rstest here (but this is fine too)

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    #[test]
+    fn resolved_id_is_served_without_asking_again() {
+        let mut discovery = discovery_for_test(&[operator().get()]);
+        discovery.resolved = Some(operator());
+
+        assert_eq!(
+            discovery.validated_from_memory(SUBNET_1, VERSION),
+            Some(operator())
+        );
```
pierugo-dfinity, line 310: `assert_eq!(discovery.resolved, Some(operator()));`

rs/orchestrator/src/cloud_engine/discovery.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    #[test]
+    fn own_subnet_admin_canister_is_accepted() {
+        let discovery = discovery_for_test(&[operator().get()]);
+
+        assert_matches!(
+            discovery.validate(operator().get(), SUBNET_1, VERSION),
+            Ok(accepted) if accepted == operator()
+        );
```
pierugo-dfinity, line 262: `assert_eq!(discovery.resolved, Some(operator()));`

rs/orchestrator/src/cloud_engine/mod.rs
pierugo-dfinity, line 267: Some unit tests for this module would be more than welcome 🙏 (especially the invalidation logic)

pierugo-dfinity (commented):
Are we confident launching this without a system test?

Would there be a way to test that the different arguments to different methods of different canisters do not drift away from production?

frankdavid (approved):
👍🏻 for node

rs/orchestrator/src/processes.rs
r-birkner: Independent of HTTPS or HTTP, it still needs to know its own domain name. That one is also coming from the engine operator canister. So, `ic-gateway` cannot really start without fetching the config. The only thing we could do to serve clients faster after an upgrade, would be to somehow cache the last config on disk and use that.

rs/orchestrator/src/cloud_engine/mod.rs
r-birkner: In the engine operator canister, there is access control, which checks that the caller is a node of the subnet it is a part of. I will adapt the comment to reflect that/not use `isNode` anymore.

rs/orchestrator/src/cloud_engine/config.rs
r-birkner: This is all as intended: the DNS API config consists of the URL and the key (below). The key is sensitive, the URL not.

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,358 @@
...
+    pub base_domains: Vec<String>,
+    pub dns_api_url: Url,
+    pub dns_api_key: String,
+    pub acme_account: AcmeAccount,
+}
+
+/// An `instant_acme::AccountCredentials`. `key_pkcs8` is the account private key
+/// as PKCS#8 DER, base64url without padding.
+#[derive(Clone, PartialEq, Eq, Serialize)]
+pub(crate) struct AcmeAccount {
```
blind-oracle, line 26: Do we really need to dive into ACME account internals? Maybe we can just store it as opaque `Vec<u8>` that is JSON internally. Since that's what `ic-gateway` expects anyway.

blind-oracle (changes requested):

rs/orchestrator/src/cloud_engine/agent.rs
pierugo-dfinity: I think `block_in_place` is necessary, see `sign_basic`'s [documentation](https://github.com/dfinity/ic/blob/a071296a7b9619ffd7522b469975c5a3a3c1f6b5/rs/interfaces/src/crypto/sign.rs#L54-L59). It is also inline with how the orchestrator [rotates its IDKG key](https://github.com/dfinity/ic/blob/a071296a7b9619ffd7522b469975c5a3a3c1f6b5/rs/orchestrator/src/registration.rs#L541-L550)

rs/orchestrator/src/processes.rs
blind-oracle: Serving over plain HTTP is baaad. We do only HTTPS redirect there.

rs/orchestrator/src/cloud_engine/agent.rs
blind-oracle: Ah it's a sync method, then I guess it's correct yes.

rs/orchestrator/src/cloud_engine/discovery.rs
r-birkner: I haven't thought that far. We just check as the engine management canister could give us some garbage, but still the damage is minimal: either the canister will just reject the call as the caller is not authorized (e.g., we call an honest operator on a different engine) or it could give us a malicious operator canister and then that canister could just give it a bad domain name/acme account. I will remove all these checks as they don't really provide much and just add complexity.

rs/orchestrator/src/processes.rs
r-birkner: Claude found it. Not me 🫣

rs/orchestrator/src/error.rs
r-birkner: I have removed all the errors as suggested.

rs/orchestrator/src/cloud_engine/agent.rs
r-birkner: Yes, changed the behavior to match registration.rs

rs/ic_os/config/tool/src/guestos/generate_ic_config.rs
```diff
@@ -203,6 +207,26 @@ fn get_config_vars(guestos_config: &GuestOSConfig) -> Result<IcConfigTemplate> {
...
+            == DeploymentEnvironment::Mainnet =>
+        {
+            Some(ic_config::cloud_engine::MAINNET_ENGINE_MANAGEMENT_CANISTER_ID)
+        }
+        None => None,
+    }
+    .map(serde_json::to_string)
+    .transpose()
+    .context("Failed to encode the engine management canister id")?
+    .unwrap_or_else(|| "null".to_string());
```
pierugo-dfinity: (as you prefer)
`serde_json::to_string` should natively serialize `None` to `null`, so this could be replaced with
```rust
let engine_management_canister_id = serde_json::to_string(match guestos_config {
...
})
.context("...")?;
```

rs/orchestrator/src/processes.rs
```diff
@@ -619,6 +740,226 @@ mod tests {
...
+    fn ensure_gateway(
+        manager: &mut IcGatewayManager,
+        engine_config: Option<EngineConfig>,
+    ) -> OrchestratorResult<()> {
+        manager.ensure_running_and_restarted_on_config_change(
+            ReplicaVersion::try_from(REPLICA_VERSION).unwrap(),
+            engine_config,
+        )
+    }
+
```
pierugo-dfinity, line 782: Could we assert on the manager's `current_config` in the tests below?

rs/orchestrator/src/processes.rs
```diff
@@ -619,6 +740,226 @@ mod tests {
...
+                config.clone(),
+            ),
+        )
+        .unwrap();
+
+        // Arguments are logged by the process runner and are world-readable via
+        // /proc/<pid>/cmdline; the environment is neither.
+        let args = format!("{:?}", process.get_args());
+        assert!(!args.contains(&config.dns_api_key), "{args}");
+        assert!(!args.contains(&config.acme_account.key_pkcs8), "{args}");
```
pierugo-dfinity, line 964: In constrast, should we assert that this key is present inside the environment?

rs/orchestrator/src/orchestrator.rs
```diff
@@ -368,6 +376,32 @@ impl Orchestrator {
             logger.clone(),
         );
 
+        // A node without an engine management canister cannot discover its
+        // operator, so it has no engine configuration to fetch.
+        let cloud_engine_manager = config.cloud_engine.engine_management_canister_id.and_then(
+            |engine_management_canister_id| {
```
pierugo-dfinity: Hey sorry I know I [suggested](https://github.com/dfinity/ic/pull/11349#discussion_r3872730959) to not run the task if we were not a cloud engine node. But this style is sensibly different than the other task constructors. So I'd just keep unconditionally calling `CloudEngineManager::new` and let the latter return an `Option` like before.

P.S. IIUC this will also run for non-cloud engine nodes since on mainnet, that config field always exists.

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,101 @@
...
+    node_ids.shuffle(&mut thread_rng());
+    node_ids
+        .iter()
+        .find_map(|node_id| {
+            let domain = registry.get_node_record(*node_id, version).ok()??.domain?;
+            Url::parse(&format!("https://{domain}/"))
+                .inspect_err(|err| {
+                    warn!(
+                        logger,
+                        "Ignoring the malformed API boundary node domain '{}': {}", domain, err
```
pierugo-dfinity:
```suggestion
                        "Ignoring the malformed API boundary node domain '{domain}': {err}",
```
nit: for consistency with the rest of the file

rs/orchestrator/src/cloud_engine/agent.rs
```diff
@@ -0,0 +1,101 @@
...
+    match nns_root_key_der_from_registry(registry, version) {
+        Ok(root_key) => {
+            agent.set_root_key(root_key);
+        }
+        Err(err) => {
+            // If we cannot determine the NNS public key, we log a warning but still proceed. The
+            // agent will use the mainnet public key hardcoded in the agent library.
+            warn!(
+                logger,
+                "Failed to get NNS public key from registry: {}", err
```
pierugo-dfinity:
```suggestion
                "Failed to get NNS public key from registry: {err}",
```

rs/orchestrator/src/cloud_engine/config.rs
```diff
@@ -0,0 +1,365 @@
...
+        Ok([
+            ("ACME_CACHE_PATH", acme_cache_dir.display().to_string()),
+            ("DOMAIN", self.base_domains.join(",")),
+            (
+                "ACME_DNS_IC_DNS_LB_URLS",
+                self.dns_api_urls
+                    .iter()
+                    .map(Url::to_string)
+                    .collect::<Vec<_>>()
+                    .join(","),
```
pierugo-dfinity, line 123: It looks like `Url::parse("https://google,com")` parses just fine, so we probably need to add some validation that rejects commas in the URL.

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+            engine_management_canister_id,
+            logger.clone(),
+        );
+
+        Self {
+            registry,
+            subnet_assignment,
+            crypto,
+            replica_url,
+            operator_agent: None,
```
pierugo-dfinity: If `CloudEngineManager` actually returns an `Option` (following my suggestion to revert not spawning the task if not a cloud engine node), we could build `operator_agent` once here in the constructor (and avoid having to store `crypto` and `replica_url`).

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+        };
+        let version = self.registry.get_latest_version();
+
+        let outcome = match self.registry.is_cloud_engine_subnet(subnet_id, version) {
+            Ok(true) => self.fetch(subnet_id, version).await,
+            // Only all-in-one nodes have an engine operator to ask.
+            Ok(false) => return,
+            Err(err) => Err(CloudEngineError::failed(format!(
+                "could not determine the type of subnet {subnet_id}: {err}"
+            ))),
```
pierugo-dfinity: It feels like we shouldn't run `self.apply` if we see that error, do we? Since at that point, we're not even sure yet that we're really part of a cloud engine

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+        );
+    }
+
+    #[test]
+    fn any_other_outcome_restores_the_not_ready_budget() {
+        let mut manager = manager_for_test();
+        for _ in 0..MAX_CONSECUTIVE_NOT_READY - 1 {
+            manager.apply(Err(CloudEngineError::NotReady));
+        }
+
```
pierugo-dfinity, line 367: I would add a `assert_eq!(manager.discovery.remembered(), Some(operator()));` here just to pin the fact that we expect the operator not to be forgotten at this point.

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+        assert_eq!(
+            fetches(&manager, OUTCOME_NOT_READY),
+            MAX_CONSECUTIVE_NOT_READY as u64
+        );
+    }
+
+    #[test]
+    fn any_other_outcome_restores_the_not_ready_budget() {
+        let mut manager = manager_for_test();
+        for _ in 0..MAX_CONSECUTIVE_NOT_READY - 1 {
```
pierugo-dfinity:
```suggestion
        for _ in 1..MAX_CONSECUTIVE_NOT_READY {
```
I would keep it like above

rs/orchestrator/src/cloud_engine/operator.rs
```diff
@@ -0,0 +1,97 @@
+//! Client for the per-engine `engine-operator` canister.
+//!
+//! Every record is deliberately all-`opt` on the canister side so that the
+//! interface can grow without a breaking change; candid drops wire fields that
+//! are absent from these types, so added fields cannot break here either.
+//! Field *names* are part of the contract though: candid keys records by name
+//! hash, so a rename on either side silently decodes as absent.
```
pierugo-dfinity: I don't think this adds much

rs/orchestrator/src/cloud_engine/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+        // The replica listens on a wildcard address, so only its port is of use
+        // here: the operator canister is on this node's own subnet.
+        let replica_url = Url::parse(&format!("http://127.0.0.1:{}", replica_listen_addr.port()))
+            .inspect_err(|err| warn!(logger, "Cannot address the local replica: {err}"))
+            .ok()?;
+        let operator_agent = agent::node_signed(
+            registry.get_registry_client(),
+            crypto,
+            replica_url,
+            registry.get_latest_version(),
```
pierugo-dfinity, line 83: Seeing this `get_latest_version` here once in the constructor was suspicious to me as a reader, because this means the agent depends on the registry but is bound to the latest known version at startup and is never updated, i.e. might use outdated information.

Digging through the code, this is exceptionally fine because it is just about the NNS root key, which is never expected to change. To not confuse future readers, I would probably call `registry.get_latest_version()` directly inside `nns_root_key_der_from_registry`. We lose some flexibility but this record is exceptional in that it should never change, so I think it's fine.

rs/orchestrator/BUILD.bazel
```diff
@@ -111,11 +111,22 @@ rust_test(
...
+        # Keep sorted.
+        "CLOUD_ENGINE_MOCK_WASM": "$(rootpath //rs/rust_canisters/cloud_engine_mock:cloud_engine_mock_canister)",
+        "POCKET_IC_BIN": "$(rootpath //rs/pocket_ic_server:pocket-ic-server)",
+    },
     # Disable optimizations to speed up compilation otherwise it will take over 9 minutes on CI.
     # It does not have an effect on the test runtime.
     rustc_flags = ["-Copt-level=0"],
     deps = [
         # Keep sorted.
+        "//packages/pocket-ic",
```
pierugo-dfinity: This is quite a big dependency. Do you think it would be possible to declare a new Bazel target for the orchestrator's integration tests and include it only there? Similar to the registry replicator

rs/orchestrator/Cargo.toml
```diff
@@ -98,6 +98,7 @@ ic-test-utilities-logger = { path = "../test_utilities/logger" }
 ic-test-utilities-registry = { path = "../test_utilities/registry" }
 ic-test-utilities-time = { path = "../test_utilities/time" }
 ic-test-utilities-types = { path = "../test_utilities/types" }
+pocket-ic = { path = "../../packages/pocket-ic" }
 libc = { workspace = true }
```
pierugo-dfinity:
```suggestion
libc = { workspace = true }
pocket-ic = { path = "../../packages/pocket-ic" }
```
to keep sorted (ah well, I see this is already broken above lol)

rs/orchestrator/src/cloud_engine/integration_tests.rs
```diff
@@ -0,0 +1,562 @@
...
+        .discovery
+        .resolve(ROOT_SUBNET, VERSION)
+        .await
+        .expect_err("an engine with no operator on file cannot resolve one");
+
+    assert_matches!(err, CloudEngineError::Failed(msg) if msg.contains("does not know an operator"));
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn an_operator_that_stays_unauthorized_is_resolved_again() {
```
pierugo-dfinity: We could add a test almost identical to this one for `NotReady`

rs/orchestrator/src/cloud_engine/integration_tests.rs
```diff
@@ -0,0 +1,562 @@
...
+    fixture.start_all();
+
+    assert_eq!(fixture.fetches(OUTCOME_INCOMPLETE), 1);
+    assert_eq!(fixture.published_config(), None);
+    assert!(!fixture.processes.is_ic_gateway_running());
+    assert_eq!(fixture.gateway_env(), None);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn an_unauthorized_operator_is_not_ready() {
```
pierugo-dfinity: This could be merged into `an_operator_that_stays_unauthorized_is_resolved_again`

PR #11351, opened 2026-08-27
refactor(cketh): move the sweeper pipeline into AutomaticDeposits

gregorydemay:
## Why

- The sweeper pipeline sat on `State`, away from the sweep queue that feeds it, and its field was public where every sibling aggregate keeps its own private.

## What

- The pipeline moves into `AutomaticDeposits`, encapsulated the way `WithdrawalTransactions` encapsulates its own: the field is private and callers reach it through delegating methods, qualified with `sweep` so they cannot be confused with the deposit requests the same struct already holds.
- `AutomaticDeposits` gains an `is_equivalent_to`. Replay compared the pipeline with its own, which sorts a queue that `reschedule_request` reorders without recording an event; nesting it under the strict equality on `automatic_deposits` would have made replay fail after any reschedule.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -29,7 +29,7 @@ use std::cell::RefCell;
 use std::collections::{BTreeMap, BTreeSet, HashSet, btree_map};
 use std::fmt::{Display, Formatter};
 use strum_macros::EnumIter;
-use transactions::{SweepId, SweeperTransactionPipeline, WithdrawalTransactions};
+use transactions::{SweepId, WithdrawalTransactions};
 
 pub mod audit;
 pub mod automatic_deposits;
@@ -78,9 +78,6 @@ pub struct State {
     pub minted_events: BTreeMap<EventSource, MintedEvent>,
     pub invalid_events: BTreeMap<EventSource, InvalidEventReason>,
     pub withdrawal_transactions: WithdrawalTransactions,
-    /// The dedicated sweeper address' transaction pipeline: sweeps sent from the sweeper address on
-    /// its own nonce sequence, independent of the main-address withdrawal pipeline.
-    pub sweeper_transactions: SweeperTransactionPipeline,
     /// Monotonic counter minting the next [`SweepId`] for the sweeper pipeline.
     pub next_sweep_id: SweepId,
     pub skipped_blocks: BTreeMap<Address, BTreeSet<BlockNumber>>,
@@ -598,8 +595,8 @@ impl State {
         if let Some(nonce) = next_sweeper_transaction_nonce {
             let nonce = TransactionNonce::try_from(nonce)
                 .map_err(|e| InvalidStateError::InvalidTransactionNonce(format!("ERROR: {e}")))?;
-            self.sweeper_transactions
-                .update_next_transaction_nonce(nonce);
+            self.automatic_deposits
+                .update_next_sweeper_transaction_nonce(nonce);
         }
         if let Some(amount) = minimum_withdrawal_amount {
             let minimum_withdrawal_amount = Wei::try_from(amount).map_err(|e| {
@@ -711,7 +708,8 @@ impl State {
             other.ledger_suite_orchestrator_id
         );
         ensure_eq!(self.ckerc20_tokens, other.ckerc20_tokens);
-        ensure_eq!(self.automatic_deposits, other.automatic_deposits);
+        self.automatic_deposits
+            .is_equivalent_to(&other.automatic_deposits)?;
         ensure_eq!(
             self.sweeper_contract_address,
             other.sweeper_contract_address
@@ -719,8 +717,6 @@ impl State {
         ensure_eq!(self.sweeper_funding, other.sweeper_funding);
         ensure_eq!(self.next_sweep_id, other.next_sweep_id);
 
-        self.sweeper_transactions
-            .is_equivalent_to(&other.sweeper_transactions)?;
         self.withdrawal_transactions
             .is_equivalent_to(&other.withdrawal_transactions)
     }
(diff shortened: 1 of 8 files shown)


rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -66,9 +71,122 @@ pub struct AutomaticDeposits {
...
+    }
+
+    /// Equality as replay defines it: the sweeper pipeline reorders its queue without recording an
+    /// event, so it compares itself rather than being compared field by field.
+    pub fn is_equivalent_to(&self, other: &Self) -> Result<(), String> {
+        use ic_utils_ensure::ensure_eq;
+
+        ensure_eq!(self.watchlist, other.watchlist);
+        ensure_eq!(self.sweep, other.sweep);
+        ensure_eq!(self.attestations, other.attestations);
```
mbjorkqvist: To make this equivalence check more robust in case more fields are added to `AutomaticDeposits`, it would make sense to deconstruct to make this a compiler error:

```
    let Self { watchlist, sweep, attestations, sweeper_transactions } = self;
```

It would make sense to do this for the similar `WithdrawalTransactions` also. The issue already seems to occur in `TransactionPipeline::is_equivalent_to`, where the `processed_requests` field is not compared.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -66,9 +71,122 @@ pub struct AutomaticDeposits {
...
+        id: SweepId,
+        receipt: &TransactionReceipt,
+    ) -> Finalized<SweepTransaction> {
+        self.sweeper_transactions
+            .record_finalized_transaction(id, receipt)
+    }
+
+    /// Equality as replay defines it: the sweeper pipeline reorders its queue without recording an
+    /// event, so it compares itself rather than being compared field by field.
+    pub fn is_equivalent_to(&self, other: &Self) -> Result<(), String> {
```
mbjorkqvist, line 180: Would it make sense to add a test for `is_equivalent_to` that fails if the calls to `is_equivalent_to` are changed back to `ensure_eq!`? E.g., rescheduling of a sweep request and asserting the two states still compare equivalent (similar to what already exists for the withdrawal flow)?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -66,9 +71,122 @@ pub struct AutomaticDeposits {
     /// entries naming a retired helper stay behind forever. [`Self::attestations_len`] is exported
     /// as a metric so that growth is visible before it needs bounding.
     attestations: BTreeMap<AttestationRequest, TransactionSignature>,
+    /// The dedicated sweeper address' transaction pipeline: sweeps sent from the sweeper address on
+    /// its own nonce sequence, independent of the main-address withdrawal pipeline.
+    sweeper_transactions: SweeperTransactionPipeline,
```
mbjorkqvist, line 76: Would it make sense to keep `pub sweeper_transactions: SweeperTransactionsPipeline` here, to avoid having to add all the new accessor methods on `AutomaticDeposits`?

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -78,9 +78,6 @@ pub struct State {
     pub minted_events: BTreeMap<EventSource, MintedEvent>,
     pub invalid_events: BTreeMap<EventSource, InvalidEventReason>,
     pub withdrawal_transactions: WithdrawalTransactions,
-    /// The dedicated sweeper address' transaction pipeline: sweeps sent from the sweeper address on
-    /// its own nonce sequence, independent of the main-address withdrawal pipeline.
-    pub sweeper_transactions: SweeperTransactionPipeline,
```
mbjorkqvist, line 83: Would it make sense to move the `next_sweep_id` field below into `AutomaticDeposits` also?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Both parts done in 46fcae878a.

Destructured `Self` in all three equivalence checks — `AutomaticDeposits`, `TransactionPipeline`, and `WithdrawalTransactions` — so a newly added field is now a compile error rather than a silent gap.

You were right about `processed_requests`: it was genuinely missing from `TransactionPipeline::is_equivalent_to`. I checked whether there was a reordering reason to skip it like the one for `pending_requests`, and there is not — it is written only at transaction-creation time and `reschedule_request` moves entries out of `pending_requests` without touching it, so it is fully event-determined. It is now compared. The integration tests that exercise `check_audit_log` still pass, so replay was in fact producing a matching map; the check just was not verifying it.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Skipping this one for now. `is_equivalent_to` has a single call site — `check_audit_log`, a query gated behind `#[cfg(feature = "debug_checks")]` — so it is a test oracle rather than production logic, and a test for it would be a test of a test helper.

The destructuring in 46fcae878a covers the failure mode that actually worries me here (a new field silently dropping out of the comparison); the `ensure_eq!`-vs-reorder property is left uncovered deliberately.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Keeping the field private. This mirrors `WithdrawalTransactions`, which also wraps its `pipeline` privately and exposes it through delegating methods — making `sweeper_transactions` public would leave the two pipelines encapsulated differently for no reason other than the accessor count.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Leaving `next_sweep_id` on `State`, for the same encapsulation reason as the sibling thread: `AutomaticDeposits` keeps its pipeline private the way `WithdrawalTransactions` does, so moving the counter in would mean two more delegating accessors rather than a plain field, and the withdrawal side keeps its equivalent bookkeeping outside the wrapper too.

PR #11354, opened 2026-08-27
chore: Great simplification of SEV key exchange

frankdavid:
We used to keep two keyslots (and two metadata tokens) in the Store's LUKS header. The previous GuestOS's key and the current one. The upgrade VM stored the previous key in a file on the Var partition, and only after it rebooted as the default VM did it unlock the Store with that key and rotate in its own keyslot. 

With detached headers, each boot slot carries its own header copy on its Var partition, so we can simplify the passphrase rotation. Each header now holds exactly one keyslot and one metadata token (valid for the corresponding GuestOS), and the previous-key file is gone.

In the new flow, the old GuestOS hands its derived key and header to the upgrade VM; the upgrade VM writes the header to its own Var partition and switches the key to its own derived key right away. This wasn't possible with attached headers because the upgrade VM could not write the shared Store partition.

Benefits:
- simpler key rotation
- only a single key/token
- no need to temporarily store the key on var


rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -38,12 +38,65 @@ const NNS_PUBLIC_KEY_PATH: &str = "/run/config/nns_public_key.pem";
 
 type ServiceClientType = DiskEncryptionKeyExchangeServiceClient<Channel>;
 
+/// Wrapper around encrypted disk operations to allow testing guest upgrades without setting up
+/// block devices which requires root.
+#[mockall::automock]
```
frankdavid, line 45: Just moved this from the guest_disk crate since it's only useful for testing the upgrade mechanism.

rs/ic_os/os_tools/guest_disk/src/sev.rs
```diff
@@ -198,68 +94,65 @@ impl DiskEncryption for SevDiskEncryption {
     }
 }
 
-/// Check whether we can open the store partition with either the previous key or the SEV derived
-/// key.
-pub fn can_open_store(
+/// Reads the launch measurement and TCB version from the SEV firmware's attestation
+/// report, for storage in the LUKS2 keyslot metadata token.
+fn get_sev_metadata_for_luks(sev_firmware: &mut dyn SevGuestFirmware) -> Result<SevMetadata> {
```
frankdavid, line 99: This used to be method, was moved to be a free-standing function.

rs/ic_os/guest_upgrade/client/src/lib.rs
```diff
@@ -202,20 +252,20 @@ impl DiskEncryptionKeyExchangeClientAgent {
...
+    /// default VM we become after the reboot can re-derive.
+    async fn adopt_store_artifacts(
+        &mut self,
+        old_key: Vec<u8>,
         luks_header: Vec<u8>,
     ) -> Result<()> {
-        let disk_encryption_key =
-            String::from_utf8(disk_encryption_key).context("Key is not valid UTF-8")?;
-
         tokio::fs::write(&self.store_luks_header_path, &luks_header)
```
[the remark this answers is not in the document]
frankdavid: Fixed but fsync is not needed, if the machine crashes, it's as if the key exchange hadn't happened.

rs/ic_os/guest_upgrade/tests/src/lib.rs
```diff
@@ -166,14 +164,26 @@ impl DiskEncryptionKeyExchangeTestFixture {
...
-        let previous_key = NamedTempFile::new().unwrap();
-        let server_store = NamedTempFile::new().unwrap();
+        let store_device = NamedTempFile::new().unwrap();
         let server_store_luks_header = NamedTempFile::new().unwrap();
         let client_store_luks_header = NamedTempFile::new().unwrap();
         std::fs::write(server_store_luks_header.path(), sample_luks_header_bytes()).unwrap();
         // Increment port for each test case so tests can run in parallel
         let server_port = FREE_PORT.fetch_add(1, Ordering::Relaxed);
 
+        let can_open_disk = config.can_open_disk;
```
Bownairo: What do you think about handling `rekey` the same way? Something like `config.should_rekey_explode`. I think that saves the `Arc` and `mut` stuff, and then all of this would be consistent and handled in the same place.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
```diff
@@ -530,8 +518,11 @@ pub fn add_sev_metadata(
     keyslot: u32,
     sev_metadata: SevMetadata,
 ) -> Result<()> {
-    let assigned_token_ids = token_ids_assigned_to_keyslot(crypt_device, keyslot)?;
-    remove_token_ids(crypt_device, assigned_token_ids)?;
+    // TODO: Legacy headers may carry more than one IC key metadata token. Once all nodes
+    // have been updated (i.e., the num_tokens metric is 1 everywhere), this removal can
+    // be deleted so that only one token is written.
```
Bownairo, line 520: I don't follow. Even once we're down to a single key, we still need to clear out the old metadata when we rekey and write a new one, don't we?

frankdavid: The LUKS api allows you to overwrite a keyslot or a token in-place. So once we're down to a single keyslot+metadata in slot 0, we can just overwrite that in-place and get rid of the entire loop. Knowing that there is a single slot at index 0 is a great invariant.

rs/ic_os/guest_upgrade/tests/src/lib.rs
frankdavid: Actually, I'd prefer going in the other direction. I find letting tests create their expectations provides a higher flexibility. To reduce code duplication, we can have helper functions instead of have a field. For one-off expectations (used by only 1 or 2 test cases), we shouldn't pollute the fixture. What do you think?

Bownairo: Sounds good to me! I think that makes it easier to understand the test on a read-through, too. Consistency was mostly what I had in mind.

rs/ic_os/os_tools/guest_disk/src/crypt.rs
Bownairo: Gotcha, so we'll just overwrite the token (the clear out and write I was thinking), and drop all of this.

This already swaps the slot in place - the token handling is just different enough that we should wait to move to that API?

rs/ic_os/guest_upgrade/tests/src/lib.rs
```diff
@@ -583,14 +578,14 @@ async fn test_server_different_chip_id() {
...
 async fn test_can_open_disk() {
     let config = TestConfig {
-        can_open_disk: true,
         // If the client can open the disk, the client should not care about the server's
         // measurement since it won't even call the server.
         server_measurement: UNREGISTERED_MEASUREMENT,
         ..Default::default()
     };
+    let mut fixture = DiskEncryptionKeyExchangeTestFixture::new(config);
+    fixture.expect_can_open(true);
```
Bownairo, line 587: nit: I guess `fixture.expect_can_open(false);` could be part of the default fixture, and then you call `fixture.expect_can_open(true);` otherwise, as long as the helper clears existing expectations before writing more?

Saves a few lines, but I'm not sure if it's worth making things more complex.

Bownairo (approved):

rs/ic_os/os_tools/guest_disk/src/crypt.rs
frankdavid: Rewrote the token management to call remove_all_tokens and then explicitly set the first token slot, now it mirrors keyslots.

rs/ic_os/guest_upgrade/tests/src/lib.rs
frankdavid: If we add `expect_can_open(false)` by default, every test must call can_open otherwise the expectation fails.

PR #11355, opened 2026-08-27
fix: Only complete auto progress after the certified time has been set

mraszyk:
Enabling auto progress on a PocketIC instance spawned the progress thread and returned immediately, i.e., before the thread executed its initial `SetCertifiedTime` operation. Hence, a client could observe a stale certified time right after enabling auto progress (also when creating an instance with auto progress enabled or making an instance live).

Now the progress thread signals via a oneshot channel once the certified time has been set for the first time and `ApiState::auto_progress` awaits that signal before returning. The instance locks are dropped before awaiting since the progress thread needs them to execute the operation.

### DIFF
--- a/packages/pocket-ic/src/lib.rs
+++ b/packages/pocket-ic/src/lib.rs
@@ -491,6 +491,8 @@ impl PocketIcBuilder {
     /// Configures the new instance to make progress automatically,
     /// i.e., periodically update the time of the IC instance
     /// to the real time and execute rounds on the subnets.
+    /// Building the instance only returns after the certified time
+    /// of the IC instance has been updated for the first time.
     pub fn with_auto_progress(mut self) -> Self {
         let config = AutoProgressConfig {
             artificial_delay_ms: None,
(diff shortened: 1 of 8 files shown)


rs/pocket_ic_server/src/state_api/state.rs
eichhorl, line 999: If I understand correctly, this operation may fail if `time` is in the past. Is it still fine to send the signal unconditionally?

packages/pocket-ic/src/lib.rs
```diff
@@ -805,6 +805,8 @@ impl PocketIc {
     /// Configures the IC to make progress automatically,
     /// i.e., periodically update the time of the IC
     /// to the real time and execute rounds on the subnets.
+    /// Only returns after the certified time of the IC
+    /// has been updated for the first time.
```
eichhorl, line 811: Does the same apply to `make_live`?

eichhorl (approved):

rs/pocket_ic_server/src/state_api/state.rs
mraszyk: Yes, at least it is not a regression since before the operation would also silently fail and we'd return immediately.

packages/pocket-ic/src/lib.rs
mraszyk: [9e821c3](https://github.com/dfinity/ic/pull/11355/commits/9e821c3aca44745dd8578dc82c1137f7bd99cbea)

PR #11357, opened 2026-08-27
feat: [DSM-148] Track unhealthy nodes in XNet payload builder

alin-at-dfinity:
Both for stream polling (where we otherwise pick an unhealthy node with the same probability as a healthy one); and, in the future, for adverts (where we want to make it more likely for healthy nodes to learn our stream header); maintain a set of recently (10s) unhealthy nodes to be avoided.

Hook it into `XNetClientImpl` (to update the set) and into `ProximityMap` (to avoid unhealthy nodes when picking).

### DIFF
--- a/rs/xnet/payload_builder/src/lib.rs
+++ b/rs/xnet/payload_builder/src/lib.rs
@@ -10,6 +10,8 @@ mod tests;
 #[cfg(test)]
 mod xnet_client_tests;
 
+pub use proximity::{GenRangeFn, ProximityMap, UNHEALTHY_NODE_TTL, UnhealthyNodes};
+
 use crate::certified_slice_pool::{
     CertifiedSliceError, CertifiedSlicePool, CertifiedSliceResult, certified_slice_count_bytes,
 };
@@ -47,7 +49,6 @@ use ic_types::{Height, NodeId, NumBytes, RegistryVersion, SubnetId};
 use ic_xnet_hyper::TlsConnector;
 use ic_xnet_uri::XNetAuthority;
 use prometheus::{Histogram, HistogramVec, IntCounter, IntCounterVec, IntGauge};
-pub use proximity::{GenRangeFn, ProximityMap};
 use rand::{Rng, rngs::StdRng, thread_rng};
 use std::collections::{BTreeMap, VecDeque};
 use std::net::SocketAddr;
@@ -336,9 +337,13 @@ impl XNetPayloadBuilderImpl {
         metrics_registry: &MetricsRegistry,
         log: ReplicaLogger,
     ) -> XNetPayloadBuilderImpl {
+        // Shared by the node selection of every component talking to other
+        // subnets: `ProximityMap` below, the advert task once it exists.
+        let unhealthy_nodes = Arc::new(UnhealthyNodes::new(UNHEALTHY_NODE_TTL, metrics_registry));
         let proximity_map = Arc::new(ProximityMap::new(
             node_id,
             registry.clone(),
+            unhealthy_nodes.clone(),
             metrics_registry,
             log.clone(),
         ));
@@ -346,6 +351,7 @@ impl XNetPayloadBuilderImpl {
             metrics_registry,
             tls_handshake,
             proximity_map.clone(),
+            unhealthy_nodes,
         ));
 
         let deterministic_rng_for_testing = Arc::new(None);
@@ -1718,6 +1724,9 @@ struct XNetClientImpl {
 
     /// Proximity map to update after every query with the time-to-first-byte.
     proximity_map: Arc<ProximityMap>,
+
+    /// Unhealthy node set to update after every query with its outcome.
+    unhealthy_nodes: Arc<UnhealthyNodes>,
 }
 
 impl XNetClientImpl {
@@ -1727,6 +1736,7 @@ impl XNetClientImpl {
         metrics_registry: &MetricsRegistry,
         tls: Arc<dyn TlsConfig>,
         proximity_map: Arc<ProximityMap>,
+        unhealthy_nodes: Arc<UnhealthyNodes>,
     ) -> XNetClientImpl {
         #[cfg(not(test))]
         let https = TlsConnector::new(tls);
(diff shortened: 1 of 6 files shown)


rs/xnet/payload_builder/src/lib.rs
```diff
@@ -1858,6 +1885,32 @@ pub enum XNetClientError {
 }
 
 impl XNetClientError {
+    /// Returns `true` if this error means that the node failed to serve the
+    /// request: we could not connect to it, it timed out, or it responded with
+    /// a 5xx status code or something we could not make sense of. Whereas a 204
+    /// (`NoContent`) or 4xx status code point to a (likely) healthy node.
+    fn is_node_failure(&self) -> bool {
```
eichhorl, line 1892: How does a node that's behind respond?

rs/xnet/payload_builder/src/proximity.rs
```diff
@@ -26,8 +26,14 @@ pub type GenRangeFn = Box<dyn Fn(u64, u64) -> u64 + Sync + Send>;
 
 const NANOS_PER_SEC: u64 = 1_000_000_000;
 
+/// How long a node is considered unhealthy after a failed request. It may be
+/// tried again after this long, so a node that stays down costs one failed
+/// request per TTL.
```
eichhorl, line 31: I guess there could be multiple requests to the same node "in flight"?

rs/xnet/payload_builder/src/proximity.rs
eichhorl: Could also mention that unhealthy nodes are excluded here

rs/xnet/payload_builder/src/proximity.rs
```diff
@@ -234,6 +249,81 @@ impl ProximityMap {
...
+}
+
+impl UnhealthyNodes {
+    pub fn new(ttl: Duration, metrics_registry: &MetricsRegistry) -> Self {
+        Self {
+            nodes: Default::default(),
+            ttl,
+            metric_unhealthy_nodes: metrics_registry.int_gauge(
+                METRIC_UNHEALTHY_NODES,
+                "Number of unhealthy nodes, i.e. whose last XNet request failed.",
```
eichhorl: I suppose the a node could still be unhealthy, but not show up because we didn't query it in a while?
```suggestion
                "Number of unhealthy nodes within the unhealthy node TTL, i.e. nodes whose last XNet request failed.",
```

rs/xnet/payload_builder/src/xnet_client_tests.rs
```diff
@@ -46,10 +49,18 @@ where
 
 fn make_xnet_client(metrics: &MetricsRegistry, log: ReplicaLogger) -> XNetClientImpl {
```
eichhorl, line 50: We could add tests here showing that a successful query clears earlier failures, and also the fact that 5xx means failure and 4xx means success

rs/xnet/payload_builder/src/proximity.rs
```diff
@@ -234,6 +249,81 @@ impl ProximityMap {
...
+        self.prune(&mut nodes);
+    }
+
+    /// Drops the unhealthy nodes from `nodes`, as long as at least `minimum`
+    /// remain; else leaves `nodes` as it is. So that a subnet-wide endpoint
+    /// problem or a local network fault cannot leave the caller with too few
+    /// nodes to choose from -- or none at all.
+    pub fn filter_at_least(&self, minimum: usize, nodes: &mut Vec<NodeId>) {
+        let mut unhealthy = self.nodes.lock().unwrap();
+        self.prune(&mut unhealthy);
```
eichhorl, line 307: Do we have a test for this `prune`?

eichhorl (approved):

rs/xnet/payload_builder/src/lib.rs
alin-at-dfinity: If it's too far behind to produce a slice that fits our expectations (i.e. if there is no overlap between the messages it has and the ones we are expecting to see next), with an `HTTP 416 Range Not Satisfiable` (which is intended for byte range requests, but I found the similarity cute and repurposed it). However, it also responds with the same 416 if it is ahead of us (i.e. if we are the replica that's behind our subnet). Logs show instances of both.

So we could optimize further by marking remote replicas that are behind as unhealthy, but we'd first need to differentiate between ahead and behind (either by using different status codes or by standardizing the response payload).

rs/xnet/payload_builder/src/proximity.rs
alin-at-dfinity: Right now, there aren't. We fire off a batch of requests to one replica on every subnet; and when notified, if the previous batch hasn't completed, we simply skip this batch.

Once we switch to advert-based, it depends. E.g. I haven't yet fully decided on what the queue for the pull task should deal with race conditions: if I'm pulling based on a header from certified height N and get notified with the (different) header from certified height N+1, do I overwrite it in the queue? should I have removed the former when I started pulling? should I start pulling immediately for the new one once I'm done with the old? There are a lot of trade-offs to think through.

rs/xnet/payload_builder/src/proximity.rs
alin-at-dfinity: `unhealthy_nodes_expire()` explicitly tests that with a TTL of zero, an `observe_failure()` followed by a `filter_at_least()` will classify all nodes as healthy.

rs/xnet/payload_builder/src/proximity.rs
alin-at-dfinity: Done, thanks for the tip.

rs/xnet/payload_builder/src/proximity.rs
alin-at-dfinity: Reworded, thanks.

rs/xnet/payload_builder/src/xnet_client_tests.rs
alin-at-dfinity: We already had a test for a `HTTP 500` response (switched it to `HTTP 503`, because that's what the endpoint actually produces). Added a test for `HTTP 416 Range Not Satisfiable`; and one for `HTTP 503` followed by `HTTP 200`.

PR #11358, opened 2026-08-27
fix: Use the same registry version in `CatchUpPackageParam` and verification 

eichhorl:
The orchestrator only fetches a CUP if it has a higher height and registry version than the local CUP. Here, the relevant registry version is the one of the DKG summary.

However, when validating the CUPs signature, we used the version of the validation context, which in principal could be smaller than the one of the DKG summary. This could lead to the CUP being verified with an old public key from the registry.

With this PR, we also use the registry version of the DKG summary for signature verification. This is fine, since under normal operation, both versions should always point to the same public key:

1. The public key can only change via recovery and subnet splitting, during which both versions are equal. 
2. A recovery CUP may only be adopted if the subnet is stalled, it is therefore not possible to create a summary whose validation context points to the new key, while the DKG version still points to the old key.
3. During a subnet split, the version of the validation context is intentionally frozen to one version below the one introducing the new key. The "scheduled split" summary (which bumps the context's version to the new key) is never part of a CUP. Instead, a new summary is derived for which both versions point to the new key (like in a recovery).

### DIFF
--- a/rs/orchestrator/src/catch_up_package_provider.rs
+++ b/rs/orchestrator/src/catch_up_package_provider.rs
@@ -715,8 +715,6 @@ pub(crate) mod tests {
         let context_registry_version = RegistryVersion::from(7);
         let cup_proto =
             cup_with_distinct_registry_versions(summary_registry_version, context_registry_version);
-        // Guards the premise of this test: the CUP must actually distinguish the two fields.
-        assert_ne!(summary_registry_version, context_registry_version);
 
         let server_addr =
             start_server(TestService::SendCup(Arc::new(cup_proto.encode_to_vec()))).await;
--- a/rs/orchestrator/src/catch_up_package_provider.rs
+++ b/rs/orchestrator/src/catch_up_package_provider.rs
@@ -293,7 +293,7 @@ impl CatchUpPackageProvider {
                 &CombinedThresholdSigOf::new(CombinedThresholdSig(protobuf.signature.clone())),
                 &CatchUpContentProtobufBytes::from(&protobuf),
                 subnet_id,
-                cup.content.block.get_value().context.registry_version,
+                cup.content.registry_version(),
             )
             .map_err(|e| format!("Failed to verify CUP signature at: {uri:?} with: {e:?}"))?;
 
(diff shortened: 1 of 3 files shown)


rs/orchestrator/src/catch_up_package_provider.rs
```diff
@@ -618,6 +641,118 @@ pub(crate) mod tests {
...
+        let (_, cup) = cup_provider
+            .fetch_and_verify_catch_up_package(&node_id, &node_record, None, SUBNET_0)
+            .await
+            .expect("failed to fetch and verify the CUP")
+            .expect("expected a CUP to be served");
+
+        assert_eq!(
+            *crypto.verified_at.lock().unwrap(),
+            Some(summary_registry_version)
+        );
```
pierugo-dfinity: Maybe also add an `assert_ne!(verified_at, context_registry_version)`

pierugo-dfinity (approved):
> the validation context, which in principal could be smaller than the one of the DKG summary

You mean the opposite? So it's actually the other way around, we could technically validate the CUP with a "future" public key of the subnet.

The fix is anyways still correct: we should use the same registry version for validation than the one we used to sign it. The latter is the one from the high threshold transcript of the summary, and the payload builder ensures it is equal to the summary's registry version, which is what you are using here. Though it does not look like this implicit invariant is actually enforced by any tests today.

eichhorl:
> > the validation context, which in principal could be smaller than the one of the DKG summary
> 
> You mean the opposite? So it's actually the other way around, we could technically validate the CUP with a "future" public key of the subnet.

No, I mean that assuming the CUP is completely forged, then that forgery could also contain a validation context with a registry version that is smaller than the one of the DKG summary

Under normal operation I agree that the risk is the one you describe (although it should never happen with the argument that recovery/splitting set both versions to the same value)

pierugo-dfinity:
Ah yes, I understand your argument

rs/orchestrator/src/catch_up_package_provider.rs
eichhorl: Done, also added a test making sure that both versions are the same for registry cups

PR #11359, opened 2026-08-27
feat(cketh): sign the attestations a batched sweep needs

gregorydemay:
## Why

- The sweep queue folds deposits into a batch per token, but nothing produced the attestation each deposit needs before it can be swept: the delegate recovers that signature against the deposit address it moves funds from, so an unattested deposit cannot be swept at all.

## What

- Each batch resolves the attestation request its deposits need and signs the ones the minter has not signed before.
- An attestation binds an account to one chain and one deposit helper — never to a token — so an account depositing several tokens is attested once and the signature is reused across its batches.
- Signatures are recorded through the event log, so replaying after an upgrade keeps them rather than paying for another threshold-ECDSA signature.
- A deposit whose attestation cannot be signed is reported and left for the next tick.

### DIFF
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -56,6 +56,14 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
             return;
         }
     };
+
+    let Some(_sweeper_contract) = read_state(|s| s.sweeper_contract_address) else {
+        log!(
+            DEBUG,
+            "[create_pending_sweeper_requests]: SKIPPING: no sweeper contract address is configured"
+        );
+        return;
+    };
 }
 
 pub async fn process_sweeper_transactions<R: CanisterRuntime>(runtime: R) {
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -68,6 +68,15 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
         return;
     };
 
+    let batch_per_token =
+        read_state(|s| s.automatic_deposits.requests_batch(MAX_DEPOSITS_PER_SWEEP));
+    if batch_per_token.is_empty() {
+        return;
+    }
+
+    // Priced only once there is something to sweep: the estimate expires on this task's own
+    // interval, so refreshing it before the queue is known to be non-empty spends an
+    // `eth_feeHistory` outcall every tick that an idle minter has no use for.
     let Some(_gas_fee_estimate) = lazy_refresh_gas_fee_estimate(runtime).await else {
         log!(
             INFO,
@@ -75,12 +84,6 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
         );
         return;
     };
-
-    let batch_per_token =
-        read_state(|s| s.automatic_deposits.requests_batch(MAX_DEPOSITS_PER_SWEEP));
-    if batch_per_token.is_empty() {
-        return;
-    }
 
     for (_token, targets) in batch_per_token {
         let Some(attestation_requests) = read_state(|s| s.attestation_requests(&targets)) else {
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -74,9 +74,6 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
         return;
     }
 
-    // Priced only once there is something to sweep: the estimate expires on this task's own
-    // interval, so refreshing it before the queue is known to be non-empty spends an
-    // `eth_feeHistory` outcall every tick that an idle minter has no use for.
     let Some(_gas_fee_estimate) = lazy_refresh_gas_fee_estimate(runtime).await else {
         log!(
             INFO,
--- a/rs/ethereum/cketh/minter/src/sweep/mod.rs
+++ b/rs/ethereum/cketh/minter/src/sweep/mod.rs
@@ -42,10 +42,11 @@ pub(crate) const SWEEP_TRANSACTION_GAS_LIMIT: GasAmount = GasAmount::new(100_000
 const SWEEP_REQUESTS_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SIGN_BATCH_SIZE: usize = 5;
 const SWEEP_TRANSACTIONS_TO_SEND_BATCH_SIZE: usize = 5;
+const MAX_DEPOSITS_PER_SWEEP: usize = 10;
 
 /// Turns the deposits the balance scan queued into the sweep requests that
 /// [`process_sweeper_transactions`] prices, signs, sends and finalizes.
-pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
+pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
     let _guard = match TimerGuard::new(TaskType::SweeperEnqueue) {
         Ok(guard) => guard,
         Err(e) => {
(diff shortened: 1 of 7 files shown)


rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -56,6 +59,76 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
...
+            "[create_pending_sweeper_requests]: SKIPPING: failed retrieving gas fee estimate"
+        );
+        return;
+    };
+
+    let batch_per_token =
+        read_state(|s| s.automatic_deposits.requests_batch(MAX_DEPOSITS_PER_SWEEP));
+    if batch_per_token.is_empty() {
+        return;
+    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in 44d30c4f0e — the queue check now runs before the estimate is refreshed.

The cadence you point at is exact, not approximate: `SWEEP_ENQUEUE_INTERVAL` is 60s and `MAX_AGE_NS` is 60s, and the freshness test is `now < last + 60s`, so a tick landing on the boundary always missed the cache. An idle minter really was paying for an `eth_feeHistory` outcall every minute.

I left the estimate in this function rather than deferring it to sweep-request construction: that consumer arrives in a later layer of the stack, and keeping the guard here is what pins the ordering now. `should_not_price_an_empty_sweep_queue` covers it — the mock carries no `expect_time`, so refreshing the estimate (which reads the clock) fails the test. I verified it fails when the old order is restored.

On the second half — a fee failure still stops attestation signing when the queue is non-empty. I am leaving that: there is no value in attesting for a sweep that cannot be priced this tick, and that is exactly the case where the estimate has to be fetched anyway. Only the empty-queue waste was worth removing.

rs/ethereum/cketh/minter/src/sweep/tests.rs
```diff
@@ -1,25 +1,169 @@
...
 
-fn state_with_a_queued_deposit() -> State {
+#[tokio::test]
+async fn should_be_no_op_when_no_deposit_helper_contract() {
     let mut state = initial_state();
-    apply_state_transition(
-        &mut state,
-        &EventType::AutomaticDepositReceived(automatic_deposit()),
+    state.log_scrapings = LogScrapings::new(BlockNumber::ONE);
+    init_state(state);
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct, and fixed in 44d30c4f0e.

I confirmed it rather than reading it off: putting a `panic!` in the deposit-helper branch left the test green, so it never reached it. The cause is what you describe — `initial_state()` builds from `valid_init_arg()`, which sets `ethereum_sweeper_contract_address: None`, so the run returned at the sweeper guard and the test was a duplicate of `should_be_no_op_when_no_sweeper_contract` under a different name.

It now starts from `state_ready_to_sign`, which configures the sweeper contract, a fresh cached estimate and a queued deposit, then clears the helper. The same `panic!` probe now fires, so the guard is genuinely exercised, and the mock carries no signing expectation so getting past the helper would fail the test.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -68,6 +68,15 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
...
 
+    let batch_per_token =
+        read_state(|s| s.automatic_deposits.requests_batch(MAX_DEPOSITS_PER_SWEEP));
+    if batch_per_token.is_empty() {
+        return;
+    }
+
+    // Priced only once there is something to sweep: the estimate expires on this task's own
+    // interval, so refreshing it before the queue is known to be non-empty spends an
+    // `eth_feeHistory` outcall every tick that an idle minter has no use for.
```
gregorydemay: remove comment

gregorydemay: 🤖 Removed in e9d0d56f0c, along with the three others I added in the same commit (two in `sweep/tests.rs`, one more here). Added as a follow-up commit rather than an amend, since this PR is already open for review.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -56,6 +59,76 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
...
+    for (_token, targets) in batch_per_token {
+        let Some(attestation_requests) = read_state(|s| s.attestation_requests(&targets)) else {
+            log!(
+                DEBUG,
+                "[create_pending_sweeper_requests]: SKIPPING: no deposit helper with subaccount is configured"
+            );
+            return;
+        };
+        sign_attestations_batch(attestation_requests, runtime).await;
+    }
```
mbjorkqvist, line 94: nit: Would it make sense to add a comment here, or on `sign_attestations_batch`, saying that for the dedupe for the signatures to work, this needs to be called sequentially (and not using `join_all`)? Or somehow programmatically enforce it, by keeping track of which signatures requests are in flight?

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -56,6 +59,76 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
...
+    let Some(_gas_fee_estimate) = lazy_refresh_gas_fee_estimate(runtime).await else {
+        log!(
+            INFO,
+            "[create_pending_sweeper_requests]: SKIPPING: failed retrieving gas fee estimate"
+        );
+        return;
+    };
+
+    for (_token, targets) in batch_per_token {
+        let Some(attestation_requests) = read_state(|s| s.attestation_requests(&targets)) else {
```
mbjorkqvist, line 86: `read_state(|s| s.attestation_requests(&targets))` only returns `None` if the deposit helper is unconfigured, and this is the same for every token. Rather than checking on every iteration, could this check be performed before the start of the `for` loop, and be done using this?

```
let Some(deposit_helper) = read_state(|s| s.deposit_helper_address()) else {
    log!(DEBUG, "[create_pending_sweeper_requests]: SKIPPING: no deposit helper with subaccount is configured");
    return;
};
```

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -56,6 +59,76 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(_runtime: &R) {
...
+                        runtime,
+                    )
+                });
+            }
+            Err(e) => errors.push((request, e)),
+        }
+    }
+
+    if !errors.is_empty() {
+        log!(INFO, "Errors encountered during signing: {errors:?}");
```
mbjorkqvist: nit: To be more consistent with other added log messages (adding the prefix), and not log an unbounded vector, how about the following:
```suggestion
        log!(INFO, "[create_pending_sweeper_requests]: failed to sign {} attestations: {errors:?}", errors.len());
```

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/sweep/mod.rs
gregorydemay: 🤖 Correct that the check is loop-invariant, but I'd rather keep it where it is. `attestation_request(s)` are deliberately the only constructors of an `AttestationRequest` outside its module (see the doc on `State::attestation_request`), so a caller can't attest under a chain or deposit helper the minter doesn't run against; a `State::deposit_helper_address()` accessor would hand `sweep` exactly that. The cost of keeping the `Option` at the call site is one redundant state read per token, and the minter supports on the order of tens of tokens, so it doesn't matter in practice.

rs/ethereum/cketh/minter/src/sweep/mod.rs
gregorydemay: 🤖 This is already enforced by a test: `should_sign_one_attestation_for_every_token_of_an_account` (`sweep/tests.rs`) queues the same account under two tokens and asserts `sign_with_ecdsa` is called exactly once, so switching the loop to `join_all` fails the test rather than silently paying for the same signature twice.

rs/ethereum/cketh/minter/src/sweep/mod.rs
gregorydemay: 🤖 Taken in f33a0c67e5: prefixed the log to match the rest of the enqueue. Kept `{errors:?}` — the vector is bounded by `MAX_DEPOSITS_PER_SWEEP` since `requests_batch` truncates per token, and the failure reasons are the useful part.

PR #11360, opened 2026-08-27
feat(cketh): sign the authorizations a batched sweep needs

gregorydemay:
## Why

- The attestations the previous layer signs name the account a sweep credits, but nothing yet lets the delegate move the funds: a deposit address only executes the sweeper contract's code once it has signed an EIP-7702 authorization delegating to it.

## What

- Each deposit address in a batch signs the tuple delegating its code to the configured sweeper contract.
- The tuple carries **nonce 0** whatever nonce the address actually holds. A deposit address is at nonce 0 exactly while it has never been delegated, so the authorization either installs the delegation or is skipped — both correct, in any ordering of the sweeps carrying them. That is what lets a sweep authorize every address it touches without tracking which ones are already delegated, at the price of a skipped tuple's intrinsic gas.
- Signatures are recorded through the event log, keyed by what was signed, so a later sweep of the same address reuses one rather than paying for another threshold-ECDSA signature. Keying by the request rather than the account is what makes that safe: re-pointing the minter at another sweeper contract changes the tuple, so it misses the cache and is signed afresh instead of reusing one that delegates the old contract.
- Attestations and authorizations are signed independently. A deposit needs both before it can be swept, but neither signature depends on the other, so authorizing an address whose attestation failed costs nothing: the tuple is recorded and reused once the attestation lands.
- Recording the tuple is a shortcut: it saves the signature but not the gas, since an address that is already delegated still gets sent a tuple the chain skips. Recording that the address is delegated, rather than what it signed, would save both — left for a later layer.

## Candid compatibility

- Needs the `CI_OVERRIDE_DIDC_CHECK` label: `get_events` returns a variant, and `AuthorizedDepositAddress` adds a case to it. A variant with more cases is not a Candid subtype of one with fewer, so the check against the merge base rejects it, exactly as it did for the attestation event in #11319.
- The case is additive: no existing case changes shape, so every event emitted so far decodes as before. Only a client reading an event of the new kind needs to know it.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -18,6 +18,7 @@ use crate::state::eth_logs_scraping::{LogScrapingId, LogScrapings};
 use crate::state::sweeper_funding::{SweeperFundingAccounting, SweeperFundingConfig};
 use crate::state::transactions::{Erc20WithdrawalRequest, TransactionCallData, WithdrawalRequest};
 use crate::timed_sized_map::{Entry, Timestamp};
+use crate::tx::AuthorizationRequest;
 use crate::tx::GasFeeEstimate;
 use crate::tx::TransactionSignature;
 use candid::Principal;
@@ -291,6 +292,36 @@ impl State {
                         self.ethereum_network.chain_id(),
                         deposit_helper,
                         *account.as_ref(),
+                    )
+                })
+                .collect(),
+        )
+    }
+
+    /// What every deposit address in `accounts` authorizes to let the configured sweeper contract
+    /// sweep it: the tuple naming this minter's chain, that contract, and nonce zero. `None` while
+    /// no sweeper contract is configured.
+    ///
+    /// The nonce is always zero, whatever the address actually holds. A deposit address is at
+    /// nonce zero exactly while it has never been delegated — applying an authorization spends it
+    /// — so the tuple either installs the delegation or is skipped, and both are correct in any
+    /// order the sweeps carrying them land. That is what lets a sweep authorize every address it
+    /// touches without tracking which ones are already delegated, at the price of the intrinsic
+    /// gas a skipped tuple still costs.
+    pub fn authorization_requests<T: AsRef<Account>>(
+        &self,
+        accounts: &[T],
+    ) -> Option<Vec<AuthorizationRequest>> {
+        let delegate = self.sweeper_contract_address?;
+        Some(
+            accounts
+                .iter()
+                .map(|account| {
+                    AuthorizationRequest::new(
+                        *account.as_ref(),
+                        self.ethereum_network.chain_id(),
+                        delegate,
+                        TransactionNonce::ZERO,
                     )
                 })
                 .collect(),
--- a/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
@@ -77,6 +77,10 @@ pub struct AutomaticDeposits {
     /// signature only delegates the chain, the sweeper contract and the nonce its request names, so
     /// re-pointing the minter at another sweeper contract misses this map rather than reusing a
     /// tuple that delegates the old one.
+    ///
+    /// Nothing prunes this map: it grows with the number of accounts that have ever been swept, and
+    /// entries naming a retired helper stay behind forever. [`Self::authorizations_len`] is exported
+    /// as a metric so that growth is visible before it needs bounding.
     authorizations: BTreeMap<AuthorizationRequest, TransactionSignature>,
     /// The dedicated sweeper address' transaction pipeline: sweeps sent from the sweeper address on
     /// its own nonce sequence, independent of the main-address withdrawal pipeline.
@@ -460,6 +464,10 @@ impl AutomaticDeposits {
 
     pub fn attestations_len(&self) -> usize {
         self.attestations.len()
+    }
+
+    pub fn authorizations_len(&self) -> usize {
+        self.authorizations.len()
     }
 
     /// Where `request`'s deposit currently stands, or `None` if the pair is neither armed nor has
--- a/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
@@ -12,7 +12,9 @@ use crate::state::transactions::{
     ResubmitTransactionError, SweepId, SweepRequest, SweeperTransactionPipeline,
 };
 use crate::timed_sized_map::{Entry, InsertError, TimedSizedMap, Timestamp};
-use crate::tx::{Finalized, GasFeeEstimate, Signed, SweepTransaction, TransactionSignature};
+use crate::tx::{
+    AuthorizationRequest, Finalized, GasFeeEstimate, Signed, SweepTransaction, TransactionSignature,
+};
 use ic_ethereum_types::Address;
 use icrc_ledger_types::icrc1::account::Account;
 use std::collections::BTreeMap;
@@ -71,6 +73,11 @@ pub struct AutomaticDeposits {
     /// entries naming a retired helper stay behind forever. [`Self::attestations_len`] is exported
     /// as a metric so that growth is visible before it needs bounding.
     attestations: BTreeMap<AttestationRequest, TransactionSignature>,
+    /// Delegation authorizations the minter has signed, keyed by exactly what each one signed. A
+    /// signature only delegates the chain, the sweeper contract and the nonce its request names, so
+    /// re-pointing the minter at another sweeper contract misses this map rather than reusing a
+    /// tuple that delegates the old one.
+    authorizations: BTreeMap<AuthorizationRequest, TransactionSignature>,
     /// The dedicated sweeper address' transaction pipeline: sweeps sent from the sweeper address on
     /// its own nonce sequence, independent of the main-address withdrawal pipeline.
     sweeper_transactions: SweeperTransactionPipeline,
@@ -184,12 +191,14 @@ impl AutomaticDeposits {
             watchlist,
             sweep,
             attestations,
+            authorizations,
             sweeper_transactions,
         } = self;
 
         ensure_eq!(watchlist, &other.watchlist);
         ensure_eq!(sweep, &other.sweep);
         ensure_eq!(attestations, &other.attestations);
+        ensure_eq!(authorizations, &other.authorizations);
         sweeper_transactions.is_equivalent_to(&other.sweeper_transactions)
     }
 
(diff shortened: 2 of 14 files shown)


rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -131,6 +134,87 @@ async fn sign_attestations_batch<R: CanisterRuntime>(
...
+/// instead of tracking which ones are delegated.
+///
+/// The one thing this forecloses is **re-delegation**: an address delegated to an earlier sweeper
+/// contract holds nonce >= 1, so a nonce-0 tuple naming the new one is skipped and its code keeps
+/// pointing at the old contract.
+fn delegation_authorization(chain_id: u64, sweeper_contract: Address) -> Authorization {
+    Authorization {
+        chain_id,
+        delegate: sweeper_contract,
+        nonce: TransactionNonce::ZERO,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct, and thanks — this is a real gap, not just a theoretical one. Filed as [DEFI-2982](https://dfinity.atlassian.net/browse/DEFI-2982).

To be precise about the chain of events, since the failure is subtler than "the authorization is stale":

- A never-delegated deposit address has nonce 0, so the nonce-0 tuple applies and the address delegates to `S1`.
- Applying it increments that address' nonce to 1.
- The sweeper contract is re-pointed to `S2` via `UpgradeArg`.
- The minter *does* sign a fresh tuple for `S2` — keying the cache on `AuthorizationRequest` rather than on `Account` means the new delegate is a cache miss, so there is no stale reuse here.
- But the fresh tuple still names nonce 0, while the authority is now at 1, so EIP-7702 silently skips it on chain. The address stays delegated to `S1`.

So the caching is fine; the bug is that nonce 0 is only ever correct for the *first* delegation. Consequence is that only never-delegated addresses pick up `S2`, and sweeps against the already-delegated ones may start reverting once `S1` is retired. No funds at risk — a reverted sweep moves nothing — but those deposits are stuck.

Nonce 0 stays a sound shortcut for the first delegation, which is why I'd rather fix the replacement path than change the nonce scheme wholesale. The three candidates are in the ticket: sign against the current on-chain nonce, track which addresses are already delegated, or forbid replacement once delegation has started.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -91,6 +93,7 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
             return;
         };
         sign_attestations_batch(attestation_requests, runtime).await;
+        let _authorizations = sign_authorizations_batch(&targets, sweeper_contract, runtime).await;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Half of this is fixed, half is intentional for this PR.

**Fixed — the repeated signing.** The authorizations are no longer discarded: each one is recorded in stable memory via a new `AuthorizedDepositAddress` event, keyed by the `AuthorizationRequest` (account + chain id + delegate + nonce) with the `TransactionSignature` as the value, the same shape as `AttestedDepositAddress`. `sign_authorizations_batch` filters out requests already in that map before signing, so a second timer tick over the same queued deposits is a cache hit and costs no threshold-ECDSA signatures. Keying on the request rather than on the account is deliberate: if the delegate or nonce ever changes, that is a cache miss and the minter re-signs, which `should_resign_the_authorization_when_the_sweeper_contract_changes` covers.

**Intentional — nothing consumes them yet.** There is no `SweepRequest.authorizations` field to attach to at this layer; building the EIP-7702 transaction comes in a later PR in the stack. This one only signs and persists. So the deposits do stay queued, but the work is no longer repeated — the signatures accumulate in state until the sweep-building layer reads them.

On excluding targets whose signing failed: also deferred, for the same reason. Nothing selects targets for a transaction yet, so there is no selection to exclude from. Signing errors are logged and the request simply stays unrecorded, which means the next tick retries it.

The stale-nonce problem you raised in the other thread is tracked separately as [DEFI-2982](https://dfinity.atlassian.net/browse/DEFI-2982).

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -71,6 +73,11 @@ pub struct AutomaticDeposits {
     /// entries naming a retired helper stay behind forever. [`Self::attestations_len`] is exported
     /// as a metric so that growth is visible before it needs bounding.
     attestations: BTreeMap<AttestationRequest, TransactionSignature>,
+    /// Delegation authorizations the minter has signed, keyed by exactly what each one signed. A
+    /// signature only delegates the chain, the sweeper contract and the nonce its request names, so
+    /// re-pointing the minter at another sweeper contract misses this map rather than reusing a
+    /// tuple that delegates the old one.
```
mbjorkqvist, line 79: Similarly to `attestations` above, how about adding the following doc regarding the unbounded growth, and also a metric?
```suggestion
    /// tuple that delegates the old one.
    ///
    /// Nothing prunes this map: it grows with the number of accounts that have ever been swept, and
    /// entries naming a retired helper stay behind forever. [`Self::authorizations_len`] is exported
    /// as a metric so that growth is visible before it needs bounding.
```

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -297,6 +298,36 @@ impl State {
...
+        let delegate = self.sweeper_contract_address?;
+        Some(
+            accounts
+                .iter()
+                .map(|account| {
+                    AuthorizationRequest::new(
+                        *account.as_ref(),
+                        self.ethereum_network.chain_id(),
+                        delegate,
+                        TransactionNonce::ZERO,
```
mbjorkqvist, line 324: Changing the sweeper contract doesn't work, since the nonce on this authorization request stays zero, and so the delegation is skipped. This seems to be tracked in [DEFI-2982](https://dfinity.atlassian.net/browse/DEFI-2982), and the `should_resign_the_authorization_when_the_sweeper_contract_changes` test looks like it may verify that changing the sweeper contract works, but it only asserts the in-canister cache map. Would it make sense to add an anvil test that verifies that a delegated deposit address ignores a zero nonce re-delegation? And/or add a test in `sweep/tests.rs` (I'm not sure exactly how that would look like, or if it's possible)?

rs/ethereum/cketh/minter/src/sweep/tests.rs
```diff
@@ -96,6 +94,129 @@ async fn should_sign_one_attestation_for_every_token_of_an_account() {
...
+}
+
+fn authorization_request(delegate: Address) -> AuthorizationRequest {
+    AuthorizationRequest::new(
+        account(),
+        initial_state().ethereum_network.chain_id(),
+        delegate,
+        TransactionNonce::ZERO,
+    )
+}
```
mbjorkqvist, line 221: [`Authorization::sign`](https://github.com/dfinity/ic/blob/2e9c52e3fc4cb37a7cf519708b98f197d514b947/rs/ethereum/cketh/minter/src/tx/eip_7702.rs#L85-L106) doesn't seem to be called from anywhere, and we therefore don't check for `chain == 0` anywhere. Would it make sense to add that check here (or in `AuthorizationRequest::new`), or somehow route through `Authorization::sign`?

rs/ethereum/cketh/minter/src/sweep/tests.rs
```diff
@@ -96,6 +94,129 @@ async fn should_sign_one_attestation_for_every_token_of_an_account() {
     );
 }
 
+#[tokio::test]
+async fn should_sign_and_record_one_authorization_for_every_account() {
```
mbjorkqvist, line 98: From what I can tell, there's no test that tests a plain cache hit - a second `create_pending_sweeper_requests` against the same sweeper contract signing nothing?

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: yes, this is part of that [reply](https://github.com/dfinity/ic/pull/11360#discussion_r3875734162) and the ticket you mentioned.

> Would it make sense to add an anvil test that verifies that a delegated deposit address ignores a zero nonce re-delegation? 

good idea! will be done in this PR because the added test should be small.

rs/ethereum/cketh/minter/src/sweep/tests.rs
gregorydemay: 🤖 Dropped in 0804bcd268: `Authorization::sign` had no callers — `sign_authorizations_batch` signs the tuple's hash through `sign_digest`, and the anvil helper builds and signs its own tuple. The `chain_id == 0` guard it carried is unreachable by construction anyway: the chain id always comes from `EthereumNetwork::chain_id()`, which is 1 or 11155111.

rs/ethereum/cketh/minter/src/sweep/tests.rs
gregorydemay: 🤖 Added in 210794db6a: `should_reuse_the_recorded_authorization_on_a_later_sweep` enqueues twice against the same sweeper contract and asserts a single `AuthorizedDepositAddress` event.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Applied in ae80c406ac, plus `authorizations_len` and a `cketh_minter_stored_authorizations` gauge next to the attestations one.

rs/ethereum/cketh/minter/src/state.rs
gregorydemay: 🤖 Agreed the unit test only covers the cache — the minter never reads the authority's on-chain nonce, so from `sweep/tests.rs` the only observable is the tuple it signs, which `authorization_digest` already pins; renamed it to `should_sign_a_fresh_authorization_when_the_sweeper_contract_changes`. The chain-level half is now pinned on anvil in 0b12b1c2bc: `a_zero_nonce_authorization_cannot_redelegate` installs a delegation with a nonce-0 tuple and shows a second nonce-0 tuple naming another delegate leaves the code unchanged.

rs/ethereum/cketh/minter/src/tx/eip_7702.rs
```diff
@@ -53,6 +54,67 @@ impl AsRef<Eip7702TransactionRequest> for Eip7702TransactionRequest {
...
+        chain_id: u64,
+        delegate: Address,
+        nonce: TransactionNonce,
+    ) -> Self {
+        Self {
+            account,
+            chain_id,
+            delegate,
+            nonce,
+        }
```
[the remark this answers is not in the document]
gregorydemay: No Account has custom PartialEq/PartialOrd

PR #11361, opened 2026-08-27
feat(ci): Disclose Hotfix workflow

basvandijk:
Adds a `workflow_dispatch` workflow that **discloses** a security hotfix built and deployed from the private dfinity/ic-private repository: it pushes the exact elected commit to this repository as the same `hotfix-*` branch, authenticated as the **"Push to RC Branch"** GitHub App. The push triggers Release Testing, which re-builds the commit, re-uploads its artifacts (rclone `--checksum --immutable`: identical bytes ⇒ no-op, differing bytes ⇒ hard failure) and mints the build-provenance attestation that release workflows and the mainnet-revisions updater verify CDN downloads against (attestation stack: #11323 → #11324 → #11325 → #11326).

Runs of this workflow have to be approved by either DRE or Infra.

### Flow

1. **Validate inputs** before touching any credential: `confirm` checkbox must be ticked (boolean `required: true` does not enforce this, same pattern as ci-kickoff-manual.yml), branch must match `hotfix-<suffix>` of `[A-Za-z0-9._-]+` (`/` excluded: branch protections and triggers match `hotfix-*` with fnmatch, where `*` does not cross `/`), commit must be a full 40-hex sha.
2. **Mint an app installation token** for `dfinity/ic` + `dfinity/ic-private` (same action pin as schedule-rc.yml).
3. **Validate remote state** via `ls-remote` only: the ic-private branch must exist with its tip **exactly** at the requested commit; on dfinity/ic the branch must be absent (proceed), already at the commit (succeed as a no-op) or anything else (fail loudly — never move an existing branch).
4. **Checkout dfinity/ic\@master**, fetch the branch from ic-private, re-check the tip (closes the validate→fetch race), reject non-commit objects, warn if the tree lacks the `attest-uploads` job (the run then can't mint an attestation).
5. **Push create-only**: `--force-with-lease="refs/heads/<branch>:"` (empty expected value = the ref must not exist, so the push can create but never move/fast-forward); a concurrent creation at the same commit is treated as a benign race; post-verify with `ls-remote`.
6. **Report** in the step summary: outcome, the triggered Release Testing run URL, and the operational reminders (uploads must be rclone no-ops; "immutable file modified" = incident signal for the infra team, never retry around it or touch CDN objects; no further pushes to the branch until `attest-uploads` is green).

### Why the tip must be exact

The attestation binds to the branch tip that Release Testing builds. A descendant or different commit would attest the wrong sha — hence exact-tip validation, the create-only lease, and the per-branch `concurrency` group (queued re-runs re-validate and converge to no-op/conflict instead of racing).

### Environment / prerequisites

- The **`DISCLOSE_HOTFIX`** environment exists (2026-08-27) with required reviewers (dfinity/dre + dfinity/infra), a master-only deployment branch policy (a Write user dispatching a modified copy of this workflow from a side branch cannot reach the credentials) and environment-scoped copies of the app credentials. It is separate from `PUSH_TO_RC_BRANCH` because required reviewers there would leave schedule-rc's nightly cron hanging.
- **Org-admin prerequisite done (2026-08-27)**: the "Push to RC Branch" app has access to `dfinity/ic-private` (Contents: read).
- In ic-private the synced workflow copy stays inert (`if: github.repository == 'dfinity/ic'`).

### Validation

- The exact `run` blocks (extracted from the YAML) were replayed against scratch bare repos: happy path creates the ref at exactly the commit; re-run is a no-op; existing-at-different-commit (including an ancestor — the silent-fast-forward case) is rejected both by the precheck and independently by the lease; private-tip-moved-after-validation aborts at the `FETCH_HEAD` re-check; benign/hostile create races behave as designed; `public-hotfix-x`, `hotfix-a/b`, `hotfix-a..b`, short shas and an unchecked `confirm` all fail input validation. 24/24 passed.
- `actionlint` (1.7.12) and shellcheck clean.
- Post-merge drill (after the attestation stack lands): disclose a scratch `hotfix-attest-drill-*` branch from ic-private at a post-stack commit; confirm push → Release Testing → rclone no-ops → `attest-uploads` mints; re-dispatch with identical inputs (no-op) and with a different commit (conflict failure); delete the branch.

### Operator usage

Actions UI → Disclose Hotfix → branch + commit + confirm, or:

```
gh workflow run disclose-hotfix.yml -R dfinity/ic -f branch=hotfix-... -f commit=<40-hex> -f confirm=true
```

### DIFF
--- a/.github/workflows/disclose-hotfix.yml
+++ b/.github/workflows/disclose-hotfix.yml
@@ -1,0 +1,264 @@
+name: Disclose Hotfix
+
+# Publish ("disclose") a security hotfix that was built and deployed from the
+# private dfinity/ic-private repository: push its exact elected commit to this
+# repository as the same hotfix-* branch. The push triggers the Release
+# Testing workflow, which re-builds the commit, re-uploads its artifacts to
+# the CDN (rclone --checksum --immutable: identical bytes are a no-op,
+# differing bytes fail the run) and mints the build-provenance attestation
+# that release workflows and the mainnet-revisions updater verify their CDN
+# downloads against.
+#
+# Invariants:
+#   - The pushed tip must be EXACTLY the requested commit: the attestation
+#     binds to the branch tip, so a descendant or different commit would
+#     attest the wrong sha. The push is create-only (--force-with-lease with
+#     an empty expected value): it can create the branch but never move an
+#     existing one.
+#   - Fail closed: the branch already existing at the requested commit is a
+#     successful no-op; existing at any other commit is a hard failure that
+#     needs manual investigation.
+#   - A downstream "immutable file modified" upload failure means the public
+#     rebuild differs from the CDN bytes that were deployed — an incident
+#     signal for the security team. Never retry around it and never
+#     delete/overwrite CDN objects to make it pass.
+#   - Do not push further commits to the branch until its attest-uploads job
+#     is green: Release Testing cancels in-flight runs per ref, and a new tip
+#     would attest a different sha.
+#
+# The DISCLOSE_HOTFIX environment gates the "Push to RC Branch" app
+# credentials: required reviewers approve each run, and the environment's
+# deployment branch policy (master only) prevents dispatching a modified copy
+# of this workflow from an unreviewed branch.
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'hotfix-* branch name on dfinity/ic-private (created here under the same name)'
+        required: true
+      commit:
+        description: 'Full 40-hex commit sha the branch tip must be at (the elected/deployed build)'
+        required: true
+      confirm:
+        description: 'Check this box to confirm the hotfix has been deployed, so it can safely be disclosed'
+        required: true
+        type: boolean
+
+# Queued re-runs for the same branch re-validate and converge to a no-op or a
+# conflict failure instead of racing each other.
+concurrency:
+  group: disclose-hotfix-${{ inputs.branch }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  disclose-hotfix:
+    name: Disclose Hotfix
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    # The synced copy of this workflow in ic-private must stay inert.
+    if: github.repository == 'dfinity/ic'
+    environment: DISCLOSE_HOTFIX
+    env:
+      GIT_TERMINAL_PROMPT: "0"
+    steps:
+      - name: Validate inputs
+        shell: bash
+        env:
+          BRANCH: ${{ inputs.branch }}
+          COMMIT: ${{ inputs.commit }}
+          CONFIRM: ${{ inputs.confirm }}
+        run: |
+          set -euo pipefail
+
+          # For boolean workflow_dispatch inputs, 'required: true' does not
+          # enforce that the box is ticked, so check it here.
+          if [ "$CONFIRM" != "true" ]; then
+            echo "confirm must be checked: it acknowledges the hotfix has been deployed and can safely be disclosed" >&2
+            exit 1
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/disclose-hotfix.yml
```diff
@@ -0,0 +1,264 @@
...
+#     binds to the branch tip, so a descendant or different commit would
+#     attest the wrong sha. The push is create-only (--force-with-lease with
+#     an empty expected value): it can create the branch but never move an
+#     existing one.
+#   - Fail closed: the branch already existing at the requested commit is a
+#     successful no-op; existing at any other commit is a hard failure that
+#     needs manual investigation.
+#   - A downstream "immutable file modified" upload failure means the public
+#     rebuild differs from the CDN bytes that were deployed — an incident
+#     signal for the security team. Never retry around it and never
```
basvandijk: Replace "security team" with "infra team" because they deal with non-reproducibility issues:
```suggestion
#     signal for the infra team. Never retry around it and never
```

.github/workflows/disclose-hotfix.yml
```diff
@@ -0,0 +1,264 @@
...
+#     delete/overwrite CDN objects to make it pass.
+#   - Do not push further commits to the branch until its attest-uploads job
+#     is green: Release Testing cancels in-flight runs per ref, and a new tip
+#     would attest a different sha.
+#
+# The DISCLOSE_HOTFIX environment gates the "Push to RC Branch" app
+# credentials: required reviewers approve each run, and the environment's
```
basvandijk:
```suggestion
# credentials: required reviewers (DRE / Infra) approve each run, and the environment's
```

.github/workflows/disclose-hotfix.yml
```diff
@@ -0,0 +1,264 @@
...
+          if [ "$fetched" != "$COMMIT" ]; then
+            echo "dfinity/ic-private $BRANCH moved to $fetched since validation; re-run" >&2
+            exit 1
+          fi
+          if [ "$(git cat-file -t "$COMMIT")" != "commit" ]; then
+            echo "$COMMIT is not a commit object" >&2
+            exit 1
+          fi
+
+          if ! git show "$COMMIT:.github/workflows/release-testing.yml" 2>/dev/null | grep -q attest-uploads; then
```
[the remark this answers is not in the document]
basvandijk: Fixed in 5533610785: the check now matches the top-level job key (`^  attest-uploads:` at the jobs indent level), so a comment, a `needs:` reference or a commented-out job no longer satisfies it.

.github/workflows/disclose-hotfix.yml
```diff
@@ -0,0 +1,264 @@
...
+            echo "- outcome: no-op — the branch was already at this commit" >>"$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+
+          echo "- outcome: **branch created; Release Testing takes over**" >>"$GITHUB_STEP_SUMMARY"
+
+          run_url=""
+          for _ in $(seq 24); do
+            run_url="$(gh api "repos/dfinity/ic/actions/workflows/release-testing.yml/runs?head_sha=$COMMIT" \
+              --jq '.workflow_runs[0].html_url // empty' 2>/dev/null || true)"
```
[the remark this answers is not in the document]
basvandijk: Fixed in 5533610785: the poll now filters by `head_sha`, `branch` and `event=push`, so it cannot select a run for another branch or a prior workflow_dispatch at the same commit.

.github/workflows/disclose-hotfix.yml
basvandijk: Applied in 5533610785 — also updated the same phrase in the Report step's summary text for consistency.

.github/workflows/disclose-hotfix.yml
basvandijk: Applied in 5533610785.

dfinityianblenke (approved):

PR #11365, opened 2026-08-28
feat: Record in-progress ingress messages after a subnet merge

mraszyk:
If the `subnet_merged` marker is set, `process_batch()` now resets it and calls `ReplicatedState::after_merge()`, which makes the following adjustments during the first round after a subnet merge:

* **Records in-progress ingress messages.** All not yet responded ingress-induced canister call contexts are recorded as `Processing` in the ingress history, so that the corresponding ingress messages can be tracked to completion. The entries are timestamped with the time of the batch being processed (`after_merge()` runs before `batch_time` is updated during round execution, so the state time would still be that of the state resulting from the merge).

  Subnet call contexts are ignored, as subnet merging ensures that the subnets being merged have no in-progress subnet call contexts.

  A message that already has an ingress history entry is left alone. Its status is expected to be `Processing`; anything else raises the new `mr_unexpected_ingress_status_after_merge` critical error.

* **Re-partitions canisters' input schedules,** exactly as `after_split()` does. Canisters hosted by the other merged subnets were remote before the merge and are local now, so their input queues may sit in the wrong input schedule; and a misplaced queue is only corrected once it becomes empty, which is not guaranteed to ever happen.

Drive-by cleanups:

* Extract `repartition_input_schedules()`: the input schedule loop was duplicated across `split()`, `after_split()` and `after_merge()`.
* Simplify the height handling in `process_batch_resets_(split|merge)_marker`, which kept a local copy of the height in sync with the `FakeStateManager` instead of just reading it back off the state manager.

### DIFF
--- a/rs/replicated_state/src/replicated_state.rs
+++ b/rs/replicated_state/src/replicated_state.rs
@@ -1,7 +1,7 @@
 use crate::canister_state::queues::{
     CanisterInput, CanisterQueuesLoopDetector, refunds::RefundPool,
 };
-use crate::canister_state::system_state::{CanisterOutputQueuesIterator, push_input};
+use crate::canister_state::system_state::{CallOrigin, CanisterOutputQueuesIterator, push_input};
 use crate::metadata_state::subnet_call_context_manager::{
     PreSignatureStash, ReshareChainKeyContext, SignWithThresholdContext,
 };
@@ -30,7 +30,7 @@ use ic_types::{
     CanisterId, NumBytes, SubnetId, Time,
     batch::{ConsensusResponse, RawQueryStats},
     consensus::idkg::IDkgMasterPublicKeyId,
-    ingress::IngressStatus,
+    ingress::{IngressState, IngressStatus},
     messages::{
         CallbackId, Ingress, MessageId, Refund, RequestOrResponse, Response, SubnetMessage,
     },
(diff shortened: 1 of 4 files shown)


rs/replicated_state/src/replicated_state.rs
```diff
@@ -1613,6 +1613,85 @@ impl ReplicatedState {
...
+
+                // Already recorded as `Processing`, nothing to do.
+                Some(IngressStatus::Known {
+                    state: IngressState::Processing,
+                    ..
+                }) => {}
+
+                // Any other status indicates a bug, report it. The existing entry is
+                // preserved, as overwriting a terminal status would be worse.
+                Some(unexpected_status) => {
```
schneiderstefan, line 1714: Maybe I am misunderstanding something fundamental here, but wouldn't that trip on call contexts that are for the old subnet that is we merge into, and that we never cooled down?

mraszyk: Every unresponded call context originating from an ingress message should be tracked as "Processing" in the ingress history (the no-op case above): after a subnet merge, newly merged call contexts might be missing and are re-inserted (the first arm). Any other ingress history status should not correspond to an unresponded call context and thus indicates a bug:
- Received => still in input queue, i.e., no call context at all;
- Replied/Rejected/Done => completed, i.e., call context is marked as responded.

rs/replicated_state/src/replicated_state.rs
```diff
@@ -1613,6 +1613,85 @@ impl ReplicatedState {
...
+    /// message can be tracked to completion.
+    ///
+    /// Only call contexts of canisters are considered; subnet call contexts are
+    /// ignored, as subnet merging ensures that the subnets being merged have no
+    /// in-progress subnet call contexts.
+    ///
+    /// A message that already has an ingress history entry is left alone; its
+    /// status is expected to be `Processing`, anything else is reported via
+    /// `on_unexpected_ingress_status()` (as it indicates a bug).
+    pub fn after_merge(
```
schneiderstefan, line 1634: Possibly not intended for this PR, but `after_split` also calls [split_input_schedules](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/replicated_state/src/canister_state/queues.rs?L1672), which you might also have to do here (or do something similar at least).

rs/replicated_state/src/replicated_state.rs
```diff
@@ -1613,6 +1613,85 @@ impl ReplicatedState {
...
+        ingress_memory_capacity: NumBytes,
+        on_unexpected_ingress_status: impl Fn(&MessageId, &IngressStatus),
+    ) {
+        assert!(
+            self.metadata.subnet_merged,
+            "Not a state resulting from a subnet merge"
+        );
+        self.metadata.subnet_merged = false;
+
+        let time = self.time();
```
schneiderstefan: Something claude pointed out is that this time is still the time of the old state, not of the new block. Not sure it matters.

rs/replicated_state/src/replicated_state.rs
schneiderstefan: Follow-up question: Is it possible that there were entries in the cooled-down subnet with status received, and we are dropping them during the merge?

schneiderstefan:
Somewhat related observation: We need to make sure the subnet is cooled down and/or halted for at least 5 minutes, so that you cannot re-submit ingress messages. Realistically, we won't be able to do everything that fast anyway though.

rs/replicated_state/src/replicated_state.rs
mraszyk: No, we'd only proceed with merging if the ingress history of the source subnet contains only Processing entries (which are restored by this PR on the target subnet). This is captured in the design [doc](https://docs.google.com/document/d/1KbcbLBGqf1oOXS-2uzdTdE8WD8bVjyAmfe8t_B67-rc/edit?tab=t.0#heading=h.95fu7ssl43o4) and there's a metric for it visualized on the subnet merging [dashboard](https://github.com/dfinity-ops/k8s/pull/2604).

mraszyk:
> Somewhat related observation: We need to make sure the subnet is cooled down and/or halted for at least 5 minutes, so that you cannot re-submit ingress messages. Realistically, we won't be able to do everything that fast anyway though.

Good point! Added a note "make sure at least five minutes have elapsed since the source subnet was marked as “cooling down”" to the design doc just before rerouting canister ranges.

rs/replicated_state/src/replicated_state.rs
schneiderstefan: Oh I see, and the entries that are in received state when we start cooling down would time out. We wouldn't execute them (execution is stopped), but they time out while in this state.

mraszyk: Exactly, they'd time out after 5 minutes and then we'd wait for 5 more minutes for them to disappear from the ingress history entirely.

rs/replicated_state/src/replicated_state.rs
mraszyk: Good catch, done in de3192b120: `after_merge()` now re-partitions the input schedules exactly like `after_split()` does. It is both safe and necessary here: canisters hosted by the other merged subnets were remote before the merge and are local now, so their input queues can sit in the remote schedule, and a misplaced queue is only corrected once it happens to become empty.

rs/replicated_state/src/replicated_state.rs
mraszyk: Right, `after_merge()` runs before `batch_time` is updated (that only happens during round execution), so this was the time of the state resulting from the merge. It does not really matter, but switched to the block time nonetheless in de3192b120: `after_merge()` now takes an explicit `batch_time` and `process_batch()` passes `batch.time`.

schneiderstefan (approved):

PR #11366, opened 2026-08-28
feat(registry): add merge_subnets endpoint

mraszyk:
The registry side of a subnet merge: a `merge_subnets` endpoint on the registry canister, and a `MergeSubnets` NNS function mapped to it under the `SubnetManagement` topic, so that a single proposal performs the range reassignment.

The endpoint takes a source and a destination subnet ID and merges the canister ID ranges of the source subnet into the canister ID range set of the destination subnet, so that the canisters hosted by the source subnet are routed to the destination subnet afterwards. The routing table is updated via `modify_routing_table`, which assigns the source subnet's ranges to the destination subnet and coalesces the resulting adjacent ranges. Only the routing table is touched: neither subnet record is modified and the source subnet is not deleted, so decommissioning a subnet is `merge_subnets` followed by `delete_subnet` — the source subnet merely does not host any canister ID range anymore.

The endpoint is restricted to governance and rejects payloads where the two subnet IDs are equal, either subnet is unknown, the source subnet hosts no canister ID range, or the source subnet's ranges overlap an ongoing canister migration.

Recovering the destination subnet at the merged state — the recovery catch-up package, the fresh DKG and the unhalting — is deliberately out of scope here and stays with `recover_subnet`.

One unrelated `StateMachine` fix comes along: the fake `SetupInitialDKGResponse` it produces for pending `setup_initial_dkg` requests had its two fields swapped, putting the `HighThreshold`-tagged transcript into `low_threshold_transcript_record` and vice versa.


rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,544 @@
...
+                format!("{what} was updated during the `setup_initial_dkg` call"),
+            );
+        }
+        assert_eq!(
+            self.get_routing_table_or_panic(post_call_registry_version)
+                .ranges(source_subnet),
+            source_ranges,
+            "{LOG_PREFIX}The canister ID ranges of subnet {source_subnet} were updated during the \
+             `setup_initial_dkg` call",
+        );
```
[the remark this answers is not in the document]
mraszyk: Good catch — fixed in 155414e1cb: the post-call guard now also re-checks that the canister migrations at the post-call registry version are disjoint from the source subnet's ranges, mirroring the pre-call validation.

rs/state_machine_tests/src/lib.rs
```diff
@@ -1865,6 +1865,42 @@ impl Default for StateMachineBuilder {
...
+            let public_key = (&ni_dkg_transcript).try_into().unwrap();
+            let public_key_der = threshold_sig_public_key_to_der(public_key).unwrap();
+            let subnet_id = PrincipalId::new_self_authenticating(&public_key_der).into();
+            let mut high_threshold_transcript_record = ni_dkg_transcript.clone();
+            high_threshold_transcript_record.dkg_id.dkg_tag = NiDkgTag::HighThreshold;
+            let mut low_threshold_transcript_record = ni_dkg_transcript;
+            low_threshold_transcript_record.dkg_id.dkg_tag = NiDkgTag::LowThreshold;
+            let initial_transcript_records = SetupInitialDKGResponse {
+                low_threshold_transcript_record: high_threshold_transcript_record.into(),
+                high_threshold_transcript_record: low_threshold_transcript_record.into(),
```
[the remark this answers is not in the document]
mraszyk: Correct, the two fields were swapped. Fixed in 155414e1cb.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,556 @@
...
+            )
+            .ok_or_else(|| {
+                format!("initial DKG subnet {initial_dkg_subnet_id} is not a known subnet")
+            })?;
+        }
+
+        let mut cup_contents = self
+            .get_subnet_catch_up_package(destination_subnet, Some(pre_call_registry_version))
+            .map_err(|err| format!("failed to get the CUP of {destination_subnet}: {err}"))?;
+        cup_contents.registry_store_uri = None;
```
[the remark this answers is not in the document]
mraszyk: Right — the destination subnet can hold no chain keys, but its existing CUP could still carry initializations from an earlier recovery. Fixed in 42a48c1aea: both `chain_key_initializations` and `ecdsa_initializations` are now cleared on the recovery CUP, matching what recovery does. The integration test seeds stale initializations into the destination subnet's CUP and asserts they are gone after the merge.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            return Err(format!(
+                "source subnet {source_subnet} does not host any canister ID range"
+            ));
+        }
+
+        // Rerouting the canister ID ranges of the source subnet would break any ongoing canister
+        // migration out of those ranges: the migrated ranges would end up being hosted by the
+        // destination subnet, which is not on the recorded migration trace.
+        if let Some(canister_migrations) = self.get_canister_migrations(version)
+            && !are_disjoint(canister_migrations.ranges(), source_ranges.iter())
```
pierugo-dfinity, line 65: Is it fine to have canister migrations from `destination_ranges`?

pierugo-dfinity (commented):
Had a quick look, will continue tomorrow

rs/nns/governance/unreleased_changelog.md
```diff
@@ -9,6 +9,13 @@ on the process that this file is part of, see
 
 ## Added
 
+* Added a new `NnsFunction` variant `MergeSubnets`, which proposes to merge a
+  subnet into another subnet: the canister ID ranges of the source subnet are
+  merged into the canister ID range set of the destination subnet, so that all
+  canisters that used to be hosted by the source subnet are routed to the
+  destination subnet. Only the routing table is updated: neither subnet record
+  is modified and the source subnet is not deleted.
```
daniel-wong-dfinity-org-twin: So, basically, this migrates all canisters from one subnet to another, right? I don't think that's what "merge" calls to mind, at least, for me it doesn't. I mean, I do not hate the name, so it's probably not worth changing it at this point, but if you are willing to entertain suggestions, how about "migrate all subnet canisters", or "evacuate subnet", or something? This would make it clearer that the two subnets survive; it's just the contents of one are dumped into another.

rs/registry/canister/canister/canister.rs
```diff
@@ -1085,6 +1086,24 @@ fn reroute_canister_ranges_(payload: RerouteCanisterRangesPayload) {
...
+
+#[candid_method(update, rename = "merge_subnets")]
+fn merge_subnets_(payload: MergeSubnetsPayload) {
+    registry_mut()
+        .merge_subnets(payload)
+        .unwrap_or_else(|error_message| {
+            trap_with(&format!(
+                "{LOG_PREFIX} Merge subnets failed: {error_message}"
+            ))
+        });
```
daniel-wong-dfinity-org-twin, line 1103: This seems unusual. I think what usually happens is panic within the impl, not trap in canister.rs. I mean, yes, I see examples of this, but I think they are in the minority.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        if source_subnet == destination_subnet {
+            return Err(format!(
+                "source subnet {source_subnet} and destination subnet {destination_subnet} must be different subnets"
+            ));
+        }
+
+        let version = self.latest_version();
+
+        self.get(&make_subnet_record_key(source_subnet).into_bytes(), version)
```
daniel-wong-dfinity-org-twin, line 43:
```suggestion
        // The subnets must exist.
        self.get(&make_subnet_record_key(source_subnet).into_bytes(), version)
```

Seems like this the most "basic" requirement. To me, this suggests that these checks would be first, but this is also fine.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// modified and, in particular, the source subnet is not deleted.
+    pub fn merge_subnets(&mut self, payload: MergeSubnetsPayload) -> Result<(), String> {
+        println!("{LOG_PREFIX}merge_subnets: {payload:?}");
+
+        let MergeSubnetsPayload {
+            source_subnet,
+            destination_subnet,
+        } = payload;
+
+        if source_subnet == destination_subnet {
```
daniel-wong-dfinity-org-twin, line 34:
```suggestion
        // The subnets must be distinct.
        if source_subnet == destination_subnet {
```

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            version,
+        )
+        .ok_or_else(|| format!("destination {destination_subnet} is not a known subnet"))?;
+
+        let routing_table = self.get_routing_table_or_panic(version);
+        let source_ranges = routing_table.ranges(source_subnet);
+        if source_ranges.is_empty() {
+            return Err(format!(
+                "source subnet {source_subnet} does not host any canister ID range"
+            ));
```
daniel-wong-dfinity-org-twin: Doesn't that just make it trivially Ok, not Err?

Err is reasonable -> do as you see fit.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        // Rerouting the canister ID ranges of the source subnet would break any ongoing canister
+        // migration out of those ranges: the migrated ranges would end up being hosted by the
+        // destination subnet, which is not on the recorded migration trace.
+        if let Some(canister_migrations) = self.get_canister_migrations(version)
```
daniel-wong-dfinity-org-twin:
```suggestion
        // The source subnet must not have any canister that is migrating away.
        if let Some(canister_migrations) = self.get_canister_migrations(version)
```

Could this restriction be lifted later?

rs/registry/canister/src/mutations/routing_table.rs
```diff
@@ -378,6 +378,24 @@ impl Registry {
         })
     }
 
+    /// Makes a registry mutation that merges all canister ID ranges currently
+    /// assigned to the `source` subnet into the canister ID range set of the
+    /// `destination` subnet. After the mutation, `source` does not host any
+    /// canister ID range anymore.
+    pub fn merge_subnets_mutation(
+        &self,
+        version: u64,
```
daniel-wong-dfinity-org-twin, line 387: Seems a little strange to me that this takes version. I mean, what if I pass not the current/latest version? How could that ever be useful?

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            && !are_disjoint(canister_migrations.ranges(), source_ranges.iter())
+        {
+            return Err(format!(
+                "source subnet {source_subnet} hosts canister ID ranges with ongoing canister migrations"
+            ));
+        }
+
+        self.maybe_apply_mutation_internal(self.merge_subnets_mutation(
```
daniel-wong-dfinity-org-twin, line 72: Interesting that you spin out the "real" work, but inline the validation. The other way makes more sense to me, because "Lead with the lede.". But this is also ok, so do as you see fit.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        // Rerouting the canister ID ranges of the source subnet would break any ongoing canister
+        // migration out of those ranges: the migrated ranges would end up being hosted by the
+        // destination subnet, which is not on the recorded migration trace.
+        if let Some(canister_migrations) = self.get_canister_migrations(version)
+            && !are_disjoint(canister_migrations.ranges(), source_ranges.iter())
+        {
+            return Err(format!(
+                "source subnet {source_subnet} hosts canister ID ranges with ongoing canister migrations"
```
daniel-wong-dfinity-org-twin, line 68:
```suggestion
                "Canisters are migrating out of the source subnet ({source_subnet})."
```

"canister ID ranges" buries the lede.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        self.get(&make_subnet_record_key(source_subnet).into_bytes(), version)
+            .ok_or_else(|| format!("source {source_subnet} is not a known subnet"))?;
+        self.get(
+            &make_subnet_record_key(destination_subnet).into_bytes(),
+            version,
+        )
+        .ok_or_else(|| format!("destination {destination_subnet} is not a known subnet"))?;
+
+        let routing_table = self.get_routing_table_or_panic(version);
```
daniel-wong-dfinity-org-twin, line 52:
```suggestion
        // The source subnet must be nonempty.
        let routing_table = self.get_routing_table_or_panic(version);
```

Not sure why though. I mean, most likely, the user is confused, but you can't be sure of that.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+        CanisterIdRange {
+            start: CanisterId::from_u64(start),
+            end: CanisterId::from_u64(end),
+        }
+    }
+
+    /// Returns a registry with two subnets, `SUBNET_1` and `SUBNET_2`, where
+    /// `SUBNET_1` hosts the canister ID ranges `[10, 19]` and `[30, 39]` and
+    /// `SUBNET_2` hosts the canister ID range `[20, 29]`.
+    fn new_two_subnets_fixture_registry() -> Registry {
```
daniel-wong-dfinity-org-twin: Consider using lazy_static pseudo-constant instead of a 0-argument fn. Fixtures are fixed, i.e. constants.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+        let mut registry = invariant_compliant_registry(0);
+
+        let (mutate_request, node_ids_and_dkg_pks) =
+            prepare_registry_with_nodes(/* start_mutation_id = */ 1, /* nodes = */ 2);
```
daniel-wong-dfinity-org-twin:
```suggestion
        let (mutate_request, node_ids_and_dkg_pks) = prepare_registry_with_nodes(
            1, // start_mutation_id
            2, // nodes
        );
```

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            destination_subnet: SUBNET_2,
+        });
+
+        // Step 3: Verify results.
+        let error_message = result.unwrap_err();
+        assert!(
+            error_message.contains(&format!("source {SUBNET_3} is not a known subnet")),
+            "{error_message}"
+        );
+    }
```
daniel-wong-dfinity-org-twin, line 272: How about, like the previous test, assert that the routing table is not modified?

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            destination_subnet: SUBNET_3,
+        });
+
+        // Step 3: Verify results.
+        let error_message = result.unwrap_err();
+        assert!(
+            error_message.contains(&format!("destination {SUBNET_3} is not a known subnet")),
+            "{error_message}"
+        );
+    }
```
daniel-wong-dfinity-org-twin, line 295: Ditto.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+            "{error_message}"
+        );
+        assert_eq!(
+            get_routing_table_entries(&registry),
+            vec![
+                (range(10, 19), SUBNET_1),
+                (range(20, 29), SUBNET_2),
+                (range(30, 39), SUBNET_1),
+            ],
+        );
```
daniel-wong-dfinity-org-twin, line 324: I guess (with my suggestions), this gets repeated a bunch of times. So, how about

```
#[track_caller]
fn assert_routing_table_unchanged_vs_fixture(registry: &Registry) {...
```

?

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+};
+
+mod common;
+use common::test_helpers::{check_error_message, check_subnet_for_canisters};
+
+/// Exercises the `merge_subnets` endpoint end to end. The payload validation
+/// itself is covered by the unit tests of `Registry::merge_subnets`, so this test
+/// only covers what those cannot: that the endpoint is reachable with a Candid
+/// encoded payload, that only governance may call it, and that the resulting
+/// routing table is visible through the canister's query API.
```
daniel-wong-dfinity-org-twin, line 25: This doesn't seem to cover much... I mean, doesn't the delcared interface vs. implemented interface checker already verify that we'll decode properly?

Anyway, this doesn't really hurt, so don't get rid of it. It's just a little sad that we have this much test code to cover such a small amount of functionality...

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+#[test]
+fn test_merge_subnets() {
+    state_machine_test_on_nns_subnet(|runtime| {
+        async move {
+            // Step 1: Prepare the world: two subnets where subnet 2 hosts the canister
+            // ID range [0, 255] and subnet 1 hosts [256, 511].
+            let (subnet_1_mutation, subnet_id_1, subnet_id_2_option, _, _) =
+                prepare_registry_with_two_node_sets(
+                    /* num_nodes_in_subnet = */ 4, /* num_unassigned_nodes = */ 4, true,
```
daniel-wong-dfinity-org-twin:
```suggestion
                    4, // num_nodes_in_subnet
                    4, // num_unassigned_nodes
                    true, // ?
```

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+                );
+            let subnet_id_2 = subnet_id_2_option.unwrap();
+            let rt_mutation = {
```
daniel-wong-dfinity-org-twin:
```suggestion
            let routing_table_mutation = {
```

"rt" is not a word known to the dictionary.

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+            )
+            .await;
+
+            let governance_fake = set_up_universal_canister(&runtime).await;
+            assert_eq!(
+                governance_fake.canister_id(),
+                ic_nns_constants::GOVERNANCE_CANISTER_ID
+            );
+
+            // Step 2: A caller other than governance may not merge subnets.
```
daniel-wong-dfinity-org-twin:
```suggestion
            // Step 2: Run the code under test.
            
            // Step 2.1: The sad case: caller is not the Governance canister.
```

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+            )
+            .await
+            .unwrap();
+
+            // Step 4: Verify results: the canisters formerly hosted by subnet 1 are now
+            // hosted by subnet 2, and the canisters of subnet 2 stay put.
+            check_subnet_for_canisters(
+                &registry,
+                vec![
+                    (CanisterId::from(0), subnet_id_2),
```
daniel-wong-dfinity-org-twin: What about non-endpoint canister IDs? I mean, since everything from 0 to 511 is taken, you could concisely check all of (0..=511), right? In fact, it would probably be even less code than this!

rs/registry/canister/tests/merge_subnets.rs
```diff
@@ -0,0 +1,115 @@
...
+                        MergeSubnetsPayload {
+                            source_subnet: subnet_id_1,
+                            destination_subnet: subnet_id_2,
+                        },
+                    )
+                    .await as Result<(), String>,
+                "not authorized",
+            );
+
+            // Step 3: Run the code under test: merge subnet 1 into subnet 2.
```
daniel-wong-dfinity-org-twin:
```suggestion
            // Step 2.2: The happy case: merging subnets succeeds, because Governance is the caller.
```

daniel-wong-dfinity-org-twin (commented):
I would have split this between registry and Governance. I mean, the governance changes are small, so I guess this is ok.

It would be nice if Governance did some validation, like, does the payload even decode to the right type, but this is a general flaw with NnsFunctions, and since split is an NnsFunction, it probably makes sense for this also to be one...

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Yes, because subnet merging does not (concurrently) move any ranges away from `destination_ranges`. But actually I'm gonna drop this check since
- the registry key `canister_migrations` is used for subnet splitting and we are definitely not gonna merge and split subnets at the same time;
- streams from/to the source subnet must be empty during subnet splitting and thus the registry key `canister_migrations` is a no-op for the source subnet anyway.

CC @alin-at-dfinity

rs/nns/governance/unreleased_changelog.md
mraszyk: The word "migrate" is already overloaded to refer to (i) subnet splitting via the registry key `canister_migrations` and (ii) single canister ID migration (routing table change) orchestrated by a dedicated NNS canister. So I'd prefer to not use it in yet another context.

I suppose your confusion comes from the fact that the source subnet survives (it must be cleaned up by a separate proposal) so how about `MergeSubnetCanisterRanges`?

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: I plan to drop this check: see https://github.com/dfinity/ic/pull/11366#discussion_r3920993403

rs/registry/canister/src/mutations/merge_subnets.rs
pierugo-dfinity: > `canister_migrations` is used for subnet splitting

I see it's also used in `prepare_canister_migration` though I'm not sure in what context is this mutation called.

mraszyk: This is the subnet splitting flow.

mraszyk: Dropped the check in c189291c6e.

rs/registry/canister/canister/canister.rs
mraszyk: Kept as is: the neighbouring routing/migration endpoints (`prepare_canister_migration`, `reroute_canister_ranges`, `split_subnet`, `complete_canister_migration`) all return a `Result` and trap here, and the `Result` is also what lets the unit tests assert on the error message.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Added in c189291c6e. I kept the order though: the distinctness check needs no registry lookup at all, so it comes first.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Added in c189291c6e.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Dropped the check entirely in c189291c6e, see https://github.com/dfinity/ic/pull/11366#discussion_r3920993403.

rs/registry/canister/src/mutations/routing_table.rs
mraszyk: Kept for consistency: every mutation builder in this file (`reroute_canister_ranges_mutation`, `remove_subnet_from_routing_table_mutation`, `migrate_canisters_to_subnet`) takes the version, and the private `modify_routing_table` needs it. The only caller passes `latest_version()`, which it also uses for the validation, so both see the same registry version.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Kept as is: `merge_subnets_mutation` sits next to its siblings (`reroute_canister_ranges_mutation` etc.) in `routing_table.rs`, which is where one looks for routing table mutations.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Moot now: the check is gone in c189291c6e.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Added in c189291c6e. The check rejects a merge that would be a no-op, which most likely means that source and destination were swapped, or that the subnet was already merged.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Done in c189291c6e: the fixture registry is now the lazy_static `TWO_SUBNETS_REGISTRY` and its routing table the constant `FIXTURE_ROUTING_TABLE_ENTRIES`.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Done in c189291c6e.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Done in c189291c6e, using the helper below.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Done in c189291c6e, using the helper below.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Done in c189291c6e: `assert_routing_table_unchanged_vs_fixture`.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Agreed. It only covers the Candid decoding and the caller check; I'm keeping it at that.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Done in c189291c6e -- and the existing comments were wrong: the parameters are `num_nodes_in_subnet1`, `num_nodes_in_subnet2` and `assign_nodes_to_subnet2`.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Done in c189291c6e.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Done in c189291c6e.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Done in c189291c6e: the test now checks all of `0..=511`. It still runs in ~11s.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Done in c189291c6e.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,304 @@
...
+        )
+        .ok_or_else(|| format!("destination {destination_subnet} is not a known subnet"))?;
+
+        // The source subnet must be nonempty.
+        let routing_table = self.get_routing_table_or_panic(version);
+        if routing_table.ranges(source_subnet).is_empty() {
+            return Err(format!(
+                "source subnet {source_subnet} does not host any canister ID range"
+            ));
+        }
```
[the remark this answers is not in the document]
mraszyk: Subnet merging comes with a set of manually checked requirements (see [design doc](https://docs.google.com/document/d/1KbcbLBGqf1oOXS-2uzdTdE8WD8bVjyAmfe8t_B67-rc/edit?tab=t.0#heading=h.95fu7ssl43o4)), the subnets not being split being one of them. Hence, omitting this particular check makes sense to me since it'd only result in a failed NNS governance proposal.

rs/registry/canister/src/mutations/merge_subnets.rs
pierugo-dfinity: > This is the subnet splitting flow.

Ah yes, the old manual flow

> we are definitely not gonna merge and split subnets at the same time

Well, if we ever accidentally did, at least this check would prevent us from nuking the IC 👀

rs/registry/canister/src/mutations/merge_subnets.rs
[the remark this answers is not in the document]
alin-at-dfinity: As long as the check does nor conflict with the manual checks, I would add it.

In particular, a canister appearing in a `canister_migrations` entry means that there may still exist messages addressed to that canister that were routed into the "wrong" stream. Since subnet merging does not touch `canister_migrations` (and thus the actual host subnet after the merge is never added to the migration path) such messages might end up rerouted forever or dropped.

rs/registry/canister/src/mutations/merge_subnets.rs
alin-at-dfinity: While I agree that if everything works according to plan we should never have non-empty streams to or from the source subnet at this point, I believe there is value in defense in depth.

If under normal circumstances this is going to be a no-op anyway, then I'd rather keep it than drop it. Also for documentation purposes: you could simply replace the comment above (since it's obvious we don't want to mix merges and splits) with "Checks before the merge should ensure that this never triggers", where "this" is documented by the error message.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+impl Registry {
+    /// Merges the canister ID ranges of the source subnet into the canister ID
+    /// range set of the destination subnet.
+    ///
+    /// After this operation, all canisters that used to be hosted by the source
+    /// subnet are routed to the destination subnet and the source subnet does
+    /// not host any canister ID range anymore.
+    ///
+    /// Note that only the routing table is updated: neither subnet record is
+    /// modified and, in particular, the source subnet is not deleted.
```
alin-at-dfinity, line 24: I was going to comment below, on `MergeSubnetsPayload` that it would be cleaner to talk about moving canisters than merging canister ranges. But then, the registry mutation is specifically the "merging canister ranges" part of the canister migration, so this is the appropriate vocabulary.

What may be missing from this doc comment though, is a very high level (read, one or two short sentences) description of the merging process pointing out that both subnets are halted at this point, with all streams to and from the source subnet empty; and that the destination subnet (with all canister states) will only be unhalted after this registry change is applied (I was going to say "after it has been observed by all subnets", but that turns out to not be necessary).

alin-at-dfinity, line 24: Taking a step back, do we have (or plan to have) a description of the subnet merging process anywhere in the code? (Essentially an abbreviated version of the process described in the design doc.)

If not, then this (or, probably better, the state tool command) would be a good place to have it.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn merge_subnets(&mut self, payload: MergeSubnetsPayload) -> Result<(), String> {
+        println!("{LOG_PREFIX}merge_subnets: {payload:?}");
```
alin-at-dfinity, line 26: Ignorant question: Is this intended as a canister debug log or is it just a leftover?

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -0,0 +1,331 @@
...
+        // Step 3.2: The canister ID ranges of both subnets are now hosted by the
+        // destination subnet, and the three adjacent ranges got merged into one.
+        assert_eq!(
+            get_routing_table_entries(&registry),
+            vec![(range(10, 39), SUBNET_2)],
+        );
+    }
+
+    #[test]
+    fn test_merge_subnets_into_subnet_without_canister_id_ranges() {
```
alin-at-dfinity, line 198: Is there any use case for "merging" canisters onto an empty subnet?

I suppose you could e.g. move all canisters from an application subnet to a cloud engine (or the other way around) this way, but wouldn't it be simpler to add a "change subnet type" proposal instead?

rs/nns/governance/api/src/types.rs
```diff
@@ -4335,6 +4335,11 @@ pub enum NnsFunction {
     /// `SetupInitialDKG` requests without an explicit subnet id are routed to the
     /// calling subnet (NNS).
     SetDefaultInitialDkgSubnet = 58,
+    /// Merge a subnet into another subnet: the canister ID ranges of the source
+    /// subnet are merged into the canister ID range set of the destination subnet,
+    /// so that all canisters that used to be hosted by the source subnet are routed
+    /// to the destination subnet.
```
alin-at-dfinity: A briefer description would be:

```suggestion
    /// Merge a subnet into another subnet: in the routing table, reassigns all
    /// canister ranges hosted by the source subnet to the destination subnet.
```

Or "remaps".

rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto
```diff
@@ -513,6 +513,12 @@ enum NnsFunction {
   // `SetupInitialDKG` requests without an explicit subnet id are routed to the
   // calling subnet (NNS).
   NNS_FUNCTION_SET_DEFAULT_INITIAL_DKG_SUBNET = 58;
+
+  // Merge a subnet into another subnet: the canister ID ranges of the source
+  // subnet are merged into the canister ID range set of the destination subnet,
+  // so that all canisters that used to be hosted by the source subnet are routed
+  // to the destination subnet.
```
alin-at-dfinity: Ditto.

rs/nns/governance/src/proposals/execute_nns_function.rs
```diff
@@ -944,6 +948,13 @@ impl ValidNnsFunction {
                 calls are routed when no subnet is specified explicitly in the request. If unset, \
                 such requests are routed to the calling subnet (NNS)."
             }
+            ValidNnsFunction::MergeSubnets => {
+                "Merge a subnet into another subnet. The canister ID ranges of the source subnet \
+                are merged into the canister ID range set of the destination subnet, so that all \
+                canisters that used to be hosted by the source subnet are routed to the \
+                destination subnet. Only the routing table is updated: neither subnet record is \
+                modified and the source subnet is not deleted."
+            }
```
alin-at-dfinity, line 955: And here.

rs/nns/governance/unreleased_changelog.md
```diff
@@ -9,6 +9,13 @@ on the process that this file is part of, see
 
 ## Added
 
+* Added a new `NnsFunction` variant `MergeSubnets`, which proposes to merge a
+  subnet into another subnet: the canister ID ranges of the source subnet are
+  merged into the canister ID range set of the destination subnet, so that all
+  canisters that used to be hosted by the source subnet are routed to the
+  destination subnet. Only the routing table is updated: neither subnet record
```
alin-at-dfinity: And here.

rs/nns/governance/unreleased_changelog.md
daniel-wong-dfinity-org-twin: Do as you see fit.

Yes, the fact that the source subnet remains afterwards is what makes me think that "merge" is not so apt. When you merge two companies, the employees do not simply move from one to the other. Rather, you are left with just one company.

The existing "migrate" operation moves a canister from one subnet to another, right? Isn't that what's going on here, just a different number of canisters? To me, that makes "migrate" an even better choice. I don't see it as conflation; I see it as consistency.

There's also my other suggestion, "evacuate", which does not conjure images of the existing thing called "migrate". This more emphasizes that ALL canisters are moved out (but does not mention that they will all go to the same destination), which to me, is not as important as the fact that existing canisters are finding a new home, have an actual DESTINATION, not just left stranded in the middle of the street, as might happen if you evacuate a building due to fire. "migrate" is less chaotic. It implies a planned destination, not just a departure.

rs/registry/canister/canister/canister.rs
daniel-wong-dfinity-org-twin: ok.

rs/registry/canister/src/mutations/merge_subnets.rs
daniel-wong-dfinity-org-twin: ok. Yes, I figured that your rationale was that no look up was needed.

rs/registry/canister/src/mutations/merge_subnets.rs
daniel-wong-dfinity-org-twin: Yes, I figured that the motivation here is "the user is very very likely be confused". It's reasonable.

rs/registry/canister/tests/merge_subnets.rs
daniel-wong-dfinity-org-twin: tangent: maybe add one negative case? Sorry, I should have thought of this before.

rs/registry/canister/src/mutations/merge_subnets.rs
```diff
@@ -149,10 +148,20 @@ mod tests {
...
     }
 
+    /// Asserts that the routing table is still the one of `TWO_SUBNETS_REGISTRY`,
+    /// i.e. that the (failed) merge did not modify it.
+    #[track_caller]
+    fn assert_routing_table_unchanged_vs_fixture(registry: &Registry) {
+        assert_eq!(
+            get_routing_table_entries(registry),
+            FIXTURE_ROUTING_TABLE_ENTRIES.to_vec(),
+        );
```
daniel-wong-dfinity-org-twin: Do as you see fit.

Ah. This is so concise now that FIXTURE_ROUTING_TABLE_ENTRIES can serve as a reference value. Inlining might make more sense.

(My suggestion to spin out made more sense when the reference value was a multi-line "literal".)

daniel-wong-dfinity-org-twin (approved):

rs/registry/canister/src/mutations/merge_subnets.rs
pierugo-dfinity: For the record: with the current implementation, "change subnet type" might be tricky because we have invariants in place that ensures that cloud engines only have nodes of reward type `Type4.*` and that application subnets only have nodes of reward type **not** `Type4.*`. So changing the subnet type "just like that" is currently impossible, but would be possible by moving all canister ranges into a different subnet of different membership, like we do here.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: This is how NNS canisters produce their logs.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: This test primarily contrasts the condition that the source subnets ranges must be non-empty while the destination subnet might have no canister ID ranges. But there can't really be a subnet with no canister ID ranges (subnets have 1M canisters in their ranges, but cannot host so many canisters so even with canister migration, you cannot really empty a subnet). So this is essentially a unit test for the condition rather than a test for a realistic scenario.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: This place might not be ideal since this registry mutation is only one of multiple registry mutations needed for subnet merging. I have a draft PR with an end-to-end (system test) for subnet merging - this system test and its (high-level) description could be a good place to mirror the design doc publicly.

rs/nns/governance/api/src/types.rs
mraszyk: Applied at all four sites in 6a8dd69d1c. I kept one extra sentence ("The source subnet is not deleted.") in the proposal description and the changelog, since that turned out to be the part readers find surprising; the enum doc comments use your wording verbatim.

rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto
mraszyk: Done in 6a8dd69d1c -- see [above](https://github.com/dfinity/ic/pull/11366#discussion_r3936012716).

rs/nns/governance/src/proposals/execute_nns_function.rs
mraszyk: Done in 6a8dd69d1c. This one is the proposal-facing description, so it also keeps "The source subnet is not deleted.".

rs/nns/governance/unreleased_changelog.md
mraszyk: Done in 6a8dd69d1c.

rs/registry/canister/src/mutations/merge_subnets.rs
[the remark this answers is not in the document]
mraszyk: Restored the check in 6a8dd69d1c, with the comment reworded as you suggested:

```rust
// Separate checks before the merge should ensure that this never triggers: rerouting the
// canister ID ranges of the source subnet would break any ongoing canister migration out
// of those ranges, since the migrated ranges would end up being hosted by the destination
// subnet, which is not on the recorded migration trace.
```

Agreed on defense in depth -- and your point about messages already routed into the "wrong" stream is a stronger reason than the invariant failure alone, so this isn't merely a safety net for a case that can't happen.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Restored the check in 6a8dd69d1c, per @alin-at-dfinity's [comment](https://github.com/dfinity/ic/pull/11366#discussion_r3931647475). It still only covers `source_ranges`: the destination subnet keeps hosting its own ranges, so the merge cannot invalidate a migration trace involving `destination_ranges`.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Added in 6a8dd69d1c:

```rust
/// This is one step of the subnet merging process: by the time this mutation
/// is applied, both subnets are halted and all streams to and from the source
/// subnet are empty. The destination subnet is only unhalted after this
/// registry change has been applied.
```

I left out the part about the canister states of the source subnet having been merged into the destination subnet's state, since that is not accurate.

rs/registry/canister/tests/merge_subnets.rs
mraszyk: Added in 6a8dd69d1c: canister ID 512, just past the merged ranges, still resolves to no subnet.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Agreed -- inlined in 6a8dd69d1c; `FIXTURE_ROUTING_TABLE_ENTRIES` reads as the reference value on its own.

rs/nns/governance/unreleased_changelog.md
mraszyk: Keeping `MergeSubnets`. @alin-at-dfinity made the case [here](https://github.com/dfinity/ic/pull/11366#discussion_r3932067362) that "merging canister ranges" is the right vocabulary for *this* registry mutation, which is one step of the larger process, and the doc comments now say up front that the source subnet survives and is cleaned up by a separate proposal -- which was the substance of your objection. Happy to revisit if the operator-facing naming confuses proposal reviewers.

rs/registry/canister/src/mutations/merge_subnets.rs
mraszyk: Made this explicit in the test comment in 6a8dd69d1c: "A subnet hosting no canister ID range is not a realistic scenario; this only exercises that code path."

rs/registry/canister/unreleased_changelog.md
```diff
@@ -21,6 +21,12 @@ on the process that this file is part of, see
   `HostosVersion` accept, so until now, it was possible to elect a version that consumers could not read
   back out of the Registry.
 
+* `merge_subnets` endpoint, callable through a `MergeSubnets` proposal. It takes a source and a
+  destination subnet ID, and merges the canister ID ranges of the source subnet into the canister
+  ID range set of the destination subnet, i.e., the canisters hosted by the source subnet are
+  routed to the destination subnet afterwards. Only the routing table is updated: neither subnet
+  record is modified and the source subnet is not deleted.
```
alin-at-dfinity: Nit: This should also be changed to the more concise wording used elsewhere.

alin-at-dfinity (approved):

rs/registry/canister/unreleased_changelog.md
mraszyk: Done in f3b912fd06 -- the entry now uses the same wording as the `NnsFunction` doc comment and the proposal description, keeping only the registry-specific note that just the routing table is updated.

PR #11369, opened 2026-08-28
feat(sns): Add upgrade options for dapp canisters.

daniel-wong-dfinity-org-twin:
SNS root already supported this from when we added support to NNS. This just adds support to SNS Governance.

In particular, the following are added:

- wasm_memory_persistence - This is the more interesting one, because it allows SNSs to have canisters written in Motoko. More precisely, the allows them to take advantage of Enhanced Orthogonal Persistence.
- skip_pre_upgrade - This could be useful in an emergency, which will hopefully never happen 🤞

This is tested in another (not yet merged PR): https://github.com/dfinity/ic/pull/11379

### DIFF
--- a/rs/sns/governance/src/governance.rs
+++ b/rs/sns/governance/src/governance.rs
@@ -113,7 +113,6 @@ use ic_nervous_system_lock::acquire;
 use ic_nervous_system_root::change_canister::ChangeCanisterRequest;
 use ic_nervous_system_timestamp::format_timestamp_for_humans;
 use ic_nns_constants::LEDGER_CANISTER_ID as NNS_LEDGER_CANISTER_ID;
-use ic_protobuf::types::v1::CanisterInstallMode as CanisterInstallModeProto;
 use ic_protobuf::types::v1::WasmMemoryPersistence as WasmMemoryPersistenceProto;
 use ic_sns_governance_proposal_criticality::ProposalCriticality;
 use ic_sns_governance_token_valuation::Valuation;
@@ -2678,7 +2677,12 @@ impl Governance {
             ));
         }
 
-        let mode = upgrade.mode_or_upgrade() as i32;
+        let mode = upgrade.mode_or_upgrade();
+        let mode = CanisterInstallMode::try_from(mode)?;
+
+        let canister_upgrade_options =
+            valid_canister_upgrade_options(mode, upgrade.canister_upgrade_options)
+                .map_err(|err| GovernanceError::new_with_message(ErrorType::InvalidCommand, err))?;
 
         let wasm = Wasm::try_from(&upgrade)
             .map_err(|err| GovernanceError::new_with_message(ErrorType::InvalidCommand, err))?;
@@ -2689,7 +2693,8 @@ impl Governance {
             upgrade
                 .canister_upgrade_arg
                 .unwrap_or_else(|| Encode!().unwrap()),
-            CanisterInstallMode::try_from(CanisterInstallModeProto::try_from(mode)?)?,
+            mode,
+            canister_upgrade_options,
         )
         .await
     }
@@ -2700,6 +2705,7 @@ impl Governance {
         wasm: Wasm,
         arg: Vec<u8>,
         mode: CanisterInstallMode,
+        canister_upgrade_options: Option<CanisterUpgradeOptions>,
     ) -> Result<(), GovernanceError> {
         // Serialize upgrade.
         let payload = {
@@ -2710,7 +2716,7 @@ impl Governance {
             // For more details, please refer to the comments above the (definition of the)
             // stop_before_installing field in ChangeCanisterRequest.
             let stop_before_installing = true;
-            let mode = CanisterInstallModeV2::from(mode);
+            let mode = assemble_mode(mode, canister_upgrade_options);
 
             let mut change_canister_arg =
                 ChangeCanisterRequest::new(stop_before_installing, mode, canister_id)
@@ -2886,6 +2892,11 @@ impl Governance {
                     Wasm::Bytes(target_wasm.clone()),
                     Encode!().unwrap(),
                     CanisterInstallMode::Upgrade,
+                    // upgrade options. skip_pre_upgrade would be dangerous, and
+                    // wasm_memory_persistence is not needed either, because no
+                    // SNS framework canister is written in Motoko (as of
+                    // August, 2026).
+                    None,
                 )
                 .await?;
             }
@@ -2953,6 +2964,11 @@ impl Governance {
                     Wasm::Bytes(target_wasm.clone()),
                     Encode!().unwrap(),
                     CanisterInstallMode::Upgrade,
+                    // upgrade options. skip_pre_upgrade would be dangerous, and
+                    // wasm_memory_persistence is not needed either, because no
+                    // SNS framework canister is written in Motoko (as of
+                    // August, 2026).
+                    None,
                 )
                 .await?;
             }
@@ -3155,6 +3171,7 @@ impl Governance {
             Wasm::Bytes(ledger_wasm),
             ledger_upgrade_arg,
             CanisterInstallMode::Upgrade,
+            None, // upgrade options
         )
         .await?;
 
--- a/rs/sns/governance/src/governance.rs
+++ b/rs/sns/governance/src/governance.rs
@@ -70,7 +70,7 @@ use crate::{
             proposal::Action,
             proposal_data::ActionAuxiliary as ActionAuxiliaryPb,
             transfer_sns_treasury_funds::TransferFrom,
-            upgrade_journal_entry, valuation,
+            upgrade_journal_entry, upgrade_sns_controlled_canister, valuation,
         },
     },
     proposal::{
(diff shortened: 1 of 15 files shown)


bjoernek (commented):
Automated correctness review of this diff. The port of NNS's `install_code.rs` logic into SNS governance looks faithful: validation, `assemble_mode`, enum conversion, and the proto/did/api/conversions wiring all line up; `CanisterInstallModeV2::from(Upgrade) == Upgrade(None)` so there is no behavior change for existing proposals; all 5 `upgrade_non_root_canister` call sites are updated; and SNS root already forwards `CanisterInstallModeV2` to both `install_code` and `install_chunked_code`. Three findings below.

rs/sns/governance/src/proposal.rs
```diff
@@ -1067,6 +1069,21 @@ async fn validate_and_render_upgrade_sns_controlled_canister(
...
+    // Inspect canister_upgrade_options.
+    match RootCanisterInstallMode::try_from(mode) {
+        Err(err) => {
+            // This would happen if mode was originally 0, Unspecified. Of
+            // course, if canister_upgrade_options is Some, that's also wrong
+            // here, but moot, so we do not add to defects for that.
+            defects.push(format!("Invalid mode: {err}"));
+        }
+        Ok(mode) => {
+            if let Err(err) = valid_canister_upgrade_options(mode, *canister_upgrade_options) {
```
bjoernek, line 1117: **HIGH: `canister_upgrade_options` is validated here but never rendered in the proposal text voters see.**

`validate_and_render_upgrade_sns_controlled_canister` destructures and checks `canister_upgrade_options`, but the final `format!` block (proposal.rs:1154) only emits target canister, Wasm info, `## Mode`, and argument info.

Concrete scenario: a proposer submits an `UpgradeSnsControlledCanister` proposal with `canister_upgrade_options = Some({ wasm_memory_persistence: Replace })` against a Motoko/EOP dapp canister. `Replace` reinitializes the canister's main memory, i.e. total state loss for an enhanced-orthogonal-persistence canister; `skip_pre_upgrade: true` similarly skips state serialization. The rendered `payload_text_rendering` is byte-identical to a benign upgrade proposal, so voters on this simple-majority (non-critical) topic cannot see what they are approving.

NNS deliberately surfaces this: `rs/nns/governance/src/proposals/install_code.rs:328` adds a `canister_upgrade_options` field to the `SelfDescribingValue`, with a dedicated `From<CanisterUpgradeOptionsPb>` impl that spells out `skip_pre_upgrade` and decodes `wasm_memory_persistence` to `"Keep"`/`"Replace"`. Suggest doing the same in the rendering here.

rs/sns/governance/src/proposal.rs
```diff
@@ -1067,6 +1069,21 @@ async fn validate_and_render_upgrade_sns_controlled_canister(
     // Assume mode is the default if it is not set
     let mode = upgrade.mode_or_upgrade();
 
+    // Inspect canister_upgrade_options.
+    match RootCanisterInstallMode::try_from(mode) {
+        Err(err) => {
+            // This would happen if mode was originally 0, Unspecified. Of
+            // course, if canister_upgrade_options is Some, that's also wrong
+            // here, but moot, so we do not add to defects for that.
+            defects.push(format!("Invalid mode: {err}"));
```
bjoernek, line 1114: **LOW: unhelpful defect message for `mode: Some(0)`.**

This `Err` arm is reachable only when `mode == Some(0)` (`mode_or_upgrade()` maps any other invalid value to `Upgrade`, and `Some(100)` etc. is already caught by the check just above). In that case the error is `CanisterInstallModeError((item as i32).to_string())` (`rs/types/management_canister_types/src/lib.rs:2099`), whose `Display` is just the raw code, so the proposer sees `Invalid mode: 0`.

NNS returns `"Unspecified or unrecognized install mode"` for the same input (`rs/nns/governance/src/proposals/install_code.rs:71`). Suggest a literal message here instead of interpolating `err`.

rs/sns/governance/src/governance.rs
```diff
@@ -6579,6 +6597,75 @@ impl TimeWarp {
...
+/// Converts canister_upgrade_options into the type required by SNS Root.
+///
+/// Returns Err in the following cases:
+///
+/// 1. mode is not Upgrade.
+/// 2. wasm_memory_persistence is not one of the allowed values:
+///    a. Keep
+///    b. Replace
+///    c. None
+pub(crate) fn valid_canister_upgrade_options(
```
bjoernek, line 6630: **MEDIUM: none of the new logic is exercised by a test.**

`valid_canister_upgrade_options`, `convert_i32_to_wasm_memory_persistence`, and `assemble_mode` are never called with a `Some(...)` value anywhere in this PR: every one of the ~15 touched test sites just adds `canister_upgrade_options: None`.

So nothing covers: rejection when `mode != Upgrade`; rejection of `WasmMemoryPersistence::Unspecified` (code 0); acceptance of `Keep`/`Replace`; or that the option actually reaches `ChangeCanisterRequest.mode` as `Upgrade(Some(..))`. An inverted Keep/Replace mapping, or a dropped option somewhere along the pb -> api -> root path, would ship silently.

NNS has direct tests for exactly these cases (`rs/nns/governance/src/proposals/install_code.rs:660-790`). The PR title ("Wait for test to merge") suggests tests are intended to land separately; worth confirming they cover the above before merge.

rs/sns/governance/src/governance.rs
```diff
@@ -6579,6 +6597,75 @@ impl TimeWarp {
     }
 }
 
+/// Combines the arguments into the Mode type required by Root (and the
+/// Management canister).
+///
+/// If mode is not Upgrade, then canister_upgrade_options does not affect the
+/// return value (but in non-release builds, this panics via debug_assert).
+fn assemble_mode(
```
bjoernek, line 6605: DAYSF / spin out, definitely not blocking this PR.

These three helpers (`assemble_mode`, `valid_canister_upgrade_options`, `convert_i32_to_wasm_memory_persistence`) are near-verbatim copies of the NNS versions in `rs/nns/governance/src/proposals/install_code.rs:88-250` -- same names, same doc comments, same `debug_assert_eq!`. That is entirely reasonable for getting the feature landed, and the port looks faithful to me; no behavior concern here.

The only thing I would note for later: when `WasmMemoryPersistence` eventually grows a third value, or the validation rules change, there are now two places to update -- three, counting the very similar i32 conversion already living in `rs/types/management_canister_types/src/lib.rs:2131-2140`. If someone wanted to fold these into something shared (`ic-nervous-system-root` seems like a natural home, since it already owns `ChangeCanisterRequest` and depends on `ic-management-canister-types-private`), a single fallible `assemble_mode(mode, options) -> Result<CanisterInstallModeV2, String>` would cover both callers and would make the "mode is not Upgrade but options are Some" state unrepresentable rather than debug-asserted.

Happy for that to be a follow-up, or never. Just flagging it so the next person doesn't assume the duplication was accidental.

rs/sns/governance/src/proposal.rs
daniel-wong-dfinity-org-twin: Yes, but it is non-obvious that only 0 can lead here. Also, I think `0` in the description is helpful enough.

If some future change makes that no longer true (unlikely), then `{err}` substitution remains faithful, because it is locked in.

rs/sns/governance/src/proposal.rs
```diff
@@ -1067,6 +1069,21 @@ async fn validate_and_render_upgrade_sns_controlled_canister(
...
+        }
+        Ok(mode) => {
+            if let Err(err) = valid_canister_upgrade_options(mode, *canister_upgrade_options) {
+                defects.push(err);
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: Björn also suggested this. I'm going to track the issue there.

rs/sns/governance/src/proposal.rs
daniel-wong-dfinity-org-twin: Done.

rs/sns/governance/src/governance.rs
```diff
@@ -6579,6 +6597,75 @@ impl TimeWarp {
...
+///
+/// 1. mode is not Upgrade.
+/// 2. wasm_memory_persistence is not one of the allowed values:
+///    a. Keep
+///    b. Replace
+///    c. None
+pub(crate) fn valid_canister_upgrade_options(
+    mode: CanisterInstallMode,
+    canister_upgrade_options: Option<upgrade_sns_controlled_canister::CanisterUpgradeOptions>,
+) -> Result<Option<CanisterUpgradeOptions>, String> {
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org-twin: I think this is the same as Björn's third feedback. I'll deal with this in his thread.

rs/sns/governance/src/governance.rs
daniel-wong-dfinity-org-twin: I'll try to do that in a follow up PR.

rs/sns/governance/src/governance.rs
daniel-wong-dfinity-org-twin: Done.

bjoernek (approved):

PR #11374, opened 2026-08-31
test(cketh): derive e2e sweep parameters from get_minter_info and re-send a stale authorization

gregorydemay:
Follow-ups to #11363:

* The e2e tests no longer hard-code sweep parameters. The sweeper address and the per-token minimum deposit amounts are retrieved via `get_minter_info` (#11240).
* A new test runs two deposit flows for the same deposit address. It shows that re-sending the EIP-7702 authorization with a stale nonce is harmless: the second sweep lands and the funds are credited exactly once.
* The live e2e tests now read as fluent action-expectation chains, matching `cketh.rs` and `ckerc20.rs`. State-changing steps (funding the fee account, upgrading the minter) and minter endpoint calls (`deposit_erc20`) are visible in each test instead of being buried in the test fixtures.

https://claude.ai/code/session_01YK3ggw1xtE3HZTGGBu66oD


rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -179,287 +175,248 @@ fn should_revert_the_whole_call_when_a_token_is_not_a_contract() {
...
+        }])
+        .expect_deposit_responses();
+    assert_eq!(
+        second_registrations[0].address, address,
+        "re-registering the pair must yield the same deposit address"
+    );
+    assert_matches!(
+        setup.await_detection(owner, DEPOSIT_SUBACCOUNT, &usdc).status,
+        DepositStatus::AwaitingSweep(detected) if detected.scanned_balance == second_deposits[0].amount
+    );
```
mbjorkqvist, line 393: Is there a potential race condition here that could make the test flaky? Running `setup.call_minter_deposit_erc20` arms the balance check, but we only wait for the mint before doing so, and a pair only leaves the sweep queue on finalization - the two are different timers. So I believe it could happen that `detected.scanned_balance` could be the amount from the `first_deposits`. If so, IIUC, one way to fix it would be to change `setup.await_detection` to also take the amount, and to poll until a detection of this amount is received.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Good catch, and it is actually a bit worse than a stale balance: `deposit_erc20` returns the existing status whenever the pair is still in the sweep queue, so it would not re-arm the pair at all and the second deposit would never be scanned. Polling `await_detection` for the new amount alone would therefore time out rather than fix it.

Fixed in a36588bb1c by waiting for the first sweep's successful `FinalizedSweeperTransaction` event (new `expect_sweeps_finalized(1)` in `live.rs`) before re-registering the pair, mirroring how the funding flow already waits for its finalization. Once that entry is gone, the first `AwaitingSweep` the test observes is the fresh detection, so the balance assertion stays as is.

gregorydemay:
🤖 Flakiness check on a36588bb1c: ran `//rs/ethereum/cketh/minter:deposit_from_cex` 10 times (`--runs_per_test=10 --local_test_jobs=1 --cache_test_results=no`, so every run executed all 7 tests against a fresh live PocketIC + anvil, one at a time).

| Runs | Tests per run | Failures | Duration per run |
|---|---|---|---|
| 10/10 passed | 7 | 0 | 198s–200s |

The per-run durations are tightly clustered, so the tick-driven waits (including the new `expect_sweeps_finalized` before re-registering the pair) are settling on the same schedule each time rather than burning spare ticks.

PR #11378, opened 2026-08-31
chore: deduplicate test-strategy crate

nmattia:
This streamlines the versions of `test-strategy` used, collapsing them to a single crate.

### DIFF
--- a/rs/execution_environment/Cargo.toml
+++ b/rs/execution_environment/Cargo.toml
@@ -86,7 +86,7 @@ proptest = { workspace = true }
 regex-lite = { workspace = true }
 rstest = { workspace = true }
 serde_bytes = { workspace = true }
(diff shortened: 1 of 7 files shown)


rs/execution_environment/Cargo.toml
```diff
@@ -86,7 +86,7 @@ proptest = { workspace = true }
 regex-lite = { workspace = true }
 rstest = { workspace = true }
 serde_bytes = { workspace = true }
-test-strategy = "0.3.1"
+test-strategy = "0.4.0"
```
basvandijk: What about moving this to the workspace such that we don't have to repeat this in `rs/messaging/Cargo.toml` below?

basvandijk (approved):

PR #11379, opened 2026-08-31
feat(sns): Added upgrade options to sns cli, and end to end test.

daniel-wong-dfinity-org-twin:
In particular, added

* --skip-pre-upgrade
* --wasm-memory-persistance


rs/sns/cli/src/upgrade_sns_controlled_canister.rs
```diff
@@ -744,6 +767,31 @@ pub async fn cycles_ledger_create_canister<C: CallCanisters>(
...
+/// Returns None if skip_pre_upgrade is false and wasm_memory_persistence is
+/// None, i.e. no "special" upgrade behavior was requested.
+fn assemble_canister_upgrade_options(
+    skip_pre_upgrade: bool,
+    wasm_memory_persistence: Option<WasmMemoryPersistence>,
+) -> Option<CanisterUpgradeOptions> {
+    let has_option = skip_pre_upgrade || wasm_memory_persistence.is_some();
+    if !has_option {
+        return None;
+    }
```
bjoernek, line 781: **MEDIUM: no unit test for `assemble_canister_upgrade_options`; the `None` branch is the load-bearing one.**

The new e2e test only covers the both-set case (`skip_pre_upgrade: true`, `wasm_memory_persistence: Some(Keep)`), so three of the four flag combinations never execute this function.

The branch I would most want pinned is this early return. If `has_option` were ever computed wrongly and we returned `Some(CanisterUpgradeOptions { skip_pre_upgrade: None, wasm_memory_persistence: None })` instead, then *every* ordinary SNS dapp upgrade submitted through this CLI would carry an empty options struct. Downstream that is visible: `validate_and_render_upgrade_sns_controlled_canister` would start emitting a `## Upgrade options` section (with `skip_pre_upgrade: false`, `wasm_memory_persistence: Unspecified`) on proposals that request no special behavior at all, and the same value would be rejected outright on any non-upgrade mode. No test in this PR would notice.

`upgrade_sns_controlled_canister_with_large_wasm.rs:158-160` does now pass `false`/`None` through the CLI, so the branch is exercised incidentally, but nothing asserts the outcome.

Suggest a small table test in this crate over the four combinations, asserting `None` for (false, None), `Some { skip_pre_upgrade: Some(true), wasm_memory_persistence: None }` for (true, None), and so on. That also documents the deliberate asymmetry on the next line (false becomes `None` rather than `Some(false)`), which is currently unexplained here: the near-identical helper in `rs/registry/admin/bin/main.rs:1565` carries two comments justifying exactly these two choices (why an empty struct is not always safe to return, and why `None` rather than `Some(false)`), and this copy dropped both.

rs/nervous_system/integration_tests/tests/upgrade_sns_controlled_canister_with_options.rs
```diff
@@ -0,0 +1,282 @@
...
+    // dropped along the way.
+    let proposal::Action::UpgradeSnsControlledCanister(UpgradeSnsControlledCanister {
+        mode,
+        canister_upgrade_options,
+        ..
+    }) = action
+    else {
+        panic!("unexpected proposal action {action:?}");
+    };
+    assert_eq!(mode, Some(CanisterInstallMode::Upgrade as i32));
```
bjoernek: **LOW: this reads a proto enum field through the management-canister enum; it passes only because the two happen to number identically.**

`mode` here is the field on the api type, whose values are `ic_protobuf::types::v1::CanisterInstallMode` codes, but `CanisterInstallMode` in this file is `ic_management_canister_types_private` (line 2). The assertion holds only because both enums put `Upgrade` at 3 (`rs/protobuf/def/types/v1/management_canister_types.proto:9` and `rs/types/management_canister_types/src/lib.rs:1940`).

That coincidence does not hold for the sibling enum, which is what makes this worth changing rather than ignoring: `ic_management_canister_types_private::WasmMemoryPersistence` has implicit discriminants (`Keep = 0`, `Replace = 1`), while the proto has `Keep = 1`, `Replace = 2`. The production code gets this right by going through the `From` impl rather than casting (`upgrade_sns_controlled_canister.rs:786`), and line 254 below correctly uses `WasmMemoryPersistenceProto::Keep as i32` for the readback. So the file already has the right import; only this line reaches across families.

Suggest importing `ic_protobuf::types::v1::CanisterInstallMode as CanisterInstallModeProto` and asserting against that, so both readback assertions are expressed in the enum the field actually holds.

bjoernek (approved):

rs/sns/cli/src/upgrade_sns_controlled_canister.rs
daniel-wong-dfinity-org-twin: Done.

rs/nervous_system/integration_tests/tests/upgrade_sns_controlled_canister_with_options.rs
daniel-wong-dfinity-org-twin: Done.

PR #11388, opened 2026-09-01
chore: bump cached

nmattia:
This aligns the crate we use in this repo with that of ic-agent (much more recent). This allows us to drop a duplicate version as well as some other deprecated crates.


basvandijk (approved):

nmattia:
This new `cached` version causes issues, I'm working on an alternative solution.

PR #11392, opened 2026-09-01
feat: replace 'cached' crate with 'lru'

nmattia:
This drops the `cached` crate, of which we only used the LRU bits. Recent versions of the `cached` crate force a dependency on wasm-bindgen when compiling to wasm (including wasm32-unknown-unknown).

Instead we use the `lru` crate directly and build some basic bookkeeping on top of it.


basvandijk (approved):
Infra code LGTM

rs/utils/mem-lru/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+/// cache.insert("a", 1);
+/// cache.insert("b", 2);
+/// assert_eq!(cache.get(&"a"), Some(1)); // hit; marks "a" most-recently-used
+/// cache.insert("c", 3); // cache is full, so this evicts "b" (least-recently-used)
+/// assert_eq!(cache.get(&"b"), None); // miss; "b" was evicted
+/// assert_eq!(cache.len(), 2);
+/// assert_eq!(cache.hits(), 1);
+/// assert_eq!(cache.misses(), 1);
+/// ```
+pub struct LruCacheWithStats<K: Eq + Hash, V> {
```
pierugo-dfinity:
```suggestion
pub struct LruCacheWithStats<K, V> {
```
I don't think the bounds are necessary here, only on the implemented methods

rs/utils/mem-lru/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        Self {
+            cache: lru::LruCache::new(max_size),
+            hits: 0,
+            misses: 0,
+        }
+    }
+
+    /// Return the value cached for `key`, marking it most-recently-used, while
+    /// recording a hit or a miss.
+    pub fn get(&mut self, key: &K) -> Option<V> {
```
pierugo-dfinity:
```suggestion
    pub fn get(&mut self, key: &K) -> Option<&V> {
```
I think it's best to stay generic and not force the cost of cloning, just return a borrow. Then the caller can choose whether they want to clone it.
It would also relax the requirements as we could remove the `V: Clone` bound.

rs/utils/mem-lru/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut cache = LruCacheWithStats::with_size(2);
+        cache.insert(1, "a");
+        cache.insert(2, "b");
+        // Touch key 1 so that key 2 becomes the least-recently-used entry.
+        assert_eq!(cache.get(&1), Some("a"));
+        cache.insert(3, "c"); // evicts key 2
+        assert_eq!(cache.get(&2), None);
+        assert_eq!(cache.get(&1), Some("a"));
+        assert_eq!(cache.get(&3), Some("c"));
+        assert_eq!(cache.len(), 2);
```
pierugo-dfinity, line 122:
```suggestion
        assert_eq!(cache.hits(), 3);
        assert_eq!(cache.misses(), 1);
        assert_eq!(cache.len(), 2);
```

pierugo-dfinity (approved):

pierugo-dfinity:
Is it also feasible to bump the crate's version? Also fine if done separately.

nmattia:
> Is it also feasible to bump the crate's version? Also fine if done separately.

Definitely. I've used the version already pulled in transitively to avoid a big rebuild, will bump the version separately.

rs/crypto/internal/crypto_lib/bls12_381/type/src/cache.rs
```diff
@@ -62,14 +51,14 @@ impl G2PublicKeyCache {
     /// Check the cache for an already checked G2 public key
     pub(crate) fn get(&self, bytes: &[u8; G2Affine::BYTES]) -> Option<G2Affine> {
         let mut cache = self.cache.lock();
-        cache.cache_get(bytes).cloned()
+        cache.get_cloned(bytes)
```
pierugo-dfinity, line 54: This could have stayed unchanged (`cache.get(bytes).cloned()`) but ok

PR #11395, opened 2026-09-01
feat: drop 'thousands' crate

nmattia:
This drops the [thousands](https://crates.io/crates/thousands) crate which is unmaintained (last release 7 years ago). The behavior is replicated in a new crate in `rs/utils/thousands`.


gregorydemay (approved):

basvandijk (approved):

rs/utils/thousands/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn separate_with_underscores<T: Display>(value: T) -> String {
+    let s = value.to_string();
+    // Locate the first run of ASCII decimal digits (skipping any sign prefix).
+    let start = s.find(|c: char| c.is_ascii_digit()).unwrap_or(s.len());
+    let end = s[start..]
+        .find(|c: char| !c.is_ascii_digit())
+        .map_or(s.len(), |i| start + i);
+    let digit_count = end - start;
+
+    let mut result = String::with_capacity(s.len() + digit_count / 3);
```
pierugo-dfinity:
```suggestion
    let mut result = String::with_capacity(s.len() + (digit_count - 1) / 3);
```
Extra nit: I think this is supposed to be like this (a string with 3 digits has 0 underscores).
But keep it like that. Easier to read and avoids a potential underflow

rs/utils/thousands/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        );
+        assert_eq!(
+            separate_with_underscores(u128::MAX),
+            "340_282_366_920_938_463_463_374_607_431_768_211_455"
+        );
+    }
+
+    #[test]
+    fn leaves_sign_prefix_and_fraction_untouched() {
+        // Only the first run of digits is grouped, matching `thousands`.
```
pierugo-dfinity:
```suggestion
        // Only the first run of digits is grouped.
```

We probably forgot we were replacing a crate at this point

rs/utils/thousands/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        assert_eq!(separate_with_underscores(-1234567_i64), "-1_234_567");
+        assert_eq!(separate_with_underscores(1234.5_f64), "1_234.5");
```
pierugo-dfinity:
```suggestion
        assert_eq!(separate_with_underscores(1234.5678_f64), "1_234.5678");
```

pierugo-dfinity (approved):

PR #11400, opened 2026-09-01
feat(core): Limit accessed heap pages per message

michael-weigelt:
This PR paves the way for large heaps! 

Currently, the heap is limited to 4GiB for 32bit canisters and 6GiB for 64bit canisters. We want to raise the limit for 64bit to much more, but the amount of heap that can be accessed in a single message has to remain small, because the nodes' RAM is limited. 

This PR uses the deterministic memory tracker stats to bound the number of memory pages accessible during a single message. The current limit is 6GiB, which is the full canister heap, so there is no behavioural change to the protocol. 
Later, we can raise the memory limit and keep this pages-per-message limit as is, such that the reasonable limits remain in place but the overall memory a canister can access is vastly bigger.

### DIFF
--- a/rs/embedders/src/wasm_utils/instrumentation.rs
+++ b/rs/embedders/src/wasm_utils/instrumentation.rs
@@ -21,6 +21,12 @@
 //! counter overflows, the value of the counter is the initial value minus the
 //! sum of cost of all executed instructions.
 //!
+//! There is a second check at each reentrant block: Whether the current message
+//! has exceeded the maximum number of accessed heap pages. It does so by checking
+//! the injected global "canister pending_trap_code", which is written to by the
+//! deterministic memory tracker if the limit is exceeded. In that case, the execution
+//! is aborted by calling "internal_trap".
+//!
 //! In more details, first, it inserts up to five System API functions:
 //!
 //! ```wasm
@@ -35,8 +41,19 @@
 //! It then inserts (and exports) a global mutable counter:
 //! ```wasm
 //! (global (;0;) (mut i64) (i64.const 0))
-//! (export "canister counter_instructions" (global 0)))
+//! (export "canister counter_instructions" (global 0))
+//! (export "canister counter_dirty_pages" (global 1))
+//! (export "canister counter_accessed_pages" (global 2))
+//! (export "canister pending_trap_code" (global 3))
 //! ```
+//!
+//! The first counts instructions; the next two count dirty and accessed pages
+//! in the stable memory. The last holds an `InternalErrorCode` to be raised at
+//! the next reentrant basic block, or zero if no trap is pending. Unlike the
+//! counters it is an i32, so that its value can be passed straight to
+//! `internal_trap`, and it is excluded from the globals persisted in the
+//! execution state (see `get_exported_globals`), so it starts at zero on every
+//! instance.
 //!
 //! An additional function is also inserted to handle updates to the instruction
 //! counter for bulk memory instructions whose cost can only be determined at
@@ -73,21 +90,29 @@
 //! global.set 0
 //! ```
 //!
-//! and a decrementation with a counter overflow check at the beginning of every
-//! reentrant block (a function or a loop body):
+//! and a decrementation followed by a pending trap check and a counter overflow
+//! check at the beginning of every reentrant block (a function or a loop body):
 //!
 //! ```wasm
 //! global.get 0
 //! i64.const 8
 //! i64.sub
 //! global.set 0
+//! global.get 3
+//! if  ;; label = @1
+//!   global.get 3
+//!   call 3           # the `internal_trap` function
+//! end
 //! global.get 0
 //! i64.const 0
 //! i64.lt_s
 //! if  ;; label = @1
-//!   (call x)
+//!   call 0           # the `out_of_instructions` function
 //! end
 //! ```
+//!
+//! The pending trap code is checked before the instruction counter so
+//! that a memory limit violation traps before an out of instructions trap.
 //!
 //! Before every bulk memory operation, a call is made to the function which
 //! will decrement the instruction counter by the "size" argument of the bulk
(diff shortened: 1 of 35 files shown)


rs/embedders/src/wasmtime_embedder.rs
```diff
@@ -581,6 +595,12 @@ impl WasmtimeEmbedder {
             .set(&mut store, Val::I64(current_accessed_limit.get() as i64))
             .expect("Couldn't set dirty page counter global");
 
+        // The global is excluded from the persisted globals, so it is already
+        // zero (no trap pending) and only needs to be read out here.
```
schneiderstefan, line 599: Does that mean that you cannot send trap code 0 (unknown) with this method?

rs/memory_tracker/src/deterministic.rs
```diff
@@ -347,6 +364,15 @@ impl DeterministicMemoryTracker {
 
         // Charge instructions.
         (self.subtract_instruction_counter.lock())(num_os_pages * self.page_overhead);
+
+        // If the number of accessed OS pages exceeds the limit, abort the execution by
+        // setting an exported Wasm global "pending_trap_code".
+        // The instrumentation will check it (at the next reentrant block) and trap.
+        let accessed_os_pages =
+            state.accessed_wasm_pages_count.get() as u64 * OS_PAGES_PER_WASM_PAGE;
```
schneiderstefan, line 372: There is already the helper function num_accessed_os_pages

rs/execution_environment/src/execution/call_or_task/tests.rs
```diff
@@ -720,6 +720,165 @@ fn hitting_access_limit_fails_non_replicated_query() {
...
+        .install_canister(canister_id, wat::parse_str(wat).unwrap())
+        .unwrap_err();
+    result.assert_contains(
+        ErrorCode::CanisterMemoryAccessLimitExceeded,
+        &expected_heap_access_error(canister_id, &limit),
+    );
+}
+
+#[test]
+fn hitting_heap_access_limit_fails_non_replicated_query() {
```
schneiderstefan, line 860: What about replicated queries (as opposed to non-replicated queries and updates)?

rs/memory_tracker/src/tests.rs
```diff
@@ -532,3 +566,89 @@ fn deterministic_memory_tracker_correctly_count_access_and_dirty_pages(
         },
     );
 }
+
+/// Accessing exactly the limit must not abort: the default limit is the whole
+/// heap, so a message touching all of it has to keep working.
+#[rstest]
+fn access_limit_is_not_exceeded_at_the_limit(
```
schneiderstefan, line 573: Could you also do some tests with multi-accesses and read then write/write then read?

rs/embedders/src/wasm_utils/instrumentation.rs
```diff
@@ -1342,6 +1387,23 @@ fn inject_metering(
                     },
                 ]);
                 if scope == Scope::ReentrantBlockStart {
+                    // Trap if we accessed too many pages.
```
schneiderstefan, line 1390: How much code could realistically run between the last reentrant block start and the end of the message?

rs/embedders/src/wasm_utils/instrumentation.rs
```diff
@@ -1342,6 +1387,23 @@ fn inject_metering(
...
+                    elems.extend([
```
schneiderstefan, line 1391: Claude notes that this extra code adds a bit of overhead, so this PR is not fully a no-op.

rs/embedders/src/wasm_utils/validation.rs
```diff
@@ -42,11 +42,12 @@ use crate::WASM_PAGE_SIZE;
 
 /// Symbols that are reserved and cannot be exported by canisters.
 #[doc(hidden)] // pub for usage in tests
-pub const RESERVED_SYMBOLS: [&str; 6] = [
+pub const RESERVED_SYMBOLS: [&str; 7] = [
     "canister counter_instructions",
     "canister_start",
     DIRTY_PAGES_COUNTER_GLOBAL_NAME,
     ACCESSED_PAGES_COUNTER_GLOBAL_NAME,
+    PENDING_TRAP_CODE_GLOBAL_NAME,
```
schneiderstefan, line 50: I guess we have to hope nobody used that global name before?

rs/memory_tracker/src/deterministic.rs
schneiderstefan, line 372: Do we have a metric for this, i.e. can we track on grafana how close messages actually get to the limit once it's not a noop?

rs/embedders/src/wasmtime_embedder.rs
```diff
@@ -634,6 +654,38 @@ impl WasmtimeEmbedder {
...
+        //
+        // SAFETY: We store a raw pointer to the Store and a copy of the Global.
+        // These remain valid for the lifetime of the WasmtimeInstance because:
+        // 1. The Store is pinned (heap-allocated) and owned by WasmtimeInstance
+        // 2. The Global is a lightweight handle that references data in the Store
+        // 3. The memory tracker (which holds this closure) is also owned by WasmtimeInstance
+        // 4. All are dropped together when WasmtimeInstance is dropped
+        let abort_execution: Arc<SignalMutex<dyn FnMut(AbortReason) + Send>> = {
+            // StorePtr::new requires a Pin, enforcing that the store is pinned.
+            let mut store_ptr = StorePtr::new(store.as_mut());
```
schneiderstefan, line 669: I am not sure I fully understand the thing claude flagged, so I just post it here verbatim:

>  A second `StorePtr::new(store.as_mut())` reborrow is created for the abort closure; under Rust's aliasing model this invalidates the raw pointer captured by `subtract_instruction_counter` at line 638, which is then still used from the signal handler.

rs/embedders/src/wasm_utils/instrumentation.rs
michael-weigelt: Yes, we instrument code differently now, which has an impact on the efficiency of canister code. However, we don't charge instructions for this new instrumentation (confirmed by tests). So while we do more work, from the user's perspective nothing should change.

rs/embedders/src/wasm_utils/validation.rs
michael-weigelt: Yes

rs/embedders/src/wasm_utils/instrumentation.rs
schneiderstefan: Except the cost of compilation?

PR #11401, opened 2026-09-01
feat: Introduce separate replica version

frankdavid:
There used to be a single version: the one under `/opt/ic/share/version.txt`, read by the orchestrator and passed to the replica. During fast GuestOS upgrade, the replica binaries are hot-swapped via an overlay while the node keeps running the old GuestOS. The base GuestOS version and the replica version must be distinguished.

- **`PlatformVersion`** (new struct in `ic-types`, next to `ReplicaVersion`) holds both `guestos_version` (the GuestOS the node booted from) and `replica_version` (the running replica binary). `replica_version: ReplicaVersion` in `ReplicaConfig` is replaced by `platform_version: PlatformVersion`.
- **Build**: the GuestOS image now ships a second version artifact, `/opt/ic/share/replica_version.txt` (plus its `-test` variant). During a fast upgrade, the sysext overlay will shadow it with the post-upgrade version, while `version.txt` — now explicitly the GuestOS version — is never shadowed.
- **Orchestrator**: reads both files (`--replica-version-file` → replica version, `--guestos-version-file` → GuestOS version) and passes them when starting the replica process.
- **Observability**: `/api/v2/status` gains an optional `guestos_version` field alongside `impl_version`.
- **Metrics**: `custom-metrics.sh` now also exports a replica_version metric.


ic-os/components/guestos/ic-replica.service
```diff
@@ -20,7 +20,7 @@ User=ic-replica
 
 Environment=RUST_BACKTRACE=1
 Environment=RUST_MIN_STACK=8192000
-ExecStart=/opt/ic/bin/orchestrator --replica-binary-dir /var/lib/ic/data/images --cup-dir /var/lib/ic/data/cups --replica-config-file /run/ic-node/config/ic.json5 --ic-boundary-env-file /opt/ic/share/ic-boundary.env --ic-gateway-env-file /opt/ic/share/ic-gateway.env --enable-provisional-registration --ic-binary-directory /opt/ic/bin --orchestrator-data-directory /var/lib/ic/data/orchestrator --version-file /opt/ic/share/version.txt
+ExecStart=/opt/ic/bin/orchestrator --replica-binary-dir /var/lib/ic/data/images --cup-dir /var/lib/ic/data/cups --replica-config-file /run/ic-node/config/ic.json5 --ic-boundary-env-file /opt/ic/share/ic-boundary.env --ic-gateway-env-file /opt/ic/share/ic-gateway.env --enable-provisional-registration --ic-binary-directory /opt/ic/bin --orchestrator-data-directory /var/lib/ic/data/orchestrator --version-file /opt/ic/share/binary_version.txt --guestos-version-file /opt/ic/share/version.txt
```
pierugo-dfinity: Should we rename `--version-file` to `--binary-version-file` or actually `--replica-version-file` if we use `replica_version` instead of `binary_version` throughout the code?

Should we also rename `version.txt` to `guestos_version.txt`?

rs/consensus/mocks/src/lib.rs
```diff
@@ -176,11 +176,18 @@ impl DependenciesBuilder {
...
             pool_config,
-            replica_config: ReplicaConfig {
-                node_id: node_test_id(0),
-                subnet_id: subnet_records[0].1,
-                replica_version: ReplicaVersion::from_str(&subnet_records[0].2.replica_version_id)
-                    .expect("Invalid replica_version_id"),
+            replica_config: {
+                let replica_version =
+                    ReplicaVersion::from_str(&subnet_records[0].2.replica_version_id)
+                        .expect("Invalid replica_version_id");
```
pierugo-dfinity: style nit: we can move these out of `Self` and avoid an unnecessary code block

rs/types/types/src/replica_config.rs
```diff
@@ -10,5 +10,5 @@ pub const SUBNET_ID_DEFAULT: u64 = 0;
 pub struct ReplicaConfig {
     pub node_id: NodeId,
     pub subnet_id: SubnetId,
-    pub replica_version: ReplicaVersion,
+    pub platform_version: PlatformVersion,
 }
```
pierugo-dfinity, line 14: Most users of `platform_version` are actually interested in `platform_version.replica_version`. Maybe we could add
```rust
impl ReplicaConfig {
    fn replica_version(&self) -> &ReplicaVersion {
        &self.platform_version.replica_version
    }
}
```
(and a similar `fn guestos_version`)

rs/orchestrator/src/orchestrator.rs
```diff
@@ -136,9 +136,14 @@ impl Orchestrator {
         let metrics_registry = MetricsRegistry::global();
         let replica_version = load_version_from_file(&logger, &args.version_file)
             .map_err(|()| OrchestratorInstantiationError::VersionFileError)?;
+        let guestos_version = load_version_from_file(&logger, &args.guestos_version_file)
+            .map_err(|()| OrchestratorInstantiationError::VersionFileError)?;
         info!(
             logger,
-            "Orchestrator started: version={}, config={:?}", replica_version, config
+            "Orchestrator started: version={}, guestos_version={}, config={:?}",
```
pierugo-dfinity:
```suggestion
            "Orchestrator started: replica_version={}, guestos_version={}, config={:?}",
```

rs/orchestrator/src/orchestrator.rs
```diff
@@ -301,6 +306,7 @@ impl Orchestrator {
                 cup_provider,
                 Arc::clone(&subnet_assignment),
                 replica_version.clone(),
+                guestos_version,
```
pierugo-dfinity: We could pass a `PlatformVersion` directly

rs/replica/bin/replica/main.rs
pierugo-dfinity, line 216: We could also report the GuestOS version as a metric

Bownairo (approved):

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -892,7 +894,7 @@ mod tests {
                     })
                     .unwrap(),
                 subnet_id: replica_config.subnet_id,
-                replica_version: replica_config.replica_version,
+                platform_version: replica_config.platform_version.clone(),
```
pierugo-dfinity: Do we need the clone?

rs/consensus/src/consensus/block_maker.rs
```diff
@@ -1019,7 +1021,7 @@ mod tests {
...
```
pierugo-dfinity: Do we need the clone?

rs/http_endpoints/public/src/lib.rs
```diff
@@ -386,13 +386,13 @@ pub fn start_server(
...
         state_reader.clone(),
-        replica_version.clone(),
+        platform_version.clone(),
     );
     let dashboard_router = DashboardService::new_router(
         config.clone(),
         subnet_type,
         state_reader.clone(),
-        replica_version,
+        platform_version.replica_version,
```
pierugo-dfinity: I think it's a good opportunity to display the GuestOS version on the dashboard too, WDYT?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -403,7 +403,11 @@ impl CanisterHttpPoolManagerImpl {
                             content_hash: ic_types::crypto::crypto_hash(&response),
                             content_size: response.content.count_bytes() as u32,
                             is_reject: response.content.is_reject(),
-                            replica_version: self.replica_config.replica_version.clone(),
+                            replica_version: self
+                                .replica_config
+                                .platform_version
+                                .replica_version
+                                .clone(),
```
```suggestion
                            replica_version: self
                                .replica_config
                                .replica_version()
                                .clone(),
```
Same a few occurrences below

rs/types/types_test_utils/src/ids.rs
```diff
@@ -158,6 +158,13 @@ pub fn test_replica_version() -> ReplicaVersion {
     ReplicaVersion::from_str("cafebabe0000ffff0000ffff0000ffff0000ffff").unwrap()
 }
 
+pub fn test_platform_version() -> PlatformVersion {
+    PlatformVersion {
+        guestos_version: test_replica_version(),
+        replica_version: test_replica_version(),
```
pierugo-dfinity: I think to stay as generic as possible, we could have different versions here. Such that tests do not implicitly depend on the versions being the same. WDYT?

rs/orchestrator/src/orchestrator.rs
pierugo-dfinity, line 243: We can also report the GuestOS version

pierugo-dfinity: I think it would be a good idea to report the GuestOS version in the orchestrator's dashboard as well.

pierugo-dfinity: In your PoC, how do you handle API BN and unassigned node upgrades? I have a feeling we should pass `guestos_version` here instead.

rs/replay/src/validator.rs
```diff
@@ -129,7 +129,10 @@ impl ReplayValidator {
         let replica_cfg = ReplicaConfig {
             node_id,
             subnet_id,
-            replica_version,
+            platform_version: PlatformVersion {
+                guestos_version: replica_version.clone(),
+                replica_version,
+            },
```
pierugo-dfinity: This can go in for now, but we might have to think a bit more on how we should replay blocks if a subnet stalls in the middle of a fast upgrade. Maybe these should be configurable to be different.

rs/replica/bin/replica/main.rs
```diff
@@ -187,15 +187,28 @@ fn main() -> io::Result<()> {
...
             "version info for the internet computer replica running.",
-            &["ic_active_version", "ic_replica_binary_hash"],
+            &[
+                "ic_active_version",
+                "ic_guestos_version",
+                "ic_replica_binary_hash",
+            ],
         );
         g.with_label_values(&[
             replica_version.as_ref(),
```
pierugo-dfinity:
```suggestion
            platform_version.replica_version.as_ref(),
```
(and remove the `.clone()` above)
Consistency

rs/replica/src/setup_ic_stack.rs
```diff
@@ -134,6 +134,7 @@ pub fn construct_ic_stack(
     // This is the first object that is required for the creation of the IC stack. Initializing the
     // persistent consensus pool is the only way for retrieving the height of the last CUP and/or
     // certification.
+    let replica_version = platform_version.replica_version.clone();
```
pierugo-dfinity:
```suggestion
```
nit: I would simply inline `&platform_version.replica_version` at the use-cases (there are really just 2 of them)
\+ the clone is not necessary

rs/http_endpoints/public/src/lib.rs
frankdavid: How frequently are the nodes polled? Because an average, a node will have different guestos & replica versions for 15min/week.

pierugo-dfinity: Not sure they're polled on that endpoint since it's HTML, I think it's more for humans. And I think it's a useful piece of information for debugging and/or incident response.

frankdavid: I misunderstood your point, I thought you meant the NNS dashboard. Fixed.

rs/consensus/src/consensus/block_maker.rs
frankdavid: Actually no.

rs/consensus/src/consensus/block_maker.rs
frankdavid: Removed

rs/orchestrator/src/orchestrator.rs
frankdavid: 4c2fcc8dfcbbcc984cc98c49aba4d494781b00d7

rs/http_endpoints/public/src/lib.rs
```diff
@@ -386,13 +386,13 @@ pub fn start_server(
...
         state_reader.clone(),
-        replica_version.clone(),
+        platform_version.clone(),
     );
     let dashboard_router = DashboardService::new_router(
         config.clone(),
         subnet_type,
         state_reader.clone(),
-        replica_version,
+        platform_version.clone(),
```
pierugo-dfinity:
```suggestion
        platform_version,
```

rs/replay/src/validator.rs
frankdavid: Does the guestos version matter for a replay? The assumption is that the observable behavior of the new replica on the old guestos is equivalent to the new replica on the new guestos, so the guestos version does not matter.

ic-os/components/guestos/ic-replica.service
frankdavid: Renamed --version-file to --replica-version-file. I'd like to keep `version.txt` for now. It's the version of the image/GuestOS and people may be used to seeing that file. We could rename it later but that will have a larger blast radius and this PR is already big.

rs/orchestrator/src/orchestrator.rs
frankdavid: We don't do quick upgrades on those nodes. However, if we did, passing replica version to BoundaryNodeManager would be the only correct thing to do. The version is used for 1) logging and 2) it ends up in IcBoundaryProcess::get_version. If we ever decided to do fast upgrades to boundary nodes, we'd need the binary version here. For now, passing the binary version is still correct (it's always == with guestos version).

frankdavid: One thing is the naming, IcBoundaryProcess should not have a "replica_version" but this name was there already before, we can still rename it later if it causes confusion.

rs/replay/src/validator.rs
pierugo-dfinity: The replay's validator (which is a production `Validator` from `ic_consensus`) might depend on `replica_config.guestos_version`. Now, I expect that only the payload builder and maybe the "upgrade-permit client" should end up depending on it, and the validator shouldn't. But who knows... maybe there's a weird bug that would make validators depend on the GuestOS version. 

Since replaying is used in unexpected recovery situations, we would be happy to be able to configure the GuestOS version if needed.

rs/orchestrator/src/orchestrator.rs
pierugo-dfinity: > We don't do quick upgrades on those nodes.

Exactly. So I would expect the upgrade loop to compare its `guestos_version` with whatever is in the registry. If it compared its `replica_version`, then a previously assigned node that already fast-upgraded but not upgraded its GuestOS yet could think that it matches the version for unassigned nodes even though its GuestOS part hasn't upgraded yet.

Thus, I think that until we do want to fast-upgrade API BNs, we should pass `guestos_version` here. Such that the orchestrator doesn't try to run `ic-boundary` and the behaviour would match the comments & logs.

pierugo-dfinity: Reopening because I don't think this was addressed

ic-os/components/guestos/ic-replica.service
pierugo-dfinity: Okok sounds good

ic-os/components/monitoring/guestos/custom-metrics.sh
```diff
@@ -77,6 +94,7 @@ function update_tee_metrics() {
 
 function main() {
     update_guestos_version_metric
+    update_binary_version_metric
     update_guestos_boot_action_metric
     update_config_version_metric
     update_tee_metrics
```
pierugo-dfinity, line 100: Should we stay consistent and replace all `binary_version` with `replica_version`? This also applies to `metrics-proxy.yaml` and `ic-os/defs.bzl`

rs/replay/src/validator.rs
frankdavid: 5aa994bed3e597d42371542423b04f16fe6529a1

rs/orchestrator/src/metrics.rs
```diff
@@ -68,7 +68,7 @@ impl OrchestratorMetrics {
             orchestrator_info: metrics_registry.int_gauge_vec(
                 "ic_orchestrator_info",
                 "version info for the internet computer orchestrator running.",
```
pierugo-dfinity:
```suggestion
                "Replica & GuestOS versions info for the running Internet Computer orchestrator",
```

rs/replay/src/cmd.rs
```diff
@@ -46,6 +46,10 @@ pub struct ReplayToolArgs {
     /// Whether or not to skip prompts for user input.
     pub skip_prompts: bool,
 
+    #[clap(long)]
+    /// The GuestOS version to report; defaults to the replica version.
```
pierugo-dfinity:
```suggestion
    /// The GuestOS version the validator uses; defaults to the replica version.
```

rs/types/types/src/replica_version.rs
```diff
@@ -102,3 +102,18 @@ mod test {
...
+/// upgrade, binaries from the target (new) GuestOS are hot-swapped in the
+/// running (old) GuestOS, so the replica version is ahead of the GuestOS
+/// version until the node reboots into the target GuestOS.
+#[derive(Clone, Eq, PartialEq, Debug, Deserialize, Serialize)]
+pub struct PlatformVersion {
+    /// The GuestOS version the node booted from.
+    pub guestos_version: ReplicaVersion,
+    /// The replica binary version, possibly hot-swapped by a fast upgrade.
+    pub replica_version: ReplicaVersion,
+}
```
pierugo-dfinity, line 119: I'm starting to lean towards introducing `type GuestosVersion = VersionOf<GuestosVersionTag>` and `type ReplicaVersion = VersionOf<ReplicaVersionTag>`, i.e. have them share same semantics (same methods, same representation in memory) while keeping the two types strictly different from one another.
I think this should be more robust on the long-run.

But this would probably deserve its own PR.

frankdavid: Let's think about this later. I'm not yet convinced that there is a considerable benefit (besides some type safety of not mixing them up). Otherwise the version formats are the same and we compare them (we'd need to implement comparison between ReplicaVersion and GuestosVersion).

To me a version is like `Age`, we don't really separate types for `UserAge` and `UsersCatAge`, we can compare `Age` objects directly.

rs/orchestrator/src/orchestrator.rs
frankdavid: Addressed now.

ic-os/components/monitoring/guestos/custom-metrics.sh
frankdavid: As per our discussion, let's use replica_version everywhere.

rs/types/types/src/replica_version.rs
pierugo-dfinity: > we compare them

Yeah that's the thing, I think this could be a footgun. If we're not careful enough, we might end up comparing unrelated things (just like how we discussed passing `guestos_version` vs `replica_version` in `BoundaryNodeManager`: they behave differently).

> Let's think about this later.

👍

pierugo-dfinity (approved):

NikolaMilosa (approved):

PR #11402, opened 2026-09-01
fix(execution): Recompute the subnet available memory before draining subnet queues

mraszyk:
The memory consumed by an inner round iteration's parallel execution threads is not subtracted from `SchedulerRoundLimits::subnet_available_memory`; it is recomputed from the state at the beginning of the next iteration instead.

Move that recomputation to the very beginning of the iteration and apply it to `SchedulerRoundLimits`, so that both the subnet messages and the canister executions of the iteration start from a value that reflects the committed state. As a side effect, subnet-message allocations are now deducted in full from the per-thread memory budget of the canister executions in the same iteration, rather than only 1/`scheduler_cores` of them (which is what recomputing the scaled value after the drain amounted to). This is more conservative, and matches what the first iteration has always done.

### DIFF
--- a/rs/execution_environment/src/scheduler.rs
+++ b/rs/execution_environment/src/scheduler.rs
@@ -435,6 +435,18 @@ impl SchedulerImpl {
         //      - Update the ingress history with the resulting ingress statuses.
         //      - Induct messages on the same subnet.
         let mut state = loop {
+            // In every iteration after the first, recompute the subnet available
+            // memory from the state.
+            //
+            // The memory consumed by the previous iteration's parallel execution
+            // threads is not subtracted from `scheduler_round_limits` (see
+            // `execute_canisters_in_inner_round()`), so this recomputation is what
+            // brings the subnet available memory back in sync with the state.
+            if !is_first_iteration {
+                scheduler_round_limits.subnet_available_memory =
+                    self.exec_env.scaled_subnet_available_memory(&state);
+            }
+
             // Execute subnet messages.
             // If new messages are inducted into the subnet input queues,
             // they are processed until the subnet messages' instruction limit is reached.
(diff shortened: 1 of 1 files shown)


schneiderstefan (approved):

rs/execution_environment/src/scheduler.rs
```diff
@@ -435,6 +435,18 @@ impl SchedulerImpl {
         //      - Update the ingress history with the resulting ingress statuses.
         //      - Induct messages on the same subnet.
         let mut state = loop {
+            // In every iteration after the first, recompute the subnet available
+            // memory from the state.
```
alin-at-dfinity, line 439: It is no longer "in every iteration after the first".

rs/execution_environment/src/scheduler.rs
```diff
@@ -754,8 +759,9 @@ impl SchedulerImpl {
         // Deduct all created callbacks from the available callbacks limit. This is a
         // pessimistic estimate, as it ignores any closed callbacks.
         round_limits.subnet_available_callbacks -= callbacks_created;
-        // `subnet_available_memory` will be recomputed at the beginning of the next
-        // iteration.
+        // The memory consumed by the threads is not deducted from
+        // `subnet_available_memory`; it is recomputed from the state at the very
+        // beginning of the next iteration instead (see `inner_round()`).
```
alin-at-dfinity, line 764: There's already a comment that says the same thing at the top of the loop. The previous comment (specifically pointing out why we don't also update `subnet_available_memory` when we do `subnet_available_callbacks`) was much more useful IMHO.

alin-at-dfinity (commented):
A couple of late comments.

rs/execution_environment/src/scheduler.rs
mraszyk: There's still this if-condition: `if !is_first_iteration`

rs/execution_environment/src/scheduler.rs
mraszyk: > There's already a comment that says the same thing at the top of the loop.

True. But I'd keep them both: (i) here to contrast with `subnet_available_callbacks`; (ii) above to justify why we do the recomputation.

> The previous comment (specifically pointing out why we don't also update subnet_available_memory when we do subnet_available_callbacks) was much more useful IMHO.

This PR didn't really remove any comment.

alin-at-dfinity: It did replace a previous comment. One that was specifically written to contrast with `subnet_available_callbacks`:
```rust
subnet_available_callbacks -= callbacks_created;
// `subnet_available_memory` will be recomputed at the beginning of the next
// iteration.
```
IMHO the new comment, in addition to being more (redundantly) verbose, also reads out of context: instead of "do A1; A2 is done elsewhere" it's now "do A; something that is unrelated to A but has an effect on some B requires B to be done elsewhere (see vague reference)".

PR #11403, opened 2026-09-01
feat(registry): Allow advancing standard engine replica version after a roll back.

daniel-wong-dfinity-org-twin:
Previously, if you rolled back by reducing deployment_progress to 0.0, the only change you would be allowed to do was increase deployment_progress, which is really not what you after you have just rolled back!

What this does is relax the rules about what you are allowed to do from there.

In particular, now, when deployment_progress is 0.0, you are allowed to change new_replica_version_id (and deployment_progress). old_replica_version_id must remain fixed.

This is very much like how before, after you reach 100%, you can then do a new new_replica_version_id. This is very much like that, because at 0%, _the fleet is not split_. That's the real constraint, which we didn't correctly recognize before.

### DIFF
--- a/rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
+++ b/rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
@@ -450,6 +450,39 @@ mod tests {
         // The assertion is at the top: #[should_panic(...)].
     }
 
+    /// The counterpart of the previous test, but for the other way that the
+    /// fleet can end up on a unique replica version: rolling all the way back
+    /// instead of all the way forward.
+    #[test]
+    #[should_panic(expected = "cannot skip a version")]
+    fn should_panic_when_skipping_a_version_after_rolling_back() {
+        // Step 1: Prepare the world. Fully roll back v1 -> v2. That is, all
+        // engines end up (back) on v1, and v2 gets abandoned.
+        let mut registry = REGISTRY.clone();
+        registry.do_update_standard_engine_replica_version(
+            UpdateStandardEngineReplicaVersionPayload {
+                new_replica_version_id: "v2".into(),
+                old_replica_version_id: "v1".into(),
+                deployment_progress: 0.0,
+            },
+        );
+
+        // Step 2: Run the code under test. The new old_replica_version_id
+        // ("v2") is the version that was just rolled back away from. The
+        // version that engines are actually running is v1, so this deployment
+        // would skip over it.
+        registry.do_update_standard_engine_replica_version(
+            UpdateStandardEngineReplicaVersionPayload {
+                new_replica_version_id: "v3".into(),
+                old_replica_version_id: "v2".into(),
+                deployment_progress: 0.1,
+            },
+        );
+
+        // Step 3: Verify result(s).
+        // The assertion is at the top: #[should_panic(...)].
+    }
+
     #[test]
     #[should_panic(expected = "Using a version that isn't elected.")]
     fn should_panic_if_new_version_not_elected() {
--- a/rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
+++ b/rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
@@ -75,9 +75,12 @@ impl Registry {
             1.0 => &old_record.new_replica_version_id,
             0.0 => &old_record.old_replica_version_id,
             _ => panic!(
-                "Invalid standard engine replica version transition: the fleet is \
-                 currently split. Current record: {old_record:?}. Proposed new \
-                 record: {new_record:?}.",
+                "Invalid StandardEngineReplicaVersionRecord transition: the fleet is \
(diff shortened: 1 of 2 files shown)


rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
```diff
@@ -321,7 +326,77 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "the current deployment is still in progress")]
+    fn should_allow_starting_a_different_deployment_once_fully_rolled_back() {
```
bjoernek, line 332: Could we also get the negative counterpart of this test? The new `0.0` arm feeds into the "cannot skip a version" `assert_eq!` below, but I don't think any test exercises that combination: starting from `(old: v1, new: v2, progress: 0.0)`, proposing `(old: v2, new: v3, progress: 0.1)` should be rejected, because after a full roll back no engine is running v2 any more.

`should_panic_when_skipping_a_version` covers the same assert via the `1.0` arm only, so as it stands the rolled-back skip path is the one piece of new logic without a test.

rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
```diff
@@ -68,21 +69,25 @@ impl Registry {
...
-                 record: {old_record:?}. Proposed new record: {new_record:?}.",
-            );
-        }
+        // Starting a new deployment is only allowed once the current one has
+        // fully settled onto one of its two versions.
+        let currently_running_replica_version_id = match old_record.deployment_progress {
+            1.0 => &old_record.new_replica_version_id,
+            0.0 => &old_record.old_replica_version_id,
+            _ => panic!(
+                "Invalid standard engine replica version transition: the fleet is \
```
bjoernek: This message loses the detail an operator needs to act on it. These panics surface as a failed proposal execution, so it helps to restate the rule and the offending value, e.g.

```
"Invalid StandardEngineReplicaVersionRecord transition: the fleet is currently \
 split; deployment_progress must be exactly 0.0 or 1.0 to start a new \
 deployment, but it is {}. Current record: {old_record:?}. Proposed new \
 record: {new_record:?}.",
 old_record.deployment_progress,
```

Related: this message says "Invalid standard engine replica version transition" while the `assert_eq!` right below still says "Invalid StandardEngineReplicaVersionRecord transition" - worth picking one.

Also, tiny wording thing on the doc comment at the top (line 27): "when the fleet is not *previously* split" reads a bit oddly; I think "not *currently* split" is what's meant.

bjoernek (commented):
The reasoning here looks right to me, and a nice side effect is that it makes the registry self-consistent: `do_split_subnet.rs:295-296` already treats `0.0 < deployment_progress < 1.0` as "the fleet is split", so before this change the two mutations disagreed about what "settled" means. I also checked that nothing else encoded the old "must be 1.0" rule: `UpdateStandardEngineReplicaVersion::validate()` on the Governance side only does static checks, and the invariant checker doesn't look at transitions, so there is nothing stale left behind. Changelog entry is there and accurate.

Two small comments inline (one missing test, one panic message). Neither is blocking.

Non-blocking follow-up idea: the "is the fleet split / what is the unique version" predicate now exists twice, computed two different ways (here as a `match`, and inline in `do_split_subnet.rs:295-296`). Since this file already has `StandardEngineReplicaVersionRecordExt`, a shared `fn unique_replica_version_id(&self) -> Option<&String>` there (or in `ic_protobuf::registry::standard_engine_replica_version`) would let `do_split_subnet` express `fleet_is_split` as `.is_none()` and keep the two from drifting apart again. Out of scope for this PR, just flagging it.

bjoernek (approved):

rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
daniel-wong-dfinity-org-twin: re "currently": I think that's ambiguous. During the operation, the state is in limbo. It's kind of like Schödinger's cat; it could go either way. It's only unambiguous BEFORE and AFTER the op where things stand.

daniel-wong-dfinity-org-twin: Changed the error message, per your suggestions.

rs/registry/canister/src/mutations/do_update_standard_engine_replica_version.rs
daniel-wong-dfinity-org-twin: Done.

daniel-wong-dfinity-org-twin:
I might do the split/unique deduping later. Claude already wrote it for me, but now that I look at it, it hardly seems like spinning out "pulls its own weight", at least not in my estimation. It's just way too easy to look at deployment_progress and realize that it's either 0.0 or 1.0. I'll consider doing it when I come back, but I haven't made up my mind, which I guess you would be ok with since, you do not seem to be insisting on it.

PR #11406, opened 2026-09-01
chore(ckdoge): DEFI-2987: Proposal to raise the ckDOGE ledger archive trigger threshold

mbjorkqvist:
Add the proposal to raise the archive trigger threshold on the ckDOGE ledger.


gregorydemay (approved):
Thanks!

PR #11407, opened 2026-09-01
feat(ledger-suite-orchestrator): forward ledger upgrade argument to managed ledgers

gregorydemay:
## Why

* Awaiting `archive_blocks` before replying makes archiving a commit point on the ledger transaction path. A trap there turns a committed transaction into a reject. The ck minters read that as "the mint did not happen" and retry, risking a double mint (DEFI-2967).
* The agreed stopgap is to disable archiving by raising the ledgers' `trigger_threshold` via an upgrade argument. This was already done for ckBTC via per-canister proposal 143757. The ckEVM ledger suites are only upgradable through the ledger suite orchestrator.

## What

* The orchestrator upgrade argument can now carry an optional ledger upgrade argument, currently limited to `change_archive_options`. It is forwarded to every managed ledger when the ledgers are upgraded.
* It requires the ledger wasm hash to be set in the same upgrade argument. Validation rejects it otherwise. The hash may be any registered version: the DEFI-2986 proposals will pass the currently-installed one, but the argument can also accompany a real version bump.
* Index and archive canisters keep the empty upgrade argument.
* Upgrade tasks persisted by the currently deployed orchestrator still deserialize.
* Integration tests cover stopping archiving via a raised threshold, a fresh suite without a spawned archive canister, and the validation.

https://claude.ai/code/session_01S568wBREAARUA9GvGP5JMP

### DIFF
--- a/rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
+++ b/rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
@@ -1,6 +1,7 @@
 use crate::scheduler::Erc20Token;
 use crate::state::{Canister, Canisters};
 use candid::{CandidType, Deserialize, Nat, Principal};
+use ic_icrc1_ledger::ChangeArchiveOptions;
 use std::fmt::{Display, Formatter};
 
 #[allow(clippy::large_enum_variant)]
@@ -26,6 +27,7 @@ pub struct UpgradeArg {
     pub archive_compressed_wasm_hash: Option<String>,
     pub cycles_management: Option<UpdateCyclesManagement>,
     pub manage_ledger_suites: Option<Vec<InstalledLedgerSuite>>,
+    pub ledger_upgrade_arg: Option<LedgerUpgradeArg>,
 }
 
 impl UpgradeArg {
@@ -33,6 +35,20 @@ impl UpgradeArg {
         self.ledger_compressed_wasm_hash.is_some()
             || self.index_compressed_wasm_hash.is_some()
             || self.archive_compressed_wasm_hash.is_some()
+    }
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Default, CandidType, Deserialize)]
+pub struct LedgerUpgradeArg {
+    pub change_archive_options: Option<ChangeArchiveOptions>,
+}
+
+impl From<LedgerUpgradeArg> for ic_icrc1_ledger::UpgradeArgs {
+    fn from(arg: LedgerUpgradeArg) -> Self {
+        ic_icrc1_ledger::UpgradeArgs {
+            change_archive_options: arg.change_archive_options,
+            ..ic_icrc1_ledger::UpgradeArgs::default()
+        }
     }
 }
 
--- a/rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
+++ b/rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
@@ -45,8 +45,11 @@ pub struct LedgerUpgradeArg {
 
 impl From<LedgerUpgradeArg> for ic_icrc1_ledger::UpgradeArgs {
     fn from(arg: LedgerUpgradeArg) -> Self {
+        let LedgerUpgradeArg {
+            change_archive_options,
+        } = arg;
         ic_icrc1_ledger::UpgradeArgs {
-            change_archive_options: arg.change_archive_options,
+            change_archive_options,
             ..ic_icrc1_ledger::UpgradeArgs::default()
         }
     }
--- a/rs/ethereum/ledger-suite-orchestrator/src/scheduler/mod.rs
+++ b/rs/ethereum/ledger-suite-orchestrator/src/scheduler/mod.rs
@@ -366,11 +366,18 @@ impl UpgradeLedgerSuiteSubtask {
                 let canisters = read_state(|s| s.managed_canisters(token_id).cloned())
                     .ok_or(UpgradeLedgerSuiteError::TokenNotFound(token_id.clone()))?;
                 let canister_id = ensure_canister_is_installed(token_id, canisters.index)?;
-                upgrade_canister::<Index, _>(canister_id, compressed_wasm_hash, runtime).await
+                upgrade_canister::<Index, _>(
+                    canister_id,
+                    compressed_wasm_hash,
+                    encode_empty_arg(),
+                    runtime,
+                )
+                .await
             }
             UpgradeLedgerSuiteSubtask::UpgradeLedger {
                 token_id,
                 compressed_wasm_hash,
+                ledger_upgrade_arg,
             } => {
                 log!(
                     INFO,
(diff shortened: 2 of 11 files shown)


rs/ethereum/ledger-suite-orchestrator/tests/tests.rs
```diff
@@ -671,6 +691,81 @@ mod upgrade {
...
+            },
+        );
+        orchestrator.advance_time_for_upgrade();
+
+        let managed_canisters = orchestrator
+            .assert_managed_canisters(&usdc_erc20_contract())
+            .assert_ledger_has_wasm_hash(&embedded_ledger_wasm_hash);
+        managed_canisters.make_transfers(10 * ARCHIVE_TRIGGER_THRESHOLD);
+
+        assert_eq!(managed_canisters.call_ledger_archives(), archives_before);
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The premise is not quite right: the assertion compares the full `Vec<ArchiveInfo>`, and `ArchiveInfo` includes `block_range_start`/`block_range_end`. If archiving had continued into the existing archive, its `block_range_end` would grow and the equality would fail, so the test already catches both a new archive spawning and blocks flowing into the existing one.

That said, a positive check that the ledger keeps serving new blocks itself is worthwhile — it mirrors how the mainnet upgrades will be verified (resident block count climbing). Added in 44b3c50a58: after the post-upgrade transfers, `icrc3_get_blocks` over the new range must return every block locally with `archived_blocks` empty.

rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
```diff
@@ -36,6 +38,20 @@ impl UpgradeArg {
...
+pub struct LedgerUpgradeArg {
+    pub change_archive_options: Option<ChangeArchiveOptions>,
+}
+
+impl From<LedgerUpgradeArg> for ic_icrc1_ledger::UpgradeArgs {
+    fn from(arg: LedgerUpgradeArg) -> Self {
+        ic_icrc1_ledger::UpgradeArgs {
+            change_archive_options: arg.change_archive_options,
+            ..ic_icrc1_ledger::UpgradeArgs::default()
+        }
```
mbjorkqvist, line 54: Would it make sense to destructure `LedgerUpgradeArg` here, before setting the `change_archive_options` field? That way, if we later added a field to `LedgerUpgradeArg`, the compiler would catch it if we forgot to update this implementation.

rs/ethereum/ledger-suite-orchestrator/src/scheduler/mod.rs
```diff
@@ -318,13 +277,52 @@ impl UpgradeLedgerSuiteBuilder {
...
+    ///
+    /// The order of the subtasks is as follows:
+    /// 1. Upgrade the index canister
+    /// 2. Upgrade the ledger canister
+    /// 3. Fetch the list of archives from the ledger and upgrade all archive canisters
+    ///
+    /// For each canister, upgrading involves 3 (potentially failing) steps:
+    /// 1. Stop the canister
+    /// 2. Upgrade the canister
+    /// 3. Start the canister
```
mbjorkqvist, line 292: Not directly related to this comment, but the affected code didn't change, so putting the comment here anyway:

If the (ledger) upgrade fails, the error is treated as recoverable (`is_recoverable() == true`), and we retry it every `RETRY_FREQUENCY` which is 5 seconds. This isn't new, but what is new is the upgrade args that we pass, and that it may be easier to get wrong, in which case the ledger may panic in the `post_upgrade`. Would it make sense to add a counter for the number of retries we do, and if we exhaust them, e.g., bump a metric so we can trigger an alert? Since it's not directly related to this PR, doing it in a separate ticket/PR would also be fine with me.

mbjorkqvist (approved):
Thanks @gregorydemay!

A couple of comments that don't affect changes code, so I put them here:

- `ArchiveOptions` for freshly installed suites is [hardcoded](https://github.com/dfinity/ic/blob/34bafc0828a62267a0e6f8ac838d7f5e35ed786e/rs/ethereum/ledger-suite-orchestrator/src/scheduler/mod.rs#L958) with `trigger_threshold: 2_000`. I'm not sure if/how we'd want to change that - changing the hardcoded constant here or in another PR, taking it as an argument when adding a new token (probably not)?
- We don't have observability into these settings - no endpoint that exposes them, not logging or metrics. So to verify if a change in the archive `trigger_threshold` was effective, we'd need to make enough transactions to pass the new/old value. For the numerical values I was thinking of adding them to the metrics (as part of DEFI-1565), which would help here, WDYT?

mbjorkqvist (approved):
The PR description (which is perhaps a bit too verbal), says that:

> upgrading each ledger with the same wasm

which (to me) suggests that we would perform an upgrade with the same wasm that the ledger is currently running (that's what we typically do when we only want to change some configuration parameters). I don't think this is currently checked/enforced anywhere, but maybe we don't even want this restriction, since it may complicate some (recovery/bugfix) scenarios in the future.

rs/ethereum/ledger-suite-orchestrator/src/candid/mod.rs
gregorydemay: 🤖 Good point — done in 17a00faa4a: the `From` impl now destructures `LedgerUpgradeArg`, so adding a field to it without updating the conversion fails to compile.

rs/ethereum/ledger-suite-orchestrator/src/scheduler/mod.rs
gregorydemay: 🤖 Agreed, and the situation is actually worse than noisy retries: a failing install returns before the start step, so the ledger is left *stopped* while the task retries every 5s — a persistently failing upgrade is a silent outage for that token.

Two notes:
- An alertable signal exists today without code changes: `orchestrator_tasks_duration_seconds{task_name="upgrade_ledger_suite", result="err"}` grows at roughly one increment per 5s retry during such a loop.
- The trap surface for the one forwarded field is narrow: `change_archive_options` only traps when the ledger was initialized without archiving at all, which cannot happen for orchestrator-created ledgers (they always get archive options at init) — though suites adopted via `manage_ledger_suites` are not covered by that guarantee.

Taking you up on the separate-ticket offer: filed [DEFI-2989](https://dfinity.atlassian.net/browse/DEFI-2989) (in the DEFI-1914 epic) for capping retries, restarting the stopped canister on exhaustion, and a dedicated metric/alert.

gregorydemay:
🤖 Answering the points from the two review summaries:

**Hardcoded `trigger_threshold: 2_000` for new suites** — agreed this is a gap in the stopgap: a token added while archiving is disabled would get a fresh ledger that starts archiving at 2 000 blocks, reintroducing the commit point for that token. Bumping the hardcoded constant in a small follow-up PR seems right (restored when DEFI-2967 lands); threading it through `AddErc20Arg` would widen a public API for a temporary measure, so agreed on "probably not".

**Observability of archive options** — +1 for exposing the numerical settings as metrics in DEFI-1565. The lack is exactly why the integration tests here (and the mainnet verification of the upcoming proposals) must infer the threshold behaviorally from the resident block count; a metric would give the direct signal.

**"Same wasm" not enforced** — correct, and intentional: validation only requires *a* registered ledger wasm hash so the argument travels with a scheduled ledger upgrade. The DEFI-2986 proposals will pass the currently-installed hash, but the mechanism deliberately also works with a real version bump — e.g. restoring the threshold together with the DEFI-2967 wasm in a single upgrade. Clarified the PR description accordingly.

PR #11408, opened 2026-09-01
fix: deflake should_flag_only_deposits_at_or_above_the_per_token_minimum

basvandijk:
## Problem

`//rs/ethereum/cketh/minter:deposit_from_cex` kept flaking after #11383 (which removed the dominant stop-canister race). Both post-#11383 `master` flakes share one new signature: `should_flag_only_deposits_at_or_above_the_per_token_minimum` fails at its first assertion with

```
assertion failed: `Scanning { valid_until: …, last_scanned_block: Some(Nat(11)), scan_count: 1 }`
does not match `DepositStatus::AwaitingSweep(detected) if …`
```

right after `credit_deposits` funded the deposit addresses on anvil (its own balance assertions passed, so the funds were readable).

## Root cause

The minter's balance scan does not fetch a block number: it `eth_call`s at its **cached** `latest_block_height` (`rs/ethereum/cketh/minter/src/balance_scan/mod.rs`, pinned as `BlockTag::Number`), and that cache is refreshed **only** by the separate 30-second `REFRESH_LATEST_BLOCK_HEIGHT_INTERVAL` timer (`deposit.rs` is its sole writer; the log scrape writes a different field). So a scan firing after the harness mined the funding block can still read every balance at a pre-funding block: the call succeeds (the token code, set through `anvil_setCode`, is visible to historical `eth_call`s — mined transfers are not), finds nothing, and `record_scan` stores `scan_count = 1` with the stale `last_scanned_block`.

The failing logs show exactly this: an ETH log scrape covering blocks `1..=11` (head ≥ 11 pre-funding), the funding mined after it, then one balance scan — `scanned 3 (address, token) pair(s), found 0 candidate(s), 0 decode error(s), 0 call error(s)` — pinned at block 11.

The harness' `await_scan` accepted **any** `Scanning` status with `scan_count >= 1` as terminal, so it returned that transient verdict instead of driving on until a scan that could actually see the funds — and the test's `AwaitingSweep` assertion failed. (The next scan corrects the verdict within a couple of blocks; the predicate just never waited for it.)

## Fix

`await_scan` now snapshots the anvil chain head at entry (the funding is always mined by then — every caller funds through `credit_deposits` first) and counts an address as scanned only once its status is:

- `AwaitingSweep` — terminal by construction (a detection only results from a scan that read a balance at or above the minimum, and a queued pair never returns to `Scanning`), or
- `Scanning` with `scan_count >= 1` **and** `last_scanned_block` at or past that entry head — a scan pinned there cannot have missed the balance, so a below-minimum verdict is genuine.

`SCAN_TICKS` grows from 4 to 8 (a budget, not a cost — driving stops the moment the condition holds): the pair's block-based backoff can demand up to 300 block-seconds before it is due for the scan the new predicate waits for, against the ~10 blocks each tick's settle mines.

## Validation

- **Deterministic reproduction**: temporarily deploying the token code before registration (so a pre-funding scan completes, recording exactly the state the stale-cache race records — `scan_count = 1` at a block where the funding is invisible) and forcing one such scan:
  - old predicate → fails with the exact CI signature, deterministically;
  - fixed predicate → converges to `AwaitingSweep` and the test passes end to end.
- `cargo check`, `cargo fmt`, `./ci/scripts/rust-lint.sh` — clean.
- Depth-2 bazel rdeps all pass: `integration_tests_tests/{ckerc20,cketh}_test`, `minter:lib_tests`, `test_utils:lib_tests`.
- `bazel test --runs_per_test=3 --jobs=3 //rs/ethereum/cketh/minter:deposit_from_cex` — 3/3 full-suite runs pass.

The remaining non-success runs in the window since #11383 merged all ran on commits that predate it (branch/merge-queue checkouts from before the merge) and show the old, already-fixed stop-canister signature.

### DIFF
--- a/rs/ethereum/cketh/test_utils/src/live.rs
+++ b/rs/ethereum/cketh/test_utils/src/live.rs
@@ -123,8 +123,10 @@ const AWAIT_DEADLINE: Duration = Duration::from_secs(60);
 const SCAN_TICK: Duration = Duration::from_secs(BALANCE_SCAN_INTERVAL.as_secs() + 5);
 
 /// A budget, not a cost: one tick refreshes the latest block height the scan needs, the next scans;
-/// the spares cover a tick lost to an outcall the jump timed out.
-const SCAN_TICKS: u32 = 4;
+/// the spares cover a tick lost to an outcall the jump timed out, and the blocks the pair's
+/// block-based backoff gap demands before it is due again — up to 300 block-seconds (25 blocks)
+/// once an address has been scanned a few times, against the ~10 blocks each tick's settle mines.
+const SCAN_TICKS: u32 = 8;
 
 /// A budget, not a cost: the sweep path crosses the enqueue, send, and finalization timers, and
 /// driving stops the moment the expected transactions are on chain.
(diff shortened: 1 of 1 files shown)


mbjorkqvist (approved):
Thanks @basvandijk!

PR #11412, opened 2026-09-01
fix: deflake //rs/tests/consensus/upgrade:upgrade_downgrade_unassigned_nodes_test

basvandijk:
This PR slightly rewrites the upgrade system tests code to reuse duplicated constants and sets the same timeout to all of them. In particular, `upgrade_downgrade_unassigned_nodes_test` now has the same timeout as its siblings `upgrade_downgrade_{app,nns}_subnet_test`. This deflakes the former as it was occasionally timing out.

👨 Revised by @pierugo-dfinity

### DIFF
--- a/rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
+++ b/rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
@@ -22,11 +22,11 @@ Success::
 
 end::catalog[] */
 
-use std::time::Duration;
-
 use anyhow::Result;
 use anyhow::bail;
-use ic_consensus_system_test_upgrade_common::elect_target_version;
+use ic_consensus_system_test_upgrade_common::{
+    UP_DOWNGRADE_OVERALL_TIMEOUT, UP_DOWNGRADE_PER_TEST_TIMEOUT, elect_target_version,
+};
 use ic_consensus_system_test_utils::{
     rw_message::install_nns_and_check_progress,
     upgrade::{deploy_guestos_to_all_unassigned_nodes, fetch_unassigned_node_version},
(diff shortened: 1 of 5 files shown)


rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
```diff
@@ -40,6 +42,15 @@ use ic_types::ReplicaVersion;
...
 use slog::info;
 
+// The test performs two sequential GuestOS upgrades of the unassigned node,
+// each of which is polled with a 600s retry budget (see
+// `upgrade_unassigned_nodes` below). The driver's default per-test timeout of
+// 10 minutes cannot even cover a single exhausted poll, so give the test an
+// explicit budget that covers its own worst case (2 x 600s polling + proposal
+// handling + slack), like its siblings in this package do.
+const UP_DOWNGRADE_OVERALL_TIMEOUT: Duration = Duration::from_secs(30 * 60);
+const UP_DOWNGRADE_PER_TEST_TIMEOUT: Duration = Duration::from_secs(25 * 60);
```
pierugo-dfinity: Let's just update the `const`s in `common.rs` to 35 and 30 minutes respectively, reuse them across `upgrade_downgrade_{app_subnet,nns_subnet,unassigned_nodes}_test.rs` and get rid of the same `const`s in the latter files. Sure the unassigned nodes test should need a bit less than that but I think it's better than to have four different pairs of constants

And I don't think the verbose comment adds much value (quite the opposite, I think it distracts the reader away from what the test actually does)

pierugo-dfinity (approved):
Thanks Bas for looking into it. Maybe we could tell Claude to yap less in PR descriptions. To me, just saying
> This aligns the test with its siblings in this package (upgrade_downgrade_{app,nns}_subnet_test set 30–35 min driver timeouts and test_timeout = "eternal")

was enough of a justification, the precise timeout numbers don't matter much. The *Note* on an independent bug and the whole *Validation* paragraph could also be trimmed out.

rs/tests/consensus/upgrade/upgrade_downgrade_unassigned_nodes_test.rs
pierugo-dfinity: https://github.com/dfinity/ic/pull/11412/commits/8761942152b0162780fbcd129cb5f59baedf13a9

PR #11415, opened 2026-09-02
test: Temporarily disable vetkeys in upgrade downgrade tests

eichhorl:
The upgrade-downgrade tests are currently flaky due to a change in vetkeys signatures. See root cause analysis here: https://github.com/dfinity/ic/pull/11410

Making production backward compatible (like the PR above suggests) is not necessary, since replicas are not expected to validate artifacts of different replica versions. Therefore, this PR simply disables vetkeys in upgrade downgrade tests, until the version containing https://github.com/dfinity/ic/pull/11333 is rolled out to all subnets.


pierugo-dfinity (approved):
> Making production backward compatible is not necessary

I agree. I guess a more robust long-term solution would be to include a `ReplicaVersion` in chain-key artifacts

PR #11416, opened 2026-09-02
refactor: Export the monotonic consumed cycles total as the existing gauge

mraszyk:
> **Stacked on #11490** (`mraszyk/backfill-consumed-cycles-monotonic`), which
> backfills the canister-level monotonic value this builds on. Review that first;
> the diff here is just the export change.
>
> Earlier parts of this stack have landed: #11465 added
> `CanisterMetrics::consumed_cycles_monotonic` (and renamed the pre-existing
> `_as_counter(s)` fields to `_monotonic`), #11464 made a subnet split refund the
> prepayment of a dropped `install_code`.

## What this changes

`replicated_state_consumed_cycles_since_replica_started` (and its
`replicated_state_consumed_cycles` alias) exported
`SubnetMetrics::consumed_cycles_total_including_canisters()`, i.e. the certified
total, which behaves like a gauge: a prepayment raises it and the matching refund
lowers it again. That makes it awkward to build monitoring on top of.

It now exports the monotonic counterpart instead: the subnet-level aggregate plus
the canisters' `consumed_cycles_monotonic`. The subnet-level aggregate needs no
monotonic twin of its own -- subnet-level use cases are only ever charged, never
refunded, and a deleted canister's consumption moves into
`consumed_cycles_by_deleted_canisters`.

**No new metric is added**, and the certified total is untouched:
`CanisterMetrics::consumed_cycles` and
`SubnetMetrics::consumed_cycles_total_including_canisters` are unchanged, so
`/subnet/<subnet_id>/metrics` is unaffected. The two differ by exactly the
prepayments whose refund is not known yet, i.e. the gauge tracks the certified
total closely and is simply better behaved.

## Why reuse the name, and why it stays a gauge

It stays a **gauge**, not a Prometheus `Counter`. The value is recomputed from the
state on every observation, so a replica that restarts mid-interval and resumes
from the last checkpoint re-exports a lower value than it had reached before; a
`Counter`'s `reset()` + `inc_by()` would report that as an increase by the whole
checkpoint value. An online subnet split lowers it for the same reason -- the
migrating canisters' consumption goes to the other subnet, and retaining it here
would count it on both.

So it still needs the same high water mark treatment on the consumer side that the
non-monotonic value already needed; it is just a much better behaved gauge in the
average case. And since it must be a gauge anyway, exporting the monotonic value
under the old name saves rewriting every rule and dashboard (including the public
dashboard) that already scrapes it.

## Tests

* `consumed_cycles_total_is_exported_net_of_outstanding_prepayments`: the exported
  gauge equals the certified total net of the outstanding prepayments.
* `consumed_cycles_gauge_accounts_for_all_subnet_level_use_cases` updated: it now
  covers the subnet-level contributions only, since the state it builds holds no
  canisters.

## Notes for the reviewer

* The follow-up that the backfill PR mentions still applies here: the same trick
  would work for `consumed_cycles_by_use_cases_monotonic` -- split `outstanding`
  into its `Instructions` and `RequestAndResponseTransmission` parts and it falls
  out. Left for later.
* Locally I ran the directly affected targets (`replicated_state`, `state_layout`,
  `state_manager`, `execution_environment`, `canonical_state`,
  `cycles_account_manager`, `messaging`, `replay`, `state_machine_tests`,
  `protobuf`, `types`) plus `rust-lint.sh`. The NNS/SNS/registry integration tests,
  which reach this change only through the regenerated protobuf bindings, exhaust
  the disk on my machine, and the two golden-state tests need internal SSH access;
  those are left to CI.

### DIFF
--- a/rs/execution_environment/src/scheduler.rs
+++ b/rs/execution_environment/src/scheduler.rs
@@ -11,7 +11,7 @@ use crate::execution_environment::{
 };
 use crate::ic00_permissions::Ic00MethodPermissions;
 use crate::metrics::MeasurementScope;
-use crate::util::process_responses;
+use crate::util::{debug_assert_or_critical_error, process_responses};
 use ic_config::embedders::Config as HypervisorConfig;
 use ic_config::flag_status::FlagStatus;
 use ic_config::subnet_config::SchedulerConfig;
@@ -1126,6 +1126,22 @@ impl SchedulerImpl {
 
                 // Abort all paused execution before the checkpoint.
                 abort_all_paused_executions(state, &self.exec_env, cost_schedule, &self.log);
+
+                // Backfill the `consumed_cycles_monotonic` of every canister from
+                // its `consumed_cycles` gauge, which predates it.
+                //
+                // Only done on checkpoint rounds, and only after the paused
+                // executions above have been aborted: a paused execution holds a
+                // prepayment that is not part of the replicated state, which would
+                // make the backfill overestimate the monotonic amount (see
+                // `SystemState::outstanding_prepayments`). Aborting materializes
+                // those prepayments into the canisters' task queues.
+                //
+                // Unconditional and idempotent, like
+                // `migrate_outcalls_cycles_to_use_cases` above: it is a no-op once a
+                // canister has been backfilled, and self-healing if a downgrade
+                // dropped the monotonic amount.
+                migrate_consumed_cycles_to_monotonic(state, &self.metrics, &self.log);
             }
             ExecutionRoundType::OrdinaryRound => {
                 self.abort_paused_executions_above_limit(state);
@@ -2160,4 +2176,62 @@ pub fn abort_all_paused_executions(
     for canister in canister_states.hot_values_mut() {
         abort_canister(canister, subnet_schedule, exec_env, cost_schedule, log);
     }
+}
+
+/// Backfills `CanisterMetrics::consumed_cycles_monotonic` of every canister from
+/// its `consumed_cycles` gauge, which predates it and thus holds the full history. See `SystemState::migrate_consumed_cycles_to_monotonic`.
+///
+/// Must only be called with no paused executions left (i.e. on a checkpoint round,
+/// after `abort_all_paused_executions`); a canister that still has one is skipped,
+/// as its prepayment is not part of the replicated state.
+fn migrate_consumed_cycles_to_monotonic(
+    state: &mut ReplicatedState,
+    metrics: &SchedulerMetrics,
+    log: &ReplicaLogger,
+) {
+    state.canisters_for_each_mut(|_id, canister| {
+        let canister_metrics = canister.system_state.canister_metrics();
+        // The outstanding prepayments could not be derived from the replicated
+        // state, i.e. the canister has a paused execution, whose prepayment is only
+        // held in memory. Unreachable when called as documented.
+        let Some(outstanding) = canister.system_state.outstanding_prepayments() else {
+            debug_assert_or_critical_error!(
+                false,
+                metrics.consumed_cycles_invariant_broken,
+                log,
+                "{}: Canister {}: cannot derive the monotonic consumed cycles, \
+                 unexpected task {:?}",
+                CONSUMED_CYCLES_INVARIANT_BROKEN,
+                canister.canister_id(),
+                canister.system_state.task_queue.paused_or_aborted_task(),
+            );
+            return;
+        };
+        let derived = canister_metrics.consumed_cycles() - outstanding;
+        match derived.cmp(&canister_metrics.consumed_cycles_monotonic()) {
+            // Not backfilled yet, or a downgrade dropped it. Only take a
+            // mutable reference here, so that this stays a read-only pass once every
+            // canister has been backfilled (`Arc::make_mut` clones the canister).
+            std::cmp::Ordering::Greater => {
+                Arc::make_mut(canister)
+                    .system_state
+                    .migrate_consumed_cycles_to_monotonic();
+            }
+            // The invariant that makes the backfill exact; see
+            // `SystemState::outstanding_prepayments`.
+            std::cmp::Ordering::Equal => {}
+            std::cmp::Ordering::Less => debug_assert_or_critical_error!(
+                false,
+                metrics.consumed_cycles_invariant_broken,
+                log,
+                "{}: Canister {}: monotonic consumed cycles {} above the gauge {} net \
+                 of the {} outstanding prepayments",
+                CONSUMED_CYCLES_INVARIANT_BROKEN,
+                canister.canister_id(),
+                canister_metrics.consumed_cycles_monotonic(),
+                canister_metrics.consumed_cycles(),
+                outstanding,
+            ),
+        }
+    });
 }
--- a/rs/replicated_state/src/canister_state/system_state.rs
+++ b/rs/replicated_state/src/canister_state/system_state.rs
@@ -346,6 +346,9 @@ impl CanisterMetrics {
     /// that [`Self::consumed_cycles`] covers, i.e. everything except HTTPS outcalls,
     /// which are only tracked at the subnet level (and, for the canister, in the
     /// by-use-case map).
+    ///
+    /// See `SystemState::outstanding_prepayments` for the invariant that ties the
+    /// two together.
     pub fn consumed_cycles_monotonic(&self) -> NominalCycles {
         self.consumed_cycles_monotonic
     }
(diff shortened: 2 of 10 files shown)


rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2237,6 +2256,144 @@ impl SystemState {
...
+    ///
+    /// Only `Instructions` and `RequestAndResponseTransmission` charges are ever
+    /// refunded (they are the only two `CyclesUseCaseRefundableKind`s) and an
+    /// outstanding prepayment of either is always recorded in the replicated state:
+    ///
+    ///  * in the `Callback` of a call whose response has not been executed yet
+    ///    (`prepayment_for_response_execution` and
+    ///    `prepayment_for_call_transmission`); or
+    ///  * in the `prepaid_execution_cycles` of a paused / aborted execution or
+    ///    `install_code`.
```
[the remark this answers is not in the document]
mraszyk: Good catch, fixed.

`drop_in_progress_management_calls_after_split()` now settles the prepayment of the dropped `AbortedInstallCode` instead of discarding it: `remove_aborted_install_code_task()` returns the `prepaid_execution_cycles` and the caller observes them as a refund of zero. That leaves the balance and both consumed-cycles gauges untouched (nothing is actually returned to the canister) and moves the prepayment onto the monotonic counters, which is where a prepayment belongs once its refund is known. The invariant is restored in the same round, and the `Instructions` by-use-case counter gets the amount too -- the checkpoint sweep alone would only have caught up the scalar.

One correction to the ordering in the report: a splitting batch returns early from `StateMachineImpl::execute_round` (`rs/messaging/src/state_machine.rs`), so the split round runs no execution and the backfill sweep never runs in it. The observable effect was as described though -- the split checkpoint would have carried a counter lagging by the dropped prepayment until a later checkpoint round bumped it. Note also that the lag was in the safe direction (the counter only ever caught *up*), so this was an exactness bug rather than a monotonicity one.

Not refunding the cycles to the canister balance is deliberate: the execution was paid for and is being discarded, and changing balances during a split is a consensus-critical semantic change that does not belong in this PR.

Covered by `settles_prepayment_of_aborted_canister_install_dropped_after_split`.

rs/replicated_state/src/metrics.rs
```diff
@@ -581,6 +597,16 @@ impl ReplicatedStateMetrics {
...
 
+        // The subnet-level aggregate is monotonic already (subnet-level use cases are
+        // only ever charged, never refunded, and a deleted canister's consumption is
+        // moved into `consumed_cycles_by_deleted_canisters`), so it can be added to
+        // the canisters' counters as-is.
+        let consumed_cycles_as_counter = state.metadata.subnet_metrics.consumed_cycles_total()
+            + consumed_cycles_by_canisters_as_counter;
+        self.consumed_cycles_as_counter.reset();
+        self.consumed_cycles_as_counter
+            .inc_by(consumed_cycles_as_counter.get() as f64);
```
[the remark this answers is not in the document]
mraszyk: The premise is right -- an online split does lower this total -- but neither suggested remedy is the one we want, so this stays as is, now with a comment explaining why.

Transferring the migrated canisters' consumption into a monotonic subnet accumulator would count it twice: subnet B keeps those canisters and their counters, so subnet A' would report consumption that is still being reported next door. It would also change `SubnetMetrics::consumed_cycles_total()`, which feeds the **certified** `/subnet/<subnet_id>/metrics` from `CertificationVersion::V29` -- not something to change as a side effect of adding a metric.

A process-local baseline would keep the exported number monotone at the cost of permanently diverging from the state it reports, and would silently mask any genuine regression in the underlying value. The `debug_assert` in `migrate_consumed_cycles_to_counters` exists precisely to catch those.

What is left is a counter reset at an online split, which Prometheus handles: `rate()`/`increase()` recognise the reset and lose one interval's increment. Worth accepting, because the drop is not an artefact -- the certified `consumed_cycles_total_including_canisters` that this metric mirrors drops for exactly the same reason, as does the existing `replicated_state_consumed_cycles_from_replica_start_as_counters`, which folds the same per-canister counters into a `CounterVec`. Making this one metric split-proof in isolation would leave it inconsistent with both.

rs/replicated_state/src/canister_state/system_state.rs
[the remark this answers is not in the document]
mraszyk: Correction to my reply above: the prepayment is now **refunded in full** rather than settled as consumed (44be0cb).

The canister has nothing to show for those cycles -- an aborted execution discards the slices it had already run and starts over when retried, and this one is never retried, while subnet A' rejects the corresponding call and unwinds the operation there too. So charging for it was wrong, and my "not refunding is deliberate" argument does not hold up: the refund is deterministic across subnet B's replicas and conserves cycles (they move back from consumed to the balance), and the split path returns early from `execute_round` before any of the cycle-conservation assertions.

Both variants settle the prepayment and so restore the invariant equally -- `G' - o' = (G-p) - (o-p) = G - o = C` -- but the refund lowers the gauges by the prepayment and adds nothing to the monotonic counters, which is what should happen when nothing was consumed.

Test renamed accordingly, and `online_split` in `rs/replicated_state/tests/replicated_state.rs` now expects the refund in its post-split state.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2237,6 +2272,148 @@ impl SystemState {
...
```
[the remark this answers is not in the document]
mraszyk: Correct, fixed. The bullet now reads "in the `prepaid_execution_cycles` of an aborted execution or an aborted `install_code`" -- `PausedExecution` has no such field.

Added a parenthetical right after the list so the paused cases are accounted for where the reader meets them, rather than only in the `Returns None` paragraph further down: a paused response execution is paid for by the callback the task carries (covered by the first bullet), and any other paused execution holds its prepayment in memory only, which is why the method returns `None` for it.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2237,6 +2272,148 @@ impl SystemState {
...
+    ///
+    /// The one place where such a prepayment leaves the state without a refund is an
+    /// aborted `install_code` dropped by a subnet split; that path settles it onto
+    /// the counters, see `Self::drop_in_progress_management_calls_after_split`.
```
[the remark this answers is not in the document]
mraszyk: Right, that paragraph was left over from the settle-as-consumed version. Reworded:

> The only way such a prepayment leaves the state other than through the refund that its execution or response issues is an aborted `install_code` dropped by a subnet split; that path refunds it in full, see `Self::drop_in_progress_management_calls_after_split`.

So it is no longer framed as an exception to the invariant -- it is just an unusual place where the refund is issued -- while still pointing a reader who wonders what happens when the task disappears at the code that handles it.

rs/execution_environment/src/scheduler.rs
```diff
@@ -2154,3 +2170,46 @@ pub fn abort_all_paused_executions(
...
+
+/// Backfills the monotonic `CanisterMetrics::consumed_cycles_as_counter` of every
+/// canister from its `consumed_cycles` gauge, which predates it and thus holds the
+/// full history. See `SystemState::migrated_consumed_cycles_as_counter`.
+///
+/// Must only be called with no paused executions left (i.e. on a checkpoint round,
+/// after `abort_all_paused_executions`); canisters that still have one are skipped,
+/// as their prepayment is not part of the replicated state.
+fn migrate_consumed_cycles_to_counters(state: &mut ReplicatedState) {
+    state.canisters_for_each_mut(|_id, canister| {
```
alin-at-dfinity: I suppose it's not ideal to subtract and add back the stats of every cold canister every round, but this is temporary code, so it's probably not worth trying to be more efficient.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -2237,6 +2272,154 @@ impl SystemState {
...
+    /// retry once paused executions have been aborted.
+    pub fn migrate_consumed_cycles_to_counter(&mut self) -> bool {
+        match self.migrated_consumed_cycles_as_counter() {
+            Some(counter) => {
+                self.canister_metrics.consumed_cycles_as_counter = counter;
+                true
+            }
+            None => false,
+        }
+    }
```
alin-at-dfinity: Sorry Claude, this is stupidly verbose. `migrated_consumed_cycles_as_counter` can be trivially inlined; and it's not used anywhere else.

rs/replicated_state/src/canister_state/system_state.rs
```diff
@@ -1656,7 +1673,23 @@ impl SystemState {
...
+        // This also keeps the consumed cycles metrics consistent. Dropping the task
+        // without refunding would leave the prepayment in the gauges with nothing left
+        // in the state to account for it, breaking the invariant on
+        // `Self::outstanding_prepayments`; and the monotonic counters, which only
+        // account for a prepayment once its refund is known, would never see it at
+        // all. A full refund lowers the gauges by the prepayment and adds nothing to
+        // the counters, which is exactly right: nothing was consumed.
+        if let Some(prepaid_execution_cycles) = self.task_queue.remove_aborted_install_code_task() {
+            self.refund_cycles(prepaid_execution_cycles, prepaid_execution_cycles);
+        }
```
alin-at-dfinity: I believe that everything after the first comment line can be safely dropped.


```suggestion
        // Refund the execution cycles prepaid for the dropped `install_code` in full.
        if let Some(prepaid_execution_cycles) = self.task_queue.remove_aborted_install_code_task() {
            self.refund_cycles(prepaid_execution_cycles, prepaid_execution_cycles);
        }
```

Better yet, the one sentence can be added to the first paragraph of the existing comment, something like "Remove aborted install code task and fully refund the [prepaid] cycles."

rs/replicated_state/src/metrics.rs
```diff
@@ -166,6 +168,15 @@ impl ReplicatedStateMetrics {
                 "replicated_state_consumed_cycles_since_replica_started",
                 "Number of cycles consumed",
             ),
+            consumed_cycles_as_counter: metrics_registry.register(
+                Counter::new(
+                    "replicated_state_consumed_cycles_since_replica_started_as_counter",
```
alin-at-dfinity: "As counter" is not a great name. It does mirror "by use cases as counters", but while the `CanisterMetrics` field does kind of act like a Prometheus counter, it is it is just a field in a struct (nothing to do with Prometheus counters).

And the actual Prometheus metric should emphatically not be a counter (as hinted to by the fact that we reset it to zero and then increment it every time): if the replica process restarts half-way through a checkpoint interval and then resumes from the checkpoint, the scraped time series will increase past the `CanisterMetrics` "counter" at the checkpoint height (say `N`), then reset back to `N` and continue from there. From the POV of Prometheus, the underlying value just jumped from `N+k` to `0` and then back to `N`, an increase of `N` cycles.

I.e. definitely make this a gauge. And ideally drop the `_as_counter` from everywhere. Just name it `_monotonic`.

Even better, there is already code that aliases the `replicated_state_consumed_cycles_since_replica_started` / `replicated_state_consumed_cycles_from_replica_start` to names that drop their misleading suffixes. Those alias names are not used anywhere (dashboards, alerts or other rules). So we could grab the chance to drop the unused aliases and export the monotonic counters cleanly as `replicated_state_consumed_cycles[_by_use_case]`.

alin-at-dfinity: And actually, since the new metric must be a gauge, we may as well avoid adding a new metric and instead simply export the new monotonic value under the old name(s).

That way we save significant effort on rules and dashboards (including the public dashboard): as said, the value is potentially not actually monotonic across replica restarts, so computing rates of increase and whatnot still requires the same high water mark approach we have for the current gauge. It's merely a much better behaved (in the average case) gauge than the existing one.

alin-at-dfinity (commented):
Logic and tests look sane.

But man, is this verbose! I'm fine with leaving any temporary migration, assert and test code (and their comments) as is, but I would try to cut down on noise for everything that's here to stay.

rs/execution_environment/src/scheduler.rs
mraszyk: Agreed, leaving it as is. Two things that soften it a bit: the sweep only runs on checkpoint rounds (it is called after `abort_all_paused_executions`, in the `CheckpointRound` arm), not every round; and it takes `Arc::make_mut` only for canisters that actually need backfilling, so once every canister has been backfilled the closure itself is read-only -- what is left is exactly the `cold_stats.sub()`/`add()` round trip that `CanisterStates::for_each_mut` does conservatively.

And, as you say, it is temporary: the whole sweep goes away once every subnet has checkpointed with the counter in place.

rs/replicated_state/src/canister_state/system_state.rs
mraszyk: Done, inlined in 2339379. `migrate_consumed_cycles_to_counter` now reads:

```rust
pub fn migrate_consumed_cycles_to_counter(&mut self) -> bool {
    let Some(outstanding) = self.outstanding_prepayments() else {
        return false;
    };
    // `max` rather than a plain assignment: the counter must never go down, not
    // even if a saturating subtraction somewhere made the gauge lag behind it.
    self.canister_metrics.consumed_cycles_as_counter = self
        .canister_metrics
        .consumed_cycles_as_counter
        .max(self.canister_metrics.consumed_cycles - outstanding);
    true
}
```

The doc comment of the removed method moved onto it, with the `Returns None` paragraph turned into the `Returns false` one.

rs/replicated_state/src/canister_state/system_state.rs
mraszyk: Done in 2339379, taking the "better yet" variant -- the sentence is now part of the existing first paragraph:

```rust
// Remove aborted install code task and fully refund the prepaid execution
// cycles.
//
// Note that this cannot be a paused install code task, because we abort all
// paused tasks before triggering the split.
if let Some(prepaid_execution_cycles) = self.task_queue.remove_aborted_install_code_task() {
    self.refund_cycles(prepaid_execution_cycles, prepaid_execution_cycles);
}
```

rs/replicated_state/src/metrics.rs
mraszyk: Both points taken, and the second one supersedes the first, so 2339379 does exactly that: **the new Prometheus metric is gone**, and the monotonic total is now exported under the existing `replicated_state_consumed_cycles_since_replica_started` gauge (and, via the existing `register_alias`, `replicated_state_consumed_cycles`), in place of the certified `consumed_cycles_total_including_canisters` it used to carry.

```rust
// The monotonic counterpart of the certified
// `SubnetMetrics::consumed_cycles_total_including_canisters()`, from which it
// differs by exactly the prepayments whose refund is not known yet: the
// subnet-level aggregate (monotonic already, ...) plus the canisters' monotonic
// counters. Exported in place of the certified total, which it tracks closely,
// so that existing rules and dashboards need no change.
//
// Still a gauge, not a Prometheus `Counter`: the value is recomputed from the
// state, so a replica resuming from a checkpoint re-exports a lower value than
// it had reached before; and an online subnet split hands the migrating
// canisters' consumption over to the other subnet (retaining it here would
// count it on both). Both need the same high water mark treatment as the
// non-monotonic value did.
self.consumed_cycles.set(
    (state.metadata.subnet_metrics.consumed_cycles_total()
        + consumed_cycles_by_canisters_as_counter)
        .get() as f64,
);
```

Your `N+k → 0 → N` argument is the decisive one and I had missed it: a `Counter` whose value is recomputed from the state on every observation is not a counter at all. The `reset()` + `inc_by()` pattern only works for a value the process itself accumulates, and this one is derived from a checkpoint the process may be resumed from. So no rules or dashboards to write, and no new metric to roll out -- just a strictly better behaved gauge under the name everything already scrapes. The help string now says so ("Monotonic, except across replica restarts and subnet splits: a prepayment is only accounted for once the refund it produces is known.").

The aliases stay as they are: since we are reusing the old names rather than introducing new ones, dropping them is unrelated to this PR (and they are still the in-flight half of the earlier rename).

What I have **not** done here is the `_as_counter` → `_monotonic` rename, in `CanisterMetrics`, the protobuf field and everywhere else. It touches the persisted field name and a good deal of surrounding code, so I would rather do it as its own commit/PR than bury it in this one -- happy to do it right away if you would prefer it in here.

Test-wise: `consumed_cycles_total_as_counter_is_exported` became `consumed_cycles_total_is_exported_net_of_outstanding_prepayments`, asserting `exported == certified_total - outstanding`; and `consumed_cycles_gauge_accounts_for_all_subnet_level_use_cases` no longer stages a canisters' part via `refresh_consumed_cycles` (the gauge folds the canisters directly now), with the canisters' part covered by the scheduler test above.

rs/replicated_state/src/metrics.rs
```diff
@@ -164,7 +164,9 @@ impl ReplicatedStateMetrics {
             ),
             consumed_cycles: metrics_registry.gauge(
                 "replicated_state_consumed_cycles_since_replica_started",
-                "Number of cycles consumed",
+                "Number of cycles consumed. Monotonic, except across replica \
+                 restarts and subnet splits: a prepayment is only accounted for \
+                 once the refund it produces is known.",
```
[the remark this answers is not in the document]
mraszyk: Right, the description was stale -- the new metric was dropped in 2339379 in favour of reusing the existing name, and the description still described the earlier revision. Updated: it now has a *"Scope, and the second behaviour change: what the existing gauge now exports"* section stating explicitly that no new Prometheus metric is added and that `replicated_state_consumed_cycles_since_replica_started` (and its `replicated_state_consumed_cycles` alias) now carries the monotonic total in place of the certified `consumed_cycles_total_including_canisters`, along with why it must stay a gauge rather than become a `Counter`.

The code is the intended state here, not the description: the reset-and-increment pattern a `Counter` would need is wrong for a value recomputed from the state on every observation, and reusing the old name spares every rule and dashboard that already scrapes it.

rs/replicated_state/src/metrics.rs
```diff
@@ -570,14 +577,24 @@ impl ReplicatedStateMetrics {
...
+        // canisters' consumption over to the other subnet (retaining it here would
+        // count it on both). Both need the same high water mark treatment as the
+        // non-monotonic value did.
         self.consumed_cycles.set(
-            state
-                .metadata
-                .subnet_metrics
-                .consumed_cycles_total_including_canisters()
+            (state.metadata.subnet_metrics.consumed_cycles_total()
+                + consumed_cycles_by_canisters_as_counter)
```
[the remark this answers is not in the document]
mraszyk: Both stale, fixed in a97772f.

`SubnetMetrics::consumed_cycles_total_including_canisters`:

```rust
/// This is the certified total: the state tree at `/subnet/<subnet_id>/metrics`
/// (from certification version `V29`) reads it here. The
/// `replicated_state_consumed_cycles_since_replica_started` gauge deliberately
/// exports the monotonic counterpart instead -- this same total, net of the
/// prepayments whose refund is not known yet -- see
/// `ReplicatedStateMetrics::observe`.
```

And the test helper, which still needs the refresh, just not for `observe` any more -- the assertions compare the exported gauge against the certified total:

```rust
/// Observes the state metrics at `height`, having first refreshed
/// `SubnetMetrics::consumed_cycles_total_including_canisters`, the derived
/// certified total that the assertions below compare the exported gauge against.
/// Production refreshes it on every `commit_and_certify`, which this harness never
/// does.
```

PR #11417, opened 2026-09-02
feat(ledger-suite-orchestrator): disable archiving for newly installed ledger suites

gregorydemay:
Stacked on #11407. Addresses the [review remark](https://github.com/dfinity/ic/pull/11407#pullrequestreview-5086956827) about the hardcoded archive options for new suites.

## Why

* A token added while archiving is disabled (DEFI-2986) would get a fresh ledger with `trigger_threshold: 2_000` and start archiving once past it, reintroducing the double-mint commit point (DEFI-2967) for that token.

## What

* Raise the default `trigger_threshold` for newly installed ledger suites to 4_200_000_000, the value used to disable archiving on the ckBTC ledger ([proposal 143757](https://dashboard.internetcomputer.org/proposal/143757)).

https://claude.ai/code/session_01S568wBREAARUA9GvGP5JMP

### DIFF
--- a/rs/ethereum/ledger-suite-orchestrator/README.adoc
+++ b/rs/ethereum/ledger-suite-orchestrator/README.adoc
@@ -73,6 +73,7 @@ To understand how ckERC20 tokens are managed, it's essential to understand the l
 
 Remarks:
 
+* Archiving is temporarily disabled for newly created ledger suites: the orchestrator sets `trigger_threshold` to `4_200_000_000` so that no archive canister is ever spawned and all blocks remain in the ledger. This is a stopgap until the minters can safely process blocks that were moved to an archive.
 * Upgrading the ledger won't upgrade any spawned archive canisters.
 * The initial version (Git hash `4472b0064d347a88649beb526214fde204f906fb`) of the ledger used doesn't use stable memory.
 * Both the index canister and archive canisters used in ckERC20 use stable memory. It's therefore possible to configure the `archive_options` in such a way that it will be very unlikely that a second archive canister is ever needed.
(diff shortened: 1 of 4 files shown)


mbjorkqvist (approved):
Thanks @gregorydemay!

A few related things we may still want to address:

- The comment in the `test_utils` [here](https://github.com/dfinity/ic/blob/52098d2b70152ce52a2fdc7bfca90645bd578a5b/rs/ethereum/ledger-suite-orchestrator/test_utils/src/pocket_ic/flow.rs#L83-L85) still mention the production value of `2_000` for the `trigger_threshold`.
- The docs [here](https://github.com/dfinity/ic/blob/52098d2b70152ce52a2fdc7bfca90645bd578a5b/rs/ethereum/ledger-suite-orchestrator/README.adoc?plain=1#L70-L72) still assume archiving will be done - a sentence could be added saying that with the current configuration, archiving is (temporarily) disabled.
- A new SNS ledger suite would still have archiving enable, as per the configuration [here](https://github.com/dfinity/ic/blob/52098d2b70152ce52a2fdc7bfca90645bd578a5b/rs/sns/init/src/lib.rs#L605). This is perhaps worth a discussion regarding if archiving in general should be disabled for SNSs also (upgrade args are currently not supported for SNS ledger suite upgrades, so this is not so straight-forward).

gregorydemay:
🤖 Thanks @mbjorkqvist! Addressed in 3c279e860e:

1. Removed the stale comment in `test_utils` mentioning the old `2_000` production value.
2. Added a remark to the README lifecycle section stating that archiving is temporarily disabled for newly created ledger suites and why.
3. SNS ledger suites: agreed this is worth a discussion, but it is out of scope for this PR. The double-mint risk motivating this change is specific to the ckEVM minters, and the SNS init configuration is owned by the NNS team.

PR #11431, opened 2026-09-03
fix: adjust each part of the response execution prepayment on its own

mraszyk:
# Summary

The cycles prepaid for a response execution are adjusted to the canister's current
Wasm execution mode before the callback is executed (#11414), so that a canister
upgraded from Wasm32 to Wasm64 (or vice versa) across a call is charged as if it had
performed the call in the Wasm execution mode in which the response is executed.

That adjustment only looked at the real part of the prepayment, i.e. the cycles
actually charged, which is zero under the **free** cost schedule. On a subnet using
that schedule the adjustment therefore did nothing, and the nominal part, which feeds
the consumed cycles metrics, stayed at the amount prepaid for the Wasm execution mode
the canister had when it performed the call. The refund of the unused instructions is
computed in the Wasm execution mode in effect at the response, so the metrics ended up
wrong in both directions.

# Impact

With a 1e9 instruction limit and 1e6 instructions executed by the callback, on a
subnet using the free cost schedule:

| Wasm execution mode at the call → at the response | reported | correct |
| --- | --- | --- |
| Wasm32 → Wasm64 | **0** | 7 000 000 |
| Wasm64 → Wasm32 | **1 006 000 000** | 6 000 000 |

- Wasm32 → Wasm64: the refund exceeds the prepayment and is capped at it, so the
  whole response execution, including the fixed per-message execution fee,
  disappears from the metrics.
- Wasm64 → Wasm32: the metrics over-report by the difference of the two prepayments
  for the full instruction limit, i.e. by an amount that scales with the instruction
  limit rather than with the instructions actually executed.

Only the metrics are affected: no cycles balance changes, since nothing is charged
under the free cost schedule to begin with.

# Fix

The adjustment now brings both the real and the nominal part of the prepayment in
line with the cost of executing the response under the cost schedule and in the Wasm
execution mode in effect at the response. Under the **normal** cost schedule, where
the two parts coincide, nothing changes: no cycle amount and no metric differs for any
subnet on that schedule. An adjustment that fails because the canister cannot cover
the shortfall still leaves the canister state unchanged, as its callers rely on.

The adjustment also no longer assumes that the prepayment and the cost at the response
carry the same cost schedule. That holds today, since no proposal changes the cost
schedule of a subnet, although the registry does not enforce it for canister-range
migrations (`prepare_canister_migration` does not compare the cost schedules of the
source and the destination subnet). A canister whose subnet's cost schedule changed
across a call now pays exactly what the cost schedule in effect at the response
requires.

The settlement of a response whose callback is not executed at all, and the refund
for the response transmission, are not changed: they were already correct.

# Testing

New tests in `rs/cycles_account_manager/tests/cycles_account_manager.rs` cover every
combination of the cost schedule and the Wasm execution mode at the call with the ones
at the response:

- an executed response is charged, and reported in the consumed cycles metrics, at
  the cost of executing it under the cost schedule and in the Wasm execution mode in
  effect at the response;
- a response whose callback is not executed is charged the fixed per-message
  execution fee only, and never more than was prepaid;
- the refund for the response transmission never exceeds the prepayment, and the
  metrics report the transmission cost;
- an adjustment that fails for lack of cycles leaves the balance and the consumed
  cycles metrics unchanged.

Reverting the fix makes the first of these fail, while the others still pass: they
cover paths that were already correct.


rs/cycles_account_manager/src/cycles_account_manager.rs
```diff
@@ -870,6 +870,40 @@ impl CyclesAccountManager {
...
+    /// on its real part, so that the adjustment applies under the free cost schedule
+    /// too: the real part of an `Instructions` amount is zero under that schedule, so
+    /// a comparison of real parts would see no difference between the Wasm execution
+    /// modes and leave the nominal part unadjusted, misreporting the consumed cycles
+    /// metrics. Under the free cost schedule nothing is withdrawn from or refunded to
+    /// the balance, hence only those metrics are adjusted; under the normal cost
+    /// schedule the real and the nominal parts coincide, so this comparison is
+    /// equivalent to comparing the real parts.
+    ///
+    /// # Assumption: the cost schedule of a subnet never changes
```
schneiderstefan: We should enforce that [here](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/registry/canister/src/mutations/prepare_canister_migration.rs?L164) then.

schneiderstefan: Hmm wait, the link I gave is about subnet splitting. We probably still want to enforce it there, but I was thinking if there was a mechanism to not allow changing the cost schedule via proposal.

rs/cycles_account_manager/tests/cycles_account_manager.rs
```diff
@@ -383,6 +384,296 @@ fn verify_no_cycles_charged_for_message_execution_on_free_schedule() {
...
+
+/// If the canister's balance does not cover the cycles missing from the
+/// prepayment, then the adjustment fails and leaves the canister state unchanged.
+/// Its callers rely on that: one replays the adjustment on a clean canister state
+/// once a paused DTS execution is resumed, the other settles the unchanged
+/// prepayment for a response that is not executed at all.
+#[test]
+fn adjust_prepayment_for_response_execution_leaves_state_unchanged_on_failure() {
+    let cycles_account_manager = cycles_account_manager();
+    let config = subnet_cycles_config(CanisterCyclesCostSchedule::Normal);
```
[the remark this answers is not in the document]
mraszyk: Good catch — the setting I picked had a shortfall in both parts, so `prepaid - required` was zero and the test could not tell the two orders apart.

The test now runs over two settings; the second is the mixed-direction one you suggested (Free/Wasm64 at the call, Normal/Wasm32 at the response): the requirement exceeds the prepayment in the real part, while the prepayment exceeds the requirement in the nominal part, so the excess to be refunded is non-zero. The prepayment is consumed up front, so both consumed cycles metrics have a non-zero baseline that the assertions pin down along with the balance.

Verified by mutation: swapping the withdrawal and the refund makes it fail with `left: (NominalCycles(1005000000), NominalCycles(0))` against `right: (NominalCycles(2005000000), NominalCycles(0))`. The previous version of the test passed with that swap in place.

rs/cycles_account_manager/src/cycles_account_manager.rs
```diff
@@ -870,6 +876,39 @@ impl CyclesAccountManager {
...
+    /// execution mode the canister had when it performed the call, misreporting the
+    /// consumed cycles metrics. Under the free cost schedule nothing is withdrawn
+    /// from or refunded to the balance, hence only those metrics are adjusted; under
+    /// the normal cost schedule the real and the nominal parts coincide.
+    ///
+    /// It also means that the adjustment does not rely on the prepayment and the
+    /// requirement carrying the same cost schedule: the prepayment recorded in the
+    /// callback carries the cost schedule in effect when the call was performed,
+    /// whereas the requirement is derived from the cost schedule in effect when the
+    /// response is executed. The two always coincide today — a subnet is assigned
```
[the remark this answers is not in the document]
mraszyk: Addressed: the PR description has been rewritten. It now describes the unconditional part-wise withdrawal and refund rather than a comparison of nominal parts, replaces the "Assumption" section with one explaining that the accounting no longer depends on the two amounts carrying the same cost schedule, and states the test matrix as the 16 combinations of (cost schedule × Wasm execution mode) at the call and at the response.

The PR is also now stacked on #11529, so its diff is just the response-execution accounting.

rs/cycles_account_manager/src/cycles_account_manager.rs
mraszyk: Resolved by dropping the assumption rather than enforcing it — thanks for pushing on this.

The adjustment no longer compares the prepayment against the requirement. It tops up
or refunds the real and the nominal part each on its own, so it does not depend on the
two amounts carrying the same cost schedule: a canister whose subnet's cost schedule
changed across a call ends up paying exactly what the cost schedule in effect at the
response requires.

I also went over the other two paths that settle a callback's prepayments, a response
whose callback is not executed and the response transmission. Under a free → normal
flip they would charge nothing real, because they only ever refund and are capped at
the prepayment, whose real part is zero under the free schedule. That is pre-existing
behavior from their no-top-up rule, not something this PR changes, and the metrics
still report the nominal cost in both cases.

mraszyk: Update: #11529 has merged, so the diff is now just the two `cycles_account_manager`
files, and the argument above is pinned by tests over every combination of the cost
schedule and the Wasm execution mode at the call and at the response, for the executed
response, the unexecuted response, the response transmission, and the failure path of
the adjustment.

On enforcing it: `validate_subnets_consistency` in `prepare_canister_migration.rs`, the
place you linked, compares the subnet types, sizes and signing keys of the source and
the destination subnet, but not their cost schedules, so a canister-range migration
between a rental subnet and a normal one is not rejected today. Adding that check is a
reasonable follow-up; it is no longer needed for this PR to be correct. (The migration
canister cannot cause a flip: it requires the canister to be stopped, which it only is
once it has no open callbacks.)

schneiderstefan (approved):

PR #11432, opened 2026-09-03
chore(cketh): disable archiving on orchestrator-managed ledgers

gregorydemay:
Updates the ledger suite orchestrator upgrade proposal text so that block archiving is effectively disabled on all ledgers managed by the orchestrator.

This is done by forwarding a ledger upgrade argument that raises the archiving trigger threshold to `4_200_000_000` blocks. The ledger Wasm module itself is not changed and remains at the version currently deployed by proposal 140856.

The proposal text is refreshed against the new orchestrator release commit, with updated Wasm hash, upgrade args hash, release notes, and verification instructions.


mbjorkqvist (approved):
Verified the new wasm hash, the upgrade args hash, and the target canister ID.

PR #11437, opened 2026-09-03
fix: pass `replica_version` when replaying for mainnet NNS Recovery

pierugo-dfinity:
Since https://github.com/dfinity/ic/pull/11087, `ic-replay` expects a `--replica-version` when replaying without a consensus pool.

The original behaviour was delivering a batch with replica version `0.9.0` which was strictly worse.


nmattia (approved):
Thank you!

eichhorl (approved):

PR #11440, opened 2026-09-03
chore: Limit port 19522 (SEV key exchange server) to be called by the peer VM only

frankdavid:


basvandijk (approved):

Bownairo (approved):

pierugo-dfinity (approved):

r-birkner (approved):
Looks good! The change will be active at the next upgrade. So, we will only see an issue at the next next upgrade.

PR #11447, opened 2026-09-04
chore: CON-1730 Use `validate_response_share` for divergence and non-flexible responses

eichhorl:
`validate_response_share` does the following for a given share:
1. Validate the callback ID
2. Ensure the claimed cycles spent is within the allowance
3. Ensure the claimed size is within the limit
4. Ensure the same signer didn't already contribute a share
5. Ensure the signer belongs to the request's committee

The function was introduced for flexible outcalls. We can also use it for non-flexible outcalls and divergence responses, which currently hand roll the same checks.


rs/https_outcalls/consensus/src/payload_builder/utils.rs
pierugo-dfinity:
```suggestion
```
I think this is now redundant with the two calls below since we check `proof.id == callback_id` and `proof.id == response.id`.

This would also mean being able to remove the `InvalidCanisterHttpPayloadReason::ShareCallbackIdMismatch` variant altogether.

GitHub again... Suggesting to remove the entire `if response_with_proof.response.id != callback_id` check.

rs/https_outcalls/consensus/src/payload_builder/utils.rs
```diff
@@ -427,7 +433,8 @@ pub(crate) fn aggregate_shares(
 /// Validates a single [`FlexibleCanisterHttpResponseWithProof`].
 ///
 /// Checks callback-id consistency, share validity (using
```
pierugo-dfinity:
```suggestion
/// Checks share validity (using
```
if applying my previous suggestion

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity, line 601: Aaah, we could almost combine this and the `validate_response_share`s below with a single call to `validate_flexible_response_with_proof`, which does the same thing.

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -613,45 +613,37 @@ impl CanisterHttpPayloadBuilderImpl {
...
             }
 
-            if valid_signers.len() < effective_threshold {
+            // Every signer belongs to the committee at this point, so all of them
+            // count towards the threshold.
+            if shares.len() < effective_threshold {
                 return invalid_artifact(InvalidCanisterHttpPayloadReason::NotEnoughSigners {
                     committee: effective_committee.into_iter().collect(),
-                    signers: valid_signers,
+                    signers: shares.iter().map(|share| share.signature.signer).collect(),
```
pierugo-dfinity: What about
```suggestion
                    signers: seen_signers.into_iter().collect(),
```
?
In that case maybe replace the `if shares.len() < effective_threshold` with `if seen_signers.len() < effective_threshold` for consistency

rs/https_outcalls/consensus/src/payload_builder.rs
```diff
@@ -699,90 +691,56 @@ impl CanisterHttpPayloadBuilderImpl {
...
-                        InvalidCanisterHttpPayloadReason::DivergenceProofDoesNotMeetDivergenceCriteria,
-                    );
-                }
+            // Defer signature verification.
+            sig_inputs.extend(response_share_sig_inputs(
+                &response.shares,
+                context.registry_version,
+            ));
+
+            if !grouped_shares_meet_divergence_criteria(&grouped_shares, faults_tolerated) {
```
pierugo-dfinity, line 740: This function is really just interested in `grouped_shares.values()` which are exactly `response.shares`. We should probably be able to pass this directly, and get rid of `grouped_shares` entirely.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -517,29 +502,25 @@ impl CanisterHttpPoolManagerImpl {
                 // single replica is allowed to consume. Free subnets charge
                 // nothing, so their spend (used only for cost accounting) may
                 // exceed the zero allowance, up to `MAX_HTTP_OUTCALL_SPEND_FREE_SUBNET`.
```
```suggestion
                // single replica is allowed to consume.
```
Feels out of place imo

pierugo-dfinity (commented):
Really nice cleanup!

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Nice, done 👍. `ShareCallbackIdMismatch` is still returned by `validate_response_share`, however

rs/https_outcalls/consensus/src/payload_builder/utils.rs
eichhorl: Done

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: Done

rs/https_outcalls/consensus/src/payload_builder.rs
eichhorl: > `grouped_shares.values()` which are exactly `response.shares`

I don't think they are the same, because `grouped_shares.values()` still returns the shares grouped by the metadata, which `grouped_shares_meet_divergence_criteria` needs, to make sure that there is indeed a divergence.

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Good point, done

rs/https_outcalls/consensus/src/payload_builder/utils.rs
pierugo-dfinity: Indeed, didn't see that for some reason

rs/https_outcalls/consensus/src/payload_builder.rs
pierugo-dfinity: > I don't think they are the same

Indeed, I don't know what I was thinking. I was probably confused by the shadowing of `grouped_shares` and thought that `grouped_shares.len()` was still `1`.

pierugo-dfinity (approved):

PR #11449, opened 2026-09-04
feat(cketh): accept and sweep plain ETH at delegated deposit addresses

gregorydemay:
[DEFI-2993](https://dfinity.atlassian.net/browse/DEFI-2993): explore whether ETH deposits from a CEX can reuse the *delegated* (EIP-7702) ERC-20 deposit address instead of requiring a separate, never-delegated per-asset address.

The proposed `CkSweeperAttested` delegate gains an empty payable `receive()` — kept empty so it fits the 2'300-gas `transfer`/`send` stipend of contract-batched CEX withdrawals — and permissionless, attestation-gated `sweepEth`/`sweepEthBatch` entry points that forward the address' ETH balance to the minter through the helper's `depositEth`.

The anvil demo test covers the full flow: a fixed 21'000-gas send works while the address has no code, fails on the sender side once delegated (nothing gets locked), and lands in `receive()` given enough gas (21'055 measured); a non-minter relayer then sweeps the ETH via both entry points. The larger delegate bytecode shifts the attested gas table slightly (re-measured constants, e2e baselines updated).

Validated on Ethereum mainnet: Binance and Kraken ETH withdrawals both reached a delegated address and were swept back — details in [DEFI-2993](https://dfinity.atlassian.net/browse/DEFI-2993).

[DEFI-2993]: https://dfinity.atlassian.net/browse/DEFI-2993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEFI-2993]: https://dfinity.atlassian.net/browse/DEFI-2993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -461,21 +461,38 @@ with Phase 2):
 | **Single shared address** with *set-and-clear* delegation (install delegate, sweep, re-delegate to `address(0)`) | One address per account | Two tECDSA signatures + ≈ 2 × 12'500–25'000 gas per sweep cycle; short window in which fixed-gas ETH transfers fail at the sender; more complex delegation lifecycle |
 | **Single shared address** with permanent delegation | One address, one authorization ever | Breaks ETH deposits entirely: plain sends to a delegated address without `receive()` revert — funds bounce at the CEX (`R12` holds, but ETH deposits are impossible) |
 
-**Mainnet data point (2026-09-04, DEFI-2993,
+**Mainnet data point (2026-09-04,
 [#11449](https://github.com/dfinity/ic/pull/11449)):** with an empty payable
-`receive()` on the delegate (kept empty to fit the 2'300-gas `transfer`/`send`
-stipend) plus attested `sweepEth`/`sweepEthBatch` entry points, real ETH
-withdrawals from **Binance** and **Kraken** both reached an
-EIP-7702-delegated address — top-level sends with gas limits of 207'128 (flat)
-and 31'830 (estimate-based) respectively, 21'055 gas used in both cases,
-matching the anvil measurement exactly — and were swept back permissionlessly.
+`receive()` on the delegate (kept minimal to fit the 2'300-gas
+`transfer`/`send` stipend) plus attested `sweepEth`/`sweepEthBatch` entry
+points, real ETH withdrawals from **Binance**
+([before delegation](https://etherscan.io/tx/0xaa13310032da7fc310f451f23d013e4b9e69c24bd201ac5cc57b54b15b0afbf2),
+[after](https://etherscan.io/tx/0x59cbb83ac893da16789b15d0494b7c3e5f382e1254f259d9c6fbcef5e45747c1))
+and **Kraken**
+([after](https://etherscan.io/tx/0x3b1dbd6c29b5f56d71c0ce1baae931f8e519852775be199fe62575ba6361b864))
+reached the delegated deposit EOA
+[`0x5e5402ed…`](https://etherscan.io/address/0x5e5402edfe155b5116f1f84d0b97b9e74f5ebc85),
+delegated to the sweeper
+[`0x3983baa1…`](https://etherscan.io/address/0x3983baa1703f69e76669768164c822918e843a25)
+— top-level sends with gas limits of 207'128 (flat) and 31'830
+(estimate-based) respectively, 21'055 gas used in both cases, matching the
+anvil measurement exactly. The funds were swept back permissionlessly through
+a private helper deployment
+[`0x3fc2640f…`](https://etherscan.io/address/0x3fc2640f8529cfafc5bf47ca2bcd05eb51d94c7b)
+to the EOA standing in for the minter,
+[`0x91eba053…`](https://etherscan.io/address/0x91eba053fc03bcac1328de38925f0a1ab9dba586):
+[delegation + first sweep](https://etherscan.io/tx/0x909cab5a08442b7ede39752cb25b497c72d5992cd0fc8c6ee9ac888eb9ff6c98)
+(type-4),
+[second Binance sweep](https://etherscan.io/tx/0xc8e7df2413212215ef2dd27b603d92b99a04ef180b8fa87eebdf6787d4cb715a)
+and
+[Kraken sweep](https://etherscan.io/tx/0x8e7f014001124afe3294f709971ed2de67f7e89f8f338724044af7a75acad8f2).
 Neither exchange hardcodes a 21'000 gas limit or refuses an address carrying a
 delegation designator, so the third variant's "ETH deposits are impossible"
 did not materialize for these two exchanges. The per-asset split stays the
 decided default (exchange behavior is heterogeneous and contract-batched
 withdrawal paths, where the 2'300-gas stipend could still bite, remain
 untested), but the shared permanently-delegated address is a viable Phase 2
-candidate; per-exchange results in DEFI-2993.
+candidate.
 
 ### Step 2: Withdraw from the CEX to the deposit address
 
@@ -1061,10 +1078,16 @@ Notes:
   is the EOA, so the batch gate cannot be written as `msg.sender == address(this)`.
 * The approval is for exactly `balance` and is consumed in full within the same
   call — no standing allowance toward the helper ever survives a sweep.
-* No `receive()`/`fallback` and no ETH sweep entry point on purpose: plain ETH
-  sends to a delegated ERC-20 address must fail rather than be silently accepted
-  (`R12`); Phase 2 ETH addresses are never delegated and sweep via a key-signed
-  `depositEth` call (step 5).
+* The delegate carries an empty payable `receive()` — guarded to revert on the
+  implementation address itself, where no attestation can ever recover and ETH
+  would be locked — plus attested `sweepEth`/`sweepEthBatch` entry points that
+  forward the balance through the helper's `depositEth`
+  ([#11449](https://github.com/dfinity/ic/pull/11449)). This keeps the
+  single-shared-address variant open for Phase 2 (see the mainnet data point in
+  step 1); under the decided per-asset layout the schema-2 ETH addresses are
+  never delegated and sweep via a key-signed `depositEth` call (step 5), and a
+  plain ETH send to a delegated ERC-20 address is accepted as a recoverable
+  wrong-asset deposit (see Non-goals) rather than bounced.
 * Reentrancy is moot: no storage, and funds can only move through the helper
   toward the minter's main address.
 * Supported tokens are assumed standard: non-fee-on-transfer and non-rebasing, so
--- a/rs/ethereum/cketh/docs/deposit_from_cex.md
+++ b/rs/ethereum/cketh/docs/deposit_from_cex.md
@@ -461,6 +461,22 @@ with Phase 2):
 | **Single shared address** with *set-and-clear* delegation (install delegate, sweep, re-delegate to `address(0)`) | One address per account | Two tECDSA signatures + ≈ 2 × 12'500–25'000 gas per sweep cycle; short window in which fixed-gas ETH transfers fail at the sender; more complex delegation lifecycle |
 | **Single shared address** with permanent delegation | One address, one authorization ever | Breaks ETH deposits entirely: plain sends to a delegated address without `receive()` revert — funds bounce at the CEX (`R12` holds, but ETH deposits are impossible) |
 
+**Mainnet data point (2026-09-04, DEFI-2993,
+[#11449](https://github.com/dfinity/ic/pull/11449)):** with an empty payable
+`receive()` on the delegate (kept empty to fit the 2'300-gas `transfer`/`send`
+stipend) plus attested `sweepEth`/`sweepEthBatch` entry points, real ETH
+withdrawals from **Binance** and **Kraken** both reached an
+EIP-7702-delegated address — top-level sends with gas limits of 207'128 (flat)
+and 31'830 (estimate-based) respectively, 21'055 gas used in both cases,
+matching the anvil measurement exactly — and were swept back permissionlessly.
+Neither exchange hardcodes a 21'000 gas limit or refuses an address carrying a
+delegation designator, so the third variant's "ETH deposits are impossible"
+did not materialize for these two exchanges. The per-asset split stays the
+decided default (exchange behavior is heterogeneous and contract-batched
+withdrawal paths, where the 2'300-gas stipend could still bite, remain
+untested), but the shared permanently-delegated address is a viable Phase 2
+candidate; per-exchange results in DEFI-2993.
+
 ### Step 2: Withdraw from the CEX to the deposit address
 
 The user pastes the deposit address into the exchange withdrawal form; the CEX
--- a/rs/ethereum/cketh/minter/CkSweeperAttested.sol
+++ b/rs/ethereum/cketh/minter/CkSweeperAttested.sol
@@ -33,16 +33,22 @@ struct SweepItem {
 /// its (r, s, v) components.
 contract CkSweeperAttested {
     address private immutable HELPER;
+    address private immutable SELF;
 
     constructor(address helper) {
         HELPER = helper;
+        SELF = address(this);
     }
 
-    /// Accepts plain ETH sends to a delegated deposit address. Deliberately
-    /// empty: batched CEX withdrawals may forward value with only the 2'300-gas
-    /// `transfer`/`send` stipend, which any extra logic would exceed, and
-    /// detection is balance-based so no event is needed.
-    receive() external payable {}
+    /// Accepts plain ETH sends to a delegated deposit address, but not to the
+    /// implementation itself: no attestation can ever recover to the contract's
+    /// own address, so ETH landing there would be locked forever. Otherwise
+    /// deliberately minimal: batched CEX withdrawals may forward value with only
+    /// the 2'300-gas `transfer`/`send` stipend, and detection is balance-based
+    /// so no event is needed.
+    receive() external payable {
+        require(address(this) != SELF, "implementation cannot be swept");
+    }
 
     /// The attestation digest: keccak256 over a fixed-length, domain-separated
     /// preimage. The ASCII prefix (first byte 0x63) cannot collide with typed
--- a/rs/ethereum/cketh/minter/CkSweeperAttested.sol
+++ b/rs/ethereum/cketh/minter/CkSweeperAttested.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.20;
 
 interface ICkDeposit {
     function depositErc20(address erc20Address, uint256 amount, bytes32 principal, bytes32 subaccount) external;
+    function depositEth(bytes32 principal, bytes32 subaccount) external payable;
 }
 
 interface IErc20Balance {
@@ -37,6 +38,12 @@ contract CkSweeperAttested {
         HELPER = helper;
     }
 
+    /// Accepts plain ETH sends to a delegated deposit address. Deliberately
+    /// empty: batched CEX withdrawals may forward value with only the 2'300-gas
+    /// `transfer`/`send` stipend, which any extra logic would exceed, and
+    /// detection is balance-based so no event is needed.
+    receive() external payable {}
+
     /// The attestation digest: keccak256 over a fixed-length, domain-separated
     /// preimage. The ASCII prefix (first byte 0x63) cannot collide with typed
     /// transactions (0x00-0x04), EIP-7702 authorizations (0x05), EIP-191/712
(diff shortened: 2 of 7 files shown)


rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -461,6 +461,22 @@ with Phase 2):
...
+EIP-7702-delegated address — top-level sends with gas limits of 207'128 (flat)
+and 31'830 (estimate-based) respectively, 21'055 gas used in both cases,
+matching the anvil measurement exactly — and were swept back permissionlessly.
+Neither exchange hardcodes a 21'000 gas limit or refuses an address carrying a
+delegation designator, so the third variant's "ETH deposits are impossible"
+did not materialize for these two exchanges. The per-asset split stays the
+decided default (exchange behavior is heterogeneous and contract-batched
+withdrawal paths, where the 2'300-gas stipend could still bite, remain
+untested), but the shared permanently-delegated address is a viable Phase 2
+candidate; per-exchange results in DEFI-2993.
```
gregorydemay: nit: remove links to internal documents (Jira). Add links to the transactions visible on Ethereum instead (with addresses for delegated EOA, sweeper, helper, what was acting as minter)

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -66,17 +73,40 @@ contract CkSweeperAttested {
...
+        }
+    }
+
     /// Permissionless batch entry point: sweeps many delegated deposit EOAs in a
     /// single transaction, each with its own attestation.
     function sweepErc20Batch(SweepItem[] calldata items, address[] calldata tokens) external {
         for (uint256 i = 0; i < items.length; ++i) {
             SweepItem calldata item = items[i];
-            CkSweeperAttested(item.deposit).sweepErc20(
+            CkSweeperAttested(payable(item.deposit)).sweepErc20(
```
gregorydemay, line 102: why this change?

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -37,6 +38,12 @@ contract CkSweeperAttested {
         HELPER = helper;
     }
 
+    /// Accepts plain ETH sends to a delegated deposit address. Deliberately
+    /// empty: batched CEX withdrawals may forward value with only the 2'300-gas
+    /// `transfer`/`send` stipend, which any extra logic would exceed, and
+    /// detection is balance-based so no event is needed.
+    receive() external payable {}
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in 000660b461: the constructor captures `SELF = address(this)` and `receive()` reverts when `address(this) == SELF`, so ETH can no longer land on the implementation. Covered by the new test `the_sweeper_implementation_rejects_plain_eth`; the guard costs ~40 gas (top-level send: 21'055 → 21'095) and still fits the stipend (see next thread).

rs/ethereum/cketh/minter/tests/deposit_from_cex_demo.rs
```diff
@@ -536,6 +541,111 @@ fn attested_sweep_rejects_a_forged_attestation() {
...
+    assert_eq!(
+        anvil.balance(&deposit),
+        0,
+        "the failed send must not leave ETH at the deposit address"
+    );
+
+    let receipt = anvil.send_eth(&cex, &deposit, ETH_DEPOSIT_AMOUNT, None);
+    assert!(
+        status_ok(&receipt),
+        "the delegate's receive() must accept a plain send given enough gas"
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Added in 000660b461: `StipendForwarder.sol` forwards value with Solidity `transfer` (the 2'300-gas stipend), and `a_stipend_transfer_from_a_contract_reaches_a_delegated_deposit_address` asserts the send into the delegated, guarded `receive()` succeeds. It passes: the EIP-7702 delegation-resolution cost is charged to the caller, not the stipend, leaving the stipend for the ~40-gas receive body.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -37,6 +38,12 @@ contract CkSweeperAttested {
         HELPER = helper;
     }
 
+    /// Accepts plain ETH sends to a delegated deposit address. Deliberately
+    /// empty: batched CEX withdrawals may forward value with only the 2'300-gas
+    /// `transfer`/`send` stipend, which any extra logic would exceed, and
+    /// detection is balance-based so no event is needed.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Updated in 000660b461: the "no receive()/fallback on purpose" bullet in the sweeper-delegate section now describes the guarded `receive()` and the attested `sweepEth`/`sweepEthBatch` entry points, and points to the mainnet data point added to the step-1 address-layout table. The per-asset layout remains the recorded decision; the shared-address variant is marked as a viable Phase 2 candidate rather than impossible.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Done in 000660b461: the Jira reference is gone; the paragraph now links the five mainnet transactions (Binance before/after, Kraken after, and the three sweeps) and the four addresses (delegated deposit EOA, sweeper delegate, helper deployment, minter stand-in EOA) on etherscan.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
gregorydemay: 🤖 Required by solc: once a contract has a payable `receive()`, converting a plain `address` to that contract type must go through `payable(...)` — without it compilation fails with "Explicit type conversion not allowed from non-payable \"address\" to \"contract CkSweeperAttested\", which has a payable fallback function". No behavior change.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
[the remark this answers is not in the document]
gregorydemay: 🤖 Also covered in the e2e suite now (66b24da96b): `should_bounce_eth_sent_to_the_sweeper_implementation` in `tests/deposit_from_cex.rs` deploys the real sweep contracts via `deploy_sweep_contracts` and asserts a value transfer to the delegate implementation is mined, reverts, and leaves its balance at zero.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -461,6 +461,39 @@ with Phase 2):
...
+([before delegation](https://etherscan.io/tx/0xaa13310032da7fc310f451f23d013e4b9e69c24bd201ac5cc57b54b15b0afbf2),
+[after](https://etherscan.io/tx/0x59cbb83ac893da16789b15d0494b7c3e5f382e1254f259d9c6fbcef5e45747c1))
+and **Kraken**
+([after](https://etherscan.io/tx/0x3b1dbd6c29b5f56d71c0ce1baae931f8e519852775be199fe62575ba6361b864))
+reached the delegated deposit EOA
+[`0x5e5402ed…`](https://etherscan.io/address/0x5e5402edfe155b5116f1f84d0b97b9e74f5ebc85),
+delegated to the sweeper
+[`0x3983baa1…`](https://etherscan.io/address/0x3983baa1703f69e76669768164c822918e843a25)
+— top-level sends with gas limits of 207'128 (flat) and 31'830
+(estimate-based) respectively, 21'055 gas used in both cases, matching the
```
mbjorkqvist, line 478: As discussed offline, the mainnet tests were run against the deployed contract without the self-check, which would bump the used gas from 21'055 to 21'095. This should not be an issue wrt. the tested centralized exchanges, since they put much higher limits, but perhaps a clarification in the doc would make sense.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -461,6 +461,39 @@ with Phase 2):
 | **Single shared address** with *set-and-clear* delegation (install delegate, sweep, re-delegate to `address(0)`) | One address per account | Two tECDSA signatures + ≈ 2 × 12'500–25'000 gas per sweep cycle; short window in which fixed-gas ETH transfers fail at the sender; more complex delegation lifecycle |
 | **Single shared address** with permanent delegation | One address, one authorization ever | Breaks ETH deposits entirely: plain sends to a delegated address without `receive()` revert — funds bounce at the CEX (`R12` holds, but ETH deposits are impossible) |
 
+**Mainnet data point (2026-09-04,
+[#11449](https://github.com/dfinity/ic/pull/11449)):** with an empty payable
```
mbjorkqvist, line 465: `receive()` is no longer empty — it carries the `SELF` guard as of 000660b461. Same wording in the note at line 1081 ("an empty payable `receive()` — guarded to revert…", which reads as a contradiction) and in the module doc at `deposit_from_cex_demo.rs:29`.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1045,10 +1078,16 @@ Notes:
   is the EOA, so the batch gate cannot be written as `msg.sender == address(this)`.
 * The approval is for exactly `balance` and is consumed in full within the same
   call — no standing allowance toward the helper ever survives a sweep.
-* No `receive()`/`fallback` and no ETH sweep entry point on purpose: plain ETH
-  sends to a delegated ERC-20 address must fail rather than be silently accepted
-  (`R12`); Phase 2 ETH addresses are never delegated and sweep via a key-signed
-  `depositEth` call (step 5).
+* The delegate carries an empty payable `receive()` — guarded to revert on the
```
mbjorkqvist, line 1081: This bullet sits in the Notes under the normative `CkSweeperViaHelper` listing — decided variant B — which has neither `receive()` nor `sweepEth`/`sweepEthBatch`. Three consequences:

* it reads as if the *decided* delegate accepts ETH;
* it contradicts line 879, "a plain ETH transfer to the now-delegated `D` reverts at the sender, since `S` has no `receive()` (`R12`)", which is still correct for variant B;
* the closing clause — "a plain ETH send to a delegated ERC-20 address is accepted as a recoverable wrong-asset deposit … rather than bounced" — is false under the decided layout.

Scoping it explicitly to `CkSweeperAttested` would fix all three. This is the part of the earlier doc-contradiction point that is still open: the update landed in the variant-B section.

Smaller and adjacent: the table row at line 462 ("Breaks ETH deposits entirely") is superseded by the new paragraph right below it and could point at it.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -32,9 +33,21 @@ struct SweepItem {
...
+    }
+
+    /// Accepts plain ETH sends to a delegated deposit address, but not to the
+    /// implementation itself: no attestation can ever recover to the contract's
+    /// own address, so ETH landing there would be locked forever. Otherwise
+    /// deliberately minimal: batched CEX withdrawals may forward value with only
+    /// the 2'300-gas `transfer`/`send` stipend, and detection is balance-based
+    /// so no event is needed.
+    receive() external payable {
+        require(address(this) != SELF, "implementation cannot be swept");
```
mbjorkqvist, line 52: Rather than adding this self-check, which costs 40 gas on each legitimate deposit, an alternative could be to add a `rescueEth()` that executes only when `address(this) == SELF`, forwarding the implementation's balance through the helper under an immutable recovery principal. Zero hot-path gas, ~120 bytes of code, and the locked-ETH case becomes *recoverable* rather than merely blocked.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -32,9 +33,21 @@ struct SweepItem {
...
+    address private immutable SELF;
 
     constructor(address helper) {
         HELPER = helper;
+        SELF = address(this);
+    }
+
+    /// Accepts plain ETH sends to a delegated deposit address, but not to the
+    /// implementation itself: no attestation can ever recover to the contract's
+    /// own address, so ETH landing there would be locked forever. Otherwise
```
mbjorkqvist: Slight overstatement: the guard closes the `receive()` path only. `selfdestruct` from another contract, and coinbase payments, still deliver ETH to the implementation, where it is locked for exactly the reason given here. A "…via `receive()`" qualifier would make the comment exact.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
```diff
@@ -66,17 +79,40 @@ contract CkSweeperAttested {
         }
     }
 
+    /// Sweeps the deposit address' entire ETH balance to the minter through the
+    /// helper's `depositEth`, under the same attestation check as `sweepErc20`.
+    function sweepEth(bytes32 principal, bytes32 subaccount, bytes32 r, bytes32 s, uint8 v) external {
```
mbjorkqvist, line 86: Flagging rather than objecting, since nothing here wires the minter to issue ETH attestations.

A wrong-asset ETH arrival swept this way is credited with no screening of the actual funder: `depositEth` emits `ReceivedEthOrErc20` with `from_address = msg.sender`, i.e. the deposit EOA (`DepositHelperWithSubaccount.sol:504`), and that is precisely the field the blocklist checks (`minter/src/deposit.rs:394`). The ERC-20 path recovers the funder from the `Transfer` log; ETH has no equivalent.

The design already books this as an accepted weakening "to review with compliance before Phase 2 ships" (`docs/deposit_from_cex.md:640-643`) — but putting `receive()` on the ERC-20 delegate extends it to Phase 1 addresses. Worth keeping visible so it does not end up inside Phase 1 sign-off by accident.

rs/ethereum/cketh/minter/tests/deposit_from_cex_demo.rs
```diff
@@ -536,6 +541,175 @@ fn attested_sweep_rejects_a_forged_attestation() {
     assert_eq!(events[0].principal, encode_principal(&principal));
 }
 
+#[test]
+fn attested_sweep_moves_plain_eth_to_the_minter() {
```
mbjorkqvist, line 545: The ETH side has no negative test. `attested_sweep_rejects_a_forged_attestation` covers `sweepErc20` only, and `sweepEth` is the permissionless entry point that moves *native* ETH — a forged or foreign-principal attestation ⇒ revert with the balance unchanged is the case I would most want pinned, and it drops in next to the existing one.

rs/ethereum/cketh/minter/tests/deposit_from_cex_demo.rs
```diff
@@ -25,7 +25,10 @@
 //! matching the proposed design where sweeping is *permissionless* but each sweep
 //! carries a minter attestation (a deposit-key signature binding the address to
 //! its IC account). Those sweeps are submitted by a non-minter relayer, and a
-//! separate test shows a forged attestation is rejected.
+//! separate test shows a forged attestation is rejected. A further test funds a
+//! *delegated* deposit address with plain ETH — accepted by the delegate's empty
```
mbjorkqvist: "empty `receive()`" here too — it carries the `SELF` guard now. See the note on `docs/deposit_from_cex.md:465`.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -282,6 +283,29 @@ fn should_fund_the_sweeper_address_by_burning_cketh_from_the_fee_account() {
     );
 }
 
+#[test]
```
mbjorkqvist: This test is anvil-only (`Anvil::start()` + `deploy_sweep_contracts`) but lives in the `size = "large"` PocketIC target, and it duplicates `the_sweeper_implementation_rejects_plain_eth` in the small anvil target. The second assertion is also close to vacuous: the contract is deployed three lines earlier, so its balance is 0 whether or not the transfer reverted. Dropping this one loses no coverage.

rs/ethereum/cketh/test_utils/src/anvil.rs
```diff
@@ -263,6 +263,30 @@ impl Anvil {
             .to_string()
     }
 
+    /// Sends `wei` from `from` (an unlocked dev account) to `to` with an explicit gas limit,
```
mbjorkqvist: This reimplements the body of the private `send_transaction_with_value` directly above it, which differs only in lacking a gas parameter; symmetrically, `send_eth` in `deposit_from_cex_demo.rs` reimplements the `send_transaction_with_value` that this same diff adds there. One helper with an optional gas limit on each side would cover both.

Minor: `!anvil.plain_transfer_succeeds(...)` reads awkwardly at the call site — returning the receipt, as the demo side does, would let the assertion say what it means.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Correct — the deployed contract predates the guard, so the 21'055 in the doc is the deployed-bytecode figure and the guarded contract measures 21'095 on anvil. The clarification will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Agreed on the doc occurrences — will be addressed with #11313, which aims at improving the docs. The `deposit_from_cex_demo.rs:29` module-doc wording is fixed in 4561a992f8.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fair — the bullet landed under the variant-B (`CkSweeperViaHelper`) listing and should be scoped to `CkSweeperAttested`; same for pointing the superseded table row at the new paragraph. Will be addressed with #11313, which aims at improving the docs.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
gregorydemay: 🤖 Fixed in 4561a992f8: the comment now scopes the guard to the receive() path and names selfdestruct/coinbase as the remaining (unrecoverable) ways ETH can strand on the implementation.

rs/ethereum/cketh/minter/tests/deposit_from_cex_demo.rs
gregorydemay: 🤖 Added in 4561a992f8: `attested_eth_sweep_rejects_a_forged_attestation` delegates a deposit EOA, funds it with ETH, and asserts a sweepEth carrying an attestation signed by a foreign key reverts with the balance unchanged — next to the ERC-20 counterpart as suggested.

rs/ethereum/cketh/minter/tests/deposit_from_cex_demo.rs
gregorydemay: 🤖 Fixed in 4561a992f8 (dropped "empty" from the module doc).

rs/ethereum/cketh/minter/CkSweeperAttested.sol
gregorydemay: 🤖 The trade-off is real: rescueEth() would zero the hot-path cost (back to 21'055, more stipend headroom) and turn the locked-ETH case into a recoverable deposit. What it needs that the guard does not is a recovery principal baked in at deployment — a governance/deployment decision (whose account absorbs strays?) plus one more permissionless entry point to review. Since the delegate is not final until the Phase 2 sign-off, proposal: keep the cheap guard in this exploratory PR and decide guard-vs-rescueEth (and the recovery principal) when the deployed delegate is specified — tracked with the design-doc work in #11313.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
gregorydemay: 🤖 Acknowledged, and worth keeping visible exactly as you frame it: receive() on the ERC-20 delegate extends the unscreened-ETH-funder weakening from Phase 2 to Phase 1 addresses, even though nothing wires the minter to issue ETH attestations yet. This belongs in the compliance paragraph the design doc already carries ("to review with compliance before Phase 2 ships") with the Phase-1 scope made explicit — will be addressed with #11313, which aims at improving the docs, so it stays out of any accidental Phase 1 sign-off.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 All three observations are accurate — anvil-only test in the large PocketIC target, coverage duplicated from the small demo target, and the balance assertion is trivially satisfied. It was added deliberately to pin the guard against the same deploy_sweep_contracts fixture the e2e suite trusts, but that is a thin justification for a duplicated test in the expensive target; happy to drop it (or fold the guard assertion into an existing e2e test) — author's call.

rs/ethereum/cketh/minter/CkSweeperAttested.sol
gregorydemay: I would like to avoid having to open that door (`rescueEth`), since if it's there it means it can be used, right 😄 ? The additional gas cost is negligible of this check is negligible.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Dropped in the latest commit: the e2e copy and the harness helper that existed only for it are removed; the guard stays covered by `the_sweeper_implementation_rejects_plain_eth` in the small anvil target.

PR #11450, opened 2026-09-04
perf: CON-1657 Validate HTTPS outcall signatures in parallel

eichhorl:
The main bottleneck of validating blocks with many HTTPS outcalls responses is signature verification. Therefore, in https://github.com/dfinity/ic/pull/10345 we started using batch verification which has better performance than validating signatures one-by-one.

With this PR we additionally pass a rayon thread pool to the HTTPS outcalls payload builder. This thread pool has 16 threads, and is already used by other payload builders to reduce validation times (the `ChainKeyPayloadBuilder` for instance). Similar to that, we use the thread pool to split all signatures of the current block in to at most 16 batches (one for each thread), and then validate each batch in parallel, with a single batch verification call. This reduces the validation time of full HTTPS outcalls payloads by up to ~80%.

### Before
```sql
                        time:   [107.00 ms 107.66 ms 108.38 ms]
                        time:   [149.01 ms 149.87 ms 150.85 ms]
                        time:   [8.2811 ms 8.3068 ms 8.3353 ms]
                        time:   [79.840 ms 80.548 ms 81.381 ms]
                        time:   [179.01 ms 180.38 ms 181.82 ms]
```
### After
```sql
                        time:   [21.519 ms 21.788 ms 22.066 ms]
                        change: [-80.032% -79.762% -79.482%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [22.973 ms 23.205 ms 23.457 ms]
                        change: [-84.695% -84.517% -84.330%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [3.0804 ms 3.1072 ms 3.1346 ms]
                        change: [-62.932% -62.595% -62.245%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [14.198 ms 14.318 ms 14.452 ms]
                        change: [-82.456% -82.225% -81.996%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [43.591 ms 43.947 ms 44.336 ms]
                        change: [-75.914% -75.637% -75.337%] (p = 0.00 < 0.05)
                        Performance has improved.
```


rs/https_outcalls/consensus/benches/payload_validation.rs
```diff
@@ -52,6 +53,21 @@ const REGISTRY_VERSION: RegistryVersion = RegistryVersion::new(1);
 /// `min_responses` of the benchmark's flexible requests.
 const FLEXIBLE_MIN_RESPONSES: u32 = 1;
 
+/// Number of threads of the payload builder's thread pool. Matches
+/// `ic_consensus::consensus::MAX_CONSENSUS_THREADS`, i.e. the size of the pool
```
pierugo-dfinity: > `ic_consensus::consensus::MAX_CONSENSUS_THREADS`

Maybe we could update the documentation of this constant. We could also move it somewhere the current crate (and `ic_consensus_chain_key`, which also duplicates it) can see it

rs/https_outcalls/consensus/benches/payload_validation.rs
```diff
@@ -52,6 +53,21 @@ const REGISTRY_VERSION: RegistryVersion = RegistryVersion::new(1);
...
+/// The thread pool the benchmarked payload builder verifies signatures on.
+fn thread_pool() -> Arc<ThreadPool> {
+    Arc::new(
+        ThreadPoolBuilder::new()
+            .num_threads(NUM_THREADS)
+            .build()
+            .expect("Failed to create thread pool"),
+    )
+}
+
```
pierugo-dfinity: This is also the third same `build_thread_pool` function

rs/https_outcalls/consensus/src/payload_builder/tests.rs
```diff
@@ -1986,6 +1987,17 @@ pub(crate) fn metadata_to_shares(
...
+/// The thread pool the payload builder under test verifies signatures on.
+/// We use only 1 thread to avoid non-deterministic test failures.
+fn test_thread_pool() -> Arc<ThreadPool> {
+    Arc::new(
+        ThreadPoolBuilder::new()
+            .num_threads(1)
+            .build()
+            .expect("Failed to create thread pool"),
+    )
+}
```
pierugo-dfinity: Fourth :D

pierugo-dfinity (approved):
Nice improvement!

I would also be interested in the performance of payloads with less signatures. Would the overhead of the thread pool and/or batch-verification be higher than the savings? 

In particular for example, for 16 signatures, is it really faster to verify 16 batches of 1 signature in parallel compared to verifying a single batch of 16 signatures?

Also, if `chunk_size` is small (like in the previous example), we might better be off verifying signatures individually if batching adds some overhead.

> if batching adds some overhead

From what I can tell by quickly skimming through the code, this does not seem to be the case, but can't tell for sure.

rs/https_outcalls/consensus/benches/payload_validation.rs
eichhorl: Good idea, done! 👍

rs/https_outcalls/consensus/benches/payload_validation.rs
eichhorl: Done 👍

rs/https_outcalls/consensus/src/payload_builder/tests.rs
eichhorl: Done 👍

eichhorl:
> I would also be interested in the performance of payloads with less signatures. Would the overhead of the thread pool and/or batch-verification be higher than the savings?

I wondered the same thing, and the break-even seems to be at around 8 signatures. I.e. verifying less than 8 signatures as a single batch without even touching the thread pool (which is probably most of the overhead) is faster than verifying each of them in parallel. But these are low microsecond differences, so I'm not sure it's worth the optimization (a single fully replicated response on a 13 node subnet already has 9 signatures).

rs/https_outcalls/consensus/BUILD.bazel
```diff
@@ -90,6 +91,7 @@ rust_test(
         "@crate_index//:proptest",
         "@crate_index//:rand",
         "@crate_index//:rand_chacha",
+        "@crate_index//:rayon",
```
pierugo-dfinity:
```suggestion
```
nit: I don't think this is a direct dependency anymore

rs/https_outcalls/consensus/BUILD.bazel
```diff
@@ -124,5 +126,6 @@ rust_bench(
         "//rs/types/cycles",
         "//rs/types/types",
         "@crate_index//:criterion",
+        "@crate_index//:rayon",
```
pierugo-dfinity:
```suggestion
```
nit: I don't think this is a direct dependency anymore

rs/https_outcalls/consensus/benches/payload_validation.rs
pierugo-dfinity: Wonderful!

pierugo-dfinity:
> I wondered the same thing, and the break-even seems to be at around 8 signatures.

Cool, thanks for looking into the numbers.

> But these are low microsecond differences

Yeah ok, let's not worry about it then

Last point: there's also no significant difference between
- ~9 signatures in parallel but individually, not as single-signature batches vs
- the current implementation: ~9 signatures in parallel as single-signature batches

?

pierugo-dfinity:
Well I guess it should be similar to the difference between verifying a signature vs a single-signature batch
```
crypto_basicsig_msg_size_32_remote_vault/Ed25519/verify
                        time:   [39.701 µs 39.742 µs 39.790 µs]
crypto_basicsig_batch_verify/batch_size_1/verify_batch
                        time:   [52.108 µs 52.162 µs 52.229 µs]
```

Should be fine :D

PR #11453, opened 2026-09-04
fix(sns): store generic nervous system function call reply on the proposal

Rachit2323:
What was wrong: when the SNS asks another canister to do something, that canister sends back an answer. But we were never actually reading that answer, we just checked "did it respond at all" and then threw the answer away. So even if the answer said "this didn't work," we'd still mark it as a success, and nobody could ever go back and see what really happened, because we deleted it.

What we fixed: we stopped deleting the answer. Now we save it, so anyone can look at it later and see exactly what came back.

What we did NOT fix: we still can't automatically tell if the answer means success or failure, because this feature can talk to any canister, and every one of them answers differently. There's no way to understand all of them automatically. So we're not trying to guess anymore, we're just making sure the answer doesn't disappear.

Small limit: if the answer is really long, we only keep part of it, so it doesn't take up too much space.

Testing: added tests to check the answer gets saved properly, that nothing changed for the failure case, and that long answers get shortened correctly.


daniel-wong-dfinity-org-twin:
Hi, Rachit2323. Sorry it has taken me a while to look at this. I just came back from a 2 week vacation.

rs/sns/governance/api/src/ic_sns_governance.pb.v1.rs
```diff
@@ -1079,6 +1079,14 @@ pub struct ProposalData {
     pub action_auxiliary: Option<proposal_data::ActionAuxiliary>,
     /// This proposal's topic.
     pub topic: Option<topics::Topic>,
+    /// The raw reply bytes returned by the target canister when this
+    /// proposal's action was an ExecuteGenericNervousSystemFunction call that
+    /// completed successfully at the IC call level. SNS does not know this
+    /// reply's Candid schema, so it is stored as-is (opaque), truncated to at
```
daniel-wong-dfinity-org-twin, line 1085: I don't think it needs to be explained why the type here is blob. I mean, mentioning it is not harmful per se, so if you really like it, keep it.

The more interesting fact is that it gets truncated.

The reason for trunctation also does not seem like it needs to be mentioned, but it's not harmful either. I mean, I think the reason can be "very easily" inferred. Everyone knows that space is a finite resource.

Ditto elsewhere, ofc.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+        id: FUNCTION_ID,
+        name: "Test Function".to_string(),
+        description: None,
+        function_type: Some(FunctionType::GenericNervousSystemFunction(
+            GenericNervousSystemFunction {
+                topic: None,
+                target_canister_id: Some(TARGET_CANISTER_ID.get()),
+                target_method_name: Some(TARGET_METHOD.to_string()),
+                validator_canister_id: Some(TARGET_CANISTER_ID.get()),
+                validator_method_name: Some(TARGET_METHOD.to_string()),
```
daniel-wong-dfinity-org-twin, line 26: For realism, this should be different from target_method_name.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    types::test_helpers::NativeEnvironment,
+};
+use ic_nervous_system_canisters::cmc::FakeCmc;
+use maplit::btreemap;
+
+const TARGET_CANISTER_ID: CanisterId = CanisterId::from_u64(600);
+const TARGET_METHOD: &str = "do_something";
+const FUNCTION_ID: u64 = 1000;
+
+fn new_valid_generic_nervous_system_function() -> NervousSystemFunction {
```
daniel-wong-dfinity-org-twin, line 15: Could you use lazy_static instead? Because this is really a constant.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+/// `execute_proposal` (mirrors the pattern used by the SNS-upgrade tests in
+/// `assorted_governance_tests.rs`).
+fn new_ready_to_execute_proposal(payload: Vec<u8>) -> ProposalData {
+    let action = Action::ExecuteGenericNervousSystemFunction(ExecuteGenericNervousSystemFunction {
+        function_id: FUNCTION_ID,
+        payload,
+    });
+
+    ProposalData {
+        action: (&action).into(),
```
daniel-wong-dfinity-org-twin, line 42: Can you do this without `.into()`?

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+        id: Some(1.into()),
```
daniel-wong-dfinity-org-twin, line 43: Ditto.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+                validator_method_name: Some(TARGET_METHOD.to_string()),
+            },
+        )),
+    }
+}
+
+/// A proposal with a single decisive Yes ballot, ready to be executed by
+/// `execute_proposal` (mirrors the pattern used by the SNS-upgrade tests in
+/// `assorted_governance_tests.rs`).
+fn new_ready_to_execute_proposal(payload: Vec<u8>) -> ProposalData {
```
daniel-wong-dfinity-org-twin, line 35: If payload were a constant, then, this could also be a lazy_static. constants are good, because they are very very predictable. Functions can do all sorts of wacky things. Ofc, this one is boring (good!), but you can't be sure of that without reading the code.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    let mut governance = new_governance(new_ready_to_execute_proposal(payload), env);
+
+    // Step 2: Run the code under test.
+    let proposal_data = execute_proposal(&mut governance, 1);
+
+    // Step 3: Verify result(s). Existing failure-path behavior is unchanged.
+    assert_eq!(proposal_data.execution_reply, None);
+    assert!(proposal_data.failure_reason.is_some());
```
daniel-wong-dfinity-org-twin, line 124: This could be more precise with assert_eq.

rs/sns/governance/src/governance/execute_generic_nervous_system_function_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    let mut governance = new_governance(new_ready_to_execute_proposal(payload), env);
+
+    // Step 2: Run the code under test.
+    let proposal_data = execute_proposal(&mut governance, 1);
+
+    // Step 3: Verify result(s).
+    let stored_reply = proposal_data.execution_reply.expect("reply should be set");
+    assert_eq!(stored_reply.len(), MAX_SCALAR_FIELD_LEN_BYTES);
+    assert_eq!(stored_reply, oversized_reply[..MAX_SCALAR_FIELD_LEN_BYTES]);
+}
```
daniel-wong-dfinity-org-twin, line 151: What about timestamp and failure_reason? I mean, sure, those are less interesting, but that's why you would assert those later.

rs/sns/governance/src/canister_control.rs
```diff
@@ -275,11 +279,17 @@ pub async fn perform_execute_generic_nervous_system_function_validate_and_render
 }
 
 /// Executes a generic nervous system function (i.e., a non-native SNS proposal).
+///
+/// On success, returns the raw reply bytes from the target canister's
+/// method. Because this is a *generic* function, the SNS does not know the
+/// reply's Candid schema, so these bytes are returned as-is (opaque) for the
+/// caller to store for transparency/auditability; they are not decoded or
+/// otherwise interpreted here.
```
daniel-wong-dfinity-org-twin, line 287: Again, I think no justification is needed for blob. But keep it if you really want to.

rs/sns/governance/src/canister_control.rs
```diff
@@ -299,17 +309,11 @@ pub async fn perform_execute_generic_nervous_system_function_call(
...
-            //
-            // This could then be converted into a Result<(), GovernanceError>.
-            // For now, any reply is considered a success.
-            Ok(())
-        }
+        // The reply's Candid schema is unknown to the SNS in general (that is
+        // the whole point of a *generic* nervous system function), so it
+        // cannot be decoded here to detect an application-level error.
+        // Instead, the raw bytes are returned so they can be stored on the
+        // proposal for a human/tooling to inspect after the fact.
```
daniel-wong-dfinity-org-twin, line 316: You just said this in the doc comment! This is REALLY superfluous.

rs/sns/governance/src/canister_control.rs
daniel-wong-dfinity-org-twin, line 306: If you wanted to, you could just do `.map_err`, since `Ok` is now just a pass through, but `match` if fine too, if you prefer.

rs/sns/governance/src/governance.rs
```diff
@@ -1762,6 +1763,20 @@ impl Governance {
...
+    ///
+    /// The SNS cannot decode this reply (its Candid schema is unknown), so
+    /// the bytes are stored as-is, truncated to at most
+    /// `MAX_SCALAR_FIELD_LEN_BYTES` bytes to bound governance's stable memory
+    /// usage against a misbehaving or malicious target canister.
+    fn set_proposal_execution_reply(&mut self, pid: u64, mut execution_reply: Vec<u8>) {
+        execution_reply.truncate(MAX_SCALAR_FIELD_LEN_BYTES);
+        if let Some(proposal) = self.proto.proposals.get_mut(&pid) {
+            proposal.execution_reply = Some(execution_reply);
+        }
```
daniel-wong-dfinity-org-twin, line 1777:
```suggestion
        let Some(proposal) = self.proto.proposals.get_mut(&pid) else {
            log!(
                ERROR,
                "Tried to record reply after executing proposal {}, but unable to find the proposal.",
                pid,
            );
            return;
        };
        
        proposal.execution_reply = Some(execution_reply);
```

I mean, I don't think `else` will ever happen, but if it does, I want a way to find out about it.

rs/sns/governance/src/governance.rs
```diff
@@ -2170,10 +2185,12 @@ impl Governance {
...
-            Action::ExecuteGenericNervousSystemFunction(call) => {
-                self.perform_execute_generic_nervous_system_function(call)
-                    .await
-            }
+            Action::ExecuteGenericNervousSystemFunction(call) => self
+                .perform_execute_generic_nervous_system_function(call)
+                .await
+                .map(|execution_reply| {
+                    self.set_proposal_execution_reply(proposal_id, execution_reply);
+                }),
```
daniel-wong-dfinity-org-twin, line 2193: What if we made this part of perform_execute_generic_nervous_system_function? My motivation is just to keep match arms as small and uniform as possible. OTOH, the amount of additional code here is pretty minimal.

rs/sns/governance/src/governance.rs
```diff
@@ -2170,10 +2185,12 @@ impl Governance {
                     Err(e) => Err(e),
                 }
             }
-            Action::ExecuteGenericNervousSystemFunction(call) => {
-                self.perform_execute_generic_nervous_system_function(call)
-                    .await
-            }
+            Action::ExecuteGenericNervousSystemFunction(call) => self
```
daniel-wong-dfinity-org-twin, line 2188: What happened to braces? Other arms have them. Did rustfmt force you to get rid of them? If not, please, make this code like the rest.

rs/sns/governance/src/governance.rs
```diff
@@ -3654,6 +3672,9 @@ impl Governance {
             is_eligible_for_rewards: true,
             action_auxiliary,
             topic: Some(i32::from(proposal_topic)),
+
+            // A new proposal has not been executed yet, so there is no reply.
+            execution_reply: ProposalData::default().execution_reply,
```
daniel-wong-dfinity-org-twin, line 3677: This seems like an extravagant way to obtain a None.

rs/sns/governance/src/canister_control_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+        id,
+        name: "Test Function".to_string(),
+        description: None,
+        function_type: Some(FunctionType::GenericNervousSystemFunction(
+            GenericNervousSystemFunction {
+                topic: None,
+                target_canister_id: Some(target_canister_id.get()),
+                target_method_name: Some(target_method.to_string()),
+                validator_canister_id: Some(target_canister_id.get()),
+                validator_method_name: Some(target_method.to_string()),
```
daniel-wong-dfinity-org-twin, line 20: ditto. Suggestion: format!("validate_{}", target_method)

rs/sns/governance/src/canister_control_tests.rs
```diff
@@ -0,0 +1,86 @@
+use super::*;
+use crate::pb::v1::nervous_system_function::{FunctionType, GenericNervousSystemFunction};
+use crate::types::test_helpers::NativeEnvironment;
+
+fn new_valid_generic_nervous_system_function(
+    id: u64,
+    target_canister_id: CanisterId,
+    target_method: &str,
```
daniel-wong-dfinity-org-twin, line 8: I don't think any callers care to pick these, so just use fixed values + lazy_static.

rs/sns/governance/src/canister_control_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+        payload,
+    };
+
+    // Step 2: Run the code under test.
+    let result = perform_execute_generic_nervous_system_function_call(&env, function, call).await;
+
+    // Step 3: Verify result(s).
+    let error = result.unwrap_err();
+    assert_eq!(error.error_type, ErrorType::External as i32);
+    // Do not assert exact wording; just look for a key phrase.
```
daniel-wong-dfinity-org-twin, line 84: This seems unnecessary, but not really harmful either.

daniel-wong-dfinity-org-twin:
Hmm. The size thing is worrying. Based on the [comment], I don't think we can increase the limit. We need to find way(s) to put this WASM on a diet.

[comment]: https://github.com/dfinity/ic/blob/cf4331ee55afe96bc0fa0d54ff7bd6dc53b790f3/rs/sns/governance/src/types.rs#L81

PR #11454, opened 2026-09-04
test: run 16 more system-tests on the local backend

basvandijk:
Namespace RBE workers have 32 CPUs, so a system-test whose `reserved_cpus` (`ceil(cpus / cpus_oversubscription_factor)`, surfaced as `exec_properties["cpu"]`) is ≤ 32 can run on the `"local"` backend there.

25 of the 64 tests carrying `backend = "farm",  # Too big for the "local" backend.` meet that bound. **9 keep the attribute anyway**: `cpus` only ever becomes a CPU reservation, and these ask for 300–500 GiB of RAM via `VmResourceOverrides`, which the local backend passes verbatim to QEMU (`network_reliability_test`, `state_sync_malicious_chunk_test`, `api_bn_performance_test`, and 6 of the `testnets`). The other **16 lose it**.

That un-`manual`s 26 `_local` targets — `subnet_splitting_v2_test`, 8× `sr_app_*`, 3× `tecdsa_*`, 2× `xnet_slo_3_subnets*` — all using default 4 GiB node VMs (≤ 48 GiB per test). `state_sync_performance` and the `medium` testnet are already `manual`, so CI is unchanged for them.

Every non-manual `local_system_test` in the repo now carries a `cpu` exec property, max **25**.

Checks: `bazel build //... --nobuild` clean, `bazel run //:buildifier` a no-op, and the non-manual `local_system_test` set goes 195 → 221 with nothing dropped. The newly-enabled tests have not been run locally (no KVM group on the dev box) — this PR's CI is the first real signal.

### DIFF
--- a/rs/tests/consensus/BUILD.bazel
+++ b/rs/tests/consensus/BUILD.bazel
@@ -260,7 +260,6 @@ system_test_nns(
 
 system_test_nns(
     name = "subnet_splitting_v2_test",
-    backend = "farm",  # Too big for the "local" backend.
     cpus = MIN_LOCAL_CPUS + 11 * DEFAULT_VCPUS_PER_VM,  # 11 IC Node VMs (1+1 System + 8 Application + 1 VerifiedApplication) * 6 vCPUs.
     enable_metrics = True,
     tags = [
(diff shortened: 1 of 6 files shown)


basvandijk:
run of local-system-tests: https://github.com/dfinity/ic/actions/runs/33881001823/job/101049365270.

dfinityianblenke (approved):

nmattia (approved):

pierugo-dfinity (approved):

PR #11456, opened 2026-09-04
perf: Reduce CertifiedSlicePool lock contention

alin-at-dfinity:
Calling `CertifiedSlicePool`'s `put()` and `append()` methods requires holding the lock on the whole pool. These methods decode protobufs, pack and unpack slices and do an expensive BLS signature verification (on the order of milliseconds). In the meantime other operations on the pool (`put`, `append`, but also `take`) are blocked.

Instead, do the heavy lifting before grabbing the lock to actually mutate the pool. `put_impl()` ensures that we will only pool a valid slice (e.g. no messages that we already inducted), so any race condition between one `put`/`append` and another is benign (a separate PR will introduce "retain the more valuable slice" logic, so we don't e.g. replace a slice with messages with a header-only slice).

Triggered by https://github.com/dfinity/ic/pull/11343#discussion_r3893417291.


rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1331,21 +1328,23 @@ impl CertifiedSlicePool {
...
             subnet_id,
-            self.certified_stream_store.as_ref(),
+            certified_stream_store,
             registry_version,
             log,
         )?;
+        let unpacked = slice.try_into()?;
 
-        self.put_impl(subnet_id, slice.try_into()?)
+        pool.lock().unwrap().put_impl(subnet_id, unpacked)
```
eichhorl, line 1347: Technically we could also return any displaced slice and drop it outside of the lock

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1357,28 +1356,33 @@ impl CertifiedSlicePool {
...
+        // Clone the pooled slice, if any, instead of removing it: this way the
+        // payload builder can make use of it while we validate; and if the merge
+        // or the validation fail, we bail out without having mutated the pool.
+        let pooled = pool.lock().unwrap().slices.get(&subnet_id).cloned();
 
-        // Query the pool for an existing slice, without removing it. This way, if
-        // unpacking or validation fail, we can simply bail out without mutating the
-        // pool.
-        let slice = match self.slices.get(&subnet_id) {
+        let partial: UnpackedStreamSlice = partial.try_into()?;
```
eichhorl, line 1382: Why did this move below the pool access? Since it is fallible, doing it first makes sense to me

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1320,7 +1313,11 @@ impl CertifiedSlicePool {
     }
 
     /// Places the provided slice into the pool, after trimming off any prefix
-    /// before the corresponding `self.stream_positions` entry.
+    /// before the corresponding `stream_positions` entry.
+    ///
+    /// Decoding, unpacking and validating the slice (requiring an expensive BLS
+    /// signature verification) is done without holding the pool lock. The lock is
+    /// only taken to insert the slice.
```
eichhorl, line 1320: Should we add a note that this function will deadlock (or panic?) if the lock is already held? Also for `append` below

rs/xnet/payload_builder/src/certified_slice_pool.rs
eichhorl, line 268: We could skip this in the  `prefix.is_empty()` case

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1357,28 +1356,33 @@ impl CertifiedSlicePool {
...
         partial: CertifiedStreamSlice,
+        certified_stream_store: &dyn CertifiedStreamStore,
         registry_version: RegistryVersion,
         log: ReplicaLogger,
     ) -> CertifiedSliceResult<()> {
-        let partial: UnpackedStreamSlice = partial.try_into()?;
+        // Clone the pooled slice, if any, instead of removing it: this way the
+        // payload builder can make use of it while we validate; and if the merge
+        // or the validation fail, we bail out without having mutated the pool.
+        let pooled = pool.lock().unwrap().slices.get(&subnet_id).cloned();
```
eichhorl, line 1380: Would it make sense to wrap the `UnpackedStreamSlice`s in `slices` in an `Arc<_>`?

rs/xnet/payload_builder/src/lib.rs
```diff
@@ -1432,10 +1433,12 @@ pub struct PoolRefillTask {
 
 impl PoolRefillTask {
     /// Starts an async task that fills the slice pool in the background.
+    #[allow(clippy::too_many_arguments)]
```
eichhorl, line 1436: `clippy.toml` actually sets the limit to `12`, so this shouldn't be needed?
https://github.com/dfinity/ic/blob/1fec1641f9a300d56fa224625d2a096f344f297e/clippy.toml#L1

rs/xnet/payload_builder/tests/certified_slice_pool.rs
```diff
@@ -1386,23 +1395,147 @@ fn pool_garbage_collect_deleted_subnet(
...
+        pool.try_lock().is_ok(),
+        "pool lock held while validating a slice"
+    );
+}
+
+/// Tests that the pool lock is not held while a slice is being validated (which
+/// is the expensive part of `put()` and `append()`); and that a concurrent
+/// `append()` that lands meanwhile is simply overwritten.
+#[test_strategy::proptest(ProptestConfig::with_cases(10))]
+fn pool_append_concurrent_append(
```
eichhorl, line 1487: We could also add tests
- `pool_append_concurrent_take` to make sure messages that were already taken aren't pooled again
- `pool_append_concurrent_garbage_collect` where the work of `append` is thrown away

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Done, although I'm not sure it's worth the extra complexity and documentation.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Done.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Why would it panic? It will block if another `put()` or `append()` is in progress, but none of them calls each other, so we never reenter the lock.

And to deadlock you'd need two locks.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: I don't think so: in this particular case we need a copy to mutate (and potentially discard if we fail), so an `Arc` wouldn't help. I suppose we could switch from `append()`, which requires the prefix to be a value; to `prepend()` which could take the prefix by reference and only clone it when we know for sure that we'll go through with the concatenation.

But that seems quite the (unrelated) change.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Claude? Moved it back.

rs/xnet/payload_builder/src/lib.rs
alin-at-dfinity: Done.

rs/xnet/payload_builder/tests/certified_slice_pool.rs
alin-at-dfinity: I have an upcoming change (https://github.com/dfinity/ic/pull/11474) that retains the "more useful" slice during a `put()` or `append()` (instead of the existing implementation's "last one wins"). Once that's in (and this test is updated accordingly), all we need to test in terms of race conditions is that:
 1. "most useful wins" does what it says; and
 2. `append()` calls "most useful wins".

rs/xnet/payload_builder/src/certified_slice_pool.rs
eichhorl: Yes, but couldn't someone try to do the following?
```rust
let guard = pool.lock().unwrap();
CertifiedSlicePool::put(...); // tries to call `lock()` a second time
```

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1331,21 +1328,24 @@ impl CertifiedSlicePool {
...
-            self.certified_stream_store.as_ref(),
+            certified_stream_store,
             registry_version,
             log,
         )?;
+        let unpacked = slice.try_into()?;
 
-        self.put_impl(subnet_id, slice.try_into()?)
+        let result = pool.lock().unwrap().put_impl(subnet_id, unpacked);
+        result.map(|_| ())
```
eichhorl: Maybe we can add a comment why this is done in two lines and not one. Or call `drop` directly

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -264,16 +264,16 @@ mod messages {
...
                 self.messages = postfix;
                 Ok((None, Some(self)))
             } else if postfix.is_empty() {
+                // Return all messages, drop `self`.
                 Ok((Some(prefix), None))
             } else {
-                // Both prefix and postfix are non-empty.
+                // Both prefix and postfix are non-empty. Retain the postfix.
+                self.count_bytes = byte_size(&postfix)?;
                 Ok((Some(prefix), Some(Self::new(postfix)?)))
```
eichhorl, line 290: It seems `Self::new(postfix)` also calls `byte_size`

rs/xnet/payload_builder/src/certified_slice_pool.rs
eichhorl: Hm, good point, although it might still allow us to do a cheap clone while holding the lock and then the expensive one outside of it. But I haven't tried it out or judged if it's worth the change.

rs/xnet/payload_builder/src/lib.rs
eichhorl, line 909: Somewhat off topic, but do we really need to do all the validation again during block making, when reading from the local pool?

eichhorl (approved):

rs/xnet/payload_builder/tests/certified_slice_pool.rs
eichhorl: I see, makes sense!

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Sure, but isn't it a bit weird to write (same thing, just with names changed for emphasis):

```rust
let guard = mutex.lock().unwrap();
do_something(&mutex);
```

You are asking for trouble if you grab a non-reentrant mutex and then also pass this same mutex to a function. Leave alone a function function explicitly documented to acquire said mutex (OTOH what else was it going to do?).

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Right, so we never have to explicitly recompute it: the first branch replaces the messages unchanged; the second branch consumes `self`; and the third branch also consumes it and returns a brand new instance of `Self`.

Duh...

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: If it's OK with you, I'd rather leave it out of this change.

We can definitely consider `Arc` and switching from `append()` to `prepend()` (which is closer to what the code does anyway) as a separate change.

Or, something that I only realized recently: relax the constraints around monotonically increasing `signals_end`; drop all of the slice merging logic (we still want pruning, though); and simply hold on to a collection of slices per stream. Then we can induct a couple of messages from this slice, a couple from that and the latest header from a third, all in the same block. But that's a pretty radical change, also affecting logic in the DSM.

rs/xnet/payload_builder/src/lib.rs
alin-at-dfinity: I was also confused about validation yesterday, until I realized that there are two places where we do different levels of validation:
 * `certified_slice_pool::validate_slice()` merely calls `CertifiedStreamStore::decode_certified_stream_slice()`, which decodes the slice and validates its witness and signature. This is what the pool uses.
 * `XNetPayloadBuilderImpl::validate_slice()` does that plus ensuring that the various message and signal indices are where we expect them to be (monotonically increasing, messages extend gap-free what we've already inducted, max signal limit is respected, etc.).

The pool only cares about the slice being correctly certified and signed by a subnet, as per the registry. It tries to have slices with the "right indices", but will hold on to slices that are slightly off.

When we actually include a slice into a payload, we must ensure that it fits exactly after the previously inducted slices. If it doesn't, we just drop it and move on. The pool will then pull it again and we can later try again to include it into a block.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Done.

rs/xnet/payload_builder/src/certified_slice_pool.rs
eichhorl: > a function function explicitly documented to acquire said mutex, what else was it going to do

I think that's fair 😆

rs/xnet/payload_builder/src/certified_slice_pool.rs
eichhorl: Totally fine with taking a closer look asynchronously, if at all 👍

rs/xnet/payload_builder/src/lib.rs
eichhorl: But then it sounds like when we include a slice into a payload, we do not need to make sure again that the slice is correctly certified and signed by a subnet?

alin-at-dfinity: Technically, yes.

Although if the slice pool put together the slice from chunks (via `append()`) and then pruned the prefix (via `garbage_collect()`); all that was done by XNet logic extending the authoritative crypto/certified state logic. And while this is tested and we have metrics that we (rarely) look at to ensure that it works as expected, it is not trusted (or tested) in the same way.

OTOH, it might be acceptable for the block maker to rely on one implementation for validation; and for the other replicas to use the other; but that creates the risk for all proposed blocks to be considered valid by their maker but invalid by everyone else, stalling the subnet.

eichhorl: I see, makes sense. I wouldn't be surprised if validating the signatures is the majority of the time spent by the payload builder, but I agree, ideally we use the same validation when ingesting into the pool and when validating received blocks.

PR #11457, opened 2026-09-04
feat: Handle stuck VMs

frankdavid:
We've seen an SEV VM getting stuck due to invalid VMCB. This PR adds additional logging and automatic restart to such VMs.

### DIFF
--- a/rs/ic_os/os_tools/guest_vm_runner/src/main.rs
+++ b/rs/ic_os/os_tools/guest_vm_runner/src/main.rs
@@ -25,7 +25,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::pin::pin;
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use strum_macros::AsRefStr;
 use tempfile::NamedTempFile;
 use tokio::io::{AsyncBufReadExt, BufReader};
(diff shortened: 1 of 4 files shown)


rs/ic_os/os_tools/guest_vm_runner/src/main.rs
```diff
@@ -344,17 +353,39 @@ impl VirtualMachine {
...
                 }
-                Ok((state, reason)) => {
-                    warn!("VM is in state {state}, reason: {reason}");
+                Ok((VIR_DOMAIN_CRASHED, reason)) => {
+                    warn!("VM crashed, reason: {reason}");
+                    break;
                 }
+                // Any other state (PAUSED, SHUTOFF, PMSUSPENDED, NOSTATE) does not execute
+                // guest code. A stuck VM never resumes on its own and must be restarted.
+                Ok((state, reason)) => match stuck_since {
```
Bownairo, line 372: nit: Potentially enumerate these, in case a new state is added? I doubt this will ever happen. We can still bind `state` with `state @ (XXX | YYY)`.

Bownairo (approved):

rs/ic_os/os_tools/guest_vm_runner/src/main.rs
frankdavid: It's an integer, not an enum, so listing it doesn't buy us anything, we'd still need a fallback arm.

PR #11458, opened 2026-09-04
fix(ci): don't parse guest_launch_measurements for HostOS proposals

r-birkner:
`repro-check -p <proposal-id>` crashed with an uncaught `KeyError: 'guest_launch_measurements'` on every HostOS version-election proposal.

`process_proposal()` read and transformed `guest_launch_measurements` from the proposal payload before branching on the proposal type. HostOS election payloads only carry `hostos_version_to_elect`, `hostos_versions_to_unelect`, `release_package_sha256_hex` and `release_package_urls` — no launch measurements. Since `main()` only catches `VerificationError` and `RuntimeError`, the `KeyError` surfaced as a traceback with exit code 1, indistinguishable from a real reproducibility mismatch. As HostOS and GuestOS are elected in separate proposals each release, this affected roughly half of all election proposals.

The measurements are only used in the GuestOS return branch anyway, so parsing them up front was both wrong and unnecessary.

*Changes:*
* Extract payload parsing into a module-level `parse_proposal_payload()`, which reads `guest_launch_measurements` only in the GuestOS branch.
* Detect the proposal type from the payload keys rather than substring-matching `json.dumps(proposal_data)` — the old check also matched versions appearing in `*_to_unelect` lists.
* Correct the `process_proposal()` return annotation (the measurements element is a `dict`, not a `str`) and drop the unused `self.proposal_launch_measurements` attribute.
* Add `ci/scripts/repro_check_test.py` (+ a `py_test` in a new `ci/scripts/BUILD.bazel`) covering a HostOS payload without measurements, the GuestOS hex→bytes conversion, and the missing-version error.

Verified with `bazel test //ci/scripts:repro_check_test`, `bazel build //ci/scripts:all --nobuild`, and `bazel run //:buildifier`.

### DIFF
--- a/ci/scripts/BUILD.bazel
+++ b/ci/scripts/BUILD.bazel
@@ -1,0 +1,8 @@
+load("@rules_python//python:defs.bzl", "py_test")
+
+py_test(
+    name = "repro_check_test",
+    size = "small",
+    srcs = ["repro_check_test.py"],
+    data = ["repro-check"],
+)
--- a/ci/scripts/repro-check
+++ b/ci/scripts/repro-check
@@ -181,6 +181,25 @@ def compare_hashes(local_hash_value: str, remote_hash_value: str, os_type: str) 
         )
         return True
 
+
+def parse_proposal_payload(payload: dict, proposal_id: int) -> tuple[str, str | None, dict | None, str | None]:
+    """
+    Extracts (git_hash, guest_os_hash, guest_os_measurements, host_os_hash) from an
+    election proposal payload. HostOS proposals carry no launch measurements.
+    """
+    sha256_hex = payload["release_package_sha256_hex"]
+    if payload.get("replica_version_to_elect"):
+        # NOTE: We convert the "human" hex format from the dashboard API to the
+        # byte format that is actually used in the proposal directly.
+        measurements = payload["guest_launch_measurements"]
+        for measurement in measurements["guest_launch_measurements"]:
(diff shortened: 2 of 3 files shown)


basvandijk (approved):

r-birkner:
Tested it with both master and this branch:
```
$ ./ci/scripts/repro-check -p 143903 --hostos
```
`master` fails:
```
$ ./ci/scripts/repro-check -p 143903 --hostos
2026/09/14 | 14:54:39 | 1789397679 [ℹ️] Using temporary directory: /home/rbirkner/.cache/repro-check/tmp-1789397679
Traceback (most recent call last):
  File "/home/rbirkner/ic/./ci/scripts/repro-check", line 1039, in <module>
    main()
  File "/home/rbirkner/ic/./ci/scripts/repro-check", line 1022, in main
    verifier.run(dry_run)
  File "/home/rbirkner/ic/./ci/scripts/repro-check", line 970, in run
    guest_os_hash, guest_os_measurements, host_os_hash = self.process_proposal()
                                                         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rbirkner/ic/./ci/scripts/repro-check", line 611, in process_proposal
    proposal_launch_measurements = proposal_data["payload"]["guest_launch_measurements"]
                                   ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'guest_launch_measurements'
```
The branch succeeds.

pietrodimarco-dfinity (approved):

PR #11459, opened 2026-09-04
feat(registry): Add ic-admin get-standard-engine-replica-version

r-birkner:
Adds a read-only `ic-admin get-standard-engine-replica-version` subcommand that
prints the singleton `StandardEngineReplicaVersionRecord` stored under the
`standard_engine_replica_version` registry key. If the record was never set, it
prints a clear message instead of panicking.

Output is routed through the existing `print_value` helper, so `--json` works.

Verified: `bazel build //rs/registry/admin:ic-admin` plus the rules_rust clippy
aspect are clean, the affected non-system Bazel tests pass, and the command was
smoke-tested against `https://ic0.app` in both plain and `--json` modes.

```
$ ic-admin --nns-url https://icp-api.io get-standard-engine-replica-version
Using NNS URLs: ["https://icp-api.io/"]
Fetching the most recent value for key: standard_engine_replica_version
Most recent version is 63855. Value:
StandardEngineReplicaVersion { new_replica_version_id: "e9fd136f3fed6cb44625379e056ad63789d52d2c", old_replica_version_id: "79c01052b5f7f49d3cf53d04d696cb2893294cd3", deployment_progress: 1.0 }
```


bjoernek (commented):
Read-only `ic-admin` subcommand, nice complement to the existing `propose-to-update-standard-engine-replica-version`. I confirmed that `Error::KeyNotPresent` is what the registry canister returns for both never-set and tombstoned keys (`rs/registry/canister/canister/canister.rs:404-407`), so the match arm is right, and the command is correctly left out of the secret-key/HSM allowlist since it is read-only. One inline comment, plus a request for a test:

**Could we cover the absent-record path in a test?** `rs/registry/admin/tests/integration_tests.rs` already runs the real `ic-admin` binary against a PocketIC NNS, and a freshly installed NNS leaves this key unset, so that path costs almost nothing to reach. It is also the one case the mainnet smoke test in the description cannot reach (mainnet has the record set), and it was the path that was wrong in the first commit before Copilot flagged it. Now that `--json` consumers will depend on the `{"version": null, "value": null}` shape, it seems worth pinning that shape down in a test rather than in a commit message.

rs/registry/admin/bin/main.rs
```diff
@@ -4957,6 +4961,46 @@ async fn main() {
...
+                .await
+            {
+                Ok((bytes, version)) => {
+                    let record = StandardEngineReplicaVersionRecord::decode(&bytes[..])
+                        .expect("Error decoding value from registry.");
+                    let record = StandardEngineReplicaVersion {
+                        new_replica_version_id: record.new_replica_version_id,
+                        old_replica_version_id: record.old_replica_version_id,
+                        deployment_progress: record.deployment_progress,
+                    };
```
bjoernek, line 4985: The local struct duplicates the record's schema with nothing tying the two together: because the fields are copied across one by one, adding a field to `standard_engine_replica_version.proto` still compiles here and simply disappears from the output, in both plain and `--json` mode. For a tool whose job is "show me what is actually in the registry", a silently incomplete record is the failure mode I would least want, since the output is exactly what you end up trusting.

Prost does give the generated type a `Debug` impl; only `Serialize` is missing, and that is opted into per proto package in the generator. So the smaller fix is one entry in `rs/protobuf/generator/src/lib.rs`, alongside the ones already there for every other registry record `ic-admin` prints (`.registry.subnet`, `.registry.replica_version`, `.registry.unassigned_nodes_config`, ...):

```rust
config.type_attribute(
    ".registry.standard_engine_replica_version",
    "#[derive(serde::Serialize, serde::Deserialize)]",
);
```

Regenerate, drop the local struct, and pass the decoded record straight to `print_value`; new proto fields then show up for free. The only visible change is that plain mode prints `StandardEngineReplicaVersionRecord { ... }` rather than `StandardEngineReplicaVersion { ... }`, and the JSON is identical.

If you would rather not regenerate the protos, destructuring instead of field access at least turns a new proto field into a compile error right here, which is the pattern the neighbouring propose command already uses (`let Self { .. } = self.clone()`, main.rs:2293):

```rust
let StandardEngineReplicaVersionRecord {
    new_replica_version_id,
    old_replica_version_id,
    deployment_progress,
} = StandardEngineReplicaVersionRecord::decode(&bytes[..])
    .expect("Error decoding value from registry.");
```

Either works for me. The current code is correct today; it just has no mechanism to stay correct.

bjoernek (approved):

schneiderstefan (approved):

PR #11463, opened 2026-09-05
fix(sns): validate logo PNG signatures

Rachit2323:
## Bug

  - SNS logo validation said the logo must be a PNG.
  - But the code only checked that the logo was base64 text.
  - That means non-image data, like base64-encoded plain text, could be accepted as a logo.

  ## Fix

  - Decode the base64 logo data.
  - Check that the decoded bytes start with the PNG signature.
  - Reuse the same validation logic for both SNS logos and token logos.

  ## Tests

  - Added a test that accepts a real PNG logo.
  - Added a test that rejects base64 text that is not a PNG.
  - Updated existing test fixtures to use a real tiny PNG.

  ## Checks

  - `cargo test -p ic-nervous-system-common test_validate_token_logo`
  - `cargo test -p ic-sns-governance --lib validate_logo`
  - `cargo test -p ic-sns-init test_token_logo_validation`
  - `cargo test -p ic-sns-init test_ledger_init_args_is_valid`
  - `cargo test -p ic-sns-governance --lib test_validate_and_render_manage_ledger_parameters`

### DIFF
--- a/rs/nervous_system/common/src/ledger_validation.rs
+++ b/rs/nervous_system/common/src/ledger_validation.rs
@@ -83,25 +83,61 @@ pub fn validate_token_name(token_name: &str) -> Result<(), String> {
     Ok(())
 }
 
-pub fn validate_token_logo(token_logo: &str) -> Result<(), String> {
+const PNG_SIGNATURE: [u8; 8] = [0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A];
+
+pub fn validate_logo(logo: &str, field_name: &str) -> Result<(), String> {
     const PREFIX: &str = "data:image/png;base64,";
 
-    if token_logo.len() > MAX_LOGO_LENGTH {
+    if logo.len() > MAX_LOGO_LENGTH {
         return Err(format!(
-            "Error: token_logo must be less than {MAX_LOGO_LENGTH} characters, roughly 256 Kb"
+            "{field_name} must be less than {MAX_LOGO_LENGTH} characters, roughly 256 Kb"
         ));
     }
 
-    if !token_logo.starts_with(PREFIX) {
+    if !logo.starts_with(PREFIX) {
         return Err(format!(
-            "Error: token_logo must be a base64 encoded PNG, but the provided \
-            string doesn't begin with `{PREFIX}`."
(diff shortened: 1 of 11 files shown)


rs/nervous_system/common/src/ledger_validation.rs
```diff
@@ -83,25 +83,61 @@ pub fn validate_token_name(token_name: &str) -> Result<(), String> {
...
+
+pub fn validate_token_logo(token_logo: &str) -> Result<(), String> {
+    validate_logo(token_logo, "token_logo")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::validate_token_logo;
+
+    const VALID_PNG_LOGO: &str = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAD0lEQVQIHQEEAPv/AAD/DwIRAQ8HgT3GAAAAAElFTkSuQmCC";
```
daniel-wong-dfinity-org-twin, line 125: You might want to explain that this image just consists of one green pixel.

rs/sns/governance/src/types/tests.rs
```diff
@@ -1014,9 +1015,18 @@ fn test_nervous_system_parameters_wont_validate_without_the_required_claimer_per
     }
 }
 
+const VALID_PNG_LOGO: &str = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAD0lEQVQIHQEEAPv/AAD/DwIRAQ8HgT3GAAAAAElFTkSuQmCC";
```
daniel-wong-dfinity-org-twin, line 1018: Can you move this to the top?

rs/sns/governance/unreleased_changelog.md
```diff
@@ -11,6 +11,9 @@ on the process that this file is part of, see
 
 ## Changed
 
+Logo validation now rejects a `logo` whose decoded bytes are not actually a PNG,
+instead of accepting any base64 that carries the `data:image/png;base64,` prefix.
```
daniel-wong-dfinity-org-twin, line 15: You might want to be more precise: we now look for PNG magic bytes.

daniel-wong-dfinity-org-twin (commented):
So, this just makes sure that the magic bytes are in place, yes? I mean, it could be invalid in other ways that aren't checked, right?

PR #11465, opened 2026-09-07
feat: Add a monotonic `consumed_cycles` total to `CanisterMetrics`

mraszyk:
`CanisterMetrics::consumed_cycles` behaves like a gauge: a prepayment raises it and the matching refund lowers it again. That makes it awkward to build monitoring on top of, which is why `consumed_cycles_by_use_cases` already has a monotonic twin, only bumped once the refund of a prepayment is known.

This adds the same twin for the scalar total: `CanisterMetrics::consumed_cycles_monotonic`, maintained by `observe_consumed_cycles_with_use_case` alongside the by-use-case amounts, and persisted in `CanisterStateBits` (proto field 71).

It is exactly the scalar equivalent of `consumed_cycles_by_use_cases_monotonic` summed over the use cases that `consumed_cycles` covers, i.e. everything except HTTPS outcalls, which are only tracked at the subnet level.

The field is absent in checkpoints written before it, and decodes as zero — the same starting point the by-use-case map had when it was introduced. Nothing consumes the new field yet: outside of tests, the only reader is `tip.rs`, which persists it. No metric or behaviour changes.

A follow-up commit renames the pre-existing `as_counter(s)` fields to `monotonic`, at both the canister and the subnet level: neither the `CanisterMetrics` / `CanisterStateBits` fields nor the Prometheus metrics they feed are actual counters. It is a pure rename of Rust and protobuf identifiers; the exported Prometheus metric names are deliberately left untouched, so existing dashboards and alerts keep working.

Note on a pre-existing inconsistency that the rename makes visible, and that Copilot flags: in `metrics.rs`, the subnet-level contribution to the monotonic Prometheus aggregate is joined from `get_consumed_cycles_by_use_case()`, i.e. the gauge map, not `get_consumed_cycles_by_use_case_monotonic()`. Since `migrate_outcalls_cycles_to_use_cases()` deliberately backfills only the gauge map, that join can export the very jump the separate monotonic map exists to avoid. This is unchanged from `master` (only the accumulator's name differs here), and correcting it would lower the exported HTTP/ECDSA outcall values by the backfilled amount, which reads as a counter reset — so it is left out of this PR.

Follow-ups, kept separate:
 * backfilling `consumed_cycles_monotonic` from the `consumed_cycles` gauge, so it starts out with the full history rather than from zero;
 * exporting the monotonic total under the `replicated_state_consumed_cycles_since_replica_started` gauge.


rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
```diff
@@ -502,6 +502,15 @@ message CanisterStateBits {
...
   repeated ConsumedCyclesByUseCase consumed_cycles_by_use_cases_as_counters = 65;
+  // Total cycles consumed by the canister, presented as a counter: unlike
+  // `consumed_cycles`, which is lowered again whenever a refund is issued, this
+  // is only ever increased, by the actually consumed amount once the refund of a
+  // prepayment is known.
+  //
+  // Covers exactly the same use cases as `consumed_cycles`, i.e. everything
+  // except HTTPS outcalls, which are only tracked at the subnet level (and, for
+  // the canister, in `consumed_cycles_by_use_cases_as_counters`).
+  types.v1.NominalCycles consumed_cycles_as_counter = 71;
```
alin-at-dfinity: As commented on your earlier draft PR, "as counter" is not a great qualifier:
 * the `CanisterMetrics` / `CanisterStateBits` fields are clearly not Prometheus counters (they're just `NominalCycles` amounts); and
 * the Prometheus metric that this feeds is also intentionally not a counter.

IMHO "monotonic" would be a much better suffix. (Feel free to also apply it to the "by use case" collection.)

Same everywhere.

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
```diff
@@ -502,6 +502,15 @@ message CanisterStateBits {
...
   // various aggregations on them more easily than their gauge counterparts.
   repeated ConsumedCyclesByUseCase consumed_cycles_by_use_cases_as_counters = 65;
+  // Total cycles consumed by the canister, presented as a counter: unlike
+  // `consumed_cycles`, which is lowered again whenever a refund is issued, this
+  // is only ever increased, by the actually consumed amount once the refund of a
+  // prepayment is known.
+  //
+  // Covers exactly the same use cases as `consumed_cycles`, i.e. everything
+  // except HTTPS outcalls, which are only tracked at the subnet level (and, for
+  // the canister, in `consumed_cycles_by_use_cases_as_counters`).
```
alin-at-dfinity: Nit: This is still too verbose. While the second paragraph may be useful for the actual `CanisterMetrics` field, it is merely redundant (and likely to fall out of sync) here.

```suggestion
  // Total cycles consumed by the canister, monotonic: unlike `consumed_cycles`,
  // which is increased on prepayment and decreased on refund, this is only ever
  // increased on refund, by the actually consumed amount.
```

rs/replicated_state/src/canister_state/tests.rs
```diff
@@ -1475,6 +1475,97 @@ fn drops_aborted_canister_install_after_split() {
...
+        CanisterCyclesCostSchedule::Free,
+    ] {
+        test::<Instructions>(cost_schedule);
+        test::<RequestAndResponseTransmission>(cost_schedule);
+    }
+}
+
+/// A use case that is never refunded is accounted for on the counter right away.
+#[test]
+fn consumed_cycles_as_counter_accounts_for_final_use_cases_at_prepayment() {
```
alin-at-dfinity: Ignorant question: Do we still call it a prepayment if it's final? (As are e.g. memory usage charges?)

rs/state_layout/src/state_layout/proto.rs
```diff
@@ -102,6 +103,12 @@ impl TryFrom<pb_canister_state_bits::CanisterStateBits> for CanisterStateBits {
         let consumed_cycles =
             try_from_option_field(value.consumed_cycles, "CanisterStateBits::consumed_cycles")
                 .unwrap_or_default();
+        // Absent in checkpoints written before the field was introduced.
```
alin-at-dfinity:
```suggestion
```

rs/state_layout/src/state_layout.rs
```diff
@@ -197,6 +197,7 @@ pub struct CanisterStateBits {
     pub global_timer_nanos: Option<u64>,
     pub canister_version: u64,
     pub canister_creation_timestamp_nanos: Option<u64>,
+    pub consumed_cycles_as_counter: NominalCycles,
```
alin-at-dfinity: Nit: I'd put this next to `consumed_cycles`. Same everywhere.

alin-at-dfinity (approved):
Only a couple of nitpicks, the only meaningful one being an objection to the "as counter" qualifier.

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
mraszyk: Agreed, and done — `monotonic` everywhere. I also took you up on the "by use case" collection, at both the canister and the subnet level (`consumed_cycles_by_use_cases_monotonic` and `consumed_cycles_by_use_case_monotonic`); leaving the new field's neighbour as `as_counters` would have been the worst of both. That rename is its own commit, so it's easy to review apart from the new field.

One deliberate exception: the exported Prometheus metric names (`replicated_state_consumed_cycles_by_use_case_as_counters` and `replicated_state_consumed_cycles_from_replica_start_as_counters`) are untouched, so existing dashboards and alerts keep working. Renaming those is externally visible, so it felt like a separate decision — happy to do it if you'd rather.

rs/protobuf/def/state/canister_state_bits/v1/canister_state_bits.proto
mraszyk: Applied — second paragraph dropped from the proto; the `CanisterMetrics` accessor keeps that detail.

One tweak to your wording: I kept "once the refund is known" rather than "only ever increased on refund", because the use cases charged without a prepayment (`Memory`, `ComputeAllocation`, `Uninstall`, `IngressInduction`, `CanisterCreation`, `BurnedCycles`) are accounted for immediately at charge time, so "on refund" would be wrong for them. Same length:

```
  // Total cycles consumed by the canister, monotonic: unlike `consumed_cycles`,
  // which is increased on prepayment and decreased on refund, this is only ever
  // increased, by the actually consumed amount once the refund is known.
```

rs/replicated_state/src/canister_state/tests.rs
mraszyk: Not ignorant at all — my terminology was wrong. `consume_cycles`'s own doc already draws the distinction: *"either for cases where a prepayment needs to be made (that will be refunded later with `refund_cycles`) or a direct charge happens without a prepayment (e.g. when paying for memory)"*.

So it's a direct charge, and I've adopted that term. The test is now `consumed_cycles_monotonic_accounts_for_direct_charges_right_away`, and the accessor doc reads "…once the refund is known; or right away, for a direct charge made without a prepayment (e.g. for memory usage)".

rs/state_layout/src/state_layout/proto.rs
mraszyk: Removed. The `unwrap_or_default()` is pinned by `test_encode_decode_consumed_cycles_monotonic`, which now covers both the round-trip and the missing-field case.

rs/state_layout/src/state_layout.rs
mraszyk: Moved, and "everywhere": `CanisterStateBits` (Rust struct and the proto, where field 71 now sits right after field 22), `CanisterMetrics` and its constructor, and the `From`/`TryFrom` conversions.

PR #11474, opened 2026-09-07
feat: Prefer more useful XNet slices

alin-at-dfinity:
Unify the criteria for when a newly pulled slice should replace a pooled slice across all `put()` and `append()` paths.

Only replace a pooled slice with one at least as useful: beginning no later; failing that, with more messages; failing that, more signals; failing that, higher stream begin. Prevents a slice from a lagging node from clobbering a good slice. New `xnet_pool_put_count` metric records the outcome.

Also, when the pooled slice begins after the requested stream position (which may regress, e.g. when the block we built a payload for was not finalized), take its signals only and retain the messages, instead of dropping the slice. Drops `CertifiedSliceError::TakeBeforeSliceBegin`.

### DIFF
--- a/rs/xnet/payload_builder/src/certified_slice_pool.rs
+++ b/rs/xnet/payload_builder/src/certified_slice_pool.rs
@@ -23,7 +23,7 @@ use ic_types::{
 };
 use messages::Messages;
 use prometheus::{Histogram, IntCounterVec, IntGauge};
-use std::cmp::Reverse;
+use std::cmp::{Ordering, Reverse};
 use std::collections::BTreeMap;
 use std::convert::{From, TryFrom, TryInto};
 use std::sync::Arc;
@@ -1440,7 +1440,7 @@ impl CertifiedSlicePool {
     }
 
     /// Garbage collects the provided slice and pools the rest, iff more useful than
-    /// the already pooled slice (see `usefulness()`).
+    /// the already pooled slice (see `compare_usefulness()`).
     ///
     /// Returns `Err(InvalidPayload)` or `Err(WitnessPruningFailed)` if
     /// `unpacked` is malformed.
@@ -1462,10 +1462,10 @@ impl CertifiedSlicePool {
             };
         }
 
-        // Retain the more useful slice (`unpacked` vs already pooled), in case e.g.
-        // `unpacked` came from a lagging node.
+        // If there's a pooled slice, only replace it if `unpacked` is more useful (i.e.
+        // has more messages or signals).
         if let Some(pooled) = self.slices.get(&subnet_id)
-            && usefulness(pooled, stream_position) >= usefulness(&unpacked, stream_position)
+            && compare_usefulness(pooled, &unpacked, stream_position).is_ge()
         {
             self.metrics.observe_put(STATUS_LESS_USEFUL);
         } else {
@@ -1597,8 +1597,7 @@ fn witness_count_bytes(
         + pruned_nodes_bytes + known_nodes_bytes + fork_nodes_bytes
 }
 
-/// Returns a measure of how useful a slice is, as a tuple for lexicographic
-/// comparison:
+/// Compares two slices by how useful they are, in decreasing order of weight:
 ///
 ///  * First useful (beyond `stream_position`) message index: a slice beginning
 ///    after it may not have the messages to satisfy the next `take_slice()`.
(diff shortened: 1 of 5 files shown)


rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1400,29 +1436,43 @@ impl CertifiedSlicePool {
...
+                    return Ok(());
+                }
             };
         }
 
-        self.slices.insert(subnet_id, unpacked);
+        // Retain the more useful slice (`unpacked` vs already pooled), in case e.g.
+        // `unpacked` came from a lagging node.
+        if let Some(pooled) = self.slices.get(&subnet_id)
+            && usefulness(pooled, stream_position) >= usefulness(&unpacked, stream_position)
```
schneiderstefan: I feel like this would be easier to follow if it was a function should_replace, taking two slices as input.

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1547,6 +1597,50 @@ fn witness_count_bytes(
...
 
+/// Returns a measure of how useful a slice is, as a tuple for lexicographic
+/// comparison:
+///
+///  * First useful (beyond `stream_position`) message index: a slice beginning
+///    after it may not have the messages to satisfy the next `take_slice()`.
+///  * Its `messages_end`: more messages are better (and subjectively weighted
+///    higher than more signals).
+///  * Its `signals_end`: more signals are better.
+///  * Its stream `begin`: a higher one allows us to garbage collect more of our
```
schneiderstefan: Is there a signals_begin to also check?

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1322,11 +1339,6 @@ impl CertifiedSlicePool {
     /// Places the provided slice into the pool, after trimming off any prefix
     /// before the corresponding `self.stream_positions` entry.
     ///
-    /// On success always replaces the pooled slice regardless of its contents,
-    /// as slices may originate from malicious replicas and we would rather
```
schneiderstefan, line 1326: So does this argument with malicious replicas just not hold, considering they are threshold signed?

rs/xnet/payload_builder/src/certified_slice_pool.rs
```diff
@@ -1547,6 +1597,50 @@ fn witness_count_bytes(
...
+/// stream position may regress whenever we built a payload for a block that did
+/// not get finalized (e.g. the rank 1 block was picked instead).
+///
+/// Safety never enters into it: the slice has been certified by its source
+/// subnet. The only question is which of two slices is worth holding on to.
+fn usefulness(
+    slice: &UnpackedStreamSlice,
+    stream_position: Option<&ExpectedIndices>,
+) -> (
+    Option<Reverse<StreamIndex>>,
```
schneiderstefan: So header-only slices lose against anything with messages, is that correct? (None < Some(x))

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Precisely. This was the trigger for switching to a "keep the more useful slice" approach, since it was obviously bogus and (IIRC) slightly inconsistent with what `append()` did.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: No. We may or may not have reject signals and they may or may not regress, but all we care about is `signals_end` (being monotonically increasing and the farther along the better)

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Right. It's a subjective choice (there may be a thousand reject signals that are all basically responses vs one message), but the next time we pull from the subnet, whether or not there are extra messages, we will get an updated header that we can tack on to the messages we're already holding on to.

But the key word is subjective.

rs/xnet/payload_builder/src/certified_slice_pool.rs
alin-at-dfinity: Done.

schneiderstefan (approved):

PR #11487, opened 2026-09-08
feat: bump base64

nmattia:
This drops base64 0.13.1 (released 4 years ago) and consolidates the base64 versions around 0.22.1.

This includes some API changes (mostly specifying the decoding options via eg `base64::prelude::BASE64_STANDARD::decode(...)` & friends) and drops the `openssh-keys` (does not support recent base64, so we inline the `ssh-rsa`-specific decoding).


basvandijk (approved):

mbjorkqvist (approved):

blind-oracle (approved):

bjoernek (approved):

bjoernek:
Non-blocking note from the governance side (no action needed, just so the consequence is on record).

The `base64::decode` -> `BASE64_STANDARD.decode` rewrite is not quite semantics-preserving: base64 0.13's `decode()` did not require padding (see the comment in its `decode.rs`, "given that we don't require padding"), whereas 0.22's `BASE64_STANDARD` is built from `GeneralPurposeConfig::new()`, whose default is `DecodePaddingMode::RequireCanonical`. So input that lacks canonical `=` padding used to decode fine and is now rejected. Encoding is unchanged (both pad).

That touches two governance validators:

- `SnsMetadata::validate_logo` (`rs/sns/governance/src/types.rs`)
- `ledger_validation::validate_token_logo` (`rs/nervous_system/common/src/ledger_validation.rs`)

The logo one sits on an upgrade path: `canister_post_upgrade` -> `canister_init_` -> `ValidGovernanceProto::try_from` -> `sns_metadata.validate()` -> `validate_logo`, and that `try_from` is wrapped in `.expect(...)`. An SNS with a stored unpadded logo would therefore trap on its next upgrade.

I checked mainnet before approving: of the 54 entries from `list_deployed_snses`, the 38 that have a Wasm installed all have canonically padded logos (the other 16 are aborted swaps, `IC0537` no Wasm module). And since the same tightened validator now also guards the `ManageSnsMetadata` and `CreateServiceNervousSystem` submission paths, no unpadded logo can enter going forward either. So this is safe as-is.

If you'd prefer to keep 0.13 semantics exactly at these two call sites, a local engine with `.with_decode_padding_mode(DecodePaddingMode::Indifferent)` would do it. Fine either way; arguably the stricter check is an improvement.

rs/crypto/internal/crypto_lib/basic_sig/ecdsa_secp256k1/tests/api_tests.rs
```diff
@@ -99,7 +100,9 @@ mod keygen {
 
     #[test]
     fn should_fail_parsing_non_secp256k1_key_without_panic() {
-        let pk_der = base64::decode(crate::ED25519_PK_DER_BASE64).unwrap();
+        let pk_der = BASE64_STANDARD_NO_PAD
+            .decode(crate::ED25519_PK_DER_BASE64)
+            .unwrap();
```
pierugo-dfinity:
```suggestion
        let pk_der = BASE64_STANDARD
            .decode(crate::ED25519_PK_DER_BASE64)
            .unwrap();
```
I would pad the `ED25519_PK_DER_BASE64` constant accordingly instead

rs/crypto/internal/crypto_lib/basic_sig/ecdsa_secp256r1/tests/api_tests.rs
```diff
@@ -106,7 +107,9 @@ mod keygen {
 
     #[test]
     fn should_fail_parsing_non_ecdsa_key_without_panic() {
-        let pk_der = base64::decode(crate::ED25519_PK_DER_BASE64).unwrap();
+        let pk_der = BASE64_STANDARD_NO_PAD
+            .decode(crate::ED25519_PK_DER_BASE64)
+            .unwrap();
```
pierugo-dfinity: Ditto

rs/crypto/internal/crypto_lib/basic_sig/ed25519/src/api/tests.rs
```diff
@@ -42,7 +43,7 @@ mod serialization {
 
     #[test]
     fn should_fail_parsing_der_encoded_pk_with_wrong_oid() {
-        let mut pk_der = base64::decode(PK_DER_BASE64).unwrap();
+        let mut pk_der = BASE64_STANDARD_NO_PAD.decode(PK_DER_BASE64).unwrap();
```
pierugo-dfinity: Ditto

pierugo-dfinity (approved):

PR #11489, opened 2026-09-08
fix(cketh): derive the balance-scan batch size from the EIP-3860 initcode limit

gregorydemay:
Reported in the [forum](https://forum.dfinity.org/t/proposal-143820-to-upgrade-the-cketh-minter-canister/75524/4).

## Problem

* The balance scan reads ERC-20 balances with a create-style `eth_call` that runs the deployless batcher as init code.
* EIP-3860 caps initcode at 49152 bytes. With 165 bytes of program, one length word and 64 bytes per call, at most 764 calls fit.
* The batch size was 1000, chosen from the provider gas cap alone.
* A chunk above the limit is rejected by every node, retried whole on the next tick, and fails again. Its deposits are never detected.

## Changes

* The batch size is derived from the batcher encoding instead of a literal, which yields 764 calls today.
* An integration test against anvil checks that a full batch succeeds and one more call is rejected with `max initcode size exceeded`.
* A unit test pins the encoded size of a full batch to the initcode limit.
* The design doc states the bound, links EIP-3860 and describes the chunking.

https://claude.ai/code/session_01AV18LEDjdNim5BecwhcqeV

[DEFI-2996]: https://dfinity.atlassian.net/browse/DEFI-2996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -128,6 +129,42 @@ fn should_read_many_balances_in_a_single_call() {
...
+
+    let full_batch = batch_of(MAX_CALLS_PER_BATCH);
+    let out = anvil
+        .eth_call_create(&dev, &encode_balance_batch(&full_batch))
+        .expect(
+            "a batch of MAX_CALLS_PER_BATCH calls must stay within the EIP-3860 initcode limit",
+        );
+    assert_eq!(
+        decode_balance_batch(&out, full_batch.len()).expect("decode failed"),
+        vec![Erc20Value::ZERO; MAX_CALLS_PER_BATCH]
```
mbjorkqvist: 🤖  No holder is funded, so this expects 764 zeros — and zero is also what almost every way of being wrong returns: a `CODECOPY` offset that walks off the args into zero-padding, a loop that exits early leaving the output region untouched, a memory-expansion bug. All of those produce a correctly-sized all-zero blob and this stays green.

The batcher's riskiest code is the hand-written per-pair offset arithmetic, and the only place it's exercised against real values is `should_read_many_balances_in_a_single_call`, at 32 pairs. Funding one holder at the far end costs a single extra transaction and turns this into "the node accepted it *and* the program addressed pair 763 correctly":

```rust
const LAST_BALANCE: u128 = 123_456;
anvil.fund(&token, &dev, &holder_at((MAX_CALLS_PER_BATCH - 1) as u64), LAST_BALANCE);

// ...

let mut expected = vec![Erc20Value::ZERO; MAX_CALLS_PER_BATCH];
*expected.last_mut().unwrap() = Erc20Value::from(LAST_BALANCE);
assert_eq!(
    decode_balance_batch(&out, full_batch.len()).expect("decode failed"),
    expected
);
```

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -14,29 +14,12 @@ use crate::state::{TaskType, mutate_state, read_state};
 use crate::time::TimeProvider;
 use crate::timed_sized_map::Timestamp;
 use batcher::BalanceOfCall;
+pub use batcher::MAX_CALLS_PER_BATCH;
```
mbjorkqvist: 🤖 Nit: `pub mod batcher` on line 1 already exposes this, so the re-export adds a second public path to the same constant — and the integration test then imports from both, one line apart (`balance_scan::MAX_CALLS_PER_BATCH` on line 11, `balance_scan::batcher::{...}` on line 12).

Suggest dropping this line, using `use batcher::{BalanceOfCall, MAX_CALLS_PER_BATCH};` here, and collapsing the test's two imports into a single one from `batcher`.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -41,6 +41,30 @@ pub const BATCHER_INITCODE: [u8; 165] = [
...
+/// 50M `eth_call` cap providers commonly apply (geth's `--rpc.gascap` default). Payloads stay
+/// small as well: 64 bytes of calldata and 32 bytes of return per call, far below the 2 MiB
+/// HTTPS-outcall limit.
+///
+/// The cap must not be set above what every provider accepts: `scan_balances` splits the
+/// registered pairs into chunks of this size and advances each chunk all-or-nothing, so a
+/// whole-call failure re-does that chunk on the next tick — and a chunk that always exceeds a
+/// provider limit fails *every* time, permanently stalling its pairs.
+pub const MAX_CALLS_PER_BATCH: usize =
+    (MAX_INITCODE_SIZE - BATCHER_INITCODE.len() - WORD) / (2 * WORD);
```
mbjorkqvist: 🤖  The derivation only covers the input side. A create-style call is bounded at both ends: EIP-3860 caps the initcode at 49'152 bytes (764 pairs here), and [EIP-170](https://eips.ethereum.org/EIPS/eip-170) caps the returned blob at 24'576 bytes (768 pairs, since we return one word per call). 764 < 768, so this is safe today, but only by four pairs.

The comment in the design doc promises the cap "follows the batcher if its program ever changes". It follows the program *length*, not the encoding *shape*: pack `(token, holder)` into one word instead of two and this recomputes to `(49152 - 165 - 32) / 32 = 1529`, returning 48'928 bytes - rejected with `max code size exceeded` by every node. That is exactly the permanent-stall failure this PR fixes, with the comment reassuring whoever makes the change that it was handled.

No behaviour change today; suggest deriving from both and taking the smaller:

```rust
const MAX_CODE_SIZE: usize = 24_576;                  // EIP-170, bounds the returned blob
const MAX_INITCODE_SIZE: usize = 2 * MAX_CODE_SIZE;   // EIP-3860, bounds the sent blob

pub const MAX_CALLS_PER_BATCH: usize = {
    let by_initcode = (MAX_INITCODE_SIZE - BATCHER_INITCODE.len() - WORD) / (2 * WORD);
    let by_return_data = MAX_CODE_SIZE / WORD;
    if by_initcode < by_return_data { by_initcode } else { by_return_data }
};
```

(`std::cmp::min` isn't const on stable, hence the block.) Worth a sentence in the design doc alongside the EIP-3860 one.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: 🤖 Done in a050db41df: `MAX_CALLS_PER_BATCH` is now the smaller of the two derivations (`by_initcode_size` vs `by_returned_code_size`, via a const block since `min` is not const-stable), with `MAX_INITCODE_SIZE` defined as `2 * MAX_CODE_SIZE` per EIP-3860. The doc comment now names both bounds and how close they sit (764 vs 768), and the design doc got the matching EIP-170 sentence.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Done in a050db41df: the test now funds `holder_at(763)` with a distinct balance and asserts it comes back in the last slot, so a correctly-sized all-zero blob no longer passes. Verified against anvil via `bazel test //rs/ethereum/cketh/minter:deposit_from_cex` (passes). Worth noting the two failure modes differ: a *token*-offset bug reverts the whole call (address 0 has no code, so the 32-byte return check fails), but a *holder*-offset bug and an early-exiting loop both silently produced zeros before — this catches those.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Done in a050db41df: dropped the re-export, `balance_scan` now imports `MAX_CALLS_PER_BATCH` from `batcher` alongside `BalanceOfCall`, and the integration test imports everything through the single `batcher` path.

PR #11491, opened 2026-09-08
docs(cketh): sync the deposit-from-CEX spec with the shipped ETH deposits and delegate rotation

gregorydemay:
### Why

* ETH deposits from a CEX use the same deposit address as ERC-20 deposits. Ethereum users expect one address per person, not one per (person, token).
* The sweeper delegate can change. Re-pointing the minter at a new delegate is not enough: already-delegated addresses must be re-delegated, at the right nonce.
* ETH deposits and delegate rotation have shipped since this spec was written. The spec had drifted from the code, in these parts and in older ones.

### What

* Records the UX rationale for the shared, permanently delegated address.
* Describes `deposit_eth`, the sibling ETH balance batcher and the ETH sweep as shipped.
* Describes delegate rotation as shipped: the delegation is read from the chain before every sweep, an already-delegated address is swept without an authorization, a re-delegation is signed at a nonce the minter tracks itself.
* States what the tracked nonce guarantees: a lower bound on the chain nonce, so a rotation can land one sweep late, never not at all.
* Aligns registration, the scanning window, caps, minimums, funding, mint, tests, bazel targets and metrics with the code.
* Moves design the code does not implement into a new "Future work" section: deposit fee, sponsored deposits, the `Transfer`-log filter and sender screening, on-demand triggers, dashboard sections.
* Lists rules for future delegate versions, including keeping the minimal payable `receive()`.


rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1443,124 @@ Notes:
...
+fallback), and a reverted sweep drops its deposits from the queue (they are
+not retried until the pair is re-armed). Meanwhile a plain ETH send to such an
+address bounces at the sender (no `receive()` on the old delegate): nothing is
+locked (`R12`), but ETH deposits to it are impossible until it is rotated.
+
+**Design.** Rotation is lazy, per address, and rides the sweeps that would
+happen anyway — an address that never receives another deposit is never
+rotated, costs nothing (`R13`) and is harmless on the old delegate.
+
+1. **Delegation record.** For every deposit address the minter keeps
```
gregorydemay: is this needed? There is already `authorizations: BTreeMap<AuthorizationRequest, TransactionSignature>` in `AutomaticDeposits`

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1443,124 @@ Notes:
...
+   `sweepErc20Batch` any delegate version qualifies (the ABI is stable across
+   versions, rule below); for `sweepEthBatch` the item needs either a record
+   naming the current delegate or a rotation tuple in the same transaction —
+   tuples apply before execution, so one transaction both rotates and sweeps.
+5. **Concurrency.** Two in-flight sweeps may carry the same `(address, n)`
+   tuple; whichever mines first applies it, the other is skipped, and the
+   record processes them in order: first applied (`n → n + 1`), second
+   skipped (`n ≠ n + 1`). Because the minter never signs two different tuples
+   for the same `(address, nonce)`, no ordering can point an address anywhere
+   the minter did not intend.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed, the enqueue design did not enforce the invariant it relied on. Fixed in 16ea366434: the signing step looks the authorization store up by (address, nonce) across delegates, so at most one tuple per (address, nonce) is ever signed. While a tuple for (address, n) naming another delegate exists and is not yet marked applied, an ERC-20 item re-carries that tuple (both delegates serve `sweepErc20`, and the tuple is idempotent) and an ETH item waits in the queue until that sweep is finalized; the rotation is then signed at n + 1. Row 3 of the table now points at this rule.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1230,9 +1268,20 @@ the on-chain account, not to key derivation, so each sibling has its own lane:
...
+  authority's nonce. Since the minter is the only holder of the key, **the nonce
+  is exactly the number of the address' authorizations ever applied**, and the
+  minter signs at most one tuple per `(address, nonce)`, always naming the
+  delegate configured at signing time — so at any moment at most one live
+  (applicable) tuple exists per address, and it points where the minter wants.
+  As shipped, every first authorization names nonce 0 and is recorded under
+  `(chain, delegate, nonce)`; a sweep re-carries the recorded tuple of every
+  address it touches, and the protocol skips the stale ones (an address at
+  nonce 1 is already delegated), at the cost of the 25'000-gas tuple charge
+  minus its 12'500 refund. Tracking each address' delegation from the minter's
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct, verified against the EIP: the nonce check (step 6) precedes the existing-account refund (step 7) and a failing step ends the tuple, so a skipped tuple costs the full 25 000 gas. Fixed in 16ea366434 at both places. The same mistake sits in the `SWEEP_GAS_PER_AUTHORIZATION` doc comment in `state/transactions/mod.rs`; the 40 000 budget still covers it, so that comment will be corrected in a code PR rather than here.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1443,124 @@ Notes:
...
+  (`sweepErc20`, `sweepErc20Batch`, `sweepEth`, `sweepEthBatch`) unchanged;
+  a version that must break one of them makes every sweep through it a
+  rotating sweep (row 3 above) until the population has moved.
+* Keep the delegate storage-less: rotation then leaves nothing behind at the
+  address, and `SELF`-style guards keep working per instance.
+* A new delegate wired to the **same helper** needs no new attestations: the
+  attestation digest names the helper, never the delegate. A new helper needs
+  both a new delegate (the helper is one of its immutables) and, since
+  attestations are keyed by helper, one new attestation per address at its
+  next sweep — the same lazy pattern, one more signature.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Agreed. R18 is now scoped in 16ea366434: replacing the delegate alone needs no new attestation (the attestation names the helper, not the delegate); replacing the helper, which also replaces the delegate wired to it, additionally re-attests each address lazily at its next sweep.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Not as a separate structure. The existing map records what was signed, and signed is not applied: a tuple may sit in a sweep still in flight or never sent, and after a delegate change the map holds tuples for both delegates at the same nonce. What is missing is only which stored authorizations were applied, and by which sweep, derived from the finalized sweeps. From that mark the delegation follows: the applied tuple with the highest nonce names the current delegate and the nonce is one above it. Reworded in 16ea366434 so the record is the existing store annotated with an applied mark, not a new `delegation` field.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1448,142 @@ Notes:
...
+   sweep** — derived from the minter's own `FinalizedSweeperTransaction`
+   events, whatever their status: walking the mined transaction's
+   authorization list, a tuple whose nonce equals the address' current nonce
+   (the number of its tuples applied so far, 0 initially) is marked applied,
+   any other was skipped. An address' delegation then follows without a
+   second structure: its applied tuple with the highest nonce names the
+   current delegate, and its nonce is one above it; an address with no applied
+   tuple is not delegated. Sweeps are finalized in sweeper-nonce order, so the
+   replay is deterministic and reconstructs the marks on upgrade (`R8`); the
+   dashboard shows the resulting delegation per address (`R9`).
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Right about the code, so the sentence is fixed in 61ef889c70; the conclusion does not follow, so the design is unchanged. `create_sweep_transactions_batch` assigns the sweeper nonce in queue order and `reschedule_sweep_request` moves a fee-insufficient request to the back of that queue, while `finalize_transactions_batch` iterates the `BTreeMap<SweepId, receipt>` returned by `fetch_finalized_receipts` — so event order is `SweepId` order and can indeed differ from mining order. The claim "sweeps are finalized in sweeper-nonce order" was therefore wrong and is gone. Sorting the replay by the created transaction's nonce is not needed, though: by rule 5 the minter signs at most one tuple per `(address, nonce)`, and a tuple at `n + 1` is only signed once the tuple at `n` is marked applied, i.e. once its sweep has finalized. Every tuple in flight for one address at any one time is therefore the *same* tuple, so any processing order marks exactly one of them applied at the same nonce naming the same delegate: the reconstructed delegation is order-independent. The text now says that explicitly instead of claiming an order the pipeline does not guarantee.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1453,159 @@ Notes:
...
+   finalized)` with the usual consensus (the `finalized_transaction_count`
+   helper already used for the sweeper's own nonce) and re-anchors the record
+   on it — at `finalized`, not `latest`, so the observation agrees with the
+   finalized receipts every other mark comes from and cannot be reorged out.
+   An observation the record cannot explain (a nonce beyond every tuple the
+   minter signed, or two unapplied tuples at one nonce, possible only in
+   state that predates the one-tuple-per-nonce invariant) is not applied: the
+   address stays excluded from sweeps, is not re-read, and is surfaced by a
+   metric for manual repair.
+   The chain nonce identifies the current delegate without `eth_getCode`: the
```
mbjorkqvist: :robot: The nonce identifies the delegate only when at most one authorization exists per nonce. Today's code files authorizations under the delegate they name and always signs nonce 0 (`state.rs:313-331`), so every address swept both before and after #11449 already holds two at nonce 0 — which makes "two unapplied tuples at one nonce, possible only in state that predates the invariant" the default outcome for that whole population rather than a corner. Those addresses then get excluded from sweeps and surfaced for manual repair, and "manual repair" currently names no mechanism.

Worth noting the ambiguity is confined to this path: ordinary replay is unambiguous, because the log records which sweep carried which authorization, so the tuple carried by the sweep that finalized while the count stood at 0 is the accepted one. It only bites here, after the marks have been discarded.

**Suggestion:** specify a one-time reconciliation when the record ships — replay resolves the duplicates exactly, so mark the applied one and permanently retire the rest. After that the invariant holds for all state and this sentence is true as written. That leaves one case the log genuinely can't answer: a log truncated by a reinstall or a snapshot restore (an ordinary upgrade is safe, since replay rebuilds everything), where every address looks never-delegated and nothing reverts because ERC-20 sweeps keep working. For that, an operator path that sets the record from `eth_getCode` — it returns the designator and names the delegate outright, costs no gas, and batches hundreds of addresses per request through the `eth_call` plumbing the balance scan already uses.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1453,159 @@ Notes:
...
+already at nonce 1. Those addresses silently stay on the old delegate. ERC-20
+sweeps keep working, because both delegates implement `sweepErc20` with the
+same ABI and the same helper — which is exactly what hides the problem — but
+an ETH sweep reverts as a whole batch (the old code has no `sweepEth` and no
+fallback), and a reverted sweep drops its deposits from the queue (they are
+not retried until the pair is re-armed). Meanwhile a plain ETH send to such an
+address bounces at the sender (no `receive()` on the old delegate): nothing is
+locked (`R12`), but ETH deposits to it are impossible until it is rotated.
+
+**Design.** Rotation is lazy, per address, and rides the sweeps that would
```
mbjorkqvist, line 1525: :robot: Lazy rotation has no trigger for an address whose only pending deposit is ETH. No `receive()` on the previous delegate means the transfer fails at the exchange, so there's no ETH balance, no scan detection, and no sweep to ride — and the only sweep such an address can get comes from an ERC-20 balance at or above its minimum. So a user in this state must make a token deposit before an ETH deposit is even possible, which isn't stated. The one mechanism that would break the loop (a tuple-only `0x04`) is explicitly excluded further down as an operational shortcut — fair for rotating everybody up front, but it takes the on-demand single-address case with it.

**Suggestion:** let a *sponsored* `deposit_eth` enqueue the rotation — that mode already has the caller paying the sweep gas with detection and sweeping on demand, so it threads R13 — or make a one-off rotation of the population an explicit delivery step before Phase 2, with an owner. Either way, surface "needs rotation" in the `deposit_eth` status: today a frontend can only tell the user to send ETH and watch it fail at the exchange with no explanation.

The exotic path does work, incidentally: ETH force-fed via `selfdestruct` or a coinbase payment is detected, gated, and rotated in the same transaction. It's the ordinary CEX withdrawal that deadlocks.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1453,159 @@ Notes:
...
+   authorization list, a tuple whose nonce equals the address' current nonce
+   (the number of its tuples applied so far, 0 initially) is marked applied,
+   any other was skipped. An address' delegation then follows without a
+   second structure: its applied tuple with the highest nonce names the
+   current delegate, and its nonce is one above it; an address with no applied
+   tuple is not delegated. Replaying those events rebuilds the marks on upgrade
+   (`R8`). Sweep ids do not follow chain order — a sweep whose prepaid gas no
+   longer covers the fee is moved back in the queue, so a lower `SweepId` can
+   carry a higher sweeper nonce — which is why the minter finalizes the
+   receipts of a batch in the order the chain executed them (block number,
```
mbjorkqvist: This isn't what master does. `fetch_finalized_receipts` returns `BTreeMap<Id, _>` (`withdraw.rs:468-470`) and `finalize_transactions_batch` iterates it, so `FinalizedSweeperTransaction` is emitted in `SweepId` order — the order this same sentence explains is *not* chain order. The stored receipt also carries no `transaction_index` (`eth_rpc_client/responses.rs:11-31`), so sorting by (block, index) would need a field it doesn't have.

**Suggestion:** either mark it as behaviour this work introduces, or specify **sweeper-nonce order** instead. Every sweep is sent from the single sweeper address, so send order *is* chain execution order, and the minter already knows it without reading anything.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1453,159 @@ Notes:
...
+   keys, so a deposit address' nonce advances only when one of the minter's
+   own tuples is applied, and every tuple the minter signs and every sweep it
+   sends is recorded and finalized by receipt — the record therefore equals
+   the chain, and the rotation tuple is signed at the *recorded* nonce with no
+   chain read. The record can drift only through an operational error that
+   makes the minter skip a mined sweep's receipt (e.g. a manual
+   `next_sweeper_transaction_nonce` override past a sent, unfinalized sweep).
+   Drift is not silent: the mistimed tuple is skipped and the ETH sweep it
+   rides reverts. That revert is the **repair trigger**: a reverted sweep
+   invalidates the record of every address it touched, and before any of them
```
mbjorkqvist: Two costs of invalidating every address the reverted sweep touched:

- **Blast radius.** A batch is up to `MAX_DEPOSITS_PER_SWEEP = 10`, reverts atomically, and a reverted sweep drops *all* of its deposits from the queue (`automatic_deposits/mod.rs:210-228`). So one drifted address costs nine others their queue entries — each needing its pair re-armed — plus nine chain reads. **Suggestion:** put rotation-carrying ETH sweeps in their own batch, or batches of one; only those can revert for this reason.
- **ERC-20 items never needed the record** (`sweepErc20` works on either delegate), so blocking them behind the re-read stalls sweeps that would have succeeded, and makes an RPC-provider hiccup a sweeping outage. **Suggestion:** let them through while repair is pending, carrying the recorded authorization — a stale one is simply refused, which also covers the record-ahead-of-chain case.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1453,159 @@ Notes:
...
+   recorded tuple at `nonce − 1` is the one that installed it. Sweeping,
+   rotation included, makes no per-address chain read on the happy path.
+4. **Entry-point gating.** A sweep item is only enqueued if the delegate its
+   address will be running *when the call executes* has the entry point: for
+   `sweepErc20Batch` any delegate version qualifies (the ABI is stable across
+   versions, rule below); for `sweepEthBatch` the item needs either a record
+   naming the current delegate or a rotation tuple in the same transaction —
+   tuples apply before execution, so one transaction both rotates and sweeps.
+5. **Concurrency: one signed tuple per `(address, nonce)`, whatever the
+   delegate.** The signing step enforces this invariant by looking the store
```
mbjorkqvist: :robot: Two notes on this invariant, since the `nonce - 1` lookup in 3 is only sound because of it.

**The key order doesn't support the lookup.** `AuthorizationRequest` is `(account, chain_id, delegate, nonce)` (`tx/eip_7702.rs:62-71`), so a fixed `(account, chain, nonce)` across delegates isn't a range query — the natural implementation is a per-delegate lookup, which enforces nothing while looking correct. **Suggestion:** re-key to `(account, chain, nonce, delegate)`, state the requirement where 3 relies on it, and unit-test it.

**The case it protects against isn't only the `UpgradeArg` switch.** An authorization is a standalone signed object, public once broadcast, and the transaction carrying it needn't concern the address it delegates — so a stranger can submit one in their own transaction (the design already expects strangers to sweep and donate gas), moving the address's nonce with no sweep of the minter's own to record it. Worth stating, along with why it's benign: the stranger can only install the delegate the minter chose, which is the same argument the code already makes for the attestation ("a public fact rather than a secret", `attestation/mod.rs:6-8`). It's benign *because* of this invariant — with a per-delegate lookup it becomes 3's permanent failure with no operator error anywhere.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -626,9 +658,13 @@ key) handed to the sweeper, and is no longer re-scanned; the pair's siblings —
 other tokens at the same address — keep scanning, and the rest cost nothing
 further this tick. A balance is only ever a trigger, never a source of truth (see
 the screening discussion below). For native ETH (Phase 2), the batcher reads the
-address' ETH balance in the same latest-block call; the balance delta is only
-the sweep trigger (`R11`) — there are no logs to confirm against, and the mint
-follows the sweep's finalized helper event like any other deposit (step 4).
+address' ETH balance in the same latest-block call: an `(address, ETH)` pair is
+appended to the calldata like a token pair, with the zero address in the token
+slot, and the program answers it with the `BALANCE` opcode on the holder instead
```
mbjorkqvist: :robot: Today a zero in the token slot is an impossible value that fails loudly. The batcher reverts the whole `eth_call` if any sub-call returns anything other than exactly 32 bytes — which is precisely what a call to a non-contract address does — and that is deliberate: *"a sub-call that reverts or does not return exactly 32 bytes (e.g. a non-contract address) is an anomaly, not 'no balance'"* (`balance_scan/batcher/mod.rs:11-29`). The tick fails, the chunk is retried, no state changes.

Giving the zero address a meaning turns that alarm into a valid instruction, so the same underlying mistake — a default-constructed `Address`, a token removed from configuration, a lookup returning zero instead of failing — escalates instead of stopping:

1. the scan returns a plausible number in a slot the minter believes is token X's balance;
2. above the minimum, the pair is promoted out of the watchlist into the sweep queue and its attestation and authorization are signed;
3. the sweep goes out for token X, the delegate calls `balanceOf` on an address with no contract, gets empty return data and **reverts** — dropping the whole batch's deposits from the queue, up to `MAX_DEPOSITS_PER_SWEEP` of them.

A cheap, loud, stateless failure at scan time becomes an expensive one at sweep time that costs other users their queue entries.

**Suggestion:** carry the two-case value the doc already specifies for internal state ("the pair's asset is `ETH | ERC-20(contract)` rather than a bare contract address, so the ETH pair can never be confused with a token whose contract address is zero", step 1) all the way into the encoder — `BalanceOfCall { asset, holder }` rather than `{ token: Address, holder }`. Then the encoder is the only thing that can emit a zero word, and only from an explicit `Eth` variant.

Separately, three consequences of editing the program that the plan doesn't mention:

- `BATCHER_INITCODE`'s credibility rests on being hand-assembled and then **validated byte-identical across all four providers** — create-style `eth_call` with an unusual program is exactly where providers differ. Adding a compare-to-zero branch and a `BALANCE` path changes the bytes, so that validation needs redoing, including re-confirming that a genuinely broken token still reverts now that zero means something.
- `MAX_CALLS_PER_BATCH` is derived from `BATCHER_INITCODE.len()`, and the initcode bound currently binds at 764 against the return-size bound of 768 — "only by four calls". Every 64 bytes added costs one pair per request, so the constant and its stated derivation both move.
- The test plan's new bullets cover the delegation record, anvil rotation and the Phase 2 end-to-end flow, but nothing covers the batcher: no "zero token word returns the holder's ETH balance", no "a non-contract token still reverts the whole call", no re-run of the provider check.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1230,9 +1276,22 @@ the on-chain account, not to key derivation, so each sibling has its own lane:
...
+  minter signs at most one tuple per `(address, nonce)`, always naming the
+  delegate configured at signing time — so at any moment at most one live
+  (applicable) tuple exists per address, and it points where the minter wants.
+  As shipped, every first authorization names nonce 0 and is recorded under
+  `(chain, delegate, nonce)`; a sweep re-carries the recorded tuple of every
+  address it touches, and the protocol skips the stale ones (an address at
+  nonce 1 is already delegated), at the cost of the full 25'000-gas tuple
+  charge: EIP-7702 checks the nonce *before* it grants the existing-account
+  refund, and a failing check ends the tuple's processing, so a skipped tuple
+  earns no refund at all. Tracking each address' delegation from the minter's
```
mbjorkqvist: :robot: This is right, per EIP-7702's step order — the nonce check is step 6, the existing-account refund step 7, and a failed step ends that tuple's processing — and it contradicts `SWEEP_GAS_PER_AUTHORIZATION` in `state/transactions/mod.rs:294-300`, which states that a skipped tuple "is charged the same 25'000 and refunded 12'500 for an authority the state trie already holds".

The 40'000 budget is unaffected either way, so nothing is broken, but the rationale is wrong and today's refused tuples cost twice what it implies. **Suggestion:** fix that comment in the step-6 PR, so the doc and the code don't disagree about the same fact.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -105,8 +109,11 @@ _Requirements are grouped by phase, not numbered sequentially: `R11` and `R12` a
   unsupported ERC-20 tokens, are not credited. No funds are ever burned or destroyed:
   they remain at a tECDSA-controlled address and remain recoverable by the minter.
 * `R5`: Every credited deposit is eventually swept to the minter's main address. A
-  sweep failure or delay never affects already-minted balances; sweeps are retried
-  until confirmed.
+  sweep failure or delay never affects already-minted balances; a sweep that does
+  not mine is resubmitted with a fee bump until it does. A sweep that mines and
```
mbjorkqvist: Two things here:

- Crediting follows a *successful* sweep, so a reverted sweep credited nothing. The exception therefore describes uncredited deposits, which R5's own subject ("every credited deposit") doesn't cover.
- "Swept when their pair is next armed" needs another `deposit_erc20`/`deposit_eth` call from the user or frontend: `record_finalized_sweep_transaction` removes the entry and doesn't re-arm (`automatic_deposits/mod.rs:210-228`).

**Suggestion:** move the exception to R15, which already covers dormant pairs, and say plainly that the minter does not re-arm by itself. Minor: the trailing `R12` is the fixed-21'000-gas requirement — R4 is the one that says funds stay at a tECDSA-controlled address and remain recoverable.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1149,8 +1189,10 @@ against 98'075 for a batch of one. Batching pays once for the 21'000 base, the c
 while each extra address adds only its 25'000 authorization, a warm inner call
 and a transfer that earns the slot-clearing refund. Operational notes: a
 tuple skipped by the protocol (e.g. stale nonce) makes the corresponding inner
-call hit a code-less address, which reverts the *whole* batch (atomic, funds
-safe, gas wasted — retry); mixed batches are fine (tuples only for
+call hit a code-less address, which reverts the *whole* batch — atomic, funds
+safe, gas wasted; the minter does not retry a reverted sweep but drops its
+deposits from the queue, and they are picked up again once their pair is
```
mbjorkqvist, line 1230: :robot: With rotation in the picture this covers two different situations. A tuple skipped for a stale nonce most often means the address is *already delegated* — to the previous delegate — where the inner call runs old code (fine for `sweepErc20`, reverts for `sweepEth`) rather than hitting a code-less address. The code-less case now needs the record to be wrong, or the delegation to have been cleared. **Suggestion:** split the two, since only one of them reverts.

Related: "Transaction 3" above still presents tuple-less later sweeps as established behaviour, which the new section correctly calls new — a pointer there would help.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -626,9 +658,13 @@ key) handed to the sweeper, and is no longer re-scanned; the pair's siblings —
 other tokens at the same address — keep scanning, and the rest cost nothing
 further this tick. A balance is only ever a trigger, never a source of truth (see
 the screening discussion below). For native ETH (Phase 2), the batcher reads the
-address' ETH balance in the same latest-block call; the balance delta is only
-the sweep trigger (`R11`) — there are no logs to confirm against, and the mint
-follows the sweep's finalized helper event like any other deposit (step 4).
+address' ETH balance in the same latest-block call: an `(address, ETH)` pair is
```
gregorydemay: Modify to match #11508, see also this [comment](https://github.com/dfinity/ic/pull/11508#discussion_r3980046144)

gregorydemay: 🤖 Done in 09c806a17c. Filter 1 now describes the ETH read as a *sibling program*: a separate 78-byte init-code blob in its own create-style `eth_call`, pinned to the same latest block as the ERC-20 one, carrying one holder word per pair (no token word), reading with `BALANCE`, with no sub-calls and therefore no failure path, and returning the same flat 32-bytes-per-entry blob so the decoder and the per-call cap stay shared. The paragraph also records why the zero-token-slot overload was rejected. The Phase 2 sequence diagram and the delivery-plan bullet were updated to match.

This is the same change the thread below on the zero token slot asked for, so #11508 settled both.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Accepted, and this is how #11508 shipped it: a second deployless program (`ETH_BATCHER_INITCODE`, 78 bytes) in its own create-style `eth_call`, pinned to the same latest block. The zero token word therefore keeps its current meaning — an impossible value that reverts the whole call — and the ERC-20 init-code bytes, with their four-provider byte-identical validation, are untouched. `MAX_CALLS_PER_BATCH` also stays derived from the ERC-20 encoding (the ETH side is far looser and never binds), so neither the constant nor its stated derivation moves.

Fixed in 09c806a17c: filter 1 now describes the sibling program and records why the zero-slot overload was rejected.

The remaining bullet on test coverage is answered in the code rather than here: #11508 carries the batcher tests (encoding, decode, the both-node-limits bound, and the anvil check of the ETH program).

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Split in f0b23df929. The operational note under Transaction 4 now separates the two: a stale-nonce skip usually means the address is *already delegated* to the previous delegate, so the inner call runs the old code — harmless for `sweepErc20`, which every delegate version implements identically, and a revert for `sweepEth`, which an older one may not have — while only a cleared delegation or a delegation read the minter got wrong leaves the call hitting a code-less address. Both revert paths still take the whole batch, so the consequence sentence is unchanged.

The pointer is in too: Transaction 3 now says the minter only sends a tuple-less sweep once it reads the designator from the chain (row 2 of the rotation section), and that until then every sweep carries a tuple the protocol skips at full price.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Superseded by the rewrite in 2d74f0ac41, which took the second half of this suggestion and made it the design rather than a repair path: which delegate an address runs is no longer inferred from recorded authorizations at all, it is read from the chain before every enqueue — a batched create-style `eth_call` of one `EXTCODECOPY(addr, 0, 32)` per candidate, decoded as no code / designator naming the current delegate / designator naming another delegate. That costs no gas, batches at the same per-call cap as the balance scan, and answers the truncated-log case for free, since nothing is reconstructed from the log.

With that, the sentence this thread was anchored to is gone: there is no "two unapplied tuples at one nonce" corner to reconcile, no population-wide exclusion, and no unnamed manual repair. What stayed internal is only the *nonce* to pin a rotation tuple at — a per-address counter advanced by finalized sweeps and rebuilt on replay — and a counter that is one low merely costs a skipped tuple that the next sweep corrects.

Resolving as addressed; the delegation-read design is under "Changing the sweeper delegate (delegate rotation)".

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 The first half is moot after the rewrite in 2d74f0ac41: nothing looks up an authorization by `(account, chain, nonce)` any more. The delegate is read from the chain (one `EXTCODECOPY(addr, 0, 32)` per candidate in a batched create-style `eth_call`) and the only internal quantity left is the per-address delegation *nonce*, a plain counter — so the `AuthorizationRequest` key order no longer carries an invariant, and there is no per-delegate-lookup trap to re-key against.

The second half survives in a weaker form and I have deliberately not written it into the spec, since it is a design judgement rather than a wording fix: a stranger can indeed replay a minter-signed tuple in their own transaction, bumping the address nonce without a minter sweep being the thing that applied it. Under the chain-read design the damage is bounded and self-healing — the stranger can only install the delegate the minter chose (the tuple is the minter's), the next read shows that delegate, and a counter left one low costs one skipped tuple that the following sweep corrects. @gregorydemay, say the word if you want a sentence on this in the rotation section; I have left the design as it stands.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Agreed on the fact, and the doc side is already right — the rotation section states that the existing-account refund is only granted to tuples that pass the nonce check, which is what makes row 2 of the tuple table worth having.

The code side is out of scope here: this PR touches no `.rs` file, and the comment in question is on `SWEEP_GAS_PER_AUTHORIZATION` in `rs/ethereum/cketh/minter/src/state/transactions/mod.rs`, still saying "refunded 12'500 for an authority the state trie already holds" on master. Leaving this thread open so the finding is not lost, and surfacing it for the PR that touches that constant.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 The mechanics are right, and they survive the rewrite: rotation still rides a sweep, ETH still cannot reach an address delegated to a `receive()`-less delegate, and eager tuple-only `0x04` rotation is still excluded — so an address whose only pending deposit is ETH has no way to rotate itself.

I am not changing the design in this PR: picking between a sponsored `deposit_eth` that enqueues the rotation and a one-off rotation of the population before Phase 2 is @gregorydemay's call, and the "needs rotation" status on `deposit_eth` follows from whichever is chosen. Leaving the thread open so it is decided rather than resolved away.

One narrowing worth recording: this only bites addresses delegated to a delegate without `receive()`. The current delegate has the minimal payable `receive()` (21'095 gas, `R12`), so the population at risk is exactly the addresses still on an older instance at the time Phase 2 ships, not every deposit address.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1394,6 +1454,141 @@ Notes:
...
+address look never-delegated: the next sweep signs and carries a fresh
+`(new delegate, nonce 0)` tuple, which the protocol *skips* on every address
+already at nonce 1. Those addresses silently stay on the old delegate. ERC-20
+sweeps keep working, because both delegates implement `sweepErc20` with the
+same ABI and the same helper — which is exactly what hides the problem — but
+an ETH sweep reverts as a whole batch (the old code has no `sweepEth` and no
+fallback), and a reverted sweep drops its deposits from the queue (they are
+not retried until the pair is re-armed). Meanwhile a plain ETH send to such an
+address bounces at the sender (no `receive()` on the old delegate): nothing is
+locked (`R12`), but ETH deposits to it are impossible until it is rotated.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Same finding as the open thread a few lines up (https://github.com/dfinity/ic/pull/11491#discussion_r3971891133), and the analysis holds: lazy rotation rides a sweep, an address on a `receive()`-less delegate cannot accumulate an ETH balance to trigger one, and eager tuple-only `0x04` rotation is deliberately excluded — so such an address needs an ERC-20 deposit before an ETH one is possible.

Choosing the trigger (a sponsored `deposit_eth` that enqueues the rotation, or a one-off rotation of the population as a delivery step before Phase 2) is a design decision I am not taking in this PR; @gregorydemay owns it, and the "needs rotation" status on `deposit_eth` follows from whichever is picked. Keeping both threads open rather than resolving them.

Scope, for the record: only addresses still on a delegate without `receive()` are affected. The current delegate has the minimal payable `receive()` (21'095 gas, `R12`), so the exposure is the addresses left on an older instance when Phase 2 ships, not the whole population.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1409,6 +1487,141 @@ Notes:
...
+
+   Because the read is at `latest`, it can show a delegation whose sweep the
+   minter has not finalized yet. The rotation tuple is then signed one nonce
+   low, the protocol skips it, and the sweep after that finalization rotates
+   correctly.
+4. **Entry-point gating** stays with the Phase 2 ETH work; what makes it
+   possible is that tuples apply before execution, so a rotation tuple in the
+   same transaction is what lets an ETH sweep run on the new code.
+
+**Operational note: switch the delegate when no sweep is in flight.** One race
```
mbjorkqvist: :robot: "The window is the operator's upgrade, so the remedy is operational" is the part I would push on — not the window itself, which is accepted and well described here.

The remedy is a *timing* precondition, and the operator does not own the timing. `ethereum_sweeper_contract_address` arrives as an `UpgradeArg` in an NNS proposal, so the change lands when governance adopts the proposal, not at a moment chosen relative to sweep activity. The precondition may well hold when the proposal is written and not when it executes, days later. Nothing exposes whether it holds either, at any moment: not `get_minter_info`, not the dashboard's sweeper section, and no `cketh_minter_*` metric — every `sweeper_*` metric is about gas funding.

**Staging the change would remove the need for the note.** Rather than asking the operator to pick a moment, defer activation until the pipeline is quiet:

1. The upgrade arg sets a pending slot rather than `sweeper_contract_address`.
2. `create_pending_sweeper_requests` returns early while a change is pending, so no new sweeps are enqueued.
3. Once the pipeline has drained — signed-but-unsent requests included, not just the sent set — a new `ActivatedSweeperContract` event installs the address and sweeping resumes.

At activation no tuple is unfinalized, so the counter equals every address' chain nonce and the first rotation tuple applies. That eliminates this race by construction rather than by operator discipline.

Costs worth recording if you take it: a pause of about one finality period plus a tick, ~13-19 minutes per rotation, during which deposits wait rather than drop; activation has to be its own event, or the replayed configuration would depend on when the drain happened to finish; and a sweep that never finalizes becomes a new way to stall rotation indefinitely, so it needs a timeout or an override arg.

Not suggesting this replaces the *Rules for the next delegate version* below — those stay useful either way, and they already cover the compatibility side (selectors and semantics unchanged, same-helper versus helper-driven re-attestation).

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1409,6 +1487,141 @@ Notes:
...
+stays open. A sweep carrying `(old delegate, n)` is in flight when an upgrade
+switches the configured delegate, and a later tick enqueues the address' other
+asset: the read still shows the old delegate, so the second sweep carries the
+rotation tuple `(new delegate, n)` — the same nonce the in-flight one pins.
+The in-flight sweep sits earlier in the sweeper's nonce lane, so its tuple
+applies and the rotation tuple is skipped. If the second sweep is an ETH one,
+it then calls `sweepEth` on the old code and reverts as a whole batch; its
+deposits leave the queue and are swept once their pair is re-armed (`R5`), the
```
mbjorkqvist: :robot: This sentence is accurate but reads lighter than the outcome. Two things a future reader would want in it:

**How many deposits.** A batch holds up to `MAX_DEPOSITS_PER_SWEEP` (10), and `record_finalized_sweep_transaction` removes the queue entry for *every* account the request named — the loop is over `request.items`, not over the address whose tuple was skipped. So one stale tuple on one rotating address drops up to nine deposits belonging to addresses that were correctly delegated and would have swept fine.

**Who re-arms.** "are swept once their pair is re-armed" is passive about the actor: re-arming is a `deposit_eth`/`deposit_erc20` call by the *owner*. The pair left the watchlist when it was funded, so nothing re-scans it, and nothing tells the owner to call again — their last observed status was `AwaitingSweep`. The funds are safe at the address (`R12`) but they wait on a user action, not on the minter.

Suggested shape: "…it then calls `sweepEth` on the old code and reverts as a whole batch; every deposit the sweep named leaves the queue — up to ten, of which only one belongs to the address that was rotating — and each is swept once its owner re-arms the pair (`R5`), the funds staying at the address throughout (`R12`)."

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1409,6 +1487,141 @@ Notes:
...
+   no longer pays the full 25'000 gas of a skipped tuple on every sweep (the
+   existing-account refund is only granted to tuples that pass the nonce
+   check), which is the type-`0x02` "Transaction 3" of the
+   [primer](#eip-7702-primer-the-sweep-one-transaction-at-a-time); the sweep's
+   gas budget charges the per-tuple cost only for the items that actually
+   carry one.
+3. **The delegation nonce.** The chain says *which* delegate an address runs;
+   what a rotation tuple still needs is the nonce to pin it at, and that the
+   minter tracks itself — a per-address counter, no `eth_getTransactionCount`
+   on deposit EOAs. The counter equals the address' chain nonce: incoming CEX
```
mbjorkqvist: :robot: "The counter equals the address' chain nonce … so only its own applied tuples do" is now contradicted by the implementation's own documentation, so the spec has ended up behind the code.

A minter-signed tuple is published complete through `get_events` (`EventType::AuthorizedDepositAddress` maps through `request.signed_with(signature)`), and EIP-7702 requires no relationship between the authority and the transaction sender — as this document itself notes in the primer. So anyone can apply a published tuple in their own transaction and advance the address' nonce without a minter sweep being the thing that applied it. `enqueue_sweep` has four `return`s reachable *after* `sign_authorizations_batch` has recorded and published the tuple (missing attestation or sweeper contract, contract changed mid-flight, no signable items, insufficient sweeper gas), so "recorded but never carried" is not a corner case.

Commit 0fffd62432 on the delegation-rotation PR already rewrote the code-side doc accordingly: the counter "never runs ahead" but "can run behind it: signed authorizations are public and EIP-7702 lets anyone send one, which costs the address one sweep with a skipped authorization before a rotation lands." The equality claim here should match.

This is also the sentence you offered to add in the key-order thread. Suggested replacement for the equality clause: "The counter never runs ahead of the address' chain nonce — the minter signs only at the tracked nonce, and incoming CEX transfers advance no nonce — but it can run behind: a signed tuple is public and EIP-7702 lets anyone send one, so a tuple the minter recorded and never carried can be applied by a stranger. The cost is bounded: the stranger can only install the delegate the minter chose, and a counter left one low costs that address one skipped tuple before a rotation lands."

rs/ethereum/cketh/docs/deposit_from_cex.md
mbjorkqvist: :robot: The follow-up this thread was held open for has landed. `fd7daf3c2a` ("docs(cketh): charge a skipped authorization its full 25'000", now on the #11550 branch) rewrote `SWEEP_GAS_PER_AUTHORIZATION`: a skipped authorization "still costs the full 25'000: [EIP-7702] refunds 12'500 for an authority that already exists only once the authorization passed every check, including the nonce", with the EIP link added. Doc and code now agree, so this can be resolved.

rs/ethereum/cketh/docs/deposit_from_cex.md
mbjorkqvist: :robot: Answered by 09c806a17c, which is in the current diff — filter 1 describes the ETH read as a sibling 78-byte program in its own create-style `eth_call` pinned to the same latest block, the zero-token-slot overload is recorded as rejected, and the Phase 2 diagram and delivery bullet match. Nothing outstanding here; resolvable.

rs/ethereum/cketh/docs/deposit_from_cex.md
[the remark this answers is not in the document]
mbjorkqvist: :robot: Same finding as the thread on line 1525, as the reply says, and the decision it needs is the same one. Suggest resolving this one and keeping 1525 as the single place the trigger gets decided, so the discussion is not split across two threads.

rs/ethereum/cketh/docs/deposit_from_cex.md
mbjorkqvist: :robot: Agreed the first half is moot under the chain-read design. On the second half — yes, worth a sentence, and I have drafted one on the counter-equality line above (line 1566), because the same finding came up independently on the rotation PR and 0fffd62432 has since rewritten the code-side doc for it. As written here the spec now claims an equality the implementation's own documentation denies, so this is a wording fix rather than the design judgement it looked like when you deferred it. Once that sentence lands, this thread resolves.

rs/ethereum/cketh/docs/deposit_from_cex.md
```diff
@@ -1409,6 +1487,141 @@ Notes:
...
+sweeps keep working, because both delegates implement `sweepErc20` with the
+same ABI and the same helper — which is exactly what hides the problem — but
+an ETH sweep reverts as a whole batch (the old code has no `sweepEth` and no
+fallback), and a reverted sweep drops its deposits from the queue (they are
+not retried until the pair is re-armed). Meanwhile a plain ETH send to such an
+address bounces at the sender (no `receive()` on the old delegate): nothing is
+locked (`R12`), but ETH deposits to it are impossible until it is rotated.
+
+**Design.** Rotation is lazy, per address, and rides the sweeps that would
+happen anyway — an address that never receives another deposit is never
```
mbjorkqvist: :robot: Wording rather than design — the design question itself is the open thread on 1525, and I am not reopening it here.

This paragraph reads as if lazy rotation covers the whole population: "an address that never receives another deposit is never rotated, costs nothing (`R13`) and is harmless on the old delegate." Two lines up the doc is honest that "ETH deposits to it are impossible until it is rotated", but it never joins the two, and the consequence of joining them is the thing a reader needs: an address on a `receive()`-less delegate has **no self-service path out**. It cannot be rotated by an ETH deposit, because the deposit cannot arrive; so it needs an ERC-20 deposit at or above that token's minimum before an ETH deposit is even possible.

As written, someone reading only the Design paragraph would conclude the lazy mechanism is complete and no trigger is needed. Worth a sentence saying the case is open and which decision closes it — a sponsored `deposit_eth` that enqueues the rotation, or a one-off rotation of the affected population as a delivery step — with Gregory's narrowing recorded alongside it, since it bounds the exposure usefully: only addresses still on a delegate without `receive()` are affected, and the current delegate has the minimal payable `receive()` (21'095 gas, `R12`), so this is whoever is left on an older instance when Phase 2 ships, not every deposit address.

Now that the Phase 2 code has merged, this is the one part of the section describing behaviour that has no decided remedy, which is a good reason to mark it as such in the doc rather than leave it implied.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Settled with @gregorydemay: this was only ever an issue for the ckSepoliaETH minter, whose first sweeper contract had no `receive()`. The production ckETH minter's delegate has had the minimal payable `receive()` from the start, so no production address can be in the state described, and no proactive rotation trigger is added. What the spec gains in f0c22fc66a is the general rule that prevents a recurrence: *Rules for the next delegate version* now requires every future delegate to keep the minimal payable `receive()`, with this very reasoning (rotation rides a sweep, a sweep needs a detected balance, an address whose delegate refuses plain ETH can never accumulate one).

rs/ethereum/cketh/docs/deposit_from_cex.md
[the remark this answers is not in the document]
gregorydemay: 🤖 Same answer as on the thread above (https://github.com/dfinity/ic/pull/11491#discussion_r3971891133): only the ckSepoliaETH minter ever ran a sweeper contract without `receive()`; the production ckETH minter's delegate had it from the start, so no production address enters this state. f0c22fc66a adds "keep the minimal payable `receive()`" to the rules for the next delegate version so that it stays that way.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Both halves are now closed by the shipped code, and f0c22fc66a brings the spec in line. Key order: nothing looks an authorization up across delegates any more; the delegate is read from the chain and the only internal quantity is the per-address `delegation_nonces` counter. Stranger replay: #11554 handles it explicitly, and the spec no longer claims the counter equals the chain nonce. Item 3 of the rotation design now says it is a lower bound: it never runs ahead (the minter only signs at the tracked nonce; a finalized sweep ahead of the counter traps), it can run behind (every signed authorization is public through `get_events` and anyone can send one, as can a manual recovery transaction), and the next sweep carrying an authorization is then skipped but still advances the counter, so a rotation lands late rather than never. The stranger can only install the delegate the minter chose.

rs/ethereum/cketh/docs/deposit_from_cex.md
gregorydemay: 🤖 Fixed on master by #11550 (fd7daf3c2a): the doc comment of `SWEEP_GAS_PER_AUTHORIZATION` now says a skipped authorization still costs the full 25'000, because EIP-7702 grants the 12'500 refund only once the authorization passed every check, including the nonce. Code and spec agree.

PR #11498, opened 2026-09-08
feat: [DSM-147] Restrict XNet pulls to the caller's own subnet

alin-at-dfinity:
Resolve the peer `NodeId` from the TLS client certificate once per connection and inject it into every request's extensions, as `ic_quic_transport` does.

`/api/v1/stream/{subnet_id}` uses it to refuse a caller that the registry does not list as a member of `subnet_id`.

This also unblocks per-node rate limiting and attribution.


rs/http_endpoints/xnet/src/lib.rs
```diff
@@ -474,6 +539,30 @@ fn handle_stream(
...
 }
 
+/// Checks that `node_id` is a member of `subnet_id` at the latest registry
+/// version. If not a member or if the registry read fails, returns an error.
+fn check_subnet_membership(
+    node_id: NodeId,
+    subnet_id: SubnetId,
+    registry_client: &dyn RegistryClient,
+) -> Result<(), String> {
+    let registry_version = registry_client.get_latest_version();
```
schneiderstefan, line 549: I guess that leaves some race conditions. For example, during a subnet split, the nodes that go to the new subnet will not be able to pull slices, even though they are held back to a registry version where they still think they are members. 

Not really an issue overall, as long as neither side crashes or or stops trying altogether, but worth being aware.

rs/http_endpoints/xnet/src/tests.rs
```diff
@@ -551,16 +589,57 @@ async fn handle_stream_nonexistent() {
...
+    assert_eq!(
+        metric_vec(&[(&[("resource", "stream"), ("status", "403")], 1)]),
+        fixture.request_counts()
+    );
+    assert_eq!(0, fixture.slice_payload_size_stats().count);
+    assert!(fixture.response_size_counts().is_empty());
+}
+
+#[tokio::test]
+async fn handle_stream_for_other_subnet() {
```
schneiderstefan, line 615: Should there be some positive tests here, considering most tests otherwise call serve_connection with None?

schneiderstefan (approved):

rs/http_endpoints/xnet/src/tests.rs
alin-at-dfinity: The tests that create a `XNetEndpoint` will indeed end up going through `serve_connection()` with `peer_node_id == None`. But all the other tests explicitly call `route_request(url, Some(DST_SUBNET_NODE))` (while querying the stream for `DST_SUBNET`).

rs/http_endpoints/xnet/src/lib.rs
alin-at-dfinity: True. There was already a race condition whenever a subnet changed its public key (e.g. during subnet recovery; but also during a subnet split), where the registry provides one public key, but the stream is signed with another. In that situation no node can (usefully) pull regardless.

We should probably keep an eye on 403 statuses to see how bad of an issue this is, whenever nodes are replaced or the subnet key changes.

Or do you have an idea for how to better deal with transition phases?

(I suppose not, because the validation context for payload building is pretty clear about which registry version we must validate against. We could be a bit more relaxed about it here, e.g. by checking membership according to the last N registry versions, but that doesn't help when it comes to payload building.)

schneiderstefan: My intuition is also that we should just accept it. At worst, some slice doesn't make it to the next blockmaker, but then gets included slightly later.

PR #11499, opened 2026-09-08
feat(cketh): deposit_eth arms the (account, ETH) pair

gregorydemay:
Stacked on #11494. Second step of the `deposit_eth` flow per the [spec](https://github.com/dfinity/ic/pull/11491): calling `deposit_eth` now arms the `(account, ETH)` pair, so the response reports a `Scanning` status — idempotent across repeated calls and canister upgrades.

* The endpoint's full candid surface lands with the endpoint: a dedicated `DepositEthResponse` and `DepositEthStatus`, whose `AwaitingSweep` carries no contract address (an ETH deposit has none) and stays unreachable until scanning lands in the follow-up PR.
* A registered pair's asset is now `ETH | ERC-20(contract)` instead of a bare contract address, so the ETH pair can never be confused with a token whose contract address is zero. The `ERC-20` variant keeps its CBOR encoding byte-identical to the bare address recorded so far, so existing audit-log events replay unchanged.
* Both deposit endpoints share one per-principal registration guard and one arming flow; ETH counts against the same per-account asset cap and has its own minimum deposit (0.005 ETH).
* The armed ETH pair is not scanned yet: the balance batcher gains its ETH slot in a follow-up PR, so until then its status stays a never-scanned `Scanning`.


rs/ethereum/cketh/minter/src/asset/tests.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[test]
+fn should_decode_a_bare_address_as_erc20() {
+    let encoded = minicbor::to_vec(usdc()).unwrap();
+
+    assert_eq!(
+        minicbor::decode::<Asset>(&encoded).unwrap(),
+        Asset::Erc20(usdc())
+    );
+}
```
gregorydemay: * Turn this into a proptest: have a strategy to generate an arbitrary Asset in test_fixtures and roundtrip.
* move the cbor part into a submodule in rs/ethereum/cketh/minter/src/cbor
* Move the cbor unit tests in rs/ethereum/cketh/minter/src/cbor/tests.rs

rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -301,16 +301,64 @@ pub struct DepositEthArg {
...
+
+impl From<DepositErc20Response> for DepositEthResponse {
+    fn from(response: DepositErc20Response) -> Self {
+        Self {
+            address: response.address,
+            minimum_deposit_amount: response.minimum_deposit_amount,
+            status: response.status,
+        }
+    }
 }
```
gregorydemay: why would that transformation ever be needed? Is they are the same, why not have a single type?

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -403,8 +408,13 @@ impl AutomaticDeposits {
                     }
                 }
             };
+            let token = match request.asset {
+                Asset::Erc20(token) => token,
+                Asset::Eth => return None,
```
gregorydemay, line 412: Here and below for `Asset::Eth => `: let's use a panicking todo `todo!` instead of silently dropping it.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -205,30 +206,32 @@ async fn minter_address() -> String {
...
     let caller = validate_caller_not_anonymous();
+    // Held for the whole call, including across the ECDSA public key fetch in `arm_deposit`, so
+    // that the status check and the registration that follows it cannot be interleaved with
+    // another deposit registration from the same principal.
+    let _guard = deposit_registration_guard(caller).unwrap_or_else(|e| {
+        ic_cdk::trap(format!(
+            "Failed retrieving guard for principal {caller}: {e:?}"
+        ))
+    });
     let DepositMode::Unsponsored { subaccount } = arg.mode;
```
gregorydemay: do a match so new variant are detected by the compiler.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -252,8 +255,17 @@ async fn deposit_erc20(arg: DepositErc20Arg) -> Result<DepositErc20Response, Dep
         owner: caller,
         subaccount,
     };
-    let request = DepositRequest::new(account, token);
-    let minimum_deposit_amount = min_deposit(&token);
+    Ok(arm_deposit(account, Asset::Erc20(token)).await?)
+}
+
+/// Report the `(account, asset)` pair's deposit status, arming the pair first if it is not armed
+/// yet. Callers hold a [`deposit_registration_guard`] across the call.
```
gregorydemay: no Rust doc on private methods

```suggestion
```

rs/ethereum/cketh/minter/src/asset/tests.rs
gregorydemay: 🤖 Done in 1c9894d3b5: the Encode/Decode impls moved to `src/cbor/asset.rs`, the round-trip and byte-identical-encoding tests are proptests in `src/cbor/tests.rs` over a new `arb_asset` strategy in `test_fixtures::arb`. `asset/tests.rs` keeps only the Display/FromStr tests.

rs/ethereum/cketh/minter/src/endpoints.rs
gregorydemay: 🤖 Merged into a single `DepositResponse` returned by both endpoints (Rust and candid); the From impl is gone.

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Done — `deposit_eth` now uses the same match as `deposit_erc20`. (A new variant would also have turned the irrefutable let into a compile error, but the match keeps the two endpoints consistent.)

rs/ethereum/cketh/minter/src/main.rs
gregorydemay: 🤖 Removed.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Applied the `todo!` in `requests_batch` — an ETH sweep entry is unreachable until `sweepEthBatch` lands, so a silent skip there would hide unfinished work.

Kept the skip in `scan_targets_iter`, deliberately: as of this PR `deposit_eth` arms ETH pairs, and the balance-scan timer walks every armed pair each tick. A `todo!` there traps the shared scan task as soon as any user arms one ETH pair, which stops scanning for **all** ERC-20 pairs for the 24h the ETH pair stays armed — a cheap DoS on the whole pipeline. Armed-but-unscanned is the specified interim behavior, documented on `scan_targets_iter` and pinned by `should_not_scan_the_eth_pair` and the `should_not_yield_a_target_for_an_eth_pair` unit test; the batcher PR on top of this one removes the skip. If you prefer the loud marker anyway, the alternative is to not arm ETH pairs in this PR and move the arming into the batcher PR.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -403,8 +407,13 @@ impl AutomaticDeposits {
...
+            };
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct observation, but deliberate and out of scope for this PR. The description states it explicitly: "The armed ETH pair is not scanned yet: the balance batcher gains its ETH slot in a follow-up PR, so until then its status stays a never-scanned `Scanning`." The `Asset::Eth => return None` in `scan_targets_iter` is exactly that placeholder; scanning lands in the follow-up PR #11508.

On the severity: no user funds can be lost. `deposit_eth`'s address is derived by the minter, which holds the corresponding key, so ETH sent there stays sweepable once the ETH scan slot lands — nothing about this PR makes it unrecoverable. Leaving the thread open so it is re-checked against #11508, where the fix belongs.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -403,8 +407,13 @@ impl AutomaticDeposits {
                     }
                 }
             };
+            let token = match request.asset {
+                Asset::Erc20(token) => token,
+                Asset::Eth => return None,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Deliberate and out of scope for this PR. The description defers it explicitly: "The armed ETH pair is not scanned yet: the balance batcher gains its ETH slot in a follow-up PR, so until then its status stays a never-scanned `Scanning`." The `Asset::Eth => return None` arm is exactly that placeholder, and the scanning path lands in the follow-up PR #11508 that stacks on this one.

Note also that this is the middle layer of a stack, so nothing here reaches a released minter on its own — the endpoint and its scanning arrive together. Leaving the thread open so it is re-checked against #11508, where the fix belongs.

rs/ethereum/cketh/minter/src/endpoints.rs
```diff
@@ -298,19 +339,107 @@ pub struct DepositEthArg {
...
-#[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
-pub struct DepositEthResponse {
-    /// The Ethereum deposit address derived for the caller.
-    pub address: String,
-}
-
 #[derive(CandidType, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub enum DepositEthError {
+    /// The account already has the maximum number of assets armed.
+    TooManyTokensForAccount,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The naming point is fair in isolation, but `TooManyTokensForAccount` is not newly invented here — it mirrors the existing `DepositErc20Error::TooManyTokensForAccount` (`endpoints.rs:448`, `cketh_minter.did:533`), which landed with the per-(deposit address, ERC-20) granularity work in #11128. Both candid variants map from the same internal `RegisterDepositError::TooManyAssetsForAccount` (`endpoints.rs:421` and `:434`), so the internal type already carries the accurate name.

That leaves two options, neither of which belongs in this PR:
* rename only the new `DepositEthError` variant — then one condition has two different names across the two deposit endpoints, which is worse for clients than the current wart;
* rename both — a breaking change to the already-published `DepositErc20Error`, well outside "second step of the `deposit_eth` flow".

Leaving this unresolved for @gregorydemay to decide, since a coordinated rename of both variants is a deliberate API break rather than a fix to this layer.

rs/ethereum/cketh/minter/src/main.rs
```diff
@@ -904,7 +918,7 @@ fn get_events(arg: GetEventsArg) -> GetEventsResult {
                         .map(|r| CandidDepositAddressRegistration {
                             owner: r.owner,
                             subaccount: r.subaccount,
-                            erc20_contract_address: r.erc20_contract_address.to_string(),
+                            erc20_contract_address: r.asset.to_string(),
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The `"ETH"` string is not an accident of `Display`: it is a deliberate, round-trippable encoding of `Asset`. See `src/asset/mod.rs:34-51` — `Display` writes `ETH` for `Asset::Eth` and the raw address otherwise, and the matching `FromStr` parses `"ETH"` back to `Asset::Eth` and anything else as an `Address`. So the field carries a well-defined asset identifier, not a malformed address, and a consumer can decode it unambiguously (an ERC-20 address is always `0x`-prefixed hex, never the literal `ETH`).

What you are really flagging is that the *candid field name* `erc20_contract_address` now lags behind the widened value domain, and that is a fair observation. But renaming it — or reshaping it into an `asset` variant — is a breaking change to the published `get_events` payload (`cketh_minter.did:827-845`), it touches `AutomaticDepositReceived` as well (which only starts carrying ETH in the follow-up #11508), and it would force a restack of the layer above this one. That is an API design decision for @gregorydemay, not a fix this layer should make unilaterally, so I am leaving the thread open rather than changing the event surface here.

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/endpoints.rs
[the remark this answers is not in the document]
gregorydemay: Leave as is

rs/ethereum/cketh/minter/src/main.rs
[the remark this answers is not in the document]
gregorydemay: agreed, this is bad and similar to this [comment](https://github.com/dfinity/ic/pull/11508#discussion_r3977078888). Note that breaking changes for the candid API in the return type of get_events is acceptable. I suggest to have a candid pendant to `Asset` to use in `DepositAddressRegistration`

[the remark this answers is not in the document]
gregorydemay: 🤖 Done: the events module gains a candid `Asset` (`variant { Eth; Erc20 : text }`) with conversions to and from the internal `Asset`. It replaces `erc20_contract_address` in the `RegisteredDepositAddresses` registrations and in `AutomaticDepositReceived`, and also the sweep request payload's `token` field, so the whole `get_events` surface breaks once instead of once per PR. The upper stack layers were restacked: the ETH detection test now asserts `Asset::Eth` on the event, and #11532's sweep mapping emits the variant.

PR #11501, opened 2026-09-08
feat(cketh): sweep an already-delegated address without a tuple

gregorydemay:
### Why

* Every sweep re-carries the nonce-0 authorization of each address it touches. On an address already delegated the tuple is skipped by the protocol, but it still costs the full 25'000 gas (EIP-7702 grants the existing-account refund only after the nonce check).
* With the applied marks of #11496 the minter knows which addresses already point at the configured sweeper contract.

### What

* A target whose recorded delegation names the configured sweeper contract gets no authorization. Its sweep item carries no tuple.
* A sweep whose items all carry no tuple is a plain EIP-1559 transaction; one with at least one tuple stays EIP-7702.
* The gas budget counts the authorization cost only for items that carry a tuple.
* A target delegated to another sweeper contract keeps today's nonce-0 request, in a branch made explicit for the rotation PR that follows.

Second PR of this stack, stacked on #11496.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state.rs
+++ b/rs/ethereum/cketh/minter/src/state.rs
@@ -308,34 +308,46 @@ impl State {
         )
     }
 
-    /// What every deposit address in `accounts` authorizes to let the configured sweeper contract
-    /// sweep it: the tuple naming this minter's chain, that contract, and nonce zero. `None` while
-    /// no sweeper contract is configured.
-    ///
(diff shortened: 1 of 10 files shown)


gregorydemay:
🤖 Closing: this stack is superseded by a leaner design in which the chain is the source of truth for each deposit address' delegate (a batched EXTCODECOPY read before every sweep) instead of a record derived from applied tuples. Replacement stack starts at #11549 on top of #11532. The branch is kept for reference.

PR #11508, opened 2026-09-09
feat(cketh): scan and detect ETH deposits

gregorydemay:
Stacked on #11499. Third step of the `deposit_eth` flow per the [spec](https://github.com/dfinity/ic/pull/11491): an armed `(account, ETH)` pair is now actually scanned and, when funded with at least 0.005 ETH, detected and queued for sweeping — `deposit_eth` reports `AwaitingSweep`.

* A dedicated 78-byte deployless batcher program reads ETH balances with the `BALANCE` opcode and returns them in the same flat layout as the ERC-20 batcher, so the decoder is shared. The validated ERC-20 program is byte-identical — chosen over extending it with a zero-token-slot branch, which would have re-assembled the whole program. Validated against a live anvil node, including agreement with the node's own view of every balance.
* The scan tick partitions its due pairs by asset kind at the watchlist — the one place the runtime `Asset` is read — into `ScanTarget<Erc20Asset>` and `ScanTarget<EthAsset>`, so a pair cannot reach the wrong batcher by construction. Both batches are pinned to the same latest block under the same timer guard, and each batch's outcomes are recorded before the next batch is awaited, so funds already observed on-chain never hinge on a later await resuming.
* Queued ETH entries are not batched into sweeps yet: the `sweepEthBatch` lane comes in a follow-up (behind the delegation-record work per the spec's delivery order), which is also why the sweep-batching `todo!` became an explicit skip. The spec's phase-2 wording (zero address in the token slot) needs a small amendment in #11491 to record the separate-program choice.

### DIFF
--- a/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
+++ b/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
@@ -53,7 +53,7 @@ const MAX_CODE_SIZE: usize = 24_576;
 /// [EIP-3860]: https://eips.ethereum.org/EIPS/eip-3860
 const MAX_INITCODE_SIZE: usize = 2 * MAX_CODE_SIZE;
 
-/// Maximum number of `balanceOf` sub-calls in a single deployless-batcher `eth_call`.
+/// Maximum number of balance reads in a single deployless-batcher `eth_call`, for both batchers.
 ///
 /// A create-style `eth_call` is bounded at both ends: the initcode it carries is rejected beyond
 /// [`MAX_INITCODE_SIZE`] (EIP-3860), and the blob the program `RETURN`s is rejected beyond
@@ -73,6 +73,12 @@ const MAX_INITCODE_SIZE: usize = 2 * MAX_CODE_SIZE;
 /// registered pairs into chunks of this size and advances each chunk all-or-nothing, so a
 /// whole-call failure re-does that chunk on the next tick — and a chunk that always exceeds a
 /// provider limit fails *every* time, permanently stalling its pairs.
+///
+/// The value is derived from the ERC-20 encoding, yet it caps [`ETH_BATCHER_INITCODE`] batches
+/// too, with margin on both ends. The returned-blob term is shared — both programs return one
+/// word per entry — so the cap can never exceed `MAX_CODE_SIZE / WORD`; and the ETH initcode
+/// side is far looser than the ERC-20 one (one word per entry after a 78-byte program, so 1532
+/// reads) and never binds. `full_batch_of_eth_balance_reads_fits_both_node_limits` pins this.
 pub const MAX_CALLS_PER_BATCH: usize = {
     let by_initcode_size = (MAX_INITCODE_SIZE - BATCHER_INITCODE.len() - WORD) / (2 * WORD);
     let by_returned_code_size = MAX_CODE_SIZE / WORD;
--- a/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
+++ b/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
@@ -83,6 +83,26 @@ pub const MAX_CALLS_PER_BATCH: usize = {
     }
 };
 
+/// Deployless ETH balance-batcher creation bytecode.
+///
+/// The native-ETH sibling of [`BATCHER_INITCODE`], executed the same way (create-style
+/// `eth_call`, `to` omitted). It reads its inputs from the calldata appended right after this
+/// bytecode (`[n][ holder x n ]`, one 32-byte word each) and, for each holder, reads its ETH
+/// balance with the `BALANCE` opcode — no sub-calls, so unlike the ERC-20 batcher nothing here
+/// can fail and the program has no revert path. On success it returns the balances as a flat
+/// `n x 32`-byte array (no ABI array header), decoded positionally by [`decode_balance_batch`].
+///
+/// The program is fixed regardless of `n` (only the appended args grow). It was assembled from
+/// the opcode listing in `eth_initcode_matches_readable_assembly` and validated against a live
+/// anvil node; see `rs/ethereum/cketh/minter/tests/deposit_from_cex.rs`.
+pub const ETH_BATCHER_INITCODE: [u8; 78] = [
+    0x60, 0x20, 0x61, 0x00, 0x4e, 0x60, 0x00, 0x39, 0x60, 0x00, 0x60, 0x20, 0x52, 0x5b, 0x60, 0x00,
+    0x51, 0x60, 0x20, 0x51, 0x10, 0x15, 0x61, 0x00, 0x44, 0x57, 0x60, 0x20, 0x60, 0x20, 0x51, 0x60,
+    0x20, 0x02, 0x61, 0x00, 0x6e, 0x01, 0x60, 0x40, 0x39, 0x60, 0x40, 0x51, 0x31, 0x60, 0x20, 0x51,
+    0x60, 0x20, 0x02, 0x60, 0x60, 0x01, 0x52, 0x60, 0x20, 0x51, 0x60, 0x01, 0x01, 0x60, 0x20, 0x52,
+    0x61, 0x00, 0x0d, 0x56, 0x5b, 0x60, 0x00, 0x51, 0x60, 0x20, 0x02, 0x60, 0x60, 0xf3,
+];
+
 /// Function selector for `balanceOf(address)`, i.e. `keccak256("balanceOf(address)")[..4]`.
 /// Embedded in [`BATCHER_INITCODE`] right after its leading `PUSH32` opcode; asserted by tests.
 #[cfg(test)]
@@ -111,6 +131,18 @@ pub fn encode_balance_batch(calls: &[BalanceOfCall]) -> Vec<u8> {
     for call in calls {
         out.extend_from_slice(&left_padded_address(&call.token));
         out.extend_from_slice(&left_padded_address(call.holder.as_address()));
+    }
+    out
+}
+
+/// Build the create-call `input` for a batch of ETH balance reads:
+/// `ETH_BATCHER_INITCODE ++ [n] ++ [ holder x n ]`, every value a 32-byte word.
+pub fn encode_eth_balance_batch(holders: &[DepositAddress]) -> Vec<u8> {
+    let mut out = Vec::with_capacity(ETH_BATCHER_INITCODE.len() + WORD * (1 + holders.len()));
+    out.extend_from_slice(&ETH_BATCHER_INITCODE);
+    out.extend_from_slice(&word_from_usize(holders.len()));
+    for holder in holders {
+        out.extend_from_slice(&left_padded_address(holder.as_address()));
     }
     out
 }
--- a/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
+++ b/rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
@@ -153,13 +153,14 @@ pub fn encode_eth_balance_batch(holders: &[DepositAddress]) -> Vec<u8> {
     out
 }
 
-/// Decode the flat `n x 32`-byte return blob into `n` balances, in call order.
+/// Decode the flat `n x 32`-byte return blob of either batcher into `n` balances, in call order.
 ///
-/// Every entry is a genuine `balanceOf` result: the batcher reverts the whole
-/// call if any sub-call fails, so a successful return means all `n` balances are
-/// present (a failed batch surfaces as an `eth_call` error upstream, never as a
-/// `0` here). Returns `Err` if the blob length is not exactly `n` words; never
-/// panics.
+/// Every entry is a genuine balance, never a `0` standing in for a failure, because neither
+/// program can return a partial result: [`BATCHER_INITCODE`] reverts the whole call if any
+/// `balanceOf` sub-call fails, and [`ETH_BATCHER_INITCODE`] has no failure path at all — the
+/// `BALANCE` opcode makes no sub-calls. A failed batch therefore surfaces as an `eth_call` error
+/// upstream, never as a `0` here. Returns `Err` if the blob length is not exactly `n` words;
+/// never panics.
 pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, BatcherDecodeError> {
     let expected = n * WORD;
     if ret.len() != expected {
--- a/rs/ethereum/cketh/minter/src/balance_scan/mod.rs
+++ b/rs/ethereum/cketh/minter/src/balance_scan/mod.rs
@@ -3,7 +3,7 @@ pub mod batcher;
 #[cfg(test)]
 mod tests;
 
-use crate::asset::Asset;
+use crate::asset::{Asset, Erc20Asset, EthAsset};
 use crate::deposit_address::DepositAddress;
 use crate::eth_rpc_client::{AnyOf, MIN_ATTACHED_CYCLES, ToReducedWithStrategy, rpc_client};
 use crate::guard::TimerGuard;
@@ -48,18 +48,13 @@ async fn scan<R: Runtime, T: TimeProvider>(
             return;
         }
     };
-    let (erc20_targets, eth_targets, watchlist_len) = read_state(|s| {
-        let mut erc20 = Vec::new();
-        let mut eth = Vec::new();
-        for target in s.automatic_deposits.scan_targets_iter(now, latest_block) {
-            match target.asset() {
-                Asset::Erc20(token) => erc20.push((target, token)),
-                Asset::Eth => eth.push(target),
-            }
-        }
-        (erc20, eth, s.automatic_deposits.watchlist_len())
+    let (targets, watchlist_len) = read_state(|s| {
+        (
+            s.automatic_deposits.due_scan_targets(now, latest_block),
+            s.automatic_deposits.watchlist_len(),
+        )
     });
-    if erc20_targets.is_empty() && eth_targets.is_empty() {
+    if targets.is_empty() {
         log!(
             DEBUG,
             "[balance_scan] SKIPPING: 0/{watchlist_len} deposits ready to be scanned"
@@ -67,9 +62,9 @@ async fn scan<R: Runtime, T: TimeProvider>(
         return;
     }
 
-    let outcomes = scan_balances(&erc20_targets, latest_block, &client).await;
+    let outcomes = scan_balances(&targets.erc20, latest_block, &client).await;
     apply_scan_outcomes(outcomes, now, latest_block, time_provider);
-    let outcomes = scan_eth_balances(&eth_targets, latest_block, &client).await;
+    let outcomes = scan_eth_balances(&targets.eth, latest_block, &client).await;
     apply_scan_outcomes(outcomes, now, latest_block, time_provider);
 }
 
@@ -114,7 +109,7 @@ enum ScanOutcome {
 /// Pairs whose chunk failed yield no outcome at all, so a failed chunk is retried next tick rather
 /// than silently advanced until its next scheduled slot.
 async fn scan_balances<R: Runtime>(
-    due: &[(ScanTarget, Address)],
+    due: &[ScanTarget<Erc20Asset>],
     latest_block: BlockNumber,
     client: &EvmRpcClient<R, CandidResponseConverter, DoubleCycles>,
 ) -> Vec<ScanOutcome> {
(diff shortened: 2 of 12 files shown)


rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -83,9 +87,11 @@ async fn scan<R: Runtime, T: TimeProvider>(
...
     // Await every balance read first, then apply the outcomes in a single `mutate_state`: the state
     // borrow must never span an await, and re-reading the watchlist after the outcalls is what this
-    // scan deliberately avoids (a concurrent `deposit_erc20` can evict a pair whose window closed
-    // mid-scan, which would drop funds already observed on-chain). See `scan_balances`.
-    let outcomes = scan_balances(&targets, latest_block, client).await;
+    // scan deliberately avoids (a concurrent registration can evict a pair whose window closed
+    // mid-scan, which would drop funds already observed on-chain). See `scan_balances`. Both scans
+    // read at the same `latest_block`, so an address' assets are observed at one block.
+    let mut outcomes = scan_balances(&erc20_targets, latest_block, &client).await;
+    outcomes.extend(scan_eth_balances(&eth_targets, latest_block, &client).await);
```
gregorydemay: The 2nd await could fail, better to register any outcomes before the 2nd await, i.e.
```
let outcomes = scan_balances(&erc20_targets, latest_block, &client).await;
//mutate state to register outcomes

let outcomes = scan_eth_balances(&eth_targets, latest_block, &client).await
//mutate state to register outcomes

```

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -665,7 +659,6 @@ pub enum RegisterDepositError {
 #[derive(Clone, Copy, Debug)]
 pub struct ScanTarget {
     request: DepositRequest,
-    token: Address,
```
gregorydemay, line 682: why this change? This should just be the `Asset` enum

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Done in 83965fed3b: each batch's outcomes are applied in their own `mutate_state` before the next batch is awaited (`apply_scan_outcomes`, called after each scan). Pinned by a new integration test that answers only the ERC-20 batch, asserts the ERC-20 pair already advanced while the tick is still suspended on the unanswered ETH batch, then answers the ETH batch.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 The asset is there, just not as its own field: `ScanTarget.request` (a `DepositRequest`) already carries the `Asset`, exposed by the `asset()` accessor a few lines below — a separate `asset: Asset` field would store the same datum twice. The removed `token: Address` was the previous PR's interim design, where the iterator filtered ETH pairs out and pre-extracted the ERC-20 contract address; with ETH pairs scannable, that extraction moved to the partition in `scan()`, where the two batch builders split. Happy to add an explicit `asset: Asset` field mirroring the request if you prefer the field over the accessor.

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -207,7 +208,70 @@ mod deposit_eth {
...
+        let ckerc20 = CkErc20Setup::default().add_supported_erc20_tokens();
+        let caller = ckerc20.caller();
+
+        let (ckerc20, before) = ckerc20
+            .call_minter_deposit_eth(caller, Some(DEFAULT_USER_SUBACCOUNT))
+            .expect_deposit_response();
+
+        let scanned_at = 4_500_000_u64;
+        ckerc20.refresh_latest_block(scanned_at);
+        ckerc20.run_balance_scan(&[MINIMUM_ETH_DEPOSIT_WEI as u128]);
```
gregorydemay: this should specific for which address there is a balance. The address must come from the response from the `deposit_eth` endpoint

rs/ethereum/cketh/minter/tests/ckerc20.rs
```diff
@@ -207,7 +208,70 @@ mod deposit_eth {
...
+
+        let (_ckerc20, detected) = ckerc20
+            .call_minter_deposit_eth(caller, Some(DEFAULT_USER_SUBACCOUNT))
+            .expect_deposit_response();
+        assert_eq!(detected.address, before.address);
+        let deposit = match &detected.status {
+            DepositStatus::AwaitingSweep(deposit) => deposit.clone(),
+            other => panic!("BUG: expected AwaitingSweep, got {other:?}"),
+        };
+        assert_eq!(deposit.erc20_contract_address, "ETH");
```
gregorydemay: That does not make sense, for ETH there is no contract address. We need a dedicated status type for ETH deposits

rs/ethereum/cketh/test_utils/src/ckerc20.rs
```diff
@@ -213,6 +214,39 @@ impl CkErc20Setup {
         }
     }
 
+    /// Like [`Self::run_balance_scan`], but for a tick where both ERC-20 and ETH pairs are
+    /// due: the scanner sends the ERC-20 batch first and the ETH batch second, each answered
+    /// with its own balances.
+    pub fn run_balance_scan_with_eth(&self, erc20_balances: &[u128], eth_balances: &[u128]) {
+        self.answer_erc20_balance_scan(erc20_balances);
+        self.answer_eth_balance_scan(eth_balances);
```
gregorydemay, line 223: this is way too brittle:

1. Need to match on the payload of the HTTPs outcall to see if an ERC-20 scan balance is made or an ETH one
2. Don't do
```
for _ in 0..MAX_TICKS {
            self.env.tick();
        }
```
this makes the test runs longer that it needs. Only minimum amount of ticks to make it work

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Done in 4e611d1365: `deposit_eth` now returns its own `DepositEthResponse` with a dedicated `DepositEthStatus` — its `AwaitingSweep` carries only `scanned_balance` and `detected_at_block`, no contract address. `deposit_status` in the state layer now reports its own types (`DepositStatusInfo`/`DepositStage`) and the candid layer shapes them per endpoint, so only `deposit_erc20` names a token. Note the candid `get_events` payloads (`AutomaticDepositReceived`, `RegisteredDepositAddresses`) still render the ETH asset as `"ETH"` in their shipped `erc20_contract_address` field — changing those records is a breaking candid change; say the word if you want a follow-up there.

rs/ethereum/cketh/minter/tests/ckerc20.rs
gregorydemay: 🤖 Done in 4e611d1365: the test now passes `(address, balance)` pairs, with the address taken from the `deposit_eth` response; the harness answers each scanned holder by looking its address up (an address absent from the map reports 0).

rs/ethereum/cketh/test_utils/src/ckerc20.rs
gregorydemay: 🤖 Done in 4e611d1365: (1) each batch is identified by the batcher program in its `eth_call` calldata — the mock matcher gained `with_eth_call_input_prefix`, and the harness parses the pending request's holders to build the response, so nothing depends on call order; (2) the blind `MAX_TICKS` loops are gone — the helpers settle with two ticks after the responses (the pre-existing `run_balance_scan` had the same loop; its rewrite drops it too).

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -564,7 +568,7 @@ impl AutomaticDeposits {
         {
             let token = match deposit_request.asset {
                 Asset::Erc20(token) => token,
-                Asset::Eth => todo!("DEFI-2931: sweep ETH deposits via sweepEthBatch"),
+                Asset::Eth => continue,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Working as intended for this layer, not a defect: the sweep lane for ETH is explicitly deferred by this PR. From the PR description: "Queued ETH entries are not batched into sweeps yet: the `sweepEthBatch` lane comes in a follow-up (behind the delegation-record work per the spec's delivery order), which is also why the sweep-batching `todo!` became an explicit skip."

The `Asset::Eth => continue` arm in `requests_batch` is that deliberate skip, and it is documented on the method itself ("ETH entries stay queued but are never batched yet: they wait for the `sweepEthBatch` lane"). The previous state of that code path was a `todo!()`, which would have trapped the sweep tick outright once ETH could be queued; the skip keeps ERC-20 sweeps working while the ETH entries wait in the queue.

No funds can be stranded by this in production: the scan lane this PR adds only observes balances at addresses the minter itself controls, and the entries stay queued (they are not dropped) until the follow-up lane picks them up. Leaving the thread open so the follow-up is tracked.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -123,59 +127,109 @@ async fn scan_balances<R: Runtime>(
...
-                    }
-                }
-                Err(e) => {
-                    decode_errors += 1;
-                    log!(INFO, "balance scan decode error: {e:?}");
-                }
-            },
+            for (target, balance) in chunk.iter().zip(balances) {
+                outcomes.push(scan_outcome(target, balance, latest_block));
+            }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Correct observation, but it is pre-existing behaviour that this PR does not change, and it is a different granularity from the one the PR description claims.

Chunk-level accumulation is exactly what is on master today — `git show origin/master:rs/ethereum/cketh/minter/src/balance_scan/mod.rs` has the same `for chunk in due.chunks(MAX_CALLS_PER_BATCH)` loop pushing into a single `outcomes` vec, with the comment "a chunk yields its outcomes together once it succeeds". This PR only factored the body of that loop into `chunk_balances` (so the ETH lane can reuse it); the accumulation across chunks is untouched.

The durability guarantee the description states is per *batch* (ERC-20 vs ETH), which is what the author asked for in the review thread above: `scan()` now does `scan_balances -> apply_scan_outcomes -> scan_eth_balances -> apply_scan_outcomes`, so ERC-20 findings are event-sourced before the ETH `eth_call` is awaited. Applying per chunk as well would be a further improvement, but it changes a pre-existing code path (and the signatures/unit tests that assert on the returned `Vec<ScanOutcome>`), so it is out of scope here — leaving this open so @gregorydemay can decide whether it becomes a follow-up.

rs/ethereum/cketh/test_utils/src/ckerc20.rs
```diff
@@ -198,19 +201,61 @@ impl CkErc20Setup {
...
+            JsonRpcRequestMatcher::new(JsonRpcProvider::Provider1, JsonRpcMethod::EthCall)
+                .with_request_params(
+                    kind.map(|kind| EthCallInputStartsWith::batcher(kind).into_filter()),
+                )
+                .find_rpc_call(&self.env)
+                .expect("BUG: no balance-scan eth_call pending");
+        let (kind, holders) = parse_scan_batch(&eth_call_input_bytes(&request));
+        let response: Vec<u128> = holders
+            .iter()
+            .map(|holder| scanned_balance(balances, holder))
```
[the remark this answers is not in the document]
gregorydemay: 🤖 The limitation is real, but no test in the suite can express the case today either, and the address-keyed API is what the review asked for.

* The API it replaces was positional (`&[2_000_000_u128]`, one entry per scanned `(address, token)` pair in scan order). The author asked for it to go precisely because it was brittle: "this should be specific for which address there is a balance. The address must come from the response from the `deposit_eth` endpoint". Two tokens at one address were distinguishable only by their position in the batch, which is the coupling to scan order the rewrite removes.
* Every current call site answers a single supported token per address (`tests/ckerc20.rs:221,289,328,349,670,710`), so nothing regresses in the existing tests: a same-address multi-token scan simply was not covered before and still is not.
* The upgrade path is open: the ERC-20 batch calldata carries the token in its first word per entry, and `parse_scan_batch` already walks the entries via `kind.words_per_entry()`/`holder_word_index()`, so keying ERC-20 balances by `(contract, holder)` is a local change to `parse_scan_batch` + `scanned_balance` whenever a test needs it. It costs the ETH and ERC-20 helpers a divergent tuple shape, which is why it is not done pre-emptively.

Leaving this unresolved for @gregorydemay: the key shape of a helper he specified in review is his call, and the change is worth making together with the first test that needs one token funded and another not at the same address.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -123,59 +127,109 @@ async fn scan_balances<R: Runtime>(
...
+
+async fn scan_eth_balances<R: Runtime>(
+    due: &[ScanTarget<EthAsset>],
+    latest_block: BlockNumber,
+    client: &EvmRpcClient<R, CandidResponseConverter, DoubleCycles>,
+) -> Vec<ScanOutcome> {
+    let mut outcomes = Vec::new();
+    let mut errors = ScanErrors::default();
+
+    for chunk in due.chunks(MAX_CALLS_PER_BATCH) {
```
mbjorkqvist, line 151: 🤖  `scan_eth_balances` chunks on `MAX_CALLS_PER_BATCH`, but that constant is derived from the ERC-20 encoding: `by_initcode_size` uses `BATCHER_INITCODE.len()` and `2 * WORD` per call, and its doc comment reasons entirely about `balanceOf` sub-calls and their gas.

It is safe for ETH today, but only incidentally. The constant is `min(by_initcode_size, MAX_CODE_SIZE / WORD)`, so it can never exceed 768 — which is exactly the ETH batch's binding limit, since the ETH program returns one word per holder and `768 * 32 = 24_576` is precisely `MAX_CODE_SIZE`. Equal, with zero margin. (The ETH initcode side is far looser, ~1532.) Nothing records that reasoning and nothing pins it.

Two suggestions:

- The ERC-20 path has `full_batch_fits_the_initcode_limit_and_one_more_call_does_not`; there is no ETH counterpart. A `MAX_CALLS_PER_BATCH`-sized `encode_eth_balance_batch` asserting the returned blob stays within `MAX_CODE_SIZE` and the initcode within `MAX_INITCODE_SIZE` would pin it.
- Worth a line in the `MAX_CALLS_PER_BATCH` doc noting it now caps both batchers, and why the ERC-20-derived value is ETH-safe.

Related: the largest ETH batch exercised against anvil is 32 holders (`should_read_many_eth_balances_in_a_single_call`), so the cap is untested end-to-end as well as at the const level.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -83,6 +83,26 @@ pub const MAX_CALLS_PER_BATCH: usize = {
...
+///
+/// The native-ETH sibling of [`BATCHER_INITCODE`], executed the same way (create-style
+/// `eth_call`, `to` omitted). It reads its inputs from the calldata appended right after this
+/// bytecode (`[n][ holder x n ]`, one 32-byte word each) and, for each holder, reads its ETH
+/// balance with the `BALANCE` opcode — no sub-calls, so unlike the ERC-20 batcher nothing here
+/// can fail and the program has no revert path. On success it returns the balances as a flat
+/// `n x 32`-byte array (no ABI array header), decoded positionally by [`decode_balance_batch`].
+///
+/// The program is fixed regardless of `n` (only the appended args grow). It was assembled from
+/// the opcode listing in `eth_initcode_matches_readable_assembly` and validated against a live
```
mbjorkqvist, line 102: 🤖  The in-repo design doc still describes the ETH balance read as sharing the ERC-20 call. `rs/ethereum/cketh/docs/deposit_from_cex.md:645`:

> For native ETH (Phase 2), the batcher reads the address' ETH balance **in the same latest-block call**

This PR implements it as a separate program and a separate `eth_call` — pinned to the same block, but not the same call. So the doc now contradicts the code.

The PR description already notes the spec in #11491 needs an amendment for the separate-program choice; this in-repo doc looks like it needs the same one.

(Anchoring here because `deposit_from_cex.md` is not in this PR's diff, and this doc comment is the nearest thing to it — the ERC-20 `BATCHER_INITCODE` doc points at that file, while this one points at the anvil test.)

mbjorkqvist (approved):

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 Both suggestions done in 3ab057b6b3.

* Added `full_batch_of_eth_balance_reads_fits_both_node_limits` next to the ERC-20 one: it encodes a `MAX_CALLS_PER_BATCH`-sized `encode_eth_balance_batch` and asserts the initcode stays within `MAX_INITCODE_SIZE` and the returned blob (one word per holder) within `MAX_CODE_SIZE`. It is not a literal mirror of the ERC-20 test — there is no "and one more does not fit" half, because the ETH side has real slack at 764 (its own binding limit is 768), so asserting a failure at 765 would assert something false.
* Extended the `MAX_CALLS_PER_BATCH` doc: its summary line now says it caps both batchers, and a new paragraph records why the ERC-20-derived value is ETH-safe.

On "only incidentally": one half of it is structural. The cap is `min(by_initcode_size, MAX_CODE_SIZE / WORD)`, and the `MAX_CODE_SIZE / WORD` term is shared, because both programs return exactly one word per entry — so the cap can never exceed the ETH return-blob limit no matter how the ERC-20 initcode term moves. The half that was genuinely unrecorded is the ETH initcode side (one word per entry after a 78-byte program => 1532 reads, so it never binds); that is what the new doc paragraph and the test now pin.

Not done: raising the anvil test above 32 holders. A 764-holder live batch would mean funding 764 anvil accounts per run, and what it would exercise are the node-side EIP-170/EIP-3860 bounds that the const-level test now pins directly; the anvil test`s job is that the program computes the right balances, which 32 holders already covers.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: 🤖 Correct, the doc does now contradict the code — but the amendment belongs in #11491, not here, and that is exactly what the PR description defers:

> The spec`s phase-2 wording (zero address in the token slot) needs a small amendment in #11491 to record the separate-program choice.

`rs/ethereum/cketh/docs/deposit_from_cex.md` *is* that spec — the in-repo file and the spec PR are the same document. #11491 is the open PR that edits it (it adds the "Changing the sweeper delegate" section to it), so touching line 645 from this layer would put a second, conflicting edit on a file another open PR already rewrites. The sentence you quote is pre-existing text, unchanged since before this stack: `git show origin/master:rs/ethereum/cketh/docs/deposit_from_cex.md` has it at the same line.

Leaving this thread open so it is not lost while #11491 carries the wording change.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -115,6 +141,18 @@ pub fn encode_balance_batch(calls: &[BalanceOfCall]) -> Vec<u8> {
     out
 }
 
+/// Build the create-call `input` for a batch of ETH balance reads:
+/// `ETH_BATCHER_INITCODE ++ [n] ++ [ holder x n ]`, every value a 32-byte word.
+pub fn encode_eth_balance_batch(holders: &[DepositAddress]) -> Vec<u8> {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Valid, fixed in e9d68a7936.

`decode_balance_batch` is now documented as decoding the return blob of *either* batcher, and the no-partial-result guarantee is stated per program rather than as an ERC-20 fact: `BATCHER_INITCODE` reverts the whole call if any `balanceOf` sub-call fails, and `ETH_BATCHER_INITCODE` has no failure path at all because the `BALANCE` opcode makes no sub-calls. The shared conclusion — a failed batch surfaces as an `eth_call` error upstream, never as a `0` here — is kept, since it now follows from both halves.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: I added a comment [here](https://github.com/dfinity/ic/pull/11491/changes#r3986722991) to not forget it.

PR #11509, opened 2026-09-09
fix(node): correct the type4 multi-GuestOS paths

q-uint:
Type4 nodes run up to 60 GuestOS per host, several components still assume one. Each commit is independent.

- **VM counts**: the two HostOS monitoring scripts keep their own copy of the per-reward-type count and were missed by #11204 (type4.2 8→15) and #11207 (dropped type4.0), so a type4.2 node monitored 8 of its 15 VMs.
- **Slot indexing**: both walked slots `0..COUNT-1`, but `start-guestos.sh` boots `guestos@{1..COUNT}`. They polled a `guestos0` that never exists — a permanent `hostos_guestos_state=1`, and never polled `guestos<COUNT>`.
- **Serial log names**: `serial_log_path()` appended the slot *after* `.log`, yielding `guestos-serial.log7` while the exporter tails `guestos-serial7.log`. No type4 node forwarded serial logs at all.
- **Lost GuestOS volume**: `install-hostos.sh` took the multi-VM branch for any `^type4(\.[0-9]+)?$`, but only `.1`–`.4` have a matching case, so `type4` and `type4.5` had `hostlvm/guestos` removed and nothing created.
- **Console and logrotate**: `limited-console` and `verbose-logging.sh` hardcoded the single-VM path, and logrotate covered only the unsuffixed name, so a 60-VM node grew 60 unrotated logs.

The count is then moved into a shared `guestos-vm-count.sh` sourced by all four scripts, since it had already drifted twice. The new `sh_test` derives its expectation by running `start-guestos.sh` rather than restating the counts, so a one-sided edit fails it.

**The Rust change is unbuilt**: `guest_vm_runner` needs x86_64 Linux (`libdevmapper`, `@dosfstools//:mkfs.fat`) and I have no such host, so `cargo fmt` and `rust-lint.sh` have not run against it. Its path logic was verified standalone across all 61 slots. The shell changes pass `//ic-os/components:test_monitor_guestos`.


rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -539,6 +541,44 @@ mod tests {
...
+    // "<name><slot>.log" under /var/log/libvirt/qemu.
+    #[test]
+    fn test_serial_log_path() {
+        for (guest_vm_type, name) in [
+            (GuestVMType::Default, "guestos-serial"),
+            (GuestVMType::Upgrade, "upgrade-guestos-serial"),
+        ] {
+            assert_eq!(
+                serial_log_path(guest_vm_type, VmSlot::Plain),
+                PathBuf::from(format!("/var/log/libvirt/qemu/{name}.log"))
```
frankdavid: These tests repeat much of the logic above (and therefore may not be too powerful). Could you make them dumber by just comparing plain strings without having logic on the right handside of the assert_eq?

e.g.
```
assert_eq!(
     serial_log_path(GuestVMType::Default, VmSlot::Plain), 
     Path::new("/var/log/libvirt/qemu/guestos-serial")
)
```

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
```diff
@@ -9,9 +9,8 @@ source /opt/ic/bin/config.sh
 node_reward_type=$(get_config_value '.icos_settings.node_reward_type')
 
 case "${node_reward_type}" in
-    type4.0) COUNT=32 ;;
     type4.1) COUNT=60 ;;
-    type4.2) COUNT=8 ;;
+    type4.2) COUNT=15 ;;
```
frankdavid: Could you add a note with all the things that this has to be consistent with? (same below)

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
```diff
@@ -16,25 +16,34 @@ case "${node_reward_type}" in
...
+    slots=(0)
+else
+    slots=($(seq 1 "$COUNT"))
+fi
+
 # Forward all the GuestOS logs
-for i in $(seq 0 "$((COUNT - 1))"); do
-    # A single GuestOS keeps the guestos-serial.log name
-    if [ "$COUNT" -eq 1 ]; then
+for slot in "${slots[@]}"; do
```
frankdavid: Shall we have a second loop for guestos-serial, upgrade-guestos-serial instead of duplicating the logic?

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
```diff
@@ -16,25 +16,34 @@ case "${node_reward_type}" in
...
+    # Slot 0, the single GuestOS, keeps the guestos-serial.log name
+    if [ "$slot" -eq 0 ]; then
```
frankdavid: @Bownairo Do we need this distinction? We already name the "default" (slot 0) guestos systemd unit as guestos@0.service (no distinction), do we gain anything by special casing log name suffixes? (This logic here is related to the logic in `VmSlot` [here](https://sourcegraph.com/r/github.com/dfinity/ic/-/blob/rs/ic_os/config/types/src/lib.rs?L370))

ic-os/components/monitoring/hostos/test-monitor-guestos.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
```
frankdavid: Instead of this test, could we consolidate the monitor and the start scripts to avoid drift?  @Bownairo

ic-os/components/setupos/install-hostos.sh
```diff
@@ -99,8 +99,11 @@ function resize_partition() {
 
     local node_reward_type=$(get_config_value '.icos_settings.node_reward_type')
 
-    # Configure multiple GuestOS if type4.X
-    if [[ $node_reward_type =~ ^type4(\.[0-9]+)?$ ]]; then
+    # Configure multiple GuestOS for the type4 reward types that run more than
```
frankdavid: I don't understand this comment :(

rs/ic_os/os_tools/guest_vm_runner/src/guest_vm_config.rs
```diff
@@ -263,16 +264,16 @@ pub fn vm_domain_uuid(guest_vm_type: GuestVMType, slot: VmSlot) -> String {
...
-        GuestVMType::Upgrade => PathBuf::from(format!(
-            "{UPGRADE_SERIAL_LOG_PATH}{suffix}",
-            suffix = slot.to_suffix()
-        )),
-    }
+    let name = match guest_vm_type {
+        GuestVMType::Default => DEFAULT_SERIAL_LOG_NAME,
+        GuestVMType::Upgrade => UPGRADE_SERIAL_LOG_NAME,
+    };
+    // The slot goes before the extension, so that every VM's log is still a
```
frankdavid: nit: Not useful comment

ic-os/components/misc/guestos-vm-count.sh
frankdavid, line 1: Consider moving this file under `ic-os/components/hostos/guestos`

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
frankdavid: Changing the naming may break some multi-guestos nodes.

q-uint: It's the existing `VmSlot` logic. `start-guestos.sh` boots `guestos@0` (single) or `guestos@{1..N}`, and `VmSlot::new(0) -> suffix ""`, so `guestos0` never exists afaik.

The log filenames do change, but as a bug fix: master appends the suffix after .log, producing guestos-serial.log1. Nothing reads those?

ic-os/components/setupos/install-hostos.sh
q-uint: The old `^type4(\.[0-9]+)?$` matched `type4/type4.0/type4.5`, which had no case branch, such a node would `lvremove` the guestos LV and create nothing. I assumed this to be unintended.

ic-os/components/monitoring/hostos/export-guestos-serial-logs.sh
Bownairo: We also had the special casing for services, but got rid of it (`guestos@0.service`, used to just be `guestos.service`), but this was just to make systemd launching easier.

I think we should probably keep the special casing, but fix the suffix, for now, just because there still are a lot of assumptions on the old naming.

ic-os/components/setupos/install-hostos.sh
Bownairo: `type4`, `type4.0`, and `type4.5` (not yet) are not in use. We can handle the logic better here, but maybe these invalid "types" should be picked up elsewhere, first.

ic-os/components/monitoring/hostos/test-monitor-guestos.sh
Bownairo: I agree, there's a lot going on here just to test. The count logic is already consolidated with this PR, can we start with that?

ic-os/components/setupos/install-hostos.sh
q-uint: I stumbled upon these issues when exploring support for new (arbitrary) hardware, i.e. `type4`. So it'd be nice that we take it into account everywhere.

ic-os/components/misc/guestos-vm-count.sh
q-uint: `guestos-vm-count.sh` is also sourced by SetupOS's `install-hostos.sh`, so filing it under `hostos/guestos/` is a slightly odd home?

PR #11522, opened 2026-09-09
feat(consensus): report the consensus status as a metric

mraszyk:
Whether a subnet has halted at a CUP height is only observable in the replica's
journal today. This PR reports it as a metric, too:

```
consensus_status{status="running"|"halting"|"halted_at_cup_height"|"unknown"}
```

an `IntGaugeVec` holding 1 for the status the finalizer last computed and 0 for
the other three.

The status is the one `status::get_status` computes: the halts that end a subnet
on a CUP -- a pending upgrade, a scheduled subnet split, and the subnet record's
`halt_at_cup_height`. The record's other halt, `is_halted`, is acted upon above
the finalizer, so the finalizer never sees it and this metric does not report it.

The finalizer computes the status for the height whose batch was delivered last.
That is the height to ask about: the halts above take effect from a summary
height on, and the batch at that height is delivered whatever the status is, so
the first height at which a subnet halts is also the last one whose batch it
delivers.

`unknown` is the status the finalizer did not compute -- an unreadable registry,
no summary block for the height, or a batch delivery that failed, which returns
an error rather than the height it got to -- rather than leaving the metric
asserting the status of an earlier height as though it still held.

The delivery path's log lines now name the status that held; one of them said
"replica is halted" for a halting subnet as well as a halted one.

This is operator-facing observability: the metric says why a subnet stopped
producing blocks where previously only the journal did. It is not a signal to
gate an operation on. It carries no height, and it does not say which of the
three halts fired, so a programmatic check that a subnet has halted at a given
checkpoint still has to look at the CUP itself -- as
`wait_until_halting_cup_reached` in `rs/tests/consensus/subnet_splitting_test.rs`
does, matching a CUP's summary registry version against the subnet record at
that version.

### DIFF
--- a/rs/consensus/src/consensus/finalizer.rs
+++ b/rs/consensus/src/consensus/finalizer.rs
@@ -19,6 +19,7 @@
 use crate::consensus::{
     batch_delivery::deliver_batches_for_finalizer,
     metrics::{BatchStats, BlockStats, FinalizerMetrics},
+    status::{self, Status},
 };
 use ic_consensus_utils::{
     crypto::ConsensusCrypto, membership::Membership, pool_reader::PoolReader,
@@ -95,7 +96,7 @@ impl Finalizer {
         }
 
         // Try to deliver finalized batches to messaging
-        let _ = deliver_batches_for_finalizer(
+        let last_delivered_batch_height = deliver_batches_for_finalizer(
             &*self.message_routing,
             &self.membership,
             pool,
@@ -106,14 +107,47 @@ impl Finalizer {
             |result, block_stats, batch_stats| {
                 self.process_batch_delivery_result(result, block_stats, batch_stats)
             },
-            |status| self.metrics.observe_status(status),
         );
+        self.metrics
+            .observe_status(self.get_status(pool, last_delivered_batch_height));
 
         // Try to finalize rounds from finalized_height + 1 up to (and including)
         // notarized_height
         (finalized_height.increment().get()..=notarized_height.get())
             .filter_map(|h| self.finalize_height(pool, Height::from(h)))
             .collect()
+    }
+
+    /// The consensus status as of the last batch delivery, or [`None`] if it
+    /// could not be computed.
+    ///
+    /// It is the status of the last height whose batch was delivered, which is
+    /// the height to ask about: the halts this status covers take effect from a
+    /// summary height on, and the batch at that height is delivered whatever
+    /// the status is. The first height at which a subnet is halting is
+    /// therefore also the last one whose batch it delivers, so asking about
+    /// that height reports the halt from the round it begins.
+    ///
+    /// A delivery that failed leaves no height to ask about and no way to tell
+    /// whether the subnet is halting, so it is reported as unknown rather than
+    /// as though everything had gone well.
+    fn get_status(
+        &self,
+        pool: &PoolReader<'_>,
+        last_delivered_batch_height: Result<Height, MessageRoutingError>,
+    ) -> Option<Status> {
+        let height = last_delivered_batch_height.ok()?;
+        let summary_block = pool.dkg_summary_block_for_finalized_height(height)?;
+
+        status::get_status(
+            height,
+            &summary_block,
+            self.registry_client.as_ref(),
+            self.replica_config.subnet_id,
+            pool,
+            self.replica_config.replica_version(),
+            &self.log,
+        )
     }
 
     /// Write logs, report metrics depending on the batch deliver result.
@@ -257,16 +291,123 @@ mod tests {
     use ic_consensus_mocks::{Dependencies, DependenciesBuilder};
     use ic_logger::replica_logger::no_op_logger;
     use ic_metrics::MetricsRegistry;
+    use ic_test_artifact_pool::consensus_pool::Round;
     use ic_test_utilities::{
         ingress_selector::FakeIngressSelector, message_routing::FakeMessageRouting,
     };
     use ic_test_utilities_registry::SubnetRecordBuilder;
-    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id};
+    use ic_test_utilities_types::ids::{node_test_id, subnet_test_id, test_replica_version};
     use ic_types::{
         RegistryVersion,
         consensus::{HasHeight, HashedBlock},
     };
+    use rstest::rstest;
     use std::sync::Arc;
+
+    /// The `consensus_status` labels whose gauge reads 1. Exactly one of them
+    /// does once the finalizer has computed a status.
+    fn observed_statuses(metrics_registry: &MetricsRegistry) -> Vec<String> {
+        metrics_registry
+            .prometheus_registry()
+            .gather()
+            .into_iter()
+            .filter(|family| family.name() == "consensus_status")
+            .flat_map(|family| family.get_metric().to_vec())
+            .filter(|metric| metric.get_gauge().value() == 1.0)
+            .map(|metric| metric.get_label()[0].value().to_string())
+            .collect()
+    }
+
+    /// The finalizer reports the consensus status of the height whose batch was
+    /// delivered last. For a subnet halting at a CUP height that is the CUP
+    /// height itself: its batch is delivered whatever the status is, so the
+    /// first height at which the subnet halts is also the last one it delivers.
+    #[rstest]
+    #[case::running(
+        /* halt_at_cup_height= */ false,
+        /* certified_up_to_cup_height= */ false,
+        "running"
+    )]
+    #[case::halting(
+        /* halt_at_cup_height= */ true,
+        /* certified_up_to_cup_height= */ false,
+        "halting"
+    )]
+    #[case::halted(
+        /* halt_at_cup_height= */ true,
+        /* certified_up_to_cup_height= */ true,
+        "halted_at_cup_height"
+    )]
+    fn test_finalizer_observes_status(
+        #[case] halt_at_cup_height: bool,
+        #[case] certified_up_to_cup_height: bool,
+        #[case] expected_status: &str,
+    ) {
+        const INTERVAL_LENGTH: u64 = 3;
+        const HALT_REGISTRY_VERSION: u64 = 10;
+        // A summary block, and hence a CUP, every `INTERVAL_LENGTH + 1` heights.
+        const CUP_HEIGHT: Height = Height::new(2 * (INTERVAL_LENGTH + 1));
+
+        ic_test_utilities::artifact_pool_config::with_test_pool_config(|pool_config| {
+            // The same replica version throughout, so that the subnet record is
+            // the only thing that can halt the subnet.
+            let record = || {
+                SubnetRecordBuilder::from(&[node_test_id(0)])
+                    .with_dkg_interval_length(INTERVAL_LENGTH)
+                    .with_replica_version(test_replica_version().as_ref())
+            };
+            let Dependencies {
+                mut pool,
+                replica_config,
+                membership,
+                registry,
+                crypto,
+                ..
+            } = DependenciesBuilder::single_subnet(
+                pool_config,
+                subnet_test_id(0),
+                vec![
+                    (1, record().build()),
+                    (
+                        HALT_REGISTRY_VERSION,
+                        record().with_halt_at_cup_height(halt_at_cup_height).build(),
+                    ),
+                ],
+            )
+            .build();
+
+            pool.advance_round_normal_operation_n(CUP_HEIGHT.get());
+            let batch_height = if certified_up_to_cup_height {
+                // One round more, so that the certified height of the finalized
+                // tip reaches the CUP height. That is what tells a subnet that
+                // is halted from one that is only halting towards that height.
+                Round::new(&mut pool)
+                    .with_certified_height(CUP_HEIGHT)
+                    .advance();
+                CUP_HEIGHT.increment()
+            } else {
+                CUP_HEIGHT
+            };
+
+            let metrics_registry = MetricsRegistry::new();
+            let message_routing = Arc::new(FakeMessageRouting::new());
+            *message_routing.next_batch_height.write().unwrap() = batch_height;
+            let finalizer = Finalizer::new(
+                replica_config,
+                registry,
+                membership,
+                crypto,
+                message_routing,
+                Arc::new(FakeIngressSelector::new()),
+                no_op_logger(),
+                metrics_registry.clone(),
+            );
+
+            let _ = finalizer.on_state_change(&PoolReader::new(&pool));
+
+            assert_eq!(observed_statuses(&metrics_registry), vec![expected_status]);
+        })
+    }
 
     /// Given a single block, just finalize it
     #[test]
--- a/rs/consensus/src/consensus/finalizer.rs
+++ b/rs/consensus/src/consensus/finalizer.rs
@@ -106,6 +106,7 @@ impl Finalizer {
             |result, block_stats, batch_stats| {
                 self.process_batch_delivery_result(result, block_stats, batch_stats)
             },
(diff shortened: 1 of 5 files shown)


rs/consensus/src/consensus/metrics.rs
```diff
@@ -196,6 +218,22 @@ pub(crate) struct FinalizerMetrics {
 
 impl FinalizerMetrics {
     pub fn new(metrics_registry: MetricsRegistry) -> Self {
+        let consensus_status = metrics_registry.int_gauge_vec(
+            "consensus_status",
+            "Whether consensus is running, halting (producing empty blocks but delivering \
+             no batches), halted (producing no blocks either) or unknown (the status could \
+             not be computed), as of the last time batch delivery looked. 1 for the status \
+             that held then, 0 for the other three.",
```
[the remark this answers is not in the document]
mraszyk: Addressed: the PR description now documents all four label values. `unknown` is kept, described there as the status the delivery path failed to compute rather than a status consensus can be in, and the help text says that all four read 0 until batch delivery computes a status for the first time.

pierugo-dfinity (commented):
We don't stricly need to update the metric for every delivered height, right? Wouldn't it be sufficient to update it only once per call to `deliver_batches` according to `last_delivered_batch_height`? More concretely, instead of introducing a new closure and ensuring that all exit points of the function call it (which will almost certainly break as soon as we add a new `break`/`return` somewhere), leave `deliver_batches` unchanged and let the finalizer call `get_status` on the `last_delivered_batch_height` and report this. It would keep this functionality constrained to a single place. In case delivery fails for the halting CUP height (and we get a `MessageRoutingError`) we could export `unknown` instead. I don't think it's bad such that it would force a manual intervention instead of assuming everything went well.

I'm in favour of this change just to have better observability, but I have a question regarding its motivation i.e. for subnet merging. I assume the condition we're looking for is: "the subnet has created the manifest at the CUP height and has reached consensus on it". This is equivalent to "the subnet has created a CUP at that height". So maybe the programmatical approach should just wait for the subnet to create a CUP whose registry version contains the subnet record change that tells the subnet to halt?
With the current metric-based approach,
- Even though the metric says "halted_at_cup_height", there's no guarantee that the manifest was computed yet, only that the CUP height was certified.
- Most importantly, even if the manifest was computed, there's still no guarantee that the subnet reached consensus on it, i.e. there could be a divergence.

mraszyk:
> We don't strictly need to update the metric for every delivered height, right? [...] leave `deliver_batches` unchanged and let the finalizer call `get_status` on the `last_delivered_batch_height` and report this.

Done in 21f2c1e — `deliver_batches` is back to what it was, minus the two log lines, and the finalizer computes the status once after delivery returns.

You're right that nothing is lost, and I think the reason is worth writing down: all three halts become true exactly at the summary height. `should_halt_due_to_upgrading` and `should_halt_by_subnet_record` key off `pool.registry_version(height)`, which is the summary block's registry version, and `should_halt_due_to_subnet_splitting` is literally `height >= summary_block.height`. Since the batch at the summary height is delivered whatever the status is, the first height at which a subnet halts is also the last one whose batch it delivers — so `get_status(last_delivered_batch_height)` sees the halt from the round it begins, and the `Halting → Halted` transition still happens because `get_status` re-reads the finalized tip's certified height on every call.

Two smaller consequences:

- `unknown` now covers an unreadable registry and a delivery that failed, as you suggested. The one behaviour difference from the closure: if the *CUP* batch itself can't be delivered (random tape missing, state sync), `last_delivered_batch_height` is still below the summary height and the metric reads `running` rather than `unknown`. I left it that way — it's honest about the last height that was actually delivered — but probing `last_delivered_batch_height.increment()` instead would restore the old behaviour if you'd rather have it.
- I dropped the "all four series start at 0" initialization. It was there because the delivery loop often doesn't run a body iteration at all; the finalizer computes a status on every `on_state_change`, so it bought nothing and the "sum of 0 means not looked yet" contract was just confusing.

> I have a question regarding its motivation i.e. for subnet merging.

You're right, and I've dropped subnet merging from the description. Both objections hold:

- `Status::Halted` only requires the finalized tip's `certified_height` to have reached the summary height. Certification hashes the certification tree (`certified_state_hash = crypto_hash_of_tree(&hash_tree)`), not the manifest, so it says nothing about the manifest being computed — that's `state_manager_last_computed_manifest_height`.
- And the manifest existing isn't agreement on it. The manifest root hash is what goes into `CatchUpContent.state_hash`, so the CUP is the agreement; a divergence at the checkpoint would leave `consensus_status` reading `halted_at_cup_height` regardless.

Your suggested check is also already the established one: `wait_until_halting_cup_reached` in `rs/tests/consensus/subnet_splitting_test.rs` fetches each node's CUP from `/_/catch_up_package`, reads its summary `dkg.registry_version`, and asserts `raw_subnet_record_at_version(...).halt_at_cup_height` — a CUP whose registry version contains the halt — then cross-checks `certification_height` and that the heights agree across nodes. Merging should do the same rather than gate on this metric.

So I'd keep this PR as operator-facing observability only: it says *why* a subnet stopped producing blocks, where before only the journal did. It carries no height and doesn't distinguish an upgrade from a `halt_at_cup_height` record change, so it isn't something to gate an operation on, and the description now says so.

mraszyk:
Two corrections to my comment above.

**Drop the `last_delivered_batch_height.increment()` offer.** It isn't the straight swap I made it sound like. On a healthy subnet the delivery loop runs up to `target_height == finalized_height`, so the last delivered height *is* the finalized height, and `dkg_summary_block_for_finalized_height` returns `None` for the next one — the metric would read `unknown` on every round. Making it work would mean taking the summary block from `get_highest_finalized_summary_block()` instead, and the only case it buys anything is a replica that cannot deliver the CUP batch at all, where it would report a halt at a height the replica hasn't reached. That seems less honest than reporting the status of the last height it actually delivered, so I'd rather keep `last_delivered_batch_height` as you suggested.

**`unknown` has three sources, not the two I listed.** An unreadable registry, a height with no summary block, and a delivery that failed — which returns an error rather than the height it got to, rather than "leaves no height to ask about" as I put it. The description and the `Finalizer::get_status` doc comment say so now (16c06ea).

rs/consensus/src/consensus/status.rs
```diff
@@ -12,17 +12,32 @@ use ic_types::{
...
     Running,
-    /// The Consensus is halting, meaning we will produce *empty* blocks but no batches will be
-    /// delivered.
+    /// The Consensus is halting towards a CUP height, meaning we will produce *empty* blocks and
+    /// deliver no batch but the one at that CUP height.
     Halting,
-    /// The Consensus is halted, meaning that no blocks are created and no batches are delivered.
+    /// The Consensus is halted at a CUP height, meaning that no blocks are created and no
+    /// batches are delivered. The batch at that CUP height has been delivered by then: the
+    /// status holds only once certification has passed the height.
```
pierugo-dfinity, line 27:
```suggestion
    /// The Consensus is halted at a CUP height, meaning that no blocks are created and no
    /// batches are delivered.
```

> The batch at that CUP height has been delivered by then

Literally mentioned 3 lines above

> the status holds only once certification has passed the height

Implementation detail that is not completely correct (only once a block, whose validation context's certified height has reached this height, is finalized, which is a bit more than just the certification) and that could change in the future (there's a good reason why we could move to `Halted` directly as soon as the height is finalized)

rs/consensus/src/consensus/finalizer.rs
```diff
@@ -95,7 +96,7 @@ impl Finalizer {
         }
 
         // Try to deliver finalized batches to messaging
-        let _ = deliver_batches_for_finalizer(
+        let last_delivered_batch_height = deliver_batches_for_finalizer(
```
pierugo-dfinity, line 99:
```suggestion
        let delivery_result = deliver_batches_for_finalizer(
```
This is still a `Result` at this point. But this might be obsolete if you apply my other comment

rs/consensus/src/consensus/finalizer.rs
```diff
@@ -115,6 +118,40 @@ impl Finalizer {
...
+    /// summary height on, and the batch at that height is delivered whatever
+    /// the status is. The first height at which a subnet is halting is
+    /// therefore also the last one whose batch it delivers, so asking about
+    /// that height reports the halt from the round it begins.
+    ///
+    /// A delivery that failed returns an error rather than the height it got
+    /// to, leaving nothing to ask about, so the status is reported as unknown
+    /// rather than as though everything had gone well. So is a height with no
+    /// summary block, and one whose status [`status::get_status`] could not
+    /// compute.
```
pierugo-dfinity, line 135: Just *slightly slightly* verbose.
```suggestion
    /// Consensus' status as of the last delivered height, or [`None`] if it
    /// could not be computed.
```

rs/consensus/src/consensus/metrics.rs
```diff
@@ -25,6 +26,28 @@ use std::sync::RwLock;
...
 
+/// The label of `consensus_status`, whose values name the statuses of
+/// [`Status`], plus `unknown` for a status that could not be computed at all.
+///
+/// [`Status::Halted`] is reported as `halted_at_cup_height` rather than
+/// `halted`, after the halt it stands for: the one that ends a subnet on a CUP.
+const STATUS_LABEL: &str = "status";
+const STATUS_RUNNING: &str = "running";
+const STATUS_HALTING: &str = "halting";
+const STATUS_HALTED_AT_CUP_HEIGHT: &str = "halted_at_cup_height";
```
pierugo-dfinity, line 37: I think I would actually prefer it to be "halted". I understand it can be confused with the `SubnetRecord.is_halted`, but since there is "status" in the name of the metric, it's clear that it refers to `Status::Halted`. Otherwise, I get confused with `SubnetRecord.halt_at_cup_height`, which is a possible reason for a `Status::Halted`, but not the only one.

rs/consensus/src/consensus/metrics.rs
```diff
@@ -25,6 +26,28 @@ use std::sync::RwLock;
...
+const STATUS_UNKNOWN: &str = "unknown";
+
+/// The label value `consensus_status` reports each status under. [`None`] is
+/// the status the finalizer failed to compute, rather than a status of its
+/// own, so that a subnet whose registry cannot be read is not mistaken for one
+/// that is running.
```
pierugo-dfinity, line 43:
```suggestion
/// The label value `consensus_status` reports each status under. [`None`] is
/// the status the finalizer failed to compute.
```

rs/consensus/src/consensus/metrics.rs
```diff
@@ -206,6 +230,19 @@ impl FinalizerMetrics {
...
+                "consensus_status",
+                "Whether consensus is running, halting towards a CUP height (producing \
+                 empty blocks and delivering no batch but the one at that height), halted \
+                 at a CUP height (producing no blocks either) or unknown (the status could \
+                 not be computed), as of the height whose batch was delivered last. 1 for \
+                 the status that held then, 0 for the other three. Reported from the first \
+                 time the finalizer computes a status, and absent before that -- including \
+                 on a replica that started while its subnet record's `is_halted` flag was \
+                 set, as consensus acts on that flag before the finalizer runs. That halt \
+                 has no status of its own here.",
```
pierugo-dfinity, line 243:
```suggestion
                "Whether consensus is running, halting or halted as of the last delivered height. 1 for the corresponding status, 0 for the others.",
```

rs/consensus/src/consensus/metrics.rs
```diff
@@ -328,6 +365,21 @@ impl FinalizerMetrics {
         }
     }
 
+    /// Records `status` as the status consensus is in, and the other three as
+    /// ones it is not. [`None`] is recorded as `unknown`, the status the
+    /// finalizer failed to compute, and [`Status::Halted`] as
+    /// `halted_at_cup_height`, the only halt the finalizer sees.
```
pierugo-dfinity, line 371:
```suggestion
    /// Records `status` as the status consensus is in, and the other three as
    /// ones it is not.
```

rs/consensus/src/consensus/metrics.rs
```diff
@@ -328,6 +365,21 @@ impl FinalizerMetrics {
...
+    ///
+    /// Reported as a gauge per status rather than a single number, so that a
+    /// dashboard can select the status it asks about by name.
+    pub fn observe_status(&self, status: Option<Status>) {
+        for (label, value) in CONSENSUS_STATUSES {
+            self.consensus_status
+                .with_label_values(&[label])
+                .set((value == status) as i64);
+        }
```
pierugo-dfinity, line 380: Unfortunately this is not atomic, so there is a race condition where a scraper scrapes between two iterations of the loop and could see multiple 1s or all 0s right?
Claude suggests the following for the first race. The second is still there, but is less problematic

```rust
    for (label, value) in CONSENSUS_STATUSES {
        if value != status {
            self.consensus_status.with_label_values(&[label]).set(0);
        }
    }
    for (label, value) in CONSENSUS_STATUSES {
        if value == status {
            self.consensus_status.with_label_values(&[label]).set(1);
        }
    }
```

rs/consensus/src/consensus/finalizer.rs
```diff
@@ -256,17 +293,124 @@ mod tests {
...
+        metrics_registry
+            .prometheus_registry()
+            .gather()
+            .into_iter()
+            .filter(|family| family.name() == "consensus_status")
+            .flat_map(|family| family.get_metric().to_vec())
+            .filter(|metric| metric.get_gauge().value() == 1.0)
+            .map(|metric| metric.get_label()[0].value().to_string())
+            .collect()
+    }
```
pierugo-dfinity: I think you could use `rs/test_utilities/metrics/lib.rs:fetch_gauge_vec`

rs/consensus/src/consensus/metrics.rs
```diff
@@ -679,3 +731,52 @@ impl PurgerMetrics {
...
+
+        assert_eq!(
+            consensus_status(&metrics_registry),
+            BTreeMap::from([
+                (STATUS_RUNNING.into(), 0),
+                (STATUS_HALTING.into(), 0),
+                (STATUS_HALTED_AT_CUP_HEIGHT.into(), 1),
+                (STATUS_UNKNOWN.into(), 0),
+            ]),
+        );
```
pierugo-dfinity, line 780:
```suggestion
        metrics.observe_status(Some(Status::Halting));
        assert_eq!(
            consensus_status(&metrics_registry),
            BTreeMap::from([
                (STATUS_RUNNING.into(), 0),
                (STATUS_HALTING.into(), 1),
                (STATUS_HALTED_AT_CUP_HEIGHT.into(), 0),
                (STATUS_UNKNOWN.into(), 0),
            ]),
        );

        metrics.observe_status(Some(Status::Halted));
        assert_eq!(
            consensus_status(&metrics_registry),
            BTreeMap::from([
                (STATUS_RUNNING.into(), 0),
                (STATUS_HALTING.into(), 0),
                (STATUS_HALTED_AT_CUP_HEIGHT.into(), 1),
                (STATUS_UNKNOWN.into(), 0),
            ]),
        );
```

pierugo-dfinity (commented):
> if the CUP batch itself can't be delivered (random tape missing, state sync), `last_delivered_batch_height` is still below the summary height and the metric reads `running` rather than `unknown`.

So now I'm starting to realize that it would probably be better to have some sort of invariant to what this metric should mean. If we'd like it to say "consensus' status at its last delivered batch height", maybe it would be best to actually still ignore the result of this call and export 
```rust
self.get_status(pool, self.message_routing.expected_batch_height().decrement())
```
Otherwise with the current approach, we might have a situation like you described where a height not being delivered might either export `running` or `unknown`. Also currently if an error occurs, it masks the previously delivered heights.

PR #11532, opened 2026-09-10
feat(cketh): sweep ETH deposits via sweepEthBatch

gregorydemay:
Final step of the `deposit_eth` flow per the [spec](https://github.com/dfinity/ic/pull/11491), on top of the merged #11494, #11499 and #11508: queued ETH deposits are now swept and minted, so `deposit_eth` covers the whole pipeline from address derivation to the ckETH credit.

* A `SweepRequest` moves one asset, ETH or a single ERC-20 token. The `sweepEthBatch` entry point takes the same `SweepItem` array as the ERC-20 one without the token list, and attestations are reused unchanged because their digest binds only the account. The asset field keeps the CBOR encoding of the previous bare token address, so recorded sweep events replay unchanged. The sweep gas limit budgets an ETH address for the one balance read and `depositEth` call `sweepEth` makes rather than for the ERC-20 approve-and-transfer shape, so a ten-deposit ETH sweep declares about 860k gas against the 413k measured.
* The live anvil coverage mirrors the ERC-20 suite: twenty deposits credited through full ten-deposit `sweepEthBatch` sweeps, threshold gating around the 0.005 ETH minimum, a second sweep of an already-delegated address despite its re-sent stale authorization, and a mixed population where one account deposits both assets at its single address and each asset is swept by its own lane. Sweep gas is asserted within 10 percent of new demo-measured ETH scenarios (413,076 for a ten-deposit first sweep, about two thirds of the ERC-20 figure).
* Deferred, as decided: entry-point gating on the delegate an address actually runs (R18). Until the delegation-record and rotation work lands, a single Sepolia address still delegated to the pre-#11449 delegate reverts the whole `sweepEthBatch` it lands in, since `sweepEth` is a void call with no per-item fallback. A failed sweep drops every deposit it named from the queue, up to ten per batch, and each needs a fresh `deposit_eth` to be armed again. Mainnet is unaffected, since no production address was ever delegated with the old delegate.

### DIFF
--- a/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
+++ b/rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
@@ -205,11 +205,11 @@ impl AutomaticDeposits {
             .sweeper_transactions
             .get_processed_request(&id)
             .expect("BUG: missing sweep request");
-        let token = request.token;
+        let asset = request.asset;
         let accounts: Vec<_> = request.items.iter().map(|item| item.item.account).collect();
 
         for account in accounts {
-            let request = DepositRequest::new(account, Asset::Erc20(token));
+            let request = DepositRequest::new(account, asset);
             let entry = self
                 .sweep
                 .remove(&request)
@@ -554,23 +554,14 @@ impl AutomaticDeposits {
         })
     }
 
-    /// The queued deposits a sweep could take next, batched by token, skipping those a sweep
+    /// The queued deposits a sweep could take next, batched by asset, skipping those a sweep
     /// already holds: taking them twice would move a balance the minter has already accounted for.
-    /// ETH entries stay queued but are never batched yet: they wait for the `sweepEthBatch` lane
-    /// (DEFI-2931).
-    pub fn requests_batch(
-        &self,
-        requested_batch_size: usize,
-    ) -> BTreeMap<Address, Vec<SweepTarget>> {
+    pub fn requests_batch(&self, requested_batch_size: usize) -> BTreeMap<Asset, Vec<SweepTarget>> {
         let mut batches = BTreeMap::new();
         for (deposit_request, sweep_entry) in
             self.sweep.iter().filter(|(_, entry)| entry.is_sweepable())
         {
-            let token = match deposit_request.asset {
-                Asset::Erc20(token) => token,
-                Asset::Eth => continue,
-            };
-            let batch: &mut Vec<_> = batches.entry(token).or_default();
+            let batch: &mut Vec<_> = batches.entry(deposit_request.asset).or_default();
             if batch.len() < requested_batch_size {
                 batch.push(SweepTarget {
                     account: deposit_request.account,
@@ -581,7 +572,7 @@ impl AutomaticDeposits {
         batches
     }
 
-    /// Record that `sweep_id` took these accounts' deposits of `token`: each leaves the pool of
+    /// Record that `sweep_id` took these accounts' deposits of `asset`: each leaves the pool of
     /// sweepable entries until the sweep is done with it.
     ///
     /// # Panics
@@ -591,11 +582,11 @@ impl AutomaticDeposits {
     pub fn record_sweep_scheduled(
         &mut self,
         sweep_id: SweepId,
-        token: Address,
+        asset: Asset,
         accounts: impl IntoIterator<Item = Account>,
     ) {
         for account in accounts {
-            let request = DepositRequest::new(account, Asset::Erc20(token));
+            let request = DepositRequest::new(account, asset);
             let entry = self
                 .sweep
                 .get_mut(&request)
--- a/rs/ethereum/cketh/minter/src/state/automatic_deposits/tests.rs
+++ b/rs/ethereum/cketh/minter/src/state/automatic_deposits/tests.rs
@@ -879,6 +879,22 @@ fn should_batch_eth_entries_alongside_tokens() {
 }
 
 #[test]
+fn should_cap_each_asset_batch_at_the_requested_size() {
+    let mut entries: Vec<(Account, Asset)> =
+        (0..11).map(|index| (account(index), Asset::Eth)).collect();
+    entries.push((account(11), Asset::Erc20(usdc())));
+    let deposits = queued(&entries);
+
+    let batches = deposits.requests_batch(10);
+
+    assert_eq!(
+        accounts_in(&batches, Asset::Eth),
+        (0..10).map(account).collect::<Vec<_>>()
+    );
+    assert_eq!(accounts_in(&batches, usdc()), vec![account(11)]);
+}
+
+#[test]
 fn should_batch_queued_deposits_by_token() {
     let deposits = queued(&[
         (account(0), usdc()),
@@ -903,22 +919,24 @@ fn should_batch_queued_deposits_by_token() {
 
 #[test]
 fn should_stop_offering_a_deposit_a_sweep_has_taken() {
-    let mut deposits = queued(&[(account(0), usdc()), (account(1), usdc())]);
+    for asset in sweepable_assets() {
+        let mut deposits = queued(&[(account(0), asset), (account(1), asset)]);
 
-    deposits.record_sweep_scheduled(SweepId(7), Asset::Erc20(usdc()), [account(0)]);
+        deposits.record_sweep_scheduled(SweepId(7), asset, [account(0)]);
 
-    let batches = deposits.requests_batch(10);
-    assert_eq!(accounts_in(&batches, usdc()), vec![account(1)]);
+        let batches = deposits.requests_batch(10);
+        assert_eq!(accounts_in(&batches, asset), vec![account(1)]);
 
-    // The taken deposit is still queued: only a settled sweep removes it.
-    assert_eq!(deposits.sweep_len(), 2);
+        // The taken deposit is still queued: only a settled sweep removes it.
+        assert_eq!(deposits.sweep_len(), 2);
+    }
 }
 
 #[test]
 fn should_offer_nothing_once_every_deposit_is_taken() {
-    let mut deposits = queued(&[(account(0), usdc()), (account(1), usdt())]);
+    let mut deposits = queued(&[(account(0), Asset::Eth), (account(1), Asset::Erc20(usdt()))]);
 
-    deposits.record_sweep_scheduled(SweepId(1), Asset::Erc20(usdc()), [account(0)]);
+    deposits.record_sweep_scheduled(SweepId(1), Asset::Eth, [account(0)]);
     deposits.record_sweep_scheduled(SweepId(2), Asset::Erc20(usdt()), [account(1)]);
 
     assert!(deposits.requests_batch(10).is_empty());
@@ -927,11 +945,21 @@ fn should_offer_nothing_once_every_deposit_is_taken() {
 
 #[test]
 #[should_panic(expected = "was already taken by another sweep")]
-fn should_refuse_to_hand_the_same_deposit_to_two_sweeps() {
-    let mut deposits = queued(&[(account(0), usdc())]);
+fn should_refuse_to_hand_the_same_erc20_deposit_to_two_sweeps() {
+    hand_a_deposit_to_two_sweeps(Asset::Erc20(usdc()));
+}
 
-    deposits.record_sweep_scheduled(SweepId(1), Asset::Erc20(usdc()), [account(0)]);
-    deposits.record_sweep_scheduled(SweepId(2), Asset::Erc20(usdc()), [account(0)]);
+#[test]
+#[should_panic(expected = "was already taken by another sweep")]
+fn should_refuse_to_hand_the_same_eth_deposit_to_two_sweeps() {
+    hand_a_deposit_to_two_sweeps(Asset::Eth);
+}
+
+fn hand_a_deposit_to_two_sweeps(asset: Asset) {
+    let mut deposits = queued(&[(account(0), asset)]);
+
+    deposits.record_sweep_scheduled(SweepId(1), asset, [account(0)]);
+    deposits.record_sweep_scheduled(SweepId(2), asset, [account(0)]);
 }
 
 #[test]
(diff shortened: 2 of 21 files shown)


rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -163,15 +164,15 @@ fn enqueue_sweep<R: CanisterRuntime>(
         if max_transaction_fee > sweeper_gas {
             log!(
                 INFO,
-                "[create_pending_sweeper_requests]: SKIPPING {token}: the sweep needs \
+                "[create_pending_sweeper_requests]: SKIPPING {asset}: the sweep needs \
```
mbjorkqvist, line 167: :robot: The gas budget behind this gate is ERC-20-shaped and is now applied unchanged to ETH.

`sweep_gas_limit` charges 165'000 per address (15'000 `balanceOf` + 110'000 approve/`depositErc20`/`transferFrom` + 40'000 authorization), so a ten-deposit sweep declares `60'000 + 10 * 165'000 = 1'710'000`. This PR's own `ETH_ATTESTED_SCENARIOS` measures a ten-deposit ETH sweep at 413'076, with a marginal `(804'603 - 413'076)/10 = 39'153` per deposit. That's 4.1x over, against 2.8x for ERC-20 (609'750) — and the ERC-20 factor is deliberate *and justified* in the constants' doc comments: `sweepErc20` walks the whole token array per address, so residue pairs each cost a balance check and possibly a transfer. None of that transfers to `sweepEth`, which reads one warm `address(this).balance` and makes one `depositEth{value: balance}` call. No token array, no `approve`, no `transferFrom`.

Not a correctness bug — unused gas is refunded — but it has a live consequence right here: ETH sweeps get skipped when the sweeper holds ~4x what they actually need, and an in-flight ETH hold crowds out ERC-20 sweeps sharing the same balance. `enqueue_sweep` already has `asset` in scope and hands `&items` straight to `sweep_gas_limit`, so threading the asset through is small. An ETH per-address budget around `SWEEP_GAS_PER_AUTHORIZATION + ~40'000` would keep the same ~2x margin the ERC-20 path was given.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -333,10 +335,13 @@ impl SweepRequest {
...
 
     /// The delegate's batch call, naming every deposit address this sweep walks and the single
-    /// token it moves.
+    /// asset it moves.
     pub fn call_data(&self) -> Vec<u8> {
         let items: Vec<_> = self.items.iter().map(|item| item.item.clone()).collect();
-        encode_sweep_erc20_batch(&items, &[self.token])
+        match self.asset {
+            Asset::Erc20(token) => encode_sweep_erc20_batch(&items, &[token]),
+            Asset::Eth => encode_sweep_eth_batch(&items),
```
mbjorkqvist, line 354: :robot: This is the one line that decides which encoder a real sweep uses, and no unit test reaches it: `grep Asset::Eth` over `src/` returns nothing in `state/transactions`, and the `sweep_lane` fixture hardcodes `asset: Asset::Erc20(Address::new([0xc0; 20]))`. The existing assertions at `tests.rs:3107` and `:3160` are `tx.data() == request.call_data()` — both sides call this same function, so swapping the two arms would pass them. The only thing that would catch a swap today is `deposit_from_cex`, which (see the review body) didn't run on this PR.

Two assertions in `unit_tests` would close it: `SweepRequest { asset: Asset::Eth, .. }.call_data() == encode_sweep_eth_batch(items)` and the ERC-20 counterpart against `encode_sweep_erc20_batch(items, &[token])`.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -205,11 +205,11 @@ impl AutomaticDeposits {
             .sweeper_transactions
             .get_processed_request(&id)
             .expect("BUG: missing sweep request");
-        let token = request.token;
+        let asset = request.asset;
```
mbjorkqvist, line 208: :robot: On a `Failure` receipt the loop below removes the queue entry for **every** account the request named, logging `DROPPING` for each — not just the one that caused the revert. And for ETH the revert really is batch-wide: `sweepEthBatch` calls `sweepEth`, a void function, so Solidity emits the `extcodesize` existence check and a single item whose address is delegated elsewhere (or not at all) reverts the whole transaction.

So one Sepolia address on the pre-#11449 delegate takes up to nine healthy deposits down with it per batch, each needing a fresh arming. The PR description reads as if only the offending address is affected; worth restating, since the scope of this failure mode is the justification for deferring R18.

rs/ethereum/cketh/minter/src/state/automatic_deposits/tests.rs
```diff
@@ -873,13 +873,9 @@ fn should_keep_eth_entries_out_of_sweep_batches() {
 
     let batches = deposits.requests_batch(10);
 
-    assert_eq!(batches.len(), 1);
+    assert_eq!(batches.len(), 2);
+    assert_eq!(accounts_in(&batches, Asset::Eth), vec![account(0)]);
```
mbjorkqvist, line 877: :robot: This proves ETH is no longer skipped, which is the important half. Two gaps remain around it:

1. **The batch-size cap is untested for any asset.** Every `requests_batch` call in this file passes `10` against two or three queued entries. The cap is now the only thing keeping an ETH batch at ten, and `assert_eth_sweep_gas_near_demo` has a baseline that is only valid for exactly ten. One test with eleven entries of a single asset would pin it.

2. **The ETH queue lifecycle is only covered by the skipped live target.** `record_sweep_scheduled` and `record_finalized_sweep_transaction` both take the asset now, but every unit test passes `Asset::Erc20(...)` — including the failure path that drops a whole batch. The cause looks mechanical: `queued(&[(Account, Address)])` can't express an ETH entry, which is why this test had to build its entry by hand through `record_automatic_deposit_received`. `automatic_deposit()` already takes `impl Into<Asset>`, so widening `queued`/`queue` to `(Account, Asset)` would let `should_stop_offering_a_deposit_a_sweep_has_taken`, `should_offer_nothing_once_every_deposit_is_taken`, `should_refuse_to_hand_the_same_deposit_to_two_sweeps` and `should_refuse_to_finalize_a_sweep_whose_deposit_left_the_queue` run for ETH as they stand.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -13,15 +13,19 @@ use ic_cketh_minter::balance_scan::batcher::{
...
     Anvil, DEV_ACCOUNT, SentTransaction, address_from_hex, deploy_mock_erc20,
 };
 use ic_cketh_test_utils::ckerc20::{CkErc20Setup, Erc20Token};
-use ic_cketh_test_utils::live::{CexDeposit, DepositPlan, LiveSetup};
+use ic_cketh_test_utils::live::{
+    CexDeposit, DepositPlan, EthCexDeposit, EthDepositPlan, LiveSetup,
+};
 use ic_cketh_test_utils::{CkEthSetup, SWEEPER_ADDRESS};
+
+const MINIMUM_ETH_DEPOSIT_WEI: u128 = 5_000_000_000_000_000;
```
mbjorkqvist: :robot: This duplicates the private `MIN_ETH_DEPOSIT` in `balance_scan/mod.rs`, while the ERC-20 twin test reads its bound from the canister via `setup.minimum_deposit_amount(&token)`. `DepositEthResponse.minimum_deposit_amount` already carries the ETH figure, so `should_flag_only_eth_deposits_at_or_above_the_minimum` can read it the same way.

Worth doing because `MIN_ETH_DEPOSIT` carries `TODO(DEFI-2961): recompute daily from the exchange-rate canister` — a hardcoded copy of a soon-to-be-dynamic threshold is exactly what that change will break.

Minor: the const also sits between two `use` lines (after `ic_cketh_test_utils::{...}`, before `ic_ethereum_types::Address`), and `cargo fmt` won't move it.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -527,13 +832,32 @@ fn should_sweep_a_second_deposit_despite_resending_a_stale_authorization() {
...
     let mints = setup
-        .minter_events()
-        .into_iter()
-        .filter(|event| matches!(event.payload, EventPayload::MintedCkErc20 { .. }))
-        .count();
+        .minter_count_events(|event| matches!(event.payload, EventPayload::MintedCkErc20 { .. }));
     assert_eq!(mints, 2, "each deposit flow must be credited exactly once");
 }
 
+fn assert_eth_sweep_gas_near_demo(sweeps: &[SentTransaction], deposits_per_sweep: u64) {
```
mbjorkqvist: :robot: This is a near-copy of `assert_sweep_gas_near_demo` differing only in the baseline constant — one function taking the baseline would do. The copy also drops the `println!` the ERC-20 version emits, which is the diagnostic you want when the 10% band trips.

Both versions also take `deposits_per_sweep` and then `assert_eq!(deposits_per_sweep, 10, ...)`, so the parameter admits exactly one value. Pre-existing in the ERC-20 one, inherited here.

mbjorkqvist (approved):
Thanks @gregorydemay!

I believe the `//rs/ethereum/cketh/minter:deposit_from_cex` test didn't run on this PR, since it's tagged `long_test`, and based on [this Slack discussion](https://dfinity.slack.com/archives/C05S62QFV1N/p1789029573114479?thread_ts=1788970237.373859&cid=C05S62QFV1N) those tests currently don't run, even if related files are changed. Would it make sense to add the `CI_ALL_BAZEL_TARGETS` label to the PR to trigger it?

gregorydemay:
> Thanks @gregorydemay!
> 
> I believe the `//rs/ethereum/cketh/minter:deposit_from_cex` test didn't run on this PR, since it's tagged `long_test`, and based on [this Slack discussion](https://dfinity.slack.com/archives/C05S62QFV1N/p1789029573114479?thread_ts=1788970237.373859&cid=C05S62QFV1N) those tests currently don't run, even if related files are changed. Would it make sense to add the `CI_ALL_BAZEL_TARGETS` label to the PR to trigger it?

Thanks for catching that @mbjorkqvist ! I added the flag and did a rebase to retrigger the build.

rs/ethereum/cketh/minter/src/sweep/mod.rs
gregorydemay: 🤖 Fixed in d58fa8543c. `sweep_gas_limit` now takes the asset and budgets an ETH address at `SWEEP_GAS_PER_ETH_DEPOSIT` (40'000, for the dispatch, the warm balance read and the `depositEth` call) plus the 40'000 authorization tuple, so a ten-deposit ETH sweep declares 860'000 against the measured 413'076 (about 2.1x, the margin the ERC-20 path has). The ERC-20 budget is unchanged. `should_scale_the_sweep_gas_limit_with_the_distinct_addresses_walked` pins both shapes and that ETH stays below ERC-20 for the same items.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Fixed in 791eebddf5. `should_encode_the_batch_call_of_the_asset_the_sweep_moves` asserts an ETH request's `call_data()` against `encode_sweep_eth_batch` and an ERC-20 request's against `encode_sweep_erc20_batch`, and that the two differ, so swapping the arms now fails a unit test.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
gregorydemay: 🤖 Agreed, the scope was understated. The PR description's deferred bullet now says a single stale-delegated Sepolia address reverts the whole `sweepEthBatch` it lands in, that a failed sweep drops every deposit it named (up to ten per batch), and that each needs a fresh `deposit_eth`. The code already did that; `should_drop_every_deposit_of_a_sweep_once_it_fails` (055688bea6) now pins the batch-wide drop for ETH and ERC-20 alike.

rs/ethereum/cketh/minter/src/state/automatic_deposits/tests.rs
gregorydemay: 🤖 Fixed in 055688bea6. (1) `should_cap_each_asset_batch_at_the_requested_size` queues eleven ETH entries plus one USDC entry and checks the ETH batch stops at ten while USDC is still offered. (2) `queued`/`queue` and the `deposits_with_enqueued_sweep` fixture now take any `impl Into<Asset>`, and the take/release/drop lifecycle tests run for both ETH and USDC (the two `should_panic` ones got an ETH twin through a shared helper).

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Fixed in 49f169f1bb. The constant is gone; the live tests read the ETH minimum through `LiveSetup::minimum_eth_deposit_amount`, which takes it from the `deposit_eth` response of an account the test deposits for anyway (the call is idempotent). The stray placement between the `use` lines went with it.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Fixed in 49c0d40673. One `assert_ten_deposit_sweep_gas_near_demo(sweeps, demo_gas)` replaces both helpers, keeps the diagnostic print for either asset, and the deposits-per-sweep parameter that only admitted ten is gone; the two demo baselines are named constants next to it.

rs/ethereum/cketh/test_utils/src/live.rs
```diff
@@ -265,6 +279,34 @@ impl LiveSetup<CkErc20Setup> {
         nat_to_u128(minimum.minimum_deposit_amount)
     }
 
+    /// The ETH minimum as `deposit_eth` reports it. Read through `caller`'s own registration,
+    /// which `deposit_eth` makes idempotently, so pass an account the test deposits for anyway.
+    pub fn minimum_eth_deposit_amount(&self, caller: Principal, subaccount: [u8; 32]) -> u128 {
+        nat_to_u128(self.deposit_eth(caller, subaccount).minimum_deposit_amount)
+    }
```
gregorydemay: That's terrible, this method is calling `deposit_eth` which has side-effects. It should get the minimum from minter_info like `minimum_deposit_amount`. One could think of making this method generic over `Into<Asset>` since this is the same logic

gregorydemay: 🤖 Fixed in d346d3328d: `get_minter_info` now reports a backward-compatible `minimum_eth_deposit_amount : opt nat` (populated from the same `min_deposit(&Asset::Eth)` the balance scan uses), and the harness reads both asset minimums through one side-effect-free `minimum_deposit_amount(asset: impl Into<Asset>)`. One deviation from the sketch: `Erc20Token` re-exports the orchestrator's `AddErc20Arg`, so the orphan rule forbids `impl From<&Erc20Token> for Asset` in test utils; ERC-20 call sites pass `contract_address(token)` instead, which converts via `From<Address>`.

gregorydemay:
@mbjorkqvist I asked for a re-review because I added a part that was missing: report the minimum eth deposition amount in `get_minter_info`, see d346d3328d844fbff40649659f6d528bdacdb46a

mbjorkqvist:
> @mbjorkqvist I asked for a re-review because I added a part that was missing: report the minimum eth deposition amount in `get_minter_info`, see [d346d33](https://github.com/dfinity/ic/commit/d346d3328d844fbff40649659f6d528bdacdb46a)

Thanks, LGTM!

mbjorkqvist (approved):

PR #11533, opened 2026-09-10
chore: cargo clippy fixes to prepare for the rustc upgrade: 1.97.1 -> 1.98.1

basvandijk:
`cargo clippy` fixes to prepare for upgrading to
[rustc-1.98.1](https://blog.rust-lang.org/2026/08/20/Rust-1.98.0/).

Clippy 0.1.98 reports **85 warnings across 23 files** that Clippy 0.1.97 does not. 11 of them
come from two lints that are new in 1.98; the other 74 come from existing lints whose reach
1.98 widened:

| lint | # | why it fires now | fix |
| --- | --: | --- | --- |
| [`useless_format`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#useless_format) | 39 | 1.98 also lints through macro expansions, so `format!("some literal")` inside `slog::debug!` and the prometheus `register_*_with_registry!` macros is now reported | pass the literal directly (see below) |
| [`result_large_err`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#result_large_err) | 28 | [#17130](https://github.com/rust-lang/rust-clippy/pull/17130) fixed the lint to also fire on `async fn` | `#[allow(clippy::result_large_err)]` |
| [`chunks_exact_to_as_chunks`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#chunks_exact_to_as_chunks) | 8 | new lint in 1.98 ([#16931](https://github.com/rust-lang/rust-clippy/pull/16931)) | rewritten to `slice::as_chunks::<N>()` / `slice::get(..N)` |
| [`single_range_in_vec_init`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#single_range_in_vec_init) | 5 | [#17146](https://github.com/rust-lang/rust-clippy/pull/17146) made it detect more range shapes | `#[allow(...)]` — false positive here, see below |
| [`map_or_identity`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#map_or_identity) | 3 | new lint in 1.98 | `x.map_or(d, \|v\| v)` → `x.unwrap_or(d)` |
| [`drain_collect`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#drain_collect) | 1 | | `v.drain(..).collect()` → `std::mem::take(&mut v)` |
| [`needless_late_init`](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#needless_late_init) | 1 | [#16746](https://github.com/rust-lang/rust-clippy/pull/16746) extended it to grouped assignments | `#[allow(...)]` — naming each binding at its assignment reads better than a 5-tuple destructure |

Three of these deserve a word:

* **`useless_format`** sites just pass the string literal now, rather than clippy's suggested
  `.to_string()`: `slog`'s `Value` is implemented for `&'static str` and the prometheus
  `register_*_with_registry!` macros take `Into<String>`, so no allocation or conversion is
  needed. That is also what these files already do elsewhere — `rs/crypto/src/sign/mod.rs` logs
  a bare literal at 90 other sites, and `ic_boundary`'s `metrics.rs` registers
  `"check_total"`/`"check_status"` that way. (The one exception is
  `manual_guestos_recovery/src/ui.rs`, where the literal is an element of a `Vec<String>` and
  so does need `.to_string()`.)

* **`result_large_err`** is silenced with `#[allow]` rather than fixed, in line with the ~39
  existing uses of that allow in this repo. Boxing these error types would churn signatures
  across their call sites for no practical gain — none of these paths is performance-sensitive,
  which is the whole point of the lint. If any of them is worth boxing, that is a change worth
  making on its own merits rather than as a side effect of a toolchain bump.

* **`single_range_in_vec_init`** is a false positive at all 5 sites: those `Vec`s intentionally
  hold canister ID ranges *as elements* and are passed to a `fn(Vec<RangeInclusive<u64>>)`, so
  clippy's "did you mean to `.collect()` the range" suggestion does not apply. Allowed at the
  module level with a comment.

Note also that clippy's machine-applicable suggestion for `chunks_exact_to_as_chunks` does not
compile at 5 of the 8 sites — a const-generic argument that is not a literal needs braces
(`as_chunks::<{ std::mem::size_of::<u64>() }>()`), and `&[]` is not a `&[T; N]` — so those were
written by hand. `blob.chunks_exact(32).next()` became the equivalent and clearer
`blob.get(..32)`.

Because this PR lands *before* the toolchain bump, it has to stay clean on 1.97.1 too, and that
decides which lints may be silenced and which must be rewritten. `#[allow]` is used only for
lints that Clippy 0.1.97 already knows (`result_large_err`, `single_range_in_vec_init`,
`needless_late_init`) — 1.98 merely widened where those apply. The two lints that are genuinely
*new* in 1.98 (`chunks_exact_to_as_chunks`, `map_or_identity`) are fixed in the code, because
allowing them would trip rustc's `unknown_lints` under `--deny warnings` on the currently
pinned toolchain.

The toolchain upgrade itself is the PR stacked on top of this one.

## Testing

`./ci/scripts/rust-lint.sh` exits 0 both on the currently pinned **1.97.1** and on **1.98.1**.

### DIFF
--- a/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
+++ b/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
@@ -52,28 +52,19 @@ fn main() -> Result<()> {
 fn process_cmdline(input: &str) -> Result<String> {
     let boot_args_re = Regex::new(r"(^|\n)BOOT_ARGS=(.*)(\s+#|\n|$)").unwrap();
 
-    let left;
-    let indent;
-    let boot_args;
-    let tail;
-    let right;
-    match boot_args_re.captures(input) {
+    let (left, indent, boot_args, tail, right) = match boot_args_re.captures(input) {
         Some(captures) => {
             let whole_match = captures.get(0).unwrap();
 
-            left = whole_match.start();
-            indent = captures.get(1).unwrap().as_str();
-            boot_args = captures.get(2).unwrap().as_str().trim().trim_matches('"');
-            tail = captures.get(3).unwrap().as_str();
-            right = whole_match.end();
+            (
+                whole_match.start(),
+                captures.get(1).unwrap().as_str(),
+                captures.get(2).unwrap().as_str().trim().trim_matches('"'),
+                captures.get(3).unwrap().as_str(),
+                whole_match.end(),
+            )
         }
-        None => {
-            left = input.len();
-            indent = "";
-            boot_args = "";
-            tail = "\n";
-            right = input.len();
-        }
+        None => (input.len(), "", "", "\n", input.len()),
     };
 
     let mut cmdline = KernelCommandLine::from_str(boot_args)?;
--- a/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
+++ b/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
@@ -49,22 +49,34 @@ fn main() -> Result<()> {
 }
 
 /// Disable checks from the kernel command line
+// Naming each binding at its assignment is clearer here than destructuring a 5-tuple,
+// where it is easy to misread which value ends up in which variable.
+#[allow(clippy::needless_late_init)]
 fn process_cmdline(input: &str) -> Result<String> {
     let boot_args_re = Regex::new(r"(^|\n)BOOT_ARGS=(.*)(\s+#|\n|$)").unwrap();
 
-    let (left, indent, boot_args, tail, right) = match boot_args_re.captures(input) {
+    let left;
+    let indent;
+    let boot_args;
+    let tail;
+    let right;
+    match boot_args_re.captures(input) {
         Some(captures) => {
             let whole_match = captures.get(0).unwrap();
 
-            (
-                whole_match.start(),
-                captures.get(1).unwrap().as_str(),
-                captures.get(2).unwrap().as_str().trim().trim_matches('"'),
-                captures.get(3).unwrap().as_str(),
-                whole_match.end(),
-            )
+            left = whole_match.start();
+            indent = captures.get(1).unwrap().as_str();
+            boot_args = captures.get(2).unwrap().as_str().trim().trim_matches('"');
+            tail = captures.get(3).unwrap().as_str();
+            right = whole_match.end();
         }
-        None => (input.len(), "", "", "\n", input.len()),
+        None => {
+            left = input.len();
+            indent = "";
+            boot_args = "";
+            tail = "\n";
+            right = input.len();
+        }
     };
 
     let mut cmdline = KernelCommandLine::from_str(boot_args)?;
--- a/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
+++ b/rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
@@ -49,8 +49,7 @@ fn main() -> Result<()> {
 }
 
 /// Disable checks from the kernel command line
-// Naming each binding at its assignment is clearer here than destructuring a 5-tuple,
-// where it is easy to misread which value ends up in which variable.
+// Naming each binding at its assignment is clearer than destructuring a 5-tuple.
 #[allow(clippy::needless_late_init)]
 fn process_cmdline(input: &str) -> Result<String> {
     let boot_args_re = Regex::new(r"(^|\n)BOOT_ARGS=(.*)(\s+#|\n|$)").unwrap();
--- a/rs/monitoring/metrics_proxy/src/client.rs
+++ b/rs/monitoring/metrics_proxy/src/client.rs
@@ -53,6 +53,8 @@ impl From<HttpError> for ScrapeError {
 ///
 /// # Errors
 /// * `ScrapeError`
+// Boxing the error would be a breaking change of this function's public API.
+#[allow(clippy::result_large_err)]
 pub async fn scrape(
     client: reqwest::Client,
     c: &crate::config::ConnectTo,
--- a/rs/monitoring/metrics_proxy/src/client.rs
+++ b/rs/monitoring/metrics_proxy/src/client.rs
@@ -53,7 +53,6 @@ impl From<HttpError> for ScrapeError {
 ///
 /// # Errors
 /// * `ScrapeError`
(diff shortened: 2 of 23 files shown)


nmattia (approved):

rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
```diff
@@ -52,28 +52,19 @@ fn main() -> Result<()> {
...
-            tail = captures.get(3).unwrap().as_str();
-            right = whole_match.end();
-        }
-        None => {
-            left = input.len();
-            indent = "";
-            boot_args = "";
-            tail = "\n";
-            right = input.len();
+            (
```
frankdavid: Is this necessary? It's much less readable and more error-prone, which value gets assigned to which variable

basvandijk: Yeah I agree. Let me find out ...

rs/monitoring/metrics_proxy/src/client.rs
```diff
@@ -53,6 +53,8 @@ impl From<HttpError> for ScrapeError {
 ///
 /// # Errors
 /// * `ScrapeError`
+// Boxing the error would be a breaking change of this function's public API.
+#[allow(clippy::result_large_err)]
```
schneiderstefan:
```suggestion
#[allow(clippy::result_large_err)]
```

This comment makes only sense if you read it in the context of what this PR is doing. Doesn't add anything for anybody reading the code in the future.

schneiderstefan (approved):

rs/ic_os/dev_test_tools/setupos-disable-checks/src/main.rs
basvandijk: You're right, and it turns out it wasn't necessary. Reverted in 8201146.

`clippy::needless_late_init` already exists in Clippy 0.1.97 — 1.98 only extended its *reach* to grouped assignments ([#16746](https://github.com/rust-lang/rust-clippy/pull/16746)) — so `#[allow(clippy::needless_late_init)]` is available here and keeps the original code clean on both toolchains.

That's the distinction that drove the rest of this PR: since it lands *before* the toolchain bump it has to stay clean on 1.97.1 too, so `#[allow]` only works for lints 0.1.97 already knows. For the two lints that are genuinely new in 1.98 (`chunks_exact_to_as_chunks`, `map_or_identity`) an allow would trip rustc's `unknown_lints` under `--deny warnings` on the current toolchain, so those had to be rewritten. I over-applied the rewrite here — this one had the easier way out.

rs/monitoring/metrics_proxy/src/client.rs
basvandijk: But wouldn't you, as a future reader, want to know why the [`result_large_err`](https://rust-lang.github.io/rust-clippy/master/#result_large_err) lint has to be allowed here?

rs/monitoring/metrics_proxy/src/client.rs
```diff
@@ -53,6 +53,8 @@ impl From<HttpError> for ScrapeError {
 ///
 /// # Errors
 /// * `ScrapeError`
+// Boxing the error would be a breaking change of this function's public API.
```
frankdavid: That's fairly clear, I don't think it's worth mentioning as a comment. Btw, there seems to be a single user of this function that shouldn't be a reason not to move to Box. The better reason is that we don't care about moving 132 bytes, since this path is not performance-sensitive. But even just boxing HttpError in the Non200 enum variant would work. I don't think we need to address any of this but the comment is misleading.

rs/monitoring/metrics_proxy/src/client.rs
basvandijk: But it might indeed be obvious to the informed reader. I'll remove the comment.

rs/tests/driver/src/driver/test_env_api.rs
```diff
@@ -304,7 +304,7 @@ impl std::fmt::Display for TopologySnapshot {
                 "\tNode id={}, ipv6={:<width$}, domain_name={}, index={}",
                 n.node_id,
                 n.get_ip_addr(),
-                n.get_domain().map_or("n/a".to_string(), |domain| domain),
+                n.get_domain().unwrap_or("n/a".to_string()),
                 idx,
```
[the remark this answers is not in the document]
basvandijk: Good catch, applied in 7fc7064 — `unwrap_or_else(|| "n/a".to_string())`.

(The pre-existing `map_or("n/a".to_string(), |domain| domain)` was eager too, so this is a small improvement over what was there rather than a regression I introduced, but there's no reason not to take it.)

rs/monitoring/metrics_proxy/src/client.rs
basvandijk: Fair enough — removed in 7fc7064, along with the equivalent comments above every other `#[allow(clippy::result_large_err)]` in this PR, for consistency.

You and @frankdavid convinced me: they were rationale for *this PR* rather than for the code, and on a closer look two of them were also just wrong (see my reply to Frank below). The reasoning now lives in the PR description, which is where the PR context belongs.

rs/monitoring/metrics_proxy/src/client.rs
basvandijk: You're right on both counts, thanks — `scrape` has exactly one caller (`proxy.rs:236`) in what is a binary crate, so "breaking change of this function's public API" was wrong, and the real reason is simply that moving 132 bytes on a non-perf-sensitive path doesn't matter.

I've dropped the comment in 7fc7064 (and every other `result_large_err` comment in this PR). While re-checking them I found a second wrong one: I'd claimed boxing the cketh minter's error would break its Candid interface, but `Box<T>` is transparent to candid, so it wouldn't have.

Agreed on not addressing the boxing itself here — if it's worth doing, it's worth doing as its own change rather than as a side effect of a toolchain bump.

mbjorkqvist (approved):

frankdavid (approved):

bjoernek (approved):

PR #11536, opened 2026-09-10
feat: [DSM-147] Strict XNet slice byte limits for suffix pulls

alin-at-dfinity:
`SizeLimitVisitor` includes the first message regardless of `byte_limit`, so a stream cannot stall behind a message the caller has no room for. A suffix pull needs no such guarantee: the caller already holds some messages and can make progress. The endpoint infers this from `witness_begin == msg_begin`, so nothing changes on the wire.

A stream that has filled its share of the pool therefore stops drawing messages, while one with nothing pooled is still guaranteed one. Which in turn lets us drop the `SLICE_BYTE_SIZE_MIN` floor: a request whose remaining allowance is too small for a message now returns a fresh header and witness, which does not grow the pool.

`encode_partial_slice_test` compared a partial slice's payload against a full slice encoded with the same `byte_limit`, which now differ whenever that limit is below one message. It takes the message count from the partial slice's own payload instead, so the assertions no longer depend on matching limit semantics.

### DIFF
--- a/rs/xnet/payload_builder/src/lib.rs
+++ b/rs/xnet/payload_builder/src/lib.rs
@@ -1413,8 +1413,8 @@ pub fn adjusted_byte_limit(slice_byte_size: usize) -> usize {
     slice_byte_size.saturating_sub(350) * 98 / 100
 }
 
-/// Computes `RefillStreamSliceIndices` for every subnet whose stream slices should be refilled
-/// in the given certified slice pool owned by the given subnet.
+/// Computes `RefillStreamSliceIndices` for every subnet with a cached stream
+/// position in the given certified slice pool owned by the given subnet.
 pub fn refill_stream_slice_indices(
     pool_lock: Arc<Mutex<CertifiedSlicePool>>,
     own_subnet_id: SubnetId,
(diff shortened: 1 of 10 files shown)


rs/xnet/payload_builder/src/lib.rs
```diff
@@ -1467,11 +1467,10 @@ pub fn refill_stream_slice_indices(
             ),
         };
 
-        if slice_byte_limit < SLICE_BYTE_SIZE_MIN {
-            // No more space left in the pool for this slice, skip it.
-            continue;
-        }
-
+        // Poll every subnet regardless of how low `slice_byte_limit` is:
```
schneiderstefan, line 1477: Doesn't that open us up for shooting over the limit constantly, especially with malicious subnets? Or is the pool so small that DOS is not really a concern?

schneiderstefan (approved):

rs/xnet/payload_builder/src/lib.rs
alin-at-dfinity: On the one hand, you don't need a malicious subnet, just a bunch of subnets spamming you (e.g. the NNS) with either 2 MB messages (so you're always holding on to at least 2 MB for every stream) or with thousands of small messages (so that you're constantly pulling, validating and replacing 200 KB witnesses for each).

On the other hand, the previous behavior prevented us from pulling any fresh stream headers (to place in front of already pooled messages) as long as we were above the pool size limit.

I suppose neither is ideal, but as you point out, there are relatively few peers that a subnet (even the NNS) communicates with concurrently, so this would amount to tens of MB at most, in practice. If it turns out to be a problem in the future, we can always introduce a hard limit on top.

alin-at-dfinity: Added a log message and counter to track XNet pulls returning more bytes than requested. Unless I got the error margins wrong and we observe the log message / counter growing, we can later replace it with returning a `CertifiedSliceError` (which would drop the slice). Or we can actively truncate the slice to what we want it to be.

PR #4220, opened 2025-03-04
test(EXC-1982): Benchmark compilation cost

adambratschikaye:
EXC-1982

Add a "benchmark" in our compilation time bench suite that reports the compilation cost of each Wasm and the expected compilation time based on that cost. This will allow us to see how accurate the compilation costs are.

The benchmark test cases previously had hard-coded compilation costs which would need to be updated if the compilation costs changed. These are no longer needed now that we have a simple way to display the compilation costs along with the compilation time benchmark results.

PR #4202, opened 2025-03-03
chore: refactor to use test_strategy in Crypto Tree Hash proptests

stiegerc:
Same story has [this](https://github.com/dfinity/ic/pull/4163).

PR #11537, opened 2026-09-10
chore(bazel): upgrade rules_rust to 0.74.0

basvandijk:
Upgrades `rules_rust` from 0.71.3 to 0.74.0, stops depending on the `dfinity/rules_rust` fork, and drops two local patches that 0.74.0 makes redundant.

## Drops the fork

The pin was an `archive_override` on `dfinity/rules_rust`, whose sole commit adds a `strip_level_select` tag class ([bazelbuild/rules_rust#4167](https://github.com/bazelbuild/rules_rust/pull/4167), still open upstream). That tag's only caller was removed in 9e2920593e when hermetic-llvm replaced zig cc — clang emits deterministic debug info, so the `strip_level_select(none)` workaround went away with it. `strip_level` now appears nowhere in the repo outside the patch file, so the fork is dead code.

```diff
-bazel_dep(name = "rules_rust", version = "0.71.3")
-archive_override(
+bazel_dep(name = "rules_rust", version = "0.74.0")
+single_version_override(
     module_name = "rules_rust",
-    integrity = "sha256-5GXwZtp5VlNxMI+dolYYxhyhG5f5ReZ5qz30gAbPsrg=",
     patch_strip = 1,
     patches = ["//bazel:rules_rust_static_runtime.patch"],
-    strip_prefix = "rules_rust-f8f382c7e351a797242ce4a92262342f8994e265",
-    type = "tar.gz",
-    urls = ["https://codeload.github.com/dfinity/rules_rust/tar.gz/f8f382c7e351a797242ce4a92262342f8994e265"],
 )
```

No integrity hash, no `strip_prefix`, no URL, and no fork to rebase on every future release — a bump is now a one-line version change. Checked before switching: BCR's `rules_rust/0.74.0/source.json` carries no patches of its own, and `rules_rust` is depended on **only by the root module** (`bazel mod explain rules_rust`), so nothing else can pull MVS off 0.74.0.

## Regenerates `bazel/rules_rust_static_runtime.patch`

The `rust_doc_test` half of the patch is **upstream as of 0.74.0** ([#4220](https://github.com/bazelbuild/rules_rust/pull/4220), commit `ed7afd2fc`) and is a superset of ours — it threads the runtime libs through `rustdoc_compile_action`'s return value rather than recomputing them, adds an empty-lib-root guard we lacked (an empty root would emit `--strip_substring=/`, which the writer applies as a plain string replacement), and ships a unit test. So the patch no longer carries it.

What remains is the dylib/cdylib static-runtime preference, which is still not upstream and has no replacement mechanism (no new `unwindlib`, toolchain or rule attribute controls C++ runtime linkage in this range). #4220 also introduced a **new** site in `rustdoc.bzl` that picks the dynamic runtime for dylib/cdylib, which the old patch would have missed — leaving a cdylib doc test with the static runtime in its action inputs but the dynamic runtime's names in its `-l` flags. The policy therefore now lives in a single exported `cc_runtime_libs` helper, shared by all three sites that choose a runtime (`collect_inputs`, `_add_native_link_flags`, and rustdoc's doc-test link flags), so inputs and link flags stay consistent by construction.

The patch applies with **zero fuzz and zero offset** (`git apply --check` and `patch -F0`), verified against the release tarball specifically, not just the git tag.

## Drops `bazel/rustix.patch`

That patch deleted rustix's `can_compile()` probe artifact from `OUT_DIR` — which rules_rust captures as a cacheable `CargoBuildScriptRun` output — because the artifact was not byte-reproducible. 0.74.0 makes it unnecessary: [#4202](https://github.com/bazelbuild/rules_rust/pull/4202) adds `--remap-path-prefix=${pwd}=.` to the `CARGO_ENCODED_RUSTFLAGS` that `cargo_build_script` passes to build scripts, and rustix's `can_compile()` forwards all of them to its probe `rustc`.

## Follow-ups (not in this PR)

- The remaining dylib/cdylib half of the static-runtime patch has no upstream issue. Opening one would let it retire the way the doc-test half just did — though upstream deliberately links the dynamic runtime for shared libraries (its own `runtime_libs_test/shared_library` asserts exactly that), so it would need a toolchain-level opt-in rather than a straight behaviour change.
- `.claude/CLAUDE.md`'s repin trigger could mention the `rules_rust` pin, since a bump on its own requires `./bin/bazel-pin.sh --force`.
- If the rustix probe ever regresses (a toolchain bump, or anything adding a remap-scope rustflag), the fix is rules_rust 0.74.0's new `remove_nondeterministic_out_dir_files`: appending `rustix_test_can_compile` to `--@rules_rust//cargo/settings:out_dir_volatile_file_basenames` (re-listing its seven defaults, since the flag replaces rather than extends) empties `OUT_DIR` again with no crate patch and no repin. That, not a resurrected `bazel/rustix.patch`.
- Worth an upstream issue: `cargo_build_script` could pass the same `${output_base}`/`${exec_root}` remaps that `rustc.bzl` already passes, closing the symlink gap described above.
- The `dfinity/rules_rust` fork branches can be deleted once this merges; nothing will reference the fork any more.


bazel/rules_rust_static_runtime.patch
```diff
@@ -1,130 +1,161 @@
...
+# symbols rust std needs, since the toolchain links with `--unwindlib=none`,
+# and the dynamic one consists of versioned files (libunwind.so.1) that a
+# `-l` flag never matches. Shared-library links don't fail on undefined
+# symbols, so the breakage only surfaces at runtime: dlopen of e.g.
+# libnss_icos.so fails, leaving GuestOS unable to resolve "hostos" and nft
+# unable to load the orchestrator's firewall rules.
+#
+# This patch prefers the static runtime for shared libraries too, falling back
+# to the dynamic one for toolchains that ship no static runtime (the local
+# toolchains' system libstdc++.so). The policy lives in one new helper,
```
nmattia: "one new helper" it's only new now. Also the helper has 30 lines of documentation (for 3 lines of code) so I think we can trim this comment down a bit here.

bazel/rules_rust_static_runtime.patch
```diff
@@ -1,130 +1,161 @@
...
+# `cc_runtime_libs`, shared by the three sites that pick a runtime: the action
+# inputs (`collect_inputs`), the rustc link flags (`_add_native_link_flags`)
+# and rustdoc's doc-test link flags.
+#
+# The `rust_doc_test` half of this fix — doc tests dropping the runtime libs'
+# `-l` flags and `-Lnative=` roots entirely — is upstream as of 0.74.0
+# (commit ed7afd2fc,
+# https://github.com/bazelbuild/rules_rust/pull/4220), so this patch no
```
nmattia: How is this relevant?

Cargo.Bazel.toml.lock
```diff
@@ -2,15 +2,6 @@
 # It is not intended for manual editing.
 version = 4
 
-[[package]]
```
nmattia, line 5: do you know why the upgrade impacted the dependency graph?

nmattia:
> bazel/rustix.patch may now be redundant: upstream https://github.com/bazelbuild/rules_rust/pull/4202 fixes issue https://github.com/dfinity/ic/pull/4201, the same rustix build-script probe defect that patch works around. Worth checking separately rather than in a version bump.

Let's check here; no point carrying a patch that's unnecessary

Cargo.Bazel.toml.lock
basvandijk: It was done by [my last `bin/bazel-pin.sh`](https://github.com/dfinity/ic/pull/11537/commits/cc292ad02a576856c4b11bb80259a41433a7b972) this morning. I don't know why these got removed.

bazel/rules_rust_static_runtime.patch
basvandijk: I trimmed it a bit.

bazel/rules_rust_static_runtime.patch
basvandijk: I personally like these references to related code since it records who to contact / what is relevant if we ever need to upstream this. But I can also see this being too noisy so I removed it.

nmattia (approved):

PR #11544, opened 2026-09-11
feat: disable unused features

nmattia:
This goes through a few crates and disables a few unused features, clearing up the dependency graph. Additionally the `crossbeam` crate is replaced with `crossbeam-utils` (smaller but exports everything we need) and `json-patch` is removed and the single function used is inlined.


basvandijk (approved):

schneiderstefan (approved):

bjoernek:
Nice cleanup. One suggestion on `rs/sns/cli/src/deploy.rs`:

I diffed the new `json_merge_patch` against `json-patch` 0.2.6's `merge` and it's semantically identical, so no behavior change. But the merge semantics used to be backed by the upstream crate's RFC 7396 test suite, and there's currently no test covering `json_merge_patch` or `merge_into_json_file`. Could you add a small unit test? Four cases would cover it: recursive object merge, `null` deletes a key, scalar replaces an object, and a non-object patch replacing the whole doc.

bjoernek (approved):

PR #11545, opened 2026-09-11
feat: [DSM-148] `CertifiedSlicePool`: highest seen peer headers

alin-at-dfinity:
Record the latest (highest stream indices / greatest certified height) stream header observed for every peer subnet in `CertifiedSlicePool`.

This is to be used for triggering and handling XNet adverts (do we have any signals to advertise relative to the peer's `begin`; and, respectively, are there any new messages or signals for us to pull).

### DIFF
--- a/rs/xnet/payload_builder/tests/certified_slice_pool.rs
+++ b/rs/xnet/payload_builder/tests/certified_slice_pool.rs
@@ -14,7 +14,7 @@ use ic_test_utilities_metrics::{HistogramStats, metric_vec};
 use ic_test_utilities_state::arb_stream_slice;
 use ic_test_utilities_types::xnet::StreamSliceBuilder;
 use ic_types::messages::MAX_XNET_PAYLOAD_SIZE_ERROR_MARGIN_PERCENT;
-use ic_types::xnet::{CertifiedStreamSlice, StreamIndex, StreamSlice};
+use ic_types::xnet::{CertifiedStreamSlice, StreamHeader, StreamIndex, StreamSlice};
 use ic_types::{CountBytes, RegistryVersion, SubnetId};
 use ic_xnet_payload_builder::certified_slice_pool::{
     CertifiedSliceError, CertifiedSlicePool, CertifiedSliceResult, InvalidAppend, InvalidSlice,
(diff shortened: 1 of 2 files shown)


rs/xnet/payload_builder/tests/certified_slice_pool.rs
```diff
@@ -1399,7 +1403,7 @@ fn pool_append_invalid_slice_to_empty(
 fn pool_take_slice_respects_signal_limit(
     #[strategy(arb_stream_slice(
         MAX_SIGNALS, // min_size
-        2 * MAX_SIGNALS, // max_size
+        MAX_SIGNALS + 100, // max_size
```
alin-at-dfinity, line 1406: Drive-by adjustment: We don't need 20k messages to validate that we only take at most 10k. This test case had a runtime of 12 seconds on my devenv, now it's down to "only" 8 seconds.

schneiderstefan (commented):
Are these high watermarks nondeterministic, as they reset on restart? Will that cause issues with our logic of when to send adverts?

alin-at-dfinity:
Yes, they are non-deterministic (both upon restart and across replicas) but it's a benign state: immediately after a replica starts up, it will appear to it that everyone needs to be notified of its signals (because everyone's `begin` is implicitly zero); and (potentially, until it receives some adverts) that it has nothing to pull from other subnets iff it had previously inducted all their messages.

But it's transient: as soon as the adverts/pulls start flying we're back to the steady state where we only advertise when necessary.

schneiderstefan (approved):
Don't you need for the other subnet to send you an advert so that you get out of this state? Or how does it exactly work to populate these fields after you send the first, unsolicited advert?

alin-at-dfinity:
That is actually a good question. I had only explicitly considered the steady state behavior (what happens when some index advances) plus, briefly, what would happen with a missing entry (default of zero means we advertise until we learn any better).

But it all works anyway: a missing entry is no different from a stale one (the indices are all assumed to be zero, but we only care about equal or not equal, i.e. less). So as long as we can handle the stale entry case, we can handle the no entry case.

And we do that by having the peer reply with its header whenever we advertise a header that tells it nothing new -- on the assumption that we do this because we think we have signals for it; but it knows that it has already acted on those signals, so we need to garbage collect them. Turns out that this mechanism works for any situation where we have an out-of-date view of the world. And if we're not actually out-of-date (i.e. we do have something to advertise) then the peer will act on our new messages/signals and it will notify us to garbage collect them.

So it's always a short-lived, transient state, with a couple of adverts sent to each peer subnet before things settle down.

PR #11546, opened 2026-09-11
feat: Don't gossip HTTPS outcalls responses that were already delivered

eichhorl:
## Background
After an (aggregated) HTTPS outcalls response is delivered to the caller, the corresponding context is moved into a separate `delivered` collection in the call context manager. The call context remains there for up to one minute, in order to accept additional asynchronous responses from peers that were not included in the original aggregate. This is because these responses may still carry cycles refunds which should be credited back to the calling canister.

## Problem
If a node sees that the aggregated response was already delivered, it will not even attempt to make the HTTP request (if it hasn't already). However, ongoing requests are still completed.

Until now, the response body of these ongoing requests was then still gossiped to peers, even if there is no chance that the response will make it into a block (at this point we are only interested in the cycles refund share).

## Proposed Changes
With this PR, we no longer gossip the response body for requests that already received a response. This means when we receive a response without a body, we defer it (instead of rejecting), because it may have been from a node that's ahead of us, and already know that the context was moved to `delivered`.

Additionally, we add a candid decoder config to the payload builder, to be consistent with other management canister types.

### DIFF
--- a/rs/artifact_pool/src/canister_http_pool.rs
+++ b/rs/artifact_pool/src/canister_http_pool.rs
@@ -180,13 +180,15 @@ impl MutablePool<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
                             .insert(ic_types::crypto::crypto_hash(&content), content);
                     }
                 }
-                CanisterHttpChangeAction::MoveToValidated(share) => {
+                CanisterHttpChangeAction::MoveToValidated {
+                    share,
+                    retain_response,
+                } => {
                     if let Some(artifact) = self.unvalidated.remove(&share) {
-                        // If there is a response associated with this share, we want to move it to the `content`
-                        // section of the pool, corresponding to valid responses. A validated share carries a
-                        // response exactly for non-fully-replicated requests, whose response must be served
-                        // to peers that pull the artifact.
-                        let served_response = if let Some(content) = artifact.response {
+                        // A response that should not be retained is dropped. This means, it will
+                        // neither be served to other peers nor put into a block by this node.
+                        let content = artifact.response.filter(|_| retain_response);
+                        let served_response = if let Some(content) = content {
                             self.content
                                 .insert(ic_types::crypto::crypto_hash(&content), content);
                             ServedResponse::Include
@@ -508,9 +510,10 @@ mod tests {
             peer_id: node_test_id(0),
             timestamp: UNIX_EPOCH,
         });
-        pool.apply(vec![CanisterHttpChangeAction::MoveToValidated(
-            share.clone(),
-        )]);
+        pool.apply(vec![CanisterHttpChangeAction::MoveToValidated {
+            share: share.clone(),
+            retain_response: true,
+        }]);
         assert_eq!(pool.get(&id).unwrap().response, Some(response));
 
         // Fully replicated, received from a peer (MoveToValidated without a
@@ -518,9 +521,10 @@ mod tests {
         let share = fake_share(4);
         let id = share.clone();
         pool.insert(to_unvalidated(share.clone()));
-        pool.apply(vec![CanisterHttpChangeAction::MoveToValidated(
-            share.clone(),
-        )]);
+        pool.apply(vec![CanisterHttpChangeAction::MoveToValidated {
+            share: share.clone(),
+            retain_response: true,
+        }]);
         assert!(pool.get(&id).unwrap().response.is_none());
 
         // Non-fully-replicated share whose response content is missing: `get`
@@ -549,6 +553,27 @@ mod tests {
             None,
         )]);
         assert!(pool.get(&id).unwrap().response.is_none());
+
+        // Already responded to, received from a peer that still attached a response:
+        // the response is dropped, so it is neither retained nor served on pull.
+        let response = fake_response(7);
+        let share = fake_share_matching(7, &response);
+        let id = share.clone();
+        let content_hash = ic_types::crypto::crypto_hash(&response);
+        pool.insert(UnvalidatedArtifact {
+            message: CanisterHttpResponseArtifact {
+                share: share.clone(),
+                response: Some(response),
+            },
+            peer_id: node_test_id(0),
+            timestamp: UNIX_EPOCH,
+        });
+        pool.apply(vec![CanisterHttpChangeAction::MoveToValidated {
+            share: share.clone(),
+            retain_response: false,
+        }]);
+        assert!(pool.get(&id).unwrap().response.is_none());
+        assert!(pool.get_response_content_by_hash(&content_hash).is_none());
     }
 
     #[test]
@@ -562,8 +587,14 @@ mod tests {
         pool.insert(to_unvalidated(share1.clone()));
 
         let result = pool.apply(vec![
-            CanisterHttpChangeAction::MoveToValidated(share2.clone()),
-            CanisterHttpChangeAction::MoveToValidated(share1.clone()),
+            CanisterHttpChangeAction::MoveToValidated {
+                share: share2.clone(),
+                retain_response: true,
+            },
+            CanisterHttpChangeAction::MoveToValidated {
+                share: share1.clone(),
+                retain_response: true,
+            },
         ]);
 
         assert!(pool.lookup_validated(&id2).is_none());
--- a/rs/artifact_pool/src/canister_http_pool.rs
+++ b/rs/artifact_pool/src/canister_http_pool.rs
@@ -42,8 +42,11 @@ enum ServedResponse {
     /// `CanisterHttpResponse` is produced by only a subset of nodes, so it must
     /// be included with the artifact.
     Include,
-    /// The request is fully replicated: every node recomputes the response
-    /// locally, so it must be excluded from the artifact.
+    /// The response must be excluded from the artifact: either because the
+    /// request is fully replicated, i.e. every node recomputes the response
+    /// locally; or because the request has already been responded to, in which
+    /// case the share is nothing but a receipt and there may be no response in
+    /// the pool at all.
     Exclude,
 }
 
@@ -170,8 +173,12 @@ impl MutablePool<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
                     }));
                     // Response content should not be served
                     self.validated.insert(share, ServedResponse::Exclude);
-                    self.content
-                        .insert(ic_types::crypto::crypto_hash(&content), content);
+                    // A share of an already answered outcall comes with no response to
+                    // retain: it can never be put into a block.
+                    if let Some(content) = content {
+                        self.content
+                            .insert(ic_types::crypto::crypto_hash(&content), content);
+                    }
                 }
                 CanisterHttpChangeAction::MoveToValidated(share) => {
                     if let Some(artifact) = self.unvalidated.remove(&share) {
@@ -360,8 +367,8 @@ mod tests {
         let content_hash = ic_types::crypto::crypto_hash(&response);
 
         let result = pool.apply(vec![
-            CanisterHttpChangeAction::AddToValidated(share.clone(), response.clone()),
-            CanisterHttpChangeAction::AddToValidated(fake_share(456), fake_response(456)),
+            CanisterHttpChangeAction::AddToValidated(share.clone(), Some(response.clone())),
+            CanisterHttpChangeAction::AddToValidated(fake_share(456), Some(fake_response(456))),
         ]);
 
         assert!(
@@ -425,10 +432,43 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_canister_http_pool_add_to_validated_without_a_response_to_retain() {
+        let mut pool = CanisterHttpPoolImpl::new(MetricsRegistry::new(), no_op_logger());
+        let response = fake_response(123);
+        let share = fake_share_matching(123, &response);
+        let id = share.clone();
+        let content_hash = ic_types::crypto::crypto_hash(&response);
+
+        let result = pool.apply(vec![CanisterHttpChangeAction::AddToValidated(
+            share.clone(),
+            None,
+        )]);
+
+        let expected_artifact = CanisterHttpResponseArtifact {
+            share: share.clone(),
+            response: None,
+        };
+
+        // Only the share is gossiped, ...
+        assert!(
+            matches!(&result.transmits[0], ArtifactTransmit::Deliver(x) if x.artifact == expected_artifact)
+        );
+        assert!(result.poll_immediately);
+        assert_eq!(result.transmits.len(), 1);
+        assert_eq!(share, pool.lookup_validated(&id).unwrap());
+        // ... and no response is retained, neither to serve to a peer that pulls the
+        // artifact, nor to put into a block.
+        assert_eq!(pool.get(&id).unwrap(), expected_artifact);
+        assert!(pool.get_response_content_by_hash(&content_hash).is_none());
+        assert_eq!(pool.get_response_content_items().count(), 0);
+    }
+
     /// A peer that *pulls* a validated artifact (via
     /// [`ValidatedPoolReader::get`]) must receive the full response for
-    /// non-fully-replicated requests, and must not receive one for fully
-    /// replicated requests.
+    /// non-fully-replicated requests that are still awaiting one, and must not
+    /// receive one for fully replicated requests, nor for requests that have
+    /// already been responded to.
     #[test]
     fn test_get_serves_response_only_for_non_fully_replicated_shares() {
         let mut pool = CanisterHttpPoolImpl::new(MetricsRegistry::new(), no_op_logger());
@@ -451,7 +491,7 @@ mod tests {
         let id = share.clone();
         pool.apply(vec![CanisterHttpChangeAction::AddToValidated(
             share.clone(),
-            response.clone(),
+            Some(response.clone()),
         )]);
         assert!(pool.get(&id).unwrap().response.is_none());
 
@@ -499,6 +539,16 @@ mod tests {
         assert!(pool.get(&id).unwrap().response.is_some());
         pool.apply(vec![CanisterHttpChangeAction::RemoveContent(content_hash)]);
         assert!(pool.get(&id).is_none());
+
+        // Already responded to, locally produced (a receipt-only share): no
+        // response is served on pull, as none was retained.
+        let share = fake_share(6);
+        let id = share.clone();
+        pool.apply(vec![CanisterHttpChangeAction::AddToValidated(
+            share.clone(),
+            None,
+        )]);
+        assert!(pool.get(&id).unwrap().response.is_none());
     }
 
     #[test]
--- a/rs/https_outcalls/consensus/src/pool_manager.rs
+++ b/rs/https_outcalls/consensus/src/pool_manager.rs
@@ -618,7 +618,12 @@ impl CanisterHttpPoolManagerImpl {
                     // Update the set of existing signed requests.
                     existing_signed_requests.insert(key_from_share(share));
                     self.metrics.shares_validated.inc();
-                    Some(CanisterHttpChangeAction::MoveToValidated(share.clone()))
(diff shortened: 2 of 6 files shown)


rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -3834,89 +3858,218 @@ pub mod test {
...
+                    );
+
+                    // While the outcall is still awaiting a response, the share is held
+                    // back, but kept: no change action is produced for it.
+                    let changes =
+                        pool_manager.validate_shares(&awaiting_response, &canister_http_pool);
+                    assert!(changes.is_empty(), "{replication:?}: {changes:?}");
+                    assert_eq!(
+                        fetch_int_gauge(&metrics_registry, "canister_http_shares_deferred"),
+                        Some(1),
```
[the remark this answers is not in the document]
eichhorl: Was a leftover

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -532,44 +547,51 @@ impl CanisterHttpPoolManagerImpl {
...
-                            ));
-                        }
-
-                        if share.content.is_reject() != response.content.is_reject() {
-                            return Some(CanisterHttpChangeAction::HandleInvalid(
-                                share.clone(),
-                                "is_reject does not match the response content".to_string(),
-                            ));
+                        match &artifact.response {
+                            Some(response) => {
```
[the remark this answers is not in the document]
eichhorl: Fixed: Moving shares of delivered contexts to the validated pool now drops their responses.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -532,44 +547,51 @@ impl CanisterHttpPoolManagerImpl {
...
+                                    ));
+                                }
+                            }
+                            // The request has already been answered, so the response is of no
+                            // use to anyone and is not gossiped along with the share.
+                            None if is_delivered => {}
+                            // The peer signed this share when it already saw the request as
+                            // answered, while our own latest state still shows it as awaiting a
+                            // response. Defer until our own state catches up.
+                            None => return None,
```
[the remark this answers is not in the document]
eichhorl: As alluded to in the comment, verifying the signature first would require some kind of cache, such that signatures of deferred shares aren't verified over and over again. Instead we can just keep it as is, since the lifetime of a context is very bounded, which limits the amount of bytes to the time x the bandwidth.

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -170,16 +173,22 @@ impl MutablePool<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
...
                     if let Some(artifact) = self.unvalidated.remove(&share) {
-                        // If there is a response associated with this share, we want to move it to the `content`
-                        // section of the pool, corresponding to valid responses. A validated share carries a
-                        // response exactly for non-fully-replicated requests, whose response must be served
-                        // to peers that pull the artifact.
-                        let served_response = if let Some(content) = artifact.response {
+                        // A response that should not be retained is dropped. This means, it will
+                        // neither be served to other peers nor put into a block by this node.
+                        let content = artifact.response.filter(|_| retain_response);
+                        let served_response = if let Some(content) = content {
```
pierugo-dfinity:
```suggestion
                        let served_response = if let Some(content) = artifact.response && retain_response {
```
Maybe this reads more naturally?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -596,7 +618,12 @@ impl CanisterHttpPoolManagerImpl {
                     // Update the set of existing signed requests.
                     existing_signed_requests.insert(key_from_share(share));
                     self.metrics.shares_validated.inc();
-                    Some(CanisterHttpChangeAction::MoveToValidated(share.clone()))
+                    // The response of an already answered request is dropped rather than
+                    // passed on to peers that pull the artifact.
+                    Some(CanisterHttpChangeAction::MoveToValidated {
+                        share: share.clone(),
+                        retain_response: !is_delivered,
```
pierugo-dfinity: Just a thought to raise a discussion.

Here, independently of whether we received the share with or without a response, we decide not to gossip the response if our local view says that the context is already delivered.

If I understand correctly, this means that, for a given share, some nodes would have the response in their validated pool, while some won't, depending on the timing when they received the artifact. This *shouldn't* lead to misbehaviour since the payload builder would not try to read the response. But validated pools diverging across nodes feels like it is the kind of subtle things that could end up backfiring in the future.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -532,44 +547,51 @@ impl CanisterHttpPoolManagerImpl {
                         }
                     }
                     Replication::NonReplicated(_) | Replication::Flexible { .. } => {
```
pierugo-dfinity, line 563: I assume we do not expect to enter `None if is_delivered => {}` when `context.replication == Replication::NonReplicated`, but this was kept this way for simplicity?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -1320,8 +1348,9 @@ pub mod test {
                     log.clone(),
                 );
 
-                // TEST 1: Non-replicated request artifact is missing the response.
-                // It should be marked as invalid.
+                // TEST 1: Non-replicated request artifact is missing the response, while
+                // the request is still awaiting one. It should be left unvalidated, to be
+                // reconsidered once our state shows the request as responded to.
```
pierugo-dfinity: Related to my previous comment. This feels like it should still be invalid right? When a request is non-replicated, we don't expect async refunds, so I don't think we should expect an artifact without a response

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: A bit unfortunate to return `None` here even if the pricing is `Legacy` (since we could invalidate it instead). But I agree it's simpler to stay agnostic to the pricing version in this component.

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -532,44 +547,51 @@ impl CanisterHttpPoolManagerImpl {
...
+                                if share.content.is_reject() != response.content.is_reject() {
+                                    return Some(CanisterHttpChangeAction::HandleInvalid(
+                                        share.clone(),
+                                        "is_reject does not match the response content".to_string(),
+                                    ));
+                                }
+                            }
+                            // The request has already been answered, so the response is of no
+                            // use to anyone and is not gossiped along with the share.
+                            None if is_delivered => {}
```
pierugo-dfinity, line 604: Maybe we could increment a metric here

rs/interfaces/src/canister_http.rs
```diff
@@ -208,9 +208,14 @@ pub type CanisterHttpPayloadValidationError =
...
-    AddToValidated(CanisterHttpResponseShare, CanisterHttpResponse),
+    // The response is `None` for an outcall that has already been answered.
+    AddToValidated(CanisterHttpResponseShare, Option<CanisterHttpResponse>),
     AddToValidatedAndGossipResponse(CanisterHttpResponseShare, CanisterHttpResponse),
-    MoveToValidated(CanisterHttpResponseShare),
+    MoveToValidated {
+        share: CanisterHttpResponseShare,
+        // `false` for an outcall that has already been answered.
+        retain_response: bool,
+    },
```
pierugo-dfinity: This becomes hard to wrap our heads around. According to Claude, maybe we could refactor these into something like
```rust
enum ResponseDisposition { Publish, KeepLocal, Discard }
```
(naming could also be different) and then
```rust
AddToValidated(CanisterHttpResponseShare, CanisterHttpResponse, ResponseDisposition),
MoveToValidated(CanisterHttpResponseShare, ResponseDisposition),
```
where
- `Publish` would be `AddToValidatedAndGossipResponse`: keep in `content`, push with advert
- `KeepLocal` would be `AddToValidated(_, Some(_))`: keep in `content`, do not push
- `Discard` would be `AddToValidated(_, None)`: do not keep in `content`, do not push

and `ResponseDisposition` could be constructed with a `fn response_disposition(replication, is_delivered)`

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: Does this scenario still make sense here?

rs/https_outcalls/consensus/src/pool_manager.rs
```diff
@@ -1902,7 +1937,10 @@ pub mod test {
 
                 // 5. ASSERT: The artifact should be successfully validated and moved to the validated pool.
                 assert_eq!(changes.len(), 1);
-                assert_matches!(&changes[0], CanisterHttpChangeAction::MoveToValidated(_));
+                assert_matches!(
+                    &changes[0],
+                    CanisterHttpChangeAction::MoveToValidated { .. }
```
mraszyk: would it make sense to assert the value of `retain_response`? also below

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: This would only matter in case of a malicious subnet, right?

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: Claude also suggested: "No metric on the deferred path. Every other early exit in validate_shares records a pool-manager event. A deferral counter would make a peer spamming response-less shares visible in dashboards." which likely refers to the early-exit a few lines below.

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: Our state could be outdated while the state of the peer sending the artifact could already have the response delivered so it omitted the response for that reason (not because of the request being fully-replicated).

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: I suppose the response could also be pruned after delivery even on nodes that initially retained it to eventually converge to the same pools.

rs/artifact_pool/src/canister_http_pool.rs
mraszyk: +1

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: why not?

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: But in this test scenario, we are testing `NonReplicated`, so receiving a share for a request that was already delivered is a sign of malicious behaviour, since it would mean that the delegated node sent a second share for the same request.
You also commented on the same exact lines, so I think we both agree...?

rs/interfaces/src/canister_http.rs
mraszyk: +1

rs/https_outcalls/consensus/src/pool_manager.rs
mraszyk: > You also commented on the same exact lines, so I think we both agree...?

I suppose you refer to https://github.com/dfinity/ic/pull/11546#discussion_r4005670158

which covers the first part of the new separate test test_share_without_response_is_deferred_until_its_context_is_delivered
and thus might not make sense to be there anymore.

mraszyk: > receiving a share for a request that was already delivered is a sign of malicious behaviour, since it would mean that the delegated node sent a second share for the same request

that sounds right and using a flexible request would be more realistic to test that path

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: What I meant is that since the concept of "delivered" responses only makes sense for the `PayAsYouGo` pricing, the `Legacy` one should stay unaffected. And since the protocol enforces that artifacts include a `response` for `NonReplicated` and `Flexible` outcalls in the `Legacy` pricing, not observing one would be a sign of malicious behaviour.

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Yep, got that suggestion too, but I think that doing so would flood the metric, since it would get incremented at each call to `on_state_change` until our state actually advances.

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Related to https://github.com/dfinity/ic/pull/11546#discussion_r4005806020, if a non-replicated request was delivered, then it is not expected that the designated node sends a new share for that same request.

eichhorl: Note that the non-replicated response may timeout before the designated node sends a share, in which case it is expected that it does not gossip the response, if the request was already in progress.

rs/artifact_pool/src/canister_http_pool.rs
eichhorl: Done 👍

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: > I suppose the response could also be pruned after delivery even on nodes that initially retained it to eventually converge to the same pools.

Good catch, I missed this. Implemented now so that we purge the response content as soon as the context moves to `delivered`

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: > it would get incremented at each call to on_state_change until our state actually advances.

that's why I didn't add it in the first place, but I think it's fine either way. Added it now

rs/interfaces/src/canister_http.rs
eichhorl: Implemented as suggested

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Done with the new enum 👍

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: I guess this relates to the comment here: https://github.com/dfinity/ic/pull/11546#discussion_r4004958823, so it is still relevant. Adapted the comment.

mraszyk: the scenario covers the first part of the new separate test test_share_without_response_is_deferred_until_its_context_is_delivered
and thus might not make sense to be there anymore.

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -40,13 +42,27 @@ type ContentCanisterHttpPoolSection =
...
     Include,
-    /// The request is fully replicated: every node recomputes the response
-    /// locally, so it must be excluded from the artifact.
+    /// The response must be excluded from the artifact: either because the
+    /// request is fully replicated, i.e. every node recomputes the response
+    /// locally; or because the request has already been responded to, in which
+    /// case the share is nothing but a receipt and there may be no response in
+    /// the pool at all.
     Exclude,
 }
```
pierugo-dfinity, line 53: As you prefer: I'm not totally against the added abstraction, but now that we have `ResponseDisposition`, I feel like we could drop this and make `ValidatedCanisterHttpPoolSection` hold `ResponseDispotion`s directly. The only call-site that doesn't just call `disposition.into()` is in `MoveToValidated` when `artifact.response` is `None` but the disposition is `Publish`. Unreachable currently, and could happen clearly only if there is a bug.

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -145,45 +161,40 @@ impl MutablePool<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
...
+                        {
                             self.content
                                 .insert(ic_types::crypto::crypto_hash(&content), content);
-                            ServedResponse::Include
+                            disposition.into()
                         } else {
+                            // There is no response to serve, whatever the disposition.
                             ServedResponse::Exclude
                         };
                         self.validated.insert(share, served_response);
```
pierugo-dfinity: Does all of that even matter if we do not push a `ArtifactTransmit::Deliver`? Does P2P advertise anything currently?

rs/https_outcalls/consensus/src/pool_manager.rs
pierugo-dfinity: Ah right!

rs/interfaces/src/canister_http.rs
pierugo-dfinity: Nice

rs/artifact_pool/src/canister_http_pool.rs
pierugo-dfinity, line 200: Maybe this could be simplified to
```suggestion
                        if let Some(content) = artifact.response {
                            self
                                .content
                                .insert(ic_types::crypto::crypto_hash(&content), content);
                        };
                        self.validated.insert(share, disposition.into());
```
and let the purger remove the content later if we see it's not relevant anymore. Would it work?

rs/artifact_pool/src/canister_http_pool.rs
```diff
@@ -145,45 +161,40 @@ impl MutablePool<CanisterHttpResponseArtifact> for CanisterHttpPoolImpl {
...
-                    self.validated.insert(share, ServedResponse::Exclude);
-                    self.content
-                        .insert(ic_types::crypto::crypto_hash(&content), content);
+                    self.validated.insert(share, disposition.into());
+                    // A discarded response is not even retained locally: it can never be
+                    // put into a block.
+                    if disposition != ResponseDisposition::Discard {
+                        self.content
+                            .insert(ic_types::crypto::crypto_hash(&content), content);
+                    }
```
pierugo-dfinity, line 184: Same here

```suggestion
                    self
                        .content
                        .insert(ic_types::crypto::crypto_hash(&content), content);
```
Just always insert the content and let the purger decide whether we want to remove.

The disposition inside `self.validated` is really what matters when delivering the artifact. Whether the content is present for a short amount of time or not doesn't really matter.

This is a quick thought. Not sure if that could actually break something fundamental

rs/https_outcalls/consensus/src/pool_manager.rs
eichhorl: Good point, removed 👍

rs/artifact_pool/src/canister_http_pool.rs
eichhorl: Switched to a different enum with two variants instead

rs/artifact_pool/src/canister_http_pool.rs
eichhorl: Done, it just means that the content will remain in the pool unnecessarily for one round, but that should be fine. It also means the previous enum reduces to two variants

rs/artifact_pool/src/canister_http_pool.rs
eichhorl: Done

rs/artifact_pool/src/canister_http_pool.rs
pierugo-dfinity: Ok yep looks really nice imo!

rs/artifact_pool/src/canister_http_pool.rs
pierugo-dfinity: Nice, looks much simpler.

pierugo-dfinity (approved):

PR #11547, opened 2026-09-11
chore: only attest the artifacts of fully tested RCs

basvandijk:
This guarantees that when RC artifacts have an attestation they were fully tested, i.e. `release-system-tests`, `ci-main`, `guest-os-qualification` and `repro-check` all passed on them.

### DIFF
--- a/.github/workflows/release-testing.yml
+++ b/.github/workflows/release-testing.yml
@@ -66,26 +66,6 @@ jobs:
       commit-sha: ${{ github.sha }}
       release-build: 'true'
 
-  attest-uploads:
-    name: Attest Uploaded Artifacts
-    runs-on: ubuntu-latest
-    needs: [ci-main]
-    # Attestations exist only for the public repo: a build from ic-private
-    # gets attested when its branch is pushed to dfinity/ic as a hotfix-*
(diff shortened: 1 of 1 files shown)


basvandijk:
Closing: #11569 makes this unimplementable.

That PR inlines the attestation into `ci-main.yml`, minting it in the `Attest` step that runs immediately after `Upload artifacts` in `bazel-test-all` and `upload-external-artifacts` — i.e. before `Run Tests` in the very same job, and independently of `release-system-tests`, `guest-os-qualification` and `repro-check`. There is no longer a point in the pipeline where "everything passed" is known and the attestation has not yet been minted, so "the attestation implies a fully tested RC" is not expressible.

That is a deliberate trade. The attestation now means *"this CI run built and uploaded these bytes"*, not *"CI Main was green"* — which also means a flaky test no longer leaves a master commit unattested. Test status stays where it already lives: the release process and `repro-check`.

If we later want a machine-readable "this RC is qualified" signal, it should be its own attestation predicate minted by a job that `needs:` the test jobs, not a gate on the build-provenance one — the consumers (#11324, #11325, #11326) verify build provenance and would not be the right place to check it.

PR #11548, opened 2026-09-11
feat: Fast upgrades: upgrade permit shares, pool, and pool manager

frankdavid:
First step of the Phase-2 rolling-reboot consensus protocol, which adds the data types and the artifact mechanism.

* Add `UpgradePermitAction` (`Request` / `Authorize` / `Return`). `UpgradePermitAuthorizationRequest` is the signed content, `UpgradePermitAuthorizationShare` is the gossiped artifact.
* Add the in-memory `UpgradePermitAuthPoolImpl` which is modeled after other pools.
* Add the `ic-consensus-upgrade` crate with `UpgradePermitAuthPoolManager`. It signs shares for requests in finalized blocks and validates gossiped shares.
* Register the new proto file with the generator and add `UpgradePermitAuthorizationRequest` (signing) and `UpgradePermitAuthorizationShare` (pool-ID hashing) domain separators.

### DIFF
--- a/rs/consensus/upgrade/src/lib.rs
+++ b/rs/consensus/upgrade/src/lib.rs
@@ -1,0 +1,207 @@
+//! The upgrade permit protocol for the Phase-2 rolling GuestOS reboots.
+
+use ic_consensus_utils::crypto::ConsensusCrypto;
+use ic_consensus_utils::membership::Membership;
+use ic_interfaces::upgrade::InvalidUpgradePayloadReason;
+use ic_logger::{ReplicaLogger, warn};
+use ic_types::consensus::UpgradePermitAuthorizationShare;
+use ic_types::{Height, NodeId, RegistryVersion};
+use std::collections::BTreeSet;
+
+pub mod pool_manager;
+
+pub(crate) struct SubnetMembership {
+    /// Current members staying even at the new CUP.
+    pub staying_members: BTreeSet<NodeId>,
+}
+
+impl SubnetMembership {
+    /// Is the node a current member staying even at the new CUP?
+    pub fn staying(&self, node: &NodeId) -> bool {
+        self.staying_members.contains(node)
+    }
+}
+
+/// Subnet membership at the block height and registry version.
+pub(crate) fn subnet_membership(
+    membership: &Membership,
+    block_height: Height,
+    block_registry_version: RegistryVersion,
+    logger: &ReplicaLogger,
+) -> SubnetMembership {
+    let registry_at_height: BTreeSet<NodeId> = membership
+        .get_nodes_at_version(block_registry_version)
+        .map(|nodes| nodes.into_iter().collect())
+        .unwrap_or_default();
+    let current_members: BTreeSet<NodeId> = match membership.get_nodes(block_height) {
+        Ok(nodes) => nodes.into_iter().collect(),
+        Err(e) => {
+            warn!(
+                logger,
+                "upgrade_payload: couldn't determine the committee at height {block_height:?}: {e:?}"
+            );
+            registry_at_height.clone()
+        }
+    };
+    let staying_members = current_members
+        .intersection(&registry_at_height)
+        .cloned()
+        .collect();
+    SubnetMembership { staying_members }
+}
+
+/// Check a share's content, staying signer, and signature. Returns the
+/// signer.
+pub(crate) fn validate_share(
+    share: &UpgradePermitAuthorizationShare,
+    requestor: NodeId,
+    request_height: Height,
+    membership: &SubnetMembership,
+    registry_version: RegistryVersion,
+    crypto: &dyn ConsensusCrypto,
+) -> Result<NodeId, InvalidUpgradePayloadReason> {
+    let signer = share.signature.signer;
+    if share.content.requestor != requestor || share.content.request_height != request_height {
+        return Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare { signer });
+    }
+    if !membership.staying(&signer) {
+        return Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare { signer });
+    }
+    crypto
+        .verify_basic_sig(
+            &share.signature.signature,
+            &share.content,
+            signer,
+            registry_version,
+        )
+        .map_err(|_| InvalidUpgradePayloadReason::AuthorizeInvalidShare { signer })?;
+    Ok(signer)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ic_interfaces_mocks::crypto::MockCrypto;
+    use ic_test_utilities_types::ids::node_test_id;
+    use ic_types::consensus::UpgradePermitAuthorizationRequest;
+    use ic_types::crypto::{BasicSig, BasicSigOf, CryptoError};
+    use ic_types::signature::BasicSignature;
+
+    const REGISTRY_VERSION: RegistryVersion = RegistryVersion::new(1);
+
+    fn share(signer: u64, requestor: u64, request_height: u64) -> UpgradePermitAuthorizationShare {
+        UpgradePermitAuthorizationShare {
+            content: UpgradePermitAuthorizationRequest {
+                requestor: node_test_id(requestor),
+                request_height: Height::from(request_height),
+            },
+            signature: BasicSignature {
+                signature: BasicSigOf::new(BasicSig(vec![])),
+                signer: node_test_id(signer),
+            },
+        }
+    }
+
+    fn membership(staying: &[NodeId]) -> SubnetMembership {
+        SubnetMembership {
+            staying_members: staying.iter().copied().collect(),
+        }
+    }
+
+    fn verifying_crypto() -> MockCrypto {
+        let mut crypto = MockCrypto::new();
+        crypto
+            .expect_verify_basic_sig_upgrade_permit_auth()
+            .returning(|_, _, _, _| Ok(()));
+        crypto
+    }
+
+    #[test]
+    fn test_rejects_requestor_mismatch() {
+        let share = share(2, 3, 10);
+        let membership = membership(&[node_test_id(1), node_test_id(2)]);
+        // No verify expectation: the share is rejected before verification.
+        let result = validate_share(
+            &share,
+            node_test_id(4),
+            Height::from(10),
+            &membership,
+            REGISTRY_VERSION,
+            &MockCrypto::new(),
+        );
+        assert_eq!(
+            result,
+            Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare {
+                signer: node_test_id(2)
+            })
+        );
+    }
+
+    #[test]
+    fn test_rejects_request_height_mismatch() {
+        let share = share(2, 3, 10);
+        let membership = membership(&[node_test_id(1), node_test_id(2)]);
+        let result = validate_share(
+            &share,
+            node_test_id(3),
+            Height::from(11),
+            &membership,
+            REGISTRY_VERSION,
+            &MockCrypto::new(),
+        );
+        assert_eq!(
+            result,
+            Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare {
+                signer: node_test_id(2)
+            })
+        );
+    }
+
+    #[test]
+    fn test_rejects_non_staying_signer() {
+        let share = share(2, 3, 10);
+        let membership = membership(&[node_test_id(1)]);
+        let result = validate_share(
+            &share,
+            node_test_id(3),
+            Height::from(10),
+            &membership,
+            REGISTRY_VERSION,
+            &MockCrypto::new(),
+        );
+        assert_eq!(
+            result,
+            Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare {
+                signer: node_test_id(2)
+            })
+        );
+    }
+
+    #[test]
+    fn test_rejects_invalid_signature() {
+        let share = share(2, 3, 10);
+        let membership = membership(&[node_test_id(1), node_test_id(2)]);
+        let mut crypto = MockCrypto::new();
+        crypto
+            .expect_verify_basic_sig_upgrade_permit_auth()
+            .returning(|_, _, _, _| {
+                Err(CryptoError::TransientInternalError {
+                    internal_error: "boom".to_string(),
+                })
+            });
+        let result = validate_share(
+            &share,
+            node_test_id(3),
+            Height::from(10),
+            &membership,
+            REGISTRY_VERSION,
+            &crypto,
+        );
+        assert_eq!(
+            result,
+            Err(InvalidUpgradePayloadReason::AuthorizeInvalidShare {
+                signer: node_test_id(2)
+            })
+        );
+    }
+}
--- a/rs/consensus/upgrade/src/lib.rs
+++ b/rs/consensus/upgrade/src/lib.rs
@@ -108,14 +108,6 @@ mod tests {
         }
     }
 
-    fn verifying_crypto() -> MockCrypto {
-        let mut crypto = MockCrypto::new();
-        crypto
-            .expect_verify_basic_sig_upgrade_permit_auth()
-            .returning(|_, _, _, _| Ok(()));
-        crypto
-    }
-
     #[test]
     fn test_rejects_requestor_mismatch() {
         let share = share(2, 3, 10);
--- a/rs/consensus/upgrade/src/pool_manager.rs
+++ b/rs/consensus/upgrade/src/pool_manager.rs
@@ -1,0 +1,676 @@
+use std::collections::BTreeSet;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use crate::{SubnetMembership, subnet_membership, validate_share};
+use ic_consensus_utils::crypto::ConsensusCrypto;
+use ic_consensus_utils::membership::Membership;
+use ic_interfaces::consensus_pool::ConsensusBlockCache;
+use ic_interfaces::p2p::consensus::{Bouncer, BouncerFactory, BouncerValue, PoolMutationsProducer};
+use ic_interfaces::upgrade::{
+    UpgradePermitAuthChangeAction, UpgradePermitAuthChangeSet, UpgradePermitAuthPool,
+};
+use ic_logger::{ReplicaLogger, info, warn};
+use ic_types::artifact::IdentifiableArtifact;
+use ic_types::batch::bytes_to_upgrade_payload;
+use ic_types::consensus::{Block, UpgradePermitAuthorizationShare, upgrade::UpgradePermitAction};
+use ic_types::{Height, NodeId};
+use num_traits::SaturatingSub;
+
+/// Shares whose request height falls this many blocks below the finalized
+/// tip are purged from the pool.
+const SHARE_EXPIRY_BLOCKS: Height = Height::new(20);
+
+/// Signs shares for requests in finalized blocks, validates gossiped shares,
+/// and purges expired ones.
+pub struct UpgradePermitAuthPoolManager {
+    node_id: NodeId,
+    crypto: Arc<dyn ConsensusCrypto>,
+    consensus_pool_cache: Arc<dyn ConsensusBlockCache>,
+    membership: Arc<Membership>,
+    /// Requests we've already signed (node, request_height).
+    signed_requests: Mutex<BTreeSet<(NodeId, Height)>>,
+    /// Last finalized height we scanned for requests.
+    last_scanned: Mutex<Height>,
+    logger: ReplicaLogger,
+}
+
+impl UpgradePermitAuthPoolManager {
+    pub fn new(
+        node_id: NodeId,
+        crypto: Arc<dyn ConsensusCrypto>,
+        consensus_pool_cache: Arc<dyn ConsensusBlockCache>,
+        membership: Arc<Membership>,
+        logger: ReplicaLogger,
+    ) -> Self {
+        Self {
+            node_id,
+            crypto,
+            consensus_pool_cache,
+            membership,
+            signed_requests: Mutex::new(BTreeSet::new()),
+            last_scanned: Mutex::new(Height::from(0)),
+            logger,
+        }
+    }
+
+    /// Membership at the finalized block's own height and registry version.
+    fn block_membership(&self, block: &Block) -> SubnetMembership {
+        subnet_membership(
+            &self.membership,
+            block.height,
+            block.context.registry_version,
+            &self.logger,
+        )
+    }
+
+    /// Scan finalized blocks for new `Request` actions and sign an auth share
+    /// for each one we haven't signed yet.
+    fn sign_shares_for_new_requests(&self) -> UpgradePermitAuthChangeSet {
+        let chain = self.consensus_pool_cache.finalized_chain();
+        let tip = chain.tip().height;
+        let mut last = self.last_scanned.lock().unwrap();
+        let start = last.increment();
+        if start > tip {
+            return vec![];
+        }
+        *last = tip;
+
+        let mut signed = self.signed_requests.lock().unwrap();
+        let mut change_set = vec![];
+
+        for height_num in start.get()..=tip.get() {
+            let height = Height::from(height_num);
+            let Ok(block) = chain.get_block_by_height(height) else {
+                continue;
+            };
+            let payload = block.payload.as_ref();
+            if payload.is_summary() {
+                continue;
+            }
+            let upgrade_bytes = &payload.as_data().batch.upgrade;
+            if upgrade_bytes.is_empty() {
+                continue;
+            }
+            let Ok(actions) = bytes_to_upgrade_payload(upgrade_bytes) else {
+                continue;
+            };
+            let membership = self.block_membership(block);
+            if !membership.staying(&self.node_id) {
+                continue;
+            }
+            for action in actions {
+                let UpgradePermitAction::Request(request) = action else {
+                    continue;
+                };
+                let key = (request.requestor, request.request_height);
+                if signed.contains(&key) {
+                    continue;
+                }
+                match self
+                    .crypto
+                    .sign(&request, self.node_id, block.context.registry_version)
+                {
+                    Ok(signature) => {
+                        signed.insert(key);
+                        info!(
+                            self.logger,
+                            "permit_auth: signed share for node {:?} at height {:?}",
+                            request.requestor,
+                            request.request_height
+                        );
+                        change_set.push(UpgradePermitAuthChangeAction::AddToValidated(
+                            UpgradePermitAuthorizationShare {
+                                content: request,
+                                signature,
+                            },
+                        ));
+                    }
+                    Err(e) => {
+                        warn!(
+                            self.logger,
+                            "permit_auth: failed to sign share for node {:?}: {:?}",
+                            request.requestor,
+                            e
+                        );
+                    }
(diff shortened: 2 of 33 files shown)


rs/types/types/src/crypto/hash/tests.rs
```diff
@@ -551,7 +551,7 @@ mod crypto_hash_stability {
         let hash = crypto_hash(&data);
         assert_eq!(
             hex::encode(hash.get_ref().0.as_slice()),
-            "764535296841f3db421a928cfadff3460be406d0182da64034eee623a9a97e99",
+            "c20a87578beb94df369dabfefc30c0d47d170c75d68236aae3b16335c0f21c4a",
```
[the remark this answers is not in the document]
eichhorl: No, it should be fine, please don't do that. The reason this fired seems to be because `test_block` above returns a data block, when it should really be a summary block (which is unaffected by this change).

rs/types/types/src/batch.rs
```diff
@@ -226,6 +232,12 @@ impl BatchPayload {
             bitcoin_adapter_responses: self.self_validating.0,
             query_stats: QueryStatsPayload::deserialize(&self.query_stats)
                 .map_err(IntoMessagesError::QueryStatsPayloadError)?,
+            upgrade: if self.upgrade.is_empty() {
+                Vec::new()
+            } else {
+                bytes_to_upgrade_payload(&self.upgrade)
+                    .map_err(IntoMessagesError::UpgradePayloadError)?
+            },
```
eichhorl: Don't think we need the condition here

rs/types/types/src/batch.rs
```diff
@@ -405,6 +419,7 @@ mod tests {
             canister_http,
             query_stats,
             chain_key,
+            upgrade: _,
```
eichhorl: We should also add a comment here if this test isn't applicable (like for xnet above)

rs/consensus/upgrade/src/pool_manager.rs
```diff
@@ -0,0 +1,676 @@
...
+
+/// Signs shares for requests in finalized blocks, validates gossiped shares,
+/// and purges expired ones.
+pub struct UpgradePermitAuthPoolManager {
+    node_id: NodeId,
+    crypto: Arc<dyn ConsensusCrypto>,
+    consensus_pool_cache: Arc<dyn ConsensusBlockCache>,
+    membership: Arc<Membership>,
+    /// Requests we've already signed (node, request_height).
+    signed_requests: Mutex<BTreeSet<(NodeId, Height)>>,
```
eichhorl: Couldn't we just check the pool if a certain signature already exists?

rs/consensus/upgrade/src/pool_manager.rs
```diff
@@ -0,0 +1,676 @@
...
+                continue;
+            }
+            let upgrade_bytes = &payload.as_data().batch.upgrade;
+            if upgrade_bytes.is_empty() {
+                continue;
+            }
+            let Ok(actions) = bytes_to_upgrade_payload(upgrade_bytes) else {
+                continue;
+            };
+            let membership = self.block_membership(block);
```
eichhorl: Wasn't the plan to also add some data to the replicated state? I think that might simplify everything here. Usually we shouldn't need to look back at the blockchain and deserialize the payloads in the client (see for example the HTTP outcalls or chain-key signer implementations). There, the data flow is the following:
1. A block maker adds some data to a block
2. When the block is executed, this data triggers some changes to the replicated state
3. The client/pool manager creates/validates new shares based on data in the replicated state

rs/types/types/src/crypto/hash.rs
```diff
@@ -222,6 +231,20 @@ impl CryptoHashDomain for EquivocationProof {
...
+    fn domain(&self) -> String {
+        DomainSeparator::UpgradePermitAuthorizationRequest.to_string()
+    }
+}
+
+impl CryptoHashDomain
+    for Signed<UpgradePermitAuthorizationRequest, BasicSignature<UpgradePermitAuthorizationRequest>>
+{
+    fn domain(&self) -> String {
+        DomainSeparator::UpgradePermitAuthorizationShare.to_string()
```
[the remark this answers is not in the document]
frankdavid: 3b60c88699f8b81cd33f1d84b5aed44ac92fa522

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,39 @@
+//! Phase-2 quick-upgrade permit types.
+//!
+//! The permit flow works in three stages:
+//!
+//! 1. **Request**: A block maker includes `UpgradePermitAction::Request` in
+//!    its block when it wants to reboot. Validators check outstanding requests
+//!    the allowed max parallel reboots.
```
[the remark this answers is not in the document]
frankdavid: f2a3e65e3419faab5f9562c1de63cfccade87409

rs/consensus/upgrade/src/pool_manager.rs
frankdavid: How would that work for blocks whose effects are not in the certified state yet?

For example:
- Node 1 requests a permit in its block.
- Node 1's permit request is not applied to the certified state yet, so Node 2 also asks for a permit in the subsequent block.
- If the validators don't check the recent payloads, they accept Node 2's block and now we have 2 outstanding requests which violates the precondition.

eichhorl: Yes, validators have to check the latest certified state and all payloads since then to make these decisions. But the code you are adding in this file is not ran by the validator, it is a separate thread responsible for maintaining the pool (i.e. the interface to P2P), by creating/validating/purging shares.

rs/types/types/src/batch.rs
frankdavid: Actually, we can add an assertion for upgrade.

frankdavid: 433eb1530123f1d20c422db706b461ab1798094e

rs/types/types/src/batch.rs
frankdavid: Done, I also added a wrapper around the payload type, see 85f87cf90314c99669a3aa3c5070fc300b56000d

rs/consensus/upgrade/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+pub(crate) fn subnet_membership(
+    membership: &Membership,
+    block_height: Height,
+    block_registry_version: RegistryVersion,
+    logger: &ReplicaLogger,
+) -> SubnetMembership {
+    let registry_at_height: BTreeSet<NodeId> = membership
+        .get_nodes_at_version(block_registry_version)
+        .map(|nodes| nodes.into_iter().collect())
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
frankdavid: File reverted for this PR, will address in follow-up.

rs/consensus/upgrade/src/pool_manager.rs
```diff
@@ -0,0 +1,676 @@
...
+                    ),
+                ));
+                continue;
+            };
+            let membership = self.block_membership(block);
+            match validate_share(
+                share,
+                share.content.requestor,
+                share.content.request_height,
+                &membership,
```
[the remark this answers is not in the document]
frankdavid: File reverted for this PR, will address in follow-up.

rs/consensus/upgrade/src/pool_manager.rs
```diff
@@ -0,0 +1,676 @@
...
+use ic_types::{Height, NodeId};
+use num_traits::SaturatingSub;
+
+/// Shares whose request height falls this many blocks below the finalized
+/// tip are purged from the pool.
+const SHARE_EXPIRY_BLOCKS: Height = Height::new(20);
+
+/// Signs shares for requests in finalized blocks, validates gossiped shares,
+/// and purges expired ones.
+pub struct UpgradePermitAuthPoolManager {
```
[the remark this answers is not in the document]
frankdavid: File reverted for this PR, will address in follow-up.

rs/consensus/upgrade/src/pool_manager.rs
```diff
@@ -0,0 +1,676 @@
...
+        *last = tip;
+
+        let mut signed = self.signed_requests.lock().unwrap();
+        let mut change_set = vec![];
+
+        for height_num in start.get()..=tip.get() {
+            let height = Height::from(height_num);
+            let Ok(block) = chain.get_block_by_height(height) else {
+                continue;
+            };
```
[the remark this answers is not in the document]
frankdavid: File reverted for this PR, will address in follow-up.

rs/consensus/upgrade/src/pool_manager.rs
frankdavid: File reverted for this PR, will address in follow-up.

rs/consensus/upgrade/src/pool_manager.rs
frankdavid: Sounds good. I reverted this file for this PR to keep it easier to review, I'll address it in the follow-up PR.

rs/consensus/upgrade/src/pool_manager.rs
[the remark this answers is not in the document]
frankdavid: File reverted, will address in follow-up.

rs/types/types/src/batch.rs
```diff
@@ -177,6 +180,7 @@ pub struct BatchPayload {
     pub canister_http: Vec<u8>,
     pub query_stats: Vec<u8>,
     pub chain_key: Vec<u8>,
+    pub upgrade: Vec<u8>,
```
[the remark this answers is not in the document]
frankdavid: Removed it for now

rs/types/types/src/batch.rs
```diff
@@ -177,6 +179,7 @@ pub struct BatchPayload {
...
```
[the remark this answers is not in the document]
frankdavid: Removed it for now

rs/types/types/src/batch.rs
frankdavid: Actually, I decided to simplify this PR and removed this code completely, I'll create a new PR with the block changes.

frankdavid:
The first version of this PR was way too big, I simplified it. I'll send out the block changes and the pool manager separately.

rs/types/types/src/consensus.rs
```diff
@@ -1766,6 +1767,84 @@ impl ConsensusMessageHashable for EquivocationProof {
...
+impl HasHeight for UpgradePermitAuthorizationRequest {
+    fn height(&self) -> Height {
+        self.request_height
+    }
+}
+
+/// An upgrade permit authorization share is a basic signature share on an
+/// [`UpgradePermitAuthorizationRequest`].
+pub type UpgradePermitAuthorizationShare =
+    Signed<UpgradePermitAuthorizationRequest, BasicSignature<UpgradePermitAuthorizationRequest>>;
```
eichhorl: Alternatively I think this is equivalent
```suggestion
pub type UpgradePermitAuthorizationShare = BasicSigned<UpgradePermitAuthorizationRequest>;
```

rs/types/types/src/consensus.rs
```diff
@@ -1766,6 +1767,84 @@ impl ConsensusMessageHashable for EquivocationProof {
     }
 }
 
+/// UpgradePermitAuthorizationRequest holds the values that are signed in an
+/// upgrade permit authorization share.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize, Serialize)]
+pub struct UpgradePermitAuthorizationRequest {
+    pub requestor: NodeId,
+    pub request_height: Height,
+}
```
eichhorl, line 1776: This file mainly holds the structs for the consensus protocol, so maybe these new types would fit better in the `types/src/consensus/upgrade.rs` submodule?

rs/interfaces/src/crypto.rs
```diff
@@ -73,6 +73,9 @@ pub trait Crypto:
     // IDkgOpeningContent
     + BasicSigner<IDkgOpeningContent>
     + BasicSigVerifier<IDkgOpeningContent>
+    // UpgradePermitAuthorizationRequest
+    + BasicSigner<UpgradePermitAuthorizationRequest>
+    + BasicSigVerifier<UpgradePermitAuthorizationRequest>
```
eichhorl, line 78: Can we move this below `VetKdProtocol`? Otherwise we kind of split the IDKG related traits

rs/types/types/src/artifact.rs
```diff
@@ -246,3 +247,57 @@ pub type IDkgMessageId = IDkgArtifactId;
...
+// -----------------------------------------------------------------------------
+// Upgrade permit authorization artifacts
+
+/// Upgrade permit authorization message identifier carries both a message hash
+/// and a height, used by the upgrade permit auth pool for lookup.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize, Serialize)]
+pub struct UpgradePermitAuthorizationShareId {
+    pub hash: CryptoHashOf<UpgradePermitAuthorizationShare>,
+    pub height: Height,
+}
```
eichhorl: I would also move all of these to `typrs/src/consensus/upgrade.rs`

rs/types/types/src/batch/upgrade.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// action can still be included.
+    pub fn serialize_with_limit(&self, byte_limit: NumBytes) -> Vec<u8> {
+        let mut proto = pb::UpgradePayload::default();
+        let mut remaining = byte_limit.get() as usize;
+        for action in &self.actions {
+            let entry = pb::UpgradeAction::from(action);
+            // One repeated field entry: the key, the varint length, and the
+            // message bytes.
+            let entry_len =
+                1 + encoded_len_varint(entry.encoded_len() as u64) + entry.encoded_len();
```
eichhorl, line 34: There also seems to be `prost::length_delimiter_len`:

```suggestion
            let len = entry.encoded_len();
            let entry_len =
                1 + prost::length_delimiter_len(len) + len;
```

rs/types/types/src/consensus.rs
```diff
@@ -1766,6 +1767,84 @@ impl ConsensusMessageHashable for EquivocationProof {
...
+    Signed<UpgradePermitAuthorizationRequest, BasicSignature<UpgradePermitAuthorizationRequest>>;
+
+impl PbArtifact for UpgradePermitAuthorizationShare {
+    type PbId = pb::UpgradePermitAuthMessageId;
+    type PbIdError = ProxyDecodeError;
+    type PbMessage = pb::UpgradePermitAuthorizationShare;
+    type PbMessageError = ProxyDecodeError;
+}
+
+impl From<&UpgradePermitAuthorizationRequest> for pb::UpgradePermitRequest {
```
eichhorl: Maybe it would make sense to use the same names for both struct variants (also applies to other new types)

rs/types/types/src/batch/upgrade.rs
```diff
@@ -0,0 +1,261 @@
...
+    for signature in signatures {
+        let signature: BasicSignature<UpgradePermitAuthorizationRequest> = signature.try_into()?;
+        if signatures_map
+            .insert(signature.signer, signature.signature)
+            .is_some()
+        {
+            return Err(ProxyDecodeError::DuplicateEntry {
+                key: format!("{:?}", signature.signer),
+                v1: "signature".to_string(),
+                v2: "signature".to_string(),
```
eichhorl, line 133: Doesn't seem very helpful

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,39 @@
...
+pub enum UpgradePermitAction {
+    /// Request permission to reboot. The block maker requests for itself.
+    /// `request_height` is the height of the block containing this request,
+    /// used for timeout tracking.
+    Request(UpgradePermitAuthorizationRequest),
+    /// Authorize a node to reboot — the signed request and the basic
+    /// signatures over it collected from the staying members.
+    Authorize {
+        request: UpgradePermitAuthorizationRequest,
+        signatures: BasicSignatureBatch<UpgradePermitAuthorizationRequest>,
```
eichhorl, line 35: Alternatively this could be one field of type `Signed<UpgradePermitAuthorizationRequest, BasicSignatureBatch<UpgradePermitAuthorizationRequest>>` which would then be compatible with some existing helpers

rs/types/types/src/batch/upgrade.rs
frankdavid: Fixed

rs/types/types/src/consensus.rs
frankdavid: I made the names more consistent, wdyt?

rs/types/types/src/crypto/hash.rs
```diff
@@ -222,6 +231,20 @@ impl CryptoHashDomain for EquivocationProof {
     }
 }
 
+impl CryptoHashDomain for UpgradePermitAuthorizationRequest {
+    fn domain(&self) -> String {
+        DomainSeparator::UpgradePermitAuthorizationRequest.to_string()
+    }
+}
```
eichhorl: Do we even hash the request anywhere, i.e. is this needed?

rs/types/types/src/crypto/hash.rs
```diff
@@ -70,6 +70,15 @@ mod private {
...
     impl CryptoHashDomainSeal for BlockPayload {}
 
+    impl CryptoHashDomainSeal for UpgradePermitAuthorizationRequest {}
+    impl CryptoHashDomainSeal
+        for Signed<
+            UpgradePermitAuthorizationRequest,
+            BasicSignature<UpgradePermitAuthorizationRequest>,
+        >
+    {
+    }
```
eichhorl: Alternatively
```suggestion
    impl CryptoHashDomainSeal for UpgradePermitAuthorizationShare {}
```

rs/types/types/src/crypto/hash.rs
```diff
@@ -222,6 +231,20 @@ impl CryptoHashDomain for EquivocationProof {
...
 
+impl CryptoHashDomain for UpgradePermitAuthorizationRequest {
+    fn domain(&self) -> String {
+        DomainSeparator::UpgradePermitAuthorizationRequest.to_string()
+    }
+}
+
+impl CryptoHashDomain
+    for Signed<UpgradePermitAuthorizationRequest, BasicSignature<UpgradePermitAuthorizationRequest>>
+{
```
eichhorl:
```suggestion
impl CryptoHashDomain for UpgradePermitAuthorizationShare {
```

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// A single action in a block's upgrade payload section. A block may carry
+/// multiple actions (e.g. `Request` for the block maker and `Authorize` for
+/// another node).
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub enum UpgradePermitAction {
+    /// Request permission to reboot. The block maker requests for itself.
+    /// `request_height` is the height of the block containing this request,
+    /// used for timeout tracking.
+    RequestUpgradePermit(UpgradePermitAuthorizationRequest),
```
eichhorl: `UpgradePermitAction::RequestUpgradePermit(UpgradePermitAuthorizationRequest {..})`
Has quite a lot of repetition, so I think here I would have kept the short variant names, but it's also fine.

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
+//! Phase-2 quick-upgrade permit types.
+//!
+//! The permit flow works in three stages:
+//!
+//! 1. **Request**: A block maker includes `UpgradePermitAction::RequestUpgradePermit` in
+//!    its block when it wants to reboot. Validators check outstanding requests
+//!    against the allowed max parallel reboots.
+//!
+//! 2. **Authorize**: After the request block is finalized, each node gossips an
```
eichhorl: Is it necessarily "each node" that gossips a share?

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+}
+
+impl TryFrom<pb::UpgradePermitAuthorizationShareId> for UpgradePermitAuthorizationShareId {
+    type Error = ProxyDecodeError;
+
+    fn try_from(id: pb::UpgradePermitAuthorizationShareId) -> Result<Self, Self::Error> {
+        Ok(Self {
+            hash: CryptoHash(id.hash.clone()).into(),
```
eichhorl: do we need to clone the hash here?

eichhorl (approved):

rs/types/types/src/consensus.rs
eichhorl: LGTM just a minor comment below 👍

rs/types/types/src/consensus/upgrade.rs
frankdavid: Nope, fixed (also above).

rs/types/types/src/consensus/upgrade.rs
frankdavid: Shortened the variant names

rs/types/types/src/crypto/hash.rs
frankdavid: Actually, has the hash is not needed, we only need to sign the request.

rs/types/types/src/consensus/upgrade.rs
frankdavid: Rewrote to "nodes"

rs/types/types/src/batch/upgrade.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub actions: Vec<UpgradePermitAction>,
+}
+
+impl UpgradePayload {
+    /// Serialize this payload into a vector.
+    ///
+    /// This function will drop actions that do not fit to guarantee that the
+    /// payload fits into the `byte_limit`. Smaller actions after a dropped
+    /// action can still be included.
+    pub fn serialize_with_limit(&self, byte_limit: NumBytes) -> Vec<u8> {
```
pierugo-dfinity: Couldn't we reuse `fn iterator_to_bytes`?

rs/types/types/src/batch/upgrade.rs
```diff
@@ -0,0 +1,262 @@
...
+            }
+            remaining -= entry_len;
+            proto.actions.push(entry);
+        }
+        proto.encode_to_vec()
+    }
+
+    /// Deserializes an [`UpgradePayload`]. An empty byte slice yields an empty
+    /// payload.
+    pub fn deserialize(data: &[u8]) -> Result<Self, ProxyDecodeError> {
```
pierugo-dfinity: And `fn slice_to_messages` here

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
+//! Phase-2 quick-upgrade permit types.
+//!
+//! The permit flow works in three stages:
+//!
+//! 1. **Request**: A block maker includes `UpgradePermitAction::RequestPermit` in
+//!    its block when it wants to reboot. Validators check outstanding requests
+//!    against the allowed max parallel reboots.
+//!
+//! 2. **Authorize**: After the request block is finalized, nodes gossip an
```
pierugo-dfinity:
```suggestion
//! 2. **Authorize**: After the request block is executed, nodes gossip an
```
Sorry, just about wording. I would expect nodes would start to gossip such shares once the `RequestPermit` actually ends up in the state?

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+/// multiple actions (e.g. `Request` for the block maker and `Authorize` for
+/// another node).
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub enum UpgradePermitAction {
+    /// Request permission to reboot. The block maker requests for itself.
+    /// `request_height` is the height of the block containing this request,
+    /// used for timeout tracking.
+    RequestPermit(UpgradePermitAuthorizationRequest),
+    /// Authorize a node to reboot — the signed request and the basic
+    /// signatures over it collected from the staying members.
```
pierugo-dfinity: There's no mention of "staying members" in the rest of the file and is not clear what it really means. Maybe "other nodes" is just as fine to stay abstract here?

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+        >,
+    ),
+    /// Release a previously authorized permit (reboot complete).
+    ReturnPermit { node: NodeId },
+}
+
+/// UpgradePermitAuthorizationRequest holds the values that are signed in an
+/// upgrade permit authorization share.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize, Serialize)]
+pub struct UpgradePermitAuthorizationRequest {
```
pierugo-dfinity: Sorry, this is really just my opinion so feel free to discard it. This name feels long and feels like `UpgradePermitRequest` would be sufficient, no? Same for `UpgradePermitAuthorizationShare` -> `UpgradeAuthorizationShare`

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub requestor: NodeId,
+    pub request_height: Height,
```
pierugo-dfinity: I see you're deriving `Ord`. If this struct is used in a `BTreeMap`, it may be interesting to order these fields according to the way we're gonna use the map. For example, if we plan to "purge any request below a certain height", we could take advantage of `PoolSection::extract_keys_below`, but that means first indexing by `request_height`, i.e. put the latter before `requestor`.

rs/types/types/src/consensus/upgrade.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+pub type UpgradePermitAuthorizationShare = BasicSigned<UpgradePermitAuthorizationRequest>;
+
+/// Upgrade permit authorization message identifier carries both a message hash
+/// and a height, used by the upgrade permit auth pool for lookup.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Deserialize, Serialize)]
+pub struct UpgradePermitAuthorizationShareId {
+    pub hash: CryptoHashOf<UpgradePermitAuthorizationShare>,
+    pub height: Height,
```
pierugo-dfinity: Same comment about the fields' ordering.

rs/types/types/src/crypto/hash.rs
```diff
@@ -70,6 +70,8 @@ mod private {
     impl CryptoHashDomainSeal for EquivocationProof {}
     impl CryptoHashDomainSeal for BlockPayload {}
 
+    impl CryptoHashDomainSeal for UpgradePermitAuthorizationShare {}
+
```
pierugo-dfinity: nit: maybe we could move this a bit further down such that it doesn't stand in the middle of consensus artifacts

rs/types/types/src/consensus/upgrade.rs
frankdavid: Ord is actually not needed for this one. I changed the field ordering and removed Ord from this struct.

pierugo-dfinity (approved):

PR #11549, opened 2026-09-11
feat(cketh): read the EIP-7702 delegation of deposit addresses in one eth_call

gregorydemay:
### Why

* Replacing the sweeper delegate contract needs the minter to know, per deposit address, whether the address is delegated and to which contract. The chain holds that answer: the code at a delegated address is the EIP-7702 designator `0xef0100 || delegate`.
* Reading it one `eth_getCode` at a time would cost one consensus outcall per address. The balance scan already reads hundreds of addresses in one deployless `eth_call`.

### What

* A third deployless batcher program next to the ERC-20 and ETH balance batchers. It copies the first 32 bytes of every address' code with `EXTCODECOPY` and returns one word per address.
* The word is classified in Rust as not delegated, delegated to a given contract, or other code. EIP-3541 keeps deployed contracts out of the `0xef` space, so a designator cannot be confused with a contract.
* The shared batch cap is unchanged; the new program is bounded by the returned blob, not by its initcode.
* Unit tests pin the bytecode against a readable opcode listing; anvil tests exercise the read against a bare address, a delegated address and a contract, and a full batch.

Nothing in the minter uses the read yet. First PR of a stack of three: next the sweeps that carry no tuple for an already-delegated address, then the rotation to a newly configured delegate.


rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -179,6 +214,80 @@ pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, Bat
...
+    }
+    Ok(ret
+        .chunks_exact(WORD)
+        .map(|word| classify_code_prefix(word.try_into().expect("BUG: chunk is exactly one word")))
+        .collect())
+}
+
+fn classify_code_prefix(prefix: &[u8; WORD]) -> Delegation {
+    if prefix.iter().all(|byte| *byte == 0) {
+        return Delegation::NotDelegated;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Half right, and the doc is now fixed accordingly — but no `EXTCODESIZE` is needed.

You are right that EIP-3541 only reserves the `0xef` prefix, so in the abstract an all-zero word does not prove an empty account: runtime code of `0x00` (`STOP`) is perfectly deployable and `EXTCODECOPY` would zero-pad it to an all-zero word. The justification the doc gave ("a shorter non-empty code would still carry a non-zero first byte") was simply wrong, and eb6589e1b2 replaces it.

What makes the classification sound here is the input type, not the prefix. `encode_delegation_batch` takes `&[DepositAddress]` — addresses the minter derives from its master threshold-ECDSA public key, i.e. `keccak(pubkey)[12..]`. A contract address is `keccak(rlp(deployer, nonce))[12..]` or the `CREATE2` hash; landing either on a given deposit address is a preimage attack on keccak-256. So no contract can ever be deployed at one of these accounts, and the only code such an account can hold is an EIP-7702 delegation designator installed by the minter itself. An all-zero word therefore does mean "no code" for every address this program is ever given.

Conveying `EXTCODESIZE` would cost a second word per entry (halving `MAX_CALLS_PER_BATCH` against the EIP-170 return bound) to distinguish a case that cannot occur, so it is deliberately not done. Instead the doc now states the real premise and explicitly warns that the `NotDelegated` classification rests on the account being a deposit address and does not generalize to arbitrary addresses — which is exactly the trap you spotted. A regression test for zero-prefixed runtime code would pin behaviour for an input the encoder cannot produce, so none was added.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -109,6 +112,38 @@ pub const ETH_BATCHER_INITCODE: [u8; 78] = [
...
+/// The program is fixed regardless of `n` (only the appended args grow). It was assembled from
+/// the opcode listing in `delegation_initcode_matches_readable_assembly` and validated against a
+/// live anvil node; see `rs/ethereum/cketh/minter/tests/deposit_from_cex.rs`.
+pub const DELEGATION_BATCHER_INITCODE: [u8; 81] = [
+    0x60, 0x20, 0x61, 0x00, 0x51, 0x60, 0x00, 0x39, 0x60, 0x00, 0x60, 0x20, 0x52, 0x5b, 0x60, 0x00,
+    0x51, 0x60, 0x20, 0x51, 0x10, 0x15, 0x61, 0x00, 0x47, 0x57, 0x60, 0x20, 0x60, 0x20, 0x51, 0x60,
+    0x20, 0x02, 0x61, 0x00, 0x71, 0x01, 0x60, 0x40, 0x39, 0x60, 0x20, 0x60, 0x00, 0x60, 0x20, 0x51,
+    0x60, 0x20, 0x02, 0x60, 0x60, 0x01, 0x60, 0x40, 0x51, 0x3c, 0x60, 0x20, 0x51, 0x60, 0x01, 0x01,
+    0x60, 0x20, 0x52, 0x61, 0x00, 0x0d, 0x56, 0x5b, 0x60, 0x00, 0x51, 0x60, 0x20, 0x02, 0x60, 0x60,
+    0xf3,
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Confirmed against anvil: a batch starting with a delegated address failed with `EVM error CreateContractStartingWithEF`. Fixed in 5e053a312e: the program returns one zero word ahead of the per-address prefixes and the decoder strips it, so the returned blob is always deployable under EIP-3541. The red test came first (f67deb30d0: `should_read_delegations_across_addresses` now reads `[delegated, bare, contract]`), and the full-batch anvil test delegates both the first and the last address. The cap derivation is unchanged: the returned blob now allows 767 addresses, still above the shared 764.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -179,6 +226,87 @@ pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, Bat
...
+/// Returns `Err` if the blob length is not exactly `n + 1` words; never panics.
+///
+/// The first classification rests on the account being a deposit address and does not generalize:
+/// [EIP-3541] reserves only the `0xef` prefix, so a contract whose runtime code starts with 32
+/// zero bytes — `STOP` padded out, say — would read as [`Delegation::NotDelegated`] here.
+///
+/// [EIP-3541]: https://eips.ethereum.org/EIPS/eip-3541
+/// [EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
+pub fn decode_delegation_batch(
+    ret: &[u8],
```
gregorydemay: provide a descriptive variable name

gregorydemay: 🤖 Renamed in 2d7f86e29f: `ret` → `returned_blob`, `n` → `address_count`, and the per-word closure argument to `code_prefix`; the doc comment follows. The sibling `decode_balance_batch` keeps its pre-existing names since it is not part of this PR.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -179,6 +226,93 @@ pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, Bat
...
+/// [EIP-3541] reserves only the `0xef` prefix, so a contract whose runtime code starts with 32
+/// zero bytes — `STOP` padded out, say — would read as [`Delegation::NotDelegated`] here.
+///
+/// [EIP-3541]: https://eips.ethereum.org/EIPS/eip-3541
+/// [EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
+pub fn decode_delegation_batch(
+    returned_blob: &[u8],
+    address_count: usize,
+) -> Result<Vec<Delegation>, BatcherDecodeError> {
+    let expected = (DELEGATION_BATCH_LEADING_WORDS + address_count) * WORD;
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Valid, fixed in 87aad23659 (test first in 8ac80ab7ef).

`decode_delegation_batch` now derives the expected length with `checked_add`/`checked_mul` and returns a new `BatcherDecodeError::UnrepresentableLength { entries }` when the count calls for a blob no `usize` can describe, so the "never panics" contract holds in debug and no wrapped length is ever compared in release. The doc comment states the new error case.

New unit test `decode_delegation_unrepresentable_length_is_err` pins both overflow sites: `usize::MAX` (the add) and `usize::MAX / WORD` (the multiply). It panicked with "attempt to add with overflow" before the fix and passes after — `bazel test //rs/ethereum/cketh/minter:lib_tests` is green.

The sibling `decode_balance_batch` computes `n * WORD` the same way, but it is pre-existing code that this PR does not touch, so it is deliberately left alone here and surfaced instead of widened into this diff.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -121,11 +168,14 @@ pub struct BalanceOfCall {
     pub holder: DepositAddress,
 }
 
-/// Error encountered while decoding a balance-batch return blob.
+/// Error encountered while decoding the return blob of a batcher program.
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub enum BatcherDecodeError {
     /// The return blob is not exactly `n` 32-byte words.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Valid, fixed in ff287b34ad.

Widening the enum doc to cover all three batchers left the `WrongLength` doc behind at the ERC-20/ETH shape. It now states the contract both decoders actually honour: one 32-byte word per entry, plus whatever fixed words the program puts ahead of them — the delegation batcher's single leading zero word for the others. `expected` and `got` already carried the concrete byte counts, so only the wording changed; no behaviour, no test affected.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
```diff
@@ -371,7 +371,7 @@ const MIN_DEPOSITS: &[(Address, Erc20Value)] = &[
     ), // ckSepoliaPEPE
 ];
 
-fn call_args(input: Vec<u8>, block: BlockNumber) -> evm_rpc_types::CallArgs {
+pub(crate) fn call_args(input: Vec<u8>, block: BlockNumber) -> evm_rpc_types::CallArgs {
```
mbjorkqvist, line 374: nit: There don't seem to be any new callers to `call_args`, so this could stay private?

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -244,6 +245,112 @@ fn should_read_many_eth_balances_in_a_single_call() {
...
+    let bare = |address_byte: u8| {
+        (
+            DepositAddress::new(Address::new([address_byte; 20])),
+            Delegation::NotDelegated,
+        )
+    };
+    let delegated = |address_byte: u8, delegate_byte: u8| {
+        let address = DepositAddress::new(Address::new([address_byte; 20]));
+        let delegate = Address::new([delegate_byte; 20]);
+        anvil.set_code(address.as_address(), &delegation_designator(&delegate));
```
mbjorkqvist, line 261: :robot: Every delegated fixture here (and in the full-batch test) is installed with `anvil_setCode`, so nothing exercises the batcher against a delegation a real EIP-7702 tuple installed — the only way it happens in production.

The gap is narrow but not purely theoretical: a real authorization produces a *typed* `Bytecode::Eip7702` account entry in revm, while `anvil_setCode` writes raw bytes. Both are specified to yield the 23 designator bytes under `EXTCODECOPY`, but they arrive through different doors and the PR only walks through the one production never uses. Note also that every existing observation of a real delegation is either `eth_getCode` (`assert_delegations_installed_at`) or the `CALL` family *following* the designator (the sweeps succeeding) — the batcher is the first consumer that needs `EXTCODECOPY` to **not** follow it.

Cheapest close: `LiveSetup::assert_delegations_installed_at` (`test_utils/src/live.rs:792`) is the single funnel both `assert_delegations_installed` and `assert_eth_delegations_installed` already pass through, running right after a sweep installed the delegations for real. Collect the addresses (the parameter is a once-consumed `impl IntoIterator`), then alongside the existing `anvil.code` check do one `eth_call_create` with `encode_delegation_batch` and assert `vec![Delegation::Delegated(*delegate); n]`. `eth_call_create` is already `pub`, `DEV_ACCOUNT`/`address_from_hex` are already imported there, and `test_utils` already depends on `ic-cketh-minter`. That covers every sweep test at once for one extra `eth_call` per assertion.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -179,6 +231,99 @@ pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, Bat
...
+        out.extend_from_slice(&left_padded_address(address.as_address()));
+    }
+    out
+}
+
+/// Decode the return blob of [`DELEGATION_BATCHER_INITCODE`] — one leading zero word, then
+/// `address_count x 32` bytes — into `address_count` delegations, in call order.
+///
+/// Each word after the first is the first 32 bytes of an account's code, zero-padded beyond its
+/// size. The
```
mbjorkqvist: nit: This format of this doc comment could be improved (it doesn't seem to be caught by any linter).

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
```diff
@@ -179,6 +231,99 @@ pub fn decode_balance_batch(ret: &[u8], n: usize) -> Result<Vec<Erc20Value>, Bat
...
+/// zero bytes — `STOP` padded out, say — would read as [`Delegation::NotDelegated`] here.
+///
+/// [EIP-3541]: https://eips.ethereum.org/EIPS/eip-3541
+/// [EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
+pub fn decode_delegation_batch(
+    returned_blob: &[u8],
+    address_count: usize,
+) -> Result<Vec<Delegation>, BatcherDecodeError> {
+    let expected = DELEGATION_BATCH_LEADING_WORDS
+        .checked_add(address_count)
```
mbjorkqvist, line 287: nit: This hardening (with checked arithmetic) could also be applied to the `decode_balance_batch` above (not touched by this PR, so fine to leave for a subsequent cleanup PR).

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
```diff
@@ -244,6 +245,112 @@ fn should_read_many_eth_balances_in_a_single_call() {
...
+        delegated(0x55, 0xa2),
+        delegated(0x66, 0xa3),
+        contract(),
+        contract(),
+        contract(),
+    ];
+
+    for batch in addresses_with_expected_delegation
+        .iter()
+        .permutations(BATCH_SIZE)
```
mbjorkqvist: :robot: nit: `permutations(3)` over 9 fixtures is 504 anvil round-trips to cover 27 ordered kind-triples.

Keeping three fixtures per kind but indexing by position gets the same coverage in 27 calls — `kinds[first][0], kinds[second][1], kinds[third][2]` over three nested `0..3` loops, so three-of-a-kind still uses three distinct addresses, and the designator-first case stays covered. Bonus: `permutations` is the only use of `itertools`, so nested loops would also drop the dev-dependency and the `Cargo.toml`/`BUILD.bazel`/`Cargo.lock` lines that came with it.

mbjorkqvist (approved):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Done in 22674c0c0f: `assert_delegations_installed_at` now collects the addresses and, next to the per-address code check, reads them all back through `encode_delegation_batch`/`decode_delegation_batch` in one `eth_call`, asserting `Delegated(delegate)` for each. Every live sweep test goes through it, so designators installed by real EIP-7702 tuples are covered; verified on `should_credit_mixed_erc20_and_eth_deposits_through_one_sweep_per_asset`.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: 🤖 Reflowed in 22674c0c0f.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
gregorydemay: 🤖 The caller is in the next PR of the stack, #11550, where the sweep task builds the same create-style call for its delegation read. Widening it here keeps that PR to its own change; it can be narrowed back if you prefer the widening to travel with its caller.

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: 🤖 Agreed, left for a follow-up as suggested: `decode_balance_batch` is not touched by this PR.

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Left for the author: that test is Greg's own commit (24d06dd5fe), so the trade between the permutation sweep and three indexed loops is his call.

rs/ethereum/cketh/minter/src/balance_scan/mod.rs
mbjorkqvist: Thanks for the pointer! Fine either way!

rs/ethereum/cketh/minter/src/balance_scan/batcher/mod.rs
gregorydemay: Will let Claude drive it 😄

rs/ethereum/cketh/minter/tests/deposit_from_cex.rs
gregorydemay: 🤖 Applied in 3d32fa9372: three nested `0..3` loops pick `kinds[first][0], kinds[second][1], kinds[third][2]`, so the 27 ordered kind triples run with three distinct addresses each, designator-first included; `itertools` and its Cargo.toml, BUILD.bazel and Cargo.lock lines are gone. The two delegation anvil tests pass in 0.4 s.

PR #11550, opened 2026-09-11
feat(cketh): sweep an address already delegated to the sweeper contract without an authorization

gregorydemay:
### Why

* Every sweep re-carries the nonce-0 authorization of each address it touches. On an address already delegated the tuple is skipped by the protocol, but it still costs the full 25'000 gas, since EIP-7702 grants the existing-account refund only after the nonce check.
* With the delegation read of #11549 the minter can see, before it builds a sweep, which addresses already run the configured sweeper contract.

### What

* Before enqueueing, the minter reads the delegation of every queued address in one batched `eth_call` at the same latest block height the balance scan uses. A failed chunk leaves its addresses queued for the next tick.
* An address delegated to the configured sweeper contract gets no authorization. A sweep whose items all carry none is a plain EIP-1559 transaction; one with at least one tuple stays EIP-7702.
* The gas budget counts the authorization cost only for the items that carry a tuple.
* An address holding other code is left out of sweeps and logged. An address delegated to another contract keeps today's nonce-0 tuple, in a branch made explicit for the rotation PR that follows.
* The two anvil tests that expected a stale tuple on the second sweep now expect a tuple-less type-2 sweep.

Second PR of the stack, stacked on #11549.


rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -84,66 +101,137 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
...
         };
 
         assert_eq!(targets.len(), attestation_requests.len());
-        assert_eq!(targets.len(), authorization_requests.len());
+
+        if targets.iter().any(|delegated| {
+            delegated
+                .authorization
+                .as_ref()
+                .is_some_and(|request| request.delegate() != destination)
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Valid finding, fixed in c568540a8d (test first in 2721f3e006).

Confirmed before fixing: with the single queued address read as `Delegated(SWEEPER_CONTRACT)` and the contract swapped during the attestation signing, the tick enqueued `SweepRequest { destination: 0x9999…, items: [… authorization: None] }` — a tuple-less sweep calling a contract the addresses were never read against.

`sweep_delegations` now returns a `DelegatedSweepBatch` carrying the `delegate` it classified against alongside the targets, and `enqueue_sweep` compares `batch.delegate != destination` unconditionally, instead of inspecting the per-target authorizations. That also makes the check independent of whether any target happened to need a tuple, and covers the second case the per-target check missed: two assets in one tick can no longer be classified against different delegates without being noticed.

`AuthorizationRequest::delegate()`, which existed only for the old per-target check, is removed.

New test: `sweep::tests::should_skip_the_tick_when_the_sweeper_contract_changed_since_a_tuple_less_read`. Verified with `bazel test //rs/ethereum/cketh/minter:lib_tests`.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -84,66 +101,134 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
         return;
     };
 
+    let Some(latest_block) = read_state(|s| s.latest_block_height) else {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in 902e043703. `enqueue_pending_sweeps` now returns as soon as the queue is non-empty but no deposit helper is configured, before the gas-fee refresh and before `read_delegations`, so a helper-less tick spends nothing. The check goes through a new `State::deposit_helper_contract()`, which `attestation_request`/`attestation_requests` now also use, so the configuration is read in one place. `should_be_no_op_when_no_deposit_helper_contract` now drives the tick with `stub_rpc_client(vec![])`: any RPC would panic the test with "No available call response", which is the proof that none is made (012f135212 shows it failing that way before the fix). The in-loop `attestation_requests` check stays, because the configuration can still be cleared across the awaits of a later iteration.

rs/ethereum/cketh/minter/src/state.rs
```diff
@@ -308,34 +312,61 @@ impl State {
...
+        };
+        let targets = targets
+            .iter()
+            .filter_map(|target| {
+                let authorization = match delegations.get(&target.address()) {
+                    Some(Delegation::Delegated(installed)) if *installed == delegate => None,
+                    Some(Delegation::NotDelegated) => authorize(target, TransactionNonce::ZERO),
+                    Some(Delegation::Delegated(_another_delegate)) => {
+                        authorize(target, TransactionNonce::ZERO)
+                    }
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Covered in 012f135212 by `should_sweep_an_address_delegated_elsewhere_with_a_nonce_zero_authorization` (`src/sweep/tests.rs`): the read answers `Delegated(ANOTHER_SWEEPER_CONTRACT)` and the test asserts the enqueued item carries `authorization_request(SWEEPER_CONTRACT).signed_with(..)`, i.e. a tuple naming the configured contract on nonce 0, and that the sweep budgets the tuple gas (225_000 instead of the tuple-less 185_000). Dropping or omitting that arm now fails the test.

rs/ethereum/cketh/minter/src/state/automatic_deposits/mod.rs
```diff
@@ -279,6 +280,66 @@ impl AutomaticDeposits {
...
+                    Some(Delegation::NotDelegated) => authorize(target, TransactionNonce::ZERO),
+                    // TODO DEFI-2997: track deposit address nonce. A tuple at nonce zero cannot
+                    // apply to an address already delegated: the address spent that nonce when it
+                    // was delegated, so the protocol skips the tuple and the address keeps the
+                    // other contract. Re-delegating it needs the nonce the address has reached,
+                    // which the minter does not track yet.
+                    Some(Delegation::Delegated(_another_delegate)) => {
+                        authorize(target, TransactionNonce::ZERO)
+                    }
+                    Some(Delegation::Other) | None => {
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Valid as a shape, but deliberately out of scope for this PR, so leaving this open for Greg.

Two points:

1. `Delegation::Other` cannot occur for a deposit address in practice. A deposit address is derived from a public key the minter holds; it is never the result of a deployer nonce or of CREATE2 inputs, so no contract can be deployed at one, and the only code it can ever hold is an EIP-7702 delegation designator. That is documented on `decode_delegation_batch`. The `Other` arm exists so the minter never sweeps on a guess if the impossible ever happened, not because it is expected.

2. The starvation shape you describe is real if it ever did happen, and the agreed follow-up is to drop such entries from the queue with an audit event and a metric, rather than to keep filling the batch past exclusions. That is a separate change and is intentionally not in this PR.

No code change here.

rs/ethereum/cketh/minter/src/test_fixtures.rs
```diff
@@ -271,6 +286,59 @@ pub fn sweeper_contract() -> Address {
     Address::new([0x5e; 20])
 }
 
+/// The block height every fixture pins its reads to, so a state built here can serve the
+/// delegation read the enqueue makes.
+pub const LATEST_BLOCK: BlockNumber = BlockNumber::new(1_000_000);
```
gregorydemay: const should be declared at the beginning of the file after the imports

rs/ethereum/cketh/minter/src/test_fixtures.rs
```diff
@@ -271,6 +286,59 @@ pub fn sweeper_contract() -> Address {
...
+    responses: Vec<Result<MultiRpcResult<Hex>, IcError>>,
+) -> EvmRpcClient<StubRuntime, CandidResponseConverter, DoubleCycles> {
+    let mut runtime = StubRuntime::new();
+    for response in responses {
+        runtime = match response {
+            Ok(result) => runtime.add_stub_response(result),
+            Err(error) => runtime.add_stub_error(error),
+        };
+    }
+    EvmRpcClient::builder(runtime, Principal::anonymous())
```
gregorydemay: It's does not make sense to use `Principal::anonymous` as canister ID for the EVM RPC canister. Can just use the real one `7hfb6-caaaa-aaaar-qadga-cai`

rs/ethereum/cketh/minter/src/test_fixtures.rs
gregorydemay: 🤖 Moved in 2dc1b4129b: `LATEST_BLOCK` now sits right after the imports, with its doc comment.

rs/ethereum/cketh/minter/src/test_fixtures.rs
gregorydemay: 🤖 Done in e012854df4: the stub client is now built with the crate's `EVM_RPC_ID_PRODUCTION` (7hfb6-caaaa-aaaar-qadga-cai) instead of the anonymous principal.

rs/ethereum/cketh/minter/src/test_fixtures.rs
```diff
@@ -145,8 +153,7 @@ pub fn automatic_deposit() -> AutomaticDeposit {
 }
 
 /// An [`AutomaticDeposits`] whose sweep queue holds exactly these funded pairs, all taken by the
-/// one sweep [`create_pending_sweeper_requests`] enqueued for them, returned along with that
-/// request.
+/// one sweep [`enqueue_pending_sweeps`] enqueued for them, returned along with that request.
```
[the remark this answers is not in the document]
gregorydemay: 🤖 Fixed in 597101f766: both fixture doc links now point at `create_pending_sweeper_requests`, the function the fixtures call.

rs/ethereum/cketh/minter/src/sweep/mod.rs
```diff
@@ -84,66 +100,141 @@ pub async fn create_pending_sweeper_requests<R: CanisterRuntime>(runtime: &R) {
...
```
mbjorkqvist, line 103: :robot: Reading delegations at `latest_block_height` introduces a reorg exposure the tuple-carrying design did not have.

`refresh_latest_block_height` maintains this from `BlockTag::Latest`, i.e. the unfinalized head. Before this PR every sweep carried a tuple, so a reorg that undid a delegation was self-healing — the tuple simply re-installed it. Now a delegation read at a height that later reorgs yields a tuple-less sweep against an address with no code, and `sweepEth`/`sweepErc20` is a void external call, so Solidity's `extcodesize` check reverts the **whole batch**. `record_finalized_sweep_transaction` then drops every deposit the sweep named (up to ten) from the queue, each needing a fresh `deposit_eth`/`deposit_erc20`. Same blast radius as the R18 note in #11532, reached by a second route.

Worth contrasting with the balance scan, which reads at the same height and is safe: a stale balance just means `sweepEth` moves whatever is actually there. Delegation is the first read where being wrong reverts rather than under-delivers.

The exposed window is read → *execute*, not read → enqueue. The `batch.delegate != destination` guard closes the enqueue window only; after that the request is immutable and nothing re-validates it. The gap spans `read_delegations` → enqueue → price → sign → send → mine, at `SWEEP_ENQUEUE_INTERVAL` (60s) plus `PROCESS_SWEEPER_TRANSACTIONS_INTERVAL` (6 min), and `resubmit_transactions_batch` extends it indefinitely, since a fee bump goes through `with_price_and_amount`, which leaves the (empty) authorization list alone.

What I think settles it is the contrast with #11554: there, `delegation_nonces` advances only on `FinalizedSweeperTransaction`, gated behind `finalized_transaction_count(sender)` at `BlockTag::Finalized` (`withdraw.rs:531`). The minter's other piece of per-address delegation state is deliberately anchored to finalized chain state so a reorg cannot desync it — and the delegation read is the one input to the same decision that is not. Two halves of one mechanism on different finality assumptions is the part worth reconciling, whichever way you decide. The asymmetry favours the read being the conservative half: a desynced nonce silently skips an address' rotation tuples, while a stale `Delegated` read reverts a whole batch.

The machinery is already there — `s.last_observed_block_number`, maintained at `s.ethereum_block_height` (`Finalized` on mainnet). An address delegated but not yet finalized would read `NotDelegated`, carry a tuple, and have it skipped, costing exactly what today costs. Steady-state addresses still get the full saving; the price is roughly one finality window before it starts.

Three things I checked that do not contribute. Provider disagreement cannot skew the read: `NoReduction` is the strictest strategy in the client (`consistent_result_or_reduce(.., Err(InconsistentResults))`), so a lagging or dishonest provider fails the chunk rather than skewing it — stricter than the balance scan's `AnyOf`, and the right choice here. `latest_block_height` being a `MinByKey` across providers lags the true head, which shrinks the window a little. And #11554's rotation cannot race an in-flight tuple-less sweep: it re-delegates through the address' next sweep rather than a separate transaction, and the sweeper's sequential nonce lane keeps a sweep at nonce `n` ahead of one at `n+1`, which executes against the contract the address still runs since `destination` is frozen with the request.

Not testable in the anvil harness (no reorgs), so this is a design decision to record rather than a test gap. Happy to be told the window is too small to care about — but it would be good to have that written down.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -296,11 +296,12 @@ const SWEEP_GAS_PER_TRANSFER: GasAmount = GasAmount::new(110_000);
...
 /// every tuple, before any of them is looked at.
 ///
-/// Budgeted for every address the sweep touches, since every one of them carries a tuple. A tuple
-/// the EVM skips — the address is already delegated, so the nonce it was signed for no longer
-/// matches — is charged the same 25'000 and refunded 12'500 for an authority the state trie
-/// already holds. That refund lands after execution and so cannot shrink the limit the transaction
-/// had to declare, leaving 25'000 the figure to budget either way. Rounded up as its siblings are.
+/// Budgeted for the authorizations the sweep carries, which are those of the addresses it still has
+/// to delegate. An authorization the EVM skips — another sweep delegated the address in between, so
+/// the nonce it was signed for no longer matches — is charged the same 25'000 and refunded 12'500
```
mbjorkqvist: :robot: This doc says a skipped authorization is "charged the same 25'000 and **refunded 12'500** for an authority the state trie already holds" — which contradicts the PR description, and the description is the one that is right.

EIP-7702 verifies the nonce (step 6) *before* adding the refund (step 7), and "if any of the above steps fail, immediately stop processing that tuple". So a nonce-mismatched tuple never reaches the refund: it costs the full 25'000 net. A *valid* tuple on an existing account is what costs 12'500 net — which is exactly the ~12'700 marginal the demo measures between its EIP-7702 and EIP-1559 columns.

The budgeted number is unaffected (25'000 is charged upfront either way, which is what limit-sizing needs), but a reader trusting this paragraph would conclude the PR saves 12'500 per already-delegated address rather than 25'000 — halving the stated benefit.

Related: nothing measures the shape this PR replaces. The demo's EIP-1559 column is already tuple-less, and the live gas assertions only cover first sweeps, so the headline saving (25'000/address, ~250'000 on a ten-deposit sweep, about 46% of it) is derived from the spec rather than measured. A "second sweep re-sending stale tuples" scenario in `deposit_from_cex_demo.rs` would pin it — and would have caught this paragraph, since the measurement comes out 12'500/address above what it predicts.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -356,7 +352,7 @@ impl SweepRequest {
     }
 
     /// The delegations the sweep installs on the way, one per deposit address it still has to
-    /// delegate. Signed for nonce zero, so a tuple whose delegation is already installed is
+    /// delegate. Signed for nonce zero, so an authorization another sweep's delegation raced is
```
mbjorkqvist, line 358: :robot: Question rather than a request: "which is what makes it a plain EIP-1559 transaction" reads as a consequence, but the fallback is actually **mandatory** — EIP-7702 declares a type-`0x04` transaction with a zero-length authorization list invalid, so a sweep with nothing left to delegate cannot simply send an empty list.

That necessity is stated only on `DelegatingSweep` in `tx/sweep.rs` ("A type-`0x04` transaction with an empty list is invalid, and encoding one traps" — accurate, `Eip7702TransactionRequest::rlp_inner` does assert non-empty). The place the decision is actually taken, `SweepTransaction::new`, has no comment on its `if authorizations.is_empty()` early return, and the enum doc above it argues only sufficiency. The contrast inside that same function is what prompts the question: the subtler chain-id guard right below gets a full explanatory paragraph, while the mandatory branch gets none.

#11554 looks like the natural place for it. `tx/sweep.rs` is in that diff, and it rewrites both this doc and the enum doc above `SweepTransaction::new` — keeping the sufficiency framing in both ("a sweep needs type `0x04` only while it touches an address not yet delegated to the configured contract") without adding the necessity. Since the paragraph is being touched anyway, one clause there ("not an optimisation: [EIP-7702] declares a type-0x04 transaction with an empty authorization list invalid") plus the EIP link would do it. Neither `tx/sweep.rs` nor `tx/eip_7702.rs` cites EIP-7702 anywhere, unlike `batcher/mod.rs` next door, which links [EIP-3541] and [EIP-7702] in exactly this situation.

Behaviour itself is well pinned — `should_sign_a_plain_sweep_as_its_eip1559_transaction` covers empty to type-2, and the CBOR round-trip covers the decode path that re-validates on event replay. It is only the rationale that does not reach the reader.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
```diff
@@ -311,31 +312,26 @@ const SWEEP_GAS_PER_AUTHORIZATION: GasAmount = GasAmount::new(40_000);
...
-    .unwrap_or(u64::MAX);
+    let addresses = items
+        .iter()
+        .map(|authorized| authorized.item.deposit)
+        .collect::<BTreeSet<_>>()
+        .len() as u64;
+    let authorizations = items
+        .iter()
+        .filter(|authorized| authorized.authorization.is_some())
+        .count() as u64;
```
mbjorkqvist, line 326: :robot: nit: `u64::try_from(..).unwrap_or(u64::MAX)` became `as u64` here and above. Lossless on every supported target (`usize` is at most 64 bits), so this is fine — noting only that it runs against the direction of travel in #11549, where the same kind of arithmetic was hardened with `checked_*`.

mbjorkqvist (commented):
Thanks @gregorydemay!

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Right, and confirmed against the EIP text: the refund (step 7) is granted only after the nonce check (step 6), and a tuple failing any step stops there, so a skipped authorization costs the full 25'000. Reworded in fd7daf3c2a; the budgeted 25'000 is unchanged. Not adding a stale-authorization demo scenario: it would measure the shape this PR removes.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Agreed that the necessity was stated only on the private newtype. 2423c52667 says it where the decision is taken, on `SweepTransaction::new`, with the EIP link: the type-0x02 fallback is a requirement of EIP-7702, not a saving.

rs/ethereum/cketh/minter/src/state/transactions/mod.rs
gregorydemay: 🤖 Intentional: the cast simplifies the code and keeps the same behaviour, since `usize` is at most 64 bits on every supported target. Kept as is.

rs/ethereum/cketh/minter/src/sweep/mod.rs
gregorydemay: 🤖 The window is real on paper but the sweeper's own nonce lane closes it, so the read stays at latest. Step by step, with a deposit address `deposit`, the sweeper EOA `sweeper` at nonce 7, and `CkSweeper` the configured delegate:

1. **First deposit swept.** The enqueue reads `deposit` at latest: not delegated. The sweep carries the authorization (`deposit` → `CkSweeper`, nonce 0) and goes out from `sweeper` at nonce 7 as type 0x04. It is mined in a block that is not yet finalized; `deposit` now holds the designator and has nonce 1.
2. **Second deposit enqueued.** The read at latest includes that block: delegated to `CkSweeper`. No authorization. The second sweep goes out from `sweeper` at nonce 8 as type 0x02.
3. **Reorg drops the block holding the first sweep.** `deposit` reverts to no code and nonce 0. The first sweep is back in the mempool, and the minter keeps it alive anyway: the sweeper's latest transaction count is still 7, so resubmission keeps bumping its fee, and a fee bump leaves the authorization list untouched (`with_price_and_amount`).
4. **The feared step.** The second sweep would call `deposit`, find no code, and revert the batch. It cannot get there: it sits at nonce 8 and is not includable until a transaction from `sweeper` at nonce 7 is. Only the minter signs for `sweeper`, so the only candidates at nonce 7 are the first sweep and its fee bumps, all carrying the same authorization. Whichever is mined, the tuple passes every check again (`deposit` has no code and nonce 0) and the delegation is re-installed before the second sweep runs — even if the first sweep's own call reverts, since EIP-7702 does not roll delegations back.
5. **First sweep mined again, then the second.** The second sweep executes against `CkSweeper` code at `deposit`. Once the first sweep finalizes, its receipt moves the per-address counter (#11554) to 1; the counter never moved on the reorged block, so it cannot be left ahead of the chain.

The same ordering covers a rotation in #11554: the rotating sweep and the authorization-less sweep that follows it are on the same lane, in that order. Reading at finalized would buy nothing here and would cost one full authorization (25'000) per address for a finality window after every first sweep. I will record the argument in the spec's delegate-rotation section on #11491.

gregorydemay: 🤖 Correction to the last line: the spec on #11491 stays as it is. The argument above is the record.

mbjorkqvist (approved):

PR #11553, opened 2026-09-11
refactor(recovery): move the reusable parts of subnet splitting into a crate of their own

mraszyk:
The subnet splitting tool carries a number of helpers that have nothing to do with splitting a subnet. This moves them into `rs/recovery/subnet_tools`, a crate of their own, which `subnet_splitting` now depends on, so that any other tool that reshapes the canister id ranges of a subnet — halting it at a CUP, working on the state it halted at, and validating what it produced — can use them instead of copying them.

What moved, all of it out of `rs/recovery/subnet_splitting`:

| Module | Content |
|---|---|
| `agent_helper` | `AgentHelper` and `StateTree`: the state tree the NNS signed, pruned to one subnet, and the subnet public key in it |
| `validation` | `validate_artifacts`: state tree → authentic subnet key → CUP signature → manifest root hash of the state the CUP names |
| `state_tool_helper` | the `state_tool` commands these tools run: computing and verifying a manifest, and splitting one |
| `utils` | reading a CUP, its block time, the state hash of a checkpoint, rendering canister id ranges, finding the expected state hash in a `state-tool split-manifests` output, and printing a dashboard URL and waiting for the operator's confirmation |
| `admin_helper` | the `ic-admin` command builders: halting a subnet at its next CUP height and the three canister migration proposals |
| `steps` | `ReadRegistryStep`, which is generic over what it reads |

The unit tests of the moved modules moved with them, as did the `fake_expected_manifests.data` they read.

`subnet_splitting` keeps what carries the logic of splitting — the layout, the state splitting step, the expected manifest cross-check, the recovery iterator driving the whole thing, and the post-split estimations — and is down to `post_split_estimations`, `subnet_splitting`, `layout`, `steps` and `target_subnet`. Per review, the `ic-admin` command builders and the small utilities that only splitting uses so far moved along with the rest rather than staying behind in files of their own; keeping them together is worth more than having each crate hold exactly what it alone uses.

**No behavior change.** The code moved as it was; the items `subnet_splitting` reaches for across the crate boundary became `pub`, everything else stayed private, and the log and error messages of the CUP validation no longer say "split" and "source subnet", since they now describe a subnet that was halted at a CUP, whichever operation halted it. `state_tool_lib` is visible to `//rs/recovery:__subpackages__` rather than to `subnet_splitting` by name.


rs/recovery/subnet_splitting/src/admin_helper.rs
```diff
@@ -1,16 +1,12 @@
...
 use ic_base_types::SubnetId;
-use ic_recovery::admin_helper::{
-    AdminHelper, CommandHelper, IcAdmin, SSH_READONLY_ACCESS_ARG, SUMMARY_ARG, quote,
-};
+use ic_recovery::admin_helper::{AdminHelper, CommandHelper, IcAdmin, SUMMARY_ARG, quote};
 use ic_registry_routing_table::CanisterIdRange;
+use ic_subnet_tools::admin_helper::{DESTINATION_SUBNET_ARG, SOURCE_SUBNET_ARG};
 
-const SOURCE_SUBNET_ARG: &str = "source-subnet";
-const DESTINATION_SUBNET_ARG: &str = "destination-subnet";
```
pierugo-dfinity, line 10: Why not keep those two in the current subnet-splitting crate?

rs/recovery/subnet_tools/src/admin_helper.rs
```diff
@@ -0,0 +1,88 @@
...
+/// between, and the subnet an `ic-admin` subnet command applies to.
+pub const SOURCE_SUBNET_ARG: &str = "source-subnet";
+pub const DESTINATION_SUBNET_ARG: &str = "destination-subnet";
+pub const SUBNET_ARG: &str = "subnet";
+
+/// Propose to make the subnet halt after reaching the next CUP height, i.e. at
+/// a checkpoint whose state is certified and whose hash the subnet agreed on.
+///
+/// Optionally adds a ssh-readonly-access key to the subnet, which is what the
+/// state is then downloaded with.
```
pierugo-dfinity:
```suggestion
/// Propose to make the subnet halt after reaching the next CUP height.
///
/// Optionally adds a ssh-readonly-access key to the subnet.
```

I think the original comment is more informative by not leaking unnecessary details

rs/recovery/subnet_tools/src/admin_helper.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[test]
+    fn get_halt_subnet_at_cup_height_command_test() {
+        let admin_helper = AdminHelper::new(
+            PathBuf::from(FAKE_IC_ADMIN),
+            Url::try_from(FAKE_NNS_URL).unwrap(),
+            /*neuron_args=*/ None,
+        );
+        let subnet_id = PrincipalId::from_str(FAKE_SUBNET_ID)
+            .map(SubnetId::from)
+            .unwrap();
```
pierugo-dfinity: I think it's fine to duplicate `fake_admin_helper` and `subnet_id_from_str`. They're already duplicated in `rs/recovery/src/admin_helper.rs` and `rs/recovery/subnet_splitting/src/admin_helper.rs`

rs/recovery/subnet_tools/src/cli.rs
```diff
@@ -0,0 +1,22 @@
...
+        Ok(url) => {
+            warn!(logger, "{}", text_to_display);
+            warn!(logger, "{}", url);
+            wait_for_confirmation(logger);
+        }
+        Err(err) => {
+            error!(logger, "Failed to parse url {}: {}", url, err);
+        }
+    }
+}
```
pierugo-dfinity: Why not put this in `utils.rs`?

rs/recovery/subnet_tools/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+//! Helpers shared by the tools that reshape the canister id ranges of a
+//! subnet -- `subnet_splitting` and `subnet_merging`.
+//!
+//! Both tools halt a subnet at a CUP, work on the state it halted at, validate
+//! what they produced against the CUP and the NNS signed state tree, and let an
+//! operator confirm a dashboard before the next proposal. This crate holds the
+//! parts of that which are the same for either operation; what is specific to
+//! splitting or to merging stays in the respective tool.
```
pierugo-dfinity:
```suggestion
```
This becomes obsolete in just a few weeks.

rs/recovery/subnet_splitting/src/admin_helper.rs
pierugo-dfinity: Obsolete if you apply the outer comment

pierugo-dfinity (commented):
I'd like to see if it's possible to avoid too much code fragmentation.
For example, what about moving the remaining functions in `subnet_splitting/{admin_helper,utils}.rs` to `subnet_tools/{admin_helper,utils}.rs` and delete the former files? Those would only be called by the subnet-spiltting crate and not by subnet-merging, but it would avoid many small files scattered around multiple crates.

rs/recovery/subnet_tools/src/lib.rs
mraszyk: Why?

rs/recovery/subnet_splitting/src/admin_helper.rs
mraszyk: Moot now — I applied your top-level suggestion, so `subnet_splitting/admin_helper.rs` is gone entirely and both constants are private to `subnet_tools::admin_helper` again.

rs/recovery/subnet_tools/src/admin_helper.rs
mraszyk: Applied, thanks.

rs/recovery/subnet_tools/src/admin_helper.rs
mraszyk: Done — the two files' test modules are merged now, so `get_halt_subnet_at_cup_height_command_test` uses `fake_admin_helper()` and `subnet_id_from_str()` rather than constructing them inline.

rs/recovery/subnet_tools/src/cli.rs
mraszyk: Moved `print_url_and_ask_for_confirmation` to `utils.rs` and dropped `cli.rs`.

mraszyk:
Done: `subnet_splitting/{admin_helper,utils}.rs` are both gone, their contents (and tests, and `test_data/fake_expected_manifests.data`) merged into the corresponding `subnet_tools` files, and `subnet_tools/cli.rs` folded into `subnet_tools/utils.rs` per the other comment. `subnet_splitting/src/lib.rs` is now just `post_split_estimations`, `subnet_splitting`, `layout`, `steps`, `target_subnet`.

rs/recovery/subnet_splitting/src/admin_helper.rs
pierugo-dfinity: They don't look private to me

rs/recovery/subnet_tools/src/lib.rs
pierugo-dfinity: This describes every function that the crate has. As soon as you add or remove a function, this documentation becomes obsolete.

rs/recovery/subnet_splitting/src/admin_helper.rs
mraszyk: they were pub in rs/recovery/subnet_tools/src/admin_helper.rs at commit 97975b84bd7676456b1ee278ac317565cf9511ba

rs/recovery/subnet_tools/src/lib.rs
mraszyk: Fair point — dropped that paragraph, only the one-line summary of what the crate is for remains.

rs/recovery/subnet_splitting/src/admin_helper.rs
mraszyk: You were right — I had left them `pub` in `subnet_tools::admin_helper`. Nothing outside that module uses them and they were private before the move, so d58cfeaa6c restores the block exactly as it was in `subnet_splitting/src/admin_helper.rs`.

pierugo-dfinity: > they were private before the move

Exactly

pierugo-dfinity (approved):

